PR #1036, opened 2019-01-05
[device_info]first unit test for native iOS, running with XCode (not CLI) only, with coverage

truongsinh:
Staring point for https://github.com/flutter/flutter/issues/10141

screen shot 2019-01-05 at 07 57 12

PR #1052, opened 2019-01-09
[image_picker]ios unit test for `image_picker`

truongsinh:
similar to #1036, help flutter/flutter#10141. This actually can be run with CI and has test coverage, but need additional changes, see https://github.com/truongsinh/flutter-plugins/pull/10 and https://codecov.io/gh/truongsinh/flutter-plugins/tree/b7460f4311a1bda9c9f890f493a07352cd5451dd/packages/image_picker
screen shot 2019-01-09 at 5 55 15 pm


cyanglaz:
Thank you for providing this. Unfortunately, we chose to use ObjC as the language of our plugins. I am closing this PR now, if you'd like to do a similar PR with ObjC that'd be great!

PR #1053, opened 2019-01-09
[cloud_functions]Add region support for cloud functions

Knupper:
Hey,

here my implementation to add region support for cloud_functions (see [#24333).](https://github.com/flutter/flutter/issues/24333)

King regards

Max


Knupper:
I signed it!

Knupper:
Any updates or feedback is appreciated.

packages/cloud_functions/lib/cloud_functions.dart
```diff
@@ -29,16 +29,20 @@ class CloudFunctions {
   /// Executes this Callable HTTPS trigger asynchronously.
   ///
   /// @param functionName The name of the callable function being triggered.
+  /// @param region The region where the callable function being triggered. Default is us-central1.
   /// @param parameters Parameters to be passed to the callable function.
   Future<dynamic> call(
-      {@required String functionName, Map<String, dynamic> parameters}) async {
+      {@required String functionName,
+      String region,
```
slightfoot: I think adding the default value in the parameters makes it more obvious, and then add an assert to the top of the function to guarantee that the region is not null and has been set.

packages/cloud_functions/test/cloud_functions_test.dart
```diff
@@ -32,25 +32,37 @@ void main() {
...
             },
           ),
+          isMethodCall(
+            'CloudFunctions#call',
+            arguments: <String, dynamic>{
+              'functionName': 'buz',
+              'region': 'us-east1',
+              'parameters': null,
+            },
+          ),
```
slightfoot: Would be good to add some expected failure cases. Like not passing a region, or passing an invalid region.

slightfoot:
@Knupper I checked `FirebaseApp` hoping we got a default region for the project through in `google-services.json` but we don't. With Firestore now going multi-region it might be worth combining efforts here if there is a more direct route. 

Other change that might be worth making is that FirebaseApp should be able to be passed in. This might already be in another PR. Might be worth making another PR for that however.

References to how Firestore plugin does it.
https://github.com/flutter/plugins/blob/master/packages/cloud_firestore/lib/src/firestore.dart#L132
https://github.com/flutter/plugins/blob/master/packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java#L80

packages/cloud_functions/test/cloud_functions_test.dart
Knupper: A test case for not passing a region is already there. 

> @Knupper I checked `FirebaseApp` hoping we got a default region for the project through in `google-services.json` but we don't. With Firestore now going multi-region it might be worth combining efforts here if there is a more direct route.
> 
> Other change that might be worth making is that FirebaseApp should be able to be passed in. This might already be in another PR. Might be worth making another PR for that however.
> 
> References to how Firestore plugin does it.
> https://github.com/flutter/plugins/blob/master/packages/cloud_firestore/lib/src/firestore.dart#L132
> https://github.com/flutter/plugins/blob/master/packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java#L80
> https://firebase.google.com/docs/reference/android/com/google/firebase/functions/FirebaseFunctions.html#getInstance%28com.google.firebase.FirebaseApp%2C+java.lang.String%29

Not sure if this should be part of this merge request. All FirebaseApp calls are not implemented yet. So i would move it into a seperate merge request to keep this as simple as possible.

slightfoot:
@kroikie hey Arthur, Needs a second review and I nominate you 😁 If your happy with it, then I leave it for you to merge and release 👍

Knupper:
@slightfoot can be closed, we have region support now :)

slightfoot:
Looks like PR #1210 has superseded this one with support for multiple Apps and Regions. Closing.

dshukertjr:
Wait, so do we have region support for cloud functions plug in now?

PR #1054, opened 2019-01-09
ADDED prefixes to try while in downloads and FIXED "jpg" suffix to ".jpg"

pushangupta:
- Added 3 URI prefixes to downloads directory
- Fixed "jpg" suffix to ".jpg" while in getPathFromRemoteUri() method


pushangupta:
@googlebot I signed it!

PR #1057, opened 2019-01-09
[IAP] Check if the payment processor is available

mklim:
The actual functionality here is extremely minimal. Most of this patch
is establishing the structure of the rest of the plugin code to follow.

1. Add a generic IAP plugin interface with platform-specific
   implementations (`*_connection.dart`.)
2. Add platform specific wrappers (`*_wrappers.dart`).
3. Add dart end to end tests covering the sample app and unit tests
   covering the native platform code. These all need to be run manually
   currently.


mklim:
I'm happy to break this up into smaller PRs if that would make it easier to review, let me know. I think there's a couple ways this could be split and it's a bit on the large side right now.

packages/in_app_purchase/lib/src/google_play_connection.dart
```diff
@@ -0,0 +1,33 @@
...
+    return await _billingClient.isReady;
+  }
+
+  @override
+  Future<bool> connect() async {
+    final BillingResponse responseCode = await _billingClient.startConnection(
+        onBillingServiceDisconnected: () {});
+    final bool finishedSuccessfully = responseCode == BillingResponse.OK;
+    if (!finishedSuccessfully) {
+      print('Failed to connect to Play Billing. Response $responseCode');
```
amirh: We probably want to expose some information about why we couldn't connect to the appm e.g if there's no internet connection (which I assume is a requirement), the app might want to say something to the user.

packages/in_app_purchase/lib/src/in_app_purchase_connection.dart
```diff
@@ -0,0 +1,35 @@
+import 'dart:async';
+import 'dart:io';
+import 'app_store_connection.dart';
+import 'google_play_connection.dart';
+
+/// Basic generic API for making in app purchases across multiple platforms.
+abstract class InAppPurchaseConnection {
+  /// Returns true if the payment platform is ready and available.
+  Future<bool> isAvailable();
```
amirh: Suggestion/Question:
I imagine this is something you'll use from Widget build methods to decide e.g whether to show a "buy" button or not, a synchronous API would be easier to use in such cases.
Since it is anyway impossible to guarantee that the value is not stale by the time you get it (as the message channel is async), would it be possible / make sense to maintain an availability status on the dart side that's updated by callbacks and make this synchronous?

[I know nothing about IAP so don't know if this makes sense, really just asking here]

packages/in_app_purchase/lib/src/google_play_connection.dart
```diff
@@ -0,0 +1,33 @@
...
+class GooglePlayConnection implements InAppPurchaseConnection {
+  BillingClient _billingClient = BillingClient();
+
+  @override
+  Future<bool> isAvailable() async {
+    return await _billingClient.isReady;
+  }
+
+  @override
+  Future<bool> connect() async {
```
amirh: The docs on `InAppPurchaseConnection` says that this does nothing if the user is already connected, if that is the wanted behavior we probably shouldn't call `startConnection` if we're connected?

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -1,20 +1,34 @@
...
+  if (!self) {
+    return nil;
+  }
+  return self;
+}
+
 - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
-  if ([@"getPlatformVersion" isEqualToString:call.method]) {
-    result([@"iOS " stringByAppendingString:[[UIDevice currentDevice] systemVersion]]);
+  if ([@"-[SKPaymentQueue canMakePayments:]" isEqualToString:call.method]) {
```
amirh: Is this going to be used?

packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
```diff
@@ -0,0 +1,92 @@
...
+  public InAppPurchasePlugin(Context context, MethodChannel channel) {
+    billingClient =
+        BillingClient.newBuilder(context)
+            .setListener(
+                new PurchasesUpdatedListener() {
+                  @Override
+                  public void onPurchasesUpdated(
+                      int responseCode, @Nullable List<Purchase> purchases) {}
+                })
+            .build();
```
amirh: Just making sure - this have no side effects right?

packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
```diff
@@ -0,0 +1,92 @@
...
+    this.billingClient = billingClient;
+    this.channel = channel;
+  }
+
+  private void startConnection(final int handle, final Result result) {
+    billingClient.startConnection(
+        new BillingClientStateListener() {
+          @Override
+          public void onBillingSetupFinished(int responseCode) {
+            result.success(responseCode);
```
amirh: It wasn't immediately clear to me how do we fail this.
If we keep it this way, lets add a comment mentioning that we pass error as "success" as well.

packages/in_app_purchase/example/android/app/src/test/java/io/flutter/plugins/inapppurchase/InAppPurchasePluginTest.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.mockito.Spy;
+
+public class InAppPurchasePluginTest {
```
amirh: Woohoo!

packages/in_app_purchase/example/ios/in_app_purchase_pluginTests/in_app_purchase_pluginTests.m
```diff
@@ -0,0 +1,36 @@
+#import <XCTest/XCTest.h>
+#import "InAppPurchasePlugin.h"
+
+@interface in_app_purchase_pluginTests : XCTestCase
+
+@end
+
+@implementation in_app_purchase_pluginTests
```
amirh: Woohoo 2 !

packages/in_app_purchase/lib/src/in_app_purchase_connection.dart
mklim: I'd want to set up long polling to make sure it didn't go totally stale on the Android song. Maybe I could add it as a configurable option? I'd rather do this in a followup PR. For iOS this is always stable, so I could cache it.

packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
```diff
@@ -0,0 +1,105 @@
...
+  private void startConnection(final int handle, final Result result) {
+    if (billingClient == null) {
+      billingClient = buildBillingClient(context);
+    }
+
+    billingClient.startConnection(
+        new BillingClientStateListener() {
+          @Override
+          public void onBillingSetupFinished(int responseCode) {
+            // Consider finishing at all a "success", leave it to the Dart side to validate the responseCode.
```
amirh: I'm not sure I understand what is meant by `Consider finishing at all a "success"`

packages/in_app_purchase/lib/src/billing_client_wrappers.dart
```diff
@@ -0,0 +1,82 @@
...
+/// converted to futures where appropriate.
+class BillingClient {
+  BillingClient() {
+    _channel.setMethodCallHandler(_callHandler);
+  }
+  static const MethodChannel _channel =
+      MethodChannel('plugins.flutter.io/in_app_purchase');
+  List<Map<String, Function>> _callbacks = <Map<String, Function>>[];
+
+  /// Wraps `BillingClient#isReady()`.
```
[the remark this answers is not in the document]
amirh: ok.
[nit: lets link to the Javadoc at least to make it easier to jump when you look at these docs]

packages/in_app_purchase/lib/src/billing_client_wrappers.dart
```diff
@@ -0,0 +1,82 @@
...
+  Future<bool> get isReady async {
+    return await _channel.invokeMethod('BillingClient#isReady()');
+  }
+
+  /// `BillingClient#startConnection(BillingClientStateListener)`
+  ///
+  /// [onBillingServiceConnected] has been converted from a callback parameter
+  /// to the Future result returned by this function. This returns the
+  /// `BillingClient.BillingResponse` `responseCode` of the connection result.
+  Future<BillingResponse> startConnection(
```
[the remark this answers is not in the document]
amirh: IIUC there is no reason for 2 `startConnection`'s to be in flight simultaneously?
I'm not sure I understand why do we need this list of callback maps and not just callback set for the current connection?

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -12,45 +11,59 @@ class MyApp extends StatefulWidget {
...
+          title: const Text('IAP Example'),
         ),
         body: Center(
-          child: Text('Running on: $_platformVersion\n'),
-        ),
+            child: FutureBuilder<List<Widget>>(
+                future: buildStorefront(),
+                builder: (BuildContext context,
+                    AsyncSnapshot<List<Widget>> snapshot) {
+                  if (!snapshot.hasData || snapshot.error != null) {
```
amirh: Shouldn't we show some error message on error and not 'Trying to connect...'?

packages/in_app_purchase/lib/src/in_app_purchase_connection.dart
mklim: I don't think so, but I can investigate for a followup. The only guarantee for `BillingClientStateListener` is that it's talking about setup being complete and successful, specifically. Practically I think it's probably tied to `isReady` but I'm not totally sure. I'd also still need at least one async call here to set it up.

packages/in_app_purchase/lib/src/billing_client_wrappers.dart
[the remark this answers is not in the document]
amirh: ok

packages/in_app_purchase/lib/src/in_app_purchase_connection.dart
amirh: ok, lets try to verify that before we first publish (we should be able to return false before we set up the callbacks so this could be synchronous)

mklim:
> LGTM but I still think we should include Dart unit tests as part of this PR as future work will probably need that unit test skeleton anyway, and as in general we try to avoid submitting untested code.

OK, done. I was going to rearrange the dart files into a more fine grained structure in a followup PR, but I think with adding in dart unit tests it also became important to make sure the structure here made more sense now. d14bac6 moves the dart files around and the new tests are in 1f61bf2.

packages/in_app_purchase/test/fake_platform_views_controller.dart
```diff
@@ -0,0 +1,28 @@
+import 'dart:async';
+import 'package:flutter/services.dart';
+
+class FakePlatformViewsController {
```
amirh: I'd expect a "FakePlatformViewsController" to be a fake implementation of the engine's [PlatformViewsController](https://github.com/flutter/engine/blob/master/shell/platform/android/io/flutter/plugin/platform/PlatformViewsController.java)

packages/in_app_purchase/test/billing_client_wrappers/billing_client_wrapper_test.dart
```diff
@@ -0,0 +1,70 @@
...
+import '../fake_platform_views_controller.dart';
+
+void main() {
+  final FakePlatformViewsController fakePlatformViewsController =
+      FakePlatformViewsController();
+  BillingClient billingClient;
+
+  setUpAll(() {
+    Channel.override = SystemChannels.platform_views;
+    SystemChannels.platform_views.setMockMethodCallHandler(
```
amirh: Why override the platform_views system channel? is it used by this plugin at all?

Can we not just set a mock handler for the `plugins.flutter.io/in_app_purchase` channel? and avoid the indirection added with the Channel class?

mklim:
Following up in #1082.

PR #1046, opened 2019-01-07
Fix Crash When StartPreview Error

bcollins19:
There is a crash due to Result already being submitted when startPreview() throws an error. This is because do result.error() in the catch block and then continue on to do result.success() later. If there is an error we should return early.

PR #1060, opened 2019-01-10
Bump camera plugin version and update changelog.

amirh:
Follow-up to https://github.com/flutter/plugins/pull/1046

### DIFF
--- a/packages/camera/pubspec.yaml
+++ b/packages/camera/pubspec.yaml
@@ -1,7 +1,7 @@
 name: camera
 description: A Flutter plugin for getting information about and controlling the
   camera on Android and iOS. Supports previewing the camera feed and capturing images.
-version: 0.2.9
+version: 0.2.9+1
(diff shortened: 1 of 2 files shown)


packages/camera/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: camera
 description: A Flutter plugin for getting information about and controlling the
   camera on Android and iOS. Supports previewing the camera feed and capturing images.
-version: 0.2.9
+version: 0.2.9+1
```
goderbauer: Why not just 0.2.10?

amirh: Following the guideline at https://www.dartlang.org/tools/pub/versioning#semantic-versions
it says that prior to 1.0.0: `...going to 0.1.2+1 indicates a change that doesn’t affect the public API.`

PR #1062, opened 2019-01-10
Url launcher ios universallinksonly

cyanglaz:
Added a setting `universalLinksOnly` for IOS with version >= 10.0

With setting this to true, user can now force the URL Launcher to launch universal links only when the app for the universal link is installed. If the app is not installed, a Platform exception will be thrown. By catching the exception, other actions can be performed. E.g. open the URL destination in a Safari View Controller. 
This is an improvement requested in https://github.com/flutter/flutter/issues/25991

Updated the version to 4.1.0 because there is a breaking change updating `launch` to return bool from void

### DIFF
--- a/packages/url_launcher/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncherPlugin.java
+++ b/packages/url_launcher/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncherPlugin.java
@@ -43,28 +43,9 @@ public class UrlLauncherPlugin implements MethodCallHandler {
     if (call.method.equals("canLaunch")) {
       canLaunch(url, result);
     } else if (call.method.equals("launch")) {
-      Intent launchIntent;
-      boolean useWebView = call.argument("useWebView");
-      boolean enableJavaScript = call.argument("enableJavaScript");
-      Activity activity = mRegistrar.activity();
-      if (activity == null) {
-        result.error("NO_ACTIVITY", "Launching a URL requires a foreground activity.", null);
-        return;
-      }
-      if (useWebView) {
-        launchIntent = new Intent(activity, WebViewActivity.class);
-        launchIntent.putExtra("url", url);
-        launchIntent.putExtra("enableJavaScript", enableJavaScript);
-      } else {
-        launchIntent = new Intent(Intent.ACTION_VIEW);
-        launchIntent.setData(Uri.parse(url));
-      }
-      activity.startActivity(launchIntent);
-      result.success(null);
+      launch(call, result, url);
     } else if (call.method.equals("closeWebView")) {
-      Intent intent = new Intent("close");
-      mRegistrar.context().sendBroadcast(intent);
-      result.success(null);
+      closeWebView(result);
     } else {
       result.notImplemented();
     }
@@ -81,6 +62,33 @@ public class UrlLauncherPlugin implements MethodCallHandler {
             && !"{com.android.fallback/com.android.fallback.Fallback}"
                 .equals(componentName.toShortString());
     result.success(canLaunch);
+  }
+
+  private void launch(MethodCall call, Result result, String url) {
+    Intent launchIntent;
+    boolean useWebView = call.argument("useWebView");
+    boolean enableJavaScript = call.argument("enableJavaScript");
+    Activity activity = mRegistrar.activity();
+    if (activity == null) {
+      result.error("NO_ACTIVITY", "Launching a URL requires a foreground activity.", null);
+      return;
+    }
+    if (useWebView) {
+      launchIntent = new Intent(activity, WebViewActivity.class);
+      launchIntent.putExtra("url", url);
+      launchIntent.putExtra("enableJavaScript", enableJavaScript);
+    } else {
+      launchIntent = new Intent(Intent.ACTION_VIEW);
+      launchIntent.setData(Uri.parse(url));
+    }
+    activity.startActivity(launchIntent);
+    result.success(true);
+  }
+
+  private void closeWebView(Result result) {
+    Intent intent = new Intent("close");
+    mRegistrar.context().sendBroadcast(intent);
+    result.success(null);
   }
 
   /*  Launches WebView activity */
--- a/packages/url_launcher/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncherPlugin.java
+++ b/packages/url_launcher/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncherPlugin.java
@@ -43,9 +43,28 @@ public class UrlLauncherPlugin implements MethodCallHandler {
     if (call.method.equals("canLaunch")) {
       canLaunch(url, result);
     } else if (call.method.equals("launch")) {
-      launch(call, result, url);
+      Intent launchIntent;
+      boolean useWebView = call.argument("useWebView");
+      boolean enableJavaScript = call.argument("enableJavaScript");
+      Activity activity = mRegistrar.activity();
+      if (activity == null) {
+        result.error("NO_ACTIVITY", "Launching a URL requires a foreground activity.", null);
+        return;
+      }
+      if (useWebView) {
+        launchIntent = new Intent(activity, WebViewActivity.class);
+        launchIntent.putExtra("url", url);
+        launchIntent.putExtra("enableJavaScript", enableJavaScript);
+      } else {
+        launchIntent = new Intent(Intent.ACTION_VIEW);
+        launchIntent.setData(Uri.parse(url));
+      }
+      activity.startActivity(launchIntent);
+      result.success(true);
     } else if (call.method.equals("closeWebView")) {
-      closeWebView(result);
+      Intent intent = new Intent("close");
+      mRegistrar.context().sendBroadcast(intent);
+      result.success(null);
     } else {
       result.notImplemented();
     }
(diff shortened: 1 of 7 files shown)


packages/url_launcher/ios/Classes/UrlLauncherPlugin.m
```diff
@@ -96,22 +101,23 @@ - (BOOL)canLaunchURL:(NSString *)urlString {
   return [application canOpenURL:url];
 }
 
-- (void)launchURL:(NSString *)urlString result:(FlutterResult)result {
+- (void)launchURL:(NSString *)urlString options:(NSDictionary<UIApplicationOpenExternalURLOptionsKey, id> *)options result:(FlutterResult)result {
   NSURL *url = [NSURL URLWithString:urlString];
   UIApplication *application = [UIApplication sharedApplication];
-  if ([application respondsToSelector:@selector(openURL:options:completionHandler:)]) {
```
amirh: I'm missing context on the fix that added this check, is this completely equivalent to checking for iOS 10.0 and above?

packages/url_launcher/ios/Classes/UrlLauncherPlugin.m
```diff
@@ -81,7 +81,12 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
     if (useSafariVC.boolValue) {
       [self launchURLInVC:url result:result];
     } else {
-      [self launchURL:url result:result];
+        NSDictionary *options = nil;
+        if (@available(iOS 10.0, *)) {
```
amirh: nit: I'd keep `handleMethodCall` focused on routing the calls and move this logic to launchUrl

packages/url_launcher/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Flutter plugin for launching a URL on Android and iOS. Supports
   web, phone, SMS, and email schemes.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/url_launcher
-version: 4.0.3
+version: 4.0.4
```
amirh: This is a new feature so we should bump the minor number.

packages/url_launcher/lib/url_launcher.dart
```diff
@@ -17,13 +17,24 @@ const MethodChannel _channel = MethodChannel('plugins.flutter.io/url_launcher');
...
+/// Set this to false if you want to use the cookies/context of the main browser
+/// of the app (such as SSO flows). This setting will nullify [universalLinksOnly]
+/// and will always launch a web content in the built-in Safari View Controller regardless
+/// if the url is a universal link or not.
+///
+/// [universalLinksOnly] is only used in iOS with iOS version >= 10.0. This setting is only validated
+/// when [forceSafariVC] is set to false. The default value of this setting is false.
+/// By default (when unset), the launcher will either launch the url in a browser (when the
+/// url is not a universal link), or launch the respective native app content (when
+/// the url is a universal link). If setting to true, the launcher will only launch the
```
amirh: nit: s/If setting to true/When set to true/

packages/url_launcher/lib/url_launcher.dart
```diff
@@ -17,13 +17,24 @@ const MethodChannel _channel = MethodChannel('plugins.flutter.io/url_launcher');
...
+/// the content if the url is a universal link and the respective app for the universal
```
amirh: nit: extra 'the'

packages/url_launcher/example/lib/main.dart
```diff
@@ -65,6 +63,20 @@ class _MyHomePageState extends State<MyHomePage> {
     }
   }
 
+  Future<void> _launchUniversalLinkIos(String url) async {
+    try {
+      if (await canLaunch('https://youtube.com')) {
+        await launch('https://youtube.com',
+          forceSafariVC: false, universalLinksOnly: true);
+      }
+    } on Exception {
```
amirh: I'm assuming you do this as you assume that it failed because the universal link is not supported? but we probably shouldn't re-try if something else causes the error.

The flow feels a little weird anyway (that we need to catch an exception to know whether we can launch a universal only link), is it possible to add some method that tells the app whether the universal link is supported?

packages/url_launcher/example/lib/main.dart
```diff
@@ -129,6 +141,13 @@ class _MyHomePageState extends State<MyHomePage> {
                   }),
               child: const Text('Launch in app(JavaScript ON)'),
             ),
+            RaisedButton(
+              onPressed: () => setState(() {
+                    _launched = _launchUniversalLinkIos(toLaunch);
+                  }),
+              child: const Text(
+                  'Universal link launch native App or fallback to Safari View Controller(Youtube)'),
```
amirh: nit: Launch a universal link in a native app, fallback to Safari.

packages/url_launcher/ios/Classes/UrlLauncherPlugin.m
cyanglaz: Yes it is equivalent; however, in recent updates xcode starts to give warning when api gets called without a version check, so this update is to suppress the warning

packages/url_launcher/example/lib/main.dart
cyanglaz: On IOS, there is no way to tells if the universal link is supported beside actually trying to launch the universal link with the universalLinksOnly flag to be true. I do agree this is a little weird. I will update the Future to return a boolean.

packages/url_launcher/ios/Classes/UrlLauncherPlugin.m
```diff
@@ -6,52 +6,66 @@
 
 #import "UrlLauncherPlugin.h"
 
+API_AVAILABLE(ios(9.0))
 @interface FLTUrlLaunchSession : NSObject <SFSafariViewControllerDelegate>
-@property(strong) SFSafariViewController *safari;
+
+@property(copy, nonatomic) FlutterResult flutterResult;
+@property(strong, nonatomic) NSURL *url;
+@property(strong, nonatomic) SFSafariViewController *safari;
```
amirh: would be easier for me to review if we do the refactoring in a separate PR

packages/url_launcher/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncherPlugin.java
```diff
@@ -43,28 +43,9 @@ public void onMethodCall(MethodCall call, Result result) {
     if (call.method.equals("canLaunch")) {
       canLaunch(url, result);
     } else if (call.method.equals("launch")) {
-      Intent launchIntent;
```
amirh: Thanks for cleaning up here!
I'd rather keep it to a separate PR to reduce noise from this one (and also, I think looking on these lines and seeing 'universal ios links' in the blame line will be a bit confusing)

packages/url_launcher/lib/url_launcher.dart
```diff
@@ -38,11 +49,12 @@ const MethodChannel _channel = MethodChannel('plugins.flutter.io/url_launcher');
 /// [statusBarBrightness] Sets the status bar brightness of the application
 /// after opening a link on iOS. Does nothing if no value is passed. This does
 /// not handle reseting the previous status bar style.
-Future<void> launch(
+Future<bool> launch(
```
amirh: we need to describe the semantics of the return value in the Dartdoc.

packages/url_launcher/example/lib/main.dart
```diff
@@ -65,6 +65,22 @@ class _MyHomePageState extends State<MyHomePage> {
     }
   }
 
+  Future<void> _launchUniversalLinkIos(String url) async {
+    if (await canLaunch('https://youtube.com')) {
+      final bool canLaunchInNativeApp = await launch(
```
amirh: nit: s/canLaunchInNativeApp/nativeAppLaunchSucceeded/

packages/url_launcher/lib/url_launcher.dart
```diff
@@ -38,11 +49,15 @@ const MethodChannel _channel = MethodChannel('plugins.flutter.io/url_launcher');
 /// [statusBarBrightness] Sets the status bar brightness of the application
 /// after opening a link on iOS. Does nothing if no value is passed. This does
 /// not handle reseting the previous status bar style.
-Future<void> launch(
+///
+/// Return a Future contains bool value. If ture, launch url is successful; if false,
```
amirh: nit:typos

amirh: This should explain that false is only returned if a universal link cannot be opened, otherwise it's a little confusing when you look at the "retry" logic.

amirh: nit: No need to say "Return a Future of a bool" the signature already says that.

packages/url_launcher/lib/url_launcher.dart
```diff
@@ -17,32 +17,47 @@ const MethodChannel _channel = MethodChannel('plugins.flutter.io/url_launcher');
...
 /// this will throw a [PlatformException].
 ///
 /// [statusBarBrightness] Sets the status bar brightness of the application
 /// after opening a link on iOS. Does nothing if no value is passed. This does
-/// not handle reseting the previous status bar style.
-Future<void> launch(
+/// not handle resetting the previous status bar style.
+///
+/// Returns true if launch url is successful; false is only returned when [universalLinksOnly]
+/// is set to true and the universal link is failed to launch.
```
amirh: nit: and the universal link failed to launch

PR #1063, opened 2019-01-11
[google_maps_flutter]support for custom map style

ialhamad:
This PR adds support for customizing the map style.


iskakaushik:
We have now added support for custom map styles.

PR #1064, opened 2019-01-11
[firebase_auth]Updated the sign in methods to use new methods

thecoffeehog:
Updated the sign in methods as getCredential is no longer present


collinjackson:
Looks like this was obsoleted by https://github.com/flutter/plugins/pull/1219

PR #1068, opened 2019-01-12
Iap productlist ios

cyanglaz:
Added SKProductWrapper and a SKRequestMaker.
Added unit test for SKProductWrapper classes and SKRequestMaker.
Added respective objc code for IOS platform functionality, including a FIAPRequestHandler to handle all the sk requests.


mklim:
I think this is probably too difficult to review right now, since it contains changes from #1057 (not yet merged, @amirh wants a few more changes on it) plus my branch [shuffling around directories](https://github.com/mklim/plugins/tree/iap_minor_directory_fixup). Do you mind waiting until the extra changes in here are merged into master and then merging again before we review?

mklim:
#1057 is merged, so these changes should be able to be merged or rebased on top of master now.

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -48,7 +48,8 @@ class _MyAppState extends State<MyApp> {
   }
 
   Future<List<Widget>> buildStorefront() async {
-    final bool available = await InAppPurchaseConnection.instance.isAvailable();
+    final InAppPurchaseConnection connection = InAppPurchaseConnection.instance;
```
mklim: This segment isn't going to work on Android yet. Could this file be reverted until a later patch where Android is also done and the unified `Product` is written?

packages/in_app_purchase/ios/Classes/FLTSKProductRequestHandler.m
```diff
@@ -0,0 +1,144 @@
+//
+//  FLTSKProductRequestWrapper.m
+//  in_app_purchase
+//
+//  Created by Chris Yang on 1/10/19.
```
mklim: I don't think we want to include these generated notices in the obj c files, but I'm not totally sure.

/cc @amirh

packages/in_app_purchase/ios/Classes/FLTSKProductRequestHandler.m
```diff
@@ -0,0 +1,144 @@
...
+    @"localizedTitle" : self.localizedTitle ?: [NSNull null],
+    @"productIdentifier" : self.productIdentifier ?: [NSNull null],
+    @"downloadable" : @(self.downloadable),
+    @"price" : self.price ?: [NSNull null],
+    @"downloadContentLengths" : self.downloadContentLengths ?: [NSNull null],
+    @"downloadContentVersion" : self.downloadContentVersion ?: [NSNull null]
+
+  }];
+  if (@available(iOS 10.0, *)) {
+    // TODO: NSLocle is a complex object, want to see the actual need of getting this expanded to
```
mklim: Can you file an issue to track this?

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -31,4 +40,26 @@ - (void)canMakePayments:(FlutterResult)result {
   result([NSNumber numberWithBool:[SKPaymentQueue canMakePayments]]);
 }
 
+- (void)getProductListWithMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result {
+  NSArray *productsIdentifiers = call.arguments[@"identifiers"];
+  [self.productRequestHandler
+      startWithProductIdentifiers:[NSSet setWithArray:productsIdentifiers]
+                completionHandler:^(SKProductsResponse *_Nullable response) {
+                  NSMutableArray *productsDetailsJSON = [NSMutableArray new];
```
mklim: nit: `productsDetailsJSON` is a confusing name because this isn't a JSON string, it's just an array of dictionaries. `productDetailsSerialized`?

packages/in_app_purchase/lib/src/in_app_purchase_connection/google_play_connection.dart
```diff
@@ -30,6 +32,7 @@ class GooglePlayConnection
   void didChangeAppLifecycleState(AppLifecycleState state) {
     switch (state) {
       case AppLifecycleState.paused:
+      case AppLifecycleState.suspending:
```
mklim: Looks like a merge conflict, this line should be deleted.

packages/in_app_purchase/lib/src/in_app_purchase_connection/google_play_connection.dart
```diff
@@ -39,6 +42,8 @@ class GooglePlayConnection
     }
   }
 
+  @override
+  Future<List<Product>> getProductList(List<String> identifiers) => null;
```
mklim: nit: Would prefer `throw UnimplementedError` to make it really clear that this isn't done yet.

https://api.dartlang.org/stable/2.1.0/dart-core/UnimplementedError-class.html

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:in_app_purchase/store_kit_wrappers.dart';
```
mklim: (optional) I think maybe we should implement this and the other `in_app_purchase_connection/` level changes in a followup patch. It's hard to judge how this should look without both APIs in.

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
```diff
@@ -0,0 +1,8 @@
...
+import 'package:flutter/foundation.dart';
+
+class Product {
+  Product({@required this.skProduct});
```
mklim: I don't think this param should be `@required`, since it's not going to be used on Android.

In general I'm a little wary of throwing something on these high level classes that requires users to write `if (platform) then {` type statements and case in their code. Ideally I was thinking that these wrappers should be as sufficient as possible without needing to know the underlying platform. That's not going to be possible 100% of the time, but I think it should be doable here. I get the benefit here of having access to all of the `SkProductWrapper` information instead of having it be lost, but I'm not sure if it's worth it.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,116 @@
+import 'dart:async';
+
+import 'package:flutter/services.dart';
+import 'package:flutter/foundation.dart';
+import 'package:in_app_purchase/src/in_app_purchase_connection/product.dart';
+
+const MethodChannel _channel =
```
mklim: nit: this came up in a merge, but you can import '../channel.dart' for this now.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,116 @@
...
+/// [SKProductRequestWrapper] wraps IOS SKProductRequest class to to retrive StoreKit product information in dart.
+///
+/// https://developer.apple.com/documentation/storekit/skproductsrequest?language=objc
+class SKProductRequestWrapper {
+  /// Get product list.
+  ///
+  /// [identifiers] is the product identifiers specified in Itunes Connect for the products that need to be retrived.
+  ///
+  /// Returns a future containing a list of [SKProduct] which then can be queried to get desired information.
+  static Future<List<Product>> getProductList(List<String> identifiers) async {
```
mklim: So in the original API the request can be started and canceled. Do you think that's worth keeping here? The Dart API equivalent is probably [`CancelableOperation<T>`](https://pub.dartlang.org/documentation/async/latest/async/CancelableOperation-class.html). You'd need iOS to pass up a handle to the request here, and pipe through another method where it can be stopped on cancel.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,117 @@
...
+    return _channel.invokeListMethod<Map<dynamic, dynamic>>(
+      'getProductList',
+      <String, Object>{
+        'identifiers': identifiers,
+      },
+    ).then<List<Product>>((List<Map<dynamic, dynamic>> productListJson) {
```
mklim: nit: This would be easier to read if it used `await` instead of `then`.

https://www.dartlang.org/tutorials/language/futures#async-await

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,117 @@
...
+      final List<Product> productList = <Product>[];
+      for (Map<dynamic, dynamic> productJson in productListJson) {
```
mklim: nit: Like before, `Json` is a confusing name here since we've serialized the data using an array of maps and not a JSON string. I'd rename references to it throughout this patch since we're not really using that format.

https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,117 @@
...
+        productList.add(Product(
+          skProduct: SKProductWrapper.fromJson(productJson),
+        ));
+      }
+      return productList;
+    });
+  }
+}
+
+/// https://developer.apple.com/documentation/storekit/skproduct/2936884-subscriptionperiod?language=objc
```
mklim: nit: Can you add a bit more documentation here? @amirh was preferring that we at least have a summary fragment in the dart doc.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,117 @@
...
+class SKProductSubscriptionPeriodWrapper {
+  SKProductSubscriptionPeriodWrapper(
+      {@required this.numberOfUnits, @required this.unit});
+
+  SKProductSubscriptionPeriodWrapper.fromJson(Map<dynamic, dynamic> json)
```
mklim: nit: I know this causes weird casting problems that bubble upstream, but the method signature should probably be `Map<String, dynamic>` for the type. The keys of the map all need to be strings and not variable, even though the values are dynamic.

packages/in_app_purchase/example/ios/in_app_purchase_pluginTests/ProductRequestHandlerTest.m
```diff
@@ -0,0 +1,180 @@
...
+      @"numberOfPeriods" : @(1),
+      @"subscriptionPeriod" : @{@"numberOfUnits" : @(0), @"unit" : @(0)},
+      @"paymentMode" : @(1)
+    },
+    @"subscriptionGroupIdentifier" : @"com.group"
+  };
+  XCTAssertNotEqualObjects(map, notMatch);
+}
+
+- (void)testRequestDelegateSetup {
```
mklim: Checking for if `delegate` is set seems like it's testing an implementation detail to me, do we need it? For this test to work it looks like a lot of private methods are being exposed just to check internal state on the class, which generally leads to brittle tests. It seems like the thing that really matters is it executes a request and passes a response when `startWithProductIdentifiers` is called. Can we test for that directly instead?

https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

packages/in_app_purchase/example/ios/in_app_purchase_pluginTests/ProductRequestHandlerTest.m
```diff
@@ -0,0 +1,180 @@
...
+@interface ProductRequestHandlerTest : XCTestCase
+
+@end
+
+@implementation ProductRequestHandlerTest
+
+- (void)testSKProductSubscriptionPeriodStubToMap {
+  SKProductSubscriptionPeriodStub *period = [[SKProductSubscriptionPeriodStub alloc] init];
+  NSDictionary *map = [period toMap];
+  NSDictionary *match = @{@"numberOfUnits" : @(0), @"unit" : @(0)};
```
mklim: nit: is It possible to clearly create the `match` and `notMatch` maps throughout here using the values from the stub? It would make the tests less brittle because they'd be actually using the stub values for the assertions instead of having them hardcoded.

Something like:

```
*match = @{@"numberOfUnits" : [period numberOfUnits], @"unit" : [period unit]};
*notMatch = @{@"numberOfUnits" : [period numberOfUnits] + 1, @"unit" : @(1)};
```

packages/in_app_purchase/example/ios/in_app_purchase_pluginTests/in_app_purchase_pluginTests.m
```diff
@@ -16,21 +16,47 @@ - (void)tearDown {
...
-                      XCTAssert(result == [NSNumber numberWithBool:YES]);
+                    result:^(id r) {
+                      [expectation fulfill];
+                      result = r;
                     }];
+  [self waitForExpectations:@[ expectation ] timeout:5];
+  XCTAssertEqual(result, [NSNumber numberWithBool:YES]);
+}
+
+- (void)testGetProductListEmpty {
```
mklim: Is there any way we can unit test an actually successful response here, or is it not mockable?

packages/in_app_purchase/ios/Classes/FLTSKProductRequestHandler.h
```diff
@@ -0,0 +1,57 @@
...
+//
+
+#import <Foundation/Foundation.h>
+#import <StoreKit/StoreKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void (^ProductRequestCompletion)(SKProductsResponse *_Nullable response);
+
+@interface FLTSKProductRequestHandler : NSObject
```
mklim: What does `FLTSK` stand for?

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -31,4 +40,26 @@ - (void)canMakePayments:(FlutterResult)result {
...
+                  for (SKProduct *product in response.products) {
+                    [productsDetailsJSON addObject:[product toMap]];
+                  }
+                  result(productsDetailsJSON);
+                }];
+  [self.productRequestHandler
```
mklim: It seems like this block is repeated twice, is that intentional?

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,116 @@
...
+  static Future<List<Product>> getProductList(List<String> identifiers) async {
+    return _channel.invokeListMethod<Map<dynamic, dynamic>>(
+      'getProductList',
+      <String, Object>{
+        'identifiers': identifiers,
+      },
+    ).then<List<Product>>((List<Map<dynamic, dynamic>> productListJson) {
+      final List<Product> productList = <Product>[];
+      for (Map<dynamic, dynamic> productJson in productListJson) {
+        productList.add(Product(
```
mklim: Since this is part of the `store_kit_wrappers/`, it should probably be returning SkProductWrapper instead of the higher level abstraction. `AppStoreConnection` can then call this and convert it internally into the more generic abstraction.

packages/in_app_purchase/ios/Classes/FLTSKProductRequestHandler.m
amirh: Yeah we don't usually add these.

packages/in_app_purchase/example/ios/in_app_purchase_pluginTests/ProductRequestHandlerTest.m
```diff
@@ -87,94 +17,87 @@ @implementation ProductRequestHandlerTest
...
+    @"subscriptionPeriod" : [product.subscriptionPeriod toMap],
+    @"introductoryPrice" : [product.introductoryPrice toMap],
+    @"subscriptionGroupIdentifier" : product.subscriptionGroupIdentifier
   };
   XCTAssertNotEqualObjects(map, notMatch);
 }
 
-- (void)testRequestDelegateSetup {
-  FLTSKProductRequestHandler *handler = [FLTSKProductRequestHandler new];
+- (void)testRequestHandler {
```
mklim: 🙌 This is great, thanks.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -54,20 +65,27 @@ class SKProductDiscountWrapper {
...
       @required this.subscriptionPeriod});
 
-  SKProductDiscountWrapper.fromJson(Map<dynamic, dynamic> json)
-      : price = json['price'],
-        numberOfPeriods = json['numberOfPeriods'],
-        paymentMode = json['paymentMode'],
-        subscriptionPeriod = json['subscriptionPeriod'] != null
-            ? SKProductSubscriptionPeriodWrapper.fromJson(
-                json['subscriptionPeriod'])
+  /// Consctruter to build with a map
```
mklim: nit: "Constructor"

packages/in_app_purchase/test/store_kit_wrappers/sk_product_request_wrapper_test.dart
```diff
@@ -0,0 +1,151 @@
...
+
+import 'package:test/test.dart';
+import 'package:in_app_purchase/src/store_kit_wrappers/sk_product_request_wrapper.dart';
+import 'package:in_app_purchase/src/channel.dart';
+import 'package:flutter/services.dart';
+import 'package:in_app_purchase/src/in_app_purchase_connection/product.dart';
+import '../fake_platform_views_controller.dart';
+
+void main() {
+  final FakePlatformViewsController fakePlatformViewsController =
```
mklim: This is a little behind master, this class was renamed.

packages/in_app_purchase/test/store_kit_wrappers/sk_product_request_wrapper_test.dart
```diff
@@ -0,0 +1,151 @@
...
+        fakePlatformViewsController.fakePlatformViewsMethodHandler);
+  });
+
+  group('canMakePayments', () {
+    test(
+        'SKProductSubscriptionPeriodWrapper should have property values consistent with map',
+        () {
+      final SKProductSubscriptionPeriodWrapper wrapper =
+          SKProductSubscriptionPeriodWrapper.fromMap(subMap);
+      expect(wrapper.numberOfUnits, 0);
```
mklim: nit: here and throughout it'll be less brittle and a little clearer if you refer to the expected values directly instead of hardcoding magic constants for them.

For example: `expect(wrapper.numberOfUnits, subMap['numberOfUnits']);`

That way if you update the expected value the tests still pass, and the person reading these doesn't have to scroll up to where the expected values are defined to make sure that the tests aren't passing based on a false positive.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,139 @@
...
+    for (Map<dynamic, dynamic> productMap in productListSerilized) {
+      productList.add(SKProductWrapper.fromMap(
+        productMap.cast<String, dynamic>(),
+      ));
+    }
+    return productList;
+  }
+}
+
+/// This class wraps the IOS SKProductSubscriptionPeriod class
```
amirh: nits: iOS, period at the end of sentence, avoid the "This class" prefix.

I'd suggest an alternative phrasing in the spirit of:

```dart
/// A subscription period.
///
/// A period is defined by a [numberOfUnits] and a [unit], e.g for a 3 months period [numberOfUnits] is 3 and [unit] is a month.
///
/// This is a wrapper for StoreKit's [SKProductSubscriptionPeriod](https://developer.apple.com/documentation/storekit/skproductsubscriptionperiod?language=objc).
``` 

Same for the other Dartdocs below.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,139 @@
...
+      {@required this.numberOfUnits, @required this.unit});
+
+  /// Constructor to build with a map
+  ///
+  /// Used for constructing the class with the map passed from the OBJC layer.
+  SKProductSubscriptionPeriodWrapper.fromMap(Map<String, dynamic> map)
+      : numberOfUnits = map['numberOfUnits'],
+        unit = map['unit'];
+
+  final int numberOfUnits, unit;
```
amirh: Dartdoc for public members.

amirh: What are the values for unit? should this be an enum instead?

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+class Product {}
```
amirh: What is this used for? Dartdoc?

packages/in_app_purchase/ios/Classes/FIAPProductRequestHandler.m
```diff
@@ -0,0 +1,105 @@
...
+    @"localizedTitle" : self.localizedTitle ?: [NSNull null],
+    @"productIdentifier" : self.productIdentifier ?: [NSNull null],
+    @"downloadable" : @(self.downloadable),
+    @"price" : self.price ?: [NSNull null],
+    @"downloadContentLengths" : self.downloadContentLengths ?: [NSNull null],
+    @"downloadContentVersion" : self.downloadContentVersion ?: [NSNull null]
+
+  }];
+  if (@available(iOS 10.0, *)) {
+    // TODO: NSLocle is a complex object, want to see the actual need of getting this expanded to
```
amirh: TODOs should be mention an owner, and reference an issue, see example here: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-checking-in-comments-that-ask-questions

amirh: nit: type NSLocale

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
```diff
@@ -0,0 +1,139 @@
...
+
+import 'dart:async';
+
+import 'package:flutter/foundation.dart';
+import '../channel.dart';
+
+/// [SKProductRequestWrapper] wraps IOS SKProductRequest class to to retrive StoreKit product information in dart.
+///
+/// https://developer.apple.com/documentation/storekit/skproductsrequest?language=objc
+class SKProductRequestWrapper {
```
amirh: I don't see much resemblance to SKProductRequestWrapper, SKRequest seems like something that represents a single App Store request and manages its state(start, cancel, callbacks).
While this a wrapper for a static method...

I'm not sure whether the bigger part of the method channel API we will end up building will be similar to what we do here, or to what we did with `canMakePayments`(basically a mirror of the App Store API).
If the former is the case, what do you guys think about having a (AppStore/AppStoreClient/FlutterAppStoreClient?) class which implements the Dart side of the method channel API?

@mklim

packages/in_app_purchase/lib/src/in_app_purchase_connection/app_store_connection.dart
```diff
@@ -26,4 +27,11 @@ class AppStoreConnection implements InAppPurchaseConnection {
     _instance = AppStoreConnection();
     return _instance;
   }
+
+  @override
+  Future<List<Product>> getProductList(List<String> identifiers) =>
+      throw UnimplementedError();
+
+  Future<List<SKProductWrapper>> getSKProductList(List<String> identifiers) =>
```
amirh: @mklim was the intention for `AppStoreConnection` and `GooglePlayConnection` to not extend the API defined in `InAppPurchaseConnection` and expose platform specific features only in the wrappers?

(I guess I'm ok with both options, just wanted to make sure I'm on the same page with you here)

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
mklim: I'm not totally sure I'm following. @amirh are you saying that this API is too different from `SkProductsRequest`? If so I agree here. Ideally I'd like everything in the `wrappers/` layer to be a thin wrapper around the original API, with as little a translation as possible but taking advantage of Dart's language features when appropriate (mostly translating callbacks to futures). I think for now it's a problem that the naming on this doesn't match up, and it's not really wrapping the response. This method should be returning the data from [`SkProductsRequestDelegate`](https://developer.apple.com/documentation/storekit/skproductsrequestdelegate?language=objc), so the [products and a list of invalid identifiers](https://developer.apple.com/documentation/storekit/skproductsresponse?language=objc). Right now it's missing information from the App Store API.

I mentioned in an earlier comment that this should ideally also be returning a cancelable instance but didn't want to block this PR on it. What I'm thinking of would still be a static method on the class but would be returning an instance of a cancelable request, which I think is a close enough match. The important thing is to make sure we're not dropping any functionality.

```
class SKProductRequestWrapper {
   static CancelableOperation<Future<SkProductsResponseWrapper>> initWithProductIdentifiers(List<String> productIdentifiers);
}

class SkProductsResponseWrapper {
  final List<SkProductsWrapper> products;
  final List<String> invalidIdentifiers;
}
```

Then it's more the job of the generic API that's a level up from here to adapt that and just return a single `List<Product>`.

>what do you guys think about having a (AppStore/AppStoreClient/FlutterAppStoreClient?) class which implements the Dart side of the method channel API?

Also not totally sure what you're suggesting here, can you explain more?

packages/in_app_purchase/lib/src/in_app_purchase_connection/app_store_connection.dart
mklim: Yeah, this layer should be the generic API only. @cyanglaz, I think you can revert this file entirely. This patch can just be isolated to the `wrapper/` layer instead of building out the generic code.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_request_wrapper.dart
mklim: OK, we caught up offline. Summarizing the discussion below.

So the goal for the `wrappers/` here is to ultimately expose all the platform functionality, and make sense at a glance given a previous understanding of how the underlying platform works. That said in a specific case like `SkRequest` wrapping each type of request as it's own class with a single static method isn't really worth it. It doesn't mirror the underlying platform enough to be instantly recognizable: `start`, `stop`, `cancel`, and the `delegate` all need to be cleared out to make sense in Dart. And the end result is really hard to read Dart, too.

So for something like this we should make a more idiomatic API, as long as the function is still totally understandable in terms of the underlying platform.

So instead of my original suggestion above,  collecting all the `SkRequest` methods into one place without redundant named classes for any of them. But clear naming for each of the methods so it's obvious what they are at a glance (or grep), and the returned data types still covering everything.

```
class SkRequester {
  Future<SkProductsResponseWrapper> startSkProductsRequest(List<String> productIdentifiers); // optionally CancelableOperation<Future<>> instead

  // etc
}

class SkProductsResponseWrapper {
  final List<SkProductsWrapper> products;
  final List<String> invalidIdentifiers;
}
```

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -0,0 +1,179 @@
...
+/// A payment mode to describe how the discounted price is paid.
+///
+/// [PayAsYouGo] allows user to pay the discounted price at each payment period.
+/// [PayUpFront] allows user to pay the discounted price upfront and receive the product for the rest of time that was paid for.
+/// [FreeTrail] user pays nothing during the discounted period.
+/// This is a wrapper for StoreKit's [SKProductDiscountPaymentMode](https://developer.apple.com/documentation/storekit/skproductdiscountpaymentmode?language=objc).
+enum ProductDiscountPaymentMode {
+  PayAsYouGo,
+  PayUpFront,
+  FreeTrail,
```
mklim: nit: "freeTrial", and these should be lowerCamelCase.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -0,0 +1,179 @@
...
+  SKProductSubscriptionPeriodWrapper(
+      {@required this.numberOfUnits, @required this.unit});
+
+  /// Constructor to build with a map.
+  ///
+  /// Used for constructing the class with the map passed from the OBJC layer.
+  SKProductSubscriptionPeriodWrapper.fromMap(Map<String, dynamic> map)
+      : numberOfUnits = map['numberOfUnits'],
+        unit = (map['unit'] != null)
+            ? SubscriptionPeriodUnit.values[map['unit']]
```
mklim: Do these enum values match across both platforms? eg `SkProductPeriodUnit.day` is 0, `SkProductDiscountPaymentMode.PayAsYouGo` is 0, etc.

cyanglaz: Yes it matches the objc code.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -0,0 +1,255 @@
...
+import 'dart:async';
+import 'package:flutter/services.dart';
+import 'package:flutter/foundation.dart';
+import 'package:in_app_purchase/src/channel.dart';
+
+/// A product request.
+///
+/// This is a wrapper for StoreKit's
+/// [SKProductsRequest](https://developer.apple.com/documentation/storekit/skproductsrequest?language=objc).
+class SKProductRequestWrapper {
```
cyanglaz: @amirh I am naming this to SKProductRequestWrapper since it is resemble the underlining objc code as well. (Without the cancel method for now)

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// A consolidated product class represents the common fields in StoreKit product and BillingClient product.
+///
+/// This product class is returned from the [getProductList] method of a [InAppPurchaseConnection] instance. Use this class
+/// if you only need to have genenric and basic implementation. If prefer a detailed platform specific impelmentation,
+/// use the platform specific class [SKProductWrapper] or [SkuDetailsWrapper] that is inside this class.
+class Product {
```
amirh: nit: I'd introduce this in the first PR that it is used (I didn't see it in use here, did I miss something?)

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -35,4 +44,49 @@ - (void)canMakePayments:(FlutterResult)result {
   result([NSNumber numberWithBool:[SKPaymentQueue canMakePayments]]);
 }
 
+- (void)startProductRequest:(FlutterMethodCall *)call result:(FlutterResult)result {
+  if (![call.arguments isKindOfClass:[NSArray class]]) {
+    result([FlutterError errorWithCode:@"storekit_invalide_argument"
```
amirh: nit: typo

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -35,4 +44,49 @@ - (void)canMakePayments:(FlutterResult)result {
...
+                               message:@"Argument type of startRequest is not array"
+                               details:call.arguments]);
+    return;
+  }
+  NSArray *productsIdentifiers = (NSArray *)call.arguments;
```
amirh: nit: productIdentifiers

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -35,4 +44,49 @@ - (void)canMakePayments:(FlutterResult)result {
...
+  FIAPRequestHandler *handler = [[FIAPRequestHandler alloc] initWithRequest:request];
+  handler.delegate = self;
+  [self.RequestHandlerSet addObject:handler];
+  [handler startProductRequestWithCompletionHandler:^(SKProductsResponse *_Nullable response,
+                                                      NSError *_Nullable error) {
+    if (error) {
+      NSString *details = [NSString stringWithFormat:@"Reason:%@\nRecoverSuggestion:%@",
+                                                     error.localizedFailureReason,
+                                                     error.localizedRecoverySuggestion];
+      result([FlutterError errorWithCode:error.domain message:error.description details:details]);
```
amirh: The error code should be probably a value in our control, otherwise there's no way to guarantee we don't report different errors with the same code.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -4,28 +4,37 @@
 
 #import "InAppPurchasePlugin.h"
 #import <StoreKit/StoreKit.h>
+#import "FIAObjectTranslator.h"
+#import "FIAPRequestHandler.h"
+
+@interface InAppPurchasePlugin () <FIAPRequestHandlerDelegate>
+
+@property(strong, nonatomic) NSMutableSet<FIAPRequestHandler *> *RequestHandlerSet;
```
amirh: Please add a comment explaining the role of this set.

packages/in_app_purchase/ios/Classes/FIAObjectTranslator.h
```diff
@@ -0,0 +1,37 @@
+//
```
amirh: replace with the standard license header

packages/in_app_purchase/ios/Classes/FIAObjectTranslator.m
```diff
@@ -0,0 +1,88 @@
+//
```
amirh: replace with the standard license header

packages/in_app_purchase/example/ios/in_app_purchase_pluginTests/TranslatorTest.m
```diff
@@ -0,0 +1,76 @@
+//
```
amirh: replace with the standard license header

cyanglaz:
@amirh I have updated the docs, tried to follow the style guide as much as possible. Could you go through the PR another round and comment on the ones need to be approved or fixed.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -4,9 +4,52 @@
...
+  ///
+  /// Used for constructing the class with map passed from the OBJC layer.
+  SkProductResponseWrapper.fromMap(Map<String, List<dynamic>> map)
+      : products = _getListFromMapList(_getProductMapListFromResponseMap(map)),
+        invalidProductIdentifiers =
+            List.castFrom<dynamic, String>(map['invalidProductIdentifiers']);
+
+  /// The list of the products.
+  ///
+  /// Can be null if the [SKProductRequestMaker]'s [startProductRequest] method does not pass correct [productIdentifiers].
```
mklim: Handling null collections is very error-prone. It would be much better if there was always a `List<>` returned here, but it was just empty in cases where there were no products. That way devs using this won't have to guard against `null` here and can just consistently iterate over the list. (This applies to a few lists in this patch.)

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -4,9 +4,52 @@
...
+  }
+}
+
 /// A unit discription the length of a subscription period.
 ///
-/// This is wrapper of StoreKit's [SKProductPeriodUnit](https://developer.apple.com/documentation/storekit/skproductperiodunit?language=objc).
+/// This is wrapper for StoreKit's
+/// [SKProductPeriodUnit](https://developer.apple.com/documentation/storekit/skproductperiodunit?language=objc).
+/// This is used as a property in the [SKProductSubscriptionPeriodWrapper].
+/// The values of the enum options are matching the [SKProductPeriodUnit]'s values. Should there be an update or addition
```
mklim: I don't think this bit about `SKProductPeriodUnit` being kept in sync is public information for the users of the API, it's more of an implementation note for developers on the API. Can you move it out of the Dart Doc into a private comment? (This also applies to `ProductDiscountPaymentMode`)

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_request_maker.dart
```diff
@@ -0,0 +1,39 @@
...
+
+import 'dart:async';
+import 'package:flutter/services.dart';
+import 'package:in_app_purchase/src/channel.dart';
+import 'sk_product_wrapper.dart';
+
+/// A request maker that handles all the requests made by SKRequest subclasses.
+///
+/// There are multiple [SKRequest] subclasses handling different requests in the [StoreKit] with multiple delegate methods,
+/// we consolidateded all the [SKRequest] subclasses into this class to make requests handling more straightforward.
```
mklim: nit: A little hard to read, "to make requests in a more straightforward way." would be more clear.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_request_maker.dart
```diff
@@ -0,0 +1,39 @@
...
+/// [SKRequest](https://developer.apple.com/documentation/storekit/skrequest?language=objc)
+class SKRequestMaker {
+  /// Starts a request to get SKProduct response.
+  ///
+  /// [productIdentifiers] should contain legit product identifiers that you decleared for the products in the Itunes Connect.
```
mklim: nit: s/decleared/declared, s/omiited/omitted. I also think it's called "App Store Connect" now.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_request_maker.dart
```diff
@@ -0,0 +1,39 @@
...
+  /// [SkProductResponseWrapper] is returned if there is no error during the request.
+  /// An exception can be thrown if an underlining error occurs when making the request from the IOS platform.
+  Future<SkProductResponseWrapper> startProductRequest(
+      List<String> productIdentifiers) async {
+    final Map<dynamic, dynamic> productResponseMap = await channel.invokeMethod(
+      '-[InAppPurchasePlugin startProductRequest:result:]',
+      productIdentifiers,
+    );
+    if (productResponseMap == null) {
+      throw PlatformException(
```
mklim: I'm not sure if throwing an error here is the right answer. This would crash the app, right? And it's completely outside of the developers control.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -43,33 +31,41 @@ - (void)canMakePayments:(FlutterResult)result {
   result([NSNumber numberWithBool:[SKPaymentQueue canMakePayments]]);
 }
 
-- (void)getProductListWithMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result {
-  NSArray *productsIdentifiers = call.arguments[@"identifiers"];
+- (void)startProductRequest:(FlutterMethodCall *)call result:(FlutterResult)result {
+  if (![call.arguments isKindOfClass:[NSArray class]]) {
+    result([FlutterError errorWithCode:@"storekit_invalid_argument"
+                               message:@"Argument type of startRequest is not array"
```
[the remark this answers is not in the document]
mklim: FYI this is still unresolved.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_request_maker.dart
```diff
@@ -7,13 +7,20 @@ import 'package:flutter/services.dart';
 import 'package:in_app_purchase/src/channel.dart';
 import 'sk_product_wrapper.dart';
 
-/// Handles all the SKRequest subclasses.
+/// A request maker that handles all the requests made by SKRequest subclasses.
 ///
+/// There are multiple [SKRequest] subclasses handling different requests in the [StoreKit] with multiple delegate methods,
+/// we consolidateded all the [SKRequest] subclasses into this class to make requests in a more straightforward way.
```
mklim: nit: s/consolidateded/conslidated

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -0,0 +1,256 @@
...
+        unit = (map['unit'] != null)
+            ? SubscriptionPeriodUnit.values[map['unit']]
+            : null;
+
+  /// The number of a certain units to represent the period, the unit is defined in the [unit] property.
+  ///
+  /// This should have a value >= 0.
+  final int numberOfUnits;
+
+  /// The unit that combined with [numberOfUnits] to define the length of the subscripton.
```
amirh: The time unit used to specify the length of this period.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -0,0 +1,256 @@
...
+                map['introductoryPrice'].cast<String, dynamic>())
+            : null;
+
+  /// The unique identifier of the product.
+  final String productIdentifier,
+
+      /// The localizedTitle of the product.
+      ///
+      /// It is localized based on the current locale.
+      localizedTitle,
```
amirh: I can't find a specific section about this in the style guide but this style is not typically used in the Flutter codebase, so I'd stay consistent.

Lets replace each of these with e.g `final String localizedTitle;`.

One reason I could think of why that would be better is that when I stare and one of these deep down in the list (e.g downloadContentVersion) I can't easily tell what the type is but have to scroll up.

@Hixie maybe worth mentioning this in the style guide?

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_request_maker.dart
```diff
@@ -0,0 +1,39 @@
...
+///
+/// There are multiple [SKRequest] subclasses handling different requests in the [StoreKit] with multiple delegate methods,
+/// we consolidated all the [SKRequest] subclasses into this class to make requests in a more straightforward way.
+/// [SKRequest](https://developer.apple.com/documentation/storekit/skrequest?language=objc)
+class SKRequestMaker {
+  /// Starts a request to get SKProduct response.
+  ///
+  /// [productIdentifiers] should contain legit product identifiers that you declared for the products in the Itunes Connect.
+  /// Duplicate values in [productIdentifiers] will be omitted. If [productIdentifiers] is null, an ["storekit_invalid_argument"] error will
+  /// be returned. If [productIdentifiers] is empty, a [SkProductResponseWrapper] will still be returned with [products] being null.
```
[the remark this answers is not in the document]
cyanglaz: products is one of the properties in the SKProductResponseWrapper

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -0,0 +1,256 @@
...
+
+/// Dart wrapper around StoreKit's [SKProductsResponse](https://developer.apple.com/documentation/storekit/skproductsresponse?language=objc).
+///
+/// Represents the response object returned by [startProductRequest].
+/// Contains information about a list of products and a list of invalid product identifers.
+class SkProductResponseWrapper {
+  SkProductResponseWrapper(
+      {@required this.products, @required this.invalidProductIdentifiers});
+
+  /// Used for constructing the class with map passed from the OBJC layer.
```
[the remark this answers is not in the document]
cyanglaz: The structure is defined in the objective-c code, is dart doc able to refer to objective-c classes?

PR #1070, opened 2019-01-13
[cloud_functions]Remove internal FIR methods

radvansky-tomas:
Internal methods caused build errors https://github.com/flutter/flutter/issues/21973 and they should not be used anyway!


amirh:
@kroikie

collinjackson:
Looks like this was addressed in fc5fab0e40050c1d9f0ae30792eacbe9af120606 but didn't remove the internal header reference; I've opened https://github.com/flutter/plugins/pull/1747 for that.

PR #1072, opened 2019-01-13
adding nil check

cyanglaz:
Adding nil checks in NSDictionary and default them to empty string to prevent crashes. 

https://github.com/flutter/flutter/issues/20761


packages/package_info/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Flutter plugin for querying information about the application
   package, such as CFBundleVersion on iOS or versionCode on Android.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/package_info
-version: 0.3.2
+version: 0.3.3
```
amirh: this should be 0.3.2+1 see: https://www.dartlang.org/tools/pub/versioning#semantic-versions

PR #1073, opened 2019-01-13
Add missing imports to README.md image_picker code example

wendyadi:
Add missing imports to README.md image_picker code example

`dart:io` for `File` type. 
`package:flutter/material.dart` for `StatefulWidget`

```
import 'dart:io';
```

### DIFF
--- a/packages/image_picker/README.md
+++ b/packages/image_picker/README.md
@@ -26,7 +26,10 @@ No configuration required - the plugin should work out of the box.
 ### Example
 
 ``` dart
+import 'dart:io';
+import 'package:flutter/material.dart';
 import 'package:image_picker/image_picker.dart';
+
 
 class MyHomePage extends StatefulWidget {
   @override


amirh:
I'm a little unsure if this sample was meant to be a standalone app (in which case we need to add not just the imports but also a main function), or a snippet showing how to use just the image picker (in which case I think we might be able to trim down the sample even further).

cc @mit-mit who wrote this sample, do you have any preferences here?

mit-mit:
The example snippets in README.md files are not expected to be self-contained. For that, we have complete, runable examples inside the `/example/` sub-folder of each plugin. I believe this is consistent across the plugins, see e.g. https://github.com/flutter/plugins/tree/master/packages/cloud_firestore

amirh:
Thanks, I'm going to close this PR.
@wendyadi feel if you're interesting in simplifying this sample further a PR will be welcomed!

PR #1074, opened 2019-01-13
[Half PR / Android] google maps plugin - icon scaling

matanshukry:
This allow one to specify width and height, which will scale the loaded asset.

Currently, this should not be accepted or iOS won't work at all. Probably anyway, I didn't actually read the iOS code.

notes:
(1) I don't have a mac, so I couldn't work on the iOS part. Just to get started, one should just override the onMethodCall (equivalent in ios), and get the correct parameters (ignoring the width and height parameters).
(2) I'm having issue with pub on my machine, so I couldn't actually run the analyze command:
pub global run flutter_plugin_tools analyze --plugins plugin_name

I managed to run the other 2 though, so it should be fine.

### DIFF
--- a/packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
+++ b/packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
@@ -4,7 +4,11 @@
 
 package io.flutter.plugins.googlemaps;
 
+import android.content.Context;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
 import android.graphics.Point;
+
 import com.google.android.gms.maps.CameraUpdate;
 import com.google.android.gms.maps.CameraUpdateFactory;
 import com.google.android.gms.maps.model.BitmapDescriptor;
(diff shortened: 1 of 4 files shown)


matanshukry:
Seems like CI fails on test+format. I ran the pub format cmd though. Or rather, I ran the flutter format command manually, the one that the pub format outputs. Got an "Unchanged" line for every file.

I'm guessing it has to do with me being on window, so would be great if someone else can run it, or reach privately to better understand the issue & solve it, allowing windows users to run format.

amirh:
Thanks for the contribution!

This will need an iOS implementation and e2e testing before we can land it, also note that I'm reworking the markers API and it might change soon.

As we currently don't have a harness for testing the platform side of plugins, current and new features are not properly covered by tests.
In order to maintain a quality bar for flutter/plugins we are going to avoid adding new features that aren't critical until they can be properly tested.

The test harness work is tracked in flutter/flutter#10824 and flutter/flutter#26080, your feedback on the testing work will be really appreciated (will it properly support the testing needs of this PR?) some design details should be available on flutter/flutter#10824 soon(contributions to that effort are welcomed!).

I'm going to close this PR for now, would you mind re-opening it after an iOS implementation and e2e tests were added? (once a test harness is ready)

matanshukry:
Well I can't really do the iOS implementation I'm afraid, so I won't be able to do that part.

Just as a side note, not sure how you define critical but I think this is an important feature. Currently, For some kind of apps (just as the one I'm making) it's a blocker, and there's no way to bypass that, that I know of.

I thought about changing the asset file size it self, but I'm pretty sure there will still be inconsistent sizes between iOS and Android, as you can see in this issue:
https://github.com/flutter/flutter/issues/24865

amirh:
Interesting, can you share some more details about your specific use case? do you only know the desired icon size at runtime?

matanshukry:
@amirh - My specific use case resembles a bit the issue I provided; I'm allowing users to create share and view different kind of poi (points of interest); each one with different icons. The icon size can be static at compile time, at least for now, but it needs to be consistent across devices; I can't have the icon really small on iOS and really big on Android, or vice versa.

PR #1075, opened 2019-01-13
[firebase_analytics]Refactor unit test to use `setMockMethodCallHandler`

truongsinh:
Following the discussion on https://github.com/flutter/flutter/pull/26456#discussion_r247304700, I figured out we have several places in https://github.com/flutter/plugins using old hacking `.private`, thus this is an continuingly attempt to clean up


truongsinh:
cc @amirh @jonahwilliams

packages/firebase_analytics/test/firebase_analytics_test.dart
```diff
@@ -117,36 +108,30 @@ void main() {
...
-        final Map<String, dynamic> args = invocation.positionalArguments[1];
+      channel.setMockMethodCallHandler((MethodCall call) async {
+        if (call.method != 'logEvent') {
+          throw ArgumentError('Only logEvent invocations expected');
+        }
+        final Map<String, dynamic> args =
+            Map<String, dynamic>.from(call.arguments);
         name = args['name'];
-        parameters = args['parameters'];
+        if (args['parameters'] != null) {
```
amirh: Should we first set `parameters = null` to make sure we're not left with a stale value from a previous invocation?

truongsinh: ok, agree better safe than sorry. I'll make the change in a moment.

truongsinh: resolve with latest push

packages/firebase_analytics/test/firebase_analytics_test.dart
```diff
@@ -117,36 +108,27 @@ void main() {
   });
 
   group('$FirebaseAnalytics analytics events', () {
-    FirebaseAnalytics analytics;
-
     String name;
-    Map<String, dynamic> parameters;
+    dynamic parameters;
```
amirh: Flutter's style is to avoid dynamic variables when possible, see: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-using-var

(same for args below)

truongsinh: I can make the change, but just asking. Should this "avoiding dynamic" be applicable to test files? I'm totally for strong-typing and safe-typing, but in test files, especially this one, (ab)using `dynamic` makes test code more verbose, and sometimes even error-prone (I forgot to reset null https://github.com/flutter/plugins/pull/1075#discussion_r248042961), while this var is always asserted/expected https://github.com/flutter/plugins/blob/521a26ea4317e4ca64334f0acf0c9366ab4f0e15/packages/firebase_analytics/test/firebase_analytics_test.dart#L138

collinjackson: I think we should follow the pattern of other tests and use the `isMethodCall` matcher in the test itself, rather than asserting in the call to `setMockMethodCallHandler`. See [cloud_functions unit tests](https://github.com/flutter/plugins/blob/master/packages/cloud_functions/test/cloud_functions_test.dart) for an example.

As a bonus this will eliminate the need for the `dynamic` variable here (I think).

If you update this PR could you please make sure to update pubspec.yaml/CHANGELOG and also convert over observer_test.dart.

Thanks!

truongsinh: I made the change to use `isMethodCall`, but i'm sorry i not sure what to update pubspec.yaml/CHANGELOG.

truongsinh: For `observer_test.dart`, I think it's good as is.

truongsinh:
@collinjackson I made the change to use isMethodCall, but i'm sorry i not sure what to update pubspec.yaml/CHANGELOG. For observer_test.dart, I think it's good as is.

PR #1079, opened 2019-01-15
Added additional sample rates for sensor events

LeonardGreulich:
### Three options for all three sensor events:
**SampleRate.low** results in 15 sensor events per second on both platforms.
**SampleRate.medium** results in 50 sensors events per second on both platforms.
**SampleRate.high** results in 100 sensors events per second on iOS and 120 on Android.

A sample rate is set before the listening with, e.g., **setSensorsSampleRate(SampleRate.medium);**.


LeonardGreulich:
I signed it!

PR #1080, opened 2019-01-15
[firebase_crashlytics]Firebase Crashlytics plugin

kroikie:
Add plugin to FlutterFire that will report Dart side errors as non-fatal errors to the Firebase console.

Fixes: https://github.com/flutter/flutter/issues/14765


packages/firebase_crashlytics/README.md
```diff
@@ -0,0 +1,103 @@
...
+Default FirebaseApp is not initialized in this process [package name].
+Make sure to call FirebaseApp.initializeApp(Context) first.
+```
+
+*Note:* When you are debugging on android, use a device or AVD with Google Play services.
+Otherwise you will not be able to use Firebase Crashlytics.
+
+### iOS Integration
+
+Add the Crachlytics run scripts
```
Taormina: Crashlytics

kroikie: Thanks fixed.

packages/firebase_crashlytics/android/src/main/java/io/flutter/plugins/firebase/crashlytics/firebasecrashlytics/FirebaseCrashlyticsPlugin.java
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * @param line Line from Dart stack trace.
+   * @return Stack trace element to be used as part of an Exception stack trace.
+   */
+  private StackTraceElement generateStackTraceElement(String line) {
+    // Split line on white spaces.
+    String[] lineParts = line.split("\\s+");
+    String fileName = lineParts[0].trim();
+    String lineNumber = lineParts[1].trim().substring(0, lineParts[1].indexOf(":"));
+    String className = lineParts[2].trim().substring(0, lineParts[2].indexOf("."));
```
slightfoot: It's possible that if `trim()` removes spaces from the start and then the call to `indexOf` will report an incorrect position and these line's can throw `IndexOutOfBoundsException`.

kroikie: @slightfoot I think it could be an issue, I'll add some checks to make this safer.

kroikie: Called trim after the string operations.

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -0,0 +1,123 @@
+import 'package:firebase_crashlytics/firebase_crashlytics.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('$Crashlytics', () {
+    const MethodChannel channel =
```
collinjackson: For other plugins I've making the channel be `@visibleForTesting` and then just using it here. This is just to avoid unnecessary duplication as we need the channel to intercept method calls. I don't really feel strongly about it though.

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -0,0 +1,123 @@
...
+        isMethodCall('Crashlytics#log',
+            arguments: <String, dynamic>{'msg': 'foo'}),
+      ]);
+    });
+
+    test('sendKeys', () async {
+      crashlytics.setKey('baz', 'qux');
+      crashlytics.setKey('quux', 1);
+      crashlytics.setKey('quuz', false);
+      await crashlytics.sendKeys();
```
collinjackson: Instead of calling `sendKeys()` here just `await Future<void>.delayed(const Duration(seconds: 0));`

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+      // https://github.com/flutter/flutter/issues/26431
+      // ignore: strong_mode_implicit_dynamic_method
+      await channel.invokeMethod('Crashlytics#log', <String, dynamic>{
+        'msg': logs.elementAt(i),
+      });
+    }
+  }
+
+  @visibleForTesting
+  Future<void> sendKeys() async {
```
collinjackson: This code should be moved inline to where you're calling it. 

Unit tests should be testing the public API only.

You can flush the method calls with another method (see below).

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+  Future<void> setUserName(String name) async {
+    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+    // https://github.com/flutter/flutter/issues/26431
+    // ignore: strong_mode_implicit_dynamic_method
+    await channel.invokeMethod(
+        'Crashlytics#setUserName', <String, dynamic>{'name': name});
+  }
+
+  @visibleForTesting
+  Future<void> sendLogs() async {
```
collinjackson: This code should be moved inline to where you're calling it. 

Unit tests should be testing the public API only.

You can flush the method calls with another method (see below).

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -0,0 +1,123 @@
...
+      await crashlytics.setUserName('foo');
+      expect(log, <Matcher>[
+        isMethodCall('Crashlytics#setUserName',
+            arguments: <String, dynamic>{'name': 'foo'})
+      ]);
+    });
+
+    test('sendLogs', () async {
+      crashlytics.log('foo');
+      await crashlytics.sendLogs();
```
collinjackson: Instead of calling `sendKeys()` here just `await Future<void>.delayed(const Duration(seconds: 0));`

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -0,0 +1,123 @@
...
+    });
+
+    test('crash', () {
+      expect(() => crashlytics.crash(), throwsStateError);
+    });
+
+    test('getVersion', () async {
+      crashlytics.getVersion();
+      expect(log,
+          <Matcher>[isMethodCall('Crashlytics#getVersion', arguments: null)]);
```
collinjackson: Nit: more newlines and trailing commas will make this look nicer

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    while (logSize > 65536) {
+      final String first = logs.removeFirst();
+      logSize -= Uint8List.fromList(first.codeUnits).length;
+    }
+  }
+
+  void setKey(String key, dynamic value) {
+    // Check that only 64 keys are set.
+    if (keys.containsKey(key) || keys.length <= 64) {
+      keys[key] = value;
```
collinjackson: I wonder if there should be more error handling in the case that a key cannot be set. It would be hard though... this is a crash plugin, where would the error go? :)

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+  Future<String> getVersion() async {
+    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+    // https://github.com/flutter/flutter/issues/26431
+    // ignore: strong_mode_implicit_dynamic_method
+    final String result = await channel.invokeMethod('Crashlytics#getVersion');
+    return result;
+  }
+
+  void log(String msg) {
+    logSize += Uint8List.fromList(msg.codeUnits).length;
```
collinjackson: It might be worthwhile to truncate individual log lines so the log won't be completely erased by a single large error message. I don't have a strong opinion about it though, this is more a usability question.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+
+  Future<bool> isDebuggable() async {
+    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+    // https://github.com/flutter/flutter/issues/26431
+    // ignore: strong_mode_implicit_dynamic_method
+    final bool result = await channel.invokeMethod('Crashlytics#isDebuggable');
+    return result;
+  }
+
+  Future<String> getVersion() async {
```
collinjackson: This should have Dartdocs

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+      });
+      print(result);
+    }
+  }
+
+  void crash() {
+    throw StateError('Error thrown by Crashlytics plugin');
+  }
+
+  Future<bool> isDebuggable() async {
```
collinjackson: This should have Dartdocs

packages/firebase_crashlytics/ios/Classes/FirebaseCrashlyticsPlugin.m
```diff
@@ -0,0 +1,75 @@
...
+  if ([@"Crashlytics#onError" isEqualToString:call.method]) {
+    NSError* error = [NSError errorWithDomain:call.arguments[@"exception"]
+                                         code:(int)call.arguments[@"code"]
+                                     userInfo:@{
+                                       @"exception" : call.arguments[@"exception"],
+                                       @"stackTrace" : call.arguments[@"stackTrace"]
+                                     }];
+    [[Crashlytics sharedInstance] recordError:error];
+    result(@"Error reported to Crashlytics.");
+  } else if ([@"Crashlytics#isDebuggable" isEqualToString:call.method]) {
```
collinjackson: nit: I would call it isDebugMode instead of isDebuggable

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+///
+/// You can get an instance by calling `Crashlytics.instance`.
+class Crashlytics {
+  static final Crashlytics instance = Crashlytics();
+
+  /// Set to true to have Errors sent to Crashlytics while in debug mode.
+  bool reportInDevMode = false;
+
+  /// Keys to be included with report.
+  @visibleForTesting
```
collinjackson: This shouldn't be `@visibleForTesting` because it's not part of the public API

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+  final Map<String, dynamic> keys = <String, dynamic>{};
+
+  /// Logs to be included with report.
+  @visibleForTesting
```
collinjackson: This shouldn't be `@visibleForTesting` because it's not part of the public API

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -0,0 +1,123 @@
...
+
+    test('isDebuggable', () async {
+      expect(await crashlytics.isDebuggable(), true);
+      expect(log,
+          <Matcher>[isMethodCall('Crashlytics#isDebuggable', arguments: null)]);
+    });
+
+    test('log', () {
+      crashlytics.log('foo');
+      expect(crashlytics.logs.length, 1);
```
collinjackson: Instead of testing private implementation details, you use the mock method call handler.

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -0,0 +1,123 @@
...
+
+    test('getVersion', () async {
+      crashlytics.getVersion();
+      expect(log,
+          <Matcher>[isMethodCall('Crashlytics#getVersion', arguments: null)]);
+    });
+
+    test('setKey', () {
+      crashlytics.setKey('foo', 'bar');
+      expect(crashlytics.keys.length, 1);
```
collinjackson: Instead of testing private implementation details, use the mock method call handler.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    logSize += Uint8List.fromList(msg.codeUnits).length;
+    logs.add(msg);
+    // Remove oldest log till logSize is no more than 64K.
+    while (logSize > 65536) {
+      final String first = logs.removeFirst();
+      logSize -= Uint8List.fromList(first.codeUnits).length;
+    }
+  }
+
+  void setKey(String key, dynamic value) {
```
collinjackson: I would call this `setValue`. It should have a Dartdoc

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    // Check that only 64 keys are set.
+    if (keys.containsKey(key) || keys.length <= 64) {
+      keys[key] = value;
+    }
+  }
+
+  void setBool(String key, bool value) {
```
collinjackson: Add dartdoc

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    setKey(key, value);
+  }
+
+  void setDouble(String key, double value) {
```
collinjackson: Add dartdoc

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    setKey(key, value);
+  }
+
+  void setInt(String key, int value) {
```
collinjackson: Add dartdoc

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    setKey(key, value);
+  }
+
+  void setString(String key, String value) {
```
collinjackson: Add dartdoc

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    setKey(key, value);
+  }
+
+  Future<void> setUserEmail(String email) async {
```
collinjackson: Add dartdoc

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+    // https://github.com/flutter/flutter/issues/26431
+    // ignore: strong_mode_implicit_dynamic_method
+    await channel.invokeMethod(
+        'Crashlytics#setUserEmail', <String, dynamic>{'email': email});
+  }
+
+  Future<void> setUserIdentifier(String identifier) async {
```
collinjackson: Add dartdoc

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+    // https://github.com/flutter/flutter/issues/26431
+    // ignore: strong_mode_implicit_dynamic_method
+    await channel.invokeMethod('Crashlytics#setUserIdentifier',
+        <String, dynamic>{'identifier': identifier});
+  }
+
+  Future<void> setUserName(String name) async {
```
collinjackson: Add dartdoc

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+
+  /// Logs to be included with report.
+  @visibleForTesting
+  final ListQueue<String> logs = ListQueue<String>(15);
+  int logSize = 0;
+
+  bool get isInDebugMode {
+    bool _inDebugMode = false;
+    if (!reportInDevMode) {
+      assert(_inDebugMode = true);
```
collinjackson: For clarity I would recommend separating out the assignment from passing the assertion.

```
    assert((){
      inDebugMode = true;
      return true;
    }());
```

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -0,0 +1,123 @@
...
+    });
+
+    test('log', () {
+      crashlytics.log('foo');
+      expect(crashlytics.logs.length, 1);
+      crashlytics.log('bar');
+      expect(crashlytics.logs.length, 2);
+    });
+
+    test('isInDebugMode', () {
```
collinjackson: I don't think this needs to be tested. It shouldn't be part of the public API

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+  /// Keys to be included with report.
+  @visibleForTesting
+  final Map<String, dynamic> keys = <String, dynamic>{};
+
+  /// Logs to be included with report.
+  @visibleForTesting
+  final ListQueue<String> logs = ListQueue<String>(15);
+  int logSize = 0;
+
+  bool get isInDebugMode {
```
collinjackson: I don't think this needs to be part of the public API. We might want to change it later. Since it's only used in one place, you can just move the logic there.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+    bool _inDebugMode = false;
```
collinjackson: No reason to underscore this, it's a private variable.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+  static const MethodChannel channel =
+      MethodChannel('plugins.flutter.io/firebase_crashlytics');
+
+  Future<void> onError(FlutterErrorDetails details) async {
+    print('Error caught by Crashlytics plugin:');
+    if (isInDebugMode && !reportInDevMode) {
+      print(Trace.format(details.stack).trimRight().split('\n'));
+    } else {
+      // Send logs
+      await sendLogs();
```
collinjackson: Can you merge the sending of logs into the `onError` call below? I'm not sure it needs its own method call.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+
+      // Send keys
+      await sendKeys();
```
collinjackson: Can you merge the keys into the onError call below? I'm not sure it needs its own method call.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
collinjackson: It seems like you're using a lot more method calls than are necessary here. You can just pass the values map over the method channel.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+/// You can get an instance by calling `Crashlytics.instance`.
+class Crashlytics {
+  static final Crashlytics instance = Crashlytics();
+
+  /// Set to true to have Errors sent to Crashlytics while in debug mode.
+  bool reportInDevMode = false;
+
+  /// Keys to be included with report.
+  @visibleForTesting
+  final Map<String, dynamic> keys = <String, dynamic>{};
```
collinjackson: I don't think this should be a public API

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+// BSD-style license that can be found in the LICENSE file.
+part of firebase_crashlytics;
+
+/// The entry point for accessing Crashlytics.
+///
+/// You can get an instance by calling `Crashlytics.instance`.
+class Crashlytics {
+  static final Crashlytics instance = Crashlytics();
+
+  /// Set to true to have Errors sent to Crashlytics while in debug mode.
```
collinjackson: Don't capitalize errors. I would clarify that you mean Flutter debug mode. You might also say that it defaults to false.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+  bool reportInDevMode = false;
```
collinjackson: I would call it `enableInDebugMode`

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
collinjackson: Needs dartdoc. Clarify that this means debuggability of the OEM parts of the app, not the Flutter app.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+
+  /// Keys to be included with report.
+  @visibleForTesting
+  final Map<String, dynamic> keys = <String, dynamic>{};
+
+  /// Logs to be included with report.
+  @visibleForTesting
+  final ListQueue<String> logs = ListQueue<String>(15);
+  int logSize = 0;
```
collinjackson: This is private and should have an underscore

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+  /// Set to true to have Errors sent to Crashlytics while in debug mode.
+  bool reportInDevMode = false;
+
+  /// Keys to be included with report.
+  @visibleForTesting
+  final Map<String, dynamic> keys = <String, dynamic>{};
+
+  /// Logs to be included with report.
+  @visibleForTesting
+  final ListQueue<String> logs = ListQueue<String>(15);
```
collinjackson: This is private and should have an underscore

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,185 @@
...
+
+  Future<String> getVersion() async {
+    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+    // https://github.com/flutter/flutter/issues/26431
+    // ignore: strong_mode_implicit_dynamic_method
+    final String result = await channel.invokeMethod('Crashlytics#getVersion');
+    return result;
+  }
+
+  void log(String msg) {
```
collinjackson: This should have Dartdocs

packages/firebase_crashlytics/example/pubspec.yaml
```diff
@@ -0,0 +1,21 @@
...
+  flutter:
+    sdk: flutter
+
+  cupertino_icons: ^0.1.2
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+
+  firebase_crashlytics:
```
danwulff: Shouldn't this be up in dependencies? (not dev_dependencies?)

packages/firebase_crashlytics/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Add the following imports to your Dart code:
+```dart
+import 'package:firebase_crashlytics/firebase_crashlytics.dart';
+```
+
+Setup `Crashlytics`:
+```dart
+void main() {
+  // Set reportInDevMode to true to see reports while in debug mode
```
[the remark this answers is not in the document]
kroikie: done

packages/firebase_crashlytics/example/lib/main.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:flutter/material.dart';
+
+import 'package:firebase_crashlytics/firebase_crashlytics.dart';
+
+void main() {
+  // Set reportInDevMode to true to see reports while in debug mode
+  // This is only to be used for confirming that reports are being
+  // submitted as expected. It should not intended to be used in production.
```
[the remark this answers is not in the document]
kroikie: done

packages/firebase_crashlytics/README.md
```diff
@@ -0,0 +1,103 @@
...
+  // This is only to be used for confirming that reports are being
+  // submitted as expected. It is not intended to be used in production.
```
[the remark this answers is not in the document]
kroikie: done

packages/firebase_crashlytics/example/lib/main.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:flutter/material.dart';
+
+import 'package:firebase_crashlytics/firebase_crashlytics.dart';
+
+void main() {
+  // Set reportInDevMode to true to see reports while in debug mode
```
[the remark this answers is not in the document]
kroikie: done

packages/firebase_crashlytics/example/lib/main.dart
```diff
@@ -0,0 +1,61 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      home: Scaffold(
+        appBar: AppBar(
+          title: const Text('Crashlytics example app'),
+        ),
+        body: Center(
+          child: Column(
+            children: <Widget>[
```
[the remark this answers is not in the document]
kroikie: Added a button that throws an error.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Used doc from Android, but will have to dig a bit deeper to get more info on what this means in both a Android and iOS context.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: If this has an underscore how will I use it in tests?

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: done

packages/firebase_crashlytics/example/pubspec.yaml
kroikie: Done

hpoul:
I got a bit eager waiting for this to get merged, so I migrating a project from `flutter_crashlytics` to this plugin by including the git branch, and stumbled on a few things.. hope you don't mind a bit very early feedback (feel free to ignore):

1. Imho there should be a way to cause a "real" crash, ie. calling [Crashlytics crash] because if I understand it correctly, crashlytics only ever sends crashes/logs/etc. when the app actually crashes and is subsequently restarted? (or even provide a `isFatal` flag in the `onError` method)
2. It would be nice if logs are also sent periodically, because otherwise if a native part of my application crashes, all "context" logging from the dart side is lost (although i guess i could simply do this manually as a user of the plugin)
3. api nitpicking .. I imagine user identifier, email, name would be usually set from the same location. how about providing one method with named parameters and only one native roundtrip :)

danwulff:
Consider updating `FlutterFire.md` while we're at it? https://github.com/flutter/plugins/blob/master/FlutterFire.md

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  void setKey(String key, dynamic value) {
+    // Check that only 64 keys are set.
+    if (keys.containsKey(key) || keys.length <= 64) {
+      keys[key] = value;
+    }
+  }
+
+  void setBool(String key, bool value) {
```
Salakar: `setInt`, `setBool` & `setDouble` are purely convenience methods on native SDKs (iOS/Android) and internally the SDK's convert these to a string value. 

Perhaps it would be better to simplify this and have a singular method in Dart that accepts a `bool`, `double`, `int` or `String` value and this method converts to a String in dart and calls only one native method (`setString`)

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,190 @@
...
+    setKey(key, value);
+  }
+
+  void setDouble(String key, double value) {
```
Salakar: `setInt`, `setBool` & `setDouble` are purely convenience methods on native SDKs (iOS/Android) and internally the SDK's convert these to a string value. 

Perhaps it would be better to simplify this and have a singular method in Dart that accepts a `bool`, `double`, `int` or `String` value and this method converts to a String in dart and calls only one native method (`setString`)

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,190 @@
...
+    setKey(key, value);
+  }
+
+  void setInt(String key, int value) {
```
Salakar: `setInt`, `setBool` & `setDouble` are purely convenience methods on native SDKs (iOS/Android) and internally the SDK's convert these to a string value. 

Perhaps it would be better to simplify this and have a singular method in Dart that accepts a `bool`, `double`, `int` or `String` value and this method converts to a String in dart and calls only one native method (`setString`)

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,190 @@
...
+        'stackTrace': details.stack.toString(),
+        'stackTraceLines': stackTraceLines,
+        'code': stackTraceLines[0].hashCode
+      });
+      print(result);
+    }
+  }
+
+  void crash() {
+    throw StateError('Error thrown by Crashlytics plugin');
```
Salakar: Should this not delegate to the native SDK methods to ensure a full fatal crash (app terminates) as Crashlytics only submits reports after a termination and subsequent app startup - wouldn't the Dart/Flutter error handler capture this in debug/dev and thus preventing this termination behaviour? And dev/debug is where you'd want to be testing crashes.

**iOS**: `[[Crashlytics sharedInstance] crash];`
**Android**: `(new CrashTest()).crashAsyncTask(50);` (crash off main thread to avoid being captured by error handlers in dev/debug)

packages/firebase_crashlytics/android/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+    }
+    lintOptions {
+        disable 'InvalidPackage'
+    }
+}
+
+dependencies {
+    implementation 'com.google.firebase:firebase-core:16.0.6'
+    implementation 'com.crashlytics.sdk.android:crashlytics:2.9.8'
```
Salakar: No Crashlytics NDK support to capture low-level errors such as in Flutter Engine?

**Dependency**: `implementation 'com.crashlytics.sdk.android:crashlytics-ndk:2.0.5'`


Users `build.gradle` (documentation step only): 

```groovy
crashlytics {
  enableNdk true
}
```

See: https://docs.fabric.io/android/crashlytics/ndk.html

charafau: I feel like that should be separate dependency

kroikie:
@hpoul thanks for the feedback:

1. Non fatal crashes are the best we can do without changes to the underlying SDKs. We considered causing a fatal crash but decided that non fatals would be better till we have better native SDK support.

2. Good point, thanks for the suggestion.

3. It would be nice, but our current approach is to stay as close to the native SDK interfaces as possible. Still a good point and something we can consider later on.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: @Salakar thanks for the feedback. We are trying to keep the plugins interface as close to the native ones as possible. So we will stick with the separate methods for now.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: @Salakar thanks for the feedback. We are trying to keep the plugins interface as close to the native ones as possible. So we will stick with the separate methods for now.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: @Salakar thanks for the feedback. We are trying to keep the plugins interface as close to the native ones as possible. So we will stick with the separate methods for now.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: @Salakar This was considered, however most Flutter level crashes don't result in the underlying native application crashing, so throwing an error here was an attempt to represent the types of crashes likely to happen an a Flutter app and how the report will ultimately be submitted to Crashlytics.

packages/firebase_crashlytics/android/build.gradle
kroikie: @Salakar thanks for the suggestion, we can definitely add this in a future version.

kroikie:
Cirrus is all green, PR has stale CI status. See finished builds here:
https://cirrus-ci.com/build/5687774492491776

packages/firebase_crashlytics/README.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+*Note:* If this section is not completed you will get an error like this:
+```
+java.lang.IllegalStateException:
+Default FirebaseApp is not initialized in this process [package name].
+Make sure to call FirebaseApp.initializeApp(Context) first.
+```
+
+*Note:* When you are debugging on android, use a device or AVD with Google Play services.
```
collinjackson: nit: capitalize Android

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
collinjackson: Still needs dartdocs

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
collinjackson: still needs dartdocs

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
collinjackson: If it's not part of the public API, it's an implementation detail and shouldn't be used in tests.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,213 @@
...
+      final List<String> stackTraceLines =
+          Trace.format(details.stack).trimRight().split('\n');
+      final List<Map<String, String>> stackTraceElements =
+          getStackTraceElements(stackTraceLines);
+      final dynamic result =
+          // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+          // https://github.com/flutter/flutter/issues/26431
+          // ignore: strong_mode_implicit_dynamic_method
+          await channel.invokeMethod('Crashlytics#onError', <String, dynamic>{
+        'exception': details.exceptionAsString(),
```
collinjackson: I think you can just attach the logs and keys here.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,213 @@
...
+        'Crashlytics#setUserName', <String, dynamic>{'name': name});
+  }
+
+  @visibleForTesting
+  Future<void> sendLogs() async {
+    for (int i = 0; i < logs.length; i++) {
+      // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+      // https://github.com/flutter/flutter/issues/26431
+      // ignore: strong_mode_implicit_dynamic_method
+      await channel.invokeMethod('Crashlytics#log', <String, dynamic>{
```
collinjackson: This seems like a lot of unnecessary async waiting, especially if you have big logs. Can we send everything to the native side as an array argument to onError and log it there?

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,213 @@
...
+      final Map<String, dynamic> crashlyticsKey = <String, dynamic>{
+        'key': key,
+        'value': value
+      };
+
+      if (value is int) {
+        // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+        // https://github.com/flutter/flutter/issues/26431
+        // ignore: strong_mode_implicit_dynamic_method
+        await channel.invokeMethod('Crashlytics#setInt', crashlyticsKey);
```
collinjackson: I think it would be cleaner to send over the params map and switch on the type of the data on the native side.

packages/firebase_crashlytics/README.md
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: I think this is in line with how things are handled on the native side.

https://firebase.google.com/docs/reference/android/com/crashlytics/sdk/android/crashlytics-core/CrashlyticsCore.html#log(java.lang.String)

I can update the dart doc to explain more clearly how this works.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Changed to _setValue since it is not part of the public API. Also no Dartdoc needed.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Should this be logged then?

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/ios/Classes/FirebaseCrashlyticsPlugin.m
kroikie: This is the name used by Android so I was sticking to that.

https://firebase.google.com/docs/reference/android/io/fabric/sdk/android/fabric/Fabric#isDebuggable()

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: After moving the logic into `onError` the implementation of this is simpler:

```
bool inDebugMode = false;
    if (!enableInDevMode) {
      assert(inDebugMode = true);
    }
```

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Removed since logs are sent as part of error reporting.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Removed since logs are sent as part of error reporting.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
kroikie: Removed since keys are set as part of error reporting

kroikie: Done

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
kroikie: Done

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
hpoul: sorry to add my two cents here again.. but imho the `sendLogs()` method was a nice way for me as a user of the plugin to periodically send the logs to the native part, because if my app is heavy on native code it's nice to have the flutter logs mingled together with native logs in case my native code crashes... even if not in real time.. i understand that this is not the core use case of this plugin, but i also don't think providing this `sendLogs()` method hurts the plugin either..

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
kroikie: No longer testing log since logs are submitted with error report.

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
kroikie: Done

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
kroikie: Done

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
kroikie: Removed this test since keys are submitted with error report.

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
kroikie: Removed this test since logs are submitted with error report.

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
kroikie: Removed this test since keys are sent with error report.

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
```diff
@@ -0,0 +1,98 @@
...
+        context: 'foo context',
+        informationCollector: (StringBuffer information) {
+          information.writeln('foo information');
+        },
+      );
+      crashlytics.enableInDevMode = true;
+      crashlytics.log('foo');
+      await crashlytics.onError(details);
+      expect(log[0].method, 'Crashlytics#onError');
+      expect(log[0].arguments['exception'], 'foo exception');
```
collinjackson: I think it would be more readable and consistent with other tests to write this as a single expect comparing log with an array of maps.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -0,0 +1,202 @@
...
+  /// TODO(kroikie): Clarify what this means in context of both Android and iOS.
+  Future<bool> isDebuggable() async {
+    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+    // https://github.com/flutter/flutter/issues/26431
+    // ignore: strong_mode_implicit_dynamic_method
+    final bool result = await channel.invokeMethod('Crashlytics#isDebuggable');
+    return result;
+  }
+
+  /// Returns the Kit Version.
```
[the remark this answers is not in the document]
kroikie: Done

packages/firebase_crashlytics/test/firebase_crashlytics_test.dart
collinjackson: It sounds like StackTrace.current might make it hard to use the isMethodCall matcher, in which case this is fine.

kroikie: After chat with @collinjackson we will leave this as is, since `Stacktrace.current` will likely not be the same on each execution.

PR #1081, opened 2019-01-15
Additional sample rates for sensor events

LeonardGreulich:
### Three options for all three sensor events:
**SampleRate.low** results in 15 sensor events per second on both platforms.
**SampleRate.medium** results in 50 sensors events per second on both platforms.
**SampleRate.high** results in 100 sensors events per second on iOS and 120 on Android.

A sample rate is set before the listening with, e.g., **setSensorsSampleRate(SampleRate.medium);**.

### DIFF
--- a/packages/sensors/README.md
+++ b/packages/sensors/README.md
@@ -20,4 +20,8 @@ accelerometerEvents.listen((AccelerometerEvent event) {
 gyroscopeEvents.listen((GyroscopeEvent event) {
  // Do something with the event.
 });
+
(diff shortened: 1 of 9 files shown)


motiongestures:
I am sorry to say, but the sensors plugin is quite useless without the ability to control the sample rate. This PR is not going far enough, it should allow fine control over the sample rate, much like Android itself does. 
But even this would be of great help. So, please, consider this feature as critical. Because it is.

PR #1082, opened 2019-01-15
[IAP] Clean up Dart unit tests

mklim:
Rename `PlatformViewsController` to `StubInAppPurchasePlatform` to more
accurately describe what it's stubbing. Remove the overcomplicated
`Channel` class and just set a mock handler on the `const MethodChannel`
used by the entire plugin itself.

### DIFF
--- a/packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
+++ b/packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
@@ -19,7 +19,7 @@ class BillingClient {
   BillingClient() {
     _channel.setMethodCallHandler(_callHandler);
   }
-  MethodChannel _channel = Channel.instance;
+  final MethodChannel _channel = channel;
 
   // Occasionally methods in the native layer require a Dart callback to be
   // triggered in response to a Java callback. For example,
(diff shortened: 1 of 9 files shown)


packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -19,7 +19,7 @@ class BillingClient {
   BillingClient() {
     _channel.setMethodCallHandler(_callHandler);
   }
-  MethodChannel _channel = Channel.instance;
+  final MethodChannel _channel = channel;
```
amirh: nit: why do we need this member? we can just use channel directly?

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -5,7 +5,7 @@ import '../channel.dart';
 
 /// A wrapper around [`SKPaymentQueue`](https://developer.apple.com/documentation/storekit/skpaymentqueue?language=objc).
 class SKPaymentQueueWrapper {
-  static MethodChannel _channel = Channel.instance;
+  static MethodChannel _channel = channel;
```
amirh: same here

packages/in_app_purchase/test/fake_in_app_purchase_platform.dart
```diff
@@ -1,9 +1,10 @@
 import 'dart:async';
 import 'package:flutter/services.dart';
 
-class FakePlatformViewsController {
+class FakeInAppPurchasePlatform {
```
amirh: As discussed, we think that this is ore of a stub than a fake, lets name appropriately.

PR #1085, opened 2019-01-16
Url launcher refactor

cyanglaz:
Refactored url launcher OBJC/JAVA code.

In OBJC, updated the block local variable to copy property to prevent potential issues. Added OS version check. Performed some method extraction. 

In Java, performed some method extraction and clean up.


packages/url_launcher/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.1.0+1
+
+* Refactored the JAVA and OBJC code. Replaced instance variables with properties in OBJC.
```
cbracken: nit: 'Obj-C' or even better 'Objective-C'.

cbracken: Also: s/JAVA/Java/

PR #1088, opened 2019-01-16
add deprecated_member_use_from_same_package to pass analyzer

cyanglaz:


dnfield:
Can you add something to your message about why this was necessary (change upstream in analyzer)?  Might be helpful for archaeologists ;)

What do we think of just removing these deprecated members?  They don't do anything anyway, and they've been marked as deprecated since November.

That would require a rev bump on this.

/cc @amirh @kroikie

dnfield:
Ahhh I may be forgetting something - it's because it's deprecated on the platform side.

I suppose we should just wait until it's removed from there to remove it here.

LGTM.

PR #1089, opened 2019-01-17
Fix a crash when selecting downloaded images on certain devices

cyanglaz:
Some devices use different path for downloaded images. Those devices will crash when trying the access the public_download URI. 
Directly use default downloaded image URI instead of hard coding the string fixes the crash.
https://github.com/flutter/flutter/issues/21863


packages/image_picker/pubspec.yaml
```diff
@@ -5,7 +5,7 @@ authors:
   - Flutter Team <flutter-dev@googlegroups.com>
   - Rhodes Davis Jr. <rody.davis.jr@gmail.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/image_picker
-version: 0.4.12
+version: 0.4.13
```
amirh: This should be 0.4.12+1

PR #1090, opened 2019-01-18
[IAP] Generate boilerplate serializers

mklim:
Brings in `json_serializable` as a build step to generate the Dart
platform channel parsing boilerplate. This dependency is still very
experimental and may be abandoned if the extra build step ends up being
a maitenance burden greater than maintaining the boilerplate. At this
point I expect it to be worth the trouble given the sheer amount of data
classes and deserialization that's going to happen in this plugin.

Also converts `SkuType` and `BillingResponse` to enums.

### DIFF
--- a/packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
+++ b/packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
@@ -115,11 +115,12 @@ class BillingClient {
 /// to call back on `BillingClient` disconnect.
 typedef void OnBillingServiceDisconnected();
 
-/// Wraps
-/// [`BillingClient.BillingResponse`](https://developer.android.com/reference/com/android/billingclient/api/BillingClient.BillingResponse),
-/// possible status codes.
+/// Possible `BillingClient` response statuses.
 ///
-/// See the `BillingResponse` docs for an explanation of the different constants.
+/// Wraps
+/// [`BillingClient.BillingResponse`](https://developer.android.com/reference/com/android/billingclient/api/BillingClient.BillingResponse).
+/// See the `BillingResponse` docs for more explanation of the different
+/// constants.
 enum BillingResponse {
   // WARNING: Changes to this class need to be reflected in our generated code.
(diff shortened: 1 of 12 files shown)


packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -118,56 +120,54 @@ typedef void OnBillingServiceDisconnected();
...
-  static const BillingResponse OK = BillingResponse._(0);
-  static const BillingResponse USER_CANCELED = BillingResponse._(1);
-  static const BillingResponse SERVICE_UNAVAILABLE = BillingResponse._(2);
-  static const BillingResponse BILLING_UNAVAILABLE = BillingResponse._(3);
-  static const BillingResponse ITEM_UNAVAILABLE = BillingResponse._(4);
-  static const BillingResponse DEVELOPER_ERROR = BillingResponse._(5);
-  static const BillingResponse ERROR = BillingResponse._(6);
-  static const BillingResponse ITEM_ALREADY_OWNED = BillingResponse._(7);
-  static const BillingResponse ITEM_NOT_OWNED = BillingResponse._(8);
+enum BillingResponse {
```
amirh: Please have the summary line say what this is.

packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -118,56 +120,54 @@ typedef void OnBillingServiceDisconnected();
...
-  @override
-  bool operator ==(dynamic other) => other is SkuType && other._type == _type;
+enum SkuType {
+  // WARNING: Changes to this class need to be reflected in our generated code.
+  // Run `flutter packages pub run build_runner watch` to rebuild and watch for
+  // further changes.
 
   /// A one time product. Acquired in a single transaction.
-  static const SkuType INAPP = SkuType._("inapp");
+  inapp,
```
amirh: Do we not need `@JsonValue` here?

PR #1091, opened 2019-01-19
rev all compileSdkVersion to 28

dnfield:
Should fix CI

cc @mklim

### DIFF
--- a/.ci/Dockerfile
+++ b/.ci/Dockerfile
@@ -1,0 +1,6 @@
+
+FROM cirrusci/flutter:latest
+
+RUN yes | sdkmanager \
+    "platforms;android-27" \
+    "build-tools;27.0.3"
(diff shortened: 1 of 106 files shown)


dnfield:
@fkorotkov this is the PR I actually meant to tag you on.  Is this container image linked to the one we make?

fkorotkov:
@dnfield yeah, but there is no tools 27 on this image: https://github.com/cirruslabs/docker-images-android/blob/master/sdk/28/Dockerfile#L6-L8

I wonder where it gets them from. And also why only some plugins are failing and others not.

dnfield:
I suspect it's getting pulled in by some other dependency in the tree that's actually specifying `buildToolsVersion`.  Looks like we took that out at some point in the past - trying to just add it back in.

dnfield:
I'm really not sure what's making some of these guys hold on to Android 27.  They shouldn't be.

@fkorotkov  Do you think it would be a problem to just add SDK 27 and Buildtools 27.0.3 to the Android image for Cirrus?

dnfield:
Closing in favor of #1104

PR #1093, opened 2019-01-19
[image_picker] Fixed a crash that would occur when called in quick succession on Android

wouterhardeman:
Resolves https://github.com/flutter/flutter/issues/19819

When the `image_picker` is called before another call is finished it should return an error with the `MethodResult` object that came with the last method call. But instead it uses the `MethodResult` object from the first method call. When finishing up the first method call, the `pendingResult` (`MethodResult` object of the first call) property will already be set to null.

### DIFF
--- a/packages/image_picker/CHANGELOG.md
+++ b/packages/image_picker/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.0+1
+
+* Fix a crash when user calls the plugin in quick succession on Android.
+
 ## 0.5.0
 
 * **Breaking change**. Migrate from the deprecated original Android Support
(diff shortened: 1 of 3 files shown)


packages/image_picker/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.4.12+2
+
+* Fix a crash when user calls the plugin in quick succession on Android
```
cyanglaz: Nit: . at the end of sentence.

cyanglaz:
The original post of the same issue can be found here https://github.com/flutter/flutter/issues/19819
Please refer this PR to https://github.com/flutter/flutter/issues/19819 instead so we can better tracking the issue.

cyanglaz:
LGTM!

PR #1094, opened 2019-01-20
[video_player]Get duration from underlying asset

cbenhagen:
Fixes https://github.com/flutter/flutter/issues/26817

### DIFF
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.8.1
+
+* iOS: Fix race condition when getting duration
+
 ## 0.8.0
 
 * Android: Upgrade ExoPlayer to 2.9.1
--- a/packages/video_player/ios/Classes/VideoPlayerPlugin.m
+++ b/packages/video_player/ios/Classes/VideoPlayerPlugin.m
@@ -225,7 +225,7 @@ static void* playbackBufferFullContext = &playbackBufferFullContext;
 }
 
 - (int64_t)duration {
-  return FLTCMTimeToMillis([[_player currentItem] duration]);
+  return FLTCMTimeToMillis([[[_player currentItem] asset] duration]);
 }
 
 - (void)seekTo:(int)location {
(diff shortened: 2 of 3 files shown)


iskakaushik:
Can you describe why getting the duration from the underlying asset fixes this issue?

cbenhagen:
The value of duration property will be reported as kCMTimeIndefinite until the duration of the underlying asset has been loaded.  So we can either wait for AVPlayerItemStatusReadyToPlay or get the duration from the underlying asset.

cbenhagen:
Please also check the the example in the linked issue: https://github.com/flutter/flutter/issues/26817

iskakaushik:
After looking at the docs at: https://developer.apple.com/documentation/avfoundation/avplayeritem/1389386-duration?language=objc, it seems to me that getting it from the asset also doesn't guarantee correctness as seen in the question at: https://stackoverflow.com/questions/23874574/avplayer-item-get-a-nan-duration.

Should we follow the doc guidelines and take one of:

1. Wait until the status of the AVPlayerItem is AVPlayerItemStatusReadyToPlay.

or

2. Register for key-value observation of the duration property, requesting the initial value. If the initial value is reported as kCMTimeIndefinite, the AVPlayerItem will notify you of the availability of the item's duration via key-value observing as soon as its value becomes known.

@cbenhagen let me know what you think.

cbenhagen:
I am not sure what would be the best option. My guess would be to wait for `AVPlayerItemStatusReadyToPlay`.

iskakaushik:
@cbenhagen agreed, can you make that change to this PR?

cbenhagen:
Hmm in the current code it actually looks like we are correctly waiting for `AVPlayerItemStatusReadyToPlay`. I am also unable to reproduce the issue with the code I posted in https://github.com/flutter/flutter/issues/26817. I guess I'll close both this PR and the issue.

cbenhagen:
Sorry I need to reopen this. While in the test code I was not being able to reproduce, in a full app I was. Need to check why. The fix in here, or another solution still needs to be applied.

recastrodiaz:
Looks like this may have already been fixed in master with https://github.com/flutter/plugins/pull/1554/files#diff-e534c460662d39cc715c4eed77a31c3fR308

I'm not seeing any duration issues on iOS anymore.

PR #1096, opened 2019-01-21
[firebase_database]Return error message from DatabaseError#toString()

dotdoom:


dotdoom:
`build-apks+java-test` tests are failing because

```
Failed to install the following Android SDK packages as some licences have not been accepted.
```

so I'm positive those failures are unrelated to this PR.

PR #1098, opened 2019-01-22
Specify sampling rate for sensor events

LeonardGreulich:
Related issue: https://github.com/flutter/flutter/issues/25971

I reopen this PR as this feature should be regarded critical (closed PR: https://github.com/flutter/plugins/pull/1081). Without the ability to specify the sample rate of the accelerometer and gyroscope sensor, the sensors plugin has very limited utility. Furthermore, the current implementation results in ~15 events per second on Android and 100 events per second on iOS which is an unreasonable gap. The added changes on Java- and Objective-C side are very minimal and if there is no testing for the native code so far, I do not think it is very rational to postpone this feature due to the missing test harness.

A sample rate is set in Hz (sensor events per second) before the listening with, e.g., **setSensorsSampleRate(50)**;. If no sample rate is set, the default is 15.


a83988029:
I need rate

a83988029:
setRate() only works once with your code, but thanks to your code though.

PR #1100, opened 2019-01-23
[android_alarm_manager][HX] Add support for setAndAllowWhileIdle

Hixie:


Hixie:
(This is just a strawman proposal. It would need tests before landing.)

PR #1101, opened 2019-01-23
[android_intent] Added 'action_app' action and getIntentExtras, getIntentData

eligt:
The "action_app" special action allows to create an Intent that launches the app's launcher activity (which for most Flutter apps, it's the Activity handling the Flutter app itself).

Also added methods `getIntentExtras` and `getIntentData` which respectively let you retrieve the current Intent's extras [arguments] and data.

These two features combined are useful when your app is divided between a foreground component (the main app running the Flutter view) and a background component (a service that runs in the background).

With these features the background component can easily trigger the main app to open (whether it was running or not) and pass extra arguments to the Intent. The main app can then load up and use these extra arguments to do some custom processing.

Please note that if your app is already opened and you launch a "action_app" intent, you need a bit of extra logic in your MainActivity to ensure that arguments are refreshed from the new Intent, something like this:

```java
  @Override
  protected void onNewIntent(Intent intent) {
      super.onNewIntent(intent);
      Bundle bundle = intent.getExtras();
      getIntent().putExtras(bundle);
  }
```


eligt:
I signed the CLA

eligt:
It looks like a bunch of checks are failing, but all I can see for an error is:
```
Failed to install the following Android SDK packages as some licences have not been accepted.
        platforms;android-27 Android SDK Platform 27
```
I don't think this is related to my changes?

eligt:
Re-implemented in https://github.com/flutter/plugins/pull/2308

PR #1102, opened 2019-01-23
cleanup

cyanglaz:


cbracken:
lgtm

packages/url_launcher/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.1.0+1
+
+* Clean up.
```
amirh: Not sure I'll know what to expect when reading this in a change log, maybe worth being a little more explicit here.

And somewhat related - I think that it's generally helpful when the PR description / commit message say what was done, in this case I think "Bump url_lanuncher's version" would have been a good description (specifically I pulled up this PR as it triggered a build failure, with a more detailed description I could probably know its not this PR fault just from the message)

PR #1103, opened 2019-01-23
[WIP] Migrate plugins to androidx

mklim:


dnfield:
I tested this locally using AndroidX plugin with current Flutter and it looks good.  

Should we bump the gradle plugins while we're at it to 3.3.0?  As in https://github.com/flutter/flutter/pull/26913

It looks like this all works and builds, so maybe we could do that as a separate patch.

Either way, LGTM

packages/android_intent/example/android/app/build.gradle
```diff
@@ -37,7 +37,7 @@ android {
         targetSdkVersion 27
```
dnfield: This should be 28 (here and elsewhere)

jonahwilliams:
Does this conflict with our mandate to keep plugins working with the last stable flutter release?

mklim:
> Does this conflict with our mandate to keep plugins working with the last stable flutter release?

I've spot checked some of the example apps with `flutter stable` and what I've seen appears to be working fine. The CI failure on the latest commits looks like something unrelated to this PR, I'm also seeing it on another PR that only updated markdown documentation and it's referencing a problem with `/home/cirrus/sdks/flutter/packages/flutter_tools/gradle/flutter.gradle`.

PR #1108, opened 2019-01-23
Remove 'init' static method

bparrishMines:
I was curious to whether 'init' was necessary. I can add this call somewhere else if it required.


packages/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.2.9+2
+
+* Remove 'init' call to avoid MissingPluginException when testing.
```
sigurdm: This description is hard to read without more context.
 
I don't think the init call is exposed in the public api, and therefore its hard for an end-user to know what are the consequences of  the change.

Also it is not obvious why having the init-call leads to a MissingPluginException when testing

bparrishMines:
After some testing with the example app, I don't receive an exception on iOS or Android when removing the `init` call. 

My steps were:
1. Start app
2. open camera
3. hot restart
4. open camera

I believe it is because the following null checks were added in the `initialize` calls:

Android:
https://github.com/flutter/plugins/blob/master/packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java#L190
iOS:
https://github.com/flutter/plugins/blob/master/packages/camera/ios/Classes/CameraPlugin.m#L678

However, removing the `init` call would cause the camera to remain open and leak when using hot restart. This is not ideal, but I don't think it's a major issue since it will only happen in development with hot restart.

bparrishMines:
For handling hot restart, I think we want to add something to the engine/framework that would inform plugins that there was a hot restart. This solution would probably be a stretch goal though. A issue was created awhile ago that addressed this https://github.com/flutter/flutter/issues/10437. I saw that you had this same solution for the video_player too.

I don't know if adding a global `camera.dart:initialize()` would be good just for hot restart. That would basically require the user to call `camera.dart:initialize()` just so they could use a feature they may not use. But, it may be the only work around right now.

sigurdm:
> I don't know if adding a global camera.dart:initialize() would be good just for hot restart. That would basically require the user to call camera.dart:initialize() just so they could use a feature they may not use. But, it may be the only work around right now.

Well if it doesn't do anything else they are not required to use it. The choice alone might be confusing though.

Not having implicit code run on initialization before main() is both a strength (everything is explicit) but sometimes also a constraint of the dart language.

PR #1110, opened 2019-01-23
Corrected CameraPosition

rubberbird:
initialCameraPosition no longer exists, no inside GoogleMapsOptions under cameraPosition.


amirh:
the initialCameraPosition was added in https://github.com/flutter/plugins/commit/5f9affc8edc2685cd34d62950244e5edc81363b9#diff-2c5ffa7b6dd1fa7c98f729362b4fa5ceR10 (same change removed the options field).

Any chance you're using the updated sample with with an older version of the plugin.

PR #1115, opened 2019-01-24
Migrate remaining plugins to AndroidX

mklim:
The original Android support libraries have been deprecated in favor of
AndroidX. This migration shouldn't affect behavior in a user noticeable
way, but does require plugin users to also migrate their app. See
[migrating to AndroidX](https://developer.android.com/jetpack/androidx/migrate).

Fixes flutter/flutter#23995.


mklim:
Thanks for the heads up @Botanium. The root cause of this error message is because this PR introduces a breaking change to the plugins. The plugins were updated from the deprecated Android support libraries to AndroidX (flutter/flutter#23995). Unfortunately this update also requires a manual migration in apps using the plugins in cases where the apps are relying on the original support libraries and/or aren't compiling to SDK 28 already. From the pastebin error, it looks like you'll need to migrate your app to use AndroidX too to keep using the latest major versions of all the plugins.

(This PR also had a bug where semantic versioning wasn't incremented correctly in the plugins that were already `>= 1.0.0`, so pub was incorrectly treating this as if it wasn't a breaking change in some cases. That's been fixed now, so if you've been pinning your app to the latest major version to avoid breaking changes like this you should get the latest revert from pub and shouldn't be seeing the error anymore.)

The gradle error is not helpful here so unfortunately this is really hard to debug when it is breaking. We already have a bug on improving our error handling for Gradle (flutter/flutter#14818), but I noted on it that fixing our error handling would help this case too.

If you want to keep using the latest major versions of the plugins, [the official Android migration guide for AndroidX is here](https://developer.android.com/jetpack/androidx/migrate). Android Studio provides an automated migration tool for apps that will probably make migrating easiest (also described on that page). You can import a flutter app into AS by hitting "Open an existing android studio project" on the IDE's opening splash screen and then opening `<my_app_root>/android/build.gradle`.

If you don't want to migrate your app off of the deprecated support libraries, then for now you can avoid this by staying pinned to the last major version and avoiding this change.

PR #1120, opened 2019-01-26
A few additional Androix dependencies

edman:
Apparently some dependencies like android.support.annotation were implicitly available, but that doesn't seem to be true anymore with AndroidX.

Tracked in flutter/flutter#23995.
Related to https://github.com/flutter/plugins/pull/1115


edman:
By the way, since AndroidX migration is a breaking change shouldn't we update the major version of affected libraries to respect semantic versioning?

edman:
Hi @mklim, could you have a look at this PR? I needed to make these changes to get your AndroidX migration working in my machine.

PR #1122, opened 2019-01-28
Add an UserAgent in  webView_flutter Plugin

tefunamu16:
Add `userAgent` in `webview_flutter` plugin.
Because I want to set an userAgent in `webview_flutter`.


tefunamu16:
I signed it!

PR #1126, opened 2019-01-28
IAP add payment translators in objc

cyanglaz:


packages/in_app_purchase/ios/Classes/FIAObjectTranslator.m
```diff
@@ -83,3 +83,97 @@ - (NSDictionary *)toMap {
...
+    [map setObject:@(self.downloadState) forKey:@"state"];
+  }
+  return map;
+}
+
+@end
+
+@implementation NSError (Coder)
+
+- (NSDictionary *)toMap {
```
cyanglaz: @amirh @mklim `SKDownload` and `SKTransaction` object contains an NSError as a property. Do we have a better way of sending the error object to dart or we just make it a map and let the dart side parse the map as I am doing here.

PR #1129, opened 2019-01-28
android_alarm_manager background execution bug fixes

bkonyi:
* Updated android_alarm_manager to fix issues with "background start not
allowed" when attempting to start AlarmService while backgrounded.
  * Switched to using a combination of BroadcastReceiver and
  JobIntentService from using a Service (launching a Service from an
  alarm Intent causes the "Background start not allowed" on Android 8+)

* Alarm events received before the AlarmService isolate is finished
initializing are now queued and processed once the isolate is
initialized.

* Updated CHANGELOG.md and README.md with new setup steps.

* Bumped version to 0.4.0.

Fixes flutter/flutter#26846.

### DIFF
--- a/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
+++ b/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
@@ -6,11 +6,9 @@ package io.flutter.plugins.androidalarmmanager;
 
 import android.app.AlarmManager;
 import android.app.PendingIntent;
-import android.app.Service;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
-import android.os.IBinder;
 import android.util.Log;
 import androidx.core.app.JobIntentService;
 import io.flutter.plugin.common.MethodChannel;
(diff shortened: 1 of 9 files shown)


packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
```diff
@@ -6,30 +6,68 @@
...
+    if (!sStarted.get()) {
+      SharedPreferences p = context.getSharedPreferences(SHARED_PREFERENCES_KEY, 0);
+      long callbackHandle = p.getLong(CALLBACK_HANDLE_KEY, 0);
+      startAlarmService(context, callbackHandle);
+    }
+  }
+
+  // Schedule the alarm to be handled by the AlarmService.
+  public static void enqueueAlarmProcessing(Context context, Intent alarmContext) {
+    enqueueWork(context, AlarmService.class, 1984, alarmContext);
```
zanderso: what is 1984?

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
```diff
@@ -72,6 +72,7 @@ public void onMethodCall(MethodCall call, Result result) {
 
   private void startService(JSONArray arguments) throws JSONException {
     long callbackHandle = arguments.getLong(0);
+    AlarmService.setCallbackDispatcher(mContext, callbackHandle);
```
zanderso: Can these two be folded together into one call?

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
bkonyi: Oh, just a fixed job ID. See the docs for [enqueueWork](https://developer.android.com/reference/android/support/v4/app/JobIntentService#enqueuework_3). I can add a comment here or define a constant for this if you'd prefer that?

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
bkonyi: They could be, but `AlarmService.startAlarmService` is also called from `AlarmService.onCreate` which retrieves this callback handle anyway, so having them separate avoids setting the callback dispatcher handle when it hasn't changed.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
zanderso: Yeah, let's define a constant like with the constant strings above.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
zanderso: ack.

bparrishMines:
cc @mklim

PR #1130, opened 2019-01-29
WebView JavasScript channels Android implementation.

amirh:
Platform implementation of the method channel API for adding and removing JavaScript channels.

#1116 adds the Dart side support, the current PR will land first.

flutter/flutter#24837

### DIFF
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -9,6 +9,7 @@ import io.flutter.plugin.common.MethodChannel;
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler;
 import io.flutter.plugin.common.MethodChannel.Result;
 import io.flutter.plugin.platform.PlatformView;
+import java.util.List;
 import java.util.Map;
 
 public class FlutterWebView implements PlatformView, MethodCallHandler {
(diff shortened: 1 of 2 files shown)


packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -18,13 +19,20 @@
   @SuppressWarnings("unchecked")
   FlutterWebView(Context context, BinaryMessenger messenger, int id, Map<String, Object> params) {
     webView = new WebView(context);
+
+    methodChannel = new MethodChannel(messenger, "plugins.flutter.io/webview_" + id);
+    methodChannel.setMethodCallHandler(this);
+
+    applySettings((Map<String, Object>) params.get("settings"));
+
+    if (params.containsKey("javascriptChannelNames")) {
```
mklim: nit, optional: Consider breaking out the "javascriptChannelNames" magic string into a private constant on this class or a named variable. It'll prevent typo problems with getting the value.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/JavaScriptChannel.java
```diff
@@ -0,0 +1,32 @@
...
+ * up.
+ *
+ * <p>Exposes a single method named `postMessage` to JavaScript, which sends a message over a method
+ * channel to the Dart code.
+ */
+class JavaScriptChannel {
+  private final MethodChannel methodChannel;
+  private final String javaScriptChannelName;
+
+  JavaScriptChannel(MethodChannel methodChannel, String javaScriptChannelName) {
```
mklim: nit: Can you also add docs explaining these params? I think it's just tricky enough that it's worth saying what they do here.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
amirh: done

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/JavaScriptChannel.java
amirh: done

PR #1132, opened 2019-01-29
Use string to save double for shared_preference(Android).

kangwang1988:
Use string to save double for shared_preference instead of converting it to float.
This will fix: https://github.com/flutter/flutter/issues/27222

### DIFF
--- a/packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
+++ b/packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
@@ -32,7 +32,7 @@ public class SharedPreferencesPlugin implements MethodCallHandler {
   // Fun fact: The following is a base64 encoding of the string "This is the prefix for a list."
   private static final String LIST_IDENTIFIER = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu";
   private static final String BIG_INTEGER_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBCaWdJbnRlZ2Vy";
-  private static final String DOUBLE_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBTeCcXmYaGwHtE";
+  private static final String DOUBLE_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBEb3VibGUu";
 
   private final android.content.SharedPreferences preferences;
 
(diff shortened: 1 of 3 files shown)


packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
```diff
@@ -88,6 +89,9 @@ private String encodeList(List<String> list) throws IOException {
           } else if (stringValue.startsWith(BIG_INTEGER_PREFIX)) {
             String encoded = stringValue.substring(BIG_INTEGER_PREFIX.length());
             value = new BigInteger(encoded, Character.MAX_RADIX);
+          } else if (stringValue.startsWith(DOUBLE_PREFIX)) {
```
mehmetf: Please double check what happens to existing keys that are storing float values.

kangwang1988: Should I add the float check logic now?
I see there is 'setDouble' logic, no 'setFloat' logic, neither in iOS nor android.

mehmetf: App which get updated need to be able to reuse the values that were saved with an older version of the app.

My comment was about making sure an app which used `setDouble() getDouble()` APIs work as intended before and after this change. If it does not, this is a breaking change and should be modified to not break the values saved in the older version.

My gut feeling is that it is not a breaking change since if you read Float back then you will return it to Flutter as a double. However, it would be nice to double check to be sure.

kangwang1988: I've double checked the old existing value will remain after this patch.
However, flutter/flutter#27222 will remain for those values.

packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
```diff
@@ -32,6 +32,7 @@
   // Fun fact: The following is a base64 encoding of the string "This is the prefix for a list."
   private static final String LIST_IDENTIFIER = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu";
   private static final String BIG_INTEGER_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBCaWdJbnRlZ2Vy";
+  private static final String DOUBLE_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBTeCcXmYaGwHtE";
```
dariotrombello: When decoding that it returns `This is the prefix for Sx'醆;D`. Seems like it was wrong encoded. Try using this: `VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGRvdWJsZS4=`

kangwang1988: nice shot.
Resolved.

kangwang1988:
@mehmetf
All issues verified/resolved.

mehmetf:
Could you please update CHANGELOG and the version in pubspec.yaml as well?

kangwang1988:
> Could you please update CHANGELOG and the version in pubspec.yaml as well?

Done
@mehmetf

PR #1133, opened 2019-01-30
Added support for persisting alarms across reboots

bkonyi:
Android doesn't automatically reschedule alarms after a reboot, so
alarms set using this plugin would have to be scheduled manually after
each reboot. With this change, alarms scheduled with the 'rescheduleOnReboot'
flag set to true will be rescheduled at boot.


Hixie:
Thanks!

I'll defer to someone more familiar with the plug-in architecture do a thorough review.

PR #1134, opened 2019-01-30
Updated connectivity to singleton

cyanglaz:
This is a fix for https://github.com/flutter/flutter/issues/21709

Creating a second instance of the Connectivity object can cause the second instance overriding the first instance's stream channel. We can force the class to be a singleton because this class is designed to work as a singleton.

### DIFF
--- a/packages/connectivity/lib/connectivity.dart
+++ b/packages/connectivity/lib/connectivity.dart
@@ -20,6 +20,16 @@ const EventChannel _eventChannel =
     EventChannel('plugins.flutter.io/connectivity_status');
 
 class Connectivity {
+  /// Constructs a singleton instance of [Connectivity]
+  factory Connectivity() {
+    return _singleton;
+  }
+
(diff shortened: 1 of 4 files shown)


packages/connectivity/lib/connectivity.dart
```diff
@@ -20,6 +20,19 @@ const EventChannel _eventChannel =
...
 
 class Connectivity {
+  /// Constructs a singleton instance of [Connectivity].
+  ///
+  /// [Connectivity] is designed to work as a singleton.
+  // When a second instance is created, the first instance will not be able to listen to the
+  // EventChannel because it is overridden. Forcing the class to be a singleton class can prevent
+  // misusage of creating a second instance from a programmer.
+  factory Connectivity() {
+    return _singleton;
```
mehmetf: Nit: Let's initialize this lazily rather than eagerly.

PR #1135, opened 2019-01-30
FIx issue with calculating iOS image orientation in certain special c…

idealopamp:
…ases (#26786)

Uses accelerometer data to determine device orientation when the photo is taken.  Fixes flutter/flutter#26786 .

### DIFF
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.0+2
+
+* Fix issue with calculating iOS image orientation in certain edge cases    
+
 ## 0.3.0+1
 
 * Remove initial method call invocation from static camera method.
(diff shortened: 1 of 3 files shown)


idealopamp:
One known issue with the current solution: You may notice that if you shake the device up/down in portrait or left/right in landscape that the orientation sometimes comes out wrong. I think this approach might be more robust against unusual amounts of acceleration, e.g. the user is in a bumpy car ride, if one were to take an average/smoothing of the accelerometer metrics over the past few seconds, but that requires a bit more work to implement, so just starting with this to see what you think.

idealopamp:
> One known issue with the current solution: You may notice that if you shake the device up/down in portrait or left/right in landscape that the orientation sometimes comes out wrong. I think this approach might be more robust against unusual amounts of acceleration, e.g. the user is in a bumpy car ride, if one were to take an average/smoothing of the accelerometer metrics over the past few seconds, but that requires a bit more work to implement, so just starting with this to see what you think.

PR #1137, opened 2019-01-30
[webview_flutter] Added basic support to load html data on both platform

awazgyawali:
As title suggests, added basic support to load html data on both platform.


amirh:
Thanks for the contribution, and sorry for the long time it took me to respond!

I'd be hesitant to add `htmlData` as a widget parameter, e.g what does it mean after `loadUrl` is called on a widget that was built with `htmlData`? (I think this should be an edge trigger rather than a level trigger, you can find some information about these kind of issues in this [article](https://medium.com/flutter-io/api-design-tip-for-complex-flutter-widgets-7794e9c7144c))

We'd also want an e2e test before we can land this.

I filed https://github.com/flutter/flutter/issues/32827 to track this issue and mentioned this PR there.
I'm going to close this PR, if you're still interested in working on this feel free to re-open (or send a new one) after changing it to an edge trigger(I put up an API proposal on the issue) and adding tests.

PR #1138, opened 2019-01-30
Add a gradle warning to the AndroidX plugins

mklim:
This warning is only printed once per plugin on build failure.

flutter/flutter#27106

### DIFF
--- a/packages/android_alarm_manager/android/build.gradle
+++ b/packages/android_alarm_manager/android/build.gradle
@@ -4,8 +4,7 @@ gradle.buildFinished { buildResult ->
     if (buildResult.failure && !rootProject.ext.has(ANDROIDX_WARNING)) {
         println '         *********************************************************'
         println 'WARNING: This version of ' + PLUGIN + ' will break your Android build if it or its dependencies aren\'t compatible with AndroidX.'
-        println '         https://goo.gl/y9dCcU describes AndroidX and how to migrate to it using Android Studio.'
-        println '         https://goo.gl/n2E4mD describes how to open a Flutter app\'s Android files for refactoring in Android Studio.'
+        println '         See https://goo.gl/CP92wY for more information on the problem and how to fix it.'
         println '         This warning prints for all Android build failures. The real root cause of the error may be unrelated.'
         println '         *********************************************************'
         rootProject.ext.set(ANDROIDX_WARNING, true);
--- a/packages/android_alarm_manager/android/build.gradle
+++ b/packages/android_alarm_manager/android/build.gradle
@@ -1,12 +1,15 @@
 def PLUGIN = "android_alarm_manager";
 def ANDROIDX_WARNING = "flutterPluginsAndroidXWarning";
-if (!rootProject.ext.has(ANDROIDX_WARNING)) {
-    println '         *********************************************************'
-    println 'WARNING: This version of ' + PLUGIN + ' will break your Android build if it isn\'t compatible with AndroidX.'
-    println '         https://goo.gl/y9dCcU describes AndroidX and how to migrate to it using Android Studio.'
-    println '         https://goo.gl/n2E4mD describes how to open a Flutter app\'s Android files for refactoring in Android Studio.'
-    println '         *********************************************************'
-    rootProject.ext.set(ANDROIDX_WARNING, true);
+gradle.buildFinished { buildResult ->
+    if (buildResult.failure && !rootProject.ext.has(ANDROIDX_WARNING)) {
+        println '         *********************************************************'
+        println 'WARNING: This version of ' + PLUGIN + ' will break your Android build if it or its dependencies aren\'t compatible with AndroidX.'
+        println '         https://goo.gl/y9dCcU describes AndroidX and how to migrate to it using Android Studio.'
+        println '         https://goo.gl/n2E4mD describes how to open a Flutter app\'s Android files for refactoring in Android Studio.'
+        println '         This warning prints for all Android build failures. The real root cause of the error may be unrelated.'
+        println '         *********************************************************'
+        rootProject.ext.set(ANDROIDX_WARNING, true);
+    }
 }
 
 group 'io.flutter.plugins.androidalarmmanager'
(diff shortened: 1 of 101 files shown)


mklim:
The website hasn't been updated with the docs on opening a Flutter app for Android editing yet so this is still WIP, but the instructions are available [here](https://github.com/flutter/website/blob/master/src/docs/development/tools/android-studio.md#editing-android-code-in-android-studio-with-full-ide-support-android-ide) for now.

This is what it looks like on build failure:

failure

This is what it looks like on build success:

success

dnfield:
Would it be feasible to check whether their build.gradle is using incompatible libraries/sdks before printing this and exiting?

packages/android_alarm_manager/android/build.gradle
```diff
@@ -1,3 +1,14 @@
+def PLUGIN = "android_alarm_manager";
+def ANDROIDX_WARNING = "flutterPluginsAndroidXWarning";
+if (!rootProject.ext.has(ANDROIDX_WARNING)) {
```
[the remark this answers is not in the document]
mklim: Discussed offline. I'm still looking into if there's something better to do here, but unfortunately I'm not sure if this is feasible since the build can break based on incompatible libraries included in other gradle files in addition to the root project's compileSdkVersion.

mklim:
The IDE instructions are available at https://flutter.io/docs/development/tools/android-studio#android-ide.

I also found a hook for build failure. Now it doesn't print anything when the build succeeds, and looks like the previous screenshot still when the build fails. This could still log when the build fails for reasons unrelated to AndroidX though.

packages/android_alarm_manager/android/build.gradle
```diff
@@ -1,3 +1,16 @@
+def PLUGIN = "android_alarm_manager";
+def ANDROIDX_WARNING = "flutterPluginsAndroidXWarning";
+gradle.buildFinished { buildResult ->
+    if (buildResult.failure && !rootProject.ext.has(ANDROIDX_WARNING)) {
+        println '         *********************************************************'
+        println 'WARNING: This version of ' + PLUGIN + ' will break your Android build if it isn\'t compatible with AndroidX.'
```
dnfield: `will break your Android build if your project or its dependencies aren't compatible with AndroidX.` (not sure if the `or its dependencies` part is true - do you know?)

packages/android_alarm_manager/android/build.gradle
```diff
@@ -1,3 +1,16 @@
...
+        println '         https://goo.gl/y9dCcU describes AndroidX and how to migrate to it using Android Studio.'
+        println '         https://goo.gl/n2E4mD describes how to open a Flutter app\'s Android files for refactoring in Android Studio.'
```
dnfield: Maybe just one more line saying that this error message prints out for any build failure, and the failure may be unrelated to this problem but that users should still validate it.  

Ideally we detect that the failure was for this problem and help them even more, but this is at least better than what we have.

packages/android_alarm_manager/android/build.gradle
mklim: I'm not sure if it's universally true, but I've managed to crash the build at least once from this. Added.

Hixie:
I found a reason why this won't be enough: we don't control all the plugins. I depend on flutter_local_notifications, and when I tried to upgrade to the latest version today, I got the horrific and inexplicable Gradle error.

mklim:
@Hixie this PR won't fix build failures in that case, but I don't think that will stop this PR from at least printing out a warning at the top of the console. Since it's still a gradle error the various scripts from our plugins will still get the on gradle failure hook and log a message.

mklim:
> @Hixie this PR won't fix build failures in that case, but I don't think that will stop this PR from at least printing out a warning at the top of the console. Since it's still a gradle error the various scripts from our plugins will still get the on gradle failure hook and log a message.

Discussed offline, I misunderstood the case here. The remaining problem is when somebody has picked up an AndroidX plugin from somewhere other than `flutter/plugins`. Since this patch just applies warnings to the plugins in that repo, it won't cover any other plugin migration failures. I'll follow up with a PR to flutter's tooling to catch build errors globally and direct to a help page for this class of error. I still think this PR is worth merging since it catches cases for anyone using a `flutter/plugins` package who isn't on the latest version of flutter itself.

dnfield:
This LGTM - I understand it as an improvement for at least some things we control, even if it's not the full solution

Hixie:
LGTM assuming that we land the tool side that hides this message very soon. (Maybe first.)

We should probably document how to do this so that other plugin authors can do the same.

mklim:
> LGTM assuming that we land the tool side that hides this message very soon. (Maybe first.)
> 
> We should probably document how to do this so that other plugin authors can do the same.

Tool PR has landed so I'm going to merge this. I added [some notes for plugin maintainers to our website](https://flutter.io/docs/development/packages-and-plugins/androidx-compatibility#for-plugin-maintainers-migrating-a-flutter-plugin-to-androidx) but didn't mention this kind of Gradle warning, I'll update it in a follow up to mention it.

sjmcdowall:
BTW -- is the Flutter team going to write up (or if they did, where is it?) how to upgrade a normal Flutter application -- in Flutter Speak -- not Android speak -- when upgrading plugins and then hitting the dreaded AndroidX cascade of errors?  I've tried TWICE now upgrading some of the core packages that are now AndroidX and had to revert after spending 2-3 hours chasing down obscure Gradle error messages ..

This is a pretty normal (?) app that has about 20+ pubspec.yaml dependencies -- many 3rd party -- which aren't converted (and may never be?) .. 

I think there needs to be a concise "Conversion" document -- what things must be changed in an existing projects gradle files .. 

This is on the beta branch BTW.. I tried creating a new project and comparing the Gradle files but don't see any changes that I think are "significant" ..  it's still using SDK 27 stuff, etc. 

Anyway, so far this has been pretty frustrating .. almost as much as using a Swift Plugin within the app .. :) 

Cheers

mklim:
Hey @sjmcdowall. I'm really sorry that this has been so frustrating. Thanks for the feedback.

This PR and the related plugins one link to this doc on our website: https://flutter.io/docs/development/packages-and-plugins/androidx-compatibility. Does this fit what you're thinking of? The goal is that this should work as a conversion doc for Flutter apps specifically. We're trying to iterate and improve it, very open to any input on it.

sjmcdowall:
@mklim -- let me take a look and see what that links says... and more importantly -- let me try (after reading it) how the upgrade to the latest AndroidX plugins (that are converted) works.. 

If I still run into issues -- is this an appropriate place to report them or start a new issue? 

Cheers!

mklim:
@sjmcdowall thanks! If you're still running into issues it would be best if you could open a new issue in the flutter/flutter repo explaining the problem. Feel free to cc me.

PR #1139, opened 2019-01-30
WebView JavaScript channels - iOS implementation.

amirh:
Platform implementation of the method channel API for adding and removing JavaScript channels.

#1116 adds the Dart side support, the current PR will land first.

flutter/flutter#24837

### DIFF
--- a/packages/webview_flutter/ios/Classes/FlutterWebView.m
+++ b/packages/webview_flutter/ios/Classes/FlutterWebView.m
@@ -54,7 +54,7 @@
     _javaScriptChannelNames = [[NSMutableSet alloc] init];
 
     WKUserContentController* userContentController = [[WKUserContentController alloc] init];
-    if (![args[@"javascriptChannelNames"] isKindOfClass:[NSNull class]]) {
+    if ([args[@"javascriptChannelNames"] isKindOfClass:[NSArray class]]) {
       NSArray* javaScriptChannelNames = args[@"javascriptChannelNames"];
       [_javaScriptChannelNames addObjectsFromArray:javaScriptChannelNames];
       [self registerJavaScriptChannels:_javaScriptChannelNames controller:userContentController];
@@ -105,7 +105,7 @@
   } else if ([[call method] isEqualToString:@"addJavascriptChannels"]) {
     [self onAddJavaScriptChannels:call result:result];
   } else if ([[call method] isEqualToString:@"removeJavascriptChannels"]) {
-    [self onRemoveJavaScriptChannes:call result:result];
+    [self onRemoveJavaScriptChannels:call result:result];
   } else {
     result(FlutterMethodNotImplemented);
   }
@@ -188,7 +188,7 @@
   result(nil);
 }
 
-- (void)onRemoveJavaScriptChannes:(FlutterMethodCall*)call result:(FlutterResult)result {
+- (void)onRemoveJavaScriptChannels:(FlutterMethodCall*)call result:(FlutterResult)result {
   // WkWebView does not support removing a single user script, so instead we remove all
   // user scripts, all message handlers. And re-register channels that shouldn't be removed.
   [_webView.configuration.userContentController removeAllUserScripts];
--- a/packages/webview_flutter/ios/Classes/JavaScriptChannelHandler.m
+++ b/packages/webview_flutter/ios/Classes/JavaScriptChannelHandler.m
@@ -12,6 +12,8 @@
 - (instancetype)initWithMethodChannel:(FlutterMethodChannel*)methodChannel
                 javaScriptChannelName:(NSString*)javaScriptChannelName {
   self = [super init];
+  NSAssert(methodChannel != nil, @"methodChannel must not be null.");
+  NSAssert(javaScriptChannelName != nil, @"javaScriptChannelName must not be null.");
   if (self) {
     _methodChannel = methodChannel;
     _javaScriptChannelName = javaScriptChannelName;
@@ -21,6 +23,9 @@
 
 - (void)userContentController:(WKUserContentController*)userContentController
       didReceiveScriptMessage:(WKScriptMessage*)message {
+  NSAssert(_methodChannel != nil, @"Can't send a message to an unitialized JavaScript channel.");
+  NSAssert(_javaScriptChannelName != nil,
+           @"Can't send a message to an unitialized JavaScript channel.");
   NSDictionary* arguments = @{
     @"channel" : _javaScriptChannelName,
     @"message" : [NSString stringWithFormat:@"%@", message.body]


packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -154,6 +179,33 @@ - (void)onEvaluateJavaScript:(FlutterMethodCall*)call result:(FlutterResult)resu
              }];
 }
 
+- (void)onAddJavaScriptChannels:(FlutterMethodCall*)call result:(FlutterResult)result {
+  NSArray* channelNames = [call arguments];
+  NSSet* channelNamesSet = [[NSSet alloc] initWithArray:channelNames];
+  [_javaScriptChannelNames addObjectsFromArray:channelNames];
+  [self registerJavaScriptChannels:channelNamesSet
+                        controller:_webView.configuration.userContentController];
```
cyanglaz: It seems to have the same logic as code around line 57 to 61. Might be refactored into a method for cleanness.

packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -154,6 +179,33 @@ - (void)onEvaluateJavaScript:(FlutterMethodCall*)call result:(FlutterResult)resu
              }];
 }
 
+- (void)onAddJavaScriptChannels:(FlutterMethodCall*)call result:(FlutterResult)result {
+  NSArray* channelNames = [call arguments];
+  NSSet* channelNamesSet = [[NSSet alloc] initWithArray:channelNames];
+  [_javaScriptChannelNames addObjectsFromArray:channelNames];
+  [self registerJavaScriptChannels:channelNamesSet
```
cyanglaz: What happens if there is a duplicated value in channelNameSet with the global _javaScriptChannelNames which has been registered before? Is it safe to run the register again with the same channel?

packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -33,6 +38,8 @@ @implementation FLTWebViewController {
   int64_t _viewId;
   FlutterMethodChannel* _channel;
   NSString* _currentUrl;
+  // The set of registered JavaScript channel names.
+  NSMutableSet* _javaScriptChannelNames;
```
cyanglaz: Can use a property and be lazily initialized.

packages/webview_flutter/ios/Classes/FlutterWebView.h
```diff
@@ -1,3 +1,7 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
```
mklim: nit: 2019?

packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -41,15 +48,29 @@ - (instancetype)initWithFrame:(CGRect)frame
...
     _viewId = viewId;
-    _webView = [[WKWebView alloc] initWithFrame:frame];
+
     NSString* channelName = [NSString stringWithFormat:@"plugins.flutter.io/webview_%lld", viewId];
     _channel = [FlutterMethodChannel methodChannelWithName:channelName binaryMessenger:messenger];
+    _javaScriptChannelNames = [[NSMutableSet alloc] init];
+
+    WKUserContentController* userContentController = [[WKUserContentController alloc] init];
+    if (![args[@"javascriptChannelNames"] isKindOfClass:[NSNull class]]) {
+      NSArray* javaScriptChannelNames = args[@"javascriptChannelNames"];
```
mklim: This is really close to being the same code as `onAddJavascriptChannels`. Would it make sense to change the method signature of that class and make a call to it here instead?

packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -154,6 +179,33 @@ - (void)onEvaluateJavaScript:(FlutterMethodCall*)call result:(FlutterResult)resu
...
+- (void)onAddJavaScriptChannels:(FlutterMethodCall*)call result:(FlutterResult)result {
+  NSArray* channelNames = [call arguments];
+  NSSet* channelNamesSet = [[NSSet alloc] initWithArray:channelNames];
+  [_javaScriptChannelNames addObjectsFromArray:channelNames];
+  [self registerJavaScriptChannels:channelNamesSet
+                        controller:_webView.configuration.userContentController];
+  result(nil);
+}
+
+- (void)onRemoveJavaScriptChannes:(FlutterMethodCall*)call result:(FlutterResult)result {
```
mklim: nit: "Channels"

packages/webview_flutter/ios/Classes/JavaScriptChannelHandler.m
```diff
@@ -0,0 +1,31 @@
...
+    _methodChannel = methodChannel;
+    _javaScriptChannelName = javaScriptChannelName;
+  }
+  return self;
+}
+
+- (void)userContentController:(WKUserContentController*)userContentController
+      didReceiveScriptMessage:(WKScriptMessage*)message {
+  NSDictionary* arguments = @{
+    @"channel" : _javaScriptChannelName,
```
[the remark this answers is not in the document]
amirh: done

packages/webview_flutter/ios/Classes/FlutterWebView.m
amirh: What's the benefit over doing it this way?

amirh:
@mklim for unit tests - ok if I wait to see in what form our e2e plugin test harness manifests?

mklim:
> @mklim for unit tests - ok if I wait to see in what form our e2e plugin test harness manifests?

Totally. Just a suggestion.

packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -41,15 +48,29 @@ - (instancetype)initWithFrame:(CGRect)frame
...
   if ([super init]) {
     _viewId = viewId;
-    _webView = [[WKWebView alloc] initWithFrame:frame];
+
     NSString* channelName = [NSString stringWithFormat:@"plugins.flutter.io/webview_%lld", viewId];
     _channel = [FlutterMethodChannel methodChannelWithName:channelName binaryMessenger:messenger];
+    _javaScriptChannelNames = [[NSMutableSet alloc] init];
+
+    WKUserContentController* userContentController = [[WKUserContentController alloc] init];
+    if (![args[@"javascriptChannelNames"] isKindOfClass:[NSNull class]]) {
```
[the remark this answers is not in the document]
cyanglaz: sounds good. and YES [foo isKindOfClass:[NSArray class]] covers all cases.

packages/webview_flutter/ios/Classes/FlutterWebView.m
cyanglaz: lazy initializing the set can save memory (insignificant tho); In the case of this class, benefit is minimum since the set is used right after class initialization; I'm ok with not lazy initializing here.
Just a thought, if we do move the code `[_javaScriptChannelNames addObjectsFromArray:javaScriptChannelNames];` out of the `initiWithFrame` method, then we can think about lazy initialize the _javaScriptChannelNames to save some memory.

amirh: We can't move it out of initWithFrame, as a set of JavaScript channels is passed with the creation params... we immediately need this field initialized...

amirh:
Looks like the build-ipas task has successfully completed on Cirrus but Github didn't pick that up for some reason.. going to land on "yellow" because this is really green.

packages/webview_flutter/ios/Classes/JavaScriptChannelHandler.m
```diff
@@ -0,0 +1,36 @@
...
+}
+
+- (void)userContentController:(WKUserContentController*)userContentController
+      didReceiveScriptMessage:(WKScriptMessage*)message {
+  NSAssert(_methodChannel != nil, @"Can't send a message to an unitialized JavaScript channel.");
+  NSAssert(_javaScriptChannelName != nil,
+           @"Can't send a message to an unitialized JavaScript channel.");
+  NSDictionary* arguments = @{
+    @"channel" : _javaScriptChannelName,
+    @"message" : [NSString stringWithFormat:@"%@", message.body]
```
KennedyChiang: if native send json string back flutter, it will got a string that hard to serialized like below.

```json
{
    action = do_somethineg;
    data =     {
        key_1 = some_info;
        key_2 = "some_info";
    };
    timestamp = 1570005270866;
    token = "jwt_token";
}
```

is it possible to add code to serialize NSDictionary and  before send message.body:

```objective-c
- (NSString *)messageFromBody:(NSString *)body {
    NSString *message;
    if ([body isKindOfClass:[NSDictionary class]] ||
        [body isKindOfClass:[NSArray class]]) {
        NSData *jsonData = [NSJSONSerialization dataWithJSONObject:body
                                                           options:NSJSONWritingPrettyPrinted
                                                             error:nil];
        message = [[NSString alloc] initWithData:jsonData encoding:NSUTF8StringEncoding];
    }
    if (message.length <= 0) {
        message = [NSString stringWithFormat:@"%@", body];
    }

    return message;
}

...

NSDictionary* arguments = @{
    @"channel" : _javaScriptChannelName,
    @"message" : [self messageFromBody: message.body]
  };
```

Thanks!

amirh: Do you mind filing this as an issue?

KennedyChiang: Ya, i'm created [HERE](https://github.com/flutter/flutter/issues/42986)

PR #1140, opened 2019-01-31
Change iOS video format back to bgra8888

bparrishMines:
Fixes https://github.com/flutter/flutter/issues/26567


packages/camera/ios/Classes/CameraPlugin.m
```diff
@@ -176,11 +175,21 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
   _captureSession = [[AVCaptureSession alloc] init];
   AVCaptureSessionPreset preset;
   if ([resolutionPreset isEqualToString:@"high"]) {
-    preset = AVCaptureSessionPreset1280x720;
-    _previewSize = CGSizeMake(1280, 720);
+    if ([_captureSession canSetSessionPreset:AVCaptureSessionPreset1920x1080]) {
+      preset = AVCaptureSessionPreset1920x1080;
+      _previewSize = CGSizeMake(1920, 1080);
+    } else {
+      preset = AVCaptureSessionPreset1280x720;
```
mklim: Do we need to potentially bail down to lower resolutions in these fallback cases too? Is it possible that we won't be able to set a preset at this slightly lower size either?

packages/camera/ios/Classes/CameraPlugin.m
```diff
@@ -280,12 +289,31 @@ - (void)captureOutput:(AVCaptureOutput *)output
 
       NSMutableArray *planes = [NSMutableArray array];
 
-      size_t planeCount = CVPixelBufferGetPlaneCount(pixelBuffer);
+      Boolean isPlanar = CVPixelBufferIsPlanar(pixelBuffer);
```
mklim: nit, optional: `const`

packages/camera/ios/Classes/CameraPlugin.m
```diff
@@ -246,6 +235,59 @@ - (void)captureToFile:(NSString *)path result:(FlutterResult)result {
                                                            cameraPosition:_captureDevice.position]];
 }
 
+- (void)setCaptureSessionPreset:(NSString *)resolutionPreset {
```
bparrishMines: This will most likely just be a temporary fix until we expose available resolution for each device.

packages/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.4.0
+
+* **Breaking Change** Change iOS image stream format to `ImageFormatGroup.bgra8888`.
```
mklim: How and why is this breaking? If I'm a dev that would be broken by upgrading to this version, how can I know? If I already installed this version and I'm currently broken, is there anything you can tell me here on how to fix the breakage?

packages/camera/ios/Classes/CameraPlugin.m
```diff
@@ -244,6 +233,59 @@ - (void)captureToFile:(NSString *)path result:(FlutterResult)result {
...
+  } else if ([resolutionPreset isEqualToString:@"medium"]) {
+    presetIndex = 2;
+  } else {
+    NSAssert([resolutionPreset isEqualToString:@"low"], @"Unknown resolution preset %@",
+             resolutionPreset);
+    presetIndex = 3;
+  }
+
+  switch (presetIndex) {
+    case 0:
```
mklim: Any reason not to just case off of `resolutionPreset` directly here?

packages/camera/ios/Classes/CameraPlugin.m
```diff
@@ -244,6 +233,59 @@ - (void)captureToFile:(NSString *)path result:(FlutterResult)result {
...
+        break;
+      }
+    default: {
+      NSException *exception = [NSException
+          exceptionWithName:@"NoAvailableCaptureSessionException"
+                     reason:@"No capture session available for current capture session."
+                   userInfo:nil];
+      @throw exception;
+    }
+  }
```
mklim: What happens if we go into any of these cases and `_captureSession canSetSessionPreset:` returns false?

packages/camera/ios/Classes/CameraPlugin.m
bparrishMines: Switch statements don't work with strings in Objective-c. They only work with primitive types.

packages/camera/ios/Classes/CameraPlugin.m
bparrishMines: So [here](https://developer.apple.com/documentation/avfoundation/avcapturesessionpreset?language=objc) is the list of all capture session presets.

For history of the plugin and why I decided on this method to choose resolution preset: 
Originally we would used [AVCaptureSessionPresetHigh](https://developer.apple.com/documentation/avfoundation/avcapturesessionpresethigh?language=objc)/[AVCaptureSessionPresetMedium](https://developer.apple.com/documentation/avfoundation/avcapturesessionpresetmedium?language=objc)/[AVCaptureSessionPresetLow](https://developer.apple.com/documentation/avfoundation/avcapturesessionpresetlow?language=objc) for choosing the resolution. Which you can see starting at this [line](https://github.com/flutter/plugins/blob/eaa1388b11ac3ee6143ad782e573b9b9e314bc4b/packages/camera/ios/Classes/CameraPlugin.m#L103). I assume the camera plugin uses low/medium/high presets to copy iOS's way of handling resolution. 

However, there is no consistent way to get the correct `previewSize` before starting the camera. You can see more about this on [stackoverflow](https://stackoverflow.com/questions/7790402/knowing-resolution-of-avcapturesessions-session-presets). We would use this [method](https://github.com/flutter/plugins/blob/eaa1388b11ac3ee6143ad782e573b9b9e314bc4b/packages/camera/ios/Classes/CameraPlugin.m#L122) to get the resolution, but it would sometimes return the wrong `previewSize`. 

Since I couldn't find any way to get the correct resolution until after returning `previewSize`, I just chose to create a ladder with defined resolution presets. This ladder is just the subset of resolutions available on iOS. And **should** work with each device. It works for iphone 6 and iphone X. Other apps would just use a [chart](https://stackoverflow.com/questions/12240625/avcapturesession-settings-of-low-medium-and-high-what-res-do-they-each-represe), but I felt that would be overly complicated for fixing this bug.

packages/camera/CHANGELOG.md
bparrishMines: This is a breaking change because the original format of the image stream buffer was `kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange` and now it is `kCVPixelFormatType_32BGRA`. This will most likely be a breaking change for a very small number of people, if any. I added this detail to the changelog.

If a developer was expecting `kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange` the best solution for them would be to use a conversion library on the platform side. Do you think I should link the iOS `v_image` library in the changelog?

packages/camera/ios/Classes/CameraPlugin.m
bparrishMines: Another option would be to make `previewSize` an asynchronous getter which we update on the platform side.

PR #1144, opened 2019-02-01
fixing iOS closeWebview method

marquesinijatinha:
UrlLauncherPlugin.m was wrongly settting currentSession to nil when the page did the initial load, and that prevented the webview to closed via code.


marquesinijatinha:
I signed it!

PR #1149, opened 2019-02-02
Allow clearing cookies for FlutterWebView

iskakaushik:
As a first step to address https://github.com/flutter/flutter/issues/27321


packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -308,6 +308,21 @@ class WebViewController {
...
+class CookieManager {
+  const CookieManager._();
+
+  static const MethodChannel _channel =
+      MethodChannel('plugins.flutter.io/cookie_manager');
+
+  /// Removes all cookies, and returns true if cookies were
+  /// present before clearing, else false.
+  Future<bool> clearCookies() async {
+    print(defaultTargetPlatform);
```
amirh: delete

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -308,6 +308,21 @@ class WebViewController {
...
 }
 
+/// Manages cookies pertaining to all [WebView]s.
+class CookieManager {
+  const CookieManager._();
+
+  static const MethodChannel _channel =
+      MethodChannel('plugins.flutter.io/cookie_manager');
+
+  /// Removes all cookies, and returns true if cookies were
```
amirh: The first paragraph will be used in the Dartdoc's table of contents. better to keep it short. I'd just have the summary paragraph say "Clears all cookies."
And then in the next paragraph expand on which webview instances are affected, and in a following paragraph describe the return value.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -319,3 +334,6 @@ void _validateUrlString(String url) {
     throw ArgumentError(e);
   }
 }
+
+/// Singleton [CookieManager]
```
amirh: Describe what it is.
also nit: end sentences with period.

packages/webview_flutter/ios/Classes/WebViewFlutterPlugin.m
```diff
@@ -7,6 +8,7 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar>*)registrar {
   FLTWebViewFactory* webviewFactory =
       [[FLTWebViewFactory alloc] initWithMessenger:registrar.messenger];
   [registrar registerViewFactory:webviewFactory withId:@"plugins.flutter.io/webview"];
+  FLTCookieManager* cookieManager = [[FLTCookieManager alloc] init:registrar.messenger];
```
amirh: What's keeping a reference to this instance?

packages/webview_flutter/example/lib/main.dart
```diff
@@ -111,6 +120,17 @@ class SampleMenu extends StatelessWidget {
       ),
     );
   }
+
+  void _onClearCookies(BuildContext context) async {
+    final bool hadCookies = await cookieManager.clearCookies();
+    String message = "There were cookies. Now, they are gone!";
```
amirh: How do we know that it worked?
Maybe we could make a JS call to get the list of cookies present it?

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
```diff
@@ -0,0 +1,52 @@
...
+import android.os.Build.VERSION_CODES;
+import android.webkit.CookieManager;
+import android.webkit.ValueCallback;
+import io.flutter.plugin.common.BinaryMessenger;
+import io.flutter.plugin.common.MethodCall;
+import io.flutter.plugin.common.MethodChannel;
+import io.flutter.plugin.common.MethodChannel.MethodCallHandler;
+import io.flutter.plugin.common.MethodChannel.Result;
+
+public class FlutterCookieManager implements MethodCallHandler {
```
amirh: nit: can we make this package private?

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private FlutterCookieManager() {
+    // Do not instantiate.
```
amirh: explain why

amirh:
Please also update the change log, and bump the plugin version in pubspec.yaml.
Versioning guidelines are here: https://www.dartlang.org/tools/pub/versioning#semantic-versions

packages/webview_flutter/ios/Classes/WebViewFlutterPlugin.m
iskakaushik: the instance is not being used, but the init method needs to be called. Removing the return reference.

packages/webview_flutter/example/lib/main.dart
iskakaushik: Good point, for now i verified that flutter.io was setting and deleting cookies. Will add a method to list all cookies.

packages/webview_flutter/ios/Classes/WebViewFlutterPlugin.m
amirh: If we don't keep a reference what prevents FLTCookieManager from being disposed by the time we got a method channel invocation?
(we are running this with ARC)

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -469,6 +469,22 @@ class WebViewController {
...
+
+  static const MethodChannel _channel =
+      MethodChannel('plugins.flutter.io/cookie_manager');
+
+  /// Clears all cookies.
+  ///
+  /// This is supported for >= IOS 9 and Android api level >= 16.
+  /// returns true if cookies were present before clearing, else false.
+  Future<bool> clearCookies() async {
+    return await _channel.invokeMethod<bool>('clearCookies');
```
jonahwilliams:
```suggestion
    return _channel.invokeMethod<bool>('clearCookies');
```

jonahwilliams: Using the type parameter will make this incompatible with stable, see the other plugin repos

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -480,3 +496,8 @@ void _validateUrlString(String url) {
     throw ArgumentError(e);
   }
 }
+
+/// Singleton [CookieManager].
+///
+/// Manages cookies pertaining to all [WebView]s.
+final CookieManager cookieManager = const CookieManager._();
```
jonahwilliams:
```suggestion
const CookieManager cookieManager = CookieManager._();
```

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -471,10 +471,19 @@ class WebViewController {
...
+      final MethodChannel methodChannel =
+          const MethodChannel('plugins.flutter.io/cookie_manager');
+      _instance = CookieManager.private(methodChannel);
+    }
+    return _instance;
+  }
 
-  static const MethodChannel _channel =
-      MethodChannel('plugins.flutter.io/cookie_manager');
+  CookieManager.private(this._channel);
```
jonahwilliams: Don't name a constructor `.private`, either its private `_` or its not

packages/webview_flutter/lib/webview_flutter.dart
amirh: good catch, these are risky days for missing these on code reviews. 😨 
https://github.com/flutter/flutter/issues/27110 would have been helpful here.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -469,6 +469,31 @@ class WebViewController {
...
+  factory CookieManager() {
+    if (_instance == null) {
+      final MethodChannel methodChannel =
+          const MethodChannel('plugins.flutter.io/cookie_manager');
+      _instance = CookieManager.private(methodChannel);
+    }
+    return _instance;
+  }
+
+  CookieManager.private(this._channel);
```
amirh: the name private has no special meaning.
this should be named `_` to make sure it is really private.

packages/webview_flutter/lib/webview_flutter.dart
iskakaushik: Ouch, fixed this. Thanks for the flag!

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -469,6 +469,32 @@ class WebViewController {
   }
 }
 
+/// Manages cookies pertaining to all [WebView]s.
+class CookieManager {
+  factory CookieManager() {
```
amirh: This needs a Dartdoc, it should mention that it returns or created the singleton instance.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -469,6 +469,32 @@ class WebViewController {
...
+  }
+
+  CookieManager.private(this._channel);
+
+  static CookieManager _instance;
+  final MethodChannel _channel;
+
+  /// Clears all cookies.
+  ///
+  /// This is supported for >= IOS 9 and Android api level >= 16.
```
amirh: no need to mention Android API level (16 is the minimum level supported by Flutter anyway)

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -469,6 +469,32 @@ class WebViewController {
   }
 }
 
+/// Manages cookies pertaining to all [WebView]s.
+class CookieManager {
+  factory CookieManager() {
+    if (_instance == null) {
+      final MethodChannel methodChannel =
```
jonahwilliams: Since we can set mockMethodCallHandlers for testing, what is the value of providing the channel this way?

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -469,6 +469,32 @@ class WebViewController {
...
+
+  CookieManager.private(this._channel);
+
+  static CookieManager _instance;
+  final MethodChannel _channel;
+
+  /// Clears all cookies.
+  ///
+  /// This is supported for >= IOS 9 and Android api level >= 16.
+  /// returns true if cookies were present before clearing, else false.
```
amirh: nit: separate the paragraph, start sentence with a capital letter.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -469,6 +469,32 @@ class WebViewController {
...
+  Future<bool> clearCookies() => _channel
+      // ignore: strong_mode_implicit_dynamic_method
```
amirh: Add the same comment we have everywhere(minus the typo 😄 ) so it's easy to find all of these at once (and we need a TODO here anyway)

packages/webview_flutter/lib/webview_flutter.dart
iskakaushik: My original thought was to make this a const field on the class. I wanted to be consistent with Battery plugin, i should maybe look through the test code there.

packages/webview_flutter/ios/Classes/FLTCookieManager.m
```diff
@@ -0,0 +1,50 @@
...
+  if ([[call method] isEqualToString:@"clearCookies"]) {
+    [self clearCookies:result];
+  } else {
+    result(FlutterMethodNotImplemented);
+  }
+}
+
+- (void)clearCookies:(FlutterResult)result {
+  NSOperatingSystemVersion ios9 = (NSOperatingSystemVersion){9, 0};
+  if ([[NSProcessInfo processInfo] isOperatingSystemAtLeastVersion:ios9]) {
```
amirh: ```if (@available(iOS 9.0, *)) {```

packages/webview_flutter/lib/webview_flutter.dart
jonahwilliams: Consistency is good to a point, but correctness is better

packages/webview_flutter/ios/Classes/FLTCookieManager.m
```diff
@@ -0,0 +1,50 @@
...
+    [self clearCookiesIos9AndLater:result];
+  } else {
+    // support for IOS-8 tracked in https://github.com/flutter/flutter/issues/27624.
+    NSLog(@"Clearing cookies is not supported for Flutter WebViews prior to iOS9.");
```
amirh: nit: s/iOS9/iOS 9/

packages/webview_flutter/lib/webview_flutter.dart
iskakaushik: curious, does this approach expose a way of injecting the wrong method channel?

packages/webview_flutter/example/lib/main.dart
```diff
@@ -125,6 +135,42 @@ class SampleMenu extends StatelessWidget {
     controller.evaluateJavascript(
         'Toaster.postMessage("User Agent: " + navigator.userAgent);');
   }
+
+  void _onClearCookies(
+      WebViewController controller, BuildContext context) async {
+    final String cookies =
+        await controller.evaluateJavascript('document.cookie');
```
amirh: What do you think about adding a menu entry that just shows a toast with the cookies?
I'd feel more confident to verify that it's working if the button that's showing me the cookies is independent of the button that's deleting the cookies.

packages/webview_flutter/lib/webview_flutter.dart
jonahwilliams: It's not necessary at all

iskakaushik: i agree that passing in a method channel is not needed for testing, but passing it though this way -- creating it on initialization vs having it static, should be equivalent?

packages/webview_flutter/example/lib/main.dart
iskakaushik: Makes sense.

packages/webview_flutter/lib/webview_flutter.dart
jonahwilliams: Generally making const fields is frowned upon. since the const is already canonicalized we are only adding overhead to the class. Furthermore, removing the initialization from the factory lets us cut down the code a bit with `??=`:

```
factory Foo() {
  return _instance ??= Foo._();
}
Foo._();
static Foo _instance;
```

packages/webview_flutter/lib/webview_flutter.dart
amirh: nit: not sure if '--' is common in Flutter dartdocs (of if it is grammatically correct), I'd just say:
```
Returns the singleton cookie manager instance.

The instance is created the first time this factory is called.
```

packages/webview_flutter/lib/webview_flutter.dart
amirh: keep it TODO(amirh), most likely someone will end up searching for all these `TODO(amirh):` occurrences at once.

packages/webview_flutter/ios/Classes/FLTCookieManager.m
amirh: ping

PR #1152, opened 2019-02-02
[android_alarm_manager] Specfiy that targetSdkVersion and comileSdkVersion need to be 28

gabrc52:
Fixes #27441


gabrc52:
Shouldn't be necessary now since this is also necessary for AndroidX migration.

gabrc52:
After it's been merged, I've noticed all steps appear as 1. 

Would adding padding fix the issue?

PR #1156, opened 2019-02-04
Add capability of using single plane buffers

bparrishMines:
This PR deals with handling the changes coming to the camera plugin https://github.com/flutter/plugins/pull/1140. 

The camera plugin uses a multi planar buffer for streaming images, but is switching back to a non-planar image buffer. This PR adds the capability to handle that change, but causes no breaking changes.


packages/firebase_ml_vision/ios/Classes/FirebaseMlVisionPlugin.m
```diff
@@ -57,70 +57,121 @@ - (FIRVisionImage *)dataToVisionImage:(NSDictionary *)imageData {
   NSString *imageType = imageData[@"type"];
 
   if ([@"file" isEqualToString:imageType]) {
-    UIImage *image = [UIImage imageWithContentsOfFile:imageData[@"path"]];
```
bparrishMines: Moved these two lines to a separate function -> 71

packages/firebase_ml_vision/CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 ## 0.4.0
 
 * **Breaking Change** Removal of base detector class `FirebaseVisionDetector`.
-* **Breaking Change** Removal of `TextRecognizer.detectInImage()`. Pleas use `TextRecognizer.processImage()`.
+* **Breaking Change** Removal of `TextRecognizer.detectInImage()`. Pleas use
+  `TextRecognizer.processImage()`.
 * **Breaking Change** Changed `FaceDetector.detectInImage()` to `FaceDetector.processImage()`.
+* iOS now handles non-planar buffers from `FirebaseVisionImage.fromBytes()`.
```
bparrishMines: 0.4.0 has not been publish as of 1/4, so I'm adding this PR to that version

PR #1160, opened 2019-02-05
Ensure that flutterError for NSError extension has prefix for Categor…

kangwang1988:
Ensure that flutterError for NSError extension has prefix for Category and method so that they will work as expected when using together.
See more at: https://github.com/flutter/flutter/issues/27361


cyanglaz:
PR looks good, waiting for @cbracken to confirm on flutter/flutter#27361 if adding prefix is the way we should do it.

cyanglaz:
Let's avoid category all together and use static C functions.

PR #1162, opened 2019-02-05
Expose nslocale and expose currencySymbol instead of currencyCode to match android

cyanglaz:
To make to code simple, NSLocale has not been fully exposed.
Expose the NSLocale inside the priceLocale layer to match the SKProduct composition; this makes future changes regarding priceLocale to be non-breaking changes.

### DIFF
--- a/packages/in_app_purchase/ios/Classes/FIAObjectTranslator.m
+++ b/packages/in_app_purchase/ios/Classes/FIAObjectTranslator.m
@@ -56,12 +56,11 @@
     @"paymentMode" : @(self.paymentMode)
   }];
 
-  if (@available(iOS 10.0, *)) {
+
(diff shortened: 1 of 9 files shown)


packages/in_app_purchase/ios/Classes/FIAObjectTranslator.m
```diff
@@ -83,3 +81,18 @@ - (NSDictionary *)toMap {
 }
 
 @end
+
+@implementation NSLocale (Coder)
+
+- (nullable NSDictionary *)toMap {
+  NSMutableDictionary *map = [[NSMutableDictionary alloc] init];
+  if (@available(iOS 10.0, *)) {
```
mklim: So do we not have access to this information on ios < 10?

PR #1163, opened 2019-02-05
Change from Rectangle<num>/Point<num> to Rect/Offset for ui convenience

bparrishMines:
Also fixes a bug where iOS was not getting corner point results.


packages/firebase_ml_vision/ios/Classes/BarcodeDetector.m
```diff
@@ -32,16 +32,16 @@ + (void)handleDetection:(FIRVisionImage *)image
 NSDictionary *visionBarcodeToDictionary(FIRVisionBarcode *barcode) {
   __block NSMutableArray<NSArray *> *points = [NSMutableArray array];
 
-  for (NSValue *point in points) {
-    [points addObject:@[ @(((__bridge CGPoint *)point)->x), @(((__bridge CGPoint *)point)->y) ]];
+  for (NSValue *point in barcode.cornerPoints) {
+    [points addObject:@[ @(point.CGPointValue.x), @(point.CGPointValue.y) ]];
```
bparrishMines: Bug fix

packages/firebase_ml_vision/ios/Classes/TextRecognizer.m
```diff
@@ -79,8 +79,8 @@ + (void)addData:(NSMutableDictionary *)addTo
             text:(NSString *)text {
   __block NSMutableArray<NSArray *> *points = [NSMutableArray array];
 
-  for (NSValue *point in points) {
-    [points addObject:@[ @(((__bridge CGPoint *)point)->x), @(((__bridge CGPoint *)point)->y) ]];
+  for (NSValue *point in cornerPoints) {
+    [points addObject:@[ @(point.CGPointValue.x), @(point.CGPointValue.y) ]];
```
bparrishMines: bug fix

packages/firebase_ml_vision/CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
 ## 0.4.0
 
+* **Breaking Change** Change `Rectangle<int>` to `Rect` in Text/Face/Barcode results.
```
bparrishMines: Version `0.4.0` has yet to be [published](https://pub.dartlang.org/packages/firebase_ml_vision#-versions-tab-), so I'm adding the change to this version.

packages/firebase_ml_vision/test/firebase_ml_vision_test.dart
```diff
@@ -141,12 +140,12 @@ void main() {
 
         final Barcode barcode = barcodes[0];
         expect(barcode.valueType, BarcodeValueType.unknown);
-        expect(barcode.boundingBox, const Rectangle<int>(1, 2, 3, 4));
+        expect(barcode.boundingBox, Rect.fromLTWH(1.0, 2.0, 3.0, 4.0));
```
mjohnsullivan: With implicit int -> conversion in Dart 2.1 could change to ints, unless backwards compatibility a requirement

bparrishMines: I think keeping backwards compatibility would be the best move.

The [current](https://github.com/flutter/plugins/blob/master/packages/firebase_ml_vision/pubspec.yaml#L24) version only requires `2.0.0` and upgrading to `2.1.0` might not be worth causing another breaking change. The only gain from using `2.1` would make the tests require slightly less code.

PR #1169, opened 2019-02-06
Iap getproductlist basic draft

cyanglaz:
This is a very basic first draft for combining android and iOS get product list.
Would like this to be solid before working into more details.


packages/in_app_purchase/lib/src/in_app_purchase_connection/in_app_purchase_connection.dart
```diff
@@ -6,12 +6,19 @@ import 'dart:async';
...
 /// Basic generic API for making in app purchases across multiple platforms.
 abstract class InAppPurchaseConnection {
   /// Returns true if the payment platform is ready and available.
   Future<bool> isAvailable();
 
+  /// Query the product list.
+  ///
+  /// Returns a List of [Product] that each [Product] uniquely matches one valid identifier in [identifiers].
+  /// Any item in the `identifiers` that is not specified in App Store Connect or Google Play Console will be ignored.
+  Future<List<Product>> queryProductDetails(List<String> identifiers);
```
cyanglaz: An alternative could be returning a custom 'Response' class something like `ProductResponse` which contains the list of the product and some errors if any. Any thoughts on this?

packages/in_app_purchase/lib/src/billing_client_wrappers/sku_details_wrapper.g.dart
```diff
@@ -20,15 +20,46 @@ SkuDetailsWrapper _$SkuDetailsWrapperFromJson(Map json) {
       sku: json['sku'] as String,
       subscriptionPeriod: json['subscriptionPeriod'] as String,
       title: json['title'] as String,
-      type: const SkuTypeConverter().fromJson(json['type'] as String),
+      type: _$enumDecode(_$SkuTypeEnumMap, json['type']),
```
mklim: For some reason this generated with extra enum boilerplate and it shouldn't have.

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+
+/// The class represents the product object, it combined the common fields in the [SkuDetailsWrapper] and [SKProductWrapper].
```
mklim: At the `in_app_purchase_connection/` layer I think we probably want to avoid referring to the underlying platforms directly at all unless we absolutely need to. The idea is that this should be a minimal but clean abstraction as much as possible. Instead of trying to explain all the things this is wrapping this should explain what it actually is and how it functions. If people want more then they can drop down to the other packages and learn what they do. This applies for all of the comments in this file.

This isn't perfect but as an example, from [this early draft from of a proof of concept](https://github.com/mklim/plugins/blob/a6ccd9ca5d18adfb19644a7845e023bf14eadce8/packages/in_app_purchase/lib/in_app_purchase.dart#L47-L50):

> /// Store listing information for a product.
> ///
> /// This information needs to be manually created in each of the various
> /// purchasing platform's app management consoles.

Mostly just describing what the class does, but still referring to the underlying platform where it's absolutely required to get this working (setting up the information in the first place).

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
```diff
@@ -0,0 +1,32 @@
...
+class Product {
+  Product({
+    @required this.productIdentifier,
+    @required this.title,
+    @required this.description,
+    @required this.price,
+  });
+
+  /// The ProductIdentifier in App Store Connect or Sku in Google Play console.
+  final String productIdentifier;
```
mklim: nit: I think `id` would be just as clear here and it's more concise.

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
```diff
@@ -0,0 +1,32 @@
...
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+
+/// The class represents the product object, it combined the common fields in the [SkuDetailsWrapper] and [SKProductWrapper].
+///
+/// You can use this object for simple IAP operations. For more detailed and platform specific IAP operations,
+/// see [SkuDetailsWrapper] and [SKProductWrapper] instead.
+/// The programmer should not need to instantiate this object; rather use after returned by [InAppPurchaseConnection.getProductList].
+class Product {
```
mklim: It's tricky but I think naming this just `Product` has the potential to be confusing with purchased products. This is the data structure representing store listings of available products to buy/product details/product information, not purchased products themselves. Maybe name this `ProductInformation`? `ProductDetails`?

packages/in_app_purchase/lib/src/in_app_purchase_connection/in_app_purchase_connection.dart
mklim: I like that idea but I don't know what information we would be setting as the response code.

`BillingClient` returns an error code sometimes, and `StoreKit` returns a list of `invalidProductIdentifiers`. How would we combine that in a way that makes sense?

mklim: nit: I would name this `query[product classname]`.

cyanglaz: That's also what blocking me to introduce the response class.

packages/in_app_purchase/lib/src/billing_client_wrappers/sku_details_wrapper.g.dart
cyanglaz: oh yeah I didn't know why this piece of code got changed, I'll discuss with you offline.

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
cyanglaz: SG! I will update them all.

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
cyanglaz: SG, old habit not using id as a variable name since `id` is a reserved key word in Objective-C

packages/in_app_purchase/lib/src/in_app_purchase_connection/product.dart
cyanglaz: SG

packages/in_app_purchase/lib/src/in_app_purchase_connection/in_app_purchase_connection.dart
cyanglaz: SG

mklim: Discussed offline, going to add a response with a list of the not found product IDs.

cyanglaz:
@mklim Updated with example app. Please take another look.

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -27,53 +27,115 @@ class _MyAppState extends State<MyApp> {
...
+            ),
+            Expanded(
+              child: FutureBuilder(
+                future: InAppPurchaseConnection.instance.queryProductDetails(
+                    <String>[
+                  'consumable',
+                  'gas',
+                  'premium',
+                  'upgrade',
+                  'somethingNotValid'
```
mklim: Ideally the example app should just be demoing the right way to use the API. I don't think it's worth deliberately checking for invalid product IDs here. All we're asking people to set up in the README so far is `consumable`, `upgrade`, and `subscription`.

packages/in_app_purchase/lib/src/in_app_purchase_connection/product_details.dart
```diff
@@ -0,0 +1,44 @@
...
+  /// The description of the product, specified in the App Store Connect or Sku in Google Play console based on the platform.
+  final String description;
+
+  /// The price of the product, specified in the App Store Connect or Sku in Google Play console based on the platform.
+  /// Formatted with currency symbol ("$0.99").
+  final String price;
+}
+
+/// The response returned by [InAppPurchaseConnection.queryProductDetails]
+class QueryProductDetailsResponse {
```
mklim: nit: I think it would be a little clearer and more concise to name this `ProductDetailsResponse`. The "Query" part is already implied by "Response" (it's a response to a query).

packages/in_app_purchase/lib/src/in_app_purchase_connection/product_details.dart
```diff
@@ -0,0 +1,44 @@
...
+  QueryProductDetailsResponse(
+      {@required this.productDetails, @required this.notFoundIDs});
+
+  ///Each [ProductDetails] uniquely matches one valid identifier in [identifiers] of [InAppPurchaseConnection.queryProductDetails].
```
mklim: nit: Extra whitespace before each sentence in the dartdoc here and below.

packages/in_app_purchase/lib/src/in_app_purchase_connection/product_details.dart
```diff
@@ -0,0 +1,44 @@
...
+  final List<ProductDetails> productDetails;
+
+  ///The list of identifiers that are in the `identifiers` of [InAppPurchaseConnection.queryProductDetails] but failed to be fetched.
+  ///
+  ///These are the identifiers not matching the `productIdentifer` or `sku` in any product on the App Store Connect or Google Play Console.
```
mklim: I'd also avoid mentioning the underlying platforms here. "failed to be fetched" is about as specific as we can safely get at this layer without going into a bunch of platform specific details I think. For example I know in the App Store case `invalidIdentifiers` will be full of "valid" products in a lot of cases where the IDs do match whatever is in App Store connect and the problem is really in some other configuration in either the app or its store listing.

Maybe we could broadly point the users to look into the underlying platforms if they're seeing this failing? Like, "There's multiple platform specific reasons that product information could fail to be fetched, ranging from products not being correctly configured in the storefront to the queried IDs not existing." That's unfortunately not very helpful on its own, but I think that's really all the information we have at this level.

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -27,53 +27,115 @@ class _MyAppState extends State<MyApp> {
...
+  List<ListTile> _buildProductList(QueryProductDetailsResponse response) {
+    List<ListTile> productDetailsCards = response.productDetails.map(
+      (ProductDetails productDetails) {
+        return buildListCard(ListTile(
+          title: Text(
+            productDetails.title,
+          ),
+          subtitle: Text(
+            productDetails.description,
+          ),
```
mklim: nit, optional: Can we display formatted price string here somewhere too? Doesn't really matter where for now, maybe just in parens as part of the title (eg "Product title ($X.XX)").

packages/in_app_purchase/lib/src/billing_client_wrappers/sku_details_wrapper.dart
```diff
@@ -124,6 +125,16 @@ class SkuDetailsWrapper {
         type.hashCode,
         isRewarded.hashCode);
   }
+
+  /// Method to convert to the wrapper to the consolidated [ProductDetails] class.
```
mklim: nit: I think it's worth warning that this conversion is one-way since we're losing data.

packages/in_app_purchase/lib/src/in_app_purchase_connection/app_store_connection.dart
```diff
@@ -26,4 +27,26 @@ class AppStoreConnection implements InAppPurchaseConnection {
     _instance = AppStoreConnection();
     return _instance;
   }
+
+  /// Query the product detail list.
+  ///
+  /// This method only returns [QueryProductDetailsResponse].
```
mklim: FYI I don't think this or `GooglePlayConnection` really need dart doc, since we're not exposing either of them as public API classes and they're overriding the documented public interface. You also don't need to delete documentation you've already written though.

packages/in_app_purchase/lib/src/in_app_purchase_connection/google_play_connection.dart
```diff
@@ -59,4 +60,33 @@ class GooglePlayConnection
...
+
+  /// Query the product detail list.
+  ///
+  /// This method only returns [QueryProductDetailsResponse].
+  /// To get detailed Google Play sku list, use [BillingClient.querySkuDetails]
+  /// to get the [SkuDetailsResponseWrapper].
+  Future<QueryProductDetailsResponse> queryProductDetails(
+      Set<String> identifiers) async {
+    SkuDetailsResponseWrapper inappResponse =
+        await _billingClient.querySkuDetails(
```
mklim: nit: It would be a little faster if these two queries happened in parallel and we blocked on them both completing in general, instead of blocking on both of them sequentially.

I think you can use [`Future.wait`](https://api.dartlang.org/stable/2.1.0/dart-async/Future/wait.html) to block on multiple futures resolving.

packages/in_app_purchase/lib/src/in_app_purchase_connection/google_play_connection.dart
```diff
@@ -59,4 +60,33 @@ class GooglePlayConnection
...
+            skuType: SkuType.inapp, skusList: identifiers.toList());
+    SkuDetailsResponseWrapper subResponse = await _billingClient
+        .querySkuDetails(skuType: SkuType.subs, skusList: identifiers.toList());
+    List<ProductDetails> inappProducts = inappResponse.skuDetailsList
```
mklim: nit, subjective: instead of writing this conversion twice what about combining these lists first?

Something like:

```
List<ProductDetails> products = (inappResponse.skuDetailsList + subResponse.skuDetailsList)
    .map((SkuDetailsWrapper wrapper) => //etc
```

packages/in_app_purchase/lib/src/in_app_purchase_connection/in_app_purchase_connection.dart
```diff
@@ -6,12 +6,17 @@ import 'dart:async';
...
 import 'app_store_connection.dart';
 import 'google_play_connection.dart';
+import 'product_details.dart';
 
 /// Basic generic API for making in app purchases across multiple platforms.
 abstract class InAppPurchaseConnection {
   /// Returns true if the payment platform is ready and available.
   Future<bool> isAvailable();
 
+  /// Query the product details list.
```
mklim: nit: I don't think we're really querying from a pre-existing list of product details here. It's more like we're querying for product details that match the given set of identifiers.

packages/in_app_purchase/lib/src/in_app_purchase_connection/product_details.dart
```diff
@@ -0,0 +1,44 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+
+/// The class represents the information of a product.
+///
+/// A list of [ProductDetails] can be obtained from the [QueryProductDetailsResponse].
```
mklim: nit, subjective: I think it would be more helpful to link to `Connection.queryProductDetails` here than the response data class specifically.

cyanglaz:
@mklim Updated with suggested reviews.

PR #1172, opened 2019-02-07
Iap add payment objc translators

cyanglaz:


cyanglaz:
@mklim Last PR caused some issue after I resolved conflicts. I reverted the commit made by last PR and this is an exact duplicate of last PR without conflict.

PR #1178, opened 2019-02-07
Iap iOS add payment dart wrappers

cyanglaz:
[iOS]
Dart wrappers for the required objects that are used when adding a payment. 
The code for adding a payment will be in a following PR after this one if approved and merged. 
https://github.com/flutter/flutter/issues/26328


packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -3,12 +3,280 @@
...
+
+  /// The app canceled the download.
+  @JsonValue(5)
+  cancelled,
+}
+
+/// Dart wrapper around StoreKit's [SKDownload](https://developer.apple.com/documentation/storekit/skdownload?language=objc).
+///
+/// When a product is created in the App Store Connect, one or more download contents can be associated with it.
+/// When the product is purchased, a List of download object will present in an [SKPaymentTransactionWrapper] object.
```
mklim: nit: "download object will present in an [SKPaymentTransactionWrapper] object" is a little hard to parse. I think it's mainly because [when "present" is used this way](https://www.merriam-webster.com/dictionary/present) it's a [transitive verb](https://en.wikipedia.org/wiki/Transitive_verb#In_English). Transitive verbs need an object to make sense, and it's missing one here. "download object" should probably also be plural since there's multiple in a List.

Maybe consider rewording to "[SkPaymentTransactionWrapper] will return a list of downloads"?

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -3,12 +3,280 @@
...
+/// Then download objects can be added to the payment queue.
+/// Read the [contentURL] to get the URL of the downloaded content after download complete.A
```
mklim: nit: "after the download completes."

"download" needs a definite article, and "to complete" needs to be in the singular form of "to complete" [so that it matches the number of downloads](https://en.wikipedia.org/wiki/Agreement_(linguistics)#Number).

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -3,12 +3,280 @@
...
+/// All downloaded files must be processed before the completion of the transaction.
+/// After the transaction is complete, any download object in the transaction will not be able to add to the payment queue
```
mklim: Is "download object" talking about a type of object that has the property of being able to be downloaded? Or is it talking about an object that has been downloaded in the past?

I'm asking because this wording is hard to parse as it is. "downloadable object" would be using "download" in the form of an adjective, and saying that the object has the quality of being able to be downloaded. "downloaded object" would be using "download" as a [past participle](https://en.wikipedia.org/wiki/Participle#Types_of_participle), and saying that the objects have definitely been downloaded at some prior point.

mklim: nit: "will not be able to add to the payment queue" -> "will not be able to be added to the payment queue"

I'm having trouble finding a formal article for why this is normally done, but it's described here as [verb tense consistency](https://www.grammarly.com/blog/verb-tense-consistency/).

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -3,12 +3,280 @@
...
+  final String transactionID;
+
+  /// The download progress, between 0.0 to 1.0.
+  final double progress;
+
+  /// The estimated time remaining for the download; if no good estimate is able to be made,
+  /// [downloadTimeUnknown] will be set to true.
+  final double timeRemaining;
+
+  /// true if [timeRemaining] cannot be estimated.
```
mklim: nit: "True"

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -3,12 +3,280 @@
...
+/// Used as the parameter to initiate a payment.
+/// In general, a developer should not need to create the payment object explicitly; instead, use
+/// [SKPaymentQueueWrapper.addPayment] directly with a product identifier to initiate a payment.
+@JsonSerializable()
+class SKPaymentWrapper {
+  SKPaymentWrapper(
+      {@required this.productIdentifier,
+      @required this.applicationUsername,
+      this.requestData,
+      this.quantity = 1,
```
mklim: I think it's worth documenting that quantity defaults to 1 if unspecified.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
cyanglaz: going to rephrase the entire paragraph.

cyanglaz: be added is the correct way to do it, I actually also found it out myself while I'm reading it myself.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
cyanglaz: This is done on the property itself, do you mean to add a dart doc here as well? Or should we just have a comment?

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -10,6 +10,9 @@ import 'package:json_annotation/json_annotation.dart';
 part 'sk_payment_queue_wrapper.g.dart';
 
 /// A wrapper around [`SKPaymentQueue`](https://developer.apple.com/documentation/storekit/skpaymentqueue?language=objc).
+///
+/// The payment queue contains payment related operations. It communicates with App Store and presents
+/// a user interface for user to process and authorize the payment.
```
mklim: nit: "for user" either needs a definite article ("for the user") or to be plural ("for users").

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -105,6 +113,8 @@ class SKPaymentTransactionWrapper {
 }
 
 /// Dart wrapper around StoreKit's [SKDownloadState](https://developer.apple.com/documentation/storekit/skdownloadstate?language=objc).
+///
+/// The state of a download operation that can be in.
```
mklim: nit: remove "of", so "The state a download operation..."

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -134,12 +144,12 @@ enum SKDownloadState {
 /// Dart wrapper around StoreKit's [SKDownload](https://developer.apple.com/documentation/storekit/skdownload?language=objc).
 ///
 /// When a product is created in the App Store Connect, one or more download contents can be associated with it.
-/// When the product is purchased, a List of download object will present in an [SKPaymentTransactionWrapper] object.
-/// Then download objects can be added to the payment queue.
-/// Read the [contentURL] to get the URL of the downloaded content after download complete.A
-/// All downloaded files must be processed before the completion of the transaction.
-/// After the transaction is complete, any download object in the transaction will not be able to add to the payment queue
-/// and the contentURL of the download object will be invalid.
+/// When the product is purchased, a List of [SKDownloadWrapper] object will present in an [SKPaymentTransactionWrapper] object.
```
mklim: nit: "[SKDownloadWrapper] objects will be present"

cyanglaz:
Fixed the last nit comments, will merge in on green

PR #1181, opened 2019-02-08
Add `openUrl` action support for webview plugin.

kangwang1988:
Add `openUrl` action support for webview plugin.
This will fix: https://github.com/flutter/flutter/issues/27560


kangwang1988:
@mklim Friendly ping.

PR #1182, opened 2019-02-08
[firebase_dynamic_links] sometimes got error NSPOSIXErrorDomain Code=53 in ios

ko2ic:
As in link, there is a case of nil at the completion of handleUniversalLink.
https://github.com/firebase/firebase-ios-sdk/issues/2303

Also, even if  sleep,  cause an error, so fixed it to tell
it to the dart side.

https://github.com/flutter/flutter/issues/27810

### DIFF
--- a/packages/firebase_dynamic_links/CHANGELOG.md
+++ b/packages/firebase_dynamic_links/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.2.1
+
+* Throw `PlatformException` if there is an error retrieving dynamic link.
+
 ## 0.2.0+4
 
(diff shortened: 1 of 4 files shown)


collinjackson:
I don't think our current `flutter drive` based integration testing can regression test this change, so I'm ok with merging it as-is. We should get rid of the NSError categories to match other plugins, but that can happen in a subsequent PR.

ko2ic:
@collinjackson @bparrishMines 

It became all green. Could you merge?

bparrishMines:
Yea I can merge it for you. I added a version bump and a corresponding part to the android side.

Thanks for the contribution! I'll merge once it goes green again.

PR #1183, opened 2019-02-08
[google_maps_flutter] Add support for map marker bitmap created “fromFile”

gahms:
Support for creating markers using “defaultMarker” or “fromAsset” is already supported. With “fromFile” support for adding local image files is added.


yuridiniz:
Any news?

KirolousNashaat:
if we could pass byte array that would be helpful too
in case we want to download it from web but without creating file on disk
with 'BitmapDescriptorFactory.fromBitmap' and create bitmap from byte array

iskakaushik:
now that we have `fromBytes` do we still need this?

PR #1184, opened 2019-02-08
[google_maps_flutter] In Google Maps, add support for map style

gahms:
On Dart side it is a string as the Google Maps style class is very closed and can only be created from a string.


iskakaushik:
We now have support for map style.

PR #1186, opened 2019-02-10
[camera] 4k camera

mclark4386:
Adds a 4k option for ResolutionPreset. Would love feedback on this!


mclark4386:
seems to have been brought in from somewhere else

PR #1187, opened 2019-02-10
Update FirebaseAuthPlugin.java

HunRex:
A lot of people have problem with the getCredential(...) method, because there is no switch case for PhoneAuthProvider, making it impossible to get the credentials for phone authentication. This few lines fix the nullPointer exception problem for phone authentication


collinjackson:
This is fixed in #1176

PR #1191, opened 2019-02-11
[camera] Added a check for pictureImageReader != null

SaltumDis:
Fixed an error `Attempt to invoke virtual method 'android.view.Surface android.media.ImageReader.getSurface()' on a null object reference`

This is the bug report:
https://github.com/flutter/flutter/issues/19595

### DIFF
--- a/packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
+++ b/packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
@@ -720,7 +720,10 @@ public class CameraPlugin implements MethodCallHandler {
       Surface previewSurface = new Surface(surfaceTexture);
       surfaces.add(previewSurface);
       captureRequestBuilder.addTarget(previewSurface);
-
+      if (pictureImageReader == null) {
+        sendErrorEvent("Failed to configure the camera for preview.");
+        return;
+      }
       surfaces.add(pictureImageReader.getSurface());
 
       cameraDevice.createCaptureSession(


SaltumDis:
I signed it!

SaltumDis:
Hello @bparrishMines ! Any plans to merge this? 😺

bparrishMines:
@SaltumDis With a second glance, I think we would need a solution that also handles why this would be null. The only way this can happen is when `startPreview()` was called on a closed camera.

SaltumDis:
I am not sure how to handle this. 
Did some debugging and it seems that
```java
          @Override
          public void onActivityResumed(Activity activity) {
            boolean wasRequestingPermission = requestingPermission;
            if (requestingPermission) {
              requestingPermission = false;
            }
            if (activity != CameraPlugin.this.activity) {
              return; # this is called
            }
            ...
```
this error occurs when `activity != CameraPlugin.this.activity`

`onActivityPaused` we close the camera, `onActivityResumed` we still leave it closed.
So openCamera callback `onOpened` is being called on closed camera. 

@bparrishMines any suggestions?

bparrishMines:
I think the underlying cause it that we attempt to control opening/closing the camera on Android. I think the best solution would be to gut this part of the plugin and require the user to handle opening/closing. Especially since there is no iOS equivalent of the same control.

 I'll create a PR within the next couple days that will remove this part and will see if it causes any problems.

SaltumDis:
https://github.com/flutter/plugins/pull/1477 - is merged - bug is fixed

PR #1192, opened 2019-02-11
Flutterfire apps no longer show a confusing log message about configuring default app

collinjackson:
Fixes flutter/flutter#16534

Note that this doesn't completely fix the issue, it just changes the error message from the previous, really confusing:

```
"4.8.1 - [Firebase/Core][I-COR000003] The default Firebase app has not yet been configured.
Add `[FIRApp configure];` (`FirebaseApp.configure()` in Swift) to your application
initialization.
```

New log message:

```
[Firebase/Core][I-COR000004] App with name __FIRAPP_DEFAULT does not exist.
```

### DIFF
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.9.1
+
+* No longer shows a confusing log message about configuring default app
+
 ## 0.9.0+1
 
 * Log a more detailed warning at build time about the previous AndroidX
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -1,6 +1,6 @@
-## 0.9.1
+## 0.9.0+2
 
-* No longer shows a confusing log message about configuring default app
+* Log messages about automatic configuration of the default app are now less confusing.
 
 ## 0.9.0+1
 
--- a/packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
+++ b/packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
@@ -278,6 +278,7 @@ const UInt8 TIMESTAMP = 136;
   self = [super init];
   if (self) {
     if (![FIRApp appNamed:@"__FIRAPP_DEFAULT"]) {
+      NSLog(@"Configuring the default Firebase app...");
       [FIRApp configure];
     }
     _listeners = [NSMutableDictionary<NSNumber *, id<FIRListenerRegistration>> dictionary];
--- a/packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
+++ b/packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
@@ -274,6 +274,7 @@ const UInt8 TIMESTAMP = 136;
     if (![FIRApp appNamed:@"__FIRAPP_DEFAULT"]) {
       NSLog(@"Configuring the default Firebase app...");
       [FIRApp configure];
+      NSLog(@"Configured the default Firebase app %@.", [FIRApp defaultApp].name);
     }
     _listeners = [NSMutableDictionary<NSNumber *, id<FIRListenerRegistration>> dictionary];
     _batches = [NSMutableDictionary<NSNumber *, FIRWriteBatch *> dictionary];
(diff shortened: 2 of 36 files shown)


packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -277,7 +277,7 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
 - (instancetype)init {
   self = [super init];
   if (self) {
-    if (![FIRApp defaultApp]) {
+    if (![FIRApp appNamed:@"__FIRAPP_DEFAULT"]) {
       [FIRApp configure];
```
mklim: This is definitely an improvement but I think the new log message is still a little confusing. Could we log out a warning about how the user needs to set up their app and to check the readme in the various else cases here?

collinjackson: There's no action for the user to take here. It's going to happen for every Flutterfire user, regardless of whether they are set up or not. I wish that we could suppress the log message that the native SDK is sending, but we don't have a way to do that -- see https://github.com/firebase/firebase-ios-sdk/pull/1003

diver2: I do agree with changing the error message to add something that "This is not a problem if you are using flutter."
I just took 20mins to understand this following threads - and at least liked the response from @collinjackson that "there's no action for the user... It is going to happen for every Flutterfire user, regardless of..."

collinjackson: I added an additional log message explaining what is happening when Flutterfire automatically configures the default Firebase app. @mklim let me know if you like this version or if you prefer the previous.

packages/cloud_firestore/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.9.1
```
cyanglaz: If there is no public api change, the version should only be updated as a patch, so 0.9.0+2
Same for everything related in this PR.
@collinjackson @mklim can you confirm?

mklim: Good catch, this is more of a bugfix and should just be bumping the patch numbers everywhere. [Sometimes it is appropriate to increment the minor version just based on the behavior of private code](https://semver.org/#spec-item-7) but I don't think this change really fits that edge case.

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
mklim: Ah understood, thanks. Can you paraphrase your comment and add it into the changelogs or commit message? That explains this a lot more clearly. This new log LGTM.

packages/cloud_firestore/CHANGELOG.md
collinjackson: I'm fine with doing this as 0.9.0+2

collinjackson:
I switched to using patch versions, will land this once the bots are happy

Hixie:
test?

packages/cloud_firestore/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.9.0+2
```
bparrishMines: Doesn't the `pubspect.yaml` need the version bump too?

packages/cloud_functions/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.1.0+2
```
bparrishMines: `pubspec.yaml` version bump

packages/firebase_dynamic_links/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.2.0+2
```
bparrishMines: `pubspec.yaml` version

packages/firebase_performance/pubspec.yaml
```diff
@@ -4,7 +4,8 @@ description: Flutter plugin for Google Performance Monitoring for Firebase, an a
   iOS.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/firebase_performance
-version: 0.1.0+2
+version: 0.1.0+3
+>>>>>>> 4f024c7404c01e1f3b55be61704d2deb53f6ca41
```
bparrishMines: merge line

collinjackson:
> test?

I don't think that our current infrastructure can test this change, because the only observable impact is in the iOS device logs.

PR #1198, opened 2019-02-12
 [shared_preferences] Added reload method 

wouterhardeman:
Resolves https://github.com/flutter/flutter/issues/27490

The `refreshCache()` function clears the local cache and fetches the shared preferences from the host platform.


cyanglaz:
Thank you for the contribution! The change looks good to me! Could you please rebase and resolve the conflicts? Also, could you add integration test for your new method? The integration tests should be in the example/test folder. Please refer to the [contribution guide](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md) for running integration test locally. .

collinjackson:
I think that `reload()` is shorter than `refreshCache()` and might make more sense (none of the other method names mention a cache). @cyanglaz do you agree?

PR #1199, opened 2019-02-12
Iap fix requesthandler crash

cyanglaz:
The FIAPRequestHandler is destroyed when exiting the function. Since FIAPRequestHandler is the delegate, when SKProductRequest tries to execute the delegate method, it will not be able to find the delegate. Hence the crash.

Adding a set to hold strong reference of the FIAPRequestHandler until it is complete fix the crash.


cyanglaz:
Should be reviewed after https://github.com/flutter/plugins/pull/1169 has landed

PR #1200, opened 2019-02-12
WebView controller let's you clear cache

iskakaushik:


iskakaushik:
@amirh I was trying to come up with ways to effectively test this or have an example for this. I wasn't able to find any JS methods that let you list the entries of the browser cache. Let me know if you have any ideas..

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -394,6 +394,16 @@ class WebViewController {
     return _channel.invokeMethod("reload");
   }
 
+  /// Clears the resource cache.
+  ///
+  /// Note that the cache is per-application, so this will clear the cache for all WebViews used.
+  Future<void> clearCache() async {
```
amirh: Looks like Android and iOS have some fine grained control over what exactly is being cleared.
I think it's ok to clear all (if we want to provide finer grained control we could add optional parameters here without it being a breaking change), but lets note in the documentation exactly what is being cleared on each platform.

iskakaushik: @amirh noted in the docs.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -394,6 +394,24 @@ class WebViewController {
     return _channel.invokeMethod("reload");
   }
 
+  /// Clears all of the following caches.
```
amirh: This shows as the summary line in the Dartdoc site, it should be self contained, probably should just say "Clears all caches used by the [WebView]."

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -394,6 +394,24 @@ class WebViewController {
...
+  ///
+  /// There are multiple caches we need to be aware of.
+  ///	1. Browser HTTP Cache.
+  ///	2. Caches API caches. https://developers.google.com/web/fundamentals/instant-and-offline/web-storage/cache-api,
```
amirh: Caches API caches? typo?

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -394,6 +394,24 @@ class WebViewController {
...
+  ///    These are not yet supported in IOS WkWebView. Service workers tend to use this cache.
```
amirh: nit: iOS

packages/webview_flutter/lib/webview_flutter.dart
amirh: Maybe would be more clear as:
```dart
/// [Caches API](https://developers.google.com/web/fundamentals/instant-and-offline/web-storage/cache-api) cache.
```

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -394,6 +394,24 @@ class WebViewController {
     return _channel.invokeMethod("reload");
   }
 
+  /// Clears all of the following caches.
+  ///
+  /// There are multiple caches we need to be aware of.
```
amirh: Suggestion: "The following caches are cleared:"

PR #1204, opened 2019-02-13
Remove assertion for firebase_core that can interfere with hot-restart

collinjackson:
Fixes flutter/flutter#17929

### DIFF
--- a/packages/firebase_core/pubspec.yaml
+++ b/packages/firebase_core/pubspec.yaml
@@ -3,7 +3,7 @@ description: Flutter plugin for Firebase Core, enabling connecting to multiple
   Firebase apps.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/firebase_core
-version: 0.3.0+1
+version: 0.3.1
 
 flutter:
   plugin:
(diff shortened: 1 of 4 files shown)


packages/firebase_core/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Flutter plugin for Firebase Core, enabling connecting to multiple
   Firebase apps.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/firebase_core
-version: 0.3.0+1
+version: 0.3.1
```
cyanglaz: Prob should be 0.3.0+2. Based on `Although semantic versioning doesn’t promise any compatibility between versions prior to 1.0.0, the Dart community convention is to treat those versions semantically as well. The interpretation of each number is just shifted down one slot: going from 0.1.2 to 0.2.0 indicates a breaking change, going to 0.1.3 indicates a new feature, and going to 0.1.2+1 indicates a change that doesn’t affect the public API.` quoted here https://www.dartlang.org/tools/pub/versioning#semantic-versions

cyanglaz:
LGTM, might need to change the version per "Pub Versioning Philosophy"

packages/firebase_core/pubspec.yaml
collinjackson: Although this is just an assertion, it feels to me like a functionality change, because it changed behavior in a non-breaking way (hot restart now works in debug mode, previously it threw errors).

cyanglaz:
download

PR #1205, opened 2019-02-13
Introduce CODEOWNERS

collinjackson:
See https://github.blog/2017-07-06-introducing-code-owners/

The idea here is to identify a default reviewer for pull requests automatically.

We don't have the protected branch feature on. That's an option to consider in the future, but I think it's relatively common to do a push of all first-party plugins without getting CODEOWNERS signoff.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,17 @@
+packages/cloud_firestore/* @collinjackson @kroikie
+packages/firebase_auth/* @collinjackson @kroikie
+packages/firebase_messaging/* @collinjackson @kroikie
+packages/camera/* @bparrishMines 
+packages/firebase_admob/* @amirh
+packages/firebase_ml_vision/* @bparrishMines
+packages/image_picker/* @cyanglaz
+packages/android_alarm_manager/* @bkonyi
+packages/google_maps_flutter/* @amirh
+packages/google_sign_in/* @cyanglaz
+packages/firebase_database/* @collinjackson @kroikie
+packages/firebase_storage/* @collinjackson @kroikie
+packages/package_info/* @cyanglaz
+packages/firebase_analytics/* @collinjackson @kroikie
+packages/connectivity/* @cyanglaz
+packages/webview_flutter/* @amirh
+
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,6 @@
+# Below is a list of Flutter team members who are suggested reviewers
+# for contributions to plugins in this repository.
+
 packages/cloud_firestore/* @collinjackson @kroikie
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,20 @@
+# Below is a list of Flutter team members who are suggested reviewers
+# for contributions to plugins in this repository.
+
+packages/cloud_firestore/* @collinjackson @kroikie
+packages/firebase_auth/* @collinjackson @kroikie
+packages/firebase_messaging/* @collinjackson @kroikie
+packages/camera/* @bparrishMines
```
amirh: @mklim ok to add you as an additional point of contact here?

CODEOWNERS
```diff
@@ -0,0 +1,20 @@
...
+packages/cloud_firestore/* @collinjackson @kroikie
+packages/firebase_auth/* @collinjackson @kroikie
+packages/firebase_messaging/* @collinjackson @kroikie
+packages/camera/* @bparrishMines 
+packages/firebase_admob/* @amirh
+packages/firebase_ml_vision/* @bparrishMines
+packages/image_picker/* @cyanglaz
+packages/android_alarm_manager/* @bkonyi
+packages/google_maps_flutter/* @amirh
+packages/google_sign_in/* @cyanglaz
```
amirh: Let's list @mehmetf here (confirmed with him offline)

CODEOWNERS
```diff
@@ -0,0 +1,20 @@
+# Below is a list of Flutter team members who are suggested reviewers
+# for contributions to plugins in this repository.
```
amirh: What do you think about emphasizing that it doesn't mean the default reviewer has to be the one reviewing the PR? ("suggested reviewers" says that already, but I'd feel better if we emphasize 😄  )

CODEOWNERS
```diff
@@ -0,0 +1,20 @@
...
+packages/android_alarm_manager/* @bkonyi
+packages/google_maps_flutter/* @amirh
+packages/google_sign_in/* @cyanglaz
+packages/firebase_database/* @collinjackson @kroikie
+packages/firebase_storage/* @collinjackson @kroikie
+packages/package_info/* @cyanglaz
+packages/firebase_analytics/* @collinjackson @kroikie
+packages/connectivity/* @cyanglaz
+packages/webview_flutter/* @amirh
+
```
amirh: we can add @mklim for IAP

CODEOWNERS
mklim: SGTM

CODEOWNERS
mklim: Yeah, please do.

PR #1207, opened 2019-02-13
Fix cloud_firestore dates on some iOS devices

collinjackson:
Fixes flutter/flutter#18470

### DIFF
--- a/packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
+++ b/packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
@@ -189,8 +189,7 @@ const UInt8 TIMESTAMP = 136;
     case DATE_TIME: {
       SInt64 value;
       [self readBytes:&value length:8];
-      NSTimeInterval time = [NSNumber numberWithLong:value].doubleValue / 1000.0;
-      return [NSDate dateWithTimeIntervalSince1970:time];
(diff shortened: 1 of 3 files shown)


packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -183,8 +183,7 @@ - (id)readValueOfType:(UInt8)type {
     case DATE_TIME: {
       SInt64 value;
       [self readBytes:&value length:8];
-      NSTimeInterval time = [NSNumber numberWithLong:value].doubleValue / 1000.0;
-      return [NSDate dateWithTimeIntervalSince1970:time];
+      return [NSDate dateWithTimeIntervalSince1970:(value / 1000.0)];
```
cyanglaz: Just out of curiosity, what is the difference between [NSNumber numberWithLong:value].doubleValue / 1000.0 and value / 1000.0

collinjackson: See https://github.com/flutter/flutter/issues/18470

collinjackson: My understanding is that one of them is an Objective-C object and the other is an unboxed value.

PR #1208, opened 2019-02-13
[firebase_storage] Support for getReferenceFromUrl

collinjackson:
Fixes flutter/flutter#20458

I think it could use an example/test that exercises the native side

### DIFF
--- a/packages/firebase_storage/lib/src/firebase_storage.dart
+++ b/packages/firebase_storage/lib/src/firebase_storage.dart
@@ -126,6 +126,8 @@ class FirebaseStorage {
     });
   }
 
+  /// Creates a [StorageReference] given a gs:// or // URL pointing to a Firebase
+  /// Storage location. 
   Future<StorageReference> getReferenceFromUrl(String fullUrl) async {
     String path = await channel.invokeMethod(
         "FirebaseStorage#getReferenceFromUrl", <String, dynamic>{
--- a/packages/firebase_storage/lib/src/firebase_storage.dart
+++ b/packages/firebase_storage/lib/src/firebase_storage.dart
@@ -127,9 +127,9 @@ class FirebaseStorage {
   }
 
   /// Creates a [StorageReference] given a gs:// or // URL pointing to a Firebase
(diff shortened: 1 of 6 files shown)


packages/firebase_storage/lib/src/firebase_storage.dart
```diff
@@ -125,6 +125,20 @@ class FirebaseStorage {
       'time': time,
     });
   }
+
+  Future<StorageReference> getReferenceFromUrl(String fullUrl) async {
```
cyanglaz: Need Dart doc for public API.

packages/firebase_storage/lib/src/firebase_storage.dart
```diff
@@ -125,6 +125,20 @@ class FirebaseStorage {
...
+    String path = await channel.invokeMethod(
+        "FirebaseStorage#getReferenceFromUrl", <String, dynamic>{
+      'app': app?.name,
```
cyanglaz: `app` and `bucket` seems not used in the platform code.

packages/firebase_storage/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Flutter plugin for Firebase Cloud Storage, a powerful, simple, and
   cost-effective object storage service for Android and iOS.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/firebase_storage
-version: 2.0.1
+version: 2.0.2
```
cyanglaz: Should be 2.1.0 since we added a new public API. 

collinjackson: good catch, thanks!

packages/firebase_storage/lib/src/firebase_storage.dart
collinjackson: They are used in the `getStorage:` method (on iOS) and at the beginning of onMethodCall (on Android) to inflate the `FirebaseStorage` object on the native side.

PR #1211, opened 2019-02-13
[video_player] Future already completed fix

Yarikk26:
Workaround for https://github.com/flutter/flutter/issues/19197.
Checked on the video from example project and on my use-cases.

### DIFF
--- a/packages/video_player/ios/Classes/VideoPlayerPlugin.m
+++ b/packages/video_player/ios/Classes/VideoPlayerPlugin.m
@@ -249,14 +249,16 @@ static void* playbackBufferFullContext = &playbackBufferFullContext;
       case AVPlayerItemStatusUnknown:
         break;
       case AVPlayerItemStatusReadyToPlay:
-        _isInitialized = true;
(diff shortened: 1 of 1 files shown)


cbenhagen:
I don't think this is an appropriate fix as [playbackLikelyToKeepUp](https://developer.apple.com/documentation/avfoundation/avplayeritem/1390348-playbacklikelytokeepup) has nothing to do with being initialized.

Either we make sure to only send the initialized event only once from the iOS side. Or we do this in dart:

```dart
if (!value.initialized) {
  value = value.copyWith(
    duration: Duration(milliseconds: map['duration']),
    size: Size(map['width']?.toDouble() ?? 0.0,
        map['height']?.toDouble() ?? 0.0),
  );
  initializingCompleter.complete(null);
  _applyLooping();
  _applyVolume();
  _applyPlayPause();
}
```

iskakaushik:
This should be addressed in version: `0.10.0+8` by this PR: https://github.com/flutter/plugins/pull/1552/ please re-open this if it's still an issue. Thanks!

PR #1214, opened 2019-02-13
Video player supported formats

iskakaushik:


iskakaushik:
https://github.com/flutter/flutter/issues/13909

packages/video_player/README.md
```diff
@@ -40,6 +40,14 @@ Ensure the following permission is present in your Android Manifest file, locate
 
 The Flutter project template adds it, so it may already be there.
 
+### Supported Formats
+
+- On iOS, the backing player is [AVPlayer](https://developer.apple.com/documentation/avfoundation/avplayer).
+  The supported formats vary depending on the version of iOS, [AVURLAsset](https://developer.apple.com/documentation/avfoundation/avurlasset) class
```
xster: Instead of linking these classes here, I would briefly look at what they output and copy out a summary of the version supports here. Since it's a Flutter plugin, we're not necessarily expecting our users to be versed in iOS APIs. They may not have any iOS code at all in their project.

PR #1224, opened 2019-02-14
[firebase_ml_vision] Flip detected barcode coordinates on Android

gi097:
Android uses reversed coordinates, flip those coordinates to draw a detection box correctly.

### DIFF
--- a/packages/firebase_ml_vision/example/lib/detector_painters.dart
+++ b/packages/firebase_ml_vision/example/lib/detector_painters.dart
@@ -6,6 +6,7 @@ import 'dart:ui' as ui;
 
 import 'package:firebase_ml_vision/firebase_ml_vision.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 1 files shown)


bparrishMines:
Hi @gi097,

Where did you see/experience that they were flipped? I didn't need to flip coordinates in the [example app](https://github.com/flutter/plugins/blob/master/packages/firebase_ml_vision/example/lib/detector_painters.dart#L19).

gi097:
I faced this issue when running the example code (the code I changed here) on a Motorola Moto G4.

bparrishMines:
Oh, thats right. I was thinking this was an issue not a PR.

Anyways, I tested on a G4 and was unable to reproduce the bug. It may have been fixed by this [PR](https://github.com/flutter/plugins/pull/1225)?

gi097:
@bparrishMines, I don't have time to test it now. Since the problem seems to be solved, I will close this PR. Thanks.

PR #1226, opened 2019-02-15
Remove categories

cyanglaz:
Fixing https://github.com/flutter/flutter/issues/27361 and https://github.com/flutter/flutter/issues/27960

Used static c function instead of category.
The plugins effected are 
cloud_firestore
firebase_admob
firebase_auth
firebase_core
firebase_database
firebase_dynamiclink
firebase_mlvision
firebase_storage
google_signin
camera

### DIFF
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.0+1
+
+* Remove categories.
+
 ## 0.4.0
 
 * **Breaking Change** Change iOS image stream format to `ImageFormatGroup.bgra8888` from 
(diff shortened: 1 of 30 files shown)


cyanglaz:
> Is there a reason why categories is a no-no? How do we prevent them from popping up again.
> 
> The PR itself LGTM as it is just an implementation detail that does not leak to Dart API.

If there are multiple categories methods that have the same name in different plugins, and an APP imported all those plugins, when in run time, objective-c will pick up a random one and use it for all of the invocations. So it can break the project since the implementations of these same named category could be different. Details in here flutter/flutter#27960.
I don't know how to restrict the category usage, maybe a custom code scanning script in CI?

PR #1227, opened 2019-02-15
fix link navigation for android webview

kmtong:
Orignally in Android, link navigation within flutter webview plugin will ask user to open with another application.

image

This PR tries to solve this issue.


kmtong:
I signed it!

PR #1229, opened 2019-02-15
[google_maps_flutter] Marker APIs are now widget based (Android)

iskakaushik:
note: This commit only deals with Android side of changes.

Additional Context: Maps Plugin is in the process of being
moved away from a controller based API to a widget based api.
This is to facilitate easier state management and address a
lot of the common issues.


iskakaushik:
I've make changes to the iOS impl at https://github.com/iskakaushik/plugins/pull/2/files, keeping it separate for now though we would want to merge it into this branch before we land this feature.

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -178,6 +153,7 @@ public void onMapReady(GoogleMap googleMap) {
     googleMap.setOnCameraIdleListener(this);
     googleMap.setOnMarkerClickListener(this);
     updateMyLocationEnabled();
+    this.markersController.setGoogleMap(googleMap);
```
amirh: nit: omit the `this.`

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/MarkerId.java
```diff
@@ -0,0 +1,43 @@
+package io.flutter.plugins.googlemaps;
+
+/**
+ * Represents the user assigned MarkerId.
+ *
+ * <p>To be kept in sync with MarkerId on dart side.
+ */
+public class MarkerId {
```
amirh: I'd just keep the marker IDs as Strings in the Android implementation, saves us some boilerplate.

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
```diff
@@ -0,0 +1,115 @@
...
+
+  boolean onMarkerTap(String googleMarkerId) {
+    MarkerId markerId = googleMapsMarkerIdToDartMarkerId.get(googleMarkerId);
+    if (markerId == null) {
+      return false;
+    }
+    methodChannel.invokeMethod("marker#onTap", Convert.toJson(markerId));
+    MarkerController markerController = markerIdToController.get(markerId);
+    if (markerController != null) {
+      return markerController.consumeTapEvents();
```
amirh: Why do we need to do this here?

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+  }
+
+  void removeMarkers(List<Object> markerIdsToRemove) {
+    if (markerIdsToRemove == null) {
+      return;
+    }
+    for (Object rawMarkerId : markerIdsToRemove) {
+      if (rawMarkerId != null) {
```
amirh: uber nit: my personal preference would be to reverse the condition and `continue` to save one nesting level. Sending you some better articulated explanation offline.

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -117,6 +118,15 @@ static Object toJson(CameraPosition position) {
     return data;
   }
 
+  static Object toJson(MarkerId markerId) {
```
amirh: Don't change it as part of this PR to stay consistent, but all of these overloads of function with the same name feels error prone to me, I'd consider given each of them a different name, Also it's not converting to JSON but to a map.

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/MarkerId.java
iskakaushik: I was thinking the same, I was hoping that having a type here will let us distinguish between dart marker id and google map api's marker id (which we are keeping as a string).

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
iskakaushik: filed: https://github.com/flutter/flutter/issues/28363

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/MarkerId.java
amirh: I think the name should cover that well enough, it's used in a relatively narrow scope anyway...

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
iskakaushik: The method contract is:
``` true if the listener has consumed the event (i.e., the default behavior should not occur); false otherwise (i.e., the default behavior should occur). The default behavior is for the camera to move to the marker and an info window to appear.
```

https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.OnMarkerClickListener.html#onMarkerClick(com.google.android.gms.maps.model.Marker)

PR #1231, opened 2019-02-15
Iap make payment objc

cyanglaz:
The objective-c portion of add payment flow.

Since there is no dart code in this PR, it could be hard to trace what is going on. Included 2 purchase flows below to better explain.

A regular flow of a user to process a purchase will be as such:
1. User queries the products by `-[InAppPurchasePlugin startProductRequest:result:]`
2. User creates a payment object using productID which is found in the result of the 1st step.
3. User make a payment calling `-[InAppPurchasePlugin addPayment:result:]` using the payment object gotten from step 2.
4. `updatedTransaction` in dart waiting for the transaction to be at state of 'purchased'
5. User finish the transaction by calling `-[InAppPurchasePlugin finishTransaction:result:]` with the transaction gotten from step 4.

A app store purchase flow will be as such:
1. Customer triggered the purchase in the app store.
2. Programmer gets a notification in a callback named `shouldAddStorePayment` in dart.
3. If `shouldAddStorePayment` return true, the payment object will be automatically added to the queue using `-[InAppPurchasePlugin addPayment:result:]` (called by our plugin code, not the programmer)
4. Same as step 4 in the regular flow from this point.

https://github.com/flutter/flutter/issues/26328

### DIFF
--- a/packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
+++ b/packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
@@ -17,7 +17,7 @@ NSString *const TestingProductID = @"testing";
 @property(nullable, copy, nonatomic) ShouldAddStorePayment shouldAddStorePayment;
 @property(nullable, copy, nonatomic) UpdatedDownloads updatedDownloads;
 
-@property (strong, nonatomic) NSMutableDictionary *transactionsSetter;
+@property(strong, nonatomic) NSMutableDictionary *transactionsSetter;
 
 @end
 
@@ -54,12 +54,13 @@ NSString *const TestingProductID = @"testing";
 }
 
 - (void)finishTransaction:(SKPaymentTransaction *)transaction {
-    @try {
-        [self.queue finishTransaction:transaction];
-    } @catch (NSException *e) {
-        // finish transaction will throw exception if the transaction type is purchasing. Raise the exception so the InAppPurchasePlugin will get notified.
-        [e raise];
-    }
+  @try {
+    [self.queue finishTransaction:transaction];
+  } @catch (NSException *e) {
+    // finish transaction will throw exception if the transaction type is purchasing. Raise the
+    // exception so the InAppPurchasePlugin will get notified.
+    [e raise];
+  }
 }
 
 #pragma mark - observing
@@ -67,9 +68,9 @@ NSString *const TestingProductID = @"testing";
 // state of transactions and finish as appropriate.
 - (void)paymentQueue:(SKPaymentQueue *)queue
     updatedTransactions:(NSArray<SKPaymentTransaction *> *)transactions {
-    for (SKPaymentTransaction *transaction in transactions) {
-        [self.transactionsSetter setObject:transaction forKey:transaction.transactionIdentifier];
-    }
+  for (SKPaymentTransaction *transaction in transactions) {
+    [self.transactionsSetter setObject:transaction forKey:transaction.transactionIdentifier];
+  }
   // notify dart through callbacks.
   if (self.transactionsUpdated) {
     self.transactionsUpdated(transactions);
@@ -121,7 +122,7 @@ NSString *const TestingProductID = @"testing";
 #pragma mark - getter
 
 - (NSDictionary *)transactions {
-    return self.transactionsSetter;
+  return self.transactionsSetter;
 }
 
 @end
--- a/packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
+++ b/packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
@@ -1,0 +1,120 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FIAPaymentQueueHandler.h"
+
+NSString *const TestingProductID = @"testing";
+
+@interface FIAPaymentQueueHandler ()
+
+@property(strong, nonatomic) SKPaymentQueue *queue;
+@property(nullable, copy, nonatomic) TransactionsUpdated transactionsUpdated;
+@property(nullable, copy, nonatomic) TransactionsRemoved transactionsRemoved;
+@property(nullable, copy, nonatomic) RestoreTransactionFailed restoreTransactionFailed;
+@property(nullable, copy, nonatomic)
+    RestoreCompletedTransactionsFinished paymentQueueRestoreCompletedTransactionsFinished;
+@property(nullable, copy, nonatomic) ShouldAddStorePayment shouldAddStorePayment;
+@property(nullable, copy, nonatomic) UpdatedDownloads updatedDownloads;
+
+@end
+
+@implementation FIAPaymentQueueHandler
+
+- (instancetype)initWithQueue:(nonnull SKPaymentQueue *)queue
+                     transactionsUpdated:(nullable TransactionsUpdated)transactionsUpdated
+                      transactionRemoved:(nullable TransactionsRemoved)transactionsRemoved
+                restoreTransactionFailed:(nullable RestoreTransactionFailed)restoreTransactionFailed
+    restoreCompletedTransactionsFinished:
+        (nullable RestoreCompletedTransactionsFinished)restoreCompletedTransactionsFinished
+                   shouldAddStorePayment:(nullable ShouldAddStorePayment)shouldAddStorePayment
+                        updatedDownloads:(nullable UpdatedDownloads)updatedDownloads {
+  self = [super init];
+  if (self) {
+    self.queue = queue;
+    self.transactionsUpdated = transactionsUpdated;
+    self.transactionsRemoved = transactionsRemoved;
+    self.restoreTransactionFailed = restoreTransactionFailed;
+    self.paymentQueueRestoreCompletedTransactionsFinished = restoreCompletedTransactionsFinished;
+    self.shouldAddStorePayment = shouldAddStorePayment;
+    self.updatedDownloads = updatedDownloads;
+  }
+  return self;
+}
+
+- (void)addPayment:(SKPayment *)payment {
+  NSString *productID = payment.productIdentifier;
+  if (self.testing) {
+    productID = TestingProductID;
+  }
+  [self.queue addPayment:payment];
+}
+
+#pragma mark - observing
+// Sent when the transaction array has changed (additions or state changes).  Client should check
+// state of transactions and finish as appropriate.
+- (void)paymentQueue:(SKPaymentQueue *)queue
+    updatedTransactions:(NSArray<SKPaymentTransaction *> *)transactions {
+  // notify dart through callbacks.
+  if (self.transactionsUpdated) {
+    self.transactionsUpdated(transactions);
+  }
+  for (SKPaymentTransaction *transaction in transactions) {
+    switch (transaction.transactionState) {
+        // The following three states indicates that the transaction has been complete.
+        // We mark the transaction to be finished and send the signal back to dart.
+      case SKPaymentTransactionStatePurchased:
+      case SKPaymentTransactionStateFailed:
+      case SKPaymentTransactionStateRestored:
+        // mark finished transaction as finished as required by OBJC api.
+        [queue finishTransaction:transaction];
+        break;
+      default:
+        break;
+    }
+  }
+}
+
+// Sent when transactions are removed from the queue (via finishTransaction:).
+- (void)paymentQueue:(SKPaymentQueue *)queue
+    removedTransactions:(NSArray<SKPaymentTransaction *> *)transactions {
+  if (self.transactionsRemoved) {
+    self.transactionsRemoved(transactions);
+  }
+}
+
+// Sent when an error is encountered while adding transactions from the user's purchase history back
+// to the queue.
+- (void)paymentQueue:(SKPaymentQueue *)queue
+    restoreCompletedTransactionsFailedWithError:(NSError *)error {
+  if (self.restoreTransactionFailed) {
+    self.restoreTransactionFailed(error);
+  }
+}
+
+// Sent when all transactions from the user's purchase history have successfully been added back to
+// the queue.
+- (void)paymentQueueRestoreCompletedTransactionsFinished:(SKPaymentQueue *)queue {
+  if (self.paymentQueueRestoreCompletedTransactionsFinished) {
+    self.paymentQueueRestoreCompletedTransactionsFinished();
+  }
+}
+
+// Sent when the download state has changed.
+- (void)paymentQueue:(SKPaymentQueue *)queue updatedDownloads:(NSArray<SKDownload *> *)downloads {
+  if (self.updatedDownloads) {
+    self.updatedDownloads(downloads);
+  }
+}
+
+// Sent when a user initiates an IAP buy from the App Store
+- (BOOL)paymentQueue:(SKPaymentQueue *)queue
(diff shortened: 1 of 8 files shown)


packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+    }
     result([FIAObjectTranslator getMapFromSKProductsResponse:response]);
     [weakSelf.requestHandlers removeObject:handler];
   }];
 }
 
+- (void)createPaymentWithProductID:(FlutterMethodCall *)call result:(FlutterResult)result {
+  if (![call.arguments isKindOfClass:[NSString class]]) {
+    result([FlutterError
+        errorWithCode:@"storekit_invalide_argument"
```
mklim:
```suggestion
        errorWithCode:@"storekit_invalid_argument"
```

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+              details:call.arguments]);
+    return;
+  }
+  NSString *productID = call.arguments;
+  SKProduct *product = [self.productsCache objectForKey:productID];
+  if (!product) {
+    result([FlutterError
+        errorWithCode:@"storekit_product_not_found"
+              message:@"Cannot find the product. To create a payment of a product, you must query "
+                      @"the product with SKProductRequestMaker.startProductRequest."
```
mklim: nit, I think this would be a little clearer.

```suggestion
                      @"the product with SKProductRequestMaker.startProductRequest first."
```

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+    return;
+  }
+  SKPayment *payment = [SKPayment paymentWithProduct:product];
+  [self.paymentsCache setObject:payment forKey:productID];
+  result([FIAObjectTranslator getMapFromSKPayment:payment]);
+}
+
+- (void)addPayment:(FlutterMethodCall *)call result:(FlutterResult)result {
+  if (![call.arguments isKindOfClass:[NSDictionary class]]) {
+    result([FlutterError errorWithCode:@"storekit_invalide_argument"
```
mklim:
```suggestion
    result([FlutterError errorWithCode:@"storekit_invalid_argument"
```

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+                               message:@"Argument type of addPayment is not a map"
+                               details:call.arguments]);
+    return;
+  }
+  NSDictionary *paymentMap = (NSDictionary *)call.arguments;
+  NSString *productID = [paymentMap objectForKey:@"productID"];
+  SKPayment *payment = [self.paymentsCache objectForKey:productID];
+  // User can use  payment object with mutable = true and add simulatesAskToBuyInSandBox = true to
+  // test the payment flow.
+  if (!payment || [paymentMap[@"mutable"] boolValue] == YES) {
```
mklim: This ends up creating a mutable payment whenever payment isn't specified, not just when `mutable` is explicitly specified as an option. Is that something we want to do?

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+    SKMutablePayment *mutablePayment = [[SKMutablePayment alloc] init];
+    mutablePayment.productIdentifier = productID;
+    NSNumber *quantity = [paymentMap objectForKey:@"quantity"];
+    if (quantity) {
```
mklim: Should we panic and bail here if we don't have a quantity?

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
```diff
@@ -0,0 +1,128 @@
...
+    self.shouldAddStorePayment = shouldAddStorePayment;
+    self.updatedDownloads = updatedDownloads;
+    self.transactionsSetter = [NSMutableDictionary new];
+  }
+  return self;
+}
+
+- (void)addPayment:(SKPayment *)payment {
+  NSString *productID = payment.productIdentifier;
+  if (self.testing) {
```
mklim: I think having this kind of if/then case in production code just for the sake of unit testing someone is generally a bad sign. It's a case for potential bugs down the line if the flag gets flipped in prod, and even when it's working it complicates the code by adding a case for just test logic. Could we instead just send in the test `productID` in the `SkPayment` param itself when we're testing?

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
```diff
@@ -0,0 +1,128 @@
...
+    productID = TestingProductID;
+  }
+  [self.queue addPayment:payment];
+}
+
+- (void)finishTransaction:(SKPaymentTransaction *)transaction {
+  @try {
+    [self.queue finishTransaction:transaction];
+  } @catch (NSException *e) {
```
mklim: What's the benefit to catching and then immediately `raise`ing the exception again here?

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
```diff
@@ -0,0 +1,128 @@
...
+#pragma mark - observing
+// Sent when the transaction array has changed (additions or state changes).  Client should check
+// state of transactions and finish as appropriate.
+- (void)paymentQueue:(SKPaymentQueue *)queue
+    updatedTransactions:(NSArray<SKPaymentTransaction *> *)transactions {
+  for (SKPaymentTransaction *transaction in transactions) {
+    [self.transactionsSetter setObject:transaction forKey:transaction.transactionIdentifier];
+  }
+  // notify dart through callbacks.
+  if (self.transactionsUpdated) {
```
mklim: Should we even allow this to be constructed if we don't have a valid callback back up into dart for when our transactions are updated? Same general question for the other callbacks. I don't know if we really want this to be able to operate if it can't reach back up into the plugin with status updates.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+      mutablePayment.quantity = quantity.integerValue;
+    }
+    NSString *applicationUsername = [paymentMap objectForKey:@"applicationUsername"];
+    mutablePayment.applicationUsername = applicationUsername;
+    if (@available(iOS 8.3, *)) {
+      mutablePayment.simulatesAskToBuyInSandbox =
+          [[paymentMap objectForKey:@"simulatesAskToBuyInSandBox"] boolValue];
+    }
+    [self.paymentQueueHandler addPayment:payment];
```
mklim: nit, subjective: You can git rid of this line and the else case here by always executing this statement at the bottom of the function, outside of the if block entirely.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+  } else {
+    [self.paymentQueueHandler addPayment:payment];
+  }
+}
+
+- (void)finishTransaction:(FlutterMethodCall *)call result:(FlutterResult)result {
+  if (![call.arguments isKindOfClass:[NSString class]]) {
+    result([FlutterError errorWithCode:@"storekit_invalide_argument"
```
mklim:
```suggestion
    result([FlutterError errorWithCode:@"storekit_invalid_argument"
```

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+  if (!transaction) {
+    result([FlutterError errorWithCode:@"storekit_platform_invalid_transaction"
+                               message:@"Invalid transaction ID is used."
+                               details:call.arguments]);
+    return;
+  }
+  @try {
+    // finish transaction will throw exception if the transaction type is purchasing. Notify dart
+    // about this exception.
+    [self.paymentQueueHandler finishTransaction:transaction];
```
mklim: What happens to `result` in the good case here?

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+}
+
+#pragma mark - delegates
+
+- (void)handleTransactionsUpdated:(NSArray<SKPaymentTransaction *> *)transactions {
+  NSMutableArray *maps = [NSMutableArray new];
+  for (SKPaymentTransaction *transcation in transactions) {
+    [maps addObject:[FIAObjectTranslator getMapFromSKPaymentTransaction:transcation]];
+  }
+  [self.callbackChannel invokeMethod:@"updatedTransaction" arguments:maps];
```
mklim: nit: I think it would be a little clearer to name these methods "<callbacktype>Transactions" here and below, since there's multiple transactions in each method.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+}
+
+- (void)handleTransactionsRemoved:(NSArray<SKPaymentTransaction *> *)transactions {
+  NSMutableArray *maps = [NSMutableArray new];
+  for (SKPaymentTransaction *transcation in transactions) {
```
mklim: nit: s/transcation/transaction/g

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+- (void)updatedDownloads:(NSArray<SKDownload *> *)downloads {
+  NSMutableArray *maps = [NSMutableArray new];
+  for (SKDownload *download in downloads) {
+    [maps addObject:[FIAObjectTranslator getMapFromSKDownload:download]];
+  }
+  [self.callbackChannel invokeMethod:@"updatedDownloads" arguments:maps];
+}
+
+- (BOOL)shouldAddStorePayment:(SKPayment *)payment product:(SKProduct *)product {
+  // We alwasy return NO here. And we send the message to dart to process the payment; and we will
```
mklim:
```suggestion
  // We always return NO here. And we send the message to dart to process the payment; and we will
```

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -70,11 +123,150 @@ - (void)handleProductRequestMethodCall:(FlutterMethodCall *)call result:(Flutter
...
+  // have a incerpection method that deciding if the payment should be processed (implemented by the
```
mklim: I think you mean interception here?

```suggestion
  // have a interception method that deciding if the payment should be processed (implemented by the
```

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
cyanglaz: The mutable keyword is indeed confusing. I created this for user to test the payment in sandbox, I have updated the mutable keyword to `usePaymentObject` in a different branch. Do you think changing the keyboard to `usePaymentObject` makes more sense here?

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
cyanglaz: I think we should update the quantity to default to 1 here, ill update.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
mklim: Yeah, I think that makes sense.

I'm still a little confused over the code silently creating an `SkMutable` payment object whenever it can't find a pre-existing payment in the map. Do we want to default to that or would it make sense to error if we can't find an existing payment instead?

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
cyanglaz: It is an unused code that not been removed. I will remove this

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
cyanglaz: There is no benefit here, I should remove this.

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
cyanglaz: Maybe we can add all the call back when user adds the observer in dart. 
Or maybe we block the add payment flow if there is no dart observer been added? What do you think? Let's discuss this offline

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
cyanglaz: it is actually a bug, it should be 
```
    [self.paymentQueueHandler addPayment:mutablePayment];
```

cyanglaz:
I signed it

cyanglaz:
Mistakenly sent out a bunch of review requests due to the file owners setting. Please ignore the requests.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
cyanglaz: Oh you are right. Going to update. I am also going to re arrange the code a bit to make it more readable.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -156,38 +156,64 @@ - (void)createPaymentWithProductID:(FlutterMethodCall *)call result:(FlutterResu
...
+  if (payment) {
+    [self.paymentQueueHandler addPayment:payment];
+    result(nil);
+    return;
+  }
+  // The regular payment flow: when a product is already fetched, we create a payment object with
+  // the product to process the payment.
+  SKProduct *product = [self.productsCache objectForKey:productID];
+  if (product) {
+    payment = [SKPayment paymentWithProduct:product];
```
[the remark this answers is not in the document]
cyanglaz: It is not supported with the regular purchase flow, user will need to use the custom payment object.

PR #1232, opened 2019-02-16
[IAP] Add the Dart API for launchBillingFlow

mklim:


packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -97,6 +97,40 @@ class BillingClient {
         arguments));
   }
 
+  /// Attempt to launch the Play Billing Flow for a given [skuDetails].
+  ///
+  /// [skuDetails] needs to have already been fetched in a [querySkuDetails]
```
cyanglaz:
```suggestion
  /// The [skuDetails] needs to have already been fetched in a [querySkuDetails]
```

packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -97,6 +97,40 @@ class BillingClient {
...
+  /// call. [accountId] is an optional hashed string associated with the user
```
cyanglaz:
```suggestion
  /// call. The [accountId] is an optional hashed string associated with the user
```
Not necessarily wrong grammar but adding 'The' in front would avoid starting the sentence with a lower case letter.
This sentence might be ambiguous: the string is unique to your app or the user is unique to your app.

packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -97,6 +97,40 @@ class BillingClient {
...
+  /// It constructs a
+  /// [`BillingFlowParams`](https://developer.android.com/reference/com/android/billingclient/api/BillingFlowParams)
+  /// instance by [setting the given
+  /// skuDetails](https://developer.android.com/reference/com/android/billingclient/api/BillingFlowParams.Builder.html#setskudetails)
+  /// and [the given
+  /// accountId](https://developer.android.com/reference/com/android/billingclient/api/BillingFlowParams.Builder.html#setAccountId(java.lang.String)).
+  Future<BillingResponse> launchBillingFlow(
+      {@required SkuDetailsWrapper skuDetails, String accountId}) async {
+    final Map<String, dynamic> arguments = <String, dynamic>{
+      'sku': skuDetails.sku,
```
cyanglaz: skuDetails.sku would cause issue if skuDetails is null.

packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -97,6 +97,40 @@ class BillingClient {
...
+      'accountId': accountId,
```
cyanglaz: can accountId be null?

packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -97,6 +97,40 @@ class BillingClient {
...
+  /// user's Google ID for this field.
+  ///
+  /// Calling this attemps to show the Google Play purchase UI. The user is free
+  /// to complete the transaction there.
+  ///
+  /// This method returns a [BillingResponse] representing the initial attempt
+  /// to show the Google Play purchase screen.
+  /// TODO(mklim, flutter/flutter#26326): Expose onPurchasesUpdated() result.
+  ///
+  /// This method calls through to
```
cyanglaz: This method calls through to
```suggestion
  /// This method calls through
```

mklim: I think we do want the "to" here to specify what method it's calling through to.

packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
mklim: Yes, we are okay with passing null down for this param when it's not set. It's optional.

packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
mklim: Yes. I was relying on `@required` to indicate that this should be null, but I guess it's true that on its own `@required` just makes sure that it's specified and not that it's non-null. I'll add an assert for it too.

mklim: Done.

PR #1234, opened 2019-02-18
[firebase_auth] Fix crash on Android during Google sign in

gi097:
Solves a crash where certain activities were not found. See: https://github.com/flutter/flutter/issues/26777

### DIFF
--- a/packages/firebase_auth/android/build.gradle
+++ b/packages/firebase_auth/android/build.gradle
@@ -47,5 +47,6 @@ android {
     }
     dependencies {
         api 'com.google.firebase:firebase-auth:16.0.5'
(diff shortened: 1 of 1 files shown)


kroikie:
If you are using Firebase Auth and Google Sign In then when you add both:

```
firebase_auth
google_sign_in
```

Then the `play-services-auth` should be added to the project from the `google_sign_in` plugin. If someone is using firebase_auth but not using Google Sign In they should not have the Google Sign In dependency added to their application.

gi097:
@kroikie thanks for the heads up! You are right. I assume we should only add this to the `google_sign_in`. I can create a PR for this later.

kroikie:
@gi097 sounds good I'll close this one.

PR #1235, opened 2019-02-18
[webview_flutter] Added the ability to activate the zoom (Android)

Dn-a:
I added the ability to enable zooming on Android WebView.
by default it is set to false.

Pop-up zoom control is disabled, dialog is not responding to touch events.
`webView.getSettings().setDisplayZoomControls(false);`

### DIFF
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -137,13 +137,13 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
       throw new UnsupportedOperationException("JavaScript string cannot be null");
     }
     webView.evaluateJavascript(
-        jsString,
-        new android.webkit.ValueCallback<String>() {
-          @Override
-          public void onReceiveValue(String value) {
-            result.success(value);
-          }
-        });
+            jsString,
+            new android.webkit.ValueCallback<String>() {
+              @Override
+              public void onReceiveValue(String value) {
+                result.success(value);
+              }
+            });
   }
 
   @SuppressWarnings("unchecked")
@@ -174,6 +174,9 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
         case "jsMode":
           updateJsMode((Integer) settings.get(key));
           break;
+        case "zoom":
+          updateZoomMode((boolean) settings.get(key));
+          break;
         default:
           throw new IllegalArgumentException("Unknown WebView setting: " + key);
       }
@@ -193,10 +196,18 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
     }
   }
 
+  private void updateZoomMode(boolean mode) {
+    //webView.getSettings().setLoadWithOverviewMode(true);
+    webView.getSettings().setUseWideViewPort(mode);
+    webView.getSettings().setBuiltInZoomControls(mode);
+    // Pop-up zoom controls disabled. This is a temporary stop because dialog is not responding to touch events.
+    webView.getSettings().setDisplayZoomControls(false);
+  }
+
   private void registerJavaScriptChannelNames(List<String> channelNames) {
     for (String channelName : channelNames) {
       webView.addJavascriptInterface(
-          new JavaScriptChannel(methodChannel, channelName), channelName);
+              new JavaScriptChannel(methodChannel, channelName), channelName);
     }
   }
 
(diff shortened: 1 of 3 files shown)


Dn-a:
> 
> 
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here (e.g. `I signed it!`) and we'll verify it.
> #### What to do if you already signed the CLA
> ##### Individual signers
> 
>     * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> 
> ##### Corporate signers
> 
>     * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google.com/docs/cla/#troubleshoot)).
> 
>     * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
>     * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> 
> Googlers can find more info about SignCLA and this PR by [following this link](go/prinfo/https%3A%2F%2Fgithub.com%2Fflutter%2Fplugins%2Fpull%2F1235).

I signed it!

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -193,6 +196,14 @@ private void updateJsMode(int mode) {
     }
   }
 
+  private void updateZoomMode(boolean mode) {
+    //webView.getSettings().setLoadWithOverviewMode(true);
+    webView.getSettings().setUseWideViewPort(mode);
```
amirh: Why is this tied to the zoom mode?

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -193,6 +196,14 @@ private void updateJsMode(int mode) {
     }
   }
 
+  private void updateZoomMode(boolean mode) {
+    //webView.getSettings().setLoadWithOverviewMode(true);
```
amirh: delete this line

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -130,7 +131,10 @@ class WebView extends StatefulWidget {
   ///
   /// A null value is equivalent to an empty set.
   final Set<JavascriptChannel> javascriptChannels;
-
+  
+  /// Enable/Disable zoom.
```
amirh: Is there a way to do the same on iOS?
If yes we should support that as well, if no the comment should mention that this is a no-op on iOS.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -130,7 +131,10 @@ class WebView extends StatefulWidget {
...
+  final bool zoom;
```
amirh: nit: I'd suggest calling this zoomEnabled (I can imagine a parameter named zoom being used to set a zoom level)

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -256,19 +260,19 @@ class _CreationParams {
 }
 
 class _WebSettings {
-  _WebSettings({
-    this.javascriptMode,
-  });
-
+  _WebSettings({this.javascriptMode, this.zoom});
```
amirh: The way this worked has changed a bit, please rebase and use the widget reference.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
Dn-a: When the value is true and the page contains the viewport meta tag, the value of the width specified in the tag is used. If the page does not contain the tag or does not provide a width, then a wide viewport will be used, so you could always leave it on. do you suggest creating another parameter?

packages/webview_flutter/lib/webview_flutter.dart
Dn-a: at the moment I have no way to test the ios environment. I will modify the comment in on-op on ios

packages/webview_flutter/lib/webview_flutter.dart
Dn-a: ok, of course

Dn-a:
@amirh I know it is not the right place but I could not find a solution to this problem [`flutter_plugin_tools`](https://stackoverflow.com/questions/54755951/problem-when-executing-pub-global-run-flutter-plugin-tools-format-plugins), could you help me?

KingWu:
Any news for this feature? The latest version does not support it

PR #1236, opened 2019-02-19
[webview_flutter]Allow specifying a navigation delegate(Android and Dart).

amirh:
This allows the app to prevent specific navigations(e.g prevent
navigating to specific URLs).

https://github.com/flutter/flutter/issues/25329

iOS implementation in https://github.com/flutter/plugins/pull/1323


packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
```diff
@@ -0,0 +1,59 @@
+package io.flutter.plugins.webviewflutter;
```
Hixie: missing license

Hixie: consider adding a test that checks that all source code files start with a license

packages/webview_flutter/example/lib/main.dart
```diff
@@ -3,11 +3,26 @@
...
 import 'dart:async';
+import 'dart:convert';
 import 'package:flutter/material.dart';
 import 'package:webview_flutter/webview_flutter.dart';
 
 void main() => runApp(MaterialApp(home: WebViewExample()));
 
+const String kNavigationExamplePage = '''
+<!DOCTYPE html><html>
+<body>
```
Hixie: the <html> is optional but the <title> is not. :-)

packages/webview_flutter/example/lib/main.dart
```diff
@@ -35,6 +50,14 @@ class WebViewExample extends StatelessWidget {
           javascriptChannels: <JavascriptChannel>[
             _toasterJavascriptChannel(context),
           ].toSet(),
+          navigationDelegate: (NavigationRequest request) {
+            if (request.url == 'https://www.youtube.com/') {
```
Hixie: startsWith?

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -33,6 +33,39 @@ class JavascriptMessage {
 /// Callback type for handling messages sent from Javascript running in a web view.
 typedef void JavascriptMessageHandler(JavascriptMessage message);
 
+/// Information about a navigation action that is about to be executed.
+class NavigationRequest {
+  NavigationRequest({this.url, this.isMainFrame});
```
Hixie: constructor is missing docs

Hixie: consider adding some sort of benchmark to check how many public members are missing docs in the repo

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -33,6 +33,39 @@ class JavascriptMessage {
...
+
+  /// The URL a navigation is requested to.
```
Hixie: this doc doesn't pass our style guide ("url" and "navigation" and "request" are in the member and class name, so that leaves just "the a is ed to" which isn't meaningful).

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -33,6 +33,39 @@ class JavascriptMessage {
...
+  final String url;
+
+  /// Whether the navigation request originated from the main HTML frame.
+  final bool isMainFrame;
+
+  @override
+  String toString() {
+    return 'NavigationRequest(url: $url, isMainFrame: $isMainFrame)';
```
Hixie: we usually use `$runtimeType` instead of hardcoding the type in the toString, it makes it work better when you subclass or rename the class or copy-paste the code

Hixie:
general design SGTM.

packages/webview_flutter/example/lib/main.dart
amirh: done.

Getting @Hixie  to review my HTML:



😄

packages/webview_flutter/example/lib/main.dart
amirh: done

packages/webview_flutter/lib/webview_flutter.dart
amirh: Made the constructor private, filed: https://github.com/flutter/flutter/issues/28599

packages/webview_flutter/lib/webview_flutter.dart
amirh: Gave it another shot.

packages/webview_flutter/lib/webview_flutter.dart
amirh: done

packages/webview_flutter/example/lib/main.dart
```diff
@@ -3,11 +3,27 @@
...
 import 'dart:async';
+import 'dart:convert';
 import 'package:flutter/material.dart';
 import 'package:webview_flutter/webview_flutter.dart';
 
 void main() => runApp(MaterialApp(home: WebViewExample()));
 
+const String kNavigationExamplePage = '''
+<!DOCTYPE html><html>
+<head><title>Navigation Delegate Example</title></head>thanks
```
Hixie: stray "thanks"

amirh:
@mklim / @bparrishMines can you take a look (Ian reviewed mainly the public API).

packages/webview_flutter/example/android/gradle.properties
```diff
@@ -1 +1,2 @@
 org.gradle.jvmargs=-Xmx1536M
+android.useAndroixX=true
```
mklim:
```suggestion
android.useAndroidX=true
```

amirh: oops, thanks fixed.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
```diff
@@ -0,0 +1,110 @@
...
+    public void success(Object shouldLoad) {
+      Boolean typedShouldLoad = (Boolean) shouldLoad;
+      if (typedShouldLoad) {
+        loadUrl();
+      }
+    }
+
+    @Override
+    public void error(String errorCode, String s1, Object o) {
+      throw new IllegalStateException("navigation ");
```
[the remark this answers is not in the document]
amirh: done

PR #1239, opened 2019-02-19
[google_maps_flutter] Marker APIs are now widget based (Dart Changes)

iskakaushik:
Additional Context: Maps Plugin is in the process of being
moved away from a controller based API to a widget based api.
This is to facilitate easier state management and address a
lot of the common issues.


packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -237,13 +259,25 @@ class _GoogleMapOptions {
...
+  Map<String, dynamic> updatesMap(
+      _GoogleMapOptions newOptions, MarkerUpdates markerUpdates) {
     final Map<String, dynamic> prevOptionsMap = toMap();
+
     return newOptions.toMap()
-      ..removeWhere(
-          (String key, dynamic value) => prevOptionsMap[key] == value);
+      ..remove('markers')
+      ..removeWhere((String key, dynamic value) => prevOptionsMap[key] == value)
+      ..putIfAbsent('markerUpdates', () => markerUpdates._toJson());
```
amirh: Took me a little bit to wrap my ahead around who's responsible for the marker updates.

It used to be that `_GoogleMapOptions` is holds all the options data, this data is passed in the creation parameters 'options' field, `_GoogleMapOptions` knows to compute update deltas which are passed to `_updateMapOptions`.

The `markers` field doesn't follow this same pattern: it is still passed as part of the options in the creation parameters, but _GoogleMapOptions is not responsible for computing updates for it, and `_updateMapOptions` ignores it.

What do you think about keeping a _markers field in _GoogleMapState?

packages/google_maps_flutter/lib/src/controller.dart
```diff
@@ -144,95 +110,76 @@ class GoogleMapController extends ChangeNotifier {
...
-    // https://github.com/flutter/flutter/issues/26431
-    // ignore: strong_mode_implicit_dynamic_method
-    await _channel.invokeMethod('marker#update', <String, dynamic>{
-      'marker': marker._id,
-      'options': changes._toJson(),
-    });
-    marker._options = marker._options.copyWith(changes);
-    notifyListeners();
+/// Handles callbacks for events on [Marker] and [InfoWindow].
+class MarkerController {
```
amirh: Making this part of the public API limits are flexibility with breaking changes here, we better keep this and MarkersController private API if we can.

packages/google_maps_flutter/lib/src/controller.dart
```diff
@@ -144,95 +110,76 @@ class GoogleMapController extends ChangeNotifier {
...
-  ///
-  /// The returned [Future] completes once listeners have been notified.
-  Future<void> removeMarker(Marker marker) async {
+  factory MarkerController.init(Marker marker) {
     assert(marker != null);
-    assert(_markers[marker._id] == marker);
-    await _removeMarker(marker._id);
-    notifyListeners();
+    final String id = marker.markerId.value;
+    // TODO (kaushik) using id in the plugin handle might not be a good idea.
```
amirh: nit: should be `TODO(iskakaushik): ...` (here and elsewhere)

packages/google_maps_flutter/lib/src/controller.dart
amirh: The name MarkerController hints to me that it controls everything related to this marker (e.g icon, position, etc...) this is more of a MarkerEventDelegate

packages/google_maps_flutter/lib/src/controller.dart
```diff
@@ -144,95 +110,76 @@ class GoogleMapController extends ChangeNotifier {
...
+///
+/// Change listeners are notified upon changes to any of the markers.
+class MarkerControllers extends ChangeNotifier {
+  MarkerControllers();
+
+  final Map<MarkerId, MarkerController> _markerControllers =
+      <MarkerId, MarkerController>{};
+
+  void update(MarkerUpdates markerUpdates) {
+    markerUpdates.markerUpdates.forEach((MarkerUpdate markerUpdate) {
```
amirh: I think the code would be simpler if we would just use the map method channel for all markers, that way we could save the management of the controllers here and e.g when we get a `marker#onTap(markerId)` all we need to do is fetch the Marker with id `markerId` from the markers set we maintain anyway and invoke the callback.

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -135,36 +141,97 @@ class _GoogleMapState extends State<GoogleMap> {
...
+      _markerUpdateHandler._onInfoWindowTap(),
+    );
     _controller.complete(controller);
     if (widget.onMapCreated != null) {
       widget.onMapCreated(controller);
     }
   }
 }
 
+class _MarkerUpdateHandler {
```
amirh: I find the role of this class a little hard to grasp, it does the following:
 * Maintains the set of current markers
 * Proxy calls to `MarkerUpdates` for computing and returning the diff for a new markers set.
 * Provides closures for invoking the marker events on a given marker.

I was also a little confused by the name - the initial mental model I had just from seeing the class name was (this is a marker update handler, it probably handles computing the updates and sending them over the platform channel).

The closure generation is particularly a trick I'd rather avoid if it's not necessary as it is harder to follow.

It might be easier to follow if we keep the markers map as a member of _GoogleMapState, then we could avoid the closures as we could have `_onMarkerTap(String markerId)` as a method of _GoogleMapState.
It looks like the update computation logic can be done in a pretty short method which we can fit in `_updateMarkers`.

packages/google_maps_flutter/lib/src/marker_updates.dart
```diff
@@ -0,0 +1,186 @@
...
+
+  @override
+  String toString() {
+    return 'MarkerUpdates{markerUpdates: $markerUpdates}';
+  }
+}
+
+/// [Marker] update event with the changes.
+@immutable
+class MarkerUpdate {
```
amirh: There's some ambiguity with this being called MarkerUpdate while it can have one [add/remove/update] event type.

amirh: Also - if we keep this class we probably don't want to make it part of the public API.

packages/google_maps_flutter/lib/src/marker_updates.dart
```diff
@@ -0,0 +1,186 @@
...
+  factory MarkerUpdate._add(Marker newMarker) {
+    return MarkerUpdate.internal(
+      updateEventType: MarkerUpdateEventType.add,
+      markerId: newMarker.markerId,
+      changes: newMarker,
+      newMarker: newMarker,
+    );
+  }
+
+  /// TODO(iskakaushik): Diff is sufficient, don't need to send the whole update.
```
amirh: TODOs should include a link to a GitHub issue otherwise they are not in our tracking system,.

packages/google_maps_flutter/lib/src/marker_updates.dart
```diff
@@ -0,0 +1,186 @@
...
+    }
+  });
+  return result;
+}
+
+/// [Marker] update events to be applied to the [GoogleMap].
+///
+/// Used in [GoogleMapController] when the map is updated.
+@immutable
+class MarkerUpdates {
```
amirh: Is this essentially a set of MarkerUpdate objects? why do we need a dedicated type for it?

amirh: If we end up keeping this class we probably don't want it to be public API.

packages/google_maps_flutter/lib/src/marker_updates.dart
```diff
@@ -0,0 +1,186 @@
...
+part of google_maps_flutter;
+
+/// Types of [Marker] updates.
+enum MarkerUpdateEventType {
+  add,
+  update,
+  remove,
+}
+
+Map<MarkerId, Marker> _toMap(Set<Marker> markers) {
```
amirh: See suggestion on _createMarkerIdMarkerMap, also this seems to be redundant with the logic there?

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -135,36 +141,97 @@ class _GoogleMapState extends State<GoogleMap> {
...
+  _MarkerUpdateHandler._(this._markers);
+
+  factory _MarkerUpdateHandler.init(Set<Marker> markers) {
+    return _MarkerUpdateHandler._(_createMarkerIdMarkerMap(markers));
+  }
+
+  Map<MarkerId, Marker> _markers;
+
+  dynamic update(Set<Marker> newMarkers) {
```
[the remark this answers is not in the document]
iskakaushik: Will change it to `Map<String, dynamic>`. It returns the change to be sent to the platform channel

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -101,13 +105,15 @@ class _GoogleMapState extends State<GoogleMap> {
...
+  _MarkerUpdateHandler _markerUpdateHandler;
   _GoogleMapOptions _googleMapOptions;
 
   @override
   Widget build(BuildContext context) {
     final Map<String, dynamic> creationParams = <String, dynamic>{
       'initialCameraPosition': widget.initialCameraPosition?._toMap(),
-      'options': _GoogleMapOptions.fromWidget(widget).toMap(),
+      'options': _googleMapOptions.toMap(),
+      'markerUpdates': _markerUpdateHandler.update(widget.markers),
```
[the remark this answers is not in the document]
iskakaushik: I am considering renaming `MarkerUpdate` class to `MarkerChange` and call this `markerChanges`. I will still call it `markers` in creationParams, but elsewhere I will call this `markerChanges` let me know what you think.

packages/google_maps_flutter/lib/src/marker_updates.dart
iskakaushik: I will rename this `MarkerChange`

packages/google_maps_flutter/lib/src/marker_updates.dart
iskakaushik: Will make an issue for this.

packages/google_maps_flutter/lib/src/marker_updates.dart
iskakaushik: This does a few more things than just being a wrapper:

1. This class provides a nice place to have `.from(previous, current)` to compute the diff.
2. Provides methods for serialization of the update chunk that we need to send over to the platform side.
3. Though this is a wrapper around a `Set<MarkerUpdate>`, it succinctly captures a snapshot of all updates done to the markers.

iskakaushik: Will make this private.

packages/google_maps_flutter/lib/src/controller.dart
```diff
@@ -119,120 +102,47 @@ class GoogleMapController extends ChangeNotifier {
...
-      <String, dynamic>{
-        'options': effectiveOptions._toJson(),
-      },
+    await _channel.invokeMethod(
+      'markers#update',
+      markerUpdates._toMap(),
     );
-    final Marker marker = Marker(markerId, effectiveOptions);
-    _markers[markerId] = marker;
     notifyListeners();
```
amirh: Not related to this PR - We should re-evaluate the single-listener model before we hit API stability.

packages/google_maps_flutter/lib/src/marker_updates.dart
amirh: still need to do this

packages/google_maps_flutter/lib/src/marker.dart
```diff
@@ -199,18 +253,51 @@ class MarkerOptions {
...
+  @override
+  String toString() {
+    return 'Marker{markerId: $markerId, alpha: $alpha, anchor: $anchor, '
+        'consumeTapEvents: $consumeTapEvents, draggable: $draggable, flat: $flat, '
+        'icon: $icon, infoWindow: $infoWindow, position: $position, rotation: $rotation, '
+        'visible: $visible, zIndex: $zIndex, onTap: $onTap}';
+  }
+}
+
+Map<MarkerId, Marker> keyByMarkerId(Iterable<Marker> markers) {
```
amirh: Make this private

packages/google_maps_flutter/lib/src/marker_updates.dart
```diff
@@ -0,0 +1,99 @@
...
+    final Set<MarkerId> prevMarkerIds = previousMarkers.keys.toSet();
+    final Set<MarkerId> currentMarkerIds = currentMarkers.keys.toSet();
+
+    Marker idToCurrentMarker(MarkerId id) {
+      return currentMarkers[id];
+    }
+
+    final Set<MarkerId> markerIdsToRemove =
+        prevMarkerIds.difference(currentMarkerIds);
+    final Set<Marker> markersToAdd = currentMarkerIds
```
amirh: optional nit: blank line before `markersToAdd` and `markersToChange`.

packages/google_maps_flutter/lib/src/marker_updates.dart
```diff
@@ -0,0 +1,99 @@
...
+  MarkerUpdates.internal({
+    @required this.markersToAdd,
+    @required this.markerIdsToRemove,
+    @required this.markersToChange,
+  })  : assert(markersToAdd != null),
+        assert(markerIdsToRemove != null),
+        assert(markersToChange != null);
+
+  /// Computes [MarkerUpdates] given previous and current [Marker]s.
+  factory MarkerUpdates.from(Set<Marker> previous, Set<Marker> current) {
```
amirh: This always returns a new instance of MarkerUpdates, so no need to make it a factory constructor.

packages/google_maps_flutter/lib/src/marker.dart
```diff
@@ -199,18 +253,51 @@ class MarkerOptions {
...
+  if (markers == null) {
+    return <MarkerId, Marker>{};
+  }
+  return Map<MarkerId, Marker>.fromEntries(markers.map(
+      (Marker marker) => MapEntry<MarkerId, Marker>(marker.markerId, marker)));
+}
+
+dynamic _serializeMarkerSet(Set<Marker> markers) {
```
amirh: nit: if this is only used in markers_updates.dart lets keep it in that file.

Also - avoid dynamic when possible.

packages/google_maps_flutter/lib/src/marker.dart
```diff
@@ -55,33 +33,135 @@ class InfoWindowText {
...
+  @override
+  int get hashCode => value.hashCode;
+
+  @override
+  String toString() {
+    return 'MarkerId{value: $value}';
+  }
+}
+
+/// An icon placed at a particular geographical location on the map's surface.
```
amirh: blank line after the summary.

amirh: Is a marker "an icon"?
Maybe just say in the summary "Marks a geographical location on the map"?

[I see that you just moved this, but might be a good opportunity anyway]

packages/google_maps_flutter/lib/src/marker.dart
```diff
@@ -55,33 +33,135 @@ class InfoWindowText {
...
+  /// * is stationary; [draggable] is false
+  /// * is drawn against the screen, not the map; [flat] is false
+  /// * has a default icon; [icon] is `BitmapDescriptor.defaultMarker`
+  /// * anchors the info window at top center; [infoWindowAnchor] is (0.5, 0.0)
+  /// * has no info window text; [infoWindowText] is `InfoWindowText.noText`
+  /// * is positioned at 0, 0; [position] is `LatLng(0.0, 0.0)`
+  /// * has an axis-aligned icon; [rotation] is 0.0
+  /// * is visible; [visible] is true
+  /// * is placed at the base of the drawing order; [zIndex] is 0.0
+  const Marker({
```
amirh: this was previously annotated \@visiblefortesting

packages/google_maps_flutter/lib/src/marker.dart
```diff
@@ -55,33 +33,135 @@ class InfoWindowText {
...
+    this.icon = BitmapDescriptor.defaultMarker,
+    this.infoWindow = InfoWindow.noText,
+    this.position = const LatLng(0.0, 0.0),
+    this.rotation = 0.0,
+    this.visible = true,
+    this.zIndex = 0.0,
+    this.onTap,
   }) : assert(alpha == null || (0.0 <= alpha && alpha <= 1.0));
 
+  /// Used to uniquely identify the [Marker].
```
amirh: nit: Uniquely identifies the marker.

packages/google_maps_flutter/lib/src/marker.dart
```diff
@@ -109,16 +189,8 @@ class MarkerOptions {
...
-  /// displayed.
-  ///
-  /// The image point is specified in normalized coordinates: An anchor of
-  /// (0.0, 0.0) means the top left corner of the image. An anchor
-  /// of (1.0, 1.0) means the bottom right corner of the image.
-  final Offset infoWindowAnchor;
-
-  /// Text content for the info window.
-  final InfoWindowText infoWindowText;
+  /// Window displayed onTap for a marker.
```
amirh:
```
/// A Google Maps InfoWindow.
///
/// The window is displayed when the marker is tapped.
```

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -135,34 +141,62 @@ class _GoogleMapState extends State<GoogleMap> {
...
       widget.onMapCreated(controller);
     }
   }
+
+  void onMarkerTap(String markerIdParam) {
+    if (markerIdParam == null) {
+      return;
+    }
+    final MarkerId markerId = MarkerId(markerIdParam);
+    _markers[markerId]?.onTap();
```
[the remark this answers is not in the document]
iskakaushik: this cannot happen, got rid of the `?`

packages/google_maps_flutter/lib/src/marker.dart
iskakaushik: This is now the preferred way to construct markers. Now that we do not have the `createWithDefaultSettings` factory method

amirh: ohh right

packages/google_maps_flutter/lib/src/marker.dart
iskakaushik: This is also used in `google_map.dart`

packages/google_maps_flutter/example/lib/place_marker.dart
```diff
@@ -25,142 +25,188 @@ class PlaceMarkerBody extends StatefulWidget {
...
+
+    if (markerCount == 12) {
+      return;
+    }
+
+    final String markerIdVal = 'marker_id_$_markerIdCounter';
+    _markerIdCounter++;
+    final MarkerId markerId = MarkerId(markerIdVal);
+
+    void _onMarkerTapped() {
```
amirh: I think it'll simplify things if we avoid this closure.
We can have a member `_onMarkerTapped(MarkerId markerId);` and below do:

```dart
    final Marker marker = Marker(
      // ...
      onTap: () { _onMarkerTapped(markerId); },
    );
```

This way the logic one sees in the _add method is purely about adding a marker.

packages/google_maps_flutter/example/lib/place_marker.dart
```diff
@@ -25,142 +25,188 @@ class PlaceMarkerBody extends StatefulWidget {
...
   void _changePosition() {
-    final LatLng current = _selectedMarker.options.position;
-    final Offset offset = Offset(
-      center.latitude - current.latitude,
-      center.longitude - current.longitude,
-    );
-    _updateSelectedMarker(
-      MarkerOptions(
-        position: LatLng(
+    _updateSelectedMarker((Marker m) {
```
amirh: I feel like `_updateSelectedMarker` is reversing the "reactive" style that we are introducing in this PR. Since this is sample code I'd err on the side of adding a little more duplication but being more clear on the nature of the API(here and in the other "mutation" methods below).

Something like(pseudo):
```dart
Marker selectedMarker = markers[_selectedMarker];
setState(() {
  markers[_selectedMarker] = selectedMaker.copyWith(position: /* .. */);
});
```

packages/google_maps_flutter/example/lib/scrolling_map.dart
```diff
@@ -76,12 +77,31 @@ class ScrollingMapBody extends StatelessWidget {
...
                       initialCameraPosition: CameraPosition(
                         target: center,
                         zoom: 11.0,
                       ),
+                      markers:
+                          // TODO(iskakaushik): Remove this when collection literals makes it to stable.
+                          // ignore: prefer_collection_literals
+                          Set<Marker>.of(
+                        <Marker>[
+                          Marker(
```
amirh: 🎉

packages/google_maps_flutter/test/marker_updates_test.dart
```diff
@@ -0,0 +1,177 @@
...
+    await tester.pumpWidget(_mapWithMarkers(_toSet(m1: m1)));
+    await tester.pumpWidget(_mapWithMarkers(_toSet(m1: m1, m2: m2)));
+
+    final FakePlatformGoogleMap platformGoogleMap =
+        fakePlatformViewsController.lastCreatedView;
+    expect(platformGoogleMap.markersToAdd.length, 1);
+
+    final Marker addedMarker = platformGoogleMap.markersToAdd.first;
+    expect(addedMarker, equals(m2));
+    expect(platformGoogleMap.markerIdsToRemove.isEmpty, true);
```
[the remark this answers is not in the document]
iskakaushik: @amirh I think this is nicer but for each line where i use this shorthand, i need to add the `TODO use collection literal deprecation` stuff, how about I change these after that makes it to stable, I will add a github issue for this.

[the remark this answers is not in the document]
iskakaushik: https://github.com/flutter/flutter/issues/28311

packages/google_maps_flutter/test/marker_updates_test.dart
```diff
@@ -0,0 +1,177 @@
...
+    expect(update.infoWindow.snippet, 'changed');
+  });
+
+  testWidgets("Multi Update", (WidgetTester tester) async {
+    Marker m1 = Marker(markerId: MarkerId("marker_1"));
+    Marker m2 = Marker(markerId: MarkerId("marker_2"));
+    final Set<Marker> prev = _toSet(m1: m1, m2: m2);
+    m1 = Marker(markerId: MarkerId("marker_1"), visible: false);
+    m2 = Marker(markerId: MarkerId("marker_2"), draggable: true);
+    final Set<Marker> cur = _toSet(m1: m1, m2: m2);
```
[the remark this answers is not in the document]
iskakaushik: Filing a github issue to track this: https://github.com/flutter/flutter/issues/28311

packages/google_maps_flutter/test/marker_updates_test.dart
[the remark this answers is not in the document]
amirh: makes sense

packages/google_maps_flutter/lib/src/marker_updates.dart
```diff
@@ -0,0 +1,98 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+part of google_maps_flutter;
+
+/// [Marker] update events to be applied to the [GoogleMap].
+///
+/// Used in [GoogleMapController] when the map is updated.
+class _MarkerUpdates {
+  _MarkerUpdates._({
```
amirh: Is this used?

packages/google_maps_flutter/example/lib/place_marker.dart
```diff
@@ -25,142 +25,191 @@ class PlaceMarkerBody extends StatefulWidget {
...
         ),
-      ),
-    );
+      );
+    });
   }
 
   void _changeAnchor() {
-    final Offset currentAnchor = _selectedMarker.options.anchor;
+    final Marker m = markers[selectedMarker];
```
amirh: s/m/marker/

PR #1240, opened 2019-02-19
[google_maps_flutter] Marker APIs are now widget based (iOS Changes)

iskakaushik:


packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -61,3 +70,172 @@ - (void)setZIndex:(int)zIndex {
...
+  if (self) {
+    _methodChannel = methodChannel;
+    _mapView = mapView;
+    _markerIdToController = [NSMutableDictionary dictionaryWithCapacity:1];
+  }
+  return self;
+}
+- (void)addMarkers:(id)markersToAdd {
+  NSArray<id>* markers = markersToAdd;
+  if (!markers) {
```
amirh: A safer check would be:
`if (![markersToAdd isKindOfClass:[NSArray class]])`, this also covers the cases that it's NSNull (if it was explicitly set to null), or that it is a wrong type.

packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -61,3 +70,172 @@ - (void)setZIndex:(int)zIndex {
...
+- (instancetype)init:(FlutterMethodChannel*)methodChannel mapView:(GMSMapView*)mapView {
+  self = [super init];
+  if (self) {
+    _methodChannel = methodChannel;
+    _mapView = mapView;
+    _markerIdToController = [NSMutableDictionary dictionaryWithCapacity:1];
+  }
+  return self;
+}
+- (void)addMarkers:(id)markersToAdd {
```
cyanglaz: The type of the parameter can be NSArray if you only want to work with NSArray, use id only when the type of the object is possible to be more than 1. 
```suggestion
- (void)addMarkers:(NSArray *)markersToAdd {
```
Same for other similar places in this PR that use id as type.

packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -3,23 +3,32 @@
...
-static uint64_t _nextMarkerId = 0;
+#import "JsonConversions.h"
 
 @implementation FLTGoogleMapMarkerController {
   GMSMarker* _marker;
   GMSMapView* _mapView;
+  BOOL _consumeTapEvents;
 }
-- (instancetype)initWithPosition:(CLLocationCoordinate2D)position mapView:(GMSMapView*)mapView {
+- (instancetype)initWithPositionAndId:(CLLocationCoordinate2D)position
```
cyanglaz: Objc usually only have the first parameter name in the method body
```suggestion
- (instancetype)initWithPosition:(CLLocationCoordinate2D)position andMarkerId:....
```

packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -61,3 +70,172 @@ - (void)setZIndex:(int)zIndex {
...
+  }
+  id draggable = data[@"draggable"];
+  if (draggable) {
+    [sink setDraggable:toBool(draggable)];
+  }
+  id icon = data[@"icon"];
+  if (icon) {
+    NSArray* iconData = icon;
+    UIImage* image;
+    if ([iconData[0] isEqualToString:@"defaultMarker"]) {
```
cyanglaz: it would crash if iconData is size 0. If you only want to escape this if block, then maybe use .firstObject of NSArray which returns nil when the array is nil or empty; and makes the if statement return NO(false).
```suggestion
    if ([iconData.firstObject isEqualToString:@"defaultMarker"]) {
```

packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -61,3 +70,172 @@ - (void)setZIndex:(int)zIndex {
...
+      CGFloat hue = (iconData.count == 1) ? 0.0f : toDouble(iconData[1]);
```
cyanglaz: Are we sure iconData.count > 1?

packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -61,3 +70,172 @@ - (void)setZIndex:(int)zIndex {
...
+      image = [GMSMarker markerImageWithColor:[UIColor colorWithHue:hue / 360.0
+                                                         saturation:1.0
+                                                         brightness:0.7
+                                                              alpha:1.0]];
+    } else if ([iconData[0] isEqualToString:@"fromAsset"]) {
+      if (iconData.count == 2) {
+        image = [UIImage imageNamed:[registrar lookupKeyForAsset:iconData[1]]];
+      } else {
```
cyanglaz: Does this `else` covers only iconData.count > 2?

packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -61,3 +70,172 @@ - (void)setZIndex:(int)zIndex {
...
+    interpretMarkerOptions(marker, controller, _registrar);
+  }
+}
+- (void)removeMarkerIds:(id)markerIdsToRemove {
+  NSArray<id>* markerIds = markerIdsToRemove;
+  if (!markerIds) {
+    return;
+  }
+  for (id markerId in markerIds) {
+    if (!markerId) {
```
cyanglaz: Can markerId be nil? The check seems not necessary.

cyanglaz:
download
Offline Discussion: We will have another PR for error handling if necessary.

amirh:
LGTM

wterrill:
If the markers are now widget based, does this mean that they can be accessed by the animation features of flutter? (i.e. AnimatedContainer, AnimatedOpacity)

JayPerfetto:
+ 1 for how to make these fade in and out when mounted - I am not having luck using any kind of animations

PR #1242, opened 2019-02-20
Bugfix documents: webview_flutter JavascriptChannel, Example is not updated.

hirohisa:
I copy `JavascriptChannel` from the document, but it has an error.
I hope to update the document.


iskakaushik:
@hirohisa if you rebase your PR it should be able to pass the tests.

hirohisa:
@iskakaushik thank you for your reply.
I pushed.

PR #1246, opened 2019-02-21
Collection literals ignores in webview flutter plugin

iskakaushik:


iskakaushik:
@mklim agreed, this feels like a band-aid fix where we should really be testing against both stable and master.

PR #1249, opened 2019-02-21
[in_app_purchase] payment queue dart ios

cyanglaz:
Dart side for add payment in StoreKit.

Fixed some bug in the Objective-C code.

Removed add payment with only product identifier in Objective-C since it seems unnecessary. 

Part of https://github.com/flutter/flutter/issues/26328


packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
```diff
@@ -57,7 +58,9 @@ - (void)finishTransaction:(SKPaymentTransaction *)transaction {
 - (void)paymentQueue:(SKPaymentQueue *)queue
     updatedTransactions:(NSArray<SKPaymentTransaction *> *)transactions {
   for (SKPaymentTransaction *transaction in transactions) {
-    [self.transactionsSetter setObject:transaction forKey:transaction.transactionIdentifier];
+    if (transaction.transactionIdentifier) {
```
mklim: Would this ever be null? Should we handle it like an error if/when it is?

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -222,9 +169,15 @@ - (void)finishTransaction:(FlutterMethodCall *)call result:(FlutterResult)result
...
-    result([FlutterError errorWithCode:@"storekit_platform_invalid_transaction"
-                               message:@"Invalid transaction ID is used."
-                               details:call.arguments]);
+    result([FlutterError
+        errorWithCode:@"storekit_platform_invalid_transaction"
+              message:[NSString
+                          stringWithFormat:@"The transaction with transactionIdentifer:%@ does not "
+                                           @"exsit. Note that if the transactionState is "
+                                           @"purchasing, the transactionIdentifier will be "
+                                           @"nil(null). And you should not finish this transaction",
```
mklim: > And you should not finish this transaction

Can you go into more detail about this?

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
   static Future<bool> canMakePayments() async =>
       await channel.invokeMethod('-[SKPaymentQueue canMakePayments:]');
+
+  /// Sets a transaction observer to listen to all the transaction events of the payment queue.
+  ///
+  /// You have to have a transaction observer to be added to the payment queue to handle the
+  /// payment follow.
+  /// You must set the observer right when App launches to avoid missing callback when your user
+  /// started a purchase flow from the App Store.
+  void setTransactionObserver(SKTransactionObserverWrapper observer) {
```
mklim: Is there any way we could take this as a param at construction time instead? It would force users to set this as soon as this object is instantiated.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
+
+  /// Adds a payment object to the payment queue.
+  ///
+  /// Prior to this call, you have to make sure At least one [SKTransactionObserverWrapper] should be added to the payment queue
+  /// using [addTransactionObserver]. You also have to make sure the [SKProductWrapper] of the payment has been fetched using [SKRequestMaker.startProductRequest].
+  /// Each payment will generate a [SKPaymentTransactionWrapper]. After a payment is being added to the payment queue, the [SKTransactionObserverWrapper] is responsible to handle
+  /// the transaction that is generated by the payment.
+  /// The [productIdentifier] must match one of the product that is returned in [SKRequestMaker.startProductRequest].
+  /// to test the payment in the [sandbox](https://developer.apple.com/apple-pay/sandbox-testing/).
+  Future<void> addPayment(SKPaymentWrapper payment) async {
```
mklim: We should link to the official dev docs on the related methods for all of these too.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
+  /// started a purchase flow from the App Store.
+  void setTransactionObserver(SKTransactionObserverWrapper observer) {
+    _observer = observer;
+  }
+
+  /// Adds a payment object to the payment queue.
+  ///
+  /// Prior to this call, you have to make sure At least one [SKTransactionObserverWrapper] should be added to the payment queue
+  /// using [addTransactionObserver]. You also have to make sure the [SKProductWrapper] of the payment has been fetched using [SKRequestMaker.startProductRequest].
+  /// Each payment will generate a [SKPaymentTransactionWrapper]. After a payment is being added to the payment queue, the [SKTransactionObserverWrapper] is responsible to handle
```
mklim:
```suggestion
  /// Each payment will generate a [SKPaymentTransactionWrapper]. After a payment is added to the payment queue, the [SKTransactionObserverWrapper] is responsible for handling
```

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
+  /// payment follow.
+  /// You must set the observer right when App launches to avoid missing callback when your user
+  /// started a purchase flow from the App Store.
+  void setTransactionObserver(SKTransactionObserverWrapper observer) {
+    _observer = observer;
+  }
+
+  /// Adds a payment object to the payment queue.
+  ///
+  /// Prior to this call, you have to make sure At least one [SKTransactionObserverWrapper] should be added to the payment queue
```
mklim:
```suggestion
  /// Prior to this call, you have to make sure at least one [SKTransactionObserverWrapper] is added to the payment queue
```

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
+  /// using [addTransactionObserver]. You also have to make sure the [SKProductWrapper] of the payment has been fetched using [SKRequestMaker.startProductRequest].
+  /// Each payment will generate a [SKPaymentTransactionWrapper]. After a payment is being added to the payment queue, the [SKTransactionObserverWrapper] is responsible to handle
+  /// the transaction that is generated by the payment.
+  /// The [productIdentifier] must match one of the product that is returned in [SKRequestMaker.startProductRequest].
```
mklim:
```suggestion
  /// The [productIdentifier] must match one of the products that are returned in [SKRequestMaker.startProductRequest].
```

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
+    assert(_observer != null,
+        '[in_app_purchase]: Trying to add a payment without an observer. Observer must be set using `setTransactionObserver` before adding a payment `setTransactionObserver`. It is mandatory to set the observer at the moment when app launches.');
+    Map requestMap = payment.toMap();
+    await channel.invokeMethod(
+      '-[InAppPurchasePlugin addPayment:result:]',
+      requestMap,
+    );
+  }
+
+  /// Finishes a transaction, remove it from the queue.
```
mklim:
```suggestion
  /// Finishes a transaction and removes it from the queue.
```

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
+  ///
+  /// This method should be called from a observer callback when receiving notification from the payment queue. You should only
+  /// call this method after the transaction is successfully processed and the functionality purchased by the user is unlocked.
+  /// Itt will throw a Platform exception if the [SKPaymentTransactionWrapper.transactionState] is [SKPaymentTransactionStateWrapper.purchasing].
```
mklim:
```suggestion
  /// It will throw a Platform exception if the [SKPaymentTransactionWrapper.transactionState] is [SKPaymentTransactionStateWrapper.purchasing].
```

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
+      SKPaymentTransactionWrapper transaction) async {
+    await channel.invokeMethod(
+        '-[InAppPurchasePlugin finishTransaction:result:]',
+        transaction.transactionIdentifier);
+  }
+
+  // Triage a method channel call from the platform and triggers the correct observer method.
+  Future<dynamic> _handleObserverCallbacks(MethodCall call) {
+    assert(_observer != null,
+        'in_app_purchase]: (Fatal)The observer has not been set but we received a purchase transaction notification. Please ensure the observer has been set using `setTransactionObserver`. Make sure the observer is added right at the App Launch.');
```
mklim:
```suggestion
        '[in_app_purchase]: (Fatal)The observer has not been set but we received a purchase transaction notification. Please ensure the observer has been set using `setTransactionObserver`. Make sure the observer is added right at the App Launch.');
```

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
+    switch (call.method) {
+      case 'updatedTransactions':
+        {
+          final List<SKPaymentTransactionWrapper> transactions =
+              _getTransactionList(call.arguments);
+          return Future<void>(() {
+            _observer.updatedTransactions(transactions: transactions);
+          });
```
mklim: I think we want `break`s after each case?

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -14,9 +16,174 @@ part 'sk_payment_queue_wrapper.g.dart';
...
+        .map<SKDownloadWrapper>(
+            (dynamic map) => SKDownloadWrapper.fromJson(map))
+        .toList();
+    return downloads;
+  }
+}
+
+/// This class is Dart wrapper around [SKTransactionObserver](https://developer.apple.com/documentation/storekit/skpaymenttransactionobserver?language=objc).
+///
+/// Must be subclassed and
```
mklim: nit: doc missing.

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
cyanglaz: Apple handles the transaction identifier strangely, the identifier is null when the state the purchasing. I should probably put a comment on it.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
cyanglaz: Sure, by apple's document finish transaction with transactionIdentifier of purchasing will throw exception; I will put something similar here.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
cyanglaz: Since the payment queue is a singleton, there is no actually a construction time. User can technically call SKPaymentQueue() everywhere to refer the defaultPaymentQueue. So I didn't want user to have to put the observer each time when they refer to the singleton.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
cyanglaz: there should be a return already for each case.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
cyanglaz: OMG lol! Sorry.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -171,12 +171,14 @@ - (void)finishTransaction:(FlutterMethodCall *)call result:(FlutterResult)result
...
-                          stringWithFormat:@"The transaction with transactionIdentifer:%@ does not "
-                                           @"exsit. Note that if the transactionState is "
-                                           @"purchasing, the transactionIdentifier will be "
-                                           @"nil(null). And you should not finish this transaction",
-                                           transaction.transactionIdentifier]
+              message:[NSString stringWithFormat:
+                                    @"The transaction with transactionIdentifer:%@ does not "
+                                    @"exsit. Note that if the transactionState is "
+                                    @"purchasing, the transactionIdentifier will be "
+                                    @"nil(null). And you should not finish this transaction. "
```
mklim: Nit: Sorry, on second thought I think everything after the sentence about transactionIdentifier being null isn't needed here. It's helpful to know but it's not necessarily relevant, and we don't want to clog up the error logs with extraneous info.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
mklim: Right. Hmm. I think it may be worth converting this from a singleton into a constructed class just to enforce that this is configured ASAP. What do you think? Could be done in a followup PR too.

cyanglaz: Let's do more offline discussion on this and will create another PR if we reach a different approach.

PR #1251, opened 2019-02-21
Update android_alarm_manager with instructions on setting WAKE_LOCK permissions

bkonyi:
The WAKE_LOCK permission is required by the [JobIntentService](https://developer.android.com/reference/android/support/v4/app/JobIntentService) on pre-O versions. Without this permission, a `java.lang.SecurityException` is thrown. 

Fixes documentation issue which arose from flutter/flutter#27921

### DIFF
--- a/packages/android_alarm_manager/pubspec.yaml
+++ b/packages/android_alarm_manager/pubspec.yaml
@@ -1,7 +1,7 @@
 name: android_alarm_manager
 description: Flutter plugin for accessing the Android AlarmManager service, and
   running Dart code in the background when alarms fire.
-version: 0.4.1+2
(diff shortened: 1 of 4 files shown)


packages/android_alarm_manager/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: android_alarm_manager
 description: Flutter plugin for accessing the Android AlarmManager service, and
   running Dart code in the background when alarms fire.
-version: 0.4.1+2
+version: 0.4.1+3
```
amirh: Say something in the changelog for this version.

Hixie:
Thanks!

PR #1252, opened 2019-02-21
[In_app_purchase]SKProduct related fixes

cyanglaz:
Fixes the price property in SKProduct and SKProductDiscount losing precision after receiving from Objective-c

Fixes the example app to display error as priority.

### DIFF
--- a/packages/in_app_purchase/lib/src/billing_client_wrappers/sku_details_wrapper.g.dart
+++ b/packages/in_app_purchase/lib/src/billing_client_wrappers/sku_details_wrapper.g.dart
@@ -20,15 +20,46 @@ SkuDetailsWrapper _$SkuDetailsWrapperFromJson(Map json) {
       sku: json['sku'] as String,
       subscriptionPeriod: json['subscriptionPeriod'] as String,
       title: json['title'] as String,
-      type: const SkuTypeConverter().fromJson(json['type'] as String),
+      type: _$enumDecode(_$SkuTypeEnumMap, json['type']),
       isRewarded: json['isRewarded'] as bool);
 }
 
+T _$enumDecode<T>(Map<T, dynamic> enumValues, dynamic source) {
+  if (source == null) {
+    throw ArgumentError('A value must be provided. Supported values: '
+        '${enumValues.values.join(', ')}');
+  }
+  return enumValues.entries
+      .singleWhere((e) => e.value == source,
+          orElse: () => throw ArgumentError(
+              '`$source` is not one of the supported values: '
+              '${enumValues.values.join(', ')}'))
+      .key;
+}
+
+const _$SkuTypeEnumMap = <SkuType, dynamic>{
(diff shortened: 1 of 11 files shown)


packages/in_app_purchase/lib/src/billing_client_wrappers/sku_details_wrapper.g.dart
```diff
@@ -20,15 +20,46 @@ SkuDetailsWrapper _$SkuDetailsWrapperFromJson(Map json) {
       sku: json['sku'] as String,
       subscriptionPeriod: json['subscriptionPeriod'] as String,
       title: json['title'] as String,
-      type: const SkuTypeConverter().fromJson(json['type'] as String),
+      type: _$enumDecode(_$SkuTypeEnumMap, json['type']),
```
mklim: This ended up broken again somehow.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -25,7 +25,7 @@ class SkProductResponseWrapper {
   /// This method should only be used with `map` values returned by [SKRequestMaker.startProductRequest].
   /// The `map` parameter must not be null.
   factory SkProductResponseWrapper.fromJson(Map map) {
-    assert(map != null);
+    assert(map != null, 'map must not be null.');
```
mklim: nit: Capitalize "M" here and below.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
```diff
@@ -146,6 +146,8 @@ class SKProductDiscountWrapper {
   /// The [subscriptionPeriod] of the discount is independent of the product's [subscriptionPeriod],
   /// and their units and duration do not have to be matched.
   final SKProductSubscriptionPeriodWrapper subscriptionPeriod;
+
+  static _priceFromJson(double price) => num.parse(price.toStringAsFixed(2));
```
mklim: Are we sure we always want to round the price to two places? Just want to make sure that we're not losing precision here.

packages/in_app_purchase/lib/src/billing_client_wrappers/sku_details_wrapper.g.dart
cyanglaz: Going to revert this file

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_product_wrapper.dart
cyanglaz: Ok, I went to check the app store connect and seems like some currencies do have 3-place decimals. Will need to think about how to handle this in a different way.

PR #1254, opened 2019-02-22
[video_player]Bump video_player version to 0.10.0+2

azeezurr:
New release to add additional documentation (see #1214).


cyanglaz:
The video player seems to be in 0.10.0+6 now so I don't think this PR is needed anymore, closing.

PR #1258, opened 2019-02-22
[cloud_firestore]Temporary fix/specify firestore 5.1.0

vmwsree:
vv


aytunch:
@vmwsree Vivek, what happened whit this commit? This problem is still out there. I can not use cloud functions plugin because of pod dependency issues. Do you have an updated fix?

collinjackson:
We're using `~> 6.0` now.
```
 s.dependency 'Firebase/Firestore', '~> 6.0'
```

PR #1259, opened 2019-02-23
[in_app_purchase] Java API for querying purchases

mklim:
Actually responds in the onPurchasesUpdated() handler automatically
called when purchase updates are received.

Exposes BillingClient#queryPurchases, which is meant to be used to get
recently purchased items from the on disk Play cache.

Also exposes BillingClient#queryPurchaseHistoryAsync, which returns the
most recent purchase of the user for each SKU of the given type.

This patch also contains some updates to the Gradle script so that the
example app version is set independently of the Flutter SDK version.

### DIFF
--- a/packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
+++ b/packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
@@ -4,6 +4,8 @@
 
 package io.flutter.plugins.inapppurchase;
 
+import static io.flutter.plugins.inapppurchase.Translator.fromPurchasesList;
+import static io.flutter.plugins.inapppurchase.Translator.fromPurchasesResult;
 import static io.flutter.plugins.inapppurchase.Translator.fromSkuDetailsList;
 
 import android.app.Activity;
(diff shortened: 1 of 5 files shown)


packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
```diff
@@ -181,7 +195,38 @@ private void launchBillingFlow(String sku, @Nullable String accountId, Result re
...
   }
 
-  private void updateCachedSkus(List<SkuDetails> skuDetailsList) {
+  private void queryPurchases(String skuType, Result result) {
+    if (billingClientError(result)) {
+      return;
+    }
+
+    // Like in our connect call, consider the billing client responding a "success" here regardless of status code.
+    result.success(fromPurchasesResult(billingClient.queryPurchases(skuType)));
```
cyanglaz: Just want to make sure the response object that returned by `billingClient.queryPurchases(skuType)` should never be null, correct?

packages/in_app_purchase/example/android/app/build.gradle
```diff
@@ -27,6 +27,8 @@ project.ext {
     KEYSTORE_STORE_PASSWORD = keystoreProperties['storePassword']
     KEYSTORE_KEY_ALIAS = keystoreProperties['keyAlias']
     KEYSTORE_KEY_PASSWORD = keystoreProperties['keyPassword']
+    VERSION_CODE = 1
```
cyanglaz: It looks ok to me, but I don't have enough understanding of gradle to review gradle in a high degree of confidence. If you are not confident of this part, please find someone else to take a look.

packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
mklim: Yes, this should always be non-null.

PR #1263, opened 2019-02-23
Add SocketTimeOut exception for CloudFunctions

radvansky-tomas:


Hixie:
I'm going through some old PRs to see what we can do with them. First, sorry for dropping this on the floor, we do appreciate your contributions though apparently we don't do a good job of showing it.

For this PR specifically, I'm going to close it because it is lacking tests and all our PRs need to have tests.

PR #1266, opened 2019-02-23
[image_picker] update Uri Authority for GooglePhotos

pepegich:
google photo version 4.10.0.233345822 on Android 9 return Uri with authority "com.google.android.apps.photos.contentprovider". for backward compatibility I changed equals to startsWith.


pepegich:
@cyanglaz hm, I install first version of google photo(1.0.0.94391081) and run on nexus 5 with android 4.4.4. 
authority of image uri equals "com.google.android.apps.photos.contentprovider". seem it is just mistake @roughike 
all version of app have same Uri.authority

PR #1267, opened 2019-02-23
Fix cloud firestore platform error 0

tomalabaster:
The changes in this [PR](https://github.com/flutter/plugins/pull/1226/files) has caused `cloud_firestore` to throw PlatformExceptions incorrectly. The `getFlutterError` method never returns `null` as if the `error` argument is `null`, the string formatting puts 0 in the `errorWithCode` parameter causing a `FlutterError` to always return.


cyanglaz:
Thank you for providing the fix. The issue has been fixed with https://github.com/flutter/plugins/pull/1274.

PR #1269, opened 2019-02-23
[image_picker] set real extension instead always jpg

pepegich:


packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java
```diff
@@ -133,9 +134,11 @@ private static String getPathFromRemoteUri(final Context context, final Uri uri)
     InputStream inputStream = null;
     OutputStream outputStream = null;
     boolean success = false;
+    String extension = getImageExtension(context, uri);
+    String extensionWithDot = TextUtils.isEmpty(extension) ? "" : "." + extension;
     try {
       inputStream = context.getContentResolver().openInputStream(uri);
-      file = File.createTempFile("image_picker", "jpg", context.getCacheDir());
+      file = File.createTempFile("image_picker", extensionWithDot, context.getCacheDir());
```
cyanglaz: what happens to the file path if extensionWithDot is ""

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java
```diff
@@ -159,6 +162,30 @@ private static String getPathFromRemoteUri(final Context context, final Uri uri)
...
+              new String[]{MediaStore.MediaColumns.MIME_TYPE},
+              null, null, null);
+
+      if (cursor != null && cursor.moveToNext()) {
+        String mimeType = cursor.getString(0);
+
+        extension = MimeTypeMap.getSingleton().getExtensionFromMimeType(mimeType);
+      }
+    } catch (Exception ignored) {
+
```
cyanglaz: ignoring exception seems odd. Do we not have to resolve this if there is an exception?

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java
pepegich: nothing, because suffix simply concatenating with prefix. suffix may be null, in which case the suffix .tmp will be used

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java
pepegich: if exception was thrown, final file name will be without extension.

cyanglaz:
@pepegich any updates after the review from @mklim

pepegich:
@mklim please review new updates

packages/image_picker/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.6.0+2
+
+* Android: images have real extension instead .jpg
```
mklim: nit:

```suggestion
* Android: images are saved with their real extension instead of always using `.jpg`.
```

PR #1273, opened 2019-02-24
[google_sign_in]Added GoogleHttpClient to google_sign_in

rishab2113:
Adding support for GoogleHttpClient to the google_sign_in package in Flutter

### DIFF
--- a/packages/google_sign_in/lib/google_sign_in.dart
+++ b/packages/google_sign_in/lib/google_sign_in.dart
@@ -11,8 +11,8 @@ import 'package:meta/meta.dart' show visibleForTesting;
 import 'src/common.dart';
 import 'src/google_http_client.dart';
 
-export 'src/google_http_client.dart';
 export 'src/common.dart';
+export 'src/google_http_client.dart';
 export 'widgets.dart';
 
 enum SignInOption { standard, games }
@@ -106,7 +106,7 @@ class GoogleSignInAccount implements GoogleIdentity {
       "Authorization": "Bearer $token",
       "X-Goog-AuthUser": "0",
     };
-    
+
     return GoogleHttpClient(headers);
   }
 
--- a/packages/google_sign_in/lib/google_sign_in.dart
+++ b/packages/google_sign_in/lib/google_sign_in.dart
@@ -7,6 +7,8 @@ import 'dart:ui' show hashValues;
 
 import 'package:flutter/services.dart' show MethodChannel, PlatformException;
 import 'package:meta/meta.dart' show visibleForTesting;
+import 'package:http/http.dart';
+import 'package:http/io_client.dart';
 
 import 'src/common.dart';
 
@@ -28,6 +30,20 @@ class GoogleSignInAuthentication {
 
   @override
   String toString() => 'GoogleSignInAuthentication:$_data';
+}
+
+class GoogleHttpClient extends IOClient {
+  GoogleHttpClient._(this._headers) : super();
+
+  Map<String, String> _headers;
+
+  @override
+  Future<StreamedResponse> send(BaseRequest request) =>
+      super.send(request..headers.addAll(_headers));
+
+  @override
+  Future<Response> head(Object url, {Map<String, String> headers}) =>
+      super.head(url, headers: headers..addAll(_headers));
 }
 
 class GoogleSignInAccount implements GoogleIdentity {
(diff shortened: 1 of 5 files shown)


rishab2113:
@cyanglaz any updates here?

packages/google_sign_in/lib/google_sign_in.dart
```diff
@@ -97,6 +113,15 @@ class GoogleSignInAccount implements GoogleIdentity {
     return GoogleSignInAuthentication._(response);
   }
 
+  Future<GoogleHttpClient> get httpClient async {
```
mehmetf: This whole file is lacking proper comments. Could you please start the snowball by commenting yours? :)

packages/google_sign_in/lib/google_sign_in.dart
```diff
@@ -30,6 +32,20 @@ class GoogleSignInAuthentication {
   String toString() => 'GoogleSignInAuthentication:$_data';
 }
 
+class GoogleHttpClient extends IOClient {
```
mehmetf: This should be in its own file under src/

packages/google_sign_in/lib/google_sign_in.dart
rishab2113: sure!

rishab2113:
@mehmetf minor version change right?

packages/google_sign_in/lib/google_sign_in.dart
```diff
@@ -15,6 +15,8 @@ import 'src/common.dart';
 export 'src/common.dart';
 export 'widgets.dart';
 
+part 'src/google_http_client.dart';
```
mehmetf: Don't use part.

import and export instead (like src/common.dart)

rishab2113:
@mehmetf looks good now?

mehmetf:
Thanks! Now the PR looks in good shape. However, I want to understand the rationale. Is this unblocking something you were not able to do before? or is it for convenience?

My concern is that the HttpClient has nothing much to do with the signin logic and adding package:http to the dependencies of every single client using this is not good.

rishab2113:
This essentially allows a developer to enable three-legged OAuth by enabling application calls to Google APIs on behalf of end users, for example to allow the app to access a user's BigTable data via a consent screen.

It is an implementation of [Google's Java HTTP Client Library](https://developers.google.com/api-client-library/java/google-http-java-client/) in Flutter.

mehmetf:
Thanks for the explanation. The motivation makes sense however could you not import a package that does this? 

Retrieving a "httpClient" for an account looks odd to me. What if, for instance, my application has its own custom httpclient that adds other headers? 

One solution could be: Does httpClient support interceptors that let you append headers? It would be nice for the account to provide an interceptor that can be installed and composed.

rishab2113:
Hmm, that makes sense, although this would serve to only initialize the Google API's in this way - 

[Source 1](https://pub.dartlang.org/documentation/googleapis/latest/googleapis.dialogflow.v2/DialogflowApi-class.html) 
```DialogflowApi(Client client, { String rootUrl: "https://dialogflow.googleapis.com/", String servicePath: "" })```

[Source 2](https://pub.dartlang.org/documentation/googleapis/latest/googleapis.bigquery.v2/BigqueryApi-class.html) 
```BigqueryApi(Client client, { String rootUrl: "https://www.googleapis.com/", String servicePath: "bigquery/v2/" })```

Allowing the user to instantiate top-level variables in the main.dart file, and pass around the authorized client to the entire application and thereby "authorizing" once.

The interceptors seem like a valuable addition though! However I would need some assistance implementing them - I'm guessing constructors would do the trick?

mehmetf:
Yeah absolutely. I have no doubt having a GoogleHttpClient that integrates well with GoogleSignIn is valuable. I am simply questioning whether it should be part of the plugin or not. You can attempt to create this as a separate package and see whether it can be done. We should be able to build Dart packages on top of plugins. You can create an HttpClient that expects a User in a named factory and generates an authenticated instance. All without touching the plugin code.

Interceptors are nice but that would require a bigger change as you point out.

cyanglaz:
@rishab2113 Any updates on this? Do you have a strong opinion on having this in the google_sign_in plugin or you also think it is better to have a separate package to do so.

rishab2113:
Personally, I think having it would benefit a lot of users, however, personally I’ve found a non-Flutter solution to this. 

I feel like having a way to automatically generate httpclients based on a user login is a valuable tool, and should be part of  the sign in package!

cyanglaz:
@rishab2113 Thank you for providing this PR. Unfortunately, after discussion with @mehmetf, we think this feature can be in a different package and we prefer not to merge this at this point. I am closing this PR now.

PR #1274, opened 2019-02-24
[cloud_firestore,firebase_database,firebase_storage] handling error nil on getFlutterError

ffeu:
When `getFlutterError` was introduced, some plugins started to throw exceptions for successful operations. I originally faced that with `firebase_database`, but I can see the same happened to `cloud_firestore`.

As you can see in the code, either the caller would have to be changed to check for `error == nil` before calling it, or we can do it inside `getFlutterError` and keep the plugin original code.

I checked the other plugins, and apparently they don't need the same change as they already do the check or use the `getFlutterError` differently.

Some of the formatting of the error code was slightly changed and I restored it to what it was, please let me know if that's ok or if that change was intentional.

intends to
 fix flutter/flutter#28344
 fix flutter/flutter#28065
 fix flutter/flutter#28103

cc @collinjackson @kroikie @cyanglaz

### DIFF
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.9.6
+
+* Fixing PlatformException(Error 0, null, null) which happened when a successful operation was performed.
+
 ## 0.9.5+1
 
 * Log messages about automatic configuration of the default app are now less confusing.
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.9.6
+## 0.9.5+2
 
 * Fixing PlatformException(Error 0, null, null) which happened when a successful operation was performed.
 
--- a/packages/firebase_database/ios/Classes/FirebaseDatabasePlugin.m
+++ b/packages/firebase_database/ios/Classes/FirebaseDatabasePlugin.m
@@ -7,7 +7,9 @@
 #import <Firebase/Firebase.h>
 
 static FlutterError *getFlutterError(NSError *error) {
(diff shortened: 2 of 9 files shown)


packages/firebase_database/ios/Classes/FirebaseDatabasePlugin.m
```diff
@@ -7,7 +7,9 @@
 #import <Firebase/Firebase.h>
 
 static FlutterError *getFlutterError(NSError *error) {
-  return [FlutterError errorWithCode:[NSString stringWithFormat:@"Error %d", (int)error.code]
+  if (error == nil) return nil;
```
cyanglaz: I personally prefer the idea to add the check before where the function is called like everywhere else. It avoids hiding a secret nil return. Also it keeps the consistency throughout the repo. Same for the cloud_firestore plugin.

packages/cloud_firestore/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.9.6
```
cyanglaz: 0.9.5+2 since it is pre 1.0.0

packages/firebase_database/ios/Classes/FirebaseDatabasePlugin.m
ffeu: I also prefer doing the check on the parent. However, when fixing code, I'd rather change it as little as possible to avoid new errors.

packages/cloud_firestore/CHANGELOG.md
ffeu: 👍

packages/firebase_database/ios/Classes/FirebaseDatabasePlugin.m
cyanglaz: fair enough!

nemoryoliver:
Amazing! Thanks a lot for the fix.

PR #1276, opened 2019-02-24
[shared_preferences] copying list for prevent mutate from outside

ivk1800:
fixed https://github.com/flutter/flutter/issues/28015


packages/shared_preferences/lib/shared_preferences.dart
```diff
@@ -127,7 +128,7 @@ class SharedPreferences {
           .invokeMethod('remove', params)
           .then<bool>((dynamic result) => result);
     } else {
-      _preferenceCache[key] = value;
+      _putToCache(key, value);
```
collinjackson: Since this is only used in one place, I don't think we need to split it out into its own method.

PR #1283, opened 2019-02-26
[webview_flutter] propagate more events back to flutter, allow customization of UserAgent, basic auth

hpoul:
Hi,
I required a few customizations for the platform WebView and wondered if it might be useful for others:

* Customization of the User-Agent string (on iOS only since 9).
* Getting notifications for page loading, finished and error events.
* Handling authorization requests (for basic authorization)

let me know if this is something which makes sense to be contributed back.

thanks,
  Herbert


hpoul:
I signed it!

hpoul:
@amirh maybe you could take a look if these are things which fit into the plans for the webview plugin? thanks.

hpoul:
i will split this PR up into smaller once. first one https://github.com/flutter/plugins/pull/1389

PR #1285, opened 2019-02-26
[shared_preferences] driver test using package:test on device

collinjackson:
Relates to flutter/flutter#26080

This PR shows how I'd like plugin tests to be organized in the future. Each plugin should include an example app, and the driver tests should live in the tests folder of the example. In addition to (or instead of) mocking out a MethodChannel implementation we can confirm functionality on simulators.

To run a test, here's the command:

```
 flutter drive test/shared_preferences.dart 
```

Right now there isn't a good way to get the individual results from package:test other than looking at the logs, my plan to get things going in the short term is to have this be cirrus task and we can improve the integration going forward.


packages/shared_preferences/example/test/shared_preferences.dart
```diff
@@ -0,0 +1,91 @@
...
+    test('removing', () async {
+      const String key = 'testKey';
+      preferences
+        ..setString(key, null)
+        ..setBool(key, null)
+        ..setInt(key, null)
+        ..setDouble(key, null)
+        ..setStringList(key, null);
+      await preferences.remove(key);
+      expect(preferences.get('testKey'), isNull);
```
cyanglaz: I am not too familiar with shared preference. Seems like we set 'testKey' with null value for all the set methods, will `expect(preferences.get('testKey'), isNull);` succeed even without `await preferences.remove(key);`

packages/shared_preferences/example/test/shared_preferences.dart
```diff
@@ -0,0 +1,91 @@
...
+    });
+
+    test('clearing', () async {
+      await preferences.clear();
+      expect(preferences.getString('String'), null);
```
cyanglaz: Maybe we can set something before we call `await preferences.clear();` so we can be sure that all the 'expect' methods are not returning a false positive.

PR #1286, opened 2019-02-26
[in_app_purchase] Adds Dart BillingClient APIs for loading purchases

mklim:
1. Wires the Purchase data class and serializers in to Dart.
2. BillingClient now requires a callback for responding to purchase
   updates to be passed in at construction time.
3. Exposes `BillingClient#queryPurchases` and
   `BillingClient#queryPurchaseHistory`.

For now GooglePlayConnection passes in a lambda that does nothing to
`BillingClient`. This will need to be fixed when all the connections are
updated.

### DIFF
--- a/packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
+++ b/packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
@@ -134,8 +134,8 @@ class BillingClient {
   /// to complete the transaction there.
   ///
   /// This method returns a [BillingResponse] representing the initial attempt
-  /// to show the Google Play purchase screen.
-  /// TODO(mklim, flutter/flutter#26326): Expose onPurchasesUpdated() result.
+  /// to show the Google Play billing flow. Actual purchase updates are
+  /// delivered via the [PurchasesUpdatedListener].
   ///
   /// This method calls through to
   /// [`BillingClient#launchBillingFlow`](https://developer.android.com/reference/com/android/billingclient/api/BillingClient#launchbillingflow).
--- a/packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
+++ b/packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
@@ -7,11 +7,32 @@ import 'package:flutter/services.dart';
 import 'package:flutter/foundation.dart';
 import 'package:json_annotation/json_annotation.dart';
 import '../channel.dart';
+import 'purchase_wrapper.dart';
 import 'sku_details_wrapper.dart';
 import 'enum_converters.dart';
 
+const String _kOnPurchasesUpdated =
+    'PurchasesUpdatedListener#onPurchasesUpdated(int, List<Purchase>)';
 const String _kOnBillingServiceDisconnected =
     'BillingClientStateListener#onBillingServiceDisconnected()';
+
+/// Callback triggered by Play in response to purchase activity.
+///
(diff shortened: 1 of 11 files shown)


packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -26,8 +47,10 @@ const String _kOnBillingServiceDisconnected =
 /// some minor changes to account for language differences. Callbacks have been
 /// converted to futures where appropriate.
 class BillingClient {
-  BillingClient() {
+  BillingClient(PurchasesUpdatedListener onPurchasesUpdated) {
+    assert(onPurchasesUpdated != null);
```
cyanglaz: maybe assert _callback != null as well?

packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -134,8 +157,52 @@ class BillingClient {
...
+  Future<PurchasesResultWrapper> queryPurchaseHistory(SkuType skuType) async {
+    assert(skuType != null);
+    return PurchasesResultWrapper.fromJson(await channel.invokeMapMethod(
+        'BillingClient#queryPurchaseHistoryAsync(String, PurchaseHistoryResponseListener)',
+        <String, dynamic>{'skuType': SkuTypeConverter().toJson(skuType)}));
+  }
+
   Future<void> _callHandler(MethodCall call) async {
     switch (call.method) {
+      case _kOnPurchasesUpdated:
```
cyanglaz: maybe assert _callback != null?

packages/in_app_purchase/lib/src/billing_client_wrappers/purchase_wrapper.dart
```diff
@@ -0,0 +1,132 @@
...
+    return typedOther.responseCode == responseCode &&
+        typedOther.purchasesList == purchasesList;
+  }
+
+  @override
+  int get hashCode => hashValues(responseCode, purchasesList);
+
+  /// The status of the operation.
+  ///
+  /// Unfortunately this can represent both the status of the "query purchase
```
cyanglaz: (Personal opinion)Unfortunately seems too casual for dart doc? I don't know if I can find any document to support my point.

packages/in_app_purchase/lib/src/billing_client_wrappers/purchase_wrapper.dart
```diff
@@ -0,0 +1,132 @@
...
+  /// history" half of the operation and the "user made purchases" transaction
+  /// itself.
+  final BillingResponse responseCode;
+
+  /// The list of succesful purchases made in this transaction. May be
+  /// empty, especially if [responseCode] is not [BillingResponse.ok].
```
cyanglaz: Can it be empty in other scenarios?

packages/in_app_purchase/lib/src/billing_client_wrappers/billing_client_wrapper.dart
mklim: Discussed offline, since this is set as non-null at construction time and never reassigned an assert isn't needed here.

packages/in_app_purchase/lib/src/billing_client_wrappers/purchase_wrapper.dart
mklim: I don't know of any, but I don't want to definitively claim this in case there's edge cases I'm not aware of.

PR #1287, opened 2019-02-27
[image_picker] generate thumbnails of images and videos

ivk1800:
this is PR adds new feature requested of https://github.com/flutter/flutter/issues/26191

added new functions:
Future<PickingResult> pickImageWithThumbnail
Future<PickingResult> pickVideoWithThumbnail
Future<File> generateImageThumbnail
Future<File> generateVideoThumbnail

also fixed:
https://github.com/flutter/flutter/issues/23503
https://github.com/flutter/flutter/issues/27950
https://github.com/flutter/flutter/issues/2848
https://github.com/flutter/flutter/issues/21571
https://github.com/flutter/flutter/issues/25469
https://github.com/flutter/flutter/issues/26194


Deftness:
@pepegich dang, I really need this feature since other resize libraries I've found do it solely in dart in comparison to the native OS.

Do we know what might be the timetable for this to actually be merged? Until then, I'm going to have the 10 second resize hiccups...

ghxstdev:
Will this work for FilePicker? (multiFilePath())

packages/image_picker/ios/Classes/ImagePickerPlugin.m
```diff
@@ -99,11 +99,60 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
+                                         timeOption:MPMovieTimeOptionNearestKeyFrame];
+
+      [self generateThumbnail:image result:result width:width height:height];
+    });
   } else {
     result(FlutterMethodNotImplemented);
   }
 }
 
+- (void)generateThumbnail:(UIImage *)image
```
cyanglaz: What is the difference of this and the regular pickImage with maxWidth and maxHeight?

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ThumbnailCreator.java
```diff
@@ -0,0 +1,92 @@
...
+import android.media.ThumbnailUtils;
+import android.os.AsyncTask;
+import android.provider.MediaStore;
+import androidx.core.util.Consumer;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+
+public class ThumbnailCreator {
```
cyanglaz: public seems to be an overkill for this class if it is only used inside the package. Can just be the default scope. So are the 'public' methods in this class.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ThumbnailCreator.java
```diff
@@ -0,0 +1,92 @@
...
+      String[] pathParts = originalVideoPath.split("/");
+      String imageName = pathParts[pathParts.length - 1];
+
+      File imageFile = new File(externalFilesDirectory, "/thumbnail" + imageName);
+      FileOutputStream fileOutput = new FileOutputStream(imageFile);
+      fileOutput.write(outputStream.toByteArray());
+      fileOutput.close();
+
+      return imageResizer.resizeImageIfNeeded(originalVideoPath, width, height);
+    } catch (IOException e) {
```
mklim: Catching and rethrowing the exception will make failures here hard to debug since it's obfuscating the original type of problem. Can we handle the I/O exceptions instead?

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/BitmapAsyncGenerator.java
```diff
@@ -0,0 +1,32 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package io.flutter.plugins.imagepicker;
+
+import android.os.AsyncTask;
+import androidx.core.util.Consumer;
+
+public class BitmapAsyncGenerator extends AsyncTask<Void, Void, String> {
```
mklim: nit: This should be package private.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ThumbnailCreator.java
```diff
@@ -0,0 +1,92 @@
...
+    BitmapAsyncGenerator generator =
+        new BitmapAsyncGenerator(
+            callback,
+            new Supplier<String>() {
+              @Override
+              public String get() {
+                return generateImageThumbnail(originalImagePath, width, height);
+              }
+            });
+    generator.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
```
mklim: I think it's safer to just use `execute()` here. We don't need these tasks to run in parallel.

pepegich: in case if need display images in grid layout, without parallel execution, bitmaps will be generated too slow. AsyncTask limit count of parallel task at same time.

cyanglaz:
I apologize for getting back on this after a bit while. The iOS code for image_picker has changed a lot since you initially created the PR. Could you try to rebase and fix the conflicts?

ivk1800:
@cyanglaz I don't have time to do this right now,

cyanglaz:
@ivk1800 Ok. After some discussion with the team, we think such functionality should be in a separate package rather than in the image_picker, extracting thumbnails from a video might be a good feature in the video player plugin but definitely not the image picker. I am closing the PR. Please feel free to discuss more with us if you think otherwise.

PR #1292, opened 2019-02-28
[firebase_auth] Make providerId 'const String' for easier use in switch statements

dotdoom:


dotdoom:
+cc @collinjackson as the original author, PTAL?

collinjackson:
Looks good, can you update the pubspec.yaml and CHANGELOG.md for release?

dotdoom:
Done!

packages/firebase_auth/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: Flutter plugin for Firebase Auth, enabling Android and iOS
   like Google, Facebook and Twitter.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/firebase_auth
-version: "0.8.2"
+version: "0.8.2+1"
```
collinjackson: IMO this change merits an update to 0.8.3 since the public API changed (albeit in a backwards compatible way)

PR #1293, opened 2019-02-28
[google_maps_flutter] The FloatingActionButton requires an icon

sam-toon:
The sample code doesn't work.  It needs an icon specified for the FloatingActionButton.


sam-toon:
I signed it!

PR #1294, opened 2019-02-28
Fix icon issue on ios and manifest on android

iskakaushik:


packages/google_maps_flutter/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -8,7 +8,8 @@
         android:name="io.flutter.app.FlutterApplication"
         android:label="google_maps_flutter_example"
         android:icon="@mipmap/ic_launcher">
-        <meta-data
+      <uses-library android:name="org.apache.http.legacy" android:required="false"/>
```
amirh: Can we add this to the plugin's manifest, should we?
What if we want to use maps in an app that needs this to be false?

PR #1299, opened 2019-02-28
[in_app_purchase]restore purchases

cyanglaz:
The iOS implementation for https://github.com/flutter/flutter/issues/26326


packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -79,6 +79,18 @@ class SKPaymentQueueWrapper {
...
   }
 
+  /// Restore transactions to maintain access to content that customers have already purchased.
+  ///
+  /// For example, when a user upgrade to a new phone, they want to keep the content they purchased in the old phone.
+  /// This call will invoke the [SKTransactionObserverWrapper.restoreCompletedTransactions] or [SKTransactionObserverWrapper.paymentQueueRestoreCompletedTransactionsFinished] as well as [SKTransactionObserverWrapper.updatedTransaction]
+  /// in the [SKTransactionObserverWrapper]. If you keep the download content in your own server, in the observer methods, you can simply finish the transaction by calling [finishTransaction] and
+  /// download the content from your own server.
+  /// If you keep the download content on Apple's server, you can access the download content in the transaction object that you get from [SKTransactionObserverWrapper.updatedTransaction] when the [SKPaymentTransactionWrapper.transactionState] is [SKPaymentTransactionStateWrapper.restored].
+  Future<void> restoreTransactions({String applicationName}) async {
```
mklim: What should `applicationName` be here? The bundle ID?

In general I think we should also link to the API docs on this method too.

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
```diff
@@ -79,6 +79,18 @@ class SKPaymentQueueWrapper {
         transaction.transactionIdentifier);
   }
 
+  /// Restore transactions to maintain access to content that customers have already purchased.
+  ///
+  /// For example, when a user upgrade to a new phone, they want to keep the content they purchased in the old phone.
```
mklim:
```suggestion
  /// For example, when a user upgrades to a new phone, they want to keep the content they purchased in the old phone.
```

packages/in_app_purchase/lib/src/store_kit_wrappers/sk_payment_queue_wrapper.dart
cyanglaz: The applicationName is the userApplicationName. And ill add documents on this parameter and attach the apple's doc here as well.

PR #1301, opened 2019-03-01
[cloud_functions]fixed error introduced by #1210

rishab2113:
Fixes [flutter/flutter#28578](https://github.com/flutter/flutter/issues/28578), which was an artifact introduced by #1210 which tried to fix [flutter/flutter#27888](https://github.com/flutter/flutter/issues/27888) and [flutter/flutter#24333](https://github.com/flutter/flutter/issues/24333).

This issue is most likely caused by #1210 as this commit is correlated with the version roll to 0.1.1, which is confirmed by [git blame](https://github.com/flutter/plugins/blame/f2b1e696eb1391fa734090faaac66b46057a0d38/packages/cloud_functions/lib/cloud_functions.dart#L25).

Thanks to @aytunch for confirming that the build works with ```cloud_functions: 0.1.0+1```

Performs a null check on the variable, reverting it to the default ```us-central1``` if not found.

### DIFF
--- a/packages/cloud_functions/lib/cloud_functions.dart
+++ b/packages/cloud_functions/lib/cloud_functions.dart
@@ -22,7 +22,7 @@ class CloudFunctionsException implements Exception {
 class CloudFunctions {
   CloudFunctions({FirebaseApp app, String region})
       : _app = app ?? FirebaseApp.instance,
-        _region = region;
+        _region = region ?? "us-central1";
 
   @visibleForTesting
   static const MethodChannel channel = MethodChannel('cloud_functions');
--- a/AUTHORS
+++ b/AUTHORS
@@ -25,3 +25,4 @@ Tuyển Vũ Xuân <netsoft1985@gmail.com>
 Sarthak Verma <sarthak@artiosys.com>
 Mike Diarmid <mike@invertase.io>
 Invertase <oss@invertase.io>
+Rishab Nayak <rishab@bu.edu>
--- a/packages/cloud_functions/test/cloud_functions_test.dart
+++ b/packages/cloud_functions/test/cloud_functions_test.dart
@@ -41,7 +41,7 @@ void main() {
             'CloudFunctions#call',
             arguments: <String, dynamic>{
               'app': '[DEFAULT]',
-              'region': null ?? 'us-central1',
+              'region': 'us-central1',
               'functionName': 'baz',
               'parameters': null,
             },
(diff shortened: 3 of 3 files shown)


packages/cloud_functions/lib/cloud_functions.dart
```diff
@@ -22,7 +22,7 @@ class CloudFunctionsException implements Exception {
 class CloudFunctions {
   CloudFunctions({FirebaseApp app, String region})
       : _app = app ?? FirebaseApp.instance,
-        _region = region;
+        _region = region ?? "us-central1";
```
cyanglaz: I feel like defaulting to a certain value behind the scene might be hard to debug for users that does not enable the "us-central1" region in their application. Do google cloud always enable "us-central1"? 
Maybe `assert (region != null) is better.
@collinjackson

rishab2113: I agree, but apparently the Cloud Functions docs say that they run in the us-central1 region by default. [<Link>](https://firebase.google.com/docs/functions/locations)

collinjackson:
Thanks for the contribution! I fixed this a different way, see https://github.com/flutter/plugins/pull/1316

gustavobrian:
help me please:
 firebase_auth: ^0.8.1+4
 cloud_firestore: ^0.9.5+2
 cloud_functions: 0.1.1+1
 firebase_storage: ^2.1.0+1
 firebase_core: ^0.3.1+1
 firebase_analytics: ^2.0.3
 firebase_messaging: ^4.0.0+1


Xcode's output:
↳
    /Volumes/Home/gustavobrian/.pub-cache/hosted/pub.dartlang.org/cloud_functions-0.1.1+1/ios/Classes/CloudFunctionsPlugin.m:49:16: error: no visible @interface for 'FIRFunctions' declares the selector 'callFunction:withObject:completion:'
        [functions callFunction:functionName
         ~~~~~~~~~ ^~~~~~~~~~~~~~~~~~~~~~~~~
    1 error generated.

rishab2113:
@gustavobrian #1316 fixed this issue, there should be a version bump for cloud_functions from 0.1.1+1 to 0.1.2 soon; that should take care of this issue for you! Till then, hold on, or, you could get the bleeding-edge version by editing your cloud_functions entry in your pubspec.yaml file to look like this

```
  cloud_functions:
    git:
      url: https://github.com/flutter/plugins.git
      path: packages/cloud_functions
```

gustavobrian:
> #1316

i love you!!!

PR #1302, opened 2019-03-01
[google_maps_flutter]ChangeNotifier is replaced with granular callbacks

iskakaushik:
This facilitates better interaction with the map
than adding a listener to the controller.

### DIFF
--- a/packages/google_maps_flutter/lib/src/google_map.dart
+++ b/packages/google_maps_flutter/lib/src/google_map.dart
@@ -8,7 +8,7 @@ typedef void MapCreatedCallback(GoogleMapController controller);
 
 /// Callback that receives updates to the camera position.
 ///
-/// This callback is triggered when the platform google map
+/// This callback is triggered when the platform Google Map
 /// registers a camera movement. This will be called with null if
 /// [GoogleMap.trackCameraPosition] is false.
 ///
@@ -78,11 +78,11 @@ class GoogleMap extends StatefulWidget {
   /// This can be initiated by the following:
   /// 1. Non-gesture animation initiated in response to user actions.
   ///    For example: zoom buttons, my location button, or marker clicks.
-  /// 2. Developer initiated animation.
+  /// 2. Programmatically initiated animation.
   /// 3. Camera motion initiated in response to user gestures on the map.
   ///    For example: pan, tilt, pinch to zoom, or rotate.
   ///
-  /// Note: This is callback is called even if trackCameraPosition is false.
+  /// Note: This is callback is called even if [trackCameraPosition] is false.
   final VoidCallback onCameraMoveStarted;
 
   /// Called repeatedly as the camera continues to move after an
@@ -91,7 +91,7 @@ class GoogleMap extends StatefulWidget {
   /// This may be called as often as once every frame and should
   /// not perform expensive operations.
   ///
-  /// This is callback is not called when trackCameraPosition is false.
+  /// This is only called if [trackCameraPosition] is true.
   final CameraPositionCallback onCameraMove;
 
   /// Called when camera movement has ended, there are no pending
--- a/packages/google_maps_flutter/lib/src/google_map.dart
+++ b/packages/google_maps_flutter/lib/src/google_map.dart
@@ -6,6 +6,11 @@ part of google_maps_flutter;
 
 typedef void MapCreatedCallback(GoogleMapController controller);
 
+/// Callback that tracks camera position.
+///
+/// Will be null if trackCameraPosition on [GoogleMap] is false.
+typedef void CameraPositionCallback(CameraPosition position);
+
 class GoogleMap extends StatefulWidget {
   const GoogleMap({
     @required this.initialCameraPosition,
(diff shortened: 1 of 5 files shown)


packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -6,6 +6,11 @@ part of google_maps_flutter;
 
 typedef void MapCreatedCallback(GoogleMapController controller);
 
+/// Callback that tracks camera position.
+///
```
amirh: Probably worth adding "Used by: " and listing the relevant GoogleMap fields.

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -58,9 +68,24 @@ class GoogleMap extends StatefulWidget {
   /// True if the map view should relay camera move events to Flutter.
   final bool trackCameraPosition;
 
-  // Markers to be placed on the map.
+  /// Markers to be placed on the map.
   final Set<Marker> markers;
 
+  /// Callback for when the camera move started.
```
amirh: This doesn't add much information.
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-useless-documentation

Documentation is definitely hard 😄 I sometimes find the prompts here helpful:
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#writing-prompts-for-good-documentation

Some questions I might have been looking for answers in the docs are: is this called on the first time the map is loaded? is this called when the position is updated programmatically?

(similar comments to the docs of the other new fields below)

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -58,9 +68,24 @@ class GoogleMap extends StatefulWidget {
...
+  final VoidCallback onCameraMoveStarted;
+
+  /// Callback for when the camera is moving.
```
amirh: This should probably say something about [GoogleMap.trackCameraPosition]

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -58,9 +68,24 @@ class GoogleMap extends StatefulWidget {
...
+  final CameraPositionCallback onCameraMove;
+
+  /// Callback for when the camera has entered an idle stage.
+  final VoidCallback onCameraIdle;
+
+  /// Callback when the [GoogleMap] options have been applied on the platform.
+  final CameraPositionCallback onMapOptionsUpdate;
```
amirh: I wonder if the onOptionsUpdate and onMarkersUpdate callback are really needed? do you have some concrete use-cases in mind?
We can always add fields later, but removing will be a breaking API change.

packages/google_maps_flutter/lib/src/google_map.dart
iskakaushik: Great read. Will revise the docs based on this.

packages/google_maps_flutter/lib/src/google_map.dart
iskakaushik: They provide a way for users to know when the platform has "committed" the update. It might be useful to take an action based on it. We could remove them, I put them here to maintain compatibility with the behavior before.

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -6,6 +6,15 @@ part of google_maps_flutter;
 
 typedef void MapCreatedCallback(GoogleMapController controller);
 
+/// Callback that receives updates to the camera position.
+///
+/// This callback is triggered when the platform google map
```
amirh: nit: Google Map

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -58,9 +70,34 @@ class GoogleMap extends StatefulWidget {
...
-  // Markers to be placed on the map.
+  /// Markers to be placed on the map.
   final Set<Marker> markers;
 
+  /// Called when the camera starts moving.
+  ///
+  /// This can be initiated by the following:
+  /// 1. Non-gesture animation initiated in response to user actions.
+  ///    For example: zoom buttons, my location button, or marker clicks.
+  /// 2. Developer initiated animation.
```
amirh: nit: I'd call it "Programmatically initiated animation."

packages/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -58,9 +70,34 @@ class GoogleMap extends StatefulWidget {
...
+  /// Note: This is callback is called even if trackCameraPosition is false.
+  final VoidCallback onCameraMoveStarted;
+
+  /// Called repeatedly as the camera continues to move after an
+  /// onCameraMoveStarted call.
+  ///
+  /// This may be called as often as once every frame and should
+  /// not perform expensive operations.
+  ///
+  /// This is callback is not called when trackCameraPosition is false.
```
amirh: link trackCameraPosotion.

amirh: nit: since the default for trackCameraPosition is false, I'll reverse the negation (This is only called if [trackCameraPosition] is false).

BTW does it never gets called if trackCameraPosition is false? in that case maybe it would make sense to get rid of trackCameraPosition and use the presence of this callback as the flag?

iskakaushik: I am pro getting rid of `trackCameraPosition` entirely, it's a notion introduced by the plugin and not something that is native to Google Map platform implementations. Cool if i do it in a follow up PR?

amirh: SGTM

iskakaushik:
merging since these are successes on https://cirrus-ci.com/build/5658597900943360

PR #1303, opened 2019-03-02
[in_app_purchase]retrieve receipt

cyanglaz:
Retrieve receipt data. The data can be retrieve as a base 64 string or dictionary. User can use base 64 string to validate the purchase with App Store; use dictionary to validate purchase locally. 
Part of https://github.com/flutter/flutter/issues/26326


packages/in_app_purchase/ios/Classes/FIAPRequestHandler.m
```diff
@@ -10,13 +10,13 @@
 @interface FIAPRequestHandler () <SKProductsRequestDelegate>
 
 @property(copy, nonatomic) ProductRequestCompletion completion;
-@property(strong, nonatomic) SKProductsRequest *request;
+@property(strong, nonatomic) SKRequest *request;
```
cyanglaz: We need to change SKProductsRequest to SKRequest here because we need to handle SKRequests other than SKProductsRequest now. (SKRefreshReceiptRequest)

PR #1305, opened 2019-03-03
add headers param as option to the Webview Widget

chakAs3:
Based on a requirement we had to pass http header to our Webview but the current one does not support header so I added the feature to specify the headers on Webview Widget


chakAs3:
I signed it!

amirh:
https://github.com/flutter/plugins/pull/1516 added support for adding custom request headers.

PR #1306, opened 2019-03-03
[cloud_firestore] Update firebase-firestore to 18.2.0

jul1u5:
Update `firebase-firestore` dependency to `18.2.0`. Previously, using `setData` method on `DocumentReference` would result in a crash.

Resolves https://github.com/flutter/flutter/issues/28082

### DIFF
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -1,4 +1,5 @@
-## 0.9.7+3
+## 0.9.8
+
 * Bump Android dependencies to latest.
 
 ## 0.9.7+2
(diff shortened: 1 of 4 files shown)


ened:
@jul1u5 could you update the PR to cloud_firestore 18.1.0 & also bump the version / update changelog?

ened:
@jul1u5 Is multidex *strictly* required? Seems like a parameter local to your project?

ened:
@collinjackson @kroikie this seems like a good PR for you to review?

The firestore version bump is 17 -> 18, should the plugin then also make a major version bump (or minor, as it's still 0.x). So 0.9.7+2 -> 0.10.0 ?

collinjackson:
If behavior changed, the version should increase, but since this isn't a breaking change (I think) and it's 0.x the increase should be minor.

PR #1308, opened 2019-03-04
[shared_preferences] release active instance for create new singleton with new mock values

ivk1800:
fixed https://github.com/flutter/flutter/issues/28837

I added new static method releaseSingleton for release active instance of singleton


pepegich:
@dav1dev thank you for your comment, I commited

collinjackson:
This is similar functionality to another PR: #1469.

It seems like a non-singleton testing constructor might be an attractive alternative to resetting singletons. Does anyone else have opinions on this?

collinjackson:
Following up on this one -- it seems like every time you're calling reset you're also calling `setMockInitialValues` at the same time. I'm wondering if `setMockInitialValues` should just clear the cache as part of its functionality. I'll open a separate PR for this.

PR #1314, opened 2019-03-05
trackCameraPosition is inferred from GoogleMap.onCameraMove

iskakaushik:


packages/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.0
+
+* trackCameraPosition is inferred from GoogleMap.onCameraMove being set.
```
amirh: Let's make it part of 0.4.0

PR #1245, opened 2019-02-20
[image_picker] Fix image picker record video has two files on IOS

TenkaiRuri:
When I record, two files will appear on ios, that not my expected.

PR #1288, opened 2019-02-27
[image_picker] delete original file if scaled

pepegich:
fixed https://github.com/flutter/flutter/issues/23500

PR #1315, opened 2019-03-05
[image_picker] remove unnecessary file path for video.

cyanglaz:
Simplified version of https://github.com/flutter/plugins/pull/1245 fix. 
When pick videos, we used to create an unnecessary extra file. This PR removes the extra temp file logic.


pepegich:
seems related https://github.com/flutter/plugins/pull/1288

cyanglaz:
> seems related #1288

Actually this PR is for iOS and https://github.com/flutter/plugins/pull/1288 is for Android. So although both are trying to solve similar problems but they are separate fixes.

amirh:
The tmp path logic was added in https://github.com/flutter/image_picker/commit/5409baccb75680547b1a799f6853da423e5fd9cb

@collinjackson do you know whether we still need it?

cyanglaz:
The original videoURL is already a tmpPath. I thought adding another tmpPath did not make sense. I guess it could be a copy paste error (copy pasted from the picking image logic) when we introduced the video capability in the plugin.

collinjackson:
lgtm

PR #1316, opened 2019-03-06
[cloud_functions] Specify version for CocoaPod and handle null regions gracefully

collinjackson:
Recently Cloud Functions changed their API in a breaking fashion, which caused build errors for some users. Also fixes a regression in the new region handling that didn't handle the default case gracefully.

Fixes flutter/flutter#28433
Fixes flutter/flutter#28578

I verified manually that the iOS app works properly.


iskakaushik:
LGTM

rishab2113:
did this go through @iskakaushik? The version 0.1.2 shows up as not matching any versions

iskakaushik:
@rishab2113 this needs to get published. @collinjackson, could you please publish this when you get a chance?

PR #1318, opened 2019-03-06
[Firebase messaging] iOS direct data messages

jaumard:
fixes https://github.com/flutter/flutter/issues/25650
fixes https://github.com/flutter/flutter/issues/29714

To use this right now do:
```yaml
firebase_messaging: #^4.0.0+1
    git:
      url: git://github.com/jaumard/plugins
      ref: feature/iosFirebaseMessagingDataMessages
      path: packages/firebase_messaging
```

### DIFF
--- a/packages/firebase_messaging/CHANGELOG.md
+++ b/packages/firebase_messaging/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.0.1+1
+
+*  Update iOS to support direct data messages
+
 ## 5.0.1
 
 * Fix error in the logs on startup if unable to retrieve token on startup on Android.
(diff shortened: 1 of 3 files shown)


amervelic:
Please fix custom notification sound on android oreo+

kroikie:
@jaumard Thanks for the PR! Could you please make this an optional operation on the Dart side?

If a developer is not using direct messages then there is no need to have this channel open. So let's follow the native SDK and give the option to the developer to enable the channel or not.

jaumard:
@kroikie yes I can, but it will change the cross platform behavior. 
Because on Android you get out of the box push notification and direct messages, shouldn't be the same on iOS ? that's why I didn't hide it behind a flag. 

Let me know what you think !

collinjackson:
Thanks, this PR looks great!

Perhaps we should call `shouldEstablishDirectChannel` in response to `configure` since at that point the `onMessage` method will actually do something.

Don't forget to update pubspec.yaml

jaumard:
Hey @collinjackson I've rebase and update pubspec and changelog, let me know if it's all good :)

jaumard:
I also moved the shouldEstablishDirectChannel in configure response

PR #1321, opened 2019-03-06
[os_tester] WIP plugin exposing EarlGrey and Espresso APIs

collinjackson:
This is a sketch of a plugin that could expose interactions with non-Flutter UI to Flutter tests.

It could also be used for taking screenshots.

Right now it is based on XCTest/EarlGrey 1.0 but I'd like to migrate it to XCUITest/EG 2.0.


packages/os_tester/ios/os_tester.podspec
```diff
@@ -0,0 +1,22 @@
...
+A new flutter plugin project.
+                       DESC
+  s.homepage         = 'http://example.com'
+  s.license          = { :file => '../LICENSE' }
+  s.author           = { 'Your Company' => 'email@example.com' }
+  s.source           = { :path => '.' }
+  s.source_files = 'Classes/**/*'
+  s.public_header_files = 'Classes/**/*.h'
+  s.dependency 'Flutter'
+  s.dependency 'EarlGrey'
```
tirodkar: EarlGrey 1.0 is currently not being actively developed. We're moving users to EarlGrey 2.0. However, it doesn't have CocoaPods support at the moment. Do you have support for Carthage or so?

collinjackson: The Flutter plugins system is CocoaPods-based. I can explore what it would take to use Carthage. I'm curious to learn more about the challenges of publishing a CocoaPod for EarlGrey 2.0

tirodkar: EG 2 Path: https://github.com/google/EarlGrey/tree/earlgrey2

We have two separate components that need to be built from source, with one being turned into a framework (The AppFramework dir). We haven't been able to get to it from our side, however it shouldnt' be a huge task.

Also, I was looking at your usage of EarlGrey - EG should implicitly wait for the view to be loaded. My guess is that you are using the GREYCondition because the rendering of the view is being done on a separate thread from the iOS main thread?

iskakaushik:
`os_tester` project also needs to have a `pubspec.yml`

collinjackson:
This has been replaced by the e2e and espresso packages (earlgrey coming soon)

PR #1323, opened 2019-03-07
Allow specifying a navigation delegate (iOS implementation).

amirh:
This is the iOS implementation of the navigation delegate method channel.

The Dart and Android implementations are in https://github.com/flutter/plugins/pull/1236

Splitting off the iOS to keep a reasonable change size.
This PR will be merged first.

https://github.com/flutter/flutter/issues/25329

### DIFF
--- a/packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.h
+++ b/packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.h
@@ -1,4 +1,4 @@
-// Copyright 2018 The Chromium Authors. All rights reserved.
+// Copyright 2019 The Chromium Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
--- a/packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.h
+++ b/packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.h
@@ -1,0 +1,21 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Flutter/Flutter.h>
+#import <WebKit/WebKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FLTWKNavigationDelegate : NSObject <WKNavigationDelegate>
+
+- (instancetype)initWithChannel:(FlutterMethodChannel*)channel;
+
+/**
+ * Whether to delegate navigation decisions over the method channel.
+ */
+@property(nonatomic, assign) BOOL hasDartNavigationDelegate;
+
+@end
+
+NS_ASSUME_NONNULL_END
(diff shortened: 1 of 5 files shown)


packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.h
```diff
@@ -0,0 +1,21 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
```
cyanglaz: 2019?

packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.m
```diff
@@ -0,0 +1,39 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
```
cyanglaz: 2019? lol

cyanglaz: In this diff it looks like it is a new file. If it is indeed created in 2018, please ignore my comment.

packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.m
```diff
@@ -0,0 +1,39 @@
...
+- (void)webView:(WKWebView*)webView
+    decidePolicyForNavigationAction:(WKNavigationAction*)navigationAction
+                    decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler {
+  if (!self.hasDartNavigationDelegate) {
+    decisionHandler(WKNavigationActionPolicyAllow);
+    return;
+  }
+  NSDictionary* arguments = @{
+    @"url" : navigationAction.request.URL.absoluteString,
+    @"isMainFrame" : [NSNumber numberWithBool:navigationAction.targetFrame.isMainFrame]
```
cyanglaz: Just a suggestion, but the code looks fine. Feel free to not update this for this PR if you don't feel like to. 
The [NSNumber numberWithBool..] can be short written as @(foo).
```suggestion
    @"isMainFrame" : @(navigationAction.targetFrame.isMainFrame)
```

packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.m
```diff
@@ -0,0 +1,39 @@
...
+  };
+  [_methodChannel invokeMethod:@"navigationRequest"
+                     arguments:arguments
+                        result:^(id _Nullable result) {
+                          NSNumber* typedResult = result;
+                          decisionHandler([typedResult boolValue] ? WKNavigationActionPolicyAllow
```
cyanglaz: Is there a possibility for the dart side code to fail and the  result block not getting triggered?

packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.h
amirh: done

packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.m
amirh: done

packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.m
amirh: good catch, it shouldn't ever happen, but added some better logging in case it does.

packages/webview_flutter/ios/Classes/FLTWKNavigationDelegate.m
amirh: sweet!

PR #1328, opened 2019-03-09
[package_info] Add fallback on iOS when CFBundleDisplayName is not available

mortenboye:
In that case, fall back to CFBundleName, before returning NSNull


mortenboye:
@cyanglaz Thank you for your review. I can follow your comment about the potential false positive, using the bundleName as a fallback, so I'm going to close this PR. Thank you

PR #1330, opened 2019-03-10
[webview_flutter] Enable geolocation on Android

iamluciano:
Quick hack for enabling geolocation on Android. Native app still needs to ask for required geo permissions. Temporary solution for requesting permissions without further modifications to the webview_flutter plugin:

pubspec.yml:
```
dependencies:
  geolocator: ^3.0.0
```

Somewhere in app's code before creating WebView:

`Geolocator().getLastKnownPosition(desiredAccuracy: LocationAccuracy.high);`

### DIFF
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -8,6 +8,9 @@ import android.annotation.TargetApi;
 import android.content.Context;
 import android.os.Build;
 import android.view.View;
+import android.webkit.GeolocationPermissions;
+import android.webkit.WebChromeClient;
+import android.webkit.WebSettings;
 import android.webkit.WebStorage;
 import android.webkit.WebView;
 import io.flutter.plugin.common.BinaryMessenger;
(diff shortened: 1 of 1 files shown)


iamluciano:
I signed CLA.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -218,4 +222,16 @@ private void registerJavaScriptChannelNames(List<String> channelNames) {
...
   }
+
+  private void enableGeolocation() {
+    webView.getSettings().setGeolocationEnabled(true);
+    webView.setWebChromeClient(
+        new WebChromeClient() {
+          @Override
+          public void onGeolocationPermissionsShowPrompt(
+              String origin, GeolocationPermissions.Callback callback) {
+            callback.invoke(origin, true, false);
```
amirh: We shouldn't do that, the developer should control which sites can access geolocation and which sites cannot.

iamluciano:
Thanks for reviewing. I agree this is not an optimal, clean solution. I forgot to mention the goal of this quick hack is to match current iOS behavior until a proper solution is developed. Both platforms currently behave differently regarding W3C geolocation API. Keep the great work 😄👍🏼

samerdernaika:
plz any update on this

ycv005:
This advance webview plugin - https://pub.dev/packages/flutter_inappwebview has the feature of the geo location in-built. You could use its' `androidOnGeolocationPermissionsShowPrompt` property to enable it. Implementation- https://github.com/pichillilorenzo/flutter_inappwebview/issues/491#issuecomment-675668165

dangtienngoc:
The issue only on Android.

It working well on React native Android https://github.com/react-native-webview/react-native-webview

But not sure why not working on Flutter web view Android

PR #1333, opened 2019-03-12
[firebase_auth] fixes Android linkWithCredential

ened:
This method call has been incorrect and needed an update to use the correct parameters as passed from the Dart side.

Tests are needed soon..

@dnfield Could you help to review, please?


dnfield:
Linter wants you to reformat this.  I think this looks good but it'd be best to have @mklim @collinjackson @kroikie or @amirh to check it out as they're more involved in these at the moment.

collinjackson:
I would ideally like to have a regression test for this but I'm still figuring out the best way to do integration testing for this particular API. I don't want to hold back the fix so I'll land it as is.

PR #1335, opened 2019-03-13
[webview_flutter] Add a page loaded callback

bparrishMines:
Fixes https://github.com/flutter/flutter/issues/27910

### DIFF
--- a/packages/webview_flutter/example/lib/main.dart
+++ b/packages/webview_flutter/example/lib/main.dart
@@ -62,9 +62,7 @@ class WebViewExample extends StatelessWidget {
             return NavigationDecision.navigate;
           },
           onPageFinished: (String url) {
-            Scaffold.of(context).showSnackBar(const SnackBar(
-              content: Text('Page finished loading!'),
-            ));
+            print('Page finished loading: $url');
           },
         );
       }),
--- a/packages/webview_flutter/example/lib/main.dart
+++ b/packages/webview_flutter/example/lib/main.dart
@@ -61,7 +61,7 @@ class WebViewExample extends StatelessWidget {
             print('allowing navigation to $request');
             return NavigationDecision.navigate;
           },
-          onPageLoaded: (String url) {
+          onPageFinished: (String url) {
             Scaffold.of(context).showSnackBar(const SnackBar(
               content: Text('Page finished loading!'),
             ));
--- a/packages/webview_flutter/example/lib/main.dart
+++ b/packages/webview_flutter/example/lib/main.dart
@@ -43,7 +43,7 @@ class WebViewExample extends StatelessWidget {
       // to allow calling Scaffold.of(context) so we can show a snackbar.
       body: Builder(builder: (BuildContext context) {
         return WebView(
-          initialUrl: 'https://flutter.io',
+          initialUrl: 'https://flutter.dev',
           javascriptMode: JavascriptMode.unrestricted,
           onWebViewCreated: (WebViewController webViewController) {
             _controller.complete(webViewController);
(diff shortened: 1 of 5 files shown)


bparrishMines:
@amirh Do you think the name `onPageLoaded` or `onPageFinished` is more appropriate. [Android](https://developer.android.com/reference/android/webkit/WebViewClient.html#onPageFinished(android.webkit.WebView,%20java.lang.String)) and [iOS](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455629-webview?language=objc) both use the word **finished**.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -189,6 +193,11 @@ class WebView extends StatefulWidget {
   ///     * When a navigationDelegate is set HTTP requests do not include the HTTP referer header.
   final NavigationDelegate navigationDelegate;
 
+  /// A function that receives notification when a page has finished loading.
```
amirh: uber duber nit: a function is called/invoked not receives notifications. I'd replace "A function that receives notification" with "Invoked when"

amirh:
I don't have any preference for one of `onPageLoaded`,`onPageFinished` over the other.

packages/webview_flutter/example/lib/main.dart
```diff
@@ -61,6 +61,11 @@ class WebViewExample extends StatelessWidget {
             print('allowing navigation to $request');
             return NavigationDecision.navigate;
           },
+          onPageFinished: (String url) {
+            Scaffold.of(context).showSnackBar(const SnackBar(
+              content: Text('Page finished loading!'),
```
amirh: nit: this might be a little annoying? maybe just print? (probably worth printing the URL as well)

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -189,6 +193,16 @@ class WebView extends StatefulWidget {
   ///     * When a navigationDelegate is set HTTP requests do not include the HTTP referer header.
   final NavigationDelegate navigationDelegate;
 
+  /// Invoked when a page has finished loading.
+  ///
+  /// This is invoked only for the main frame.
+  ///
+  /// When [onPageFinished] is invoked, the page being rendered may not be
```
amirh: Are you sure that it may not be updated yet? (what does updated mean in this context?) if not I'd just omit this part.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -189,6 +193,16 @@ class WebView extends StatefulWidget {
...
+  /// updated yet. However, any Javascript code that is embedded directly in the
+  /// HTML has been loaded and code injected with
+  /// `WebViewController.evaluateJavascript` can assume this.
```
amirh: Include a doc link

packages/webview_flutter/lib/webview_flutter.dart
bparrishMines: In the [WebViewClient](https://developer.android.com/reference/android/webkit/WebViewClient#onPageFinished(android.webkit.WebView,%20java.lang.String)) docs this line is included for onPageFinished

```
This method is called only for main frame. When onPageFinished() is called, the rendering picture may not be updated yet.
```

"Picture" being the canvas the WebView is drawn on. I may need to include that this applies to Android. There's no reference to this in iOS.

hpoul:
hmm.. what magic wand did you have to get the attention of @amirh within 2 days.. waiting for 3 weeks for any signs of life https://github.com/flutter/plugins/pull/1283 .. just tell me to leave this repo alone if you don't like external PR's and i'll happily stop annoying anyone here...

PR #1342, opened 2019-03-15
add driver test command to cirrus

collinjackson:
Fixes flutter/flutter#26080

This teaches Cirrus to run the new drive-examples command added in flutter/plugin_tools#29 on iOS and Android simulators.

I'm sharding the plugins tests more and requesting more powerful Linux VMs to ensure that the tests don't time out now that they're running simulators.


.cirrus.yml
```diff
@@ -22,11 +22,18 @@ task:
...
     - name: analyze
       script: ./script/incremental_build.sh analyze
-    - name: build-apks+java-test
+    - name: build-apks+java-test+drive-examples
       env:
         matrix:
           PLUGIN_SHARDING: "--shardIndex 0 --shardCount 2"
           PLUGIN_SHARDING: "--shardIndex 1 --shardCount 2"
+      create_device_script:
+        echo no | avdmanager create avd --force \
```
amirh: Is this slow? is so maybe worth keeping it as part of the docker image?

collinjackson: According to https://medium.com/cirruslabs/speeding-up-android-builds-with-cirrus-ci-9e20ab7c5150 it's pretty fast but also I don't see a downside to putting it in the Dockerfile so I'll try that

amirh: If it's fast probably simpler to keep it here

packages/shared_preferences/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+[test change to make bots run the tests]
```
collinjackson: will be reverted prior to merging

.cirrus.yml
```diff
@@ -22,11 +22,18 @@ task:
       test_script: ./script/incremental_build.sh test
     - name: analyze
       script: ./script/incremental_build.sh analyze
-    - name: build-apks+java-test
+    - name: build-apks+java-test+drive-examples
       env:
         matrix:
           PLUGIN_SHARDING: "--shardIndex 0 --shardCount 2"
           PLUGIN_SHARDING: "--shardIndex 1 --shardCount 2"
+          PLUGIN_SHARDING: "--shardIndex 2 --shardCount 2"
```
amirh: should shardCount say 4?

collinjackson: good catch, I'll see if I can get the tests to pass with just 2 shards and if not increase to 4

PR #1345, opened 2019-03-15
[cloud_firestore] fix NoSuchMethodError regression and add test

collinjackson:
Fixes a `NoSuchMethodError` in getDocuments(), introduced in #1339

Adds a simple driver test for the fix.

Fixes flutter/flutter#29423

### DIFF
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.9.7
 
-* Fixes an iOS regression in getDocuments introduced in 0.9.6.
+* Fixes a NoSuchMethodError when using getDocuments on iOS (introduced in 0.9.6).
 * Adds a driver test for getDocuments.
 
 ## 0.9.6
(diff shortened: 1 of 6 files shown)


packages/cloud_firestore/example/test_driver/cloud_firestore.dart
```diff
@@ -0,0 +1,35 @@
...
+  group('$Firestore', () {
+    Firestore firestore;
+
+    setUp(() async {
+      final FirebaseApp app = await FirebaseApp.configure(
+        name: 'test',
+        options: const FirebaseOptions(
+          googleAppID: '1:79601577497:ios:5f2bcc6ba8cecddd',
+          gcmSenderID: '79601577497',
+          apiKey: 'AIzaSyArgmRGfB5kiQT6CunAOmKRVKEsxKmy6YI-G72PVU',
```
amirh: Is this a real API key that anyone can use?

packages/cloud_firestore/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.9.7
+
+* Fixes an iOS regression in getDocuments introduced in 0.9.6.
```
amirh: Would be useful to describe what was the regression (both here and in the PR description)

packages/cloud_firestore/example/test_driver/cloud_firestore.dart
collinjackson: Yes, it's the same credentials used by the example app. We can lock down the permissions on the server side.

PR #1350, opened 2019-03-16
[google_maps_flutter] add My location button enabled option

pepegich:
fixed: https://github.com/flutter/flutter/issues/27492


marinat:
when it will be added?

KlausJokinen:
I've tested this and for me the code won't work, the location button is still visible.

pepegich:
@KlausJokinen you need set myLocationEnabled = true, without it MyLocationButton not toggle

KlausJokinen:
> @KlausJokinen you need set myLocationEnabled = true, without it MyLocationButton not toggle

Thanks, The example works fine but with my own app the  myLocationButtonEnabled: false, wont do anything.

iskakaushik:
@pepegich I've taken the liberty to add a few tests, fix merge conflicts and improve the documentation. As usual, thanks for your excellent contribution.

@KlausJokinen, can you give me a sample of your app. I tried this out on both my iPhone and Android. this seems to work fine for me. I've also added an integration test for the same.

Let me know if you are still having issues with it.

annakrih:
Hi, I just got these changes on my project to disable the locationButton. It works great on iOS but it doesn't work at all on Android. On Android I don't see any change in wether the value is true or false. 

I'm working on a Mac and using a Nexus 5X emulator for Android, but I've also tried it on a Windows machine with a Nexus emulator. 

Can you look into this? what other information can I give you?

KlausJokinen:
@annakrih You need to crete variable ex. _enableutton and assign it to myLocationButton parameter. After the Map has been created you set the _enableButton to false and trigger setState.

annakrih:
@KlausJokinen Thanks that worked! 

It does feel a bit awkward to do it like this though, hope this can be handled on the Android side in the future.

iskakaushik:
@KlausJokinen : Sorry about that, let me double check this for Android. I'll update the progress at: https://github.com/flutter/flutter/issues/31930

PR #1351, opened 2019-03-16
[google_maps_flutter] call result.success in markers#update method on android

pepegich:
add missing call result.success on android


iskakaushik:
This should be fixed now.

PR #1352, opened 2019-03-16
[google_maps_flutter] Add method getVisibleRegion

pepegich:
fixed: https://github.com/flutter/flutter/issues/25755


erf:
You don't think we should rather get a `Projection` object with this method ( as available for iOS and Android ). So we can also translate generic points to/from geographic coordinates to screen coordinates.

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -131,6 +131,13 @@ private static Object toJson(LatLng latLng) {
     return Arrays.asList(latLng.latitude, latLng.longitude);
   }
 
+  public static Object toJson(LatLngBounds latLngBounds) {
+    final Map<String, Object> arguments = new HashMap<>(2);
+    arguments.put("southwest", toJson(latLngBounds.southwest));
+    arguments.put("northeast", toJson(latLngBounds.southwest));
```
iskakaushik: this should be `latLngBounds.northeast`

packages/google_maps_flutter/example/lib/map_ui.dart
```diff
@@ -31,6 +31,8 @@ class MapUiBody extends StatefulWidget {
 class MapUiBodyState extends State<MapUiBody> {
   MapUiBodyState();
 
+  GoogleMapController mapController;
```
iskakaushik: could you create a new page and add it to `main.dart`, i'd like to leave this one without having a controller. Also that way we can maybe get around to having a text field that gets populated with the lat lng bounds on button press rather than an alert dialog. What do you think?

packages/google_maps_flutter/lib/src/controller.dart
```diff
@@ -129,4 +129,17 @@ class GoogleMapController {
       'cameraUpdate': cameraUpdate._toJson(),
     });
   }
+
+  /// Return [LatLngBounds] defining the polygon that is visible in a map.
```
iskakaushik: nit: `polygon` -> `region`

iskakaushik:
@erf , as I understand it based on [this](https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_projection), seems like projection is a useful utility to have, i think there was a PR (which I will track down later) which had some geo utils and I also vaguely remember that it had some utilities around projections. Maybe that could solve most use cases.

erf:
@iskakaushik both [ios](https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_projection) and the [android](https://developers.google.com/android/reference/com/google/android/gms/maps/Projection) SDK have a `Projection` object which does conversion between lat lng and screen coordinates in addition to returning a `visibleRegion`. I think mapping all of these functions using their native SDK, through `platform channels` and a `Projection` object would be a good idea, keeping it close to the original implementation.

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
pepegich: done

packages/google_maps_flutter/example/lib/map_ui.dart
pepegich: done

packages/google_maps_flutter/lib/src/controller.dart
pepegich: done

pepegich:
@iskakaushik I added tests

packages/google_maps_flutter/example/test_driver/google_maps.dart
```diff
@@ -57,4 +57,43 @@ void main() {
     compassEnabled = await inspector.isCompassEnabled();
     expect(compassEnabled, true);
   });
+
+  test('testGetVisibleRegion', () async {
```
iskakaushik: I think we can make this test stronger by making sure the bounds are what we expect. I made a few changes at https://github.com/flutter/plugins/commit/36ad3891bdd31284a10112a59083626e3693e457 to make this test slightly better, if you agree with them, please make those changes.

pepegich: test with your changes not passed on my pixel 2 and nvidia shiesl tablet. seems the reason for the different screen. because of this I wrote not stronger test.

iskakaushik: Oh, interesting, I tried to make the tests unaware of the resolution. I made a mistake in the last 2 lines of the test, mind changing them to:

```
    expect(secondVisibleRegion.contains(southWest), isTrue);
    expect(secondVisibleRegion.contains(northEast), isTrue);
```

If they still fail. Could you paste the errors here? I can help you out.

pepegich: still error on line
expect(secondVisibleRegion.contains2(northEast), isTrue);. actual false

iskakaushik: I'm getting this error on Pixel 3 as well. I'm now looking at what is going on.

iskakaushik: Oops, I had illegal value for the latLng bounds before. They were outside of earth 😞

These should work well, could you please try these:

```
const LatLng southWest = LatLng(60, 75);
const LatLng northEast = LatLng(65, 80);
```

packages/google_maps_flutter/example/lib/map_data_ui.dart
```diff
@@ -0,0 +1,98 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+import 'package:google_maps_flutter/google_maps_flutter.dart';
+import 'package:google_maps_flutter_example/page.dart';
```
iskakaushik: you want to import this as `import 'page.dart';`

packages/google_maps_flutter/example/lib/main.dart
```diff
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 import 'package:flutter/material.dart';
+import 'package:google_maps_flutter_example/map_data_ui.dart';
```
iskakaushik: you want to import this as `import 'map_data_ui.dart'`

packages/google_maps_flutter/example/test_driver/google_maps.dart
pepegich: very strange behavior, to test passed on tablet with portrait and landscape orientations need add next line, it is need for portrait orientation:
`    await Future<dynamic>.delayed(Duration(seconds: 3));`
after 
`    final GoogleMapController mapController =
    await mapControllerCompleter.future;`

and change line to, it is need for landscape orientation:
`    await mapController
        .moveCamera(CameraUpdate.newLatLngBounds(latLngBounds, 1));`

may be it specifically only for my device?

iskakaushik: can you push a commit with the latest changes, I will try to run it and see if I can reproduce this. Thanks for your patience @pepegich

packages/google_maps_flutter/example/lib/map_data_ui.dart
pepegich: done

packages/google_maps_flutter/example/lib/main.dart
pepegich: done

packages/google_maps_flutter/example/test_driver/google_maps.dart
pepegich: I pushed. I tested on another tablet Lenovo TAB 2 A10 70L. and test not passed without above workaround.

mehdok:
@iskakaushik, Any news on `Projection` methods?

durduliu2009:
Hello guys ? Some news about Projection ? :) Many thanks in advance

PR #1353, opened 2019-03-17
[firebase_messaging] Update example

pepegich:
seems after Breaking Change need update example.


amervelic:
How to set custom sound on Oreo and later

PR #1355, opened 2019-03-18
[firebase_storage] Fix putFile method's Content-Type auto-detection for iOS

mono0926:
- `putFile` should use `putFile` iOS SDK method correctly
- `putFile` should auto-detect Content-Type automatically by file extension if not specified

I tested at this sample on **iOS**:
https://github.com/flutter/plugins/blob/master/packages/firebase_storage/example/lib/main.dart.

This uses `putFile` here:
https://github.com/flutter/plugins/blob/04d6f52e3b09193462a18a38a7249da92b901ba5/packages/firebase_storage/example/lib/main.dart#L66

## Before this pull request

Uploaded file:
https://firebasestorage.googleapis.com/v0/b/flutter-firebase-plugins.appspot.com/o/text%2Ffoob3d87700-3260-11e9-d838-d7a1f7b2a3d4.txt

`contentType` is `application/octet-stream`, but this is unintended value because `putFile` should auto-detect `contentType` as written this document.

> The putFile: method automatically infers the content type from the NSURL filename extension, but you can override the auto-detected type by specifying contentType in the metadata. If you do not provide a contentType and Cloud Storage cannot infer a default from the file extension, Cloud Storage uses application/octet-stream.
> https://firebase.google.com/docs/storage/ios/upload-files#add_file_metadata

In this case, the file's extension is `txt`, so `contentType` should be `plain/text`.

### After this pull request

Uploaded file:
https://firebasestorage.googleapis.com/v0/b/flutter-firebase-plugins.appspot.com/o/text%2Ffoo09aa2e30-3260-11e9-efe1-e3ed438ea03f.txt

`contentType` is `plain/text`, which is intended value.

### DIFF
--- a/packages/firebase_storage/CHANGELOG.md
+++ b/packages/firebase_storage/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.1.2
+## 2.1.1+2
 
 * On iOS, use `putFile` instead of `putData` appropriately to detect `Content-Type`.
 
--- a/packages/firebase_storage/CHANGELOG.md
+++ b/packages/firebase_storage/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.1.2
+
+* On iOS, use `putFile` instead of `putData` appropriately to detect `Content-Type`.
+
 ## 2.1.1+1
 
 * On iOS, gracefully handle the case of uploading a nonexistent file without crashing.
--- a/packages/firebase_storage/example/test_driver/firebase_storage.dart
+++ b/packages/firebase_storage/example/test_driver/firebase_storage.dart
@@ -44,6 +44,7 @@ void main() {
       final String url = await ref.getDownloadURL();
       final http.Response downloadData = await http.get(url);
       expect(downloadData.body, kTestString);
+      expect(downloadData.headers['content-type'], 'text/plain');
       final File tempFile = File('${systemTempDir.path}/tmp$uuid.txt');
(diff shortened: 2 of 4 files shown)


mono0926:
This is Android's uploaded file URL:
https://firebasestorage.googleapis.com/v0/b/flutter-firebase-plugins.appspot.com/o/text%2Ffoo1c94d560-3260-11e9-8a4d-ff3d6d715989.txt

`contentType` is `application/octet-stream`, but I do not know why this happens.

`putFile` seems to use `putFile` correctly 🤔 
https://github.com/flutter/plugins/blob/04d6f52e3b09193462a18a38a7249da92b901ba5/packages/firebase_storage/android/src/main/java/io/flutter/plugins/firebase/storage/FirebaseStoragePlugin.java#L258-L272

mono0926:
Sorry, I'll fix format.

mono0926:
@collinjackson 

I've added commits:

> Could you please write a CHANGELOG and update pubspec.yaml for release?

https://github.com/flutter/plugins/pull/1355/commits/1a1563b84856dab4435ad0b39a136e51e3172c05

> It looks like it might be possible to update the integration test to include the test for the correct behavior here. Would you be willing to add that?

https://github.com/flutter/plugins/pull/1355/commits/778d6a605a8101270c6a908c9673d9fb1a997f15

Could you check that please?

collinjackson:
Thanks for the contribution!

PR #1356, opened 2019-03-18
[url_launcher]ios whose view is not in the window hierarchy!

LDXu:
ios error whose view is not in the window hierarchy!find presentedViewController and use it

### DIFF
--- a/packages/url_launcher/ios/Classes/UrlLauncherPlugin.m
+++ b/packages/url_launcher/ios/Classes/UrlLauncherPlugin.m
@@ -63,8 +63,6 @@ API_AVAILABLE(ios(9.0))
 
 @interface FLTUrlLauncherPlugin ()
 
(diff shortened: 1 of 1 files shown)


cyanglaz:
Could you explain the issue that this PR is trying to fix in more detail? Or link this PR to an issue ticket?

ashkryab:
when Navigator.push called new page openned under th webView (safari controller)... launched url didnt close

cyanglaz:
@ashkryab Do you want to explain the detail in an issue ticket?

stuartmorgan:
Closing since it's still not clear what issue this is trying to address. If you still want to move forward with this, please file an issue clearly describing the issue, with a reproducible sample, then open a PR referencing that issue. We would also need tests associated with the PR. Thanks!

PR #1360, opened 2019-03-19
[webview_flutter] Create WebView client depending on version and hasDelegate

bparrishMines:
Fixes https://github.com/flutter/flutter/issues/29446

This changes handles three situations
1. The WebView doesn't have a NavigationDelegate. We use a [WebViewClient](https://developer.android.com/reference/android/webkit/WebViewClient).

2. The WebView has a NavigationDelegate and Android version is > 24. We use a [WebViewClient](https://developer.android.com/reference/android/webkit/WebViewClient).

3. The WebView has a NavigationDelegate and Android version is < 24. We use a [WebViewClientCompat](https://developer.android.com/reference/androidx/webkit/WebViewClientCompat). On top of that, the method

```
public boolean shouldOverrideUrlLoading (WebView view, WebResourceRequest request)
```

should only work in api 21+. [WebResourceRequest](https://developer.android.com/reference/android/webkit/WebResourceRequest) and [SHOULD_OVERRIDE_WITH_REDIRECTS](https://developer.android.com/reference/androidx/webkit/WebViewFeature.html#SHOULD_OVERRIDE_WITH_REDIRECTS)


sky126:
Great

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
```diff
@@ -32,8 +34,7 @@ void setHasNavigationDelegate(boolean hasNavigationDelegate) {
   }
```
amirh: This can be called after the initial call to createWebViewClient (e.g the app starts without a navigation delegate and later rebuilds the webview with a navigation delegate)

bparrishMines: I was wondering about this. I thought the `WebView` would be a constant widget, but realized that platform view do rebuild on state changes. I'll update the PR.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -41,7 +41,7 @@
       registerJavaScriptChannelNames((List<String>) params.get(JS_CHANNEL_NAMES_FIELD));
     }
 
-    webView.setWebViewClient(flutterWebViewClient.createWebViewClient());
+    webView.setWebViewClient(flutterWebViewClient.createWebViewClient(false));
```
amirh: applySetting is called before this, so if we started with a navigation delegate this will remove it.
Can we remove this line and rely on applySettings to set it?

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -186,7 +184,8 @@ private void applySettings(Map<String, Object> settings) {
           updateJsMode((Integer) settings.get(key));
           break;
         case "hasNavigationDelegate":
-          flutterWebViewClient.setHasNavigationDelegate((boolean) settings.get(key));
+          boolean hasNavigationDelegate = (boolean) settings.get(key);
+          webView.setWebViewClient(flutterWebViewClient.createWebViewClient(hasNavigationDelegate));
```
amirh: nit: extra line

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
```diff
@@ -90,6 +85,53 @@ private void notifyOnNavigationRequest(
     }
   }
 
+  // This method attempts to avoid using WebViewClientCompat due to bug
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=925887
```
amirh: Let's also reference https://github.com/flutter/flutter/issues/29446

PR #1364, opened 2019-03-20
[firebase_ml_vision] Update ImageDetector for iOS

bparrishMines:
Fixes https://github.com/flutter/flutter/issues/23259 for iOS

Android Companion: https://github.com/flutter/plugins/pull/1367

90% of this is name changing. I included comments explaining name changing of files.


packages/firebase_ml_vision/ios/Classes/ImageLabeler.m
```diff
@@ -0,0 +1,57 @@
+#import "FirebaseMlVisionPlugin.h"
```
bparrishMines: This file was renamed to `ImageLabeler.m` changes can be seen [here](https://github.com/flutter/plugins/pull/1364/commits/c9c42d8175dd30f21243d926165af466e887f590).

packages/firebase_ml_vision/lib/src/image_labeler.dart
```diff
@@ -0,0 +1,127 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
```
bparrishMines: This file was renamed from `label_detector.dart` changes can be seen [here](https://github.com/flutter/plugins/pull/1364/commits/c9c42d8175dd30f21243d926165af466e887f590).

packages/firebase_ml_vision/test/firebase_ml_vision_test.dart
```diff
@@ -625,175 +621,6 @@ void main() {
       });
     });
 
-    group('$LabelDetector', () {
```
bparrishMines: The code below was moved to its own file `image_labeler_test.dart` changes can be seen [here](https://github.com/flutter/plugins/pull/1364/commits/c9c42d8175dd30f21243d926165af466e887f590).

packages/firebase_ml_vision/test/image_labeler_test.dart
```diff
@@ -0,0 +1,165 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
bparrishMines: The code below was moved to its own file from `firebase_ml_vision_test.dart` changes can be seen [here](https://github.com/flutter/plugins/pull/1364/commits/c9c42d8175dd30f21243d926165af466e887f590).

packages/firebase_ml_vision/example/ios/Podfile
```diff
@@ -0,0 +1,68 @@
+# Uncomment this line to define a global platform for your project
+platform :ios, '9.0'
```
bparrishMines: Everything except this line is generated in this `PodFile`.

packages/firebase_ml_vision/ios/Classes/ImageLabeler.m
collinjackson: I would probably capitalize the L in MLVision

bparrishMines:
@collinjackson Changing the 'L' in the header name would require adding another filename change to this PR. And we would have to change it on the Android side too since code is generated based on the name we provide in the plugin [pubspec.yaml](https://github.com/flutter/plugins/blob/master/packages/firebase_ml_vision/pubspec.yaml#L21). It may not be worth doing in this PR.

collinjackson:
You're right -- the capitalization issue is pre-existing and I agree that it shouldn't block merging this. Thank you.

kuprel:
Great, I was able to get this to install on iOS by pointing pubspec.yaml to a local directory.  Can the dart pub be updated to 0.7.0 too?

collinjackson:
@kuprel 
It looks like version 0.7.0 has been published to pub. https://pub.dartlang.org/packages/firebase_ml_vision

kuprel:
Saw that, thanks!

PR #1365, opened 2019-03-20
Template type invokemethod

cyanglaz:
Bump min flutter version to 1.2.0 for most plugins.
Add template type parameter for `invokeMethod` calls.
Updated invokeMethod and invokeMethod to using invokeListMethod and invokeMapMethod. Reopen the PR.

Anything that will change the type of a public API should not be included in this PR. 

`Sensor` plugin has not been changed since no `invokeMethod` is called in the plugin.
The plugin version of `in app purchase` has not be updated because it is still under development. 

Need to publish all plugins after this PR is merged, except `in_app_purchase` and `sensor`.

### DIFF
--- a/packages/android_alarm_manager/lib/android_alarm_manager.dart
+++ b/packages/android_alarm_manager/lib/android_alarm_manager.dart
@@ -43,10 +43,7 @@ void _alarmManagerCallbackDispatcher() {
 
   // Once we've finished initializing, let the native portion of the plugin
   // know that it can start scheduling alarms.
-  // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-  // https://github.com/flutter/flutter/issues/26431
-  // ignore: strong_mode_implicit_dynamic_method
-  _channel.invokeMethod('AlarmService.initialized');
+  _channel.invokeMethod<void>('AlarmService.initialized');
 }
 
 /// A Flutter plugin for registering Dart callbacks with the Android
@@ -70,10 +67,7 @@ class AndroidAlarmManager {
       return false;
     }
     final dynamic r = await _channel
-        // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-        // https://github.com/flutter/flutter/issues/26431
-        // ignore: strong_mode_implicit_dynamic_method
-        .invokeMethod('AlarmService.start', <dynamic>[handle.toRawHandle()]);
+        .invokeMethod<dynamic>('AlarmService.start', <dynamic>[handle.toRawHandle()]);
     return r ?? false;
   }
 
@@ -117,10 +111,7 @@ class AndroidAlarmManager {
     if (handle == null) {
       return false;
     }
-    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-    // https://github.com/flutter/flutter/issues/26431
-    // ignore: strong_mode_implicit_dynamic_method
-    final dynamic r = await _channel.invokeMethod('Alarm.oneShot', <dynamic>[
+    final dynamic r = await _channel.invokeMethod<dynamic>('Alarm.oneShot', <dynamic>[
       id,
       exact,
       wakeup,
@@ -172,10 +163,7 @@ class AndroidAlarmManager {
     if (handle == null) {
       return false;
     }
-    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-    // https://github.com/flutter/flutter/issues/26431
-    // ignore: strong_mode_implicit_dynamic_method
-    final dynamic r = await _channel.invokeMethod('Alarm.periodic', <dynamic>[
+    final dynamic r = await _channel.invokeMethod<dynamic>('Alarm.periodic', <dynamic>[
       id,
       exact,
       wakeup,
@@ -196,10 +184,7 @@ class AndroidAlarmManager {
   /// failure.
   static Future<bool> cancel(int id) async {
     final dynamic r =
-        // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-        // https://github.com/flutter/flutter/issues/26431
-        // ignore: strong_mode_implicit_dynamic_method
-        await _channel.invokeMethod('Alarm.cancel', <dynamic>[id]);
+        await _channel.invokeMethod<dynamic>('Alarm.cancel', <dynamic>[id]);
     return (r == null) ? false : r;
   }
 }
--- a/packages/android_alarm_manager/lib/android_alarm_manager.dart
+++ b/packages/android_alarm_manager/lib/android_alarm_manager.dart
@@ -66,7 +66,8 @@ class AndroidAlarmManager {
     if (handle == null) {
       return false;
     }
-    return _channel.invokeMethod<bool>('AlarmService.start', <dynamic>[handle.toRawHandle()]);
+    final bool r = await _channel.invokeMethod<bool>('AlarmService.start', <dynamic>[handle.toRawHandle()]);
+    return r ?? false;
   }
 
   /// Schedules a one-shot timer to run `callback` after time `delay`.
@@ -109,7 +110,7 @@ class AndroidAlarmManager {
     if (handle == null) {
       return false;
     }
-   return _channel.invokeMethod<bool>('Alarm.oneShot', <dynamic>[
+   final bool r = await _channel.invokeMethod<bool>('Alarm.oneShot', <dynamic>[
       id,
       exact,
       wakeup,
@@ -181,8 +182,7 @@ class AndroidAlarmManager {
   /// Returns a [Future] that resolves to `true` on success and `false` on
   /// failure.
   static Future<bool> cancel(int id) async {
-    final dynamic r =
-        final bool r = await _channel.invokeMethod<bool>('Alarm.cancel', <dynamic>[id]);
+    final bool r = await _channel.invokeMethod<bool>('Alarm.cancel', <dynamic>[id]);
     return (r == null) ? false : r;
   }
 }
--- a/packages/android_alarm_manager/lib/android_alarm_manager.dart
+++ b/packages/android_alarm_manager/lib/android_alarm_manager.dart
@@ -66,9 +66,7 @@ class AndroidAlarmManager {
     if (handle == null) {
       return false;
     }
-    final dynamic r = await _channel
     return _channel.invokeMethod<bool>('AlarmService.start', <dynamic>[handle.toRawHandle()]);
-    return r ?? false;
   }
 
   /// Schedules a one-shot timer to run `callback` after time `delay`.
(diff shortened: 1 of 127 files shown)


bkonyi:
android_alarm_manager changes LGTM.

packages/android_alarm_manager/lib/android_alarm_manager.dart
```diff
@@ -70,10 +67,7 @@ class AndroidAlarmManager {
       return false;
     }
     final dynamic r = await _channel
-        // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-        // https://github.com/flutter/flutter/issues/26431
-        // ignore: strong_mode_implicit_dynamic_method
-        .invokeMethod('AlarmService.start', <dynamic>[handle.toRawHandle()]);
+        .invokeMethod<dynamic>('AlarmService.start', <dynamic>[handle.toRawHandle()]);
```
jonahwilliams:
```suggestion
    return _channel.invokeMethod<bool>('AlarmService.start', <dynamic>[handle.toRawHandle()]);
```

jonahwilliams: Its not necessary to do the explicit cast, that is the point of the invokeMethod parameter

packages/android_alarm_manager/lib/android_alarm_manager.dart
```diff
@@ -117,10 +111,7 @@ class AndroidAlarmManager {
     if (handle == null) {
       return false;
     }
-    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-    // https://github.com/flutter/flutter/issues/26431
-    // ignore: strong_mode_implicit_dynamic_method
-    final dynamic r = await _channel.invokeMethod('Alarm.oneShot', <dynamic>[
+    final dynamic r = await _channel.invokeMethod<dynamic>('Alarm.oneShot', <dynamic>[
```
jonahwilliams:
```suggestion
   return _channel.invokeMethod<bool>('Alarm.oneShot', <dynamic>[
```

packages/android_alarm_manager/lib/android_alarm_manager.dart
```diff
@@ -172,10 +163,7 @@ class AndroidAlarmManager {
     if (handle == null) {
       return false;
     }
-    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-    // https://github.com/flutter/flutter/issues/26431
-    // ignore: strong_mode_implicit_dynamic_method
-    final dynamic r = await _channel.invokeMethod('Alarm.periodic', <dynamic>[
+    final dynamic r = await _channel.invokeMethod<dynamic>('Alarm.periodic', <dynamic>[
```
jonahwilliams:
```suggestion
    return _channel.invokeMethod<bool>('Alarm.periodic', <dynamic>[
```

packages/android_alarm_manager/lib/android_alarm_manager.dart
```diff
@@ -196,10 +184,7 @@ class AndroidAlarmManager {
   /// failure.
   static Future<bool> cancel(int id) async {
     final dynamic r =
-        // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-        // https://github.com/flutter/flutter/issues/26431
-        // ignore: strong_mode_implicit_dynamic_method
-        await _channel.invokeMethod('Alarm.cancel', <dynamic>[id]);
+        await _channel.invokeMethod<dynamic>('Alarm.cancel', <dynamic>[id]);
```
jonahwilliams:
```suggestion
        final bool r = await _channel.invokeMethod<bool>('Alarm.cancel', <dynamic>[id]);
```

packages/camera/lib/camera.dart
```diff
@@ -227,10 +224,7 @@ class CameraController extends ValueNotifier<CameraValue> {
     }
     try {
       _creatingCompleter = Completer<void>();
-      // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-      // https://github.com/flutter/flutter/issues/26431
-      // ignore: strong_mode_implicit_dynamic_method
-      final Map<dynamic, dynamic> reply = await _channel.invokeMethod(
+      final Map<dynamic, dynamic> reply = await _channel.invokeMethod<Map<dynamic, dynamic>>(
```
jonahwilliams: https://github.com/flutter/flutter/blob/5169ab5974571918121fa8a7ddb0d3e0623bc987/packages/flutter/lib/src/services/platform_channel.dart#L329

jonahwilliams:
There are many places where you are using `<dynamic>` and then doing a manual cast - those should be refactored to use `<correctType>`

packages/camera/lib/camera.dart
cyanglaz: cool, going to do a massive change on all the invokeMethod that has map type

cyanglaz:
Closing the PR now for refactoring using invokeListMethod and invokeMapMethod

cyanglaz:
Updated invokeMethod<Map> and invokeMethod<List> to using invokeListMethod and invokeMapMethod. Reopen the PR.

packages/cloud_functions/lib/cloud_functions.dart
```diff
@@ -43,10 +43,7 @@ class CloudFunctions {
       {@required String functionName, Map<String, dynamic> parameters}) async {
```
jonahwilliams: I don't know what this is supposed to return, but you could add a similar feature to the invokeMethod to allow users to use inference instead of casting:

```
Future<T> call(	  Future<dynamic> call<T>(
      {@required String functionName, Map<String, dynamic> parameters}) async {
```

Where if `T` isn't specified it defaults to dynamic. Example:

```
final bool x = cloudFunctions.call(functionName: 'hello'); // <T> inferred as bool
final y = cloudFunctions.call<String>(functionName: 'bar'); // <T> specified as String
```

cyanglaz: SG, I think we can follow this up in a different PR. 
Also, by looking at the implementation, this method should return Future<Map<String, dynamic>>, however it would be breaking change since call is a public method. I would avoid that in this PR. And after this is merged and if we think it is necessary to break, then we can create another PR to fix this.
There are other places that has public API returning <dynamic> which I am also avoiding in this PR.

packages/camera/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.4.3+1
+
+* Bump the minimum flutter version to 1.2.0.
+* Add Template type parameter to `invokeMethod` calls.
+
+
```
amirh: nit: extra empty line here

packages/cloud_firestore/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
-## 0.9.7
+## 0.9.6+1
```
amirh: this should be 7+1

packages/android_alarm_manager/lib/android_alarm_manager.dart
```diff
@@ -43,10 +43,7 @@ void _alarmManagerCallbackDispatcher() {
 
   // Once we've finished initializing, let the native portion of the plugin
   // know that it can start scheduling alarms.
-  // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-  // https://github.com/flutter/flutter/issues/26431
-  // ignore: strong_mode_implicit_dynamic_method
-  _channel.invokeMethod('AlarmService.initialized');
+  _channel.invokeMethod<void>('AlarmService.initialized');
```
amirh: This should be bool following the fix in https://github.com/flutter/plugins/pull/1384

packages/firebase_auth/lib/src/firebase_user.dart
```diff
@@ -190,10 +166,7 @@ class FirebaseUser extends UserInfo {
   ///   • `ERROR_REQUIRES_RECENT_LOGIN` - If the user's last sign-in time does not meet the security threshold. Use reauthenticate methods to resolve.
   Future<void> unlinkFromProvider(String provider) async {
     assert(provider != null);
-    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-    // https://github.com/flutter/flutter/issues/26431
-    // ignore: strong_mode_implicit_dynamic_method
-    return await FirebaseAuth.channel.invokeMethod(
+    return await FirebaseAuth.channel.invokeMethod<void>(
```
amirh: This seems to return a `Map<String, Object>`:
https://github.com/flutter/plugins/blob/04033879181f284cf7511db6067048e08a792e3e/packages/firebase_auth/android/src/main/java/io/flutter/plugins/firebaseauth/FirebaseAuthPlugin.java#L455

packages/firebase_auth/pubspec.yaml
```diff
@@ -29,4 +29,4 @@ dev_dependencies:
 
 environment:
   sdk: ">=2.0.0-dev.28.0 <3.0.0"
-  flutter: ">=0.1.4 <2.0.0"
+  flutter: ">=1.2.0 <2.0.0"
```
amirh: I got so far

packages/firebase_auth/lib/src/firebase_auth.dart
```diff
@@ -126,14 +116,10 @@ class FirebaseAuth {
     @required String email,
   }) async {
     assert(email != null);
-    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-    // https://github.com/flutter/flutter/issues/26431
-    // ignore: strong_mode_implicit_dynamic_method
-    final List<dynamic> providers = await channel.invokeMethod(
+    return await channel.invokeListMethod<String>(
```
bparrishMines: Wouldn't [invokeListMethod](https://docs.flutter.io/flutter/services/MethodChannel/invokeListMethod.html) crash when `result.cast<T>();` is called and `providers == null`.

This makes me wonder if we should use `invokeListMethod` and `invokeMapMethod` in this PR because in some cases we want to return null values.

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
```diff
@@ -23,11 +23,9 @@ class FirebaseDynamicLinks {
   /// there is no pending dynamic link or any call to this method after the
   /// the first attempt.
   Future<PendingDynamicLinkData> retrieveDynamicLink() async {
-    final Map<dynamic, dynamic> linkData =
-        // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-        // https://github.com/flutter/flutter/issues/26431
-        // ignore: strong_mode_implicit_dynamic_method
-        await channel.invokeMethod('FirebaseDynamicLinks#retrieveDynamicLink');
+    final Map<String, dynamic> linkData =
+        await channel.invokeMapMethod<String, dynamic>(
```
bparrishMines: This may crash due to the cast call as well. `linkData` could equal `null`.

cyanglaz:
I am going to close this one and create different PRs for each plugin. I feel like much safer that way and easier to review.

PR #1366, opened 2019-03-20
Update the controller's onPageFinished reference on widget rebuild.

amirh:


amirh:
This landed as part of https://github.com/flutter/plugins/pull/1369

PR #1367, opened 2019-03-20
[firebase_ml_vision] Android side of upgrade to new ImageLabeler

bparrishMines:
Android Companion to: https://github.com/flutter/plugins/pull/1364

90% of this is name changing. I included comments explaining name changing of files.


packages/firebase_ml_vision/android/src/main/java/io/flutter/plugins/firebasemlvision/ImageLabeler.java
```diff
@@ -0,0 +1,90 @@
+package io.flutter.plugins.firebasemlvision;
```
bparrishMines: This was a name change from `LabelDetector.java`. Changes can be seen [here](https://github.com/flutter/plugins/pull/1367/commits/d7e0bb819c779ad573760a8b1ae22138d9802835).

PR #1368, opened 2019-03-21
[firebase_crashlytics] use static framework

iqbalmineraltown:
similar to other firebase plugins: need to be set as static framework in podspec


iqbalmineraltown:
already fixed in 0.0.2

PR #1370, opened 2019-03-21
[camera] Pause/resume video recording for Android

huulbaek:
I made this for a small project that needed the functionality. I'm not well versed in the plugin structure and JAVA, so if it's completely off, let me know. Also, I would implement it for iOS too, but it doesn't seem that straightforward.


huulbaek:
I signed it!

packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -392,6 +392,34 @@ public void stopVideoRecording(@NonNull final Result result) {
...
 
+  public void pauseVideoRecording(@NonNull final Result result) {
+    if (!recordingVideo) {
+      result.success(null);
+      return;
+    }
+
+    try {
+      mediaRecorder.pause();
+      result.success(null);
```
mklim: nit: This should be outside of the `try` block here and below, just in the weird offhand chance calling this throws an `IllegalStateException`. If it did `result.error` would also throw in the `catch` block.

PR #1373, opened 2019-03-21
[shared_preferences] Add contains method

pepegich:
fixed https://github.com/flutter/flutter/issues/16330

### DIFF
--- a/packages/shared_preferences/lib/shared_preferences.dart
+++ b/packages/shared_preferences/lib/shared_preferences.dart
@@ -70,6 +70,9 @@ class SharedPreferences {
   /// String.
   String getString(String key) => _preferenceCache[key];
 
+  /// Returns true if persistent storage the contains the given [key].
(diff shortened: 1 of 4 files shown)


packages/shared_preferences/lib/shared_preferences.dart
```diff
@@ -70,6 +70,9 @@ class SharedPreferences {
   /// String.
   String getString(String key) => _preferenceCache[key];
 
+  /// Returns true if persistent storage the contains the given [key].
+  bool contains(String key) => _preferenceCache.containsKey(key);
```
collinjackson: I think this should be named `containsKey` for consistency with the `Map` API. (In retrospect we probably should have just make `SharedPreferences` implement `Map` but that's out of scope for this change.)

pepegich: yuo right! done!

PR #1384, opened 2019-03-22
Send success result for `AlarmService.initialized` method call

bkonyi:


bkonyi:
> LGTM, thanks.
> 
> Make sure to update the changelog and bump the version before submitting.

I have some other changes for this plugin that should be going in shortly, so I'll hold off bumping the version for now and include this change in that release, particularly since this change shouldn't actually change any behavior for users.

PR #866, opened 2018-10-23
[image_picker] Fix to keep Exif and so on in case of iOS.

ko2ic:
## Description

The image that can be acquired with image_picker is missing some information, even in the case of the original size.

1. ```CreateDate``` of Exif.
2. ```GPSLatitude```,  ```GPSLatitude```, ```GPSLongitudeRef ```, ```GPSLongitude``` of GPS Tags.
3. Animation of GIF
4. PNG transparency (It solves with https://github.com/flutter/plugins/pull/742, but the solution method is different )

This pull request is a modification of these.

Limitation

* The Image of tiff and Unknown magic byte draws as JPEG. (But this has the same problem with the current version.) 

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [ ] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [ ] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

PR #1392, opened 2019-03-25
[image_picker] Fix exif orientation for pictures taken from camera

charafau:
Added ability to automatically rotate picture taken from camera, with exif preservation.

## Description

Problem:

Some phones take pictures with rotation `right, top` which causes photo to "rotate" 90 degrees when exif data is stripped.  This PR adds one variable to automatically rotate picture and write right exif data.

## Related Issues

This fixes this: https://github.com/flutter/flutter/issues/19458

looking forward to suggestions / changes to this PR. Thanks!

### DIFF
--- a/packages/image_picker/CHANGELOG.md
+++ b/packages/image_picker/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.0+10
+
+* Fixed photos rotated on 90 degrees.
+
 ## 0.5.0+9
 
 * Remove unnecessary temp video file path.
--- a/packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
+++ b/packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
@@ -426,8 +426,9 @@ public class ImagePickerDelegate
     if (pendingResult != null) {
       Double maxWidth = methodCall.argument("maxWidth");
       Double maxHeight = methodCall.argument("maxHeight");
+      Boolean rotate = methodCall.argument("rotate");
 
-      String finalImagePath = imageResizer.resizeImageIfNeeded(path, maxWidth, maxHeight);
+      String finalImagePath = imageResizer.resizeImageIfNeeded(path, maxWidth, maxHeight, rotate);
       finishWithSuccess(finalImagePath);
 
       //delete original file if scaled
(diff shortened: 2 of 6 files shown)


Hixie:
I think this needs end-to-end tests to verify that for each physical orientation of the phone, the output image has the correct orientation and EXIF data.

charafau:
@Hixie good idea, will try to add some.

cyanglaz:
@charafau I think what we really should do is to copy all the meta data over to the image, instead of just copying rotation. Also, it is prob not necessary to have it as an optional flag. The captured/picked image/video should have the correct meta data by default.

charafau:
Hello @cyanglaz sorry for lack of movement with this one, I was busy with different plugin. Yes I think you're right, I should copy all exif data. and add tests. Not sure if we want to make it a default behaviour though, I will leave that decision to you ;)

cyanglaz:
@charafau Thank you for the quick response! I was thinking about this yesterday and I think that since the cropped image is actually not the original image, does it make sense to copy the exif from the original image?
Instead, we can also return the original image path from the pickImage/video API along with the cropped image path, that way if you want to access the original image's exif, you can get it from the original image. And the cropped image will contain whatever metadata the iOS or android platform decides put inside the image data. Will that solve the problem?
I think it will be more robust and easier to maintain, since we don't have to do these surgeries on metadata of the image.

charafau:
@cyanglaz sorry for late reply. This code doesn't crop the image, it just rotates it, so it will be displayed correctly in portrait mode, when you strip exif data (some phones mount camera differently).

About your suggestions - yes we could just strip the exif data, but I'm not sure if that's what users would like to do (Imagine photo gallery with location of pictures and camera info). 

Other aspect of it would be, is current API good for it? Meaning, would it be better to pass parameter `rotate` or create new method like `pickRotatedImage` ?

cyanglaz:
@charafau You are right. I have been reading all the issues that people have opened for image picker and I think people do want the exif to be copied over to the new image. Let's do that?
There is a PR doing something similar on the iOS side, for your reference. https://github.com/flutter/plugins/pull/866

Are you able to keep working on this PR?

charafau:
@cyanglaz Yes, I think I should be able to. Might ask for help with tests on gitter but I will make it somehow.

cyanglaz:
@charafau I have a PR related to this ready if you want to take a look.
https://github.com/flutter/plugins/pull/1586#pullrequestreview-235881807

NGhebreial:
I made a plugin to fix this, hoping it's useful for you:
[https://pub.dev/packages/flutter_exif_rotation](https://pub.dev/packages/flutter_exif_rotation)

charafau:
Sorry for slow progress, I'm on vacation now and cannot will be back soon-ish. Plan to do that when I'm back

cyanglaz:
@charafau I'm going to close this PR since it has been inactive for a while. :) I encourage you to open another one when you are ready with the fix!

PR #1406, opened 2019-03-27
[webview_flutter] Add initial e2e tests

amirh:
## Description

Add initial flutter driver based e2e tests for webview_flutter.

@collinjackson a few notes for stuff I ran into while writing this:
  * I'm not using the example app at all other than it's iOS project setup, not a big deal but might make sense to consider supporting such an integration test out of the `example/` folder.
  * I couldn't use the WidgetTester as it is asserting that the platform is flutter_tester, maybe we should change that?
  * Because I couldn't use the widget tester I used `runApp()` directly in each test, while it works it's not starting each test from a clean widget tree state(that's why I included global keys for the widgets to ensure their state is lost). I also experimented with doing `runApp(Container())` in `setUp()` but I think it didn't work as nothing was waiting for the widget tree to be pumped so I never got a frame with `Container()`, I guess that's one of the things WidgetTester is providing. 

All in all this looks like this approach can get us pretty decent test coverage for the webview plugin.

Fixes https://github.com/flutter/flutter/issues/29637.

## Related Issues

https://github.com/flutter/flutter/issues/29637

## Checklist

- [] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/example/test_driver/webview.dart
+++ b/packages/webview_flutter/example/test_driver/webview.dart
@@ -14,46 +14,44 @@ void main() {
   enableFlutterDriverExtension(handler: (_) => allTestsCompleter.future);
   tearDownAll(() => allTestsCompleter.complete(null));
 
-  group('$WebView', () {
-    test('initalUrl', () async {
-      final Completer<WebViewController> controllerCompleter =
-          Completer<WebViewController>();
-      runApp(
-        Directionality(
-          textDirection: TextDirection.ltr,
-          child: WebView(
-            key: GlobalKey(),
-            initialUrl: 'https://flutter.dev/',
-            onWebViewCreated: (WebViewController controller) {
-              controllerCompleter.complete(controller);
-            },
-          ),
+  test('initalUrl', () async {
+    final Completer<WebViewController> controllerCompleter =
+        Completer<WebViewController>();
+    runApp(
+      Directionality(
+        textDirection: TextDirection.ltr,
+        child: WebView(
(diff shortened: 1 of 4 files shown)


packages/webview_flutter/example/test_driver/webview.dart
```diff
@@ -0,0 +1,57 @@
...
+
+void main() {
+  final Completer<String> allTestsCompleter = Completer<String>();
+  enableFlutterDriverExtension(handler: (_) => allTestsCompleter.future);
+  tearDownAll(() => allTestsCompleter.complete(null));
+
+  test('initalUrl', () async {
+    final Completer<WebViewController> controllerCompleter =
+        Completer<WebViewController>();
+    runApp(
```
collinjackson: Consider moving into an async helper function to avoid code duplication with the test below.

amirh: I typically prefer to have code duplication across tests vs. extracting "repetitive" stuff from test cases to a common method so when a test fail and you're coming to look at it, you can tell right away what the test is doing without having to go and understand the what the common methods are doing.

Curious what are your thoughts about this?

collinjackson: I don’t feel strongly about it and I could go either way. It feels somewhat specific to the way that this plugin initializes itself so I wouldn’t generalize this decision to other plugins. If you like it better with repetition then let’s leave it.

PR #1409, opened 2019-03-27
[google_maps] Add initial google_maps tests

iskakaushik:
- Also fixes a bug where we were not doing returning success when we update the markers.


packages/google_maps_flutter/example/test_driver/google_maps.dart
```diff
@@ -0,0 +1,84 @@
...
+  test('testCompassToggle', () async {
+    GoogleMapStateSnapshot mapStateSnapshot;
+
+    mapStateSnapshot = await testController.mapStateSnapshot();
+    expect(mapStateSnapshot.compassEnabled, false);
+
+    mapTestState.toggleCompass();
+
+    // This delay exists for platform channel propagation.
+    await Future<void>.delayed(Duration(seconds: 1), () {});
```
collinjackson: Maybe we should make `toggleCompass` a `Future<void>` so we can `await` it in tests and get reliable timing?

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -199,6 +199,13 @@ public void onMethodCall(MethodCall call, MethodChannel.Result result) {
           markersController.changeMarkers((List<Object>) markersToChange);
           Object markerIdsToRemove = call.argument("markerIdsToRemove");
           markersController.removeMarkers((List<Object>) markerIdsToRemove);
+          result.success(null);
+          break;
+        }
+      case "map#stateSnapshot":
```
collinjackson: IMO it would be more consistent with other plugins to have a separate async MethodCall name for each property that you want to read off of the googleMap and expose this data as a public API of the plugin. This gives developers a suite of useful API for getting for properties of the map that can also be used in integration testing.

packages/google_maps_flutter/example/test_driver/google_maps_test_controller.dart
```diff
@@ -0,0 +1,36 @@
...
+  final MethodChannel _channel;
+
+  Future<GoogleMapStateSnapshot> mapStateSnapshot() async {
+    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+    // https://github.com/flutter/flutter/issues/26431
+    // ignore: strong_mode_implicit_dynamic_method
+    final List<dynamic> stateSnapshot =
+        await _channel.invokeMethod('map#stateSnapshot');
+    return GoogleMapStateSnapshot(
+      compassEnabled: stateSnapshot[0],
```
collinjackson: I think it's clearer to use named arguments rather than indexed arguments for plugin method calls.

packages/google_maps_flutter/example/test_driver/google_maps.dart
iskakaushik: The problem here is that we ideally want for a rebuild event to occur before the next assert. using testWidget wrapper we would have pumped a new widget, but it's not easy to do this here. Will fix this in the future.

packages/google_maps_flutter/example/test_driver/google_maps.dart
```diff
@@ -0,0 +1,87 @@
...
+    testController = GoogleMapTestController(controller.getMethodChannel());
+  });
+
+  test('testCompassToggle', () async {
+    bool compassEnabled;
+
+    compassEnabled = await testController.isCompassEnabled();
+    expect(compassEnabled, false);
+
+    mapTestState.toggleCompass();
```
amirh: Looking just at this test case (e.g if I were coming here to debug a test failure), it's kind of hard to understand what's going on.

I tend to favor keeping test cases more verbose, and having repetition across test cases such that when you read the test case you know what it is doing without having to learn the common facilities shared across the test cases in this file.

In this case I'd prefer to see something like:
```dart
pumpWidget(Directionality(
      child: GoogleMap(
        initialCameraPosition: _kInitialCameraPosition,
        compassEnabled: false,
        onMapCreated: (GoogleMapController controller) {
          widget._controllerCompleter.complete(controller);
        },
      ),
    ));
 
 compassEnabled = await testController.isCompassEnabled();
    expect(compassEnabled, false);

pumpWidget(Directionality(
      child: GoogleMap(
        initialCameraPosition: _kInitialCameraPosition,
        compassEnabled: true,
        onMapCreated: (GoogleMapController controller) {
          widget._controllerCompleter.complete(controller);
        },
      ),
    ));

compassEnabled = await testController.isCompassEnabled();
    expect(compassEnabled, true);
```

More verbose, but it's clear what the test is doing (also I believe this is the typical style of widget tests in the framework)

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -199,6 +199,13 @@ public void onMethodCall(MethodCall call, MethodChannel.Result result) {
           markersController.changeMarkers((List<Object>) markersToChange);
           Object markerIdsToRemove = call.argument("markerIdsToRemove");
           markersController.removeMarkers((List<Object>) markerIdsToRemove);
+          result.success(null);
```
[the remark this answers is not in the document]
iskakaushik: updated the PR description.

packages/google_maps_flutter/example/test_driver/google_maps.dart
iskakaushik: Created `test_widgets.dart` that has `pumpWidget`, we can potentially move it to a shared location, but fits nicely into the framework pattern.

packages/google_maps_flutter/example/ios/Runner/AppDelegate.m
```diff
@@ -4,10 +4,11 @@
 
 @implementation AppDelegate
 
-- (BOOL)application:(UIApplication *)application
-    didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+- (BOOL)application:(UIApplication*)application
+    didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
   // Provide the GoogleMaps API key.
-  [GMSServices provideAPIKey:@"YOUR KEY HERE"];
+  NSString* mapsApiKey = [[NSProcessInfo processInfo] environment][@"MAPS_API_KEY"];
```
amirh: This might be confusing for users, what do you think about still using the key "YOUR KEY HERE" if the env variable is not set?

packages/google_maps_flutter/example/pubspec.yaml
```diff
@@ -10,11 +10,11 @@ dependencies:
...
 
 dev_dependencies:
-  flutter_test:
-    sdk: flutter
-
   google_maps_flutter:
     path: ../
+  flutter_driver:
+    sdk: flutter
+  test: any
```
amirh: depend on a specific version with caret notation

packages/google_maps_flutter/example/test_driver/google_maps_test_controller.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright 2019, the Chromium project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:flutter/services.dart';
+
+class GoogleMapTestController {
```
amirh: nit: Controller feels a little weird here, as this is not controlling the map but rather allows inspecting it(TestController sounds to me like it's controlling the test). Maybe call it something like GoogleMapInspectors?

packages/google_maps_flutter/example/ios/Runner/AppDelegate.m
iskakaushik: I'm thinking of making it so that in the future for local runs we require the api key to be set via env var. That way we do not need to edit these files, and flutter run will automatically pick them up. But for short term, i'm cool with making this be `"YOUR KEY HERE"`.

packages/google_maps_flutter/example/test_driver/google_map_inspector.dart
```diff
@@ -4,8 +4,8 @@
 
 import 'package:flutter/services.dart';
 
-class GoogleMapTestController {
-  GoogleMapTestController(this._channel);
+class GoogleMapInspector {
```
amirh: nit: add a doc comment explaining the role of this class.

packages/google_maps_flutter/example/test_driver/google_maps.dart
collinjackson: I agree that we should move it to a common place and I'm happy to take the lead on that.

packages/google_maps_flutter/example/test_driver/google_map_inspector.dart
collinjackson: I'm confused about why this is its own class. Why not put methods on the `GoogleMapController`?

PR #1435, opened 2019-04-03
[cloud_firestore] Remove usage of `invokeMapMethod`

cyanglaz:
`invokeMapMethod` and `invokeListMethod`  currently crashes if the return value from methodChannel is null. Remove usages of these 2 methods to prevent crashes.

### DIFF
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.9.8+1
+
+* Remove `invokeMapMethod` calls to prevent crash.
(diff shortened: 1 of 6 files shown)


packages/cloud_firestore/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.9.8+1
```
collinjackson: I think it would be 0.9.9 rather than 0.9.8+1 since a behavior change is evident

collinjackson:
lgtm once bots are happy

PR #1436, opened 2019-04-03
Fix formatting in main.dart of firebase_database, firebase_storage, google_sign_in, shared_preference

cyanglaz:
Fix formatting on the main.dart files in example apps for firebase_database, firebase_storage, google_sign_in, shared_preference


iskakaushik:
LGTM

PR #1440, opened 2019-04-03
[image_picker] Check camera permissions and return error

wwwdata:
## Description

Under iOS, the camera permissions have not been correctly checked. If the user did not allow camera access, the camera view was launched and displayed a black screen. It was then also possible to make a photo of the black screen and use it. The camera view controller is now only launched, when the user has granted permissions. Otherwise it will not launch and flutter will also receive an error.

## Related Issues

https://github.com/flutter/flutter/issues/30378

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/image_picker/CHANGELOG.md
+++ b/packages/image_picker/CHANGELOG.md
@@ -162,7 +162,6 @@
 
 * Updated dependencies
 * **Breaking Change**: You need to add a maven section with the "https://maven.google.com" endpoint to the repository section of your `android/build.gradle`. For example:
-
 ```gradle
 allprojects {
     repositories {
--- a/packages/image_picker/CHANGELOG.md
+++ b/packages/image_picker/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.5.1+1
+
+* Check iOS camera authorizationStatus and return an error, if the access was
+  denied.
+
 ## 0.5.1
 
 * Android: Do not delete original image after scaling if the image is from gallery.
@@ -157,6 +162,7 @@
 
 * Updated dependencies
 * **Breaking Change**: You need to add a maven section with the "https://maven.google.com" endpoint to the repository section of your `android/build.gradle`. For example:
+
 ```gradle
 allprojects {
     repositories {
--- a/packages/image_picker/CHANGELOG.md
+++ b/packages/image_picker/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.5.1+1
+## 0.5.2
 
 * Check iOS camera authorizationStatus and return an error, if the access was
   denied.
--- a/packages/image_picker/ios/Classes/ImagePickerPlugin.m
+++ b/packages/image_picker/ios/Classes/ImagePickerPlugin.m
@@ -7,6 +7,7 @@
 #import <MobileCoreServices/MobileCoreServices.h>
 #import <Photos/Photos.h>
 #import <UIKit/UIKit.h>
+#import <AVFoundation/AVFoundation.h>
 
 @interface FLTImagePickerPlugin () <UINavigationControllerDelegate, UIImagePickerControllerDelegate>
 @end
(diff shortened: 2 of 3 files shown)


packages/image_picker/ios/Classes/ImagePickerPlugin.m
```diff
@@ -118,6 +119,42 @@ - (void)showCamera {
...
+            }
+          });
+        } else {
+          dispatch_async(dispatch_get_main_queue(), ^{
+            [self errorNoAccess];
+          });
+        }
+      }];
+    }; break;
+    case AVAuthorizationStatusDenied:
```
cyanglaz: It is prob a good idea to send the failed statuses to the dart code so the developers can have more details on what caused the failure.

packages/image_picker/ios/Classes/ImagePickerPlugin.m
```diff
@@ -118,6 +119,42 @@ - (void)showCamera {
...
+    case AVAuthorizationStatusRestricted:
+    default:
+      [self errorNoAccess];
+      break;
+  }
+}
+
+- (void)errorNoAccess {
+  _result([FlutterError errorWithCode:@"camera_access_denied"
```
cyanglaz: maybe the error code can be more specific on which case caused the failure?

wwwdata:
Thank you for the super fast response. Makes a lot of sense I added your requested changes so flutter now knows if the user denied access or if the access was restricted altogether. Also formatted the code with clang-format so hopefully the build will be green.

packages/image_picker/CHANGELOG.md
```diff
@@ -153,6 +158,7 @@
 
 * Updated dependencies
 * **Breaking Change**: You need to add a maven section with the "https://maven.google.com" endpoint to the repository section of your `android/build.gradle`. For example:
+
```
cyanglaz: Please remove this line.

wwwdata: oops that was my prettier plugin. Rebased, fixed and pushed again.

cyanglaz:
@wwwdata 
Thank you for the update. Sorry about the back and force. There is one more thing that I would like to mention. Since adding this change makes `showCamera` being async, we need to add a guard to prevent `[_viewController presentViewController:_imagePickerController animated:YES completion:nil];` in `showCamera` being called twice. You can prob get a nasty result by doing this: Fresh install the app, hit the camera button twice very fast. Now both `checkAuthorization` gets triggered and it goes into `AVAuthorizationStatusNotDetermined`. And you wait for a second and hit grant access. 
A simple fix for this could be something like 
```
- (void)showCamera {
  @synchronized (self) {
    if (_imagePickerController.beingPresented) {
      return;
    }
    // Camera is not available on simulators
    if ([UIImagePickerController isSourceTypeAvailable:UIImagePickerControllerSourceTypeCamera]) {
      [_viewController presentViewController:_imagePickerController animated:YES completion:nil];
    } else {
      [[[UIAlertView alloc] initWithTitle:@"Error"
                                  message:@"Camera not available."
                                 delegate:nil
                        cancelButtonTitle:@"OK"
                        otherButtonTitles:nil] show];
    }
  }
}
```
I know it's a bit more to ask. If you don't have to time to test and work on this, I am happy to shepherd this PR from here.

wwwdata:
@cyanglaz very good catch! This was also a bug that was already present in the plugin. Even without calling it asynchronously (by granting camera permission, so it would directly call the code to launch the view) I could make it crash by rapidly tapping a button in flutter to launch the camera view. Tested it, and it doesn't crash anymore now.

cyanglaz:
@wwwdata Thank you so much! Sorry that I would like to change one more thing which is the version. This change actually changing the behavior a little to the public API. (Now we can catch the error when trying to launch the image picker on iOS). So we have to increment the the version to 0.5.2.  You might also want to mention that in the documentation of `pickImage`, but I can do that documentation after we merge in this PR if you don't feel like doing it. 
As long as you can update the version to 0.5.2, I will merge this PR!
Again, thanks for fixing this!

wwwdata:
OK. I am not so sure how to exactly document this so I will leave it up to you. Thank you too for the fast and good feedback. My next pull request is hopefully a lot smoother then :)

PR #1446, opened 2019-04-04
[webview_flutter] iOS specific initialization configuration, for enabling video playback.

hpoul:
## Description

Right now video playback is not supported on iOS because it's by default disabled, while on Android it is enabled by default.
WKWebView provides configuration options for configuring video playback. I have implemented most options to configure them. This required a iOS specific configuration object.

Unfortunately I am not sure how to test this behavior, because it simply allows video playback which does not provide any checkable events afaik? I could provide a method to query those iOS configuration options, but since this is only useful in tests, i'm not sure if this makes sense..

I have changed the example application to allow inline video playback (and removed the blocking of embedded youtube videos, but youtube.com itself is still blocked. I wasn't sure what the initial intent was to block youtube.com, so hope this is okay).

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/example/lib/main.dart
+++ b/packages/webview_flutter/example/lib/main.dart
@@ -53,8 +53,12 @@ class WebViewExample extends StatelessWidget {
           javascriptChannels: <JavascriptChannel>[
             _toasterJavascriptChannel(context),
           ].toSet(),
+          iOSWebViewConfiguration: const IOSWebViewConfiguration(
+            allowsInlineMediaPlayback: true,
+          ),
           navigationDelegate: (NavigationRequest request) {
-            if (request.url.startsWith('https://www.youtube.com/')) {
+            if (request.url.startsWith('https://www.youtube.com/') &&
+                !request.url.contains('embed')) {
               print('blocking navigation to $request}');
               return NavigationDecision.prevent;
             }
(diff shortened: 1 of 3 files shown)


hpoul:
@amirh maybe you could give a short comment if this implementation could fit into the project, then i could start working on basing basic http auth support based on this PR.. (because supporting authentication requires error handling callbacks and I wouldn't want to completely refactor it again..) thanks. :)

stuartmorgan:
Thanks for the submission! We’re currently working through a large backlog of PRs, and this will require non-trivial review (since we need to evaluate whether there are API changes we need to make at the same time), so it will take some time before we’re able to review it. As explained in [CONTRIBUTING.md](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#the-review-process), votes for the corresponding issue are the primary way we’re prioritizing non-trivial reviews, so we encourage you to file an issue and associate it with this PR.

Note that it's not clear that all of these are actually things that would only ever apply to iOS, rather than be generic across platforms; an approach that frames things in more generic terms where applicable, with comments about what platforms they currently apply to, would be much more future-poof.

hpoul:
I think in the meantime the webview plugin supports this out of the box through `initialMediaPlaybackPolicy: webview.AutoMediaPlaybackPolicy.always_allow`

PR #985, opened 2018-12-13
[google_maps_flutter]add support for map tapping

RafaO:
Deliver the tap event in the map with the LatLng object where it happened. 

Enables something like this:

          GoogleMap(
            onMapTapped: (LatLng location) {
              showDialog(context: context, builder: (BuildContext context) {
                return AlertDialog(
                  title: new Text("Map tapped!"),
                  content: new Text("You tapped in " + location.latitude.toString() + ", " + location.longitude.toString()),
                  actions: <Widget>[
                    // usually buttons at the bottom of the dialog
                    new FlatButton(
                      child: new Text("Close"),
                      onPressed: () {
                        Navigator.of(context).pop();
                      },
                    ),
                  ],
                );
              });
            },
          )

PR #1447, opened 2019-04-04
Update changelog and pubspec for onTap

iskakaushik:
Forgot to identify them being missed in https://github.com/flutter/plugins/pull/985


collinjackson:
lgtm but use 0.5.2 instead of 0.6.0

PR #1450, opened 2019-04-05
Link with email and link

creativecreatorormaybenot:
Note that I am not so sure about the error codes, i.e. I copied them from `linkWithCredential` but left out weak password as there is no password to set in this case.


creativecreatorormaybenot:
@collinjackson I actually have it checked, so you should be able to commit to the `linkWithEmailAndLink` branch, however, I am not sure what else I can change.

The test that is failing seems to be due to the existing example add. I will add your unit test.

creativecreatorormaybenot:
I looked at the example app and concluded that it would probably not make much sense to include usage for linking as there is non for linking a credential anyway. It would be a copy of `_EmailLinkSignInSection` where there is a sign in anonymously first and then simply the linking. I can do that if you want.

creativecreatorormaybenot:
#1458

PR #1403, opened 2019-03-26
[Camera] Improved resolution presets and matched quality between Android and iOS

quentinleguennec:
## Description

This PR is here to address issues in the quality of video recording for the camera plugin on Android, as well as discrepancies in quality level between Android and iOS. It also allows to control the still picture quality (which was so far only maximum quality). The existing resolution presets have been updated to match between Android and iOS, and 2 new presets have been added (veryHigh and veryLow). 

There are no breaking changes but the quality of the presets have been adjusted which will result in different picture/video quality on apps using this plugin.

We believe this change is necessary as, for the same resolution preset, Android video quality was very poor and the quality on iOS was to high, generating 4k files that are too heavy to upload.

## Related Issues

https://github.com/flutter/plugins/pull/1186
https://github.com/flutter/plugins/pull/1107
https://github.com/flutter/flutter/issues/30459
https://github.com/flutter/flutter/issues/29500
https://github.com/flutter/flutter/issues/26959

## Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]). => There are no tests on this plugin.
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.
But: Existing presets have been updated, this will affect the quality of pictures and videos in existing apps.

PR #1307, opened 2019-03-04
[video_player] Fix some videos being rendered black on iOS

recastrodiaz:
I've found another instance of a video file that does not render using the video_player. The video settings are as follows:

```LOG
TX, TY, WIDTH, HEIGHT, ROTATION (degrees)
1080.0, 0.0, 848.0, 480.0, 90
```

Fixes https://github.com/flutter/flutter/issues/28831

For some reason the .tx value is incorrect. Setting it to 848.0 correctly renders the video.
I have refactored the code so .tx and .ty are always set (when rotationDegrees is not 0) regardless of their original values.

PR #1451, opened 2019-04-05
[video_player] Fixed orientation and position issue for some videos metadata.

quentinleguennec:
## Description

The video was not always displaying properly on iOS depending on the rotation metadata (iOS need the tx and ty of the transform to have an offset if the video is rotated, unlike Android or VLC). This fixes it.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/29951
Fixes https://github.com/flutter/flutter/issues/27201
Fixes https://github.com/flutter/flutter/issues/29500

NOTE: The camera plugin also need an update to add rotation metadata to the recorded videos. This is done in this other PR: https://github.com/flutter/plugins/pull/1452

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/video_player/ios/Classes/VideoPlayerPlugin.m
+++ b/packages/video_player/ios/Classes/VideoPlayerPlugin.m
@@ -154,19 +154,27 @@ static void* playbackBufferFullContext = &playbackBufferFullContext;
   // At least 2 user videos show a black screen when in portrait mode if we directly use the
   // videoTrack.preferredTransform Setting tx to the height of the video instead of 0, properly
   // displays the video https://github.com/flutter/flutter/issues/17606#issuecomment-413473181
-  if (transform.tx == 0 && transform.ty == 0) {
-    NSInteger rotationDegrees = (NSInteger)round(radiansToDegrees(atan2(transform.b, transform.a)));
-    NSLog(@"TX and TY are 0. Rotation: %ld. Natural width,height: %f, %f", (long)rotationDegrees,
-          videoTrack.naturalSize.width, videoTrack.naturalSize.height);
+  NSInteger rotationDegrees = (NSInteger)round(radiansToDegrees(atan2(transform.b, transform.a)));
+  // This prevents the video to be rendered out of the screen if the metadata contain a rotation but
+  // no translation to compensate the shift induced by the rotation.
+  if (rotationDegrees != 0 && transform.tx == 0 && transform.ty == 0) {
+    NSLog(@"Adding translation to compensate rotation. Rotation = %ld. Natural (width, height) = "
+          @"(%f, %f)",
+          (long)rotationDegrees, videoTrack.naturalSize.width, videoTrack.naturalSize.height);
+    NSLog(@"Uncompensated transform (a, b, c, d, tx, ty) = (%f, %f, %f, %f, %f, %f)", transform.a,
+          transform.b, transform.c, transform.d, transform.tx, transform.ty);
     if (rotationDegrees == 90) {
-      NSLog(@"Setting transform tx");
       transform.tx = videoTrack.naturalSize.height;
       transform.ty = 0;
+    } else if (rotationDegrees == 180) {
+      transform.tx = videoTrack.naturalSize.width;
+      transform.ty = videoTrack.naturalSize.height;
     } else if (rotationDegrees == 270) {
-      NSLog(@"Setting transform ty");
       transform.tx = 0;
       transform.ty = videoTrack.naturalSize.width;
     }
+    NSLog(@"Compensated transform (a, b, c, d, tx, ty) = (%f, %f, %f, %f, %f, %f)", transform.a,
+          transform.b, transform.c, transform.d, transform.tx, transform.ty);
   }
   return transform;
 }
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.0+5
+
+* iOS: Fixed orientation and position issue for some videos metadata.
+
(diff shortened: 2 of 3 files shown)


ened:
Do you have a sample video to show before/after?

quentinleguennec:
> Do you have a sample video to show before/after?

The difference is that the video will always display in the correct orientation when viewed on the video_player (with no distortion when the aspect ratio was not matching the video aspect ratio after a rotation).
**BUT** this doesn't change the fact that the video player container will rotate *side way* if the phone is rotated. This is not a plugin issue and to prevent it the screen orientation should either be lock to landscape, or a rotation (using a `RotatedBox`) should be applied to the VideoPlayer on orientation change to counter the screen rotation.

quentinleguennec:
I did this fix when working on this PR for the camera plugin on iOS https://github.com/flutter/plugins/pull/1452
When I add the translation directly in the metadata it displays properly (and the above fix is not needed). This translation is only needed on iOS though and displays properly on Android without it. I think it's best to fix it in the video_player directly because some videos (like the one produced by the camera plugin on Android) don't have the translation metadata and will not display properly on iOS without this fix.

packages/video_player/ios/Classes/VideoPlayerPlugin.m
```diff
@@ -154,19 +154,27 @@ - (CGAffineTransform)fixTransform:(AVAssetTrack*)videoTrack {
...
+    NSLog(@"Adding translation to compensate rotation. Rotation = %ld. Natural (width, height) = "
+          @"(%f, %f)",
+          (long)rotationDegrees, videoTrack.naturalSize.width, videoTrack.naturalSize.height);
+    NSLog(@"Uncompensated transform (a, b, c, d, tx, ty) = (%f, %f, %f, %f, %f, %f)", transform.a,
+          transform.b, transform.c, transform.d, transform.tx, transform.ty);
     if (rotationDegrees == 90) {
-      NSLog(@"Setting transform tx");
       transform.tx = videoTrack.naturalSize.height;
       transform.ty = 0;
+    } else if (rotationDegrees == 180) {
```
recastrodiaz: @quentinleguennec Thanks for your PR. I just tested your fix and unfortunately this line is not rendering videos taken at 180°. tx and ty must be 0 for videos in landscape and reverse landscape mode. Furthermore, some videos metadata include incorrect tx, ty values when in portrait mode. In https://github.com/flutter/plugins/pull/1307/files I've added a fix to properly display these videos.

quentinleguennec: It might be because I was using the fixed rotation metadata from this PR:  https://github.com/flutter/plugins/pull/1452
On my side with both those PRs the videos display properly. (even when the video is recorded with a rotation).

recastrodiaz: Have you tried recording your videos with the native Camera app? That's the one I used to record a video in 180° that displays a black screen with tx and ty set to non 0.

quentinleguennec: Good point, I didn't think about checking with the native camera app.
I just did and it is working with the current code when you record a video on your iPhone and play it with the video_player. But it is not working (black screen) if you record the video from the native camera app of an Android an play it with the video_player on an iPhone.

That's because the native iPhone camera app add the translation metadata to the video file (which can be seen with ffprobe). But Android doesn't add it (and VLC doesn't need it either, just iOS). 

With the fix I added it displays the video recorded on Android properly when played on an iPhone. And it also works with a video recorded from an iPhone and played on an iPhone from the video_player.

quentinleguennec:
I'll drop this here while we wait for the PR review, it's a fix for the camera plugin and video plugin recording quality and rotation issues. For people interested in a fix have a look at the discussion on this issue: https://github.com/flutter/flutter/issues/29951

Copy-paste:
When I worked on solving those issue I did 4 fixes in the camera plugin and video_player plugin. And it is quite possible it will only work well only with the 4 fixes together. Here is the list of PR (only one of them has been merged so far and I have yet to get any answer from reviewers on the others):

    flutter/plugins#1452
    flutter/plugins#1403
    flutter/plugins#1451
    flutter/plugins#1470

I created a branch with those 4 fixes merged together, I suggest you try this branch first and see if it works on your side:
https://github.com/Tengio/plugins/tree/tested4you_video_player_camera_merge

You can use it by adding this in your pubspec.yaml:
```
 video_player:
    git:
      url: git://github.com/tengio/plugins.git
      ref: 310a8f06c4981121a5553b0f6f3a8cdf475b8af1
      path: packages/video_player
  camera:
    git:
      url: git://github.com/tengio/plugins.git
      ref: 310a8f06c4981121a5553b0f6f3a8cdf475b8af1
      path: packages/camera
```

cbenhagen:
@quentinleguennec thank you for your contribution! Unfortunately the structure of the plugin has changed lately to support implementations on more platforms. Mainly the plugin was moved to a new subdirectory and now uses a common platform interface. Adapting your PR to the new structure should be straight forward. If you have questions please ask!

Please also make sure to include tests for the new functionality.

kuyazee:
Currently using the ff
- camera: 0.5.8+11
- video_player: 0.10.12+5

Issue still persists

stuartmorgan:
Thanks for the contribution! Based on the discussion above, it sounds like this PR may not actually work in isolation, which would be a blocker for moving forward with review and landing. It also would need to be updated for the changes in plugin structure as noted above. Given that, and the long period without updates, I'm going to close this, but if you're still interested in moving forward with this please do open a new PR (which should be triaged quickly under our new process) with a version that works in isolation—combining fixes if necessary.

PR #1455, opened 2019-04-05
[connectivity]Added integration test.

cyanglaz:
Add test driver to the plugin.

### DIFF
--- a/packages/connectivity/CHANGELOG.md
+++ b/packages/connectivity/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.4.2+2
 
-* Add test driver.
+* Add integration test.
 
 ## 0.4.2+1
 
--- a/packages/connectivity/CHANGELOG.md
+++ b/packages/connectivity/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.2+2
+
+* Add test driver.
+
 ## 0.4.2+1
 
 * Bump the minimum Flutter version to 1.2.0.
--- a/packages/connectivity/example/test_driver/connectivity.dart
+++ b/packages/connectivity/example/test_driver/connectivity.dart
@@ -15,14 +15,9 @@ void main() {
       _connectivity = Connectivity();
     });
 
-    test('test _checkConnectivity', () async {
-      final ConnectivityResult result = await _connectivity.checkConnectivity();
-      expect(result, isNotNull);
-    });
-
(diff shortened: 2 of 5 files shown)


packages/connectivity/example/test_driver/connectivity.dart
```diff
@@ -0,0 +1,36 @@
...
+  tearDownAll(() => completer.complete(null));
+
+  group('Connectivity test driver', () {
+    Connectivity _connectivity;
+
+    setUpAll(() async {
+      _connectivity = Connectivity();
+    });
+
+    test('test _checkConnectivity', () async {
```
collinjackson: I would remove test and underscore

packages/connectivity/example/test_driver/connectivity.dart
```diff
@@ -0,0 +1,36 @@
...
+      final ConnectivityResult result = await _connectivity.checkConnectivity();
+      expect(result, isNotNull);
+    });
+
+    test('test connectivity result', () async {
```
collinjackson: We could probably combine this test with the one above.

packages/connectivity/example/test_driver/connectivity.dart
```diff
@@ -0,0 +1,36 @@
...
+import 'package:flutter_driver/driver_extension.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:connectivity/connectivity.dart';
+
+void main() {
+  final Completer<String> completer = Completer<String>();
+  enableFlutterDriverExtension(handler: (_) => completer.future);
+  tearDownAll(() => completer.complete(null));
+
+  group('Connectivity test driver', () {
```
collinjackson: Elsewhere in the plugins repo we use the stringified class name '$Connectivity' to identify group so you may want to be consistent with that, though it's not required. I wouldn't say "test driver" though.

packages/connectivity/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.4.2+2
+
+* Add test driver.
```
collinjackson: I would say "Added integration test." Although we're using Flutter driver to launch the test, we don't use the FlutterDriver API to do anything, so it may be confusing otherwise.

collinjackson:
lgtm modulo the nits above

collinjackson:
looks ready to merge

PR #1459, opened 2019-04-07
[webview_flutter] Added possibility to activate the zoom webview Android - no-op on IOS (Android)

Dn-a:
## Description

replaces previous  [pull request #1235 ](https://github.com/flutter/plugins/pull/1235)


Dn-a:
@amirh 
are there any news?

PR #1460, opened 2019-04-07
[shared_preferences] Make a clone of mutable containers before set/get.

devxpy:
Fixes https://github.com/flutter/flutter/issues/28015

### DIFF
--- a/packages/shared_preferences/lib/shared_preferences.dart
+++ b/packages/shared_preferences/lib/shared_preferences.dart
@@ -80,13 +80,9 @@ class SharedPreferences {
   /// exception if it's not a string set.
   List<String> getStringList(String key) {
     List<Object> list = _preferenceCache[key];
-    if (list != null) {
-      list = List<String>.from(list);
-      if (list is! List<String>) {
-        _preferenceCache[key] = list;
-      }
-    }
-    return list;
+    List<String> newList;
+    if (list != null) newList = List<String>.from(list);
+    return newList;
   }
 
   /// Saves a boolean [value] to persistent storage in the background.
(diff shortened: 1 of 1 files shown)


packages/shared_preferences/lib/shared_preferences.dart
```diff
@@ -74,9 +80,11 @@ class SharedPreferences {
   /// exception if it's not a string set.
   List<String> getStringList(String key) {
     List<Object> list = _preferenceCache[key];
-    if (list != null && list is! List<String>) {
-      list = list.cast<String>().toList();
-      _preferenceCache[key] = list;
+    if (list != null) {
+      list = List<String>.from(list);
+      if (list is! List<String>) {
```
apaatsio: Can this `if` condition ever be true? On line 84, we assign `list = List<String>.from(list)`, so is it even possible for `list` to be anything else than `List<String>` on line 85?

neelu004:
Hi, I am adding the dependency of shared_preference and it is working fine when I am running my project on android using windows but when I am using Mac(Xcode) it is giving me error as ld: warning: directory not found for option '-L/Users/ghanshyam/Development/Android/Workspace/my_app/build/ios/Debug-iphoneos/shared_preferences'
ld: library not found for -lshared_preferences
clang: error: linker command failed with exit code 1 (use -v to see invocation)
I had tried using the following methods but that doesn't works for me. They are:-
pod install
pod upgrade
flutter clean
flutter packages get
flutter packages upgrade
flutter run using command it is working fine but works on simulator still could not run on my real iPhone device using the same command.
also when I am using flutter build ios, this will creates the Runner.app but how can I install this Runner.app on my iPhone device.
Please suggest some solution for the same.
Thanks!

packages/shared_preferences/lib/shared_preferences.dart
devxpy: Looks like an overlook on my part. Sorry! Will issue a fix.

devxpy: @apaatsio Does it look better now?

stuartmorgan:
This was addressed in https://github.com/flutter/plugins/pull/1276, so closing.

PR #1469, opened 2019-04-09
[package_info] [firebase_remote_config] Allow reset of static variables in testing

thomas-stockx:
## Description

This adds a reset method to package_info and firebase_remote_config, to allow making sure that the behavior is always the same between multiple tests.

## Related Issues

We would have discrepancies when certain tests were run in a different order. Resetting the static variables between tests solves this.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


collinjackson:
Can you give me an example of what a test looks like that would use this test API? Is it a local unit test, a live test that runs on a device, or a driver test? It feels like we might want to push these singletons more in the direction of dependency injection but I'd like to understand more about how they're being used.

cyanglaz:
For [package_info], we can add an optional parameter for `PackageInfo.fromPlatform()` to let it get a refreshed result. This way we don't break the API.
For example 
```
  static Future<PackageInfo> fromPlatform({bool refreshResult = false}) async {
    if (_fromPlatform == null || refreshResult) {
      final Completer<PackageInfo> completer = Completer<PackageInfo>();

      // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
      // https://github.com/flutter/flutter/issues/26431
      // ignore: strong_mode_implicit_dynamic_method
      _kChannel.invokeMethod('getAll').then((dynamic result) {
        final Map<dynamic, dynamic> map = result;

        completer.complete(PackageInfo(
          appName: map["appName"],
          packageName: map["packageName"],
          version: map["version"],
          buildNumber: map["buildNumber"],
        ));
      }, onError: completer.completeError);

      _fromPlatform = completer.future;
    }
    return _fromPlatform;
  }
```

thomas-stockx:
> For [package_info], we can add an optional parameter for `PackageInfo.fromPlatform()` to let it get a refreshed result. This way we don't break the API.
> For example
> 
> ```dart
>   static Future<PackageInfo> fromPlatform({bool refreshResult = false}) async {
>     if (_fromPlatform == null || refreshResult) {
>       final Completer<PackageInfo> completer = Completer<PackageInfo>();
> 
>       // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
>       // https://github.com/flutter/flutter/issues/26431
>       // ignore: strong_mode_implicit_dynamic_method
>       _kChannel.invokeMethod('getAll').then((dynamic result) {
>         final Map<dynamic, dynamic> map = result;
> 
>         completer.complete(PackageInfo(
>           appName: map["appName"],
>           packageName: map["packageName"],
>           version: map["version"],
>           buildNumber: map["buildNumber"],
>         ));
>       }, onError: completer.completeError);
> 
>       _fromPlatform = completer.future;
>     }
>     return _fromPlatform;
>   }
> ```

Thanks for the feedback, and I agree this should probably be added.

But wouldn't this mean I'd always need to set refreshResult to true for my widget tests to succeed?
As a result, my code in production will also always refresh package_info, even though I only need to refresh during testing (which is why I annotated the reset method with @visibleInTesting in the first place).

collinjackson:
> We noticed that methods like PackageInfo.fromPlatform() which return a Future, would only resolve in the first widget test, while following tests would fail due to the future not resolving anymore.

Can you give a simple test case of some code that would cause this to happen? You should be able to `await` the same `Future` returned by `PackageInfo.fromPlatform()` in multiple widget tests, and it should always return the same value, so I want to understand why additions to the API are needed.

stuartmorgan:
Thanks for the submission! However, we’re no longer accepting non-critical PRs for this plugin, as we hope in time to be able to transition users to the corresponding plugin in [the Flutter Community Plus Plugins repository](https://plus.fluttercommunity.dev/). We encourage you to submit your PR there.

PR #1476, opened 2019-04-10
[cloud_firestore] support for pagination using startAtDocument, endAtDocument, etc.

collinjackson:
## Description

Adds `startAtDocument` and other DocumentSnapshot methods to `cloud_firestore` `Query`

## Related Issues

Originally based on #1426 authored by @radvansky-tomas
Edits, unit and integration tests added by @collinjackson 

## Breaking Change

Does your PR require plugin user to manually update their apps to accommodate your change?


collinjackson:
@radvansky-tomas per the request from @googlebot can you please confirm that you're ok with these commits being contributed to this project?

radvansky-tomas:
Amazing I am happy with any change you made to my code ;)

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -419,26 +496,21 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
+    getQuery(call.arguments, ^(FIRQuery *query) {
+      id<FIRListenerRegistration> listener = [query
+          addSnapshotListener:^(FIRQuerySnapshot *_Nullable snapshot, NSError *_Nullable error) {
+            if (snapshot == nil) {
+              result(getFlutterError(error));
+              return;
+            }
+            NSMutableDictionary *arguments = [parseQuerySnapshot(snapshot) mutableCopy];
+            [arguments setObject:handle forKey:@"handle"];
+            [self.channel invokeMethod:@"QuerySnapshot" arguments:arguments];
```
cyanglaz: Using `self` in this block seems can cause retain cycle? Maybe use weakSelf instead?

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -477,8 +549,26 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
         result(getFlutterError(error));
         return;
       }
-      result(parseQuerySnapshot(snapshot));
+      [self.channel invokeMethod:@"DocumentSnapshot"
+                       arguments:@{
+                         @"handle" : handle,
+                         @"path" : snapshot ? snapshot.reference.path : [NSNull null],
+                         @"data" : snapshot && snapshot.exists ? snapshot.data : [NSNull null],
```
cyanglaz: nit: `snapshot && snapshot.exists ?` can be replaced as `snapshot.exists ?` since if snapshot is nil, snapshot.exists will also return false.
```suggestion
                         @"data" : snapshot.exists ? snapshot.data : [NSNull null],
```

packages/cloud_firestore/lib/src/query.dart
```diff
@@ -53,7 +53,10 @@ class Query {
     StreamController<QuerySnapshot> controller; // ignore: close_sinks
     controller = StreamController<QuerySnapshot>.broadcast(
       onListen: () {
-        _handle = Firestore.channel.invokeMethod<int>(
+        // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+        // https://github.com/flutter/flutter/issues/26431
+        // ignore: strong_mode_implicit_dynamic_method
+        _handle = Firestore.channel.invokeMethod(
```
cyanglaz: There seems to be a merging/rebase issue where invokeMethod's type parameters got reverted back in this PR in multiple places.

packages/cloud_firestore/lib/src/query.dart
```diff
@@ -162,6 +170,46 @@ class Query {
     return _copyWithParameters(<String, dynamic>{'orderBy': orders});
   }
 
+  /// Takes a documentSnapshot, creates and returns a new [Query] that starts after
+  /// the provided document reference to the order of the query.
+  ///
+  /// documentSnapshot must be in order of [orderBy] filters.
```
cyanglaz: document sentence need to start with capital letters. also highlight the parameters Maybe: 
```suggestion
  /// The `documentSnapshot` must be in order of [orderBy] filters.
```
Same applies to other new query methods

packages/cloud_firestore/lib/src/query.dart
```diff
@@ -162,6 +170,46 @@ class Query {
...
+  ///
+  /// Cannot be used in combination with [startAtDocument], [startAt].
+  Query startAfterDocument(DocumentSnapshot documentSnapshot) {
```
cyanglaz: Maybe add documents to point people to other querying methods? 
like:
```
/// See also:
/// * [startAtDocument] <add some explaination>
/// * [endAtDocument] <add some explaination>
/// .....
```

packages/cloud_firestore/lib/src/query.dart
```diff
@@ -162,6 +170,46 @@ class Query {
...
+    assert(!_parameters.containsKey('startAtDocument'));
+    return _copyWithParameters(<String, dynamic>{
+      'startAfterDocument': <dynamic>[
+        documentSnapshot.documentID,
+        documentSnapshot.data
+      ]
+    });
+  }
+
+  /// Takes a documentSnapshot, creates and returns a new [Query] that starts after
```
cyanglaz: nit:
```suggestion
  /// Takes a documentSnapshot, creates and returns a new [Query] that starts at
```

packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
```diff
@@ -187,12 +188,75 @@ private Query getQuery(Map<String, Object> arguments) {
       query = query.orderBy(orderByFieldName, direction);
     }
     @SuppressWarnings("unchecked")
+    ArrayList startAtDocument = (ArrayList) parameters.get("startAtDocument");
```
cyanglaz: Feels like the below 4 blocks of code of 4 different querying are pretty identical, can refactor them out to a private method.

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -67,22 +68,98 @@
...
   }
+  id startAtDocument = parameters[@"startAtDocument"];
+  if (startAtDocument) {
+    NSArray *startAtValues = startAtDocument;
+    NSMutableArray *startAtQueryValues = [[NSMutableArray alloc] init];
+    query = [query queryOrderedByFieldPath:FIRFieldPath.documentID descending:NO];
+    NSString *startAtDocId = startAtValues[0];
+    NSDictionary *startAtData = startAtValues[1];
+    if (orderBy) {
+      for (id item in orderBy) {
```
cyanglaz:
```suggestion
      for (NSArray *item in orderBy) {
```

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -67,22 +68,98 @@
     NSArray *startAtValues = startAt;
     query = [query queryStartingAtValues:startAtValues];
   }
+  id startAtDocument = parameters[@"startAtDocument"];
+  if (startAtDocument) {
+    NSArray *startAtValues = startAtDocument;
+    NSMutableArray *startAtQueryValues = [[NSMutableArray alloc] init];
+    query = [query queryOrderedByFieldPath:FIRFieldPath.documentID descending:NO];
```
cyanglaz: Since we are at it now, should we expose `descending` to dart as well?

packages/cloud_firestore/lib/src/query.dart
collinjackson: This sentence actually just doesn't make sense, removing

packages/cloud_firestore/lib/src/query.dart
collinjackson: You're right, I've fixed this

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
collinjackson: It's a reasonable question, but I don't think this has have a corresponding equivalent in the Android or iOS API, so I don't think we need to worry about it.

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -67,22 +68,98 @@
     NSArray *startAtValues = startAt;
     query = [query queryStartingAtValues:startAtValues];
   }
+  id startAtDocument = parameters[@"startAtDocument"];
+  if (startAtDocument) {
+    NSArray *startAtValues = startAtDocument;
```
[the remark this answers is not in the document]
collinjackson: I think we should use maps instead of lists to communicate information that can't be serialized with a MethodChannel codec. I'm switch this PR over to doing that.

packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/CloudFirestorePlugin.java
collinjackson: Good point. I have done this, PTAL

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -67,22 +68,98 @@
...
+        NSArray *orderByParameters = item;
+        NSString *fieldName = orderByParameters[0];
+        [startAfterQueryValues addObject:[startAfterData objectForKey:fieldName]];
+      }
+    }
+
+    [startAfterQueryValues addObject:startAfterDocId];
+    query = [query queryStartingAfterValues:startAfterQueryValues];
+    return handler(query);
+  }
```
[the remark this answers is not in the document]
collinjackson: Done, PTAL

collinjackson:
@cyanglaz can you please review and let me know if this ok to merge or if you have further comments. thanks!

packages/cloud_firestore/ios/Classes/CloudFirestorePlugin.m
```diff
@@ -311,12 +346,12 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
-          [self.channel invokeMethod:@"DoTransaction"
-                           arguments:call.arguments
-                              result:^(id doTransactionResult) {
-                                self->transactionResults[transactionId] = doTransactionResult;
-                                dispatch_semaphore_signal(semaphore);
-                              }];
+          [weakSelf.channel invokeMethod:@"DoTransaction"
+                               arguments:call.arguments
+                                  result:^(id doTransactionResult) {
+                                    self->transactionResults[transactionId] = doTransactionResult;
```
cyanglaz:
```suggestion
                                    weakSelf.transactionResults[transactionId] = doTransactionResult;
```

creativecreatorormaybenot:
@collinjackson `startAfterDocument` does not work for me on Android.

PR #1481, opened 2019-04-11
[image_picker]remove unnecessary dev dependency.

cyanglaz:
Removed video_player dev dependency since we don't seem to use video_player anywhere in this plugin.


cyanglaz:
Closing this for now, this can only land until https://github.com/dart-lang/pub/issues/2101 is fixed.

PR #1485, opened 2019-04-11
Fix unused vars and rename

iskakaushik:


packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -154,6 +154,13 @@ static Object latlngBoundsToJson(CameraPosition position) {
     return data;
   }
 
+  static Object latlngBoundsToJson(LatLngBounds latLngBounds) {
```
cyanglaz: The return type can be `Map<String, Object> arguments`

iskakaushik: I was being consistent with all the other return methods in this class. These are in most cases passed on to the dart side and not really used for their return nature, so left them as those. Ideally, I agree that we would use the most specific return type, but in this case I don't think it matters as much.

cyanglaz: No problem!

PR #1487, opened 2019-04-11
[firebase_auth] Migrate FlutterAuthPlugin from deprecated APIs

mehmetf:
This is mirroring a change Firebase team has done in Google. It is not a breaking change.

### DIFF
--- a/packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m
+++ b/packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m
@@ -32,13 +32,14 @@ NSDictionary *toDictionary(id<FIRUserInfo> userInfo) {
 
 @implementation FLTFirebaseAuthPlugin
 
-// Handles are ints used as indexes into the NSMutableDictionary of active observers
+// Handles are ints used as indexes into the NSMutableDictionary of active
+// observers
 int nextHandle = 0;
 
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
-  FlutterMethodChannel *channel =
-      [FlutterMethodChannel methodChannelWithName:@"plugins.flutter.io/firebase_auth"
-                                  binaryMessenger:[registrar messenger]];
+  FlutterMethodChannel *channel = [FlutterMethodChannel
+      methodChannelWithName:@"plugins.flutter.io/firebase_auth"
+            binaryMessenger:[registrar messenger]];
   FLTFirebaseAuthPlugin *instance = [[FLTFirebaseAuthPlugin alloc] init];
   instance.channel = channel;
   instance.authStateChangeListeners = [[NSMutableDictionary alloc] init];
@@ -51,7 +52,8 @@ int nextHandle = 0;
     if (![FIRApp appNamed:@"__FIRAPP_DEFAULT"]) {
       NSLog(@"Configuring the default Firebase app...");
       [FIRApp configure];
-      NSLog(@"Configured the default Firebase app %@.", [FIRApp defaultApp].name);
+      NSLog(@"Configured the default Firebase app %@.",
+            [FIRApp defaultApp].name);
     }
   }
   return self;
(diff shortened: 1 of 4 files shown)


collinjackson:
In the future could you please open PRs using branches on your fork https://github.com/mehmetf/plugins instead of the official repo https://github.com/flutter/plugins?

collinjackson:
I fixed the build on this one, I think. I'll merge it once the bots pass.

mehmetf:
As long as that can be done entirely from browser. I will try editing my own fork and then pushing the PR for the next one. I don't have a frequently used developer flow for making github changes so it is pretty heavy for me to setup a github client etc for infrequent small changes like this.

packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m
```diff
@@ -170,13 +170,13 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
                       }];
   } else if ([@"reauthenticateWithCredential" isEqualToString:call.method]) {
     [[self getAuth:call.arguments].currentUser
-        reauthenticateWithCredential:[self getCredential:call.arguments]
+        reauthenticateAndRetrieveDataWithCredential:[self getCredential:call.arguments]
```
mehmetf: This diverges from the google3 change. I will make this change and make sure it is compatible and if not, I need to talk to firebase folks and see what they recommend. See history of FirebaseAuthPlugin.m in google3 if you are curious.

mehmetf:
Thanks Collin. See my comment on the code though. Please wait for me to check this in google3 before merging otherwise we are back to square one.

mehmetf:
The compilation problems are unlikely to be resolved here. I am not working on a local branch and my iOS build fails because of the `FIRAuthErrorUserInfoNameKey` as well.

How do we control which version of Firebase we use? I don't see anything in the podspec.

mehmetf:
Very odd. I upgraded flutter and plugins and when I try to do `flutter build ios` in the firebase_auth example dir, it fails with 4 errors including:

```
    /Users/mehmetf/github/plugins/packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m:10:37: error: use of undeclared identifier
    'FIRAuthErrorUserInfoNameKey'; did you mean 'FIRAuthErrorUserInfoEmailKey'?
      NSString *code = [error userInfo][FIRAuthErrorUserInfoNameKey];
                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~
                                        FIRAuthErrorUserInfoEmailKey
```

Let's look at this Monday together and figure it out.

packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m
collinjackson: It looks like in google3 `reauthenticateWithCredential` and `reauthenticateAndRetrieveDataWithCredential` do the same thing, so this change should work fine.

collinjackson:
I think the issue is that you're using FirebaseAuth functionality that was introduced in version 5 of the FirebaseAuth CocoaPod, so we need to specify that dependency in the manifest of the firebase_auth plugin. I've added the version constraint to your change and updated the CHANGELOG.

mehmetf:
Thanks Collin, since this change is required by google3, let's not merge it for now. I will talk to you on Monday and see what we can do. (I did the version restriction to my git client as well but I am still failing to compile :-/, I did `flutter clean` and I see it running "Installing Pods").

Also, if you are constraining version to 5 and above, you should be able to use `reauthenticateWithCredential` as opposed to `reauthenticateAndRetrieveDataWithCredential` which I believe is deprecated starting in version 5.

collinjackson:
My understanding is that `reauthenticateAndRetrieveDataWithCredential` is the new API and `reauthenticateWithCredential` is the deprecated one.

e.g. the docs for [https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Classes/FIRUser#/c:objc(cs)FIRUser(im)reauthenticateWithCredential:completion:](`reauthenticateWithCredential:completion:`) say "Please use reauthenticateAndRetrieveDataWithCredential:completion: for Objective-C or reauthenticateAndRetrieveData(WithCredential:completion:) for Swift instead."

mehmetf:
Thanks Collin. I created a thread with the Firebase folks. Let's hold off on this.

mehmetf:
LGTM

packages/firebase_auth/ios/firebase_auth.podspec
```diff
@@ -16,7 +16,7 @@ Firebase Auth plugin for Flutter.
   s.public_header_files = 'Classes/**/*.h'
   s.ios.deployment_target = '8.0'
   s.dependency 'Flutter'
-  s.dependency 'Firebase/Auth'
+  s.dependency 'Firebase/Auth', '~> 5'
```
paulb777: This is meaningless. To lock to 5.x, use '~> 5.0'.  See https://guides.cocoapods.org/syntax/podspec.html#dependency

PR #1489, opened 2019-04-12
BitmapDescriptor#fromBytes should be consistent across platforms

iskakaushik:
- Android Bitmaps scale images to account for dpi.
- iOS UIImages do not, we need to account for that.


ened:
@cyanglaz @iskakaushik please publish on pub on every release..

iskakaushik:
@ened sorry about that, just published. Will try to make it a part of our CI.

PR #1492, opened 2019-04-14
[firebase_analytics] Initial integration test

collinjackson:
## Description

Starting point for integration testing. I don't know how to check the results yet, so just making sure that exceptions aren't thrown.

## Breaking Change

### DIFF
--- a/packages/firebase_analytics/example/test_driver/firebase_analytics.dart
+++ b/packages/firebase_analytics/example/test_driver/firebase_analytics.dart
@@ -1,0 +1,37 @@
+// Copyright 2019, the Chromium project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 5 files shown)


packages/firebase_analytics/example/test_driver/firebase_analytics.dart
```diff
@@ -0,0 +1,38 @@
...
+      if (Platform.isIOS) {
+        expect(analytics.android, isNull);
+      }
+      if (Platform.isAndroid) {
+        await analytics.android.setMinimumSessionDuration(9000);
+      }
+    });
+
+    test('logging', () async {
+      await analytics.setAnalyticsCollectionEnabled(true);
```
cyanglaz:
```suggestion
      expect(analytics.setAnalyticsCollectionEnabled(true), completes);
```

cyanglaz: same as other places where we are testing APIs.

PR #1493, opened 2019-04-15
[android_alarm_manager] Added comments and refactored android_alarm_manager plugin for clarity.

matthew-carroll:
## Description

I started reading through android_alarm_manager to determine how it would work with the new Android plugin API but I had a difficult time tracing the execution paths and interpreting the intent of the code. So I decided to go through and document everything I could to make better sense of it. I'd like to commit those comments and minor refactorings so that hopefully it's a little more clear for everyone.

Added comments and refactored android_alarm_manager plugin for clarity.

- Added many comments, both on public members and within private source code.
- Moved all static behavior above instance behavior in `AlarmService` to clearly demonstrate how much behavior has been defined statically (which I think we very much want to move to instance behavior).
- Renamed some methods to shift them from ambiguous naming to explicit.
- Created request data structures to parse and hold information about desired one-shot and periodic tasks.

Recommendations for further refactoring:

- I think we should move all the plugin behavior out of `AlarmService` and into its own object because `AlarmService` lives and dies based on the OS, regardless of whether that timing is convenient for our plugin behavior.
- On a similar note, we should adjust the API to work with multiple Flutter execution contexts instead of just one. This will be a requirement for add-to-app scenarios, regardless of what the new Android embedding API looks like.
- I would recommend codifying the method channel messages in its own object instead of combining it with the associated logic. This will also make it much easier in the future to replace hand-coded message handling with a generated version.

## Checklist

- [x ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x ] I updated/added relevant documentation (doc comments with `///`).
- [x ] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [ x] I read and followed the [Flutter Style Guide].
- [x ] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x ] I signed the [CLA].
- [x ] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

- [ x] No, this is *not* a breaking change.

### DIFF
--- a/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
+++ b/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
@@ -164,7 +164,7 @@ public class AndroidAlarmManagerPlugin implements MethodCallHandler, ViewDestroy
   }
 
   /** A request to schedule a one-shot Dart task. */
-  final static class OneShotRequest {
+  static final class OneShotRequest {
     static OneShotRequest fromJson(JSONArray json) throws JSONException {
       int requestCode = json.getInt(0);
       boolean exact = json.getBoolean(1);
@@ -173,7 +173,8 @@ public class AndroidAlarmManagerPlugin implements MethodCallHandler, ViewDestroy
       boolean rescheduleOnReboot = json.getBoolean(4);
       long callbackHandle = json.getLong(5);
 
-      return new OneShotRequest(requestCode, exact, wakeup, startMillis, rescheduleOnReboot, callbackHandle);
+      return new OneShotRequest(
+          requestCode, exact, wakeup, startMillis, rescheduleOnReboot, callbackHandle);
     }
 
     final int requestCode;
@@ -200,7 +201,7 @@ public class AndroidAlarmManagerPlugin implements MethodCallHandler, ViewDestroy
   }
 
   /** A request to schedule a periodic Dart task. */
-  final static class PeriodicRequest {
+  static final class PeriodicRequest {
     static PeriodicRequest fromJson(JSONArray json) throws JSONException {
       int requestCode = json.getInt(0);
       boolean exact = json.getBoolean(1);
--- a/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
+++ b/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
@@ -19,32 +19,33 @@ import org.json.JSONException;
 
 /**
  * Flutter plugin for running one-shot and periodic tasks sometime in the future on Android.
- * <p>
- * Plugin initialization goes through these steps:
+ *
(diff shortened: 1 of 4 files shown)


packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
```diff
@@ -73,49 +149,115 @@ public void onMethodCall(MethodCall call, Result result) {
     }
   }
 
-  private void startService(JSONArray arguments) throws JSONException {
-    long callbackHandle = arguments.getLong(0);
-    AlarmService.setCallbackDispatcher(mContext, callbackHandle);
-    AlarmService.startAlarmService(mContext, callbackHandle);
+  /**
+   * Transitions the Flutter execution context that owns this plugin from foreground execution
+   * to background execution.
```
ened: @matthew-carroll A question about this: Is there any case or special handling required when an App transitions from background to foreground execution? E.g. there's a long-running alarm manager callback and the user meanwhile opens the App?

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
```diff
@@ -16,25 +17,76 @@
...
-    channel.setMethodCallHandler(plugin);
-    backgroundChannel.setMethodCallHandler(plugin);
+    alarmManagerPluginChannel.setMethodCallHandler(plugin);
+    backgroundCallbackChannel.setMethodCallHandler(plugin);
     registrar.addViewDestroyListener(plugin);
-    AlarmService.setBackgroundChannel(backgroundChannel);
+
+    // The AlarmService expects to hold a static reference to the plugin's background
+    // method channel.
+    // TODO(mattcarroll): this static reference implies that only one instance of this plugin
```
ened: It's unlikely that this happens with alarm manager, but some plugins may offer background and foreground execution. In which case this will break, isn't it?

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
matthew-carroll: I don't believe there are any such race conditions. By the time the `BroadcastReceiver` runs, the application class and app process must exist. At that point, either an Activity is displayed, or not. So as long as you account for those 2 possibilities, I think everything should be OK. But I welcome any insights about counter-examples.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
matthew-carroll: Unfortunately, I think the entire concept of background vs foreground is a bit ambiguous at the moment. But to your specific point, can you provide an example what you think would cause a breakage here?

matthew-carroll:
@mklim @jason-simmons @bkonyi - Do any of you have any concerns or comments about this PR, outside of the minor formatting complaints coming from PR validation?

PR #1505, opened 2019-04-17
[image_picker] Fix path of returned File objects when picking videos

LinusU:
## Description

Returning `.description` on a `NSURL` gave back an url that starts with `file:///`, when passing this url to the `File` object it constructed a path like `/file:///foo/bar/baz.ext`.

This fixes this logic to use `.path` instead, which returns the path to the file.

Here is some more background to how I found this: https://github.com/firebase/firebase-ios-sdk/pull/2458/files#r276273826

## Checklist

- [x] I signed the [CLA].

## Breaking Change

-----

Not sure how to add tests since the tests since to mock out the Objective-C part of the plugins, any help would be appreciated!

### DIFF
--- a/packages/image_picker/CHANGELOG.md
+++ b/packages/image_picker/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.6.0
+
+* Breaking change iOS: Returned `File` objects when picking videos now always holds the correct path. Before this change, the path returned could have `file://` prepended to it.
+
 ## 0.5.4+3
 
 * Fix the example app failing to load picked video.
@@ -16,7 +20,7 @@
 
 ## 0.5.3+2
 
-* Android: fix a crash when the MainActivity is destroyed after selecting the image/video. 
+* Android: fix a crash when the MainActivity is destroyed after selecting the image/video.
 
 ## 0.5.3+1
 
--- a/packages/image_picker/ios/Classes/ImagePickerPlugin.m
+++ b/packages/image_picker/ios/Classes/ImagePickerPlugin.m
@@ -191,7 +191,7 @@ static const int SOURCE_GALLERY = 1;
     return;
   }
   if (videoURL != nil) {
-    _result(videoURL.description);
+    _result(videoURL.path);
   } else {
     if (image == nil) {
       image = [info objectForKey:UIImagePickerControllerOriginalImage];
--- a/packages/image_picker/pubspec.yaml
+++ b/packages/image_picker/pubspec.yaml
@@ -5,7 +5,7 @@ authors:
   - Flutter Team <flutter-dev@googlegroups.com>
   - Rhodes Davis Jr. <rody.davis.jr@gmail.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/image_picker
-version: 0.5.4+3
+version: 0.6.0
 
 flutter:
   plugin:


collinjackson:
I tried manually testing this in the iOS simulator and unfortunately I'm getting "Error Loading Video"

Screen Shot 2019-04-20 at 19 59 11

cyanglaz:
> I tried manually testing this in the iOS simulator and unfortunately I'm getting "Error Loading Video" (both before and after the change)
> 
> Screen Shot 2019-04-20 at 19 59 11

Just curious, how did you test camera with simulator?

collinjackson:
> > I tried manually testing this in the iOS simulator and unfortunately I'm getting "Error Loading Video" (both before and after the change)
> 
> Just curious, how did you test camera with simulator?

I didn't, I just dragged and dropped a MOV file onto the simulator to get it into the videos library.

cyanglaz:
@LinusU Thank you for providing this. the .description returns the absolute path to the source. However .path seems to be more relevant to iOS file system and Apple seems suggesting to use .path.

Could you provide the use case that the absolute path failed? It seems that dart doesn't care which way the file path was expressed and works for both paths.

Also, it is probably a breaking change since the format of the returning value is changed, could you update the minor version? to 0.6.0?

LinusU:
> Thank you for providing this. the .description returns the absolute path to the source. However .path seems to be more relevant to iOS file system and Apple seems suggesting to use .path.

Since this is an `NSURL` instance, it returns an absolute _url_, not path. So the problem is that it returned `file:///foo/bar/baz`, instead of `/foo/bar/baz`.

When creating a dart `File` with the string `file:///foo/bar/baz`, it will not create a file that points to `/foo/bar/baz`, but rather a file that points to `/file:///foo/bar/baz`, which won't work.

> Could you provide the use case that the absolute path failed? It seems that dart doesn't care which way the file path was expressed and works for both paths.

I'm getting an error that Dart cannot read the file `/file:///foo/bar/baz`, I don't see how this can have ever worked, unless iOS 12.2 changed the behaviour of `.description` on `NSURL`.

> Also, it is probably a breaking change since the format of the returning value is changed, could you update the minor version? to 0.6.0?

I don't think it's a breaking change since this works around an error. Even if `description` previously returned the file path without prepending `file:///` in front of it, it shouldn't change now...

Hope this clears it up ☺️ 

I'll rebase on `master` to fix the conflicts! 🚀

cyanglaz:
> I'm getting an error that Dart cannot read the file `/file:///foo/bar/baz`, I don't see how this can have ever worked, unless iOS 12.2 changed the behaviour of `.description` on `NSURL`.

I just checked with the the current flutter, it is able to successfully read the file at `file:///....`, if you run the example app on image picker, does it pick up the file?

wreppun:
I can confirm this happens for us as well.  We have 

```
File _stripUriScheme(File file) {
  if (file.path.startsWith("file:///")) {
    final filePath = file.path.substring(7);
    debugPrint('File URL pattern dropped... path is now: $filePath');
    return File(filePath);
  }
  // N.B. file://<some-host>/<some-path> is not handled
  // ...but I don't think that happens in practice
  return file;
}
```

as a workaround until this gets merged.

IIRC, it was as simple as running

```
File f = await ImagePicker.pickVideo(source: ImageSource.gallery);
file.lengthSync();
```

cyanglaz:
> I can confirm this happens for us as well. We have
> 
> ```
> File _stripUriScheme(File file) {
>   if (file.path.startsWith("file:///")) {
>     final filePath = file.path.substring(7);
>     debugPrint('File URL pattern dropped... path is now: $filePath');
>     return File(filePath);
>   }
>   // N.B. file://<some-host>/<some-path> is not handled
>   // ...but I don't think that happens in practice
>   return file;
> }
> ```
> as a workaround until this gets merged.
> 
> IIRC, it was as simple as running
> 
> ```
> File f = await ImagePicker.pickVideo(source: ImageSource.gallery);
> file.lengthSync();
> ```

@LinusU As commented here by @wreppun, people might be doing some surgery on the return value. So it is probably a breaking change for them. I still think we should bump the minor version and make it 0.6.0.

collinjackson:
I agree, let's update the CHANGELOG with a description of new behavior and call it 0.6.0

LinusU:
Rebased and changed to breaking change 👌

packages/image_picker/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.6.0
+
+* Breaking change: Path of returned File objects when picking videos does no longer start with `file:///`, but are instead a normal absolute path.
```
cyanglaz: please specify it is on iOS platform. Also the path starts with `file:///` is the absolute path, you might want to update the wording a little bit?

LinusU: Is this better?

> Breaking change iOS: Returned `File` objects when picking videos now always holds an absolute path. Before this change, a `File` object with a path beginning with `/file:///` could be returned.

cyanglaz: Maybe 
```Breaking change iOS: Returned `File` objects when picking videos now always holds the correct path. Before this change, the absolute path (Begin with file:///) was returned.```

Comment (LinusU) [reply packages/image_picker/CHANGELOG.md]: > (Begin with file:///)

just to clarify, the path returned on the `File` object (e.g. `print(file.path)`) starts with a single forward slash, and then `file:`, followed by three forward slashes. At least on my iPhone 8, iOS 12.2.

This is the code that we are currently using in our app to work around it:

```dart
      if (file.path.startsWith('/file:///')) {
        file = File(file.path.substring(8));
      }
```

LinusU: Changed to

> Breaking change iOS: Returned `File` objects when picking videos now always holds the correct path. Before this change, the path returned could have `file://` prepended to it.

What do you think?

cyanglaz: > > (Begin with file:///)
> 
> just to clarify, the path returned on the `File` object (e.g. `print(file.path)`) starts with a single forward slash, and then `file:`, followed by three forward slashes. At least on my iPhone 8, iOS 12.2.
> 
> This is the code that we are currently using in our app to work around it:
> 
> ```dart
>       if (file.path.startsWith('/file:///')) {
>         file = File(file.path.substring(8));
>       }
> ```

yeah that was a typo of mine, you can change it to /file:/// in the changelog.

PR #1503, opened 2019-04-17
Update firebase_auth CocoaPod dependency

collinjackson:
See discussion in https://github.com/flutter/plugins/commit/8bdd77661c1cfc899eaad9cd14b9222611f49a39

Changes to firebase_auth require a "~> 5.4" version of FirebaseAuth CocoaPod to ensure that `FIRAuthErrorUserInfoNameKey` is available.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [X] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #1508, opened 2019-04-17
[firebase_auth] Increase iOS Firebase/Auth CocoaPods dependency to 5.19

collinjackson:
This is necessary to pick up the FirebaseAuth 5.4 CocoaPod.

This will actually achieve what I was trying to achieve in https://github.com/flutter/plugins/pull/1503

See discussion in 8bdd776


collinjackson:
Merging without waiting for the iOS bots since they're going to time out anyway

PR #1510, opened 2019-04-18
[webview] Webview supports setting headers via initialRequest

iskakaushik:
## Description

WebView allows setting headers.

## Related Issues

https://github.com/flutter/flutter/issues/25831

### DIFF
--- a/packages/webview_flutter/lib/request.dart
+++ b/packages/webview_flutter/lib/request.dart
@@ -1,0 +1,37 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Represents a request made by the [WebView] widget.
+///
+/// See also:
+/// * [WebView.initialRequest] for how this gets used.
+class Request {
+  /// Constructs a [Request] object with the given [url]
+  /// and [headers].
+  const Request({
+    this.url,
+    this.headers,
+  });
+
+  /// The URL to load.
+  final String url;
+
+  /// The headers  to be passed in when making the request.
+  final Map<String, String> headers;
+
+  Map<String, dynamic> toMap() {
+    final Map<String, dynamic> optionsMap = <String, dynamic>{};
+
+    void addIfNonNull(String fieldName, dynamic value) {
+      if (value != null) {
+        optionsMap[fieldName] = value;
+      }
+    }
+
+    addIfNonNull('url', url);
+    addIfNonNull('headers', headers);
+
+    return optionsMap;
+  }
+}
--- a/packages/webview_flutter/lib/webview_flutter.dart
+++ b/packages/webview_flutter/lib/webview_flutter.dart
@@ -9,6 +9,8 @@ import 'package:flutter/gestures.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 
+import 'request.dart';
+
 typedef void WebViewCreatedCallback(WebViewController controller);
 
 enum JavascriptMode {
@@ -108,10 +110,14 @@ class WebView extends StatefulWidget {
   /// `onWebViewCreated` callback once the web view is created.
   ///
   /// The `javascriptMode` parameter must not be null.
+  ///
+  /// initialUrl has been deprecated, use [initialRequest] instead.
   const WebView({
     Key key,
     this.onWebViewCreated,
-    this.initialUrl,
+    // ignore: deprecated_member_use_from_same_package
+    @deprecated this.initialUrl,
+    this.initialRequest,
     this.javascriptMode = JavascriptMode.disabled,
     this.javascriptChannels,
     this.navigationDelegate,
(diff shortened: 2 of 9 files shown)


packages/webview_flutter/lib/request.dart
```diff
@@ -0,0 +1,37 @@
...
+  /// and [headers].
+  const Request({
+    this.url,
+    this.headers,
+  });
+
+  /// The URL to load.
+  final String url;
+
+  /// The headers  to be passed in when making the request.
```
collinjackson: There's an extra space here

packages/webview_flutter/lib/request.dart
```diff
@@ -0,0 +1,37 @@
...
+/// Represents a request made by the [WebView] widget.
+///
+/// See also:
+/// * [WebView.initialRequest] for how this gets used.
+class Request {
+  /// Constructs a [Request] object with the given [url]
+  /// and [headers].
+  const Request({
+    this.url,
+    this.headers,
```
collinjackson: Consider initializing this to an empty mutable Map if not provided. Compare to package:http

https://github.com/dart-lang/http/blob/master/lib/src/base_request.dart#L85

packages/webview_flutter/lib/request.dart
```diff
@@ -0,0 +1,37 @@
...
+
+/// Represents a request made by the [WebView] widget.
+///
+/// See also:
+/// * [WebView.initialRequest] for how this gets used.
+class Request {
+  /// Constructs a [Request] object with the given [url]
+  /// and [headers].
+  const Request({
+    this.url,
```
collinjackson: I would make this @required and assert non-null

packages/webview_flutter/lib/request.dart
```diff
@@ -0,0 +1,37 @@
...
+    this.headers,
+  });
+
+  /// The URL to load.
+  final String url;
+
+  /// The headers  to be passed in when making the request.
+  final Map<String, String> headers;
+
+  Map<String, dynamic> toMap() {
```
collinjackson: I would hide this from the public API with underscore

packages/webview_flutter/lib/request.dart
```diff
@@ -0,0 +1,37 @@
...
+    final Map<String, dynamic> optionsMap = <String, dynamic>{};
+
+    void addIfNonNull(String fieldName, dynamic value) {
```
collinjackson: Could you include all the fields even if they're null and check for null on the native side? Or assert that `headers != null` if you implement my suggestion above about having a default value.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -108,10 +110,14 @@ class WebView extends StatefulWidget {
   /// `onWebViewCreated` callback once the web view is created.
   ///
   /// The `javascriptMode` parameter must not be null.
+  ///
+  /// initialUrl has been deprecated, use [initialRequest] instead.
```
collinjackson: square brackets around `initialUrl`

packages/webview_flutter/test/webview_flutter_test.dart
```diff
@@ -47,9 +49,23 @@ void main() {
...
+      ),
+    );
+
+    expect(await controller.currentUrl(), 'https://youtube.com');
+  });
+
   testWidgets('Javascript mode', (WidgetTester tester) async {
     await tester.pumpWidget(const WebView(
-      initialUrl: 'https://youtube.com',
+      initialRequest: Request(url: 'https://youtube.com'),
```
collinjackson: const?

packages/webview_flutter/test/webview_flutter_test.dart
```diff
@@ -59,7 +75,7 @@ void main() {
     expect(platformWebView.javascriptMode, JavascriptMode.unrestricted);
 
     await tester.pumpWidget(const WebView(
-      initialUrl: 'https://youtube.com',
+      initialRequest: Request(url: 'https://youtube.com'),
```
collinjackson: const?

packages/webview_flutter/test/webview_flutter_test.dart
```diff
@@ -382,7 +398,7 @@ void main() {
   testWidgets('Cookies can be cleared once', (WidgetTester tester) async {
     await tester.pumpWidget(
       const WebView(
-        initialUrl: 'https://flutter.io',
+        initialRequest: Request(url: 'https://flutter.io'),
```
collinjackson: const?

packages/webview_flutter/test/webview_flutter_test.dart
```diff
@@ -394,7 +410,7 @@ void main() {
       (WidgetTester tester) async {
     await tester.pumpWidget(
       const WebView(
-        initialUrl: 'https://flutter.io',
+        initialRequest: Request(url: 'https://flutter.io'),
```
collinjackson: const?

packages/webview_flutter/test/webview_flutter_test.dart
```diff
@@ -597,7 +613,7 @@ void main() {
 
     testWidgets('onPageFinished is null', (WidgetTester tester) async {
       await tester.pumpWidget(const WebView(
-        initialUrl: 'https://youtube.com',
+        initialRequest: Request(url: 'https://youtube.com'),
```
collinjackson: const?

amirh:
We should probably add the option to set request headers in the controller's loadUrl method (initialUrl is a convenient parameter that can only be used for the initial page to load, `WebViewController#loadUrl` can be used for any page including the first one).

I would even consider starting with adding the headers option just in a controller method and keeping initialUrl just for url as it is there just for making the most simple use case more convenient. Not saying we shouldn't do it, just to consider if it's common enough to make the simple common use case a little more boilerplate-y (another alternative that might be worth considering is adding another optional `initialHeaders` widget parameter?).

SakinaBoriwala21:
Hey @iskakaushik Can I use this in my App? And when will this be merged? Thanks so much.

namratajain44:
Hi, 

When I try to replace initialUrl with initialRequest, I get error saying "The named parameter 'initialRequest' isn't defined."

Is this feature implemented in the webview_flutter library yet. I am using version(webview_flutter: ^2.0.0)

Thanks,
Namrata

PR #1512, opened 2019-04-18
[Image_picker] retrieve lost image.

cyanglaz:
## Description

On Android, the MainActivity can be killed and restarted after launching the image picker for various of reasons. this can be tested by selecting `don't keep activities alive' option in the developer options in phone's setting app.

This PR adds a method `retrieveLostData` to retrieve the lost image/videos.

## Related Issues

flutter/flutter#17950

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/image_picker/README.md
+++ b/packages/image_picker/README.md
@@ -89,3 +89,5 @@ Future<void> retrieveLostData() async {
   }
 }
 ```
+
+There's no way to detect when this happens, so calling this method at the right place is essential. We recommend to wire this into some kind of start up check. Please refer to the example app to see how we used it.
--- a/packages/image_picker/README.md
+++ b/packages/image_picker/README.md
@@ -65,7 +65,7 @@ class _MyHomePageState extends State<MyHomePage> {
 }
 ```
 
-### Handling MainActivity Destruction on Android
+### Handling MainActivity destruction on Android
 
 Android system -- although very rarely -- sometimes kills the MainActivity after the image_picker finishes. When this happens, we lost the data selected from the image_picker. You can use `retrieveLostData` to retrieve the lost data in this situation. For example: 
 
--- a/packages/image_picker/README.md
+++ b/packages/image_picker/README.md
@@ -64,3 +64,28 @@ class _MyHomePageState extends State<MyHomePage> {
   }
 }
 ```
+
+### Handling MainActivity Destruction on Android
+
+Android system -- although very rarely -- sometimes kills the MainActivity after the image_picker finishes. When this happens, we lost the data selected from the image_picker. You can use `retrieveLostData` to retrieve the lost data in this situation. For example: 
+
+```dart
+Future<void> retrieveLostData() async {
+  final RetrieveLostDataResponse response =
+      await ImagePicker.retrieveLostData();
+  if (response == null) {
+    return;
+  }
+  if (response.file != null) {
+    setState(() {
+      if (response.type == RetrieveType.video) {
+        _handleVideo(response.file);
+      } else {
+        _handleImage(response.file);
+      }
+    });
+  } else {
+    _handleError(response.exception);
+  }
+}
+```
--- a/packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerCache.java
+++ b/packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerCache.java
@@ -1,0 +1,159 @@
+package io.flutter.plugins.imagepicker;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.net.Uri;
+import androidx.annotation.Nullable;
+import io.flutter.plugin.common.MethodCall;
+import java.util.HashMap;
+import java.util.Map;
+
+class ImagePickerCache {
+
+  static final String MAP_KEY_PATH = "path";
+  static final String MAP_KEY_MAX_WIDTH = "maxWidth";
+  static final String MAP_KEY_MAX_HEIGHT = "maxHeight";
+  private static final String MAP_KEY_TYPE = "type";
+  private static final String MAP_KEY_ERROR_CODE = "errorCode";
+  private static final String MAP_KEY_ERROR_MESSAGE = "errorMessage";
+
+  private static final String FLUTTER_IMAGE_PICKER_IMAGE_PATH_KEY =
+      "flutter_image_picker_image_path";
+  private static final String SHARED_PREFERENCE_ERROR_CODE_KEY = "flutter_image_picker_error_code";
+  private static final String SHARED_PREFERENCE_ERROR_MESSAGE_KEY =
+      "flutter_image_picker_error_message";
+  private static final String SHARED_PREFERENCE_MAX_WIDTH_KEY = "flutter_image_picker_max_width";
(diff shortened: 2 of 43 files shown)


packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -423,20 +462,28 @@ public void onPathReady(String path) {
   }
 
   private void handleImageResult(String path, boolean shouldDeleteOriginalIfScaled) {
-    if (pendingResult != null) {
+    if (methodCall != null) {
       Double maxWidth = methodCall.argument("maxWidth");
```
mklim: Now that we're parsing these in multiple places, would it make sense to parse them once and then store them somewhere instead? Reduces the surface where we could get typing errors here.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -476,4 +529,117 @@ private void clearMethodCallAndResult() {
     methodCall = null;
     pendingResult = null;
   }
+
+  private void saveResult(String path, String errorCode, String errorMessage) {
```
mklim: optional: `@Nullable` annotations on these params?

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -476,4 +529,117 @@ private void clearMethodCallAndResult() {
...
+  }
+
+  private void setType(String type) {
+    if (ImagePickerPlugin.getFilePref == null) {
+      return;
+    }
+    ImagePickerPlugin.getFilePref.edit().putString(SHARED_PREFERENCE_TYPE_KEY, type).apply();
+  }
+
+  private void persistsPendingCameraMediaUriPath(String path) {
```
mklim: nit: `persistPendingCameraMediaUriPath`. That way it matches the same tense as the rest of the methods.

packages/image_picker/lib/image_picker.dart
```diff
@@ -75,4 +85,83 @@ class ImagePicker {
...
+  /// successful image/video selection, or a failure.
+  ///
+  /// Returns null if there is no lost data.
+  ///
+  /// Calling this on a non-Android platform will throw [UnimplementedError] exception.
+  ///
+  /// See also:
+  /// * [RetrieveLostDataResponse], for what's included in the response.
+  /// * [Android Activity Lifecycle](https://developer.android.com/reference/android/app/Activity.html), for more information on MainActivity destruction.
+  static Future<RetrieveLostDataResponse> retrieveLostData() async {
```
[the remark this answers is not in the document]
cyanglaz: Optional is a good choice but I don't want to bring in that just for this since users would need to learn Optional just for this. Some might not be too familiar with Optional and rather deal with null.

I also thought about returning an empty response with `path` to be null. However, In this case the assertion of `assert(result.containsKey('path') ^ result.containsKey('errorCode'));` would fail and I do want to keep that there as a debug check.

Maybe I can add another variable in the RetrieveLostDataResponse: `isEmpty` to represent empty response?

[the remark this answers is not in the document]
mklim: Sure, `isEmpty` sounds good too.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -182,11 +194,22 @@ public void onScanCompleted(String path, Uri uri) {
...
+    }
+    cleanPendingResultPref();
+  }
+
   public void chooseVideoFromGallery(MethodCall methodCall, MethodChannel.Result result) {
     if (!setPendingMethodCallAndResult(methodCall, result)) {
       finishWithAlreadyActiveError(result);
       return;
     }
+    setType("video");
```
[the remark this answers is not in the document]
cyanglaz: > Also: is there any way we could unit all the cache logic into a single point and always route through it? Right now it's scattered throughout all these different methods as we're going through the processes depending on the specific image/video type, and then at different points in the process of picking them. It would be hard for anybody adding an extra type here or changing these methods to realize the full flow and realize that they need to extend it or maintain it for whatever extra channel that they're building.

Strongly agree. I am also 'disgusted' by all these scattered code of handling persistent storage right now. I might have to completely change my code structure here or even refactor the old code a bit in order to do so. I will ping you once I have some idea.

cyanglaz:
@mklim Updated with your suggestions. Could you do another pass?

packages/image_picker/README.md
```diff
@@ -64,3 +64,28 @@ class _MyHomePageState extends State<MyHomePage> {
   }
 }
 ```
+
+### Handling MainActivity destruction on Android
+
+Android system -- although very rarely -- sometimes kills the MainActivity after the image_picker finishes. When this happens, we lost the data selected from the image_picker. You can use `retrieveLostData` to retrieve the lost data in this situation. For example:
```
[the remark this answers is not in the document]
mklim: I think this suggestion still applies.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerCache.java
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  static void saveDemensionWithMethodCall(MethodCall methodCall) {
+    Double maxWidth = methodCall.argument("maxWidth");
+    Double maxHeight = methodCall.argument("maxHeight");
+    setMaxDimension(maxWidth, maxHeight);
+  }
+
+  private static void setMaxDimension(Double maxWidth, Double maxHeight) {
+    if (getFilePref != null) {
```
mklim: subjective: consider early returns/guard clauses here instead of the extra nesting

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -182,6 +183,33 @@ public void onScanCompleted(String path, Uri uri) {
     this.fileUtils = fileUtils;
   }
 
+  void saveStateBeforeResult() {
+    if (methodCall != null) {
```
mklim: Same subjective early return suggestion here.

packages/image_picker/lib/image_picker.dart
```diff
@@ -75,4 +88,97 @@ class ImagePicker {
...
+
+  /// Retrieve the lost image file when [pickImage] or [pickVideo] failed because the  MainActivity is destroyed. (Android only)
+  ///
+  /// Image or video can be lost if the MainActivity is destroyed. And there is no guarantee that the MainActivity is always alive.
+  /// Call this method to retrieve the lost data and process the data according to your APP's business logic.
+  ///
+  /// Returns a [LostDataResponse] if successfully retrieved the lost data. The [LostDataResponse] can represent either a
+  /// successful image/video selection, or a failure.
+  ///
+  /// Returns null if there is no lost data.
```
mklim: nit: this line no longer applies

PR #1516, opened 2019-04-19
[webview_flutter] controller headers loadurl

iskakaushik:
# Description
WebView controller allows setting headers.

# Related Issues
flutter/flutter#25831

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.6
+
+* Add an optional `headers` field to the controller.
+
 ## 0.3.5+5
 
 * Fixed error in documentation of `javascriptChannels`.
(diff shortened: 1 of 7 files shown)


collinjackson:
Per our conversation - since Android requires HTTPS and scooterlabs doesn't support that, maybe postman-echo.com would work.

iskakaushik:
@collinjackson , postman echo was not echoing all headers. I deployed https://github.com/iskakaushik/flutter-header-echo/blob/master/index.js as a heroku app, to do this. Let me know if you think that is concerning for any reason.

collinjackson:
I think this approach is fine for now. If we wanted to have fewer dependencies we could run a small Dart server locally (either on the host or in the simulator) and connect to it from the test, but that would come with its own set of issues and it's probably best to not to spend too much time on this one test when we have a lot of other tests to write.

PR #1520, opened 2019-04-21
[cloud_functions] update Dart API to replace call with getHttpsCallable

collinjackson:
Updates cloud_functions API to more closely match native SDK API

## Breaking Change

- [X] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).


packages/cloud_functions/lib/src/https_callable.dart
```diff
@@ -0,0 +1,53 @@
...
+  final String _functionName;
+
+  Future<HttpsCallableResult> call([dynamic parameters]) async {
+    try {
+      final MethodChannel channel = CloudFunctions.channel;
+      final dynamic response =
+          // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+          // https://github.com/flutter/flutter/issues/26431
+          // ignore: strong_mode_implicit_dynamic_method
+          await channel.invokeMethod('CloudFunctions#call', <String, dynamic>{
```
bparrishMines: nit: You could use the new `invokeMethod<T>` here and remove the comment above.

packages/cloud_functions/lib/src/https_callable.dart
```diff
@@ -0,0 +1,53 @@
...
+        'functionName': _functionName,
+        'parameters': parameters,
+      });
+      return HttpsCallableResult._(response);
+    } on PlatformException catch (e) {
+      if (e.code == 'functionsError') {
+        final String code = e.details['code'];
+        final String message = e.details['message'];
+        final dynamic details = e.details['details'];
+        print('throwing firebase functions exception');
```
bparrishMines: nit: Is this print needed?

packages/cloud_functions/lib/src/https_callable.dart
```diff
@@ -0,0 +1,53 @@
...
+        throw CloudFunctionsException._(code, message, details);
+      } else {
+        print('throwing generic exception');
```
bparrishMines: nit: this one too

packages/cloud_functions/lib/src/https_callable.dart
```diff
@@ -0,0 +1,53 @@
...
+        throw Exception('Unable to call function ' + _functionName);
+      }
+    } catch (e) {
+      print(e);
```
bparrishMines: nit: this one too

packages/cloud_functions/lib/src/https_callable.dart
```diff
@@ -0,0 +1,53 @@
...
+      rethrow;
+    }
+  }
+
+  Duration _timeout;
+
+  Duration get timeout => _timeout;
+  set duration(Duration value) {
```
bparrishMines: Should this be:

```suggestion
  set timeout(Duration value) {
```

packages/cloud_functions/lib/src/https_callable.dart
collinjackson: it's not, it's left over from previous implementation. removing it, thanks.

packages/cloud_functions/lib/src/https_callable.dart
collinjackson: it's not, it's left over from previous implementation. removing it, thanks.

packages/cloud_functions/lib/src/https_callable.dart
collinjackson: it's not, it's left over from previous implementation. removing it, thanks.

PR #1528, opened 2019-04-22
[webview_flutter] Remove un-used method params

iskakaushik:


packages/webview_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.6+1
+
+* Remove un-used method params in webview\_flutter
```
mklim: nit:

```suggestion
* Remove un-used method params in webview_flutter
```

cyanglaz:
```suggestion
* Remove un-used method params.
```

iskakaushik: in markdown '_' needs to be escaped, or it waits for it to be closed as an underline.

mklim: Oh, right. My mistake!

PR #1531, opened 2019-04-22
[image_picker] example app video load error fix.

cyanglaz:
Fixes the example app loading video error.
https://github.com/flutter/plugins/pull/1505#issuecomment-485220511

### DIFF
--- a/packages/image_picker/example/lib/main.dart
+++ b/packages/image_picker/example/lib/main.dart
@@ -37,7 +37,7 @@ class _MyHomePageState extends State<MyHomePage> {
   dynamic _pickImageError;
   bool isVideo = false;
   VideoPlayerController _controller;
-  VoidCallback listener = () {};
+  VoidCallback listener;
   String _retrieveDataError;
 
   void _onImageButtonPressed(ImageSource source) async {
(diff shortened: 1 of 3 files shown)


packages/image_picker/example/lib/main.dart
```diff
@@ -85,6 +85,14 @@ class _MyHomePageState extends State<MyHomePage> {
     super.dispose();
   }
 
+  @override
+  void initState() {
+    super.initState();
+    listener = () {
```
collinjackson: It seems like `listener` is set in `initState` and never changed after that.

Would it make sense to make this a private method instead of a mutable VoidCallback member data? e.g. `void _onChanged() { setState(() { }); }`

Or if we wanted to update the API, the VideoPlayerController could expose a `Stream` of `onChanged` events and we could use a `StreamListener` to update the widget tree.

cyanglaz: Exposing stream from video player is def better. For this PR since it is only to fix the image picker's example app, let's keep it in sync with the current video player public API.

We cannot use a private method here because the video player controller needs to remove listener. So we'd have to have a variable to keep track of the callback.

collinjackson: > Exposing stream from video player is def better. For this PR since it is only to fix the image picker's example app, let's keep it in sync with the current video player public API.

Sounds good, let's land this bug fix and then we can update the API in a future change. Filed flutter/flutter#31504

> We cannot use a private method here because the video player controller needs to remove listener. So we'd have to have a variable to keep track of the callback.

Is there something different about an immutable tear-off method that makes it not possible to remove compared to a mutable `VoidCallback` variable?

PR #1535, opened 2019-04-23
[webview_flutter] Test: wait for page to load before checking the content

iskakaushik:
https://github.com/flutter/flutter/issues/31510 seems to exist. Using it to make the header loading test not flake


collinjackson:
Let's remove the skip from the test and release as a new version if the bots are happy

PR #1491, opened 2019-04-14
[cloud_firestore] Support for FieldValue.increment

collinjackson:
## Description

This PR adds support for incrementing using `FieldValue`.

Example usage:

```dart
      final DocumentReference ref = firestore.collection('mycollection').document();
      await ref.setData(<String, dynamic>{
        'value': 1,
        'created_at': FieldValue.serverTimestamp(),
      });
      await ref.updateData(<String, dynamic>{
        'value': FieldValue.increment(1),
      });
      DocumentSnapshot snapshot = await ref.get();
      expect(snapshot.data['value'], 2);
```

I thought about naming the API `incrementDouble/incrementInteger` but leaning towards overloading `increment`. I'd love to get feedback on this.

## Related Issues

Fixes flutter/flutter#31034

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [X] No, this is *not* a breaking change.

PR #1543, opened 2019-04-26
[cloud_firestore] Fix wrong FieldValue

creativecreatorormaybenot:
Slight mistake that occured in https://github.com/flutter/plugins/pull/1491.

As version `0.10.0` has not even been rolled out to Pub yet, I do not think that I need to bump the version.

Still, I am wondering why `0.10.0` is still not on Pub as it has been 3 days.


creativecreatorormaybenot:
@collinjackson

collinjackson:
Thanks for the fix @creativecreatorormaybenot. I've updated the integration test to ensure that it can catch this bug.

We're working on a better system (#27258) for ensuring that changes are published promptly after merges, although it looks like in this case it was helpful to get your additional feedback.

PR #1552, opened 2019-04-29
[video_player] Fix player initialization and other warnings

iskakaushik:
1. When a player has not initialized it's presentationSize is
   returned as CGSizeZero.

2. The enum type must be AVPlayerItemStatusFailed not AVPlayerStatusFailed

3. Fix a few other IDE warnings.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/19197

### DIFF
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.10.0+8
+
+* iOS: Fix an issue where the player sends initialization message incorrectly.
+
+* Fix a few other IDE warnings.
+
+
 ## 0.10.0+7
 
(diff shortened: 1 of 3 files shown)


quentinleguennec:
Hi @iskakaushik @cyanglaz ,
This part of the PR is breaking our app:
```
// The player has not yet initialized.
    if (height == CGSizeZero.height && width == CGSizeZero.width) {
      return;
    }
```

We are using the video_player to play mp3 files (since there is no well maintained audio plugin). It was working perfectly but since the file have a width and height of 0 this `if` breaks the plugin for us on iOS. Could we remove this limitation?

khietlam:
I'm using this video_player to play mp3 files as @quentinleguennec . Now it breaks my app. 

Please help with this issue. Thanks.

PR #1553, opened 2019-04-29
[google_maps_flutter] Avoid calling null callbacks

matthewlloyd:
## Description

Avoid calling null callbacks.

There is no requirement that the callback parameters need to be non-null,
so without these checks, the plugin can generate exceptions, e.g. when
overlays that have no onTap callback are clicked.

Also guards against overlay or marker IDs not being present in the
dictionaries, which could happen by race condition (e.g. a click event
could be slightly delayed and get handled after the associated object has
already been removed).

## Checklist

- [x] I signed the [CLA].

## Breaking Change


iskakaushik:
Curious, when would this get called on a `markerId` that has been removed?

Also, is there a way to write an integration test for this over at: https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/example/test_driver/google_maps.dart ?

matthewlloyd:
Suppose a marker (or polyline, soon polygon, circle, etc.) gets tapped: the tap event is received on the iOS side of the bridge and forwarded and put on Dart's event queue for processing by some thread (I don't know enough about iOS / Flutter architecture to specify which one). Then suppose that at the same time, before Dart handles that message, it either is already processing, or now processes, some other previously queued event or state change that causes the marker to be removed from _markers. After the marker is removed, the tap event doesn't also get removed from Dart's event queue, so onMarkerTap would be called right after the marker has been removed from _markers. It is unlikely, but possible, and perhaps the additional null check it's unnecessary if someone can prove that it cannot happen in today's (or tomorrow's) Flutter architecture. However, given the concurrent nature of Flutter's architecture, with event loops on both sides of a message-passing bridge, it may not be a risk worth taking, and the null check is cheap anyway.

I think it would be hard to reproduce a race condition like that reliably an integration test...

iskakaushik:
@matthewlloyd thanks for the explanation. LGTM

PR #1556, opened 2019-05-01
[firebase_perfomance] iOS side of reformat of Firebase Performance

bparrishMines:
## Description

iOS side of possible reformat of Firebase Performance plugin

## Related Issues

Android: https://github.com/flutter/plugins/pull/1529
Dart: https://github.com/flutter/plugins/pull/1530

## Breaking Change


packages/firebase_performance/ios/Classes/FirebasePerformancePlugin.h
```diff
@@ -1,4 +1,21 @@
+#import <Firebase/Firebase.h>
 #import <Flutter/Flutter.h>
 
 @interface FLTFirebasePerformancePlugin : NSObject <FlutterPlugin>
++ (void)addMethodHandler:(NSNumber *_Nonnull)handle
```
collinjackson: By splitting things into multiple files, you're exposing a lot more class and method surface area here than we were previously. This could become a maintenance burden in the future if people use these APIs and we have to worry about breaking them. I think we should try to make the classes/methods private. Perhaps move `addMethodHandler` to a file named  `FirebasePerformancePlugin+Internal.h`.

It might also make sense to move FirebasePerformancePlugin.h into a Public/ subfolder, if you can get that to work with the build system.

See how Firebase CocoaPods are laid out for reference:

https://github.com/firebase/firebase-ios-sdk/tree/master/Functions/FirebaseFunctions

packages/firebase_performance/ios/Classes/FLTFirebasePerformance.m
```diff
@@ -0,0 +1,101 @@
...
+
+@interface FLTFirebasePerformance ()
+@property FIRPerformance *performance;
+@end
+
+@implementation FLTFirebasePerformance
++ (void)registerWithRegistrar:(nonnull NSObject<FlutterPluginRegistrar> *)registrar {
+}
+
++ (void)sharedInstanceWithCall:(FlutterMethodCall *)call result:(FlutterResult)result {
```
collinjackson: From an Objective-C developer perspective, I feel this name could be confusing. Normally in Objective-C `sharedInstance` would be an idempotent static method that returns a singleton instance of the class. Here, you're performing a non-idempotent allocation action and returning nothing. Also you're not doing anything with the result argument. Perhaps another name would make it more clear what's going on.

packages/firebase_performance/ios/Classes/FirebasePerformancePlugin+Internal.h
```diff
@@ -0,0 +1,23 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FirebasePerformancePlugin.h"
+
+@interface FLTFirebasePerformancePlugin (Internal)
++ (void)addMethodHandler:(NSNumber *_Nonnull)handle
+           methodHandler:(id<FlutterPlugin> _Nonnull)handler;
```
collinjackson: I am not sure how I feel about using `id<FlutterPlugin>` as the type here. The empty `registerWithRegistrar` implementations feel like this is a hack that we wouldn't want to propagate to other plugins. These objects aren't really `FlutterPlugin`s, they are proxied classes. It might be a good idea to define a protocol makes this clear, perhaps something we can re-use for other plugins in the future.

packages/firebase_performance/example/test_driver/firebase_performance.dart
```diff
@@ -2,14 +2,15 @@ import 'dart:async';
...
 import 'package:flutter_test/flutter_test.dart';
-import 'package:firebase_performance/firebase_performance.dart';
 
 void main() {
   final Completer<String> completer = Completer<String>();
   enableFlutterDriverExtension(handler: (_) => completer.future);
   tearDownAll(() => completer.complete(null));
 
   group('firebase_performance test driver', () {
+    // TODO(bparrishMines): Rewrite integration tests when iOS portion is written.
```
collinjackson: Reminder to address this

packages/firebase_performance/ios/Classes/FirebasePerformancePlugin+Internal.h
bparrishMines: I was thinking the same thing when I was implementing the iOS side. On Android, they are called `MethodHandler`s and don't have static methods. 

I could create a separate protocol such as

```
@protocol MethodHandler
@required
- (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result
@end


@interface FLTTrace : NSObject <MethodHandler>
@end
```

bparrishMines:
Closing in favor of https://github.com/flutter/plugins/pull/1557

PR #1558, opened 2019-05-01
[google_maps_flutter] Android: update myLocationButton preference on map load

iskakaushik:
## Related Issues

* fixes https://github.com/flutter/flutter/issues/31930


iskakaushik:
@amirh I added a test for this at https://github.com/flutter/plugins/pull/1350, once we have tests running on Android, we will be able to catch errors like these before hand.

PR #1565, opened 2019-05-03
[path_provider] Release as 1.0 and add integration tests

collinjackson:
Fixes flutter/flutter#32042

## Breaking Change

This is not a breaking change but I'm updating the version to 1.0 since the plugin is mature enough for that.

### DIFF
--- a/packages/path_provider/example/test_driver/path_provider.dart
+++ b/packages/path_provider/example/test_driver/path_provider.dart
@@ -46,5 +46,4 @@ void main() {
       expect(result.path.length, isNonZero);
     }
   });
-
 }
--- a/packages/path_provider/example/test_driver/path_provider.dart
+++ b/packages/path_provider/example/test_driver/path_provider.dart
@@ -1,0 +1,50 @@
+// Copyright 2019, the Chromium project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'dart:io';
+import 'package:flutter_driver/driver_extension.dart';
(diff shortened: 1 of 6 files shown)


mklim:
Not seeing any new integration tests here?

packages/path_provider/example/test_driver/path_provider.dart
```diff
@@ -0,0 +1,50 @@
...
+import 'package:path_provider/path_provider.dart';
+import 'package:uuid/uuid.dart';
+
+void main() {
+  final Completer<String> allTestsCompleter = Completer<String>();
+  enableFlutterDriverExtension(handler: (_) => allTestsCompleter.future);
+  tearDownAll(() => allTestsCompleter.complete(null));
+
+  test('getTemporaryDirectory', () async {
+    Directory result = await getTemporaryDirectory();
```
mklim: Do you think it worth trying to check the path name to make sure that it's actually the temp directory and not something else? Same for the other function calls below. I'm not a huge fan of hardcoding what we think those directories should be and then casing off of the platform and asserting that it matches, it would be very brittle, but right now all these are really testing is that we're getting a valid path at all.

collinjackson:
> Not seeing any new integration tests here?

Checked in the missing folder, PTAL

packages/path_provider/example/test_driver/path_provider.dart
collinjackson: Yeah, I thought about that, but I don't think we want to check the path name here. As you pointed out, it would be testing platform-specific implementation details that could (harmlessly) change in future updates. Further testing would make the test more brittle.

collinjackson:
Note that there's a follow-up to this change in https://github.com/flutter/plugins/pull/1566

PR #1566, opened 2019-05-03
[path_provider] add getApplicationSupportDirectory

collinjackson:
Fixes flutter/flutter#13474

This allows developers to store files that aren't user-generated in the correct folder on iOS.

Also updated the documentation:

1. push devs towards using `getStorageDirectory` on Android when storing user-generated content that should be visible to the user.
1. encourage the use of `getTemporaryDirectory` for storing caches of files that don't need to be backed up.

See https://developer.apple.com/icloud/documentation/data-storage/index.html

/cc @sroddy

### DIFF
--- a/packages/path_provider/lib/path_provider.dart
+++ b/packages/path_provider/lib/path_provider.dart
@@ -47,8 +47,7 @@ Future<Directory> getApplicationSupportDirectory() async {
   if (path == null) {
     return null;
   }
-  Directory result = Directory(path);
-  return result;
+  return Directory(path);
 }
 
 /// Path to a directory where the application may place data that is
--- a/packages/path_provider/lib/path_provider.dart
+++ b/packages/path_provider/lib/path_provider.dart
@@ -39,13 +39,18 @@ Future<Directory> getTemporaryDirectory() async {
 /// On iOS, this uses the `NSApplicationSupportDirectory` API.
 ///
 /// On Android, this uses the `getDataDirectory` API on the context.
+///
+/// If the directory does not exist, it will be created.
 Future<Directory> getApplicationSupportDirectory() async {
   final String path =
       await _channel.invokeMethod<String>('getApplicationSupportDirectory');
   if (path == null) {
     return null;
   }
-  return Directory(path);
+  Directory result = Directory(path);
+  if (!result.existsSync())
+    result.createSync();
+  return result;
 }
 
 /// Path to a directory where the application may place data that is
--- a/packages/path_provider/lib/path_provider.dart
+++ b/packages/path_provider/lib/path_provider.dart
@@ -44,7 +44,7 @@ Future<Directory> getApplicationSupportDirectory() async {
   if (!Platform.isIOS)
     throw UnsupportedError("getApplicationSupportDirectory requires iOS");
   final String path =
-  await _channel.invokeMethod<String>('getApplicationSupportDirectory');
+      await _channel.invokeMethod<String>('getApplicationSupportDirectory');
   if (path == null) {
     return null;
   }
--- a/packages/path_provider/lib/path_provider.dart
+++ b/packages/path_provider/lib/path_provider.dart
@@ -52,7 +52,9 @@ Future<Directory> getApplicationSupportDirectory() async {
 ///
 /// On iOS, this uses the `NSDocumentDirectory` API.
 ///
-/// On Android, this uses the `getDataDirectory` API on the context.
+/// On Android, this uses the `getDataDirectory` API on the context. Consider
+/// using getExternalStorageDirectory instead if data is intended to be visible
+/// to the user.
 Future<Directory> getApplicationDocumentsDirectory() async {
   final String path =
       await _channel.invokeMethod<String>('getApplicationDocumentsDirectory');
--- a/packages/path_provider/lib/path_provider.dart
+++ b/packages/path_provider/lib/path_provider.dart
@@ -50,7 +50,8 @@ Future<Directory> getApplicationSupportDirectory() async {
 /// Path to a directory where the application may place data that is
 /// user-generated, or that cannot otherwise be recreated by your application.
 ///
-/// On iOS, this uses the `NSDocumentDirectory` API.
+/// On iOS, this uses the `NSDocumentDirectory` API. Consider using
+/// [getApplicationSupportDirectory] instead if the data is not user-generated.
 ///
 /// On Android, this uses the `getDataDirectory` API on the context. Consider
 /// using getExternalStorageDirectory instead if data is intended to be visible
--- a/packages/path_provider/lib/path_provider.dart
+++ b/packages/path_provider/lib/path_provider.dart
@@ -21,7 +21,8 @@ const MethodChannel _channel =
 ///
 /// On Android, this uses the `getCacheDir` API on the context.
 Future<Directory> getTemporaryDirectory() async {
-  final String path = await _channel.invokeMethod<String>('getTemporaryDirectory');
+  final String path =
+      await _channel.invokeMethod<String>('getTemporaryDirectory');
   if (path == null) {
     return null;
   }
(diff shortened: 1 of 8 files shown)


collinjackson:
It looks like this directory isn't automatically created. Should we create it automatically? I've done that to make the integration test pass, @mklim please let me know if you have opinions.

packages/path_provider/lib/path_provider.dart
```diff
@@ -21,29 +22,46 @@ const MethodChannel _channel =
...
+///
+/// Use this for files you don’t want exposed to the user. Your app should not
+/// use this directory for user data files.
+///
+/// On iOS, this uses the `NSApplicationSupportDirectory` API.
+/// This directory is not guaranteed to exist and you can call `createSync` to
+/// create it if it doesn't exist.
+///
+/// On Android, this uses the `getDataDirectory` API on the context.
+Future<Directory> getApplicationSupportDirectory() async {
```
mklim: A couple things:

* I think it's worth mentioning how to check that this exists too in the DartDoc if we're expecting users to check for it every time.
* If this is getting translated automatically to some other Android concept when it's called there, not sure how I feel about the iOS specific name. I'd almost rather it non-op or throw on Android if we're trying to map it to a specific iOS folder directly. WDYT?

packages/path_provider/lib/path_provider.dart
```diff
@@ -21,29 +22,46 @@ const MethodChannel _channel =
...
+}
+
+/// Path to a directory where the application may place application support
+/// files.
+///
+/// Use this for files you don’t want exposed to the user. Your app should not
+/// use this directory for user data files.
+///
+/// On iOS, this uses the `NSApplicationSupportDirectory` API.
+/// This directory is not guaranteed to exist and you can call `createSync` to
```
mklim: nit:

```suggestion
/// This directory is not guaranteed to exist. You can call `createSync` to
```

packages/path_provider/lib/path_provider.dart
```diff
@@ -21,29 +22,46 @@ const MethodChannel _channel =
...
-/// is deleted.
+/// Path to a directory where the application may place data that is
+/// user-generated, or that cannot otherwise be recreated by your application.
 ///
-/// On iOS, this uses the `NSDocumentsDirectory` API.
+/// On iOS, this uses the `NSDocumentDirectory` API. Consider using
+/// [getApplicationSupportDirectory] instead if the data is not user-generated.
 ///
-/// On Android, this returns the AppData directory.
+/// On Android, this uses the `getDataDirectory` API on the context. Consider
```
mklim: (May be out of scope for this PR): It seems like this method is supposed to be used differently on iOS vs Android in a kind of confusing way:

* iOS: use for user generated data
* Android: use for any persistent data that should be invisible to the user

Those use cases aren't really correlated. I feel like they shouldn't really be bundled into the same API if they're really supposed to be used this way, they're different enough that just casing off of Platform version to try and use these correctly would be difficult. I may be reading the docs wrong?

packages/path_provider/lib/path_provider.dart
```diff
@@ -57,14 +75,12 @@ Future<Directory> getApplicationDocumentsDirectory() async {
 /// On iOS, this function throws an UnsupportedError as it is not possible
 /// to access outside the app's sandbox.
 ///
-/// On Android this returns getExternalStorageDirectory.
+/// On Android this returns the external storage directory.
```
mklim: I think the previous comment actually had a specific meaning that gets lost by making this plain English, it was referencing https://developer.android.com/reference/android/os/Environment#getExternalStorageDirectory(). Maybe link it here still?

packages/path_provider/lib/path_provider.dart
collinjackson: I'm fine with having it throw on Android for now.

packages/path_provider/lib/path_provider.dart
collinjackson: I may not be articulating the use case for these paths very effectively in the docs. I'm ok with making the Android and iOS methods separate rather than trying to shoehorn the two platforms onto an imperfect lowest common denominator.

collinjackson:
Review feedback addressed, PTAL

nrkramer:
As an update:

This works on Android and returns a path:
'/data/user/0/apps.blanket.showtap/files'

Yet still throws an exception:
E/MethodChannel#plugins.flutter.io/path_provider(18577): Failed to handle method call
E/MethodChannel#plugins.flutter.io/path_provider(18577): java.lang.IllegalStateException: Reply already submitted
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at io.flutter.embedding.engine.dart.DartMessenger$Reply.reply(DartMessenger.java:135)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler$1.notImplemented(MethodChannel.java:235)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at io.flutter.plugins.pathprovider.PathProviderPlugin.onMethodCall(PathProviderPlugin.java:44)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:222)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at io.flutter.embedding.engine.dart.DartMessenger.handleMessageFromDart(DartMessenger.java:96)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at io.flutter.embedding.engine.FlutterJNI.handlePlatformMessage(FlutterJNI.java:643)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at android.os.MessageQueue.nativePollOnce(Native Method)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at android.os.MessageQueue.next(MessageQueue.java:326)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at android.os.Looper.loop(Looper.java:160)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at android.app.ActivityThread.main(ActivityThread.java:6669)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at java.lang.reflect.Method.invoke(Native Method)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
E/MethodChannel#plugins.flutter.io/path_provider(18577):        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
E/DartMessenger(18577): Uncaught exception in binary message listener   
E/DartMessenger(18577): java.lang.IllegalStateException: Reply already submitted
E/DartMessenger(18577):         at io.flutter.embedding.engine.dart.DartMessenger$Reply.reply(DartMessenger.java:135)
E/DartMessenger(18577):         at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:240)
E/DartMessenger(18577):         at io.flutter.embedding.engine.dart.DartMessenger.handleMessageFromDart(DartMessenger.java:96)
E/DartMessenger(18577):         at io.flutter.embedding.engine.FlutterJNI.handlePlatformMessage(FlutterJNI.java:643)
E/DartMessenger(18577):         at android.os.MessageQueue.nativePollOnce(Native Method)
E/DartMessenger(18577):         at android.os.MessageQueue.next(MessageQueue.java:326)
E/DartMessenger(18577):         at android.os.Looper.loop(Looper.java:160)
E/DartMessenger(18577):         at android.app.ActivityThread.main(ActivityThread.java:6669)
E/DartMessenger(18577):         at java.lang.reflect.Method.invoke(Native Method)
E/DartMessenger(18577):         at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
E/DartMessenger(18577):         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)

PR #973, opened 2018-12-08
[firebase_auth] Ability to choose between sign-in or link on auto verification of phone number on Android

cqhchan:
- Fixed Android Link with credentials bug.
- Added in ability to choose between sign-in or link on auto verification of phone number

PR #1568, opened 2019-05-05
[firebase_auth] Removed automatic signing in behaviour when signing in with phone auth.

awazgyawali:
This is a fix to issue https://github.com/flutter/flutter/issues/22304.

The plugin demands auth credential while logging in with facebook, google, emailPassword. Which gives developer the option to create new account with it or link it to the existing user. But the phone auth used to behave differently, it directly used to sign in and create new user which was preventing developers to link it to the existing user. Thus this PR removes that behaviour and gives ability to extract phone auth cred and use it as required.

I have used Gson to parse and send the PhoneAuthCredential to the dart side. Not sure if its a good idea, suggestions are welcomed.

Note: This automatic signing in for phone auth was not enabled on the IOS side thus no changes have been made to the IOS code, let me know if I am missing something here.

To use this PR, copy the following lines to your pubspec.yaml file.
```  
firebase_auth:
    git:
      url: git://github.com/awazgyawali/plugins/
      path: packages/firebase_auth
```

### DIFF
--- a/packages/firebase_auth/CHANGELOG.md
+++ b/packages/firebase_auth/CHANGELOG.md
@@ -1,6 +1,8 @@
+## 0.9.0
 
 * **Breaking change**: `PhoneVerificationCompleted` now provides an `AuthCredential` that can
   be used with `signInWithCredential` or `linkWithCredential` instead of signing in automatically.
+* **Breaking change**: Remove internal counter `nextHandle` from public API.
 
 ## 0.8.4+5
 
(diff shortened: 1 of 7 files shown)


awazgyawali:
I think this is a very small and serious thing to address as the SDK is supposed to behave similarly to all the possible ways to sign in. Also, this PR will break the previous user's application as it will not sign in automatically and the user has to `signInWithCredential`  themselves after the callback.

creativecreatorormaybenot:
This actually seems better than https://github.com/flutter/plugins/pull/973.

PR #1569, opened 2019-05-06
[firebase_dynamic_links] support clicking on link while app is running.

roeierez:
## Background:
On Android there is a different behavior than in iOS which I think is the expected one.
1. When clicking on a link while the app is running and then trying to get the link via a call to "retrieveDynamicLink" the result is null.
2. Calling to "retrieveDynamicLink" doesn't clean the old value which makes it impossible for the application to know if/when the user has clicked a link or it is a  result of an old click.

To solve the first problem I added a check for the existence of a dynamic link on every new intent instead of the current behavior that checks only on the fluter activity intent.

In addition (for the second problem) I added a cleanup for the old link  every time it was retrieved
successfully by the user.

This PR matches the behavior on iOS.

## Breaking Change
The `retrieveDynamicLinks` will now clean the value that it returns, means it will return a specific link only once.

### DIFF
--- a/packages/firebase_dynamic_links/CHANGELOG.md
+++ b/packages/firebase_dynamic_links/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.4.0+1
+
+* Fixed bug where link persists after starting an app with a Dynamic Link.
+* Fixed bug where retrieving a link would fail when app was already running.
+
 ## 0.4.0
 
 * Update dependency on firebase_core to 0.4.0.
--- a/packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
+++ b/packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
@@ -24,22 +24,20 @@ import java.util.Map;
 /** FirebaseDynamicLinksPlugin */
 public class FirebaseDynamicLinksPlugin implements MethodCallHandler {
   private Registrar registrar;
-  private Map<String, Object> dynamicLinkData;
-  private Task extractIntentLinkTask;
+  private Intent latestIntent;
 
   private FirebaseDynamicLinksPlugin(Registrar registrar) {
     this.registrar = registrar;
+    latestIntent = registrar.activity().getIntent();
 
     registrar.addNewIntentListener(
         new PluginRegistry.NewIntentListener() {
           @Override
           public boolean onNewIntent(Intent intent) {
-            checkLinkOnIntent(intent);
+            latestIntent = intent;
             return false;
           }
         });
-
-    checkLinkOnIntent(registrar.activity().getIntent());
   }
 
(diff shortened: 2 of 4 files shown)


bparrishMines:
Hi @roeierez, 

Thanks for the contribution!

 You said that you were unable to retrieve the link while the app was already running. The plugin should already be able to handle this. Where did you click a link that led to `retrieveDynamicLink` to be null? It may fail in a specific case. Using `registrar.activity().getIntent()` should suffice as the `FlutterActivity` is the only activity in a full Flutter app.
 
Also, according to the [documentation](https://firebase.google.com/docs/dynamic-links/android/receive), the call to `getDynamicLink()` already clears the data on first call. We modeled the iOS side to behave what the Android side should do automatically.

roeierez:
Hi @bparrishMines,

Thanks for your quick reply!

> You said that you were unable to retrieve the link while the app was already running. The plugin should already be able to handle this. Where did you click a link that led to `retrieveDynamicLink` to be null? It may fail in a specific case. Using `registrar.activity().getIntent()` should suffice as the `FlutterActivity` is the only activity in a full Flutter app.

Hi, sorry for not being clear. The case is exactly as follows:
1. open the app.
2. click the link.
3. on resuming, call to retrieveDynamicLink. - the result is null.

I think the problem resides in the fact that the code always use `registrar.activity().getIntent()` which is the intent that started the flutter activity and not the one that was created when the user clicks the link. 
flutter activity is configured as singleTop `android:launchMode="singleTop"` and  according to android documentation (https://developer.android.com/reference/android/app/Activity.html#onNewIntent(android.content.Intent)) in these cases the onNewIntent should be called for the new click intent. I am using "singleTask" but I don't think it matters for this discussion.
In the PR I am checking this intent via the `onNewIntent` listener which contains the link data.



> Also, according to the [documentation](https://firebase.google.com/docs/dynamic-links/android/receive), the call to `getDynamicLink()` already clears the data on first call

Yes I see that although from my experience (I tested again) after I start the app with a link every call to "retrieveDynamicLink" returns the same link without cleaning it.

Baileypollard:
I’m having this exact issue on IOS. Android is working fine for me. Could this be a problem with our fire base configuration?

bparrishMines:
Hi @roeierez,

I was unable to reproduce the bug for the first problem. Based on your explanation and the documentation for onNewIntent, you should be correct. If this PR solves this problem for you, then we can go ahead and submit.

I was able to reproduce the second bug and changed your code a little to cache the intent rather than storing the task and data.

bparrishMines:
@Baileypollard which problem are you running into on iOS?

> 1. When clicking on a link while the app is running and then trying to get the link via a call to "retrieveDynamicLink" the result is null.
> 2. Calling to "retrieveDynamicLink" doesn't clean the old value which makes it impossible for the application to know if/when the user has clicked a link or it is a  result of an old click.

Would you mind creating a separate issue for iOS? If not, I can create it.

roeierez:
@bparrishMines thanks. Caching the intent looks good to me.

DharmaTech123:
Still facing the same issue

PR #1570, opened 2019-05-06
[path_provider] Add Desktop Support for `getApplicationDocumentsDirectory`

AppleEducate:
## Description

Returns the correct path for MacOS, Windows and Linux when running Flutter on the desktop on the call `getApplicationDocumentsDirectory()`.

Inspired by @lesnitsky


collinjackson:
Thanks for the PR!

If we're going to support this plugin on Linux and Mac, we should probably come up with a documented behavior for all the methods. Would you like to take a crack at the other ones?

It feels like the approach of doing some of the work in Dart and some work native-land might become challenging to maintain over time, but I do see value in exposing these paths.

See also flutter/plugins#1566

AppleEducate:
@collinjackson Sure thing! I'll see what I can do.

The reason it is split right now between dart and native land is because there is no official way for plugin maintainers to add plugin desktop support that i am aware of, and each plugin comes but with their own way. If there was an official way to publish a plugin package to support desktop with platform channels then i agree it should go there an be in Dart.

stuartmorgan:
The desktop platform have platform channel support, it's just not stable yet. Working around that within the official plugin repository is not the right path forward here (and won't be possible in general, so this doesn't establish an approach that can be replicated). Instead we should be adding desktop support the right way, as the APIs stabilize and the tooling for pulling the native sides matures.

If people experimenting with desktop before things stabilize want to use a plugin, they can implement the native side locally using the [current platform channel support](https://github.com/google/flutter-desktop-embedding/blob/master/plugins/README.md), or in cases like this they can do the workarounds themselves in their Dart code above the level of the plugin's Dart code.

lesnitsky:
@AppleEducate seems like this one should live [here](https://github.com/google/flutter-desktop-embedding/tree/master/plugins/flutter_plugins) for now

lesnitsky:
@stuartmorgan maybe worth moving #1716 to FDE repo? I wasn't aware of this PR and was about to start reimplementing the same. IMO having desktop "adapters" for plugins in FDE repo is totally working

lesnitsky:
> This PR isn't a correct implementation

I should have been more specific in first comment, I meant actual implementation of path_provider in swift, so agree

lesnitsky:
Thanks! One more step forward to real "unified" codebases for mobile and
desktop apps 🔥🎉

stuartmorgan:
> This PR isn't a correct implementation though (`${Platform.environment['HOME']}/.config` is not the documents directory on macOS, for instance).

@collinjackson I would recommend closing the PR given this.

collinjackson:
Closing per @stuartmorgan's recommendation.

PR #1572, opened 2019-05-06
Update Flutterfire Android dependencies.

collinjackson:
This updates Flutterfire plugins to the latest Firebase Android releases.

All the plugins that have firebase_core dependencies have been updated to the firebase_core release that was just published (https://github.com/flutter/plugins/pull/1576)
Updated firebase_auth dependency on firebase_dynamic_links to fix compile failures.
Updated firebase_messaging to handle breaking change on Android.


creativecreatorormaybenot:
@collinjackson I am not exactly sure why, however, all Android builds that contain this commit, i.e. `firebase_analytics` `3.0.0` & `3.0.1` fail.  
Gradle builds fail on the following task: `checkDuplicateClasses`  
I can replicate this very consistently. I changed dependencies around and cleared Pub and Gradle caches for about an hour now and only rolling `firebase_analytics` back to `2.1.1+3` solves the issue.

Also, you might want to check out https://github.com/flutter/flutter/issues/32031.

PR #1578, opened 2019-05-07
[firebase_auth] Update to latest CocoaPod

collinjackson:
Most of our Flutterfire plugins don't specify a particular version, but right now we're using the optimistic operator with `Firebase/Auth`, '~> 5.19'. This change gets us onto Firebase 6.x, which should increase the chance of find compatible pods for other plugins that are using 6.x stuff in the future.

I think I need to get Cirrus to run `pod repo update` and pick up 6.0 before this change will pass tests. The CocoaPods just went live a few hours ago.

### DIFF
--- a/packages/firebase_auth/CHANGELOG.md
+++ b/packages/firebase_auth/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.10.1
+## 0.10.0+1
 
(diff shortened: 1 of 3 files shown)


collinjackson:
Tests are passing now that the CocoaPod is live.

PR #1581, opened 2019-05-07
Fix break in firebase_analytics CocoaPod

collinjackson:
Fixes flutter/flutter#32252.

Resolves compile errors due to removal of the `FIRAnalyticsConfiguration` class in CocoaPod version 6.0.

### DIFF
--- a/packages/firebase_analytics/ios/Classes/FirebaseAnalyticsPlugin.m
+++ b/packages/firebase_analytics/ios/Classes/FirebaseAnalyticsPlugin.m
@@ -57,7 +57,7 @@
     result(nil);
   } else if ([@"setAnalyticsCollectionEnabled" isEqualToString:call.method]) {
     NSNumber *enabled = [NSNumber numberWithBool:call.arguments];
-    [[FIRAnalyticsConfiguration sharedInstance] setAnalyticsCollectionEnabled:[enabled boolValue]];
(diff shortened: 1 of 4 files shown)


packages/firebase_analytics/ios/Classes/FirebaseAnalyticsPlugin.m
```diff
@@ -57,7 +57,7 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
     result(nil);
   } else if ([@"setAnalyticsCollectionEnabled" isEqualToString:call.method]) {
     NSNumber *enabled = [NSNumber numberWithBool:call.arguments];
-    [[FIRAnalyticsConfiguration sharedInstance] setAnalyticsCollectionEnabled:[enabled boolValue]];
+    [FIRAnalytics setAnalyticsCollectionEnabled:[enabled boolValue]];
```
bparrishMines: Does this method exist in the previous pod version? It would be a breaking change, if no.

collinjackson: Looks like it was moved here fairly recently.

mollywoodnini:
Can you please publish your changes?

robisaks:
@collinjackson can you post here when 3.0.1 is pushed to pub.dev? Looks like this is blocking a lot of active development.

collinjackson:
Done, thanks for your patience.

PR #1582, opened 2019-05-08
[firebase_core] set user agent

kroikie:
## Description

Send the user agent on Android side of firebase_core plugin.


collinjackson:
Relates to flutter/flutter#32267

lgtm, we can test that this is kept up to date in a subsequent PR.

collinjackson:
I pushed an iOS implementation and script that checks that the value is kept up to date, PTAL

script/check_hard_coded_version.sh
```diff
@@ -0,0 +1,48 @@
...
+  for package_name in "$@"; do
+    local dir="$REPO_DIR/packages/$package_name"
+    echo "Checking that $package_name has the correct hard coded version, if any."
+    PACKAGE_VERSION="$(cd "$dir" && cat pubspec.yaml | grep -E "^version: " | awk '{print $2}')"
+    IOS_VERSION="$(cd "$dir" && grep -r "#define LIBRARY_VERSION" ios/Classes/*.m | awk '{print $3}')"
+    ANDROID_VERSION="$(cd "$dir" && grep -r LIBRARY_VERSION android/src/main/java/* | awk '{print $8}')"
+    if [[ "$IOS_VERSION" == 0 && "$ANDROID_VERSION" == 0 ]]; then
+      echo "No hard coded version found"
+    elif [[ "$IOS_VERSION" == "@\"$PACKAGE_VERSION\"" && "$ANDROID_VERSION" == "\"$PACKAGE_VERSION\";" ]]; then
+      error "Hard coded version matched: $PACKAGE_VERSION"
```
kroikie: should we use echo here?

packages/firebase_core/ios/Classes/FirebaseCorePlugin.m
```diff
@@ -6,6 +6,9 @@
 
 #import <Firebase/Firebase.h>
 
+#define LIBRARY_NAME @"flutter-firebase_core"
```
davidair: Can we do "flutter-fire-core" instead?

packages/firebase_core/android/src/main/java/io/flutter/plugins/firebase/core/FlutterFirebaseAppRegistrar.java
```diff
@@ -0,0 +1,18 @@
+package io.flutter.plugins.firebase.core;
+
+import com.google.firebase.components.Component;
+import com.google.firebase.components.ComponentRegistrar;
+import com.google.firebase.platforminfo.LibraryVersionComponent;
+import java.util.Collections;
+import java.util.List;
+
+public class FlutterFirebaseAppRegistrar implements ComponentRegistrar {
+  private static final String LIBRARY_NAME = "flutter-firebase_core";
```
davidair: I'd recommend "flutter-fire-core"

PR #1585, opened 2019-05-09
Return image picker method call results on the platform thread

jason-simmons:
The MediaScanConnection.scanFile callback is invoked on a background thread.
However, MethodChannel results must be provided on the platform thread.

Fixes https://github.com/flutter/flutter/issues/32315


cyanglaz:
Actually, we also need to update the CHANGELOG and pubspec.yaml

jason-simmons:
Updated version to 0.6.0+3 and added a release note

cyanglaz:
@jason-simmons Looks good! Thank you for doing this.

PR #1590, opened 2019-05-10
[in_app_purchase] Propagate launchBillingFlow failures

mklim:
## Description

Previously the buy methods were just returning `void`, so the immediate
response of `BillingClient#launchBillingFlow` was unaccessible. Change
it to return a `Future<bool>` and return whether or not the launch flow
succeeded in google_play_connection. Always return true in
app_store_connection since there's no equivalent status there.

## Related Issues

flutter/flutter#31417
Fixes flutter/flutter#32518
flutter/flutter/issues/32512

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/in_app_purchase/lib/src/in_app_purchase/app_store_connection.dart
+++ b/packages/in_app_purchase/lib/src/in_app_purchase/app_store_connection.dart
@@ -44,20 +44,21 @@ class AppStoreConnection implements InAppPurchaseConnection {
   Future<bool> isAvailable() => SKPaymentQueueWrapper.canMakePayments();
 
   @override
-  void buyNonConsumable({@required PurchaseParam purchaseParam}) {
-    _skPaymentQueueWrapper.addPayment(SKPaymentWrapper(
+  Future<bool> buyNonConsumable({@required PurchaseParam purchaseParam}) async {
+    await _skPaymentQueueWrapper.addPayment(SKPaymentWrapper(
         productIdentifier: purchaseParam.productDetails.id,
         quantity: 1,
         applicationUsername: purchaseParam.applicationUserName,
         simulatesAskToBuyInSandbox: purchaseParam.sandboxTesting,
         requestData: null));
+    return true; // There's no error feedback from iOS here to return.
   }
 
   @override
-  void buyConsumable(
+  Future<bool> buyConsumable(
       {@required PurchaseParam purchaseParam, bool autoConsume = true}) {
     assert(autoConsume == true, 'On iOS, we should always auto consume');
-    buyNonConsumable(purchaseParam: purchaseParam);
+    return buyNonConsumable(purchaseParam: purchaseParam);
   }
 
   @override
--- a/packages/in_app_purchase/lib/src/in_app_purchase/in_app_purchase_connection.dart
+++ b/packages/in_app_purchase/lib/src/in_app_purchase/in_app_purchase_connection.dart
@@ -81,14 +81,18 @@ abstract class InAppPurchaseConnection {
   /// You always need to restore all the non consumable products for user when
   /// they switch their phones.
   ///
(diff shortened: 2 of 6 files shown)


packages/in_app_purchase/lib/src/in_app_purchase/app_store_connection.dart
```diff
@@ -44,20 +44,21 @@ class AppStoreConnection implements InAppPurchaseConnection {
...
-  void buyNonConsumable({@required PurchaseParam purchaseParam}) {
-    _skPaymentQueueWrapper.addPayment(SKPaymentWrapper(
+  Future<bool> buyNonConsumable({@required PurchaseParam purchaseParam}) async {
+    await _skPaymentQueueWrapper.addPayment(SKPaymentWrapper(
         productIdentifier: purchaseParam.productDetails.id,
         quantity: 1,
         applicationUsername: purchaseParam.applicationUserName,
         simulatesAskToBuyInSandbox: purchaseParam.sandboxTesting,
         requestData: null));
+    return true; // There's no error feedback from iOS here to return.
```
mklim: May be worth catching any `PlatformException`s here and returning false in that case? I didn't add it because I thought the behavior was a little too special cased, but that may be better than always returning true.

packages/in_app_purchase/lib/src/in_app_purchase/in_app_purchase_connection.dart
```diff
@@ -81,14 +81,18 @@ abstract class InAppPurchaseConnection {
...
+  /// triggering this method, purchase updates will be sent to
+  /// [purchaseUpdatedStream]. You should [Stream.listen] to
+  /// [purchaseUpdatedStream] to get [PurchaseDetails] objects in different
+  /// [PurchaseDetails.status] and update your UI accordingly. When the
+  /// [PurchaseDetails.status] is [PurchaseStatus.purchased] or
+  /// [PurchaseStatus.error], you should deliver the content or handle the
+  /// error. On iOS, you also need to call [completePurchase] to finish the
+  /// purchasing process.
+  ///
+  /// This method does return whether or not the purchase request was initially
```
mklim: The purchases going to the Stream instead of being returned in an immediate Future is already fairly unidiomatic for Dart, even though I do think it makes the most sense here given our options. However I'm worried that this change is going to make it even more confusing, since now this method _would_ be returning a `Future`... but not a `Future` that actually has to do with the final purchase outcome, just whether the initial request to show a purchase dialog to the user went through. I'm not sure what the right call to try and expose that extra info without being extra confusing is here.

packages/in_app_purchase/lib/src/in_app_purchase/app_store_connection.dart
cyanglaz: Yeah, there shouldn't be any exception to be caught here. If there is an exception here, that means the developer has to fix it.

packages/in_app_purchase/lib/src/in_app_purchase/in_app_purchase_connection.dart
cyanglaz: I think it is necessary to have it return a BOOL. It looks good with the updated documentation.
The other alternative that I can think of is to create a unified response code to specify the response just like Android but I feel like it would make things more confusing.

LinusU:
@mklim did you consider returning a `Future<void>` that rejects with an error in case initiating the flow doesn't succeed?

I think that as it is now, it will be easy to accidentally miss handling this. E.g. since this function used to return `void`, I think that many people will not check the return value, and thus the error will go silently ignored instead of showing up as an unhandled exception in the debugger.

In general I feel that returning `bool`s wheter an error occured or not isn't the best API design in a language that has native exceptions (and, in this case, native `Future`s with rejections).

Maybe I've missed something though? ☺️

PR #1591, opened 2019-05-10
[cloud_firestore] Fixed startAt-, startAfter-, endAt-, endBeforeDocument

creativecreatorormaybenot:
I am not exactly sure why this is the case. Someone who has the ability might want to contact the Cloud Firestore team on this, however, this appears to be a recent change to Cloud Firestore that fundamentally changed how indexes work.  
Ordering by `documentID` ascending does not work (anymore), check https://github.com/flutter/flutter/issues/32031. Now, there is also no index required when ordering by `documentID` descending.

I have only tested this for Android and only changed the Android specific code in this PR. You can just change the iOS code as well if you want to test that.


collinjackson:
The solution looks ok to me, I'm a little surprised that our pagination integration test didn't notice the issue though. I'd like to have test coverage for this sort of regression. It might be that the issue is Android only.

collinjackson:
I thought that perhaps that the reason why this wasn't showing up in our tests is that our pagination integration test doesn't use a compound query.

I've updating the test to try to get it to repro this bug. First I got test failures like this:

```
I/flutter ( 9116): 00:04 +3: Firestore pagination [E]
I/flutter ( 9116):   PlatformException(Error performing getDocuments, FAILED_PRECONDITION: The query requires an index. You can create it here: <LINK>
I/flutter ( 9116):   package:flutter/src/services/message_codecs.dart 564:7                                       StandardMethodCodec.decodeEnvelope
I/flutter ( 9116):   package:flutter/src/services/platform_channel.dart 302:33                                    MethodChannel.invokeMethod
I/flutter ( 9116):   ===== asynchronous gap ===========================
I/flutter ( 9116):   dart:async/future_impl.dart 22:43                                                            _Completer.completeError
I/flutter ( 9116):   dart:async-patch/async_patch.dart 40:18                                                      _AsyncAwaitCompleter.completeError
I/flutter ( 9116):   package:flutter/src/services/platform_channel.dart                                           MethodChannel.invokeMethod
I/flutter ( 9116):   ===== asynchronous gap ===========================
I/flutter ( 9116):   dart:async/zone.dart 1053:19                                                                 _CustomZone.registerUnaryCallback
I/flutter ( 9116):   dart:async-patch/async_patch.dart 77:23                                                      _asyncThenWrapperHelper
I/flutter ( 9116):   package:flutter/src/services/platform_chann
I/flutter ( 9116): 00:04 +3 -1: (tearDownAll)
I/flutter ( 9116): 00:04 +4 -1: Some tests failed.
```

After clicking on the index creation link, the pagination test pass on Flutter stable and master:

```
I/flutter (15322): Observatory listening on http://127.0.0.1:40544/SIE-3W4-fgE=/
[info ] FlutterDriver: Connecting to Flutter application at http://127.0.0.1:50425/SIE-3W4-fgE=/
[trace] FlutterDriver: Isolate found with number: 16386418
[trace] FlutterDriver: Isolate is paused at start.
[trace] FlutterDriver: Attempting to resume isolate
[trace] FlutterDriver: Waiting for service extension
I/flutter (15322): 00:00 +0: Firestore getDocuments
[info ] FlutterDriver: Connected to Flutter application.
I/flutter (15322): 00:01 +1: Firestore increment
I/flutter (15322): 00:03 +2: Firestore runTransaction
I/flutter (15322): 00:04 +3: Firestore pagination
I/flutter (15322): 00:07 +4: (tearDownAll)
I/flutter (15322): 00:07 +5: All tests passed!
Stopping application instance.
```

So, I'm still not reproducing the issue you're seeing. Can you confirm if these updated Android pagination integration tests pass or fail for you (on Flutter stable)?

```
cd packages/cloud_firestore/example
flutter drive test_driver/cloud_firestore.dart
```

creativecreatorormaybenot:
@collinjackson The ascending query still fails in my project and it is not possible to create the index that the error message links to (single-field). I was not able to run the integration tests successfully (crashed before running).

RollyPeres:
@collinjackson Also when using `.startAtDocument(someDoc).endAtDocument(someOtherDoc)` it errors. Isn't that something that it should be supported ?

rockwotj:
Firebaser here:

Please note that ordering DESCENDING is not correct here.

You can't actually use an `orderBy` at all with this `{start,end}{After,At}Document` feature.

As soon as you do something like:

`order by foo`

then do `startAfter(<doc>)` you're starting after `<doc>` in the field `foo` not `__name__`. I believe these methods need to be updated to make sure your orderBy clause is empty. /cc @collinjackson 

Additionally, `order by foo DESC` is normalized to `order by foo DESC, __name__ DESC` which is what our builtin indexes provide. There is no builtin index for <field> DESC, __name__ ASC, which is why you get an error to build an index.

See the ORDER BY normalization here:

https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/query.proto#L186-L204

>  this appears to be a recent change to Cloud Firestore that fundamentally changed how indexes work. Ordering by documentID ascending does not work (anymore), check flutter/flutter#32031. Now, there is also no index required when ordering by documentID descending.

Indexes did not change. And I'm not sure what you mean by ordering by documentID ascending does not work? Can you give an explicit example/snippet where that is the case?

creativecreatorormaybenot:
@rockwotj Does this mean that the order by the document id should be discarded as a whole?  
As the Flutter implementation does not use either the Android nor iOS implementations of `{start,end}{After,At}Document`, this would partially make this useless because e.g. `startAfterDocument` would only start after the other fields of the query and not **exactly** after that specific document.

For example: `order by foo, bar` and using `startAfterDocument(exampleDocument)`, where `exampleDocument` has `foo = 3` and `bar = 4`, it would start after `foo = 3` and `bar = 4` but not at *that specific document*.  
At least as far as I am concerned, this is not the way the native implementations of the `{start,end}{After,At}Document` feature work as it would be pretty flawed for most pagination cases like this.

rockwotj:
Sorry, I misread the code. I missed it was populating fields based on the whole document, not just the ID.

The only bug here as far as I can tell is the explicit order by. It should match the Firestore rules and then things should mostly work for single order bys.

collinjackson:
I've updated the issue so that the tests should pass.

@rockwotj I'm not sure what "bug" you're referring to -- is this something that can be fixed or at least documented to be less confusing? Is this PR ok to merge?

Thanks.

collinjackson:
@rockwotj am I understanding you correctly that you think we should handle the case where there's a single orderBy differently?

I am happy to merge this once we get the integration test to pass, but right now it's asking me to create an index that isn't possible to create (because of the single orderBy)

creativecreatorormaybenot:
@collinjackson That is actually interesting. In my project, it only asks to create an index when using ascending orderBy document id.

Do I understand the tests correctly that they only fail on iOS? I only tested my project on Android and it works with descending orderBy document id, which is the fix I originally added to the PR.

`build-apks+java-test+drive-examples` seems to pass, so maybe there is just an issue on the iOS side with this PR currently.

collinjackson:
We don't have Android integration tests on our CI currently.

This is what I see in the console UI:
Screen Shot 2019-06-06 at 15 39 28

creativecreatorormaybenot:
@collinjackson This is exactly what I see when I do not use my fix. Are you sure you ran it with the *descending* version and did you try on Android?

creativecreatorormaybenot:
@collinjackson https://github.com/flutter/flutter/issues/32031#issuecomment-500361091

rockwotj:
Sorry for the delay.

@creativecreatorormaybenot I'm saying this CL only moves the issue. Now `ASCENDING` queries will require an index.

You need to correctly normalize the direction of the document ID field. Here is an example of doing this in java: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java#L313-L321

Basically look at the last order by clause the user specified and use that as the sort direction (or default to ASCENDING). I can open a PR if needed.

creativecreatorormaybenot:
@rockwotj Thanks a lot. I finally understand exactly what you mean.  

This also explains why @collinjackson got the error with the PR and I got it without the PR (the last order field was descending in my case). I will create a new PR with what you said in mind and also fixing https://github.com/flutter/flutter/issues/34135.

collinjackson:
Thanks so much @creativecreatorormaybenot. I'm happy to review/land your new PR.

PR #1593, opened 2019-05-11
remove iOS dependency on Firebase/Database and Firebase/Auth

long1eu:


collinjackson:
lgtm. Thanks for the contribution.

PR #1594, opened 2019-05-11
[google_maps_flutter] Fix LatLng returning slight different values

jld3103:
## Description
Previously
```dart
LatLng location = LatLng(50, 6.123456789);
print(location.latitude.toString() + ', ' + location.longitude.toString());
```
would produce:
```dart
50, 6.123456788999988
```
Now it produces:
```dart
50, 6.123456789
```

### DIFF
--- a/packages/google_maps_flutter/lib/src/location.dart
+++ b/packages/google_maps_flutter/lib/src/location.dart
@@ -18,7 +18,7 @@ class LatLng {
         assert(longitude != null),
         latitude =
             (latitude < -90.0 ? -90.0 : (90.0 < latitude ? 90.0 : latitude)),
-        longitude = (longitude + 180.0) % 360.0 - 180.0;
+        longitude = longitude > 180 ? (longitude + 180.0) % 360.0 - 180.0 : longitude;
 
   /// The latitude in degrees between -90.0 and 90.0, both inclusive.
   final double latitude;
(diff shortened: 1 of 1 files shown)


jld3103:
I signed it!

packages/google_maps_flutter/lib/src/location.dart
```diff
@@ -18,7 +18,8 @@ class LatLng {
         assert(longitude != null),
         latitude =
             (latitude < -90.0 ? -90.0 : (90.0 < latitude ? 90.0 : latitude)),
-        longitude = (longitude + 180.0) % 360.0 - 180.0;
+        longitude =
+            longitude > 180 ? (longitude + 180.0) % 360.0 - 180.0 : longitude;
```
matthewlloyd: I think this should be

`longitude = (longitude < -180.0 || longitude >= 180.0) ? (longitude + 180.0) % 360.0 - 180.0 : longitude;`

For two reasons:
- As it stands the new code will map +180 to +180, whereas the old code mapped it to 0. The convention for longitude is normally that it should be in the interval [-180, +180), and that interval doesn't include +180.
- The new code doesn't wrap values less than -180 at all. The old code did.

jld3103: You're right

iskakaushik:
this PR needs tests.

zubyf09:
@google-admin @joaoatgoogle @mravn-google 
Google Map Android provides KML support in maps.
https://developers.google.com/maps/documentation/android-sdk/utility/kml
How we can use KML support in flutter ?

iskakaushik:
@zubyf09 please create a new issue here https://github.com/flutter/flutter/issues/new

iskakaushik:
Since it's been almost a month and the committer has not addressed the comments. I'm closing the pull request. Please re-open a new one when you are ready. Thanks!

PR #1482, opened 2019-04-11
[camera] allow camera plugin to be installed when android api < 21

jaumard:
## Description

Currently camera plugin is working for API 21, but flutter is since 16, so camera should be able to be installed on API 16 even if it doesn't work. That will allow developers to target API 16 and just hide the camera features for API 21 or more.

## Related Issues
fix https://github.com/flutter/flutter/issues/30854

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #1598, opened 2019-05-14
[camera] Enable camera plugin to compile with earlier android apis with custom AndroidManifest settings

bparrishMines:
## Description

Enable camera plugin to compile with earlier android sdks with custom AndroidManifest settings. 
Catching `CameraAccessException`s in the `onMethodCall` prevented compilation on <21. This separates handling them into another method. 

## Related Issues

Original PR: https://github.com/flutter/plugins/pull/1482
Fixes https://github.com/flutter/flutter/issues/30854

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.5.1
+
+* Can now be compiled with earlier Android sdks when
+`<uses-sdk tools:overrideLibrary="io.flutter.plugins.camera"/>` has been added to the project
+`AndroidManifest.xml`.
+
 ## 0.5.0
 
 * **Breaking Change** This plugin no longer handles closing and opening the camera on Android
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## 0.5.1
 
-* Can now be compiled with earlier Android sdks when
+* Can now be compiled with earlier Android sdks below 21 when
 `<uses-sdk tools:overrideLibrary="io.flutter.plugins.camera"/>` has been added to the project
-`AndroidManifest.xml`.
+`AndroidManifest.xml`. For sdks below 21, the plugin won't be registered and calls to it will throw
+a `MissingPluginException.`
 
 ## 0.5.0
 
--- a/packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
+++ b/packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
@@ -200,6 +200,9 @@ public class CameraPlugin implements MethodCallHandler {
     }
   }
 
+  // We move catching CameraAccessException out of onMethodCall because it causes a crash
+  // on plugin registration for sdks incompatible with Camera2 (< 21). We want this plugin to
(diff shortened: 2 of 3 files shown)


jaumard:
I can confirm that it works, I just get
```
Android dependency 'androidx.exifinterface:exifinterface' has different version for the compile (1.0.0-rc01) and runtime (1.0.0) classpath. You should manually set the same version via DependencyResolution
``` 
not sure where it come from :/

packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
```diff
@@ -126,8 +128,8 @@ public void onMethodCall(MethodCall call, final Result result) {
             cameras.add(details);
           }
           result.success(cameras);
-        } catch (CameraAccessException e) {
-          result.error("cameraAccess", e.getMessage(), null);
+        } catch (Exception e) {
```
mklim: In general it's better not to catch and rethrow `Exception`s broadly like this, it'll make stack traces hard to track down for people debugging the problem. Could you rewrite `handleException` to just take `CameraAccessException` and then continue to just catch that still?

packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
```diff
@@ -198,6 +200,15 @@ public void onMethodCall(MethodCall call, final Result result) {
     }
   }
 
+  @SuppressWarnings("ConstantConditions")
```
mklim: Why is this being suppressed?

packages/camera/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.5.1
+
+* Can now be compiled with earlier Android sdks when
```
mklim: I think it's worth saying that this plugin still doesn't actually work on older SDKs and is just a non-op, just to be clear.

mklim:
> I can confirm that it works, I just get
> 
> ```
> Android dependency 'androidx.exifinterface:exifinterface' has different version for the compile (1.0.0-rc01) and runtime (1.0.0) classpath. You should manually set the same version via DependencyResolution
> ```
> 
> not sure where it come from :/

In general it sounds related to flutter/flutter#27254?

packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
bparrishMines: So the app crashes when registering the plugin and you attempt to catch `CameraAccessException` in `onMethodCall`. So far, nothing else has worked including the use of `@RequiresAPI` and `@TargetAPI` annotations. 

Weirdly enough, you can still catch `CameraAccessException`s in the inner `Camera` class. 

I haven't been able to find another solution.

packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
bparrishMines: The casting of `Exception` to `RuntimeException`. From my understanding of Java, `CameraAccessException` is the only checked exception that will come from these methods. Otherwise it should be a `RuntimeException`.

PR #441, opened 2018-03-27
Make SharedPreferences return failures

kurtisnelson:
On Android, SharedPreferences#apply() will silently fail if there is an
issue. Using #commit() ensures we can propogate errors up the future
chain.
Also updated the example to reflect the fact that preference writes are
not instant.
Deprecated commit() as neither platform actually implements it
correctly.

PR #1600, opened 2019-05-14
[shared_preferences] Asynchronous commit() callback

awazgyawali:
Replaced .commit() to .apply() to get asyncronous performance.

### DIFF
--- a/packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
+++ b/packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
@@ -180,7 +180,8 @@ public class SharedPreferencesPlugin implements MethodCallHandler {
           result.success(getAllPrefs());
           return;
         case "remove":
-          result.success(preferences.edit().remove(key).commit());
+          editor = preferences.edit().remove(key);
+          backgroundTask(editor, result);
           break;
         case "clear":
           Set<String> keySet = getAllPrefs().keySet();
--- a/packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
+++ b/packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
@@ -123,12 +123,12 @@ public class SharedPreferencesPlugin implements MethodCallHandler {
     try {
       switch (call.method) {
         case "setBool":
-          status = preferences.edit().putBoolean(key, (boolean) call.argument("value")).commit();
+          preferences.edit().putBoolean(key, (boolean) call.argument("value")).apply();
           break;
         case "setDouble":
           double doubleValue = ((Number) call.argument("value")).doubleValue();
           String doubleValueStr = Double.toString(doubleValue);
-          status = preferences.edit().putString(key, DOUBLE_PREFIX + doubleValueStr).commit();
+          preferences.edit().putString(key, DOUBLE_PREFIX + doubleValueStr).apply();
           break;
         case "setInt":
           Number number = call.argument("value");
@@ -139,7 +139,7 @@ public class SharedPreferencesPlugin implements MethodCallHandler {
           } else {
             editor.putLong(key, number.longValue());
           }
-          status = editor.commit();
+          editor.apply();
           break;
         case "setString":
           String value = (String) call.argument("value");
@@ -150,11 +150,11 @@ public class SharedPreferencesPlugin implements MethodCallHandler {
                 null);
             return;
           }
-          status = preferences.edit().putString(key, value).commit();
+          preferences.edit().putString(key, value).apply();
           break;
         case "setStringList":
           List<String> list = call.argument("value");
-          status = preferences.edit().putString(key, LIST_IDENTIFIER + encodeList(list)).commit();
+          preferences.edit().putString(key, LIST_IDENTIFIER + encodeList(list)).apply();
           break;
         case "commit":
           // We've been committing the whole time.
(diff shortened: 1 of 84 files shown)


collinjackson:
paging @mehmetf 

Looking at https://github.com/flutter/plugins/pull/441 it seems like the decision to `commit()` on every set is intentional because it allows Dart code to be notified about errors.

It's possible that we should have an optional param { bool commit: true } to the setter APIs to give developers a choice here.

mehmetf:
Yep. We don't want this. `apply()` has no guarantee that the data is actually saved.

If we want async writes to disk, it would be better to execute the commit()'s off UI thread and return the result off of a callback.

awazgyawali:
@mehmetf Please let me know should I use a boolean param for .apply() or should I create an AsyncTask for it without asking bool param. I would love to refactor this PR.

mehmetf:
Thanks @awazgyawali for pushing for the correct fix! Yes, we should call `commit` from AsyncTask or [equivalent](https://medium.com/@frank.tan/using-a-thread-pool-in-android-e3c88f59d07f) and return the result with result.success() on the UI thread.

awazgyawali:
Hello @mehmetf, I have changed the PR accordingly, please review and let me know if there is anything that I should change. Also the `hard_coded_version` test, I am not sure what does that mean and how to solve it.

packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
```diff
@@ -116,30 +116,44 @@ private String encodeList(List<String> list) throws IOException {
     return filteredPrefs;
   }
 
+  void backgroundTask(final Editor editor, final MethodChannel.Result result) {
+    new Thread(
+            new Runnable() {
+              @Override
+              public void run() {
+                boolean status = editor.commit();
+                result.success(status);
```
mehmetf: You can't do this. result objects need to be called from UiThread. You can use `AsyncTask` if you want a nice API to work with.

packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
```diff
@@ -116,30 +117,47 @@ private String encodeList(List<String> list) throws IOException {
     return filteredPrefs;
   }
 
+  void backgroundTask(final Editor editor, final MethodChannel.Result result) {
```
mehmetf: Change method name to `commitAsync`

awazgyawali:
Added AsyncTask now.

packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
mehmetf: `private`

packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
```diff
@@ -116,30 +117,47 @@ private String encodeList(List<String> list) throws IOException {
...
+    new AsyncTask<Void, Void, Boolean>() {
+      @Override
+      protected Boolean doInBackground(Void... voids) {
+        return editor.commit();
+      }
+
+      @Override
+      protected void onPostExecute(Boolean value) {
+        super.onPostExecute(value);
```
mehmetf: Is this marked as "must call super"?

packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
```diff
@@ -116,30 +117,47 @@ private String encodeList(List<String> list) throws IOException {
...
+        result.success(value);
+      }
+    }.execute();
+  }
+
   @Override
   public void onMethodCall(MethodCall call, MethodChannel.Result result) {
     String key = call.argument("key");
-    boolean status = false;
+    Editor editor;
```
mehmetf: No need to define this here. You are not using it in the scope.

packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
```diff
@@ -116,30 +117,47 @@ private String encodeList(List<String> list) throws IOException {
...
-          status = preferences.edit().putBoolean(key, (boolean) call.argument("value")).commit();
+          editor = preferences.edit().putBoolean(key, (boolean) call.argument("value"));
+          backgroundTask(editor, result);
           break;
         case "setDouble":
           double doubleValue = ((Number) call.argument("value")).doubleValue();
           String doubleValueStr = Double.toString(doubleValue);
-          status = preferences.edit().putString(key, DOUBLE_PREFIX + doubleValueStr).commit();
+          editor = preferences.edit().putString(key, DOUBLE_PREFIX + doubleValueStr);
+          backgroundTask(editor, result);
```
mehmetf: Once you change method name, you can do:

`commitAsync(preferences.edit().putString(...), result);`

packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java
awazgyawali: https://stackoverflow.com/a/20637495/3931968 This says no, so I removed it.

mehmetf:
@collinjackson You have the final word in review and merge of this PR.

awazgyawali:
As a reminder: Builds are green now.

awazgyawali:
It's been 4 days and there is no newer version for this plugin. Isn't the upload to pub.dev automated?

collinjackson:
Changes are now published. Thanks for your patience.

PR #1618, opened 2019-05-20
Allow changing the webview's platform specific implementation

amirh:
The goal of this PR is to allow separate packages to extend the plugin to work on more platforms.
This is done by abstracting all the platform specific logic behind the `WebViewPlatform` and `WebViewBuilder` interfaces(in `platform_interface.dart`), and adding a static `WebView.platformBuilder` field that can be set to change the the platform specific code.

For example, in order to extend the webview to support Fuchsia, a `fuchsia_webview_flutter` plugin can implement  `WebViewBuilder` and `WebViewPlatform`(reference implementations available in `webview_android.dart` and `webview_cupertino.dart`).

When running on Fuchsia the plugin(or the app developer if the plugin doesn't have the right hooks) can then set itself as the implementation for the webview by running:
```dart
WebView.platformBuilder = FuchsiaWebViewBuilder();
```

### DIFF
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -29,7 +29,8 @@ abstract class WebViewPlatform {
   int get id;
 }
 
-typedef WebViewCreatedCallback = void Function(WebViewPlatform webViewPlatform);
+typedef WebViewPlatformCreatedCallback = void Function(
+    WebViewPlatform webViewPlatform);
 
 /// Interface for a platform specific webview implementation.
 ///
@@ -40,7 +41,7 @@ abstract class WebViewBuilder {
   Widget build({
     BuildContext context,
     Map<String, dynamic> creationParams,
-    WebViewCreatedCallback onWebViewCreated,
+    WebViewPlatformCreatedCallback onWebViewPlatformCreated,
     Set<Factory<OneSequenceGestureRecognizer>> gestureRecognizers,
   });
 }
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -8,11 +8,19 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/widgets.dart';
 
-/// Interface for talking to the webview controller on the platform side.
+/// Interface for talking to the webview's platform implementation.
 ///
-/// An instance implementing this interface is passed to the `onWebViewCreated` callback that is
-/// passed to [WebViewPlatformInterface#onWebViewCreated].
+/// An instance implementing this interface is passed to the `onWebViewPlatformCreated` callback that is
+/// passed to [WebViewPlatformBuilder#onWebViewPlatformCreated].
 abstract class WebViewPlatform {
+  /// Loads the specified URL.
+  ///
+  /// If `headers` is not null and the URL is an HTTP URL, the key value paris in `headers` will
+  /// be added as key value pairs of HTTP headers for the request.
+  ///
+  /// `url` must not be null.
+  ///
+  /// Throws an ArgumentError if `url` is not a valid URL string.
   Future<void> loadUrl(
     String url,
     Map<String, String> headers,
@@ -32,14 +40,33 @@ abstract class WebViewPlatform {
 typedef WebViewPlatformCreatedCallback = void Function(
     WebViewPlatform webViewPlatform);
 
-/// Interface for a platform specific webview implementation.
+/// Interface building a platform WebView implementation.
 ///
-/// [WebView#iplatformBuilder] controls the platform interface that is used by [WebView].
-/// [WebViewAndroidImplementation] and [WebViewIosImplementation] are the default implementations
+/// [WebView#platformBuilder] controls the builder that is used by [WebView].
+/// [AndroidWebViewPlatform] and [CupertinoWebViewPlatform] are the default implementations
 /// for Android and iOS respectively.
 abstract class WebViewBuilder {
+  /// Builds a new WebView.
+  ///
+  /// Returns a Widget tree that embeds the created webview.
+  ///
+  /// `creationParams` are the initial parameters used to setup the webview.
+  ///
+  /// `onWebViewPlatformCreated` will be invoked after the platform specific [WebViewPlatform]
+  /// implementation is created with the [WebViewPlatform] instance as a parameter.
+  ///
+  /// `gestureRecognizers` specifies which gestures should be consumed by the web view.
+  /// It is possible for other gesture recognizers to be competing with the web view on pointer
+  /// events, e.g if the web view is inside a [ListView] the [ListView] will want to handle
+  /// vertical drags. The web view will claim gestures that are recognized by any of the
+  /// recognizers on this list.
+  /// When `gestureRecognizers` is empty or null, the web view will only handle pointer events for gestures that
+  /// were not claimed by any other gesture recognizer.
   Widget build({
     BuildContext context,
+    // TODO(amirh): convert this to be the actual parameters.
+    // I'm starting without it as the PR is starting to become pretty big.
+    // I'll followup with the conversion PR.
     Map<String, dynamic> creationParams,
     WebViewPlatformCreatedCallback onWebViewPlatformCreated,
     Set<Factory<OneSequenceGestureRecognizer>> gestureRecognizers,
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -1,0 +1,44 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/widgets.dart';
+
+/// Interface for talking to the webview controller on the platform side.
+/// 
+/// An instance implementing this interface is passed to the `onWebViewCreated` callback that is
+/// passed to [WebViewPlatformInterface#onWebViewCreated].
+abstract class WebViewPlatformControllerInterface {
+  Future<void> loadUrl(
+      String url,
+      Map<String, String> headers,
+      );
+
+  // As the PR currently focus about the wiring I've only moved loadUrl to the new way, so
+  // the discussion is more focused.
+  // In this temporary state WebViewController still uses a method channel directly for all other
+  // method calls so we need to expose the webview ID.
+  // TODO(amirh): remove this before submitting this PR(after getting an LGTM for the overall approach).
+  int get id;
+}
+
+typedef WebViewCreatedCallback = void Function(WebViewPlatformControllerInterface);
+
+/// Interface for a platform specific webview implementation.
+///
+/// [WebView#implementation] controls the platform interface that is used by [WebView].
+/// [WebViewAndroidImplementation] and [WebViewIosImplementation] are the default implementations
+/// for Android and iOS respectively.
+abstract class WebViewPlatformInterface {
+  Widget build({
+    BuildContext context,
+    Map<String, dynamic> creationParams,
+    WebViewCreatedCallback onWebViewCreated,
+    Set<Factory<OneSequenceGestureRecognizer>> gestureRecognizers,
+  });
+}
+
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -25,7 +25,7 @@ abstract class WebViewPlatform {
   // the discussion is more focused.
   // In this temporary state WebViewController still uses a method channel directly for all other
   // method calls so we need to expose the webview ID.
-  // TODO(amirh): remove this before submitting this PR(after getting an LGTM for the overall approach).
+  // TODO(amirh): remove this before publishing this package.
   int get id;
 }
 
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -9,14 +9,17 @@ import 'package:flutter/gestures.dart';
 import 'package:flutter/widgets.dart';
 
 /// Interface for talking to the webview controller on the platform side.
-/// 
+///
 /// An instance implementing this interface is passed to the `onWebViewCreated` callback that is
 /// passed to [WebViewPlatformInterface#onWebViewCreated].
-abstract class WebViewPlatformControllerInterface {
+abstract class WebViewPlatform {
   Future<void> loadUrl(
-      String url,
-      Map<String, String> headers,
-      );
+    String url,
+    Map<String, String> headers,
+  ) {
+    throw UnimplementedError(
+        "WebView loadUrl is not implemented on the current platform");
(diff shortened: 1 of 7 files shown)


packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -0,0 +1,44 @@
...
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/widgets.dart';
+
+/// Interface for talking to the webview controller on the platform side.
+/// 
+/// An instance implementing this interface is passed to the `onWebViewCreated` callback that is
+/// passed to [WebViewPlatformInterface#onWebViewCreated].
+abstract class WebViewPlatformControllerInterface {
```
amirh: Actually I think we better make this class non-abstract and have a default implementation(that throws?) for all methods, this way a new implementor doesn't have to implement the full API surface all at once (and when we're adding new methods here it's going to take time for implementors to catch-up).

I'll update the PR to make this non abstract tomorrow.

nkoroste: nit: maybe "WebViewControllerPlatformInterface

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -0,0 +1,44 @@
...
+}
+
+typedef WebViewCreatedCallback = void Function(WebViewPlatformControllerInterface);
+
+/// Interface for a platform specific webview implementation.
+///
+/// [WebView#implementation] controls the platform interface that is used by [WebView].
+/// [WebViewAndroidImplementation] and [WebViewIosImplementation] are the default implementations
+/// for Android and iOS respectively.
+abstract class WebViewPlatformInterface {
```
nkoroste: nit: maybe move each interface into it's own file to emphasize their isolation

Hixie:
> Note: This PR is not ready to be merged

FWIW, GitHub has a "create PR as WIP" mode when you create the PR that will make the PR appear grey instead of green until you say it's ready.

packages/webview_flutter/lib/platform_interface.dart
nkoroste: I  think it can stay abstract but you can just add default implementation which others will inherit when extending it

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -0,0 +1,44 @@
...
+
+/// Interface for talking to the webview controller on the platform side.
+/// 
+/// An instance implementing this interface is passed to the `onWebViewCreated` callback that is
+/// passed to [WebViewPlatformInterface#onWebViewCreated].
+abstract class WebViewPlatformControllerInterface {
+  Future<void> loadUrl(
+      String url,
+      Map<String, String> headers,
+      );
```
Hixie: trivial nit: indenting

packages/webview_flutter/lib/src/webview_ios.dart
```diff
@@ -0,0 +1,26 @@
...
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+
+import '../platform_interface.dart';
+import 'webview_method_channel.dart';
+
+class WebViewIosImplementation implements WebViewPlatformInterface {
```
Hixie: Avoid using `Ios` in classes, the correct capitalization is `iOS`. Use `Cupertino` if necessary.

packages/webview_flutter/lib/src/webview_ios.dart
```diff
@@ -0,0 +1,26 @@
...
+  @override
+  Widget build({BuildContext context, Map<String, dynamic> creationParams, WebViewCreatedCallback onWebViewCreated, Set<Factory<OneSequenceGestureRecognizer>> gestureRecognizers}) {
+    return UiKitView(
```
Hixie: assert onWebViewCreated isn't null

packages/webview_flutter/lib/src/webview_method_channel.dart
```diff
@@ -0,0 +1,33 @@
+import 'dart:async';
```
Hixie: missing copyright statement

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -121,6 +125,28 @@ class WebView extends StatefulWidget {
...
+  static set implementation(WebViewPlatformInterface implementation) {
+    _implementation = implementation;
+  }
+
+  static WebViewPlatformInterface get implementation {
+    if (_implementation != null) {
+      return implementation;
+    }
+
+    if (defaultTargetPlatform == TargetPlatform.android) {
```
Hixie: prefer using a switch when comparing on an enum. see style guide.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -121,6 +125,28 @@ class WebView extends StatefulWidget {
...
+      return WebViewAndroidImplementation();
+    }
+
+    if (defaultTargetPlatform == TargetPlatform.iOS) {
+      return WebViewIosImplementation();
+    }
+
+    throw UnsupportedError("Trying to use the default webview implementation for $defaultTargetPlatform but there isn't a default one");
+  }
```
Hixie: did you intend to cache the returned value?

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -0,0 +1,44 @@
...
+
+  // As the PR currently focus about the wiring I've only moved loadUrl to the new way, so
+  // the discussion is more focused.
+  // In this temporary state WebViewController still uses a method channel directly for all other
+  // method calls so we need to expose the webview ID.
+  // TODO(amirh): remove this before submitting this PR(after getting an LGTM for the overall approach).
+  int get id;
+}
+
+typedef WebViewCreatedCallback = void Function(WebViewPlatformControllerInterface);
```
Hixie: nit: we like to name the arguments in typedefs. In some places, without a name, the type becomes the name (e.g. `void bar(Foo)` means `void bar(dynamic Foo)`) so it's better to be consistent throughout.

packages/webview_flutter/lib/platform_interface.dart
Hixie: how about WebViewPlatform and FuchsiaWebViewPlatform/AndroidWebViewPlatform/etc?

Hixie: what @nkoroste says

hterkelsen:
LGTM for the overall approach.

I like how you are abstracting the "communicate with the platform side" part of the implementation with `WebViewPlatformControllerInterface`, but if something similar will be required for all plugins with implementations that wish to avoid platform channel overhead, perhaps this idea should be pushed into the framework itself.

I think the problem is that Flutter assumes that communication with the platform-side _must_ be done by encoding the data into bytes, but with the new platform implementations (Fuchsia, web) we know that is not necessarily the case.

packages/webview_flutter/lib/platform_interface.dart
amirh: Ok I think II'm going with:
`WebViewPlatform`/`AndroidWebViewPlatform` and then `WebViewBuilder`/`AndroidWebViewBuilder`(instead of `WebViewPlatformInterface`)

This will mean to set it up for Fuchsia we will do:
`WebView.platformBuilder = FuchsiaWebViewBuilder();`

amirh: ok, added a default implementation that throws.

packages/webview_flutter/lib/platform_interface.dart
amirh: done

packages/webview_flutter/lib/platform_interface.dart
amirh: done

packages/webview_flutter/lib/src/webview_ios.dart
amirh: done

packages/webview_flutter/lib/src/webview_ios.dart
amirh: done

packages/webview_flutter/lib/src/webview_method_channel.dart
amirh: done

packages/webview_flutter/lib/webview_flutter.dart
amirh: done

packages/webview_flutter/lib/webview_flutter.dart
amirh: done

amirh:
Looks like we're generally ok with this approach.

I'm going to start landing it, to make review easier I'd rather land this in pieces, this first piece sets up the plumbing, and moves a single method `loadUrl` to the Dart interface.

I plan to followup with 2 PRs:
1. Move the rest of the method channel methods behind the Dart interface.
2. Replace the `creationParams` map with Dart objects.

As each of these PRs will be a "breaking change" from the current PR, I'm not going to publish the plugin until everything landed.

amirh:
I think this is ready for review now @cyanglaz @nkoroste

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -385,6 +395,7 @@ class _WebSettings {
 class WebViewController {
```
nkoroste: is the long term plan to get rid of this class in favor of WebViewPlatform?

packages/webview_flutter/lib/platform_interface.dart
Hixie: Do WebViewBuilder and WebViewPlatform need to be distinct?

Hixie: If the default implementation throws, it's probably best if the default is abstract (so that the analyzer tells you to implement it). Otherwise you risk crashes at runtime and there's no lint to tell you what you need to implement. If it makes sense to have an actual default that doesn't throw, I'd go with that. That way you can get up and going quickly on a new platform.

packages/webview_flutter/lib/webview_flutter.dart
amirh: This is the public API we expose to our users(vs the public `WebViewPlatform` which is the API we expose to platform implementors), while the public API for these 2 currently matches I'm not sure we should require it to match.

Also note that renaming WebViewController right now will be a breaking change, I guess we could eventually make WebViewController extend WebViewPlatform if we want to save one indirection layer, and we extract all the utility code(like the logic in `_updateJavaScriptChannels`) out of it.

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -0,0 +1,74 @@
...
+/// passed to [WebViewPlatformBuilder#onWebViewPlatformCreated].
+abstract class WebViewPlatform {
+  /// Loads the specified URL.
+  ///
+  /// If `headers` is not null and the URL is an HTTP URL, the key value paris in `headers` will
+  /// be added as key value pairs of HTTP headers for the request.
+  ///
+  /// `url` must not be null.
+  ///
+  /// Throws an ArgumentError if `url` is not a valid URL string.
```
cyanglaz: How do we enforce the specific platform implementation to honor the parameter constraints mentioned here?

packages/webview_flutter/lib/platform_interface.dart
amirh: We have a single `WebViewPlatform` per a platform webview instance, while a `WebViewBuilder` can create zero or more `WebViewPlatform` instances.

I don't see a nice way to merge them(other than adding a `webviewId` parameter to all the methods and then using a single instance of that merges `WebViewPlatform` to serve all of the webview instance).

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -0,0 +1,74 @@
...
+typedef WebViewPlatformCreatedCallback = void Function(
+    WebViewPlatform webViewPlatform);
+
+/// Interface building a platform WebView implementation.
+///
+/// [WebView#platformBuilder] controls the builder that is used by [WebView].
+/// [AndroidWebViewPlatform] and [CupertinoWebViewPlatform] are the default implementations
+/// for Android and iOS respectively.
+abstract class WebViewBuilder {
+  /// Builds a new WebView.
```
cyanglaz: Nit: Should have an extra space above the dart doc?

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -0,0 +1,74 @@
...
+
+typedef WebViewPlatformCreatedCallback = void Function(
+    WebViewPlatform webViewPlatform);
+
+/// Interface building a platform WebView implementation.
+///
+/// [WebView#platformBuilder] controls the builder that is used by [WebView].
+/// [AndroidWebViewPlatform] and [CupertinoWebViewPlatform] are the default implementations
+/// for Android and iOS respectively.
+abstract class WebViewBuilder {
```
nkoroste: nit: still think you should move this to it's own file

packages/webview_flutter/lib/webview_flutter.dart
nkoroste: I agree that this should not be part of this PR, but in general I think this controller and WebViewPlatform going to look fairly similar so it would be nice if we can unify them

packages/webview_flutter/lib/platform_interface.dart
amirh: Not sure if we can enforce it. The closet thing we have is the e2e test which we should hope implementors of new platforms will run.

packages/webview_flutter/lib/platform_interface.dart
amirh: Should we? I wasn't aware that we enforce something like that, I can't find anything about it in the style guide, are we enforcing this convention?

packages/webview_flutter/lib/platform_interface.dart
amirh: I'd prefer to extract it if this file ever becomes too big, I kind of like the idea of having a single file that we can point implementors of new platforms to where everything they need to implement is specified...

If you feel strongly about it I'll split it now.

packages/webview_flutter/lib/webview_flutter.dart
amirh: Let's re-discuss it after we move all the methodchannel stuff out of WebViewController?

I'm actually not sure which option I prefer, I can imagine having a value in separating the interface that users are using to control the webview from the interface that platforms need to provide for the plugin to control the webview (e.g I can see us adding some logic in WebViewController that don't belong in the platform implementation).
But I do see your point about the code duplication.

We should decide at the end of this PR series before publishing to avoid a breaking change.

packages/webview_flutter/lib/platform_interface.dart
cyanglaz: I asked because it seems other places in this file is doing that. If it is not enforced then I think we are ok. However, I do think it is better to keep the style consistent throughout a file or even a package.

amirh: Not sure I understand, how can weI enforce that `"onWebViewPlatformCreated will be invoked after the platform specific [WebViewPlatform] implementation is created with the [WebViewPlatform] instance as a parameter."`

PR #1621, opened 2019-05-21
Change LocalAuth to use Biometric API

mehmetf:
## Description

Migrates the LocalAuth plugin on Android to Biometric API for showing the fingerprint dialog. This will allow us to implement more biometric authentication schemes (and others such as PIN) in the future.

## Related Issues
https://github.com/flutter/flutter/issues/18690

## Breaking Change

This is a breaking change because it requires Flutter applications on Android to extend a FragmentActivity rather than a regular Activity. The fix is to use `FlutterFragmentActivity` instead of `FlutterActivity`. The y version has been bumped due to this.

### DIFF
--- a/packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
+++ b/packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
@@ -125,11 +125,13 @@ class AuthenticationHelper extends BiometricPrompt.AuthenticationCallback
 
   @Override
   public void onAuthenticationError(int errorCode, CharSequence errString) {
-    if (!activityPaused || !isAuthSticky) {
-      // Either the authentication got cancelled by user or we are not interested
-      // in sticky auth, so return failure.
-      stop(false);
+    if (activityPaused && isAuthSticky) {
+      return;
     }
+
+    // Either the authentication got cancelled by user or we are not interested
+    // in sticky auth, so return failure.
+    stop(false);
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


mehmetf:
@jiajiabingcheng Please take a look at this as well. It would be good if you copied this to Leafy's codebase and gave it a try.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
```diff
@@ -147,99 +123,53 @@ private void stop(boolean success) {
...
-   * If the activity is paused or stopped, we have to stop listening for fingerprint. Otherwise,
-   * user can still interact with fp reader in the background.. Sigh..
-   */
   @Override
-  public void onActivityPaused(Activity activity) {
-    if (call.argument("stickyAuth")) {
-      pause();
-    } else {
+  public void onAuthenticationError(int errorCode, CharSequence errString) {
+    if (!activityPaused || !isAuthSticky) {
```
mklim: optional, subjective: The negatives combined with the logical or makes this kind of hard to read. Can we rephrase this?

```java
if (activityPaused && isAuthSticky) {
  return;
}

// Either the authentication got cancelled by user or we are not interested
// in sticky auth, so return failure.
stop(false);
```

mehmetf: Good suggestion. Applied.

awazgyawali:
It seems that we have to change the Activity manually, shouldn't this be mentioned on the readme?

PR #1624, opened 2019-05-22
[webview_flutter] platform_interface: Use Dart objects for creationParams and webSettings.

amirh:
As a temporary step we passed these objects as maps that were sent across the method channel. This PR replaces the maps with proper Dart objects.

### DIFF
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -59,9 +59,10 @@ class WebSettings {
     this.debuggingEnabled,
   });
 
+  /// The JavaScript execution mode to be used by the webview.
   final JavascriptMode javascriptMode;
 
-  /// Whether a [NavigationDelegate] should be used for this webview.
+  /// Whether the [WebView] has a [NavigationDelegate] set.
   final bool hasNavigationDelegate;
 
   /// Whether to enable the platform's webview content debugging tools.
@@ -92,7 +93,7 @@ class CreationParams {
 
   /// The initial set of JavaScript channels that are configured for this webview.
   ///
-  /// For each value in this list the platform's webview should make sure that a corresponding
+  /// For each value in this set the platform's webview should make sure that a corresponding
   /// property with a postMessage method is set on `window`. For example for a JavaScript channel
   /// named `Foo` it should be possible for JavaScript code executing in the webview to do
   ///
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -8,6 +8,8 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/widgets.dart';
 
+import 'webview_flutter.dart';
+
 /// Interface for talking to the webview's platform implementation.
 ///
 /// An instance implementing this interface is passed to the `onWebViewPlatformCreated` callback that is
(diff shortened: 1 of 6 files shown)


amirh:
@nkoroste

packages/webview_flutter/lib/src/webview_method_channel.dart
```diff
@@ -32,6 +32,42 @@ class MethodChannelWebViewPlatform implements WebViewPlatform {
     });
   }
 
+  @override
+  Future<void> updateSettings(WebSettings settings) {
+    // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
+    // https://github.com/flutter/flutter/issues/26431
+    // ignore: strong_mode_implicit_dynamic_method
+    final Map<String, dynamic>  updatesMap = _webSettingsToMap(settings);
+    if (updatesMap.isEmpty) {
```
cyanglaz: Another option could be `assert(!updatesMap.isEmpty)`

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -307,85 +307,51 @@ class _WebViewState extends State<WebView> {
...
-    return _WebSettings(
-      javascriptMode: widget.javascriptMode,
-      hasNavigationDelegate: widget.navigationDelegate != null,
-      debuggingEnabled: widget.debuggingEnabled,
-    );
+WebSettings _webSettingsUpdate(WebSettings currentValue, WebSettings newValue) {
+  JavascriptMode javascriptMode;
+  bool hasNavigationDelegate;
+  bool debuggingEnabled;
+  if (currentValue.javascriptMode != newValue.javascriptMode) {
```
cyanglaz: if the newValue.javascriptMode is null and the currentValue.javascriptMode is non-null, we are setting the old value to null.  Same as below. It seems to violate the comment here:
```  
/// Updates the webview settings.
///
/// Any non null field in `settings` will be set as the new setting value.
/// All null fields in `settings` are ignored.
```

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -37,6 +48,67 @@ abstract class WebViewPlatform {
...
+
+  final JavascriptMode javascriptMode;
+
+  /// Whether a [NavigationDelegate] should be used for this webview.
+  final bool hasNavigationDelegate;
+
+  /// Whether to enable the platform's webview content debugging tools.
+  ///
+  /// See also: [WebView.debuggingEnabled].
+  final bool debuggingEnabled;
```
nkoroste: nit: isDebuggingEnabled

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -37,6 +48,67 @@ abstract class WebViewPlatform {
...
+  final String initialUrl;
+
+  /// The initial [WebSettings] for the new webview.
+  ///
+  /// This can later be updated with [WebViewPlatform.updateSettings].
+  final WebSettings webSettings;
+
+  /// The initial set of JavaScript channels that are configured for this webview.
+  ///
+  /// For each value in this list the platform's webview should make sure that a corresponding
```
nkoroste: nit: "in this **set**"

packages/webview_flutter/lib/src/webview_method_channel.dart
```diff
@@ -32,6 +32,42 @@ class MethodChannelWebViewPlatform implements WebViewPlatform {
...
+    }
+
+    _addIfNonNull('jsMode', settings.javascriptMode?.index);
+    _addIfNonNull('hasNavigationDelegate', settings.hasNavigationDelegate);
+    _addIfNonNull('debuggingEnabled', settings.debuggingEnabled);
+    return map;
+  }
+
+
+  static Map<String, dynamic> creationParamsToMap(CreationParams creationParams) {
```
nkoroste: add dart doc or should it be private?

packages/webview_flutter/lib/webview_flutter.dart
nkoroste: +1

amirh: The fields in `currentValue` should never be null (the initial value built by _WebSettingFromWidget can never have null values).

I made it clear in the method comment and added assertions.

packages/webview_flutter/lib/src/webview_method_channel.dart
amirh: It's actually legit to send a `settings` with all null fields to this method.

We could potentially move the short-circuting logic to webview_flutter.dart, but that requires some ugliness or maintaining an `operator==` for `WebSettings` which I'm hesitant to do from a forward-compatibility point of view(I can see us adding a list field to WebSettings, at which point naive looking `webSettings==webSettings2` might actually be expensive.

packages/webview_flutter/lib/src/webview_method_channel.dart
amirh: done

packages/webview_flutter/lib/platform_interface.dart
amirh: done

packages/webview_flutter/lib/platform_interface.dart
amirh: I think it makes sense to keep it consistent with the `WebView` parameter name, if we want to change both it will be a breaking change at this point.

Note that as far as I know the Flutter style guide doesn't prefer an "is" prefix for flags.

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -37,6 +48,67 @@ abstract class WebViewPlatform {
...
+/// Initial settings are passed as part of [CreationParams], settings updates are sent with
+/// [WebViewPlatform#updateSettings].
+class WebSettings {
+  WebSettings({
+    this.javascriptMode,
+    this.hasNavigationDelegate,
+    this.debuggingEnabled,
+  });
+
+  final JavascriptMode javascriptMode;
```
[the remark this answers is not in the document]
amirh: done

packages/webview_flutter/lib/webview_flutter.dart
cyanglaz: I was thinking if the fields in newValue is null and this will set the corresponding fields in the currentValue to null. But the dart doc in the `updateSetting` in the WebViewSettings class mentioned that the null fields should be ignored.

cyanglaz: After offline discussion with @amirh; we confirmed that the fields in both currentValue and newValue could never be null because they are both created by _webSettingsFromWidget which takes the setting from the `WebView` widget, which should never be null. Can possibly add assertions in the code but it will create a lot of code overhead.

amirh: Added more assertions here

amirh:
Thanks.
Waiting for @nkoroste's approval before merging.

packages/webview_flutter/lib/webview_flutter.dart
amirh: On the platform side, at least for the platforms we've been dealing with so far, all these "settings" are updated by imperative calls, e.g on Android we call [setJavaScriptEnabled](https://developer.android.com/reference/android/webkit/WebSettings.html#setJavaScriptEnabled(boolean)) to update the "javascriptMode".
We do not want to call [setJavaScriptEnabled](https://developer.android.com/reference/android/webkit/WebSettings.html#setJavaScriptEnabled(boolean)) every time the widget is rebuilt, so we're keeping track of the previous `javascriptMode` value we had the last time the widget was built, and only when we see that the value has changed we call `setJavaScriptEnabled`.

The `WebSettings` class groups these kind of settings. We track the previous `WebSettings` value, and  when the widget is rebuilt this method(`_webSettingsUpdate`) computes a `WebSettings` instance that only includes the values that have changed and has to be updated on the platform side. This value is then passed to `WebViewPlatform.updateSettings` which ignores the null fields and updates the platform webview with the new values for all non-null fields in WebSettings.

nkoroste: I see, so you're nullfying all the unchanged fields on purpose because `WebViewPlatform.updateSettings` expects it that way. I guess the method name confused me, it's more like `nullfyUnchangedValues` which is a horrible name I agree. Anyway, thanks for the detailed explanation.

amirh: Renamed to `_clearUnchangedWebSettings`

PR #1641, opened 2019-05-28
[android_alarm_manager] Move method channel usage to the platform main thread

jason-simmons:
Fixes flutter/flutter#32976

### DIFF
--- a/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
+++ b/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
@@ -9,6 +9,7 @@ import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.os.Handler;
 import android.util.Log;
 import androidx.core.app.JobIntentService;
 import io.flutter.plugin.common.MethodChannel;
@@ -421,7 +422,7 @@ public class AlarmService extends JobIntentService {
    * callbacks have been executed.
    */
   @Override
-  protected void onHandleWork(Intent intent) {
+  protected void onHandleWork(final Intent intent) {
     // If we're in the middle of processing queued alarms, add the incoming
     // intent to the queue and return.
     synchronized (sAlarmQueue) {
@@ -434,6 +435,13 @@ public class AlarmService extends JobIntentService {
 
     // There were no pre-existing callback requests. Execute the callback
     // specified by the incoming intent.
-    executeDartCallbackInBackgroundIsolate(intent);
+    new Handler(getMainLooper())
+        .post(
+            new Runnable() {
+              @Override
+              public void run() {
+                executeDartCallbackInBackgroundIsolate(intent);
+              }
+            });
   }
 }


audkar:
Are you sure this is valid usage of [onhandlework](https://developer.android.com/reference/android/support/v4/app/JobIntentService#onhandlework) method? It is run on background thread for a reason. After that method is synchronously executed wake lock is released and Service is stopped so further background work is mostly impossible.

jason-simmons:
@bkonyi @matthew-carroll any suggestions here?

Would there be any benefit to having `AlarmService.onHandleWork` block until the platform->Dart message has been sent on the main thread?

The execution of the background work happens asynchronously on a Dart thread after `onHandleWork` returns, and we don't have a way for `executeDartCallbackInBackgroundIsolate` to wait until the Dart work completes.  So it looks like blocking until `sBackgroundChannel.invokeMethod` completes would not provide any additional guarantees.  `sBackgroundChannel` is a static that is set during plugin registration and apparently is never cleared, so it should be safe to use after shutdown of the `AlarmService`.

audkar:
> sBackgroundChannel is a static that is set during plugin registration and apparently is never cleared, so it should be safe to use after shutdown of the AlarmService.

Theoretically that would be true. But if application doesn't have active Activities/Services/Broadcast receivers whole application is going to be stopped in very short period of time.

Also latest Android version have very strict rules for background work to conserve battery. So even if application is not killed, some services like network/gps chip can be suspended if no wakelock is registered. And letting `onHandleWork` method pass, wakeLock is released.

> You do not need to use WakefulBroadcastReceiver when using this class. When running on Android O, the JobScheduler will take care of wake locks for you (holding a wake lock from the time you enqueue work until the job has been dispatched and while it is running).

I am not sure if appropriate to have `alarm_manager` plugin which is based on hope that background work will be done in time. I know that flutter communicates with platforms in async way. But that is not right give users such flaky way to execute scheduled tasks. Maybe then need to use some thread `Lock` and manage that in async way.

matthew-carroll:
@jason-simmons I think blocking in `onHandleWork()` until the Dart work completes is probably a good idea. That would theoretically allow us to execute that method as it was intended and thereby not trigger any unexpected shutdowns from the OS. I don't have any specific knowledge to back that up, but blockings seems safer than the alternative.

bkonyi:
I'm not an expert when it comes to Android lifecycles, but I'm under the assumption that the JobIntentService grabs a wake lock while a job is being processed in `onHandleWork`, releasing it when `onHandleWork` completes. I'd be hesitant to push any work that's currently done in that method into another queue or thread as that may prevent the alarm from being handled properly while backgrounded. If that's what we need to do, I'd probably also recommend blocking in `onHandleWork` to make sure the application doesn't shut down while we wait for the alarm to be handled elsewhere.

jason-simmons:
`executeDartCallbackInBackgroundIsolate` just sends a fire-and-forget message to Dart.  Prior to this PR the message was sent on the thread that called `onHandleWork`.  With this PR the message is sent on the main thread.  But in both cases `executeDartCallbackInBackgroundIsolate` did not wait for a reply from the Dart side.

It sounds like it would be safer to:
  * have `executeDartCallbackInBackgroundIsolate` signal an event when it receives a `MethodChannel.Result` callback
  * have `onHandleWork` queue the `executeDartCallbackInBackgroundIsolate` call to the main thread and then block until the event is signaled

Any concerns with this approach?

`executeDartCallbackInBackgroundIsolate` is also called from `AlarmService.onInitialized`, which runs on the main thread.  Those calls would need to be non-blocking.

matthew-carroll:
That proposal sounds good to me! Thanks for looking into it.

bkonyi:
+1, SGTM!

amirh:
I think we missed publishing this fix and updating the changelog, I believe it got published with `0.4.1+7` (https://github.com/flutter/plugins/pull/1849). @jason-simmons  / @bkonyi maybe we should retroactively update the changelog for that version?

PR #1644, opened 2019-05-28
[cloud_firestore] Add user agent submission to Firestore plugin

kroikie:
## Description

Add ability for Firestore plugin to submit Flutter user agent: `flutter-firebase_cloud_firestore`

### DIFF
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.12.0+1
+
+* Send user agent to Firebase.
+
 ## 0.12.0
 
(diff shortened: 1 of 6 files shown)


creativecreatorormaybenot:
@kroikie Why did you add hardcoded versions with this PR?
Is this necessary because it is somewhat of a pain to change all of the version codes for every revision.

https://github.com/flutter/plugins/blob/ef8b493bb18f245b049b1ca69599fe9270bef69e/packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/FlutterFirebaseAppRegistrar.java#L11

kroikie:
@creativecreatorormaybenot It is necessary for now so we can submit usage metrics for our Firebase plugins. We are looking into ways to automatically get the version number of the plugin so we can remove this manual step.

PR #1595, opened 2019-05-11
Updated Gradle tooling to match Android Studio 3.4

ened:
## Description

As a follow up PR to https://github.com/flutter/flutter/pull/32550, this updates ALL the plugins gradle & gradle plugin versions to the latest.

It's unfortunately quite a beast to review, therefore please let me know if this needs to be split in any way.

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ ] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #1653, opened 2019-05-29
[shared_preferences] Updated Gradle tooling to match Android Studio 3.4.

collinjackson:
This is a partial landing of #1595, updating the Gradle to match Android Studio 3.4.

If there are no issues, we'll go ahead and update all the other plugins.


collinjackson:
> Are the non gradle changes required?

didn't mean to include those. I've removed them, thanks

PR #1654, opened 2019-05-29
[path_provider] Cast NSInteger to long

mehmetf:
For arm32, arm64 compatibility


collinjackson:
LGTM. In the future, can you please push your changes to a public fork of flutter/plugins, instead of a branch on the flutter/plugins repo? We're trying to avoid having lots of branches on the main repo.

PR #1657, opened 2019-05-30
[firebase_core] replace invokeMethod with invokeMapMethod, add type parameters to invokeMethod calls,  bump min Flutter version to 1.5.0

cyanglaz:


collinjackson:
Note: For this plugin you'll have to updated the hard coded version to get the bots to be happy. I'm working on making this more automatic in the future. (flutter/flutter#32267)

PR #1667, opened 2019-05-31
[firebase_crashlytics]add missing type parameters and pump min flutter version to 1.5.0

cyanglaz:


packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -41,11 +41,8 @@ class Crashlytics {
           Trace.format(details.stack).trimRight().split('\n');
       final List<Map<String, String>> stackTraceElements =
           _getStackTraceElements(stackTraceLines);
-      final dynamic result =
-          // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-          // https://github.com/flutter/flutter/issues/26431
-          // ignore: strong_mode_implicit_dynamic_method
-          await channel.invokeMethod('Crashlytics#onError', <String, dynamic>{
+      final String result =
```
cyanglaz: @kroikie I saw that it returns string from the platform so I changed it to String. Please confirm it is correct.

packages/firebase_crashlytics/pubspec.yaml
```diff
@@ -1,12 +1,13 @@
...
 description: Flutter plugin for Firebase Crashlytics. It reports uncaught errors to the
   Firebase console.
-version: 0.0.4+3
+version: 0.0.4+4
 author: Flutter Team <flutter-dev@google.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/firebase_crashlytics
 
 environment:
   sdk: ">=2.0.0-dev.68.0 <3.0.0"
+  flutter: ">=1.5.0"
```
cyanglaz: Note that i set the min flutter version to 1.5.0 here.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
cyanglaz: It actually fails a test. I reverted to have dynamic. I'll let you to update it if need to.

PR #1669, opened 2019-05-31
[firebase_dynamic_links]Add missing template type parameter to  calls. Bump minimum Flutter v…

cyanglaz:

### DIFF
--- a/packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
+++ b/packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
@@ -23,11 +23,9 @@ class FirebaseDynamicLinks {
   /// there is no pending dynamic link or any call to this method after the
   /// the first attempt.
   Future<PendingDynamicLinkData> retrieveDynamicLink() async {
-    final Map<dynamic, dynamic> linkData =
-        // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-        // https://github.com/flutter/flutter/issues/26431
-        // ignore: strong_mode_implicit_dynamic_method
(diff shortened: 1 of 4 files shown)


packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
```diff
@@ -23,11 +23,9 @@ class FirebaseDynamicLinks {
   /// there is no pending dynamic link or any call to this method after the
   /// the first attempt.
   Future<PendingDynamicLinkData> retrieveDynamicLink() async {
-    final Map<dynamic, dynamic> linkData =
-        // TODO(amirh): remove this on when the invokeMethod update makes it to stable Flutter.
-        // https://github.com/flutter/flutter/issues/26431
-        // ignore: strong_mode_implicit_dynamic_method
-        await channel.invokeMethod('FirebaseDynamicLinks#retrieveDynamicLink');
+    final Map<String, dynamic> linkData =
```
bparrishMines: There is a chance that `linkData` could be `null` based on the line below it.
```
if (linkData == null) return null;
```

I think this case will require using `invokeMethod<dynamic>`.

cyanglaz: `channel.invokeMapMethod<String, dynamic>` can also return null now if the platform pass a null to dart. See here: https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/services/platform_channel.dart#L343-L345
If that's what you meant.

bparrishMines: Awesome! I didn't know they changed it.

PR #1672, opened 2019-05-31
[local_auth] The thread for executor needs to be UI thread.

mehmetf:
## Description

Two bug fixes:
- The thread passed to FP needs to be UI thread. Callbacks are executed on that thread and we also don't want any synchronization issues.
- We should not be calling .error() then .success() on the same result object. This does not cause any exceptions but it is bad form.

### DIFF
--- a/packages/local_auth/CHANGELOG.md
+++ b/packages/local_auth/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.5.2
+* Executor thread needs to be UI thread.
+
 ## 0.5.1
 * Fix crash on Android versions earlier than 28.
 * [`authenticateWithBiometrics`](https://pub.dev/documentation/local_auth/latest/local_auth/LocalAuthentication/authenticateWithBiometrics.html) will not return result unless Biometric Dialog is closed.
--- a/packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
+++ b/packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
@@ -24,7 +24,7 @@ import androidx.biometric.BiometricPrompt;
 import androidx.core.hardware.fingerprint.FingerprintManagerCompat;
 import androidx.fragment.app.FragmentActivity;
 import io.flutter.plugin.common.MethodCall;
-import java.util.concurrent.Executors;
+import java.util.concurrent.Executor;
 
 /**
  * Authenticates the user with fingerprint and sends corresponding response back to Flutter.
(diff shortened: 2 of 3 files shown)


mehmetf:
@mklim I need to submit this asap since the plugin is currently broken. Please review when you get the chance.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
```diff
@@ -164,19 +170,17 @@ public void onActivityPaused(Activity ignored) {
...
-              },
-              100);
+      uiThreadExecutor.handler.postDelayed(
+          new Runnable() {
+            @Override
+            public void run() {
+              prompt.authenticate(promptInfo);
+            }
+          },
+          100);
```
mklim: I realize the code was doing this already, but why the `postDelayed` with the 100ms delay here?

packages/local_auth/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.5.2
```
mklim: Ideally this version would actually be `0.5.1+1` since it's a bugfix. This is technically incrementing the minor digit. Reason is that pub has inconsistent semantic versioning policies depending on how mature the package is. For packages that are < `1.0.0` it uses `0.major.minor+patch` for semver, and then for anything at or above 1 it uses the standard `major.minor.patch`. Not a big deal though.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
mehmetf: See comment above:

```
// When activity is resuming, we cannot show the prompt right away. We need to post it to the
// UI queue.
```

mklim: My confusion is around what's actually stopping us from showing the prompt immediately, and why 100ms specifically as the delay. Is it because we're waiting for `onResume` to totally finish, and we're hoping that `100ms` is enough time? There's an [onPostResume](https://developer.android.com/reference/android/app/Activity.html#onPostResume%28%29) callback that isn't used much but I think would work more consistently here. I'm not clear from the comment if it's `onResume` specifically that's the issue or if there's some other underlying state that we're waiting for.

Either way like I said, the code was doing this already--just noticed it now. I can file an issue to clean this up if needed.

mehmetf: I don't like it either. We had to use it because onPostResume is only available after [Android Q](https://developer.android.com/reference/android/app/Application.ActivityLifecycleCallbacks.html#onActivityPostResumed(android.app.Activity)).

The 100ms delay? Completely random. I just tried a regular `post` instead of `postDelayed` and it worked. I will send a PR.

PR #1676, opened 2019-05-31
[local_auth] Use post instead of postDelayed

mehmetf:


packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
```diff
@@ -173,14 +173,13 @@ public void onActivityResumed(Activity ignored) {
       final BiometricPrompt prompt = new BiometricPrompt(activity, uiThreadExecutor, this);
       // When activity is resuming, we cannot show the prompt right away. We need to post it to the
```
mklim: nit: This comment can be deleted now.

mehmetf: That's still valid. I just removed the delay. We are still posting it to UI thread rather than showing it immediately.

PR #1687, opened 2019-06-03
[firebase_dynamic_links] Change architecture to work correctly on async links and launch links

jherencia:
## Description

This pull request changes firebase_dynamic_links architecture to be able to difference between the dynamic link which opened the app and links clicked on app active or in background.

## Related Issues

* https://github.com/flutter/flutter/issues/28062
* https://github.com/flutter/flutter/issues/24629
* https://github.com/flutter/flutter/issues/27439
* https://github.com/flutter/flutter/issues/21456

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/firebase_dynamic_links/README.md
```diff
@@ -117,24 +118,38 @@ class MyHomeWidgetState extends State<MyHomeWidget> {
   @override
   void initState() {
     super.initState();
-    _retrieveDynamicLink();
+    this.initDynamicLinks();
   }
 
-  Future<void> _retrieveDynamicLink() async {
-    final PendingDynamicLinkData data = await FirebaseDynamicLinks.instance.retrieveDynamicLink();
+  void initDynamicLinks() async {
```
collinjackson: Seems like this should be `Future<void>` right?

If it doesn't need to be a public method, using an underscore seems more appropriate

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
```diff
@@ -104,3 +140,13 @@ class PendingDynamicLinkDataIOS {
   /// declares higher [minimumVersion] than currently installed.
   final String minimumVersion;
 }
+
+class OnLinkErrorException {
```
collinjackson: I think we might want to call this class `OnLinkError` or `OnLinkErrorData` since it doesn't extend `Exception`

packages/firebase_dynamic_links/README.md
jherencia: Fixed in the latest commit: https://github.com/flutter/plugins/pull/1687/commits/b5c45c262941f09ed00b90c4076b7fbfd67871dc

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
jherencia: Fixed in the latest commit: https://github.com/flutter/plugins/pull/1687/commits/b5c45c262941f09ed00b90c4076b7fbfd67871dc

Thank you :)

bparrishMines:
Hi @jherencia,

I'm unable to reproduce these bugs with the latest version of `firebase_dynamic_links`: [0.4.0+3](https://pub.dev/packages/firebase_dynamic_links/versions/0.4.0+3). I think we just forgot to close all the bugs related to the same issue. What version are you using? I believe these bugs were fixed with https://github.com/flutter/plugins/pull/1569

jherencia:
Hello @bparrishMines,

This PR had two goals:

1. to fix that issue, which I see it is fixed in the latest version
2. to change the architecture of the plugin to be able to subscribe to dynamic links events while the application is active or in background in a consistent way with firebase_messaging. I believe that it is a better approach than needing to subscribe to application lifecycle changes.

I've made another PR to firebase_messaging (https://github.com/flutter/plugins/pull/1709) in the same direction.

jherencia:
Hi @bparrishMines, I assume you didn't have time to see this.

Do you know when you'll be able to check this, I'm going to be off in some days and to be able to answer your questions. I really think this is a really important improvement.

bparrishMines:
@jherencia Unfortunately, this plugin is still lacking integration testing and we are still working on our side to potentially get expresso to work with flutter testing tools (e.g. flutter driver). Since '1.' has already been solved, I don't believe we should go forward with an architecture change without integration tests to make sure everything will still work. 

Also, we typically want the dart api to match the native api. The Android API doesn't use a stream to receive links, so I don't believe an architecture change is necessary if it isn't necessary to solve a bug or crash.

jherencia:
Thank you @bparrishMines for the response, I'm going to try explain why I think this PR is needed even if we do not have integration tests.

1. First of all, the main reason is that the current implementation / architecture does not match the native API either, my PR goes in the direction of providing a consistent way to retrieve the dynamic link that launched the application.

As easy as:
`
final PendingDynamicLinkData data = await FirebaseDynamicLinks.instance.retrieveDynamicLink();
`
which follows how Firebase library works in the native part:
`
FirebaseDynamicLinks.getInstance()
        .getDynamicLink(registrar.activity().getIntent())
        .addOnSuccessListener(
            registrar.activity(),
            new OnSuccessListener<PendingDynamicLinkData>() {
              @Override
              public void onSuccess(PendingDynamicLinkData pendingDynamicLinkData) {
                 // onSuccess logic
              }
            })
        .addOnFailureListener(
            registrar.activity(),
            new OnFailureListener() {
              @Override
              public void onFailure(@NonNull Exception e) {
                // onFailure logic
              }
            });
`
but the current architecture requires the Dart application to listen to to AppLifecycleChanges, which adds a lot of complexity and risks of failure in the Dart side. Why? It should not be that hard to retrieve the dynamic link.

2. My PR proposal clearly separates launchLinks and links open when the application is running, which gives the developer more power and follows how Firebase does this in the native way. You subscribe a listener to a new intent as an event, you do not have to check on lifecycle changes:

`
@Override
  public boolean onNewIntent(Intent intent) {
    FirebaseDynamicLinks.getInstance()
        .getDynamicLink(intent)
        .addOnSuccessListener(
            registrar.activity(),
            new OnSuccessListener<PendingDynamicLinkData>() {
              @Override
              public void onSuccess(PendingDynamicLinkData pendingDynamicLinkData) {
                // onSucessLogic
              }
            })
        .addOnFailureListener(
            registrar.activity(),
            new OnFailureListener() {
              @Override
              public void onFailure(@NonNull Exception e) {
                // onFailure logic
              }
            });

    return false;
  }
`

I did this using a Future following the best practices used in other plugins and in this one.

3. In conclusion, we've developed 2 different applications with the architecture and lead us to a lot of complexity which was eased with our change.

So I understand why not having integration tests could be a blocker, but I think asking developers to subscribe to app lifecycle changes and to use an inconsistent API is blocker too, as long as makes their life hard and increases the risk of bugs and edge cases in which the "dynamic link" is not syncronized between the native part and the dart part, which has happened a lot in the past and lead to "not so" elegant solutions like:

https://github.com/flutter/plugins/commit/bc04a77dbc55a765000e9b0c9c0558d3433c75d3#diff-3f6c424218a2662c8ddc4b322f30bf66

Please, do not miss understand my comment, I understand the hard work behind this but I'm just trying to point that it could me improved with this PR.

I'm happy to make any changes that you think could be better.

Best regards.

bparrishMines:
Thanks for taking the time to write this out. I agree with your reasoning and believe this feature should be implemented. However, I think it should be implemented as a convenience method rather than an architecture change.

I could imagine an instance where someone would only want to retrieve a single dynamic link, and not want to listen for callbacks. For example, [email link authentication](https://firebase.google.com/docs/auth/android/email-link-auth). Not to mention some users may currently be using `retrieveDynamicLink` with their Lifecycle flow and this would cause a regression for them.

Would it be possible to keep the current architecture and add the listener? I also think an `EventChannel` would be better suited for the listener. The [sensors](https://github.com/flutter/plugins/blob/master/packages/sensors/android/src/main/java/io/flutter/plugins/sensors/SensorsPlugin.java) plugin is a good example of using this.

LMK what your thoughts on this are.

collinjackson:
We may be able to draw some inspiration from the React Native Firebase implementation here.  Looks like there are two ways that they allow developers to receive links:

1. [`getInitialLink()`](https://github.com/invertase/react-native-firebase/blob/master/packages/links/lib/index.d.ts#L515) which which looks at `launchOptions[UIApplicationLaunchOptionsURLKey]` (we don't seem to be doing this yet) as well as `continueUserActivity` on iOS. 
1. [`onLink`](https://github.com/invertase/react-native-firebase/blob/master/packages/links/lib/index.d.ts#L545) which provides notifications of all links received after the callback is set.

Interestingly errors aren't passed over the bridge. They're simply logged to the console. https://github.com/invertase/react-native-firebase/blob/978b298f94940db7776cd6c4117eb6b618c92824/packages/links/ios/RNFBLinks/RNFBLinksAppDelegateInterceptor.m#L74

In the Dart version, `onLink` might be more Dart-y to expose as a `Stream` rather than setting callbacks but I could go either way on that. If we do callbacks `onLink` as a setter might make sense rather than a `configure()` method.

`EventChannel` does seem more appropriate than `MethodChannel` here but it doesn't affect the developer either way.

FYI @Ehesp

jherencia:
@collinjackson 

Yes, the architecture proposed is based in RNFirebase.

- `getInitialLink()` is what I called `getLauchLink` to be in line with firebase_messaging naming
- `onLink` is configure with two callbacks OK and KO, again to be in line with firebase_messaging which uses this same behaviour

I am happy to do any change you think is needed: change the naming, migrate from `MethodChannel` to `EventChannel`, etc.

See https://github.com/flutter/plugins/tree/master/packages/firebase_messaging and my other PR (https://github.com/flutter/plugins/pull/1709)

@bparrishMines 
The architecture change comes because the dependence on Lifecycle subscription is wrong and leads to mistakes as long as forces developers to handle things in an architecture that is harder than it should be. 

RNFirebase does this exact thing:
- offers a way to get the launch / link => await `getLaunchLink`
- offers a way to be notified when a new link is pressed => `onLink` which I called `configure`

The PR follows the same architecture.

packages/firebase_dynamic_links/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.5.0
+
+* **Breaking change**. Changed architecture and method names to be able to differentiate between
```
bparrishMines: I would include what methods are no longer available and what methods replaced them.

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
```diff
@@ -4,6 +4,9 @@
 
 part of firebase_dynamic_links;
 
+typedef Future<dynamic> OnLinkSuccessHandler(PendingDynamicLinkData linkData);
```
bparrishMines: I would use the newer `Function` declaration. It's now preferred by the dart team. [Style guide](https://dart.dev/guides/language/effective-dart/design#dont-use-the-legacy-typedef-syntax)

```dart
typedef OnLinkSuccessCallback = Future<dynamic> Fucntion(PendingDynamicLinkData linkData);
```

Also, I believe `Callback` is preferred over `Handler` for these. (e.g. [VoidCallback](https://api.flutter.dev/flutter/dart-ui/VoidCallback.html))

packages/firebase_dynamic_links/README.md
```diff
@@ -97,7 +97,8 @@ Receiving dynamic links on *iOS* requires a couple more steps than *Android*. If
 applinks:YOUR_SUBDOMAIN.page.link
 ```
 
-4. To receive a dynamic link, call the `retrieveDynamicLink()` method from `FirebaseDynamicLinks`:
+4. To receive a dynamic link, call the `getLaunchLink()` method from `FirebaseDynamicLinks` and
```
bparrishMines: I would include that get launch link only retrieves the link that opens the app.

bparrishMines:
@jherencia After giving this more thought, I think we should go through with the change and avoid depending on the Lifecycle changes. I left an initial review with a few comments

packages/firebase_dynamic_links/CHANGELOG.md
jherencia: Done!

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
jherencia: Done!

packages/firebase_dynamic_links/README.md
jherencia: Done!

jherencia:
@bparrishMines 

Great news :).

I pushed two commits:
1.  With your changes
2. Renaming configure method to onLink as @collinjackson suggested. It's easy to revert if you both decide between on or the other.

I haven't migrated from `MethodChannel` to `EventChannel` as long as it is a big change and probably affects to tests. If you think it should be changed, let me know and I'll work on that.

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
```diff
@@ -50,6 +61,33 @@ class FirebaseDynamicLinks {
...
+  }
+
+  Future<dynamic> _handleMethod(MethodCall call) async {
+    switch (call.method) {
+      case "onLinkSuccess":
+        final Map<dynamic, dynamic> data =
+            call.arguments.cast<dynamic, dynamic>();
+        final PendingDynamicLinkData linkData =
+            getPendingDynamicLinkDataFromMap(data);
+        return _onLinkSuccess(linkData);
```
bparrishMines: Wouldn't this cause a `NullPointerException` if `onLink` was not called.

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
```diff
@@ -50,6 +61,33 @@ class FirebaseDynamicLinks {
...
+      case "onLinkError":
+        final Map<dynamic, dynamic> data =
+            call.arguments.cast<dynamic, dynamic>();
+        final OnLinkErrorException e = OnLinkErrorException._(
+            data['code'], data['message'], data['details']);
+        return _onLinkError(e);
```
bparrishMines: Same as above.

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
```diff
@@ -104,3 +142,13 @@ class PendingDynamicLinkDataIOS {
   /// declares higher [minimumVersion] than currently installed.
   final String minimumVersion;
 }
+
+class OnLinkErrorException {
```
bparrishMines: I think you may have changed this back to `OnLinkErrorException`.

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
jherencia: If `onLink` is not called, then `_handleMethod` won't be call either, as long as `channel.setMethodCallHandler` wouldn't have been called.

Am I wrong?

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
jherencia: What you mean? It's already `OnLinkErrorException`

bparrishMines: In the comment here: https://github.com/flutter/plugins/pull/1687#discussion_r294047856, you said you changed it to `OnLinkErrorData`, but here its name is `OnLinkErrorException`.

I might be looking at the wrong commits tho.

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
bparrishMines: You're right, good point. I missed where you called `channel.setMethodCallHandler`. However, what happens if a user passes `null` to `onLink`? Can a user unregister from `onLink` callbacks or is this bad practice?

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
jherencia: Ok :)

packages/firebase_dynamic_links/lib/src/firebase_dynamic_links.dart
jherencia: Well, I think that's a bad practice and we could check that, but that problem exists in almost every plugin where we depend on the developers to follow the documentation and they do it wrong:

https://github.com/flutter/plugins/blob/master/packages/firebase_messaging/lib/firebase_messaging.dart

jherencia:
I've changed the name as suggested.

Thank you both :)

mohammed932:
thanks for help me to solve my issue

PR #1695, opened 2019-06-03
[url_launcher] Add missing template type parameter to `invokeMethod` calls. Bump minimum Flutter version to 1.5.0. Replace invokeMethod with invokeMapMethod wherever necessary.

cyanglaz:


iskakaushik:
er, this should be a patch version change

PR #1697, opened 2019-06-03
[google_maps_flutter] Add support for styling google maps

iskakaushik:
- supports both Android and iOS
- fixes: https://github.com/flutter/flutter/issues/24147

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/google_maps_flutter/example/lib/map_ui.dart
+++ b/packages/google_maps_flutter/example/lib/map_ui.dart
@@ -4,6 +4,7 @@
 
 import 'package:flutter/material.dart';
 import 'package:google_maps_flutter/google_maps_flutter.dart';
+import 'package:flutter/services.dart' show rootBundle;
 
 import 'page.dart';
 
@@ -49,6 +50,8 @@ class MapUiBodyState extends State<MapUiBody> {
   bool _zoomGesturesEnabled = true;
   bool _myLocationEnabled = true;
   bool _myLocationButtonEnabled = true;
+  GoogleMapController _controller;
+  bool _nightMode = false;
 
   @override
   void initState() {
@@ -183,6 +186,36 @@ class MapUiBodyState extends State<MapUiBody> {
     );
   }
 
+  Future<String> _getFileData(String path) async {
+    return await rootBundle.loadString(path);
+  }
+
+  void _setMapStyle(String mapStyle) {
+    setState(() {
+      _nightMode = true;
+      _controller.setMapStyle(mapStyle);
+    });
+  }
+
+  Widget _nightModeToggler() {
+    if (!_isMapCreated) {
+      return null;
+    }
+    return FlatButton(
+      child: Text('${_nightMode ? 'disable' : 'enable'} night mode'),
+      onPressed: () {
+        if (_nightMode) {
+          setState(() {
+            _nightMode = false;
+            _controller.setMapStyle(null);
+          });
+        } else {
+          _getFileData('assets/night_mode.json').then(_setMapStyle);
+        }
+      },
+    );
+  }
+
   @override
   Widget build(BuildContext context) {
     final GoogleMap googleMap = GoogleMap(
@@ -236,6 +269,7 @@ class MapUiBodyState extends State<MapUiBody> {
               _zoomToggler(),
               _myLocationToggler(),
               _myLocationButtonToggler(),
+              _nightModeToggler(),
             ],
           ),
         ),
@@ -256,6 +290,7 @@ class MapUiBodyState extends State<MapUiBody> {
 
   void onMapCreated(GoogleMapController controller) {
     setState(() {
+      _controller = controller;
       _isMapCreated = true;
     });
   }
--- a/packages/google_maps_flutter/ios/Classes/GoogleMapController.m
+++ b/packages/google_maps_flutter/ios/Classes/GoogleMapController.m
@@ -228,8 +228,12 @@ static double ToDouble(NSNumber* data) { return [FLTGoogleMapJsonConversions toD
     result(isMyLocationButtonEnabled);
   } else if ([call.method isEqualToString:@"map#setStyle"]) {
     NSString* mapStyle = [call arguments];
-    NSNumber* mapStyleSet = @([self setMapStyle:mapStyle]);
-    result(mapStyleSet);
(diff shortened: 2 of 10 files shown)


packages/google_maps_flutter/example/lib/map_ui.dart
```diff
@@ -183,6 +186,36 @@ class MapUiBodyState extends State<MapUiBody> {
...
+    if (!_isMapCreated) {
+      return null;
+    }
+    return FlatButton(
+      child: Text('${_nightMode ? 'disable' : 'enable'} night mode'),
+      onPressed: () {
+        if (_nightMode) {
+          setState(() {
+            _nightMode = false;
+            _controller.setMapStyle(null);
```
amirh: Can `_controller` be null here?

packages/google_maps_flutter/ios/Classes/GoogleMapController.m
```diff
@@ -308,6 +312,22 @@ - (void)setMyLocationButtonEnabled:(BOOL)enabled {
   _mapView.settings.myLocationButton = enabled;
 }
 
+- (BOOL)setMapStyle:(NSString*)mapStyle {
+  if (mapStyle == (id)[NSNull null] || mapStyle.length == 0) {
```
amirh: nit: we don't seem to cast to `(id)` elsewhere, do we need it?

packages/google_maps_flutter/ios/Classes/GoogleMapController.m
```diff
@@ -308,6 +312,22 @@ - (void)setMyLocationButtonEnabled:(BOOL)enabled {
...
+    _mapView.mapStyle = nil;
+    return YES;
+  }
+  NSError* error;
+  GMSMapStyle* style = [GMSMapStyle styleWithJSONString:mapStyle error:&error];
```
amirh: is there any useful information in error we should send back to Dart?

packages/google_maps_flutter/lib/src/controller.dart
```diff
@@ -173,6 +173,20 @@ class GoogleMapController {
...
+  /// Sets the styling of the base map.
+  ///
+  /// Set to `null` to clear any previous custom styling.
+  ///
+  /// Returns `true` if the style was successfully parsed; `false` if problems were
+  /// detected with the [mapStyle], including un-parsable styling JSON, unrecognized
+  /// feature type, unrecognized element type, or invalid styler keys. If the return
+  /// value is `false`, the current style is left unchanged.
+  ///
+  /// The style string can be generated using https://mapstyle.withgoogle.com/
```
amirh: Is there any documentation for the json format we can refer to?

domesticmouse: Here's the iOS style reference: https://developers.google.com/maps/documentation/ios-sdk/style-reference
I believe it is identical to Android: https://developers.google.com/maps/documentation/android-sdk/style-reference

packages/google_maps_flutter/ios/Classes/GoogleMapController.m
iskakaushik: I get an IDE warning about comparison between incompatible pointer types, we could either cast it to `(NSString*)` or `(id)` to address that and I chose the latter. It shouldn't be a big deal if we choose to remove the cast but seems reasonable to leave it.

packages/google_maps_flutter/ios/Classes/GoogleMapController.m
iskakaushik: The error has some more details about what the issue with the JSON is. Though it is useful to know what the error is for fixing the style, I don't think we can take any action based on the specifics of the error on the dart side in this case.

Typical action on error would be to keep the existing style (which is the default) or change to an empty/alternate style, which the user can currently do by checking the return value.

packages/google_maps_flutter/ios/Classes/GoogleMapController.m
amirh: Totally reasonable was just wondering why this is different from the other similar checks in this file. Taking another look I guess it's because the other instances here are access `NSArray` elements.

packages/google_maps_flutter/lib/src/controller.dart
iskakaushik: Thanks @domesticmouse , will add these to the docs.

packages/google_maps_flutter/ios/Classes/GoogleMapController.m
amirh: We should pass back whatever information we have to the Dart side. e.g it's possible that I'm building an app for editing and previewing map styles, in which case I would even want to show the error to the user.

packages/google_maps_flutter/example/lib/map_ui.dart
iskakaushik: No, this is only called from `_nightModeToggler` where we first check for `_isMapCreated` which is set alongside with `_controller`.

iskakaushik:
landing after a verbal LGTM from @amirh 👍

pedromassango:
Is there any date to publish a new version with this changes?

iskakaushik:
@pedromassango it's published `0.5.16`

McCossum:
Is there an example of a working implementation somewhere?

loint:
Thanks for great improvement guys

nephix:
Is there any documentation on how to use this?

iskakaushik:
@nephix the dart docs on the method over at https://github.com/flutter/plugins/pull/1697/files#diff-41940e5867e958815f51ef18747727edR189 should answer most questions, also look at the example app for how this is used.

BoLindJensen:
is it intended behavior that you can only style the "normal" map? 

mapType: MapType.hybrid,    // No styling applied. :(
mapType: MapType.satellite,  // No labels or markers is used, no styling can be applied :(
mapType: MapType.normal,   // Styling is used. :)
mapType: MapType.terrain,    // No styling applied. :(

Or am I doing something wrong?

ventr1x:
Nah, it is just not working and no documentation was provided at all.
Even if I copy the dark mode json from the PR, it simply is being ignored.
Welcome to google plugins! This is btw "production ready"...

PR #1701, opened 2019-06-04
[cloud_firestore] Invoke on ui thread only

kroikie:
## Description

In coming versions of the Flutter engine calls to the Dart side from the Java side must be done on the UI thread. The Firestore plugin implements transactions by making calls to the Dart side from the Java side. This change ensures that those calls are made on the UI thread.

## Related Issues

Resolves: https://github.com/flutter/flutter/issues/32657

### DIFF
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.12.2
+
+* Ensure that all channel calls to the Dart side from the Java side are done
+  on the UI thread. This change allows Transactions to work with upcoming
+  Engine restrictions, which require channel calls be made on the UI thread. 
+
 ## 0.12.1
 
 * Added support for `Source` to `Query.getDocuments()` and `DocumentReference.get()`.
--- a/packages/cloud_firestore/CHANGELOG.md
+++ b/packages/cloud_firestore/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 * Ensure that all channel calls to the Dart side from the Java side are done
   on the UI thread. This change allows Transactions to work with upcoming
-  Engine restrictions, which require channel calls be made on the UI thread. 
+  Engine restrictions, which require channel calls be made on the UI thread.
+  **Note** this is an Android only change, the iOS implementation was not impacted.
+* Updated the Firebase reporting string to `flutter-fire-fst` to be consistent
+  with other reporting libraries.
 
 ## 0.12.1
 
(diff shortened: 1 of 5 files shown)


packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/FlutterFirebaseAppRegistrar.java
```diff
@@ -7,8 +7,8 @@
 import java.util.List;
 
 public class FlutterFirebaseAppRegistrar implements ComponentRegistrar {
-  private static final String LIBRARY_NAME = "flutter-firebase_cloud_firestore";
-  private static final String LIBRARY_VERSION = "0.12.1";
+  private static final String LIBRARY_NAME = "flutter-fire-fst";
```
collinjackson: Should be mentioned in changelog

packages/cloud_firestore/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.12.2
+
+* Ensure that all channel calls to the Dart side from the Java side are done
```
collinjackson: You may want to mention that this is an Android-only change.

packages/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/cloudfirestore/FlutterFirebaseAppRegistrar.java
kroikie: Done

packages/cloud_firestore/CHANGELOG.md
kroikie: Done

packages/cloud_firestore/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+  on the UI thread. This change allows Transactions to work with upcoming
+  Engine restrictions, which require channel calls be made on the UI thread. 
+
```
larssn: Does this take into account that Firestore's Android SDK will not allow transactions on the UI thread?

kroikie: The transaction itself is not being executed on the UI thread, only the message being passed over the channel from native side to Dart side is being done on the UI thread.

ride4sun:
This does not fix flutter #18870

PR #1702, opened 2019-06-04
[google_maps_flutter]Marker drag event

duzenko:
## Description

Flutter bridge to OnMarkerDragListener 

## Related Issues

https://github.com/flutter/flutter/issues/28929


duzenko:
signed it

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -369,6 +371,17 @@ public boolean onMarkerClick(Marker marker) {
     return markersController.onMarkerTap(marker.getId());
   }
 
+  @Override
+  public void onMarkerDragStart(Marker marker) {}
```
iskakaushik: It would be great if you add support for `onMarkerDragStart` and `onMarkerDrag` as part of this PR as well.

packages/google_maps_flutter/lib/src/marker.dart
```diff
@@ -216,6 +217,8 @@ class Marker {
   /// Callbacks to receive tap events for markers placed on this map.
   final VoidCallback onTap;
 
+  final ValueChanged<LatLng> onDragEnd;
```
iskakaushik: Can you add an example to the `examples` directory that illustrates how we can use this feature?

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
duzenko: I could do that but I can't imagine the use case for it and thus can't add an example code for it. Moreover, the way complex types are passed between dart and native code (via JSON serialization) I don't want to introduce unnecessary CPU load when marker is being dragged.

packages/google_maps_flutter/lib/src/marker.dart
duzenko: Showing a dialog with the new position in the example app.
In our business app we use the drag function to edit map polygons but that's a bit too much work to repeat in the example.

duzenko:
IPA build check is seemingly broken. All recent PR's fail to pass it.

brux88:
are there any news?

Henkvanslochteren:
It looks like all the requests are done but its still showing up as one missing. @duzenko Any idea on how to fix that?

duzenko:
> 
> 
> It looks like all the requests are done but its still showing up as one missing. @duzenko Any idea on how to fix that?

@iskakaushik ?

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
stevenspiel: My use case is that I have polygons with draggable points. As the marker moves, the polygon would need to redraw. Is there some way to bypass the CPU load when onMarkerDrag isn't supplied?

duzenko: Do you absolutely need to update your poly's while dragging? I'm not sure if it's even possible.

Henkvanslochteren:
@iskakaushik this should be good to go, right?

duzenko:
> 
> 
> Yup. This looks good to me. @duzenko can you please update the pubspec and changelog?

Done

iskakaushik:
sorry for the trouble, but could you please fix the conflicts, i will keep and eye out and make sure i merge this PR in first :-)

duzenko:
> 
> 
> sorry for the trouble, but could you please fix the conflicts, i will keep and eye out and make sure i merge this PR in first :-)

@iskakaushik 
No problem 
But I'm traveling this and the next weeks and AFK most of the time. I'll try to find some time for this.

duzenko:
Fixed

joaquinperaza:
The callback receives the new position after drag event in the marker object?

duzenko:
> 
> 
> The callback receives the new position after drag event in the marker object?

That was the idea

PR #1707, opened 2019-06-05
Ensure that CocoaPods on Cirrus is always the latest version.

collinjackson:
Fixes test failures on the plugins iOS builder

```
Warning: CocoaPods minimum required version 1.6.0 or greater not installed. Skipping pod install.
  CocoaPods is used to retrieve the iOS and macOS platform side's plugin code that responds to your plugin usage on the Dart side.
  Without CocoaPods, plugins will not work on iOS or macOS.
  For more info, see https://flutter.dev/platform-plugins
To upgrade:
  brew upgrade cocoapods
  pod setup

```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -72,8 +72,7 @@ task:
     - brew install libimobiledevice
     - brew install ideviceinstaller
     - brew install ios-deploy
-    - brew install cocoapods
-    - brew link --overwrite cocoapods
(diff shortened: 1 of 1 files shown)


collinjackson:
> LGTM!
> Interesting that we had to 're-install' all the other dependencies but only need to upgrade cocoapods. Is cocoapods special that it is pre installed somehow?

Seems like it is, yeah.

PR #1729, opened 2019-06-12
Fix build failures due to CocoaPods version

collinjackson:
Recently the plugins CI started [failing](https://cirrus-ci.com/task/5233412101636096) after a CocoaPods version change in the framework repo: flutter/flutter@c8bf7abeee7cc06b985aa719396603ff43821507.

This change brings the plugins CI in line with what the main Flutter repository is doing


amirh:
Another case where https://github.com/flutter/flutter/issues/33440 would have saved us some trouble.

I wonder how this change affects our users, do they need to manually upgrade cocoapods when they get the new framework version?

PR #1733, opened 2019-06-12
[firebase_core] Automate version retrieval

kroikie:
## Description

* Automate the retrieval of the plugin's version for user agent reporting to Firebase.

### DIFF
--- a/packages/firebase_core/android/build.gradle
+++ b/packages/firebase_core/android/build.gradle
@@ -56,7 +56,7 @@ android {
     defaultConfig {
         minSdkVersion 16
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
-        buildConfigField "String", "FLUTTER_FIRE_VERSION", "\"${flutterFireVersion()}\""
+        versionName flutterFireVersion()
     }
     lintOptions {
         disable 'InvalidPackage'
--- a/packages/firebase_core/android/build.gradle
+++ b/packages/firebase_core/android/build.gradle
@@ -34,12 +34,29 @@ rootProject.allprojects {
 
 apply plugin: 'com.android.library'
 
+/**
+ * Get the current version of the plugin from the plugin's pubspec.yaml file.
+ *
+ * @return String representing version of the plugin.
+ */
+def flutterFireVersion() {
+    File pubspec = new File('../../pubspec.yaml')
+    def lines = pubspec.readLines()
+    for (line in lines) {
+        if (line.contains('version: ')) {
+            return "${line}".substring(9).trim()
+        }
+    }
+    throw new Exception('Version not found in pubspec.yaml file.')
+}
+
 android {
     compileSdkVersion 28
 
     defaultConfig {
         minSdkVersion 16
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        buildConfigField "String", "FLUTTER_FIRE_VERSION", "\"${flutterFireVersion()}\""
     }
     lintOptions {
         disable 'InvalidPackage'
--- a/packages/firebase_core/android/src/main/java/io/flutter/plugins/firebase/core/FlutterFirebaseAppRegistrar.java
+++ b/packages/firebase_core/android/src/main/java/io/flutter/plugins/firebase/core/FlutterFirebaseAppRegistrar.java
@@ -7,13 +7,10 @@ import java.util.Collections;
 import java.util.List;
 
 public class FlutterFirebaseAppRegistrar implements ComponentRegistrar {
-  private static final String LIBRARY_NAME = "flutter-fire-core";
-  private static final String LIBRARY_VERSION =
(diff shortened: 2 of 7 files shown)


packages/firebase_core/android/src/main/java/io/flutter/plugins/firebase/core/FlutterFirebaseAppRegistrar.java
```diff
@@ -8,7 +8,7 @@
 
 public class FlutterFirebaseAppRegistrar implements ComponentRegistrar {
   private static final String LIBRARY_NAME = "flutter-fire-core";
-  private static final String LIBRARY_VERSION = "0.4.0+3";
+  private static final String LIBRARY_VERSION = BuildConfig.FLUTTER_FIRE_VERSION;
```
collinjackson: I don't think you need to declare it here; you can just use `BuildConfig.VERSION_NAME` below.

packages/firebase_core/android/build.gradle
```diff
@@ -34,12 +34,29 @@ rootProject.allprojects {
...
+    throw new Exception('Version not found in pubspec.yaml file.')
+}
+
 android {
     compileSdkVersion 28
 
     defaultConfig {
         minSdkVersion 16
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        buildConfigField "String", "FLUTTER_FIRE_VERSION", "\"${flutterFireVersion()}\""
```
collinjackson: I think I would call this VERSION_NAME; FLUTTER_FIRE is a bit non standard compared to what other SDKs are doing

kroikie: Done

packages/firebase_core/android/src/main/java/io/flutter/plugins/firebase/core/FlutterFirebaseAppRegistrar.java
kroikie: done

packages/firebase_core/android/build.gradle
```diff
@@ -34,12 +34,29 @@ rootProject.allprojects {
 
 apply plugin: 'com.android.library'
 
+/**
+ * Get the current version of the plugin from the plugin's pubspec.yaml file.
+ *
+ * @return String representing version of the plugin.
+ */
+def flutterFireVersion() {
+    File pubspec = new File('../../pubspec.yaml')
```
Salakar: This may be cleaner:

```groovy
File pubspec = new File(project.projectDir.parentFile,  'pubspec.yaml')
```

project.projectDir = `<pluginRoot>/android`
parentFile = `<pluginRoot>`

----

Otherwise should probably use `File.separator` for Windows support instead of `/`

packages/firebase_core/android/src/main/java/io/flutter/plugins/firebase/core/FlutterFirebaseAppRegistrar.java
```diff
@@ -7,12 +7,10 @@
...
 public class FlutterFirebaseAppRegistrar implements ComponentRegistrar {
-  private static final String LIBRARY_NAME = "flutter-fire-core";
-  private static final String LIBRARY_VERSION = "0.4.0+3";
-
   @Override
   public List<Component<?>> getComponents() {
     return Collections.<Component<?>>singletonList(
-        LibraryVersionComponent.create(LIBRARY_NAME, LIBRARY_VERSION));
+        LibraryVersionComponent.create(
+            "flutter-fire-core", BuildConfig.VERSION_NAME.replaceAll("\\+", "-")));
```
Salakar: Does this replace all need to be done at runtime - can we move the logic to build time inside of the `flutterFireVersion()` method?

packages/firebase_core/ios/firebase_core.podspec
```diff
@@ -18,4 +18,11 @@ A new flutter plugin project.
   s.dependency 'Firebase/Core'
   s.ios.deployment_target = '8.0'
   s.static_framework = true
+
+  s.prepare_command = <<-CMD
```
Salakar: CocoaPods already provide the versions via an `NSBundle` for each podspec, as shown in my other PR - not sure if code gen is needed if we already have the version at runtime

packages/firebase_core/ios/firebase_core.podspec
```diff
@@ -18,4 +18,11 @@ A new flutter plugin project.
...
+    PUBSPEC_VERSION=`cat ../pubspec.yaml | grep version: | sed 's/version: //g'`
```
Salakar: It's possible to read the yaml file via the default Ruby yaml library that comes with OSX

```ruby
require 'yaml'
pubspec = YAML.load_file(File.join('..', 'pubspec.yaml'))

# ...

Pod::UI.puts "This plugin version is '#{pubspec['version']}'"
```

ened:
This PR currently breaks plugin integrations as shown in #34385.

PR #1735, opened 2019-06-12
Remove hard coded version check

collinjackson:
Previously we were checking to make sure that hard coded versions stayed in sync with pubspec.yaml. This is no longer necessary; see https://github.com/flutter/plugins/pull/1733


collinjackson:
Holding off on merging this until #1736 is finished

kroikie:
Should we wait till https://github.com/flutter/plugins/pull/1748 is applied to all FlutterFire plugins?

collinjackson:
> Should we wait till #1748 is applied to all FlutterFire plugins?

I'll wait until you've applied #1748 to all FlutterFire plugins.

collinjackson:
Auto reporting landed in https://github.com/flutter/plugins/pull/1760 so I've merged this

PR #1737, opened 2019-06-13
[google_maps_flutter] Allow BitmapDescriptor scaling override

duzenko:
## Description

Patch for BitmapDescriptor.fromAssetImage that respects ImageConfiguration.devicePixelRatio

## Related Issues

https://github.com/flutter/flutter/issues/34386

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/google_maps_flutter/lib/src/bitmap.dart
```diff
@@ -59,6 +59,13 @@ class BitmapDescriptor {
     AssetBundle bundle,
     String package,
   }) async {
+    if (configuration.devicePixelRatio != null) {
+      return BitmapDescriptor._(<dynamic>[
+        'fromAssetImage',
+        assetName,
+        configuration.devicePixelRatio,
+      ]);
+    }
```
iskakaushik: line 76 can be `configuration.devicePixelRatio ?? assetBundleImageKey.scale`

iskakaushik:
please also change the pubspec.yml and changelog files.

packages/google_maps_flutter/lib/src/bitmap.dart
duzenko: That would invoke unnecessary calls to 
```
    final AssetImage assetImage =
        AssetImage(assetName, package: package, bundle: bundle);
    final AssetBundleImageKey assetBundleImageKey =
        await assetImage.obtainKey(configuration);
```
I don't think we need to waste CPU time and battery doing that when scaling is already provided.

packages/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.17+1
```
iskakaushik: Can you change this to `0.5.19` please. Thanks

packages/google_maps_flutter/lib/src/bitmap.dart
iskakaushik: Makes sense.

packages/google_maps_flutter/CHANGELOG.md
duzenko: Done

duzenko:
I would still like to have a sync and/or contextless version of this function

asanoki:
This CL breaks loading assets of correct scale. It passes the path to 1.0 scaled asset, and the scale ratio to Java, where however the scale is discarded. So even if mipmaps are available, always the 1.0x asset will be loaded. Can we revert?

duzenko:
> 
> 
> This CL breaks loading assets of correct scale. It passes the path to 1.0 scaled asset, and the scale ratio to Java, where however the scale is discarded. So even if mipmaps are available, always the 1.0x asset will be loaded. Can we revert?

1. Can you attach a repro case/pull request for the example app
2. Did you try configuration.devicePixelRatio = null?

asanoki:
> This change should only kick in when devicePixelRatio != null which is exactly the idea behind it so what's broken?

That's true, but in order to load the correct mipmap on Flutter, I can't pass null. Here's what I'm doing:

    final imageConfiguration = createLocalImageConfiguration(context);
    _markerIcon = await BitmapDescriptor.fromAssetImage(
        imageConfiguration, 'assets/images/marker.png');

With this CL, mipmap of scale 1.0 will be loaded, even though the imageConfiguration has scale 3.0. Something is clearly broken, as the incorrect asset is loaded when you pass a scale in ImageConfiguration.

> Did you try configuration.devicePixelRatio = null?

Passing null doesn't help, as then mipmaps won't be used. Just tried. Also, not sure why would I override devicePixelRatio with null, if the one provided by Flutter's createLocalImageConfiguration is correct. It's just after this CL, the value is ignored.

duzenko:
Right
So what you need is a BitmapDescriptor of the image mipmap closest to given pixel ratio
While I need a BitmapDescriptor of the image with given pixel ratio.
This is a breaking change. I will commit a resolve now.

PR #1741, opened 2019-06-13
pubspec relative path incorrect

praneethfdo:
pubspec relative path incorrect


collinjackson:
Issue was fixed in #1742.

PR #1736, opened 2019-06-13
[draft] accessing a pubspec version at runtime

Salakar:
## Description

This PR is for example purposes (requested by @collinjackson) and is not intended to be merged.

This PR demos retrieving the `pubspec` version at runtime in Android & iOS - to be used for things like the Firebase plugin registrar.

### Notes

 - Only added this to one package, `cloud_firestore` - just as a demo
 - Android gradle code can be moved into a single gradle plugin - then it becomes a 1/2 line change for each package (we do this on RNFirebase)
 - iOS Bunde & version code could be moved into a shared package and exposed as a utility (e.g. on Firebase Core)
 - The `podspec` changes on its own might be a good one to apply already as is for all Flutter plugins - saves duplicating versions, descriptions, authors etc.

### Screenshots

Added some screenshots to show each platform works as expected.

#### Pod install

image

#### iOS via `NSBundle`

image

#### Android via `BuildConfig`

image

## Related Issues

N/A

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ ] All existing and new tests are passing.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [ ] I read and followed the [Flutter Style Guide].
- [ ] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [ ] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [ ] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [ ] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [ ] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[...]

PR #1742, opened 2019-06-13
[firebase_core] roll back to 0.4.0+3

collinjackson:
This PR reverts firebase_core to the 0.4.0+3 release due to a build issue with the 0.4.0+4 release that was not caught by our CI.

The automatic version checking feature will be re-landed once the build issue is resolved.

Fixes flutter/flutter#34410 flutter/flutter#34385 flutter/flutter#34393

See also https://github.com/flutter/plugins/pull/1736 which has some suggestions for additional improvements to automatic version checking.

### DIFF
--- a/packages/firebase_core/CHANGELOG.md
+++ b/packages/firebase_core/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 0.4.0+5
+
+* Rollback of automatic plugin version retrieval.
+
+## 0.4.0+4
+
+* Automate the retrieval of the plugin's version when reporting usage to Firebase.
+
 ## 0.4.0+3
 
 * Add missing template type parameter to `invokeMethod` calls.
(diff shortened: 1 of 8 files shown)


collinjackson:
> I don't think we need to update these files for the rollback:
> 
> 1. packages/firebase_core/example/ios/Runner.xcodeproj/project.pbxproj
> 2. packages/firebase_core/ios/Classes/.gitignore

I think we should roll back to the state before 0.4.0+4 was released, which includes reverting those files. We can re-land the changes together.

Salakar:
I left some comments on https://github.com/flutter/plugins/pull/1733 yesterday just before the PR landed, specifically:

image


Using the built-in gradle API would solve the pathing issue without having to use relative path strings, would also a probable Windows issue with having an incorrect path separator.

If I can help with anything let me know 👍

kroikie:
@Salakar Thanks for bringing that up, will definitely use that.

PR #1562, opened 2019-05-02
[video_player] Fix race condition on disposing the VideoController.

ianko:
## Description

After, or while in process of, disposing the VideoController, the `bufferingUpdate` is still being fired which causes the `ChangeNotifier._debugAssertNotDisposed` to fail.

I am adding the verification of the `_dispose` flag before  applying any change to the `value` variable. 

## Related Issues

flutter/flutter#25936

## Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #1743, opened 2019-06-13
[video_player] Make `initialize` not depend on duration, fix the live stream

ianko:
## Description

The `0.10.1` version introduced the condition to have the `duration` in order to consider the player initialized. That change made the **live stream videos** not able to play anymore, since it does not have a duration.

## Related Issues

flutter/flutter#35402

## Checklist

- [x] I signed the [CLA].

## Breaking Change


collinjackson:
Is there a better way to prevent the division by zero error that #1554 was trying to fix?

@iskakaushik @ened

ianko:
@collinjackson Didn't know that the division by zero was the issue there. I will definitely work in a solution that solves both cases.

collinjackson:
> @collinjackson Didn't know that the division by zero was the issue there. I will definitely work in a solution that solves both cases.

Great, thanks! Let me know when this PR is ready for re-review.

collinjackson:
Checking in, it looks like you're still working on this PR since the tests don't pass so I'm marking it WIP. Just let us know when it's ready for re-review. Thanks!

ianko:
@collinjackson We have a working solution already.

Taking some time to fix the errors from the example app. Will be finished by tomorrow.

bduyng:
@googlebot confirmed!

ianko:
@collinjackson I am done with the PR.

Here are some things to notice:

1) We have added a better validation for the `duration` in the VideoPlayerPlugin.m file;
2) Better validation and checks in the `video_player.dart`;
3) Added a `Live` tab in the example app;

When testing the example app, I've noticed that, everytime we switch to a tab that plays a video from the assets, an exception is thrown and the other videos start to play as well. **This was not introduced by this PR.** I went back and tested previous versions and it does the same.

I refactored the whole example app trying to find the source of the problem but it's not in it. It's something from the player itself and I couldn't find the reason. That's probably why there was a `controller.setVolume(0.0)` in the `deactive()` method. To hide the fact that the video was playing.

I am sending the PR with the refactored example app (all the features remain the same). I believe it's cleaner this way. Let me know what you think.

**About the failing CLA:** @bduyng is the co-author of this PR. He already signed the CLA and also sent a [confirmation message here](https://github.com/flutter/plugins/pull/1743#issuecomment-508790942).

cyanglaz:
@ianko Thank you for the contribution. The PR seems pretty long and it includes different things. Do you mind to separate the refactor and the race condition fixing to different PRs?

ianko:
> @ianko Thank you for the contribution. The PR seems pretty long and it includes different things. Do you mind to separate the refactor and the race condition fixing to different PRs?

@cyanglaz,

It was already a different PR (#1562). I had to bring to try to fix the issue with the video from assets that I mentioned, but it does not solve this case.

Added a reference to a related issue flutter/flutter#35402 opened recently.

cyanglaz:
@ianko Thanks, this PR seems still contains multiple stuff: the duration method addition and a refactoring of the example app. Do you mind separate them so one PR only contains a single feature/fix?

ianko:
@cyanglaz Done.

#1562 - Fix race condition
#1743 - (this) Fix the live stream (initialize does not depend on duration)
#1829 - Example app refactor

marianoarga:
I can confirm that 0.10.1 causes this issue
https://github.com/flutter/flutter/issues/35402
Using the previous version (0.10.0+8) works as spected.

Thank you @ianko

packages/video_player/ios/Classes/VideoPlayerPlugin.m
```diff
@@ -335,6 +331,17 @@ - (int64_t)position {
 }
 
 - (int64_t)duration {
+  if (CMTIME_IS_INDEFINITE([[_player currentItem] duration])) {
```
cyanglaz: I assume this is computing the length of loaded times in livestream. I personally would think the duration is the length of the whole video. Maybe create a different method for getting the "loadedTimes" for livestream videos?

packages/video_player/lib/video_player.dart
```diff
@@ -369,6 +370,19 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     );
   }
 
+  /// The duration in the current video.
+  Future<Duration> get duration async {
```
cyanglaz: It would seem to fail if called on Android.

packages/video_player/ios/Classes/VideoPlayerPlugin.m
```diff
@@ -305,10 +305,6 @@ - (void)sendInitialized {
     if (height == CGSizeZero.height && width == CGSizeZero.width) {
       return;
     }
-    // The player may be initialized but still needs to determine the duration.
```
cyanglaz: I think the quicker fix can be something like
```objc
if ([self duration] == 0 && ![self isDurationIndefinate])
```

And create a method to check if the duration isDurationIndefinate as simple as 
```objc
return CMTIME_IS_INDEFINITE([[_player currentItem] duration]);
```

This way we can avoid the breakage also we won't need to introduce the new api for duration anymore.

aletorrado:
Hi, could you please review the CLA task that is pending to move forward? This is pretty much needed. Thanks

ianko:
> Hi, could you please review the CLA task that is pending to move forward? This is pretty much needed. Thanks

I have signed already.

I believe that @bduyng [has also signed](https://github.com/flutter/plugins/pull/1743#issuecomment-508790942).

aletorrado:
Ok, I think a project admin is needed to change the `cla:no` tag to `cla:yes`.

cyanglaz:
@ianko Are you still planning to work on this? It seems that I left some review comments that haven't been addressed.

ianko:
> @ianko Are you still planning to work on this? It seems that I left some review comments that haven't been addressed.

@cyanglaz Unfortunately not in a short term.

HenglyEver:
This bug seems to be very old and needed for everyone as I was reading through a lot of threads, so can you please fix it?

VictorLekweuwa:
Please, is there a fix for this??

PR #1748, opened 2019-06-16
Auto retrieve version to report user agent

kroikie:
## Description

Automatically retrieve plugin version from `pubspec.yaml` to allow reporting of usage to Firebase.

This is the second attempt of https://github.com/flutter/plugins/pull/1733 which updated `firebase_core` plugin, this PR attempts to implement the auto version retrieval in the `firebase_remote_config` plugin to isolate any issues to a single plugin. If no issues result this will be implemented in the other FlutterFire plugins.


packages/firebase_remote_config/ios/firebase_remote_config.podspec
```diff
@@ -1,6 +1,11 @@
 #
 # To learn more about a Podspec see http://guides.cocoapods.org/syntax/podspec.html
 #
+
+require 'yaml'
+pubspec = YAML.load_file(File.join('..', 'pubspec.yaml'))
+ffVersion = pubspec['version'].gsub('+', '-')
```
collinjackson: I would name this something other than ffVersion as this feels like a non standard variable name. Perhaps just libraryVersion

packages/firebase_remote_config/ios/firebase_remote_config.podspec
```diff
@@ -18,5 +23,11 @@ Firebase Remote Config plugin for Flutter.
   s.dependency 'Flutter'
   s.dependency 'Firebase/RemoteConfig'
   s.static_framework = true
+
+  s.prepare_command = <<-CMD
```
collinjackson: I see you're doing this using the prepare_command that I demonstrated as opposed to the NSBundle approach that @Salakar suggested.

After thinking about it I think that @Salakar's bundle based approach seems a bit cleaner and idiomatic to CocoaPods. I like that it automates more of the content of the podspec file (as opposed to having a fake version of 0.0.1 being reported) and I think we should consider doing it version to be reported when Flutter runs pod install, and I think we should consider doing it in the plugin template in the future (filed flutter/flutter#34565)

The reason why it wasn't working for you is that a line that needs to be added to the podspec to make it work. (Confusingly, once you build and remove that line it still works, which is probably why @Salakar didn't include it in the PR.)

I've pushed a commit to @Salakar's branch with the missing line: https://github.com/flutter/plugins/pull/1736/commits/aedbe20db0181cd5c699cd76fe2517726f2cf66e

.gitignore
```diff
@@ -33,3 +33,4 @@ GeneratedPluginRegistrant.java
 build/
 .flutter-plugins
 
+UserAgent.h
```
collinjackson: Can remove this change once switching to the NSBundle based approach (see comments below)

packages/firebase_remote_config/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.2.0+3
+
+* Auto retrieve pubspec version to report usage to
+  Firebase.
```
collinjackson: Slightly more formal here, and also I don't think you need the extra linebreak.

```suggestion:-2+1
* Automatically use version from pubspec.yaml when reporting usage to Firebase.
```

.gitignore
collinjackson: I think I prefer to put this in a plugin-specific .gitignore rather than the root of the first-party plugins repo since it feels sort of like an implementation detail of a subset of the plugins.

packages/firebase_remote_config/android/user-agent.gradle
```diff
@@ -0,0 +1,25 @@
...
+String libraryVersionName = "UNKNOWN"
+String libraryName = "flutter-fire-rc"
+File pubspec = new File(project.projectDir.parentFile,  'pubspec.yaml')
+
+if (pubspec.exists()) {
+    String yaml = pubspec.text
+    // Using \s*['|"]?([^\n|'|"]*)['|"]? to extract version number.
+    Matcher versionMatcher = Pattern.compile("^version:\\s*['|\"]?([^\\n|'|\"]*)['|\"]?\$", Pattern.MULTILINE).matcher(yaml)
+    if (versionMatcher.find()) libraryVersionName = versionMatcher.group(1).replaceAll("\\+", "-")
+    println(libraryVersionName)
```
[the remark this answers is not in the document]
kroikie: Done

packages/firebase_remote_config/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Fri Jun 14 11:52:32 PDT 2019
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.1.1-all.zip
```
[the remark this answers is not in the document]
kroikie: Removed, we don't need to check in this file.

packages/firebase_remote_config/android/user-agent.gradle
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// Apply the library name and version to default config
```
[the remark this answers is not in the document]
kroikie: Done

packages/firebase_remote_config/ios/firebase_remote_config.podspec
kroikie: used `libraryVersion`

packages/firebase_remote_config/ios/firebase_remote_config.podspec
kroikie: Given the limitation of `CFBundleShortVersionString` to report the full version string. I'd stick with the header file approach.

packages/firebase_remote_config/CHANGELOG.md
kroikie: Done

.gitignore
kroikie: Used plugin specific .gitignore

PR #1751, opened 2019-06-18
[in_app_purchase] Fixed code example error & README links 

creativecreatorormaybenot:
```dart
const Set<String> _kIds = <String>['product1', 'product2'].toSet();
```

That is not a valid assignment. Hence, I fixed it and on the way I also updated all the links in the README that either did not work at all or did not work from the [pub page](https://pub.dev/packages/in_app_purchase#-readme-tab-).


creativecreatorormaybenot:
@cyanglaz @mklim I thought this was a quick one. Is there something wrong with it?

packages/in_app_purchase/README.md
```diff
@@ -26,14 +26,14 @@ guides:
 * [In-App Purchase (App Store)](https://developer.apple.com/in-app-purchase/)
 * [Google Play Biling Overview](https://developer.android.com/google/play/billing/billing_overview)
 
-You can check out the [example app README](example/README.md) for steps on how
+You can check out the [example app README](https://github.com/flutter/plugins/blob/master/packages/in_app_purchase/example/README.md) for steps on how
```
cyanglaz: I remember @mklim mentioned that we couldn't use the absolute path here for some reason, maybe I remembered incorrectly.

packages/in_app_purchase/README.md
```diff
@@ -87,7 +87,8 @@ if (!available) {
 ### Loading products for sale
 
 ```dart
-const Set<String> _kIds = <String>['product1', 'product2'].toSet();
+// Set literals require Dart 2.2. Alternatively, remove `const` and use `<String>['product1', 'product2'].toSet()`.
```
cyanglaz: Nits:
```suggestion
// Set literals require Dart 2.2. Alternatively, use `Set<String> _kIds = <String>['product1', 'product2'].toSet()`.
```

packages/in_app_purchase/README.md
creativecreatorormaybenot: For me, the reason was that the links do not work on Pub. The absolute paths work from anywhere.

mklim: I think I was saying that they're slightly less convenient than working relative links in that we'll need to update them more if these files move around at all, but they should work fine. It is better to fix them than leave them broken.

creativecreatorormaybenot:
@mklim Done.

PR #1769, opened 2019-06-24
Update README to reflect the requirement to use fragment activity

mehmetf:
No code change. README update to clarify what type of activity is required. Resolves https://github.com/flutter/flutter/issues/33429


mehmetf:
Not sure README updates require a pub publish? If so, let me know and I will update the version too.

PR #1774, opened 2019-06-26
Add pre-submit checks to see if plugin versions have been incremented correctly

iskakaushik:
Addresses https://github.com/flutter/flutter/issues/29815


script/version_checker/lib/main.dart
```diff
@@ -0,0 +1,82 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
```
amirh: What do you think about putting this in the flutter_plugin_tools?

iskakaushik: Ah, did not realize that it was a thing. Makes more sense for this to live there. Will add it.

iskakaushik:
Will move this to: https://github.com/flutter/plugin_tools/tree/master/lib/src

iskakaushik:
moved to https://github.com/flutter/plugin_tools/pull/34/files

PR #1776, opened 2019-06-26
[all_plugins] Compile all plugins together

bparrishMines:
## Description

Add pubspec generating tool to `build_all_plugins_app.sh` script. This will add all the plugins in the repo to the **all_plugins** app and test compilation.


examples/all_plugins/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="io.plugins.all_plugins">
+    package="io.plugins.all_plugins"
+    xmlns:tools="http://schemas.android.com/tools">
+
+    <uses-sdk tools:overrideLibrary="io.flutter.plugins.camera"/>
```
bparrishMines: Required because camera plugin is 21+

script/build_all_plugins_app.sh
```diff
@@ -4,14 +4,15 @@
...
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
 REPO_DIR="$(dirname "$SCRIPT_DIR")"
 
 source "$SCRIPT_DIR/common.sh"
 check_changed_packages > /dev/null
 
 cd $REPO_DIR/examples/all_plugins
 flutter clean > /dev/null
+(cd "$REPO_DIR" && pub global run flutter_plugin_tools gen-pubspec --exclude firebase_core,firebase_ml_vision)
```
bparrishMines: Exclude `firebase_core` because all the other firebase plugins depend on it. This caused a problem when compiling. 

Exclude `firebase_ml_vision` because it requires iOS 9.0+ which caused a problem when compiling with `google_maps_flutter`. I didn't want to change either plugin in this PR.

examples/all_plugins/pubspec.yaml
```diff
@@ -1,72 +1,81 @@
...
+  flutter: 
     sdk: flutter
-
-  # The following adds the Cupertino Icons font to your application.
-  # Use with the CupertinoIcons class for iOS style icons.
-  cupertino_icons: ^0.1.2
+  android_alarm_manager: 
+    path: /Users/bmparr/Work/plugins/packages/android_alarm_manager
+  android_intent: 
+    path: /Users/bmparr/Work/plugins/packages/android_intent
```
collinjackson: I don't think you want to check in these local paths. Maybe revert and add pubspec.yaml to .gitignore or something.

script/build_all_plugins_app.sh
collinjackson: sgtm. note that we'll be moving these plugins out of this repo soon.

PR #1786, opened 2019-07-01
Disable version-check until we fix the patch version

iskakaushik:


script/incremental_build.sh
```diff
@@ -24,7 +24,7 @@ else
     echo "No changes detected in packages."
   else
     (cd "$REPO_DIR" && pub global run flutter_plugin_tools "${ACTIONS[@]}" --plugins="$CHANGED_PACKAGES" $PLUGIN_SHARDING)
-    echo "Running version check for changed packages"
-    (cd "$REPO_DIR" && pub global run flutter_plugin_tools version-check --base_sha="$(get_branch_base_sha)")
+    # echo "Running version check for changed packages"
```
amirh: Please explain why this is commented out and reference https://github.com/flutter/flutter/issues/35411

PR #1795, opened 2019-07-03
[image_picker] Don't use modules

mehmetf:
Our internal build system is allergic to modules. We try to avoid them as much as we can and in here it seems harmless to change upstream code.


mehmetf:
@mklim PTAL. This is a small change that already exists internally.

mehmetf:
Folks I am submitting this because without it we cannot update the image_picker plugin internally. Please comment post submit if you have concerns.

PR #1804, opened 2019-07-05
[image_picker] Removed cursor to prevent crash

AmolGangadhare:
Removed cursor and added MimeTypeMap.getFileExtensionFromUrl to prevent crash occurring in some scenarios.

## Description

*App is crashing in some scenarios while image picked from the file manager or other sources. In a line:182 while getting the mimeType for an image it was causing the app to crash as sometimes there is no mime type available for the image. Now, to get image extension cursor is removed and now extension will be taken from MimeTypeMap.getFileExtensionFromUrl*

## Related Issues

*Fixes flutter/flutter#35632*

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/image_picker/android/build.gradle
+++ b/packages/image_picker/android/build.gradle
@@ -21,7 +21,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.0'
+        classpath 'com.android.tools.build:gradle:3.4.1'
     }
 }
 
--- a/packages/image_picker/android/build.gradle
+++ b/packages/image_picker/android/build.gradle
@@ -21,7 +21,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.1'
+        classpath 'com.android.tools.build:gradle:3.3.0'
     }
 }
 
--- a/packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java
+++ b/packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java
@@ -172,15 +172,19 @@ class FileUtils {
     String extension = null;
 
     try {
-      extension = MimeTypeMap.getFileExtensionFromUrl(uriImage.getPath());
+      String imagePath = uriImage.getPath();
+      if (imagePath != null && imagePath.lastIndexOf(".") != -1) {
+        extension = imagePath.substring(imagePath.lastIndexOf(".") + 1);
+      }
     } catch (Exception e) {
       extension = null;
     }
 
-    if (extension == null) {
+    if (extension == null || extension.isEmpty()) {
       //default extension for matches the previous behavior of the plugin
       extension = "jpg";
     }
+
     return "." + extension;
   }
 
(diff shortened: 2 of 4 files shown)


packages/image_picker/android/build.gradle
```diff
@@ -21,7 +21,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.0'
+        classpath 'com.android.tools.build:gradle:3.4.1'
```
cyanglaz: Seems not related to this PR?

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java
```diff
@@ -168,25 +168,13 @@ private static String getPathFromRemoteUri(final Context context, final Uri uri)
...
-      if (cursor != null && cursor.moveToNext()) {
-        String mimeType = cursor.getString(0);
-
-        extension = MimeTypeMap.getSingleton().getExtensionFromMimeType(mimeType);
-      }
-    } finally {
-      if (cursor != null) {
-        cursor.close();
-      }
+      extension = MimeTypeMap.getFileExtensionFromUrl(uriImage.getPath());
```
cyanglaz: I did some research on this method and I found this: https://stackoverflow.com/questions/14320527/android-should-i-use-mimetypemap-getfileextensionfromurl-bugs/14321470
it seems getFileExtensionFromUrl is not sufficient for all scenarios?

packages/image_picker/android/build.gradle
AmolGangadhare: gradle was updated after the PR made, it would be better if added with this PR else not necessary.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java
AmolGangadhare: agreed, would be better to find the extension manually but in some scenarios, I found that file uri doesn't have an extension so in that case, it will also not sufficient. so checking for null or empty extension and applying manual extension will be working.

cyanglaz: Sure, do you mind to add code to manually check the extension as well? 
Something like 
```
extension = fromURI
if (extension == null) {
  extension = manuallyParsing
}
if (extension == null) {
  extension = ".jpg" (i believe this is already handled)
}

```

packages/image_picker/android/build.gradle
cyanglaz: ok, it'd be nice to rebase the PR so we won't see this diff here

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java
AmolGangadhare: did the change, null check for extension was already there, in addition to that I added an empty check to prevent rare scenarios.

cyanglaz:
LGTM! Please update the pubspec and changelog with the a new version, please also fix the formatting.

packages/image_picker/android/build.gradle
cyanglaz: Sorry i misunderstood your comment here. I think what I wanted to say is that let's remove change in this particular PR, since it is not related to the PR. Also please remove the related bit in the CHANGELOG

AmolGangadhare: ok, I'll remove gradle change and comment from the changelog.

AmolGangadhare: @cyanglaz I've restored the gradle version and updated changelog.

PR #1807, opened 2019-07-06
Fix firebase_admob issues caused by un-handled exception situations

archanpaul:
## Description

While using BannerAd from fairly complex application which expects BannerAd to be disposed gracefully there are multiple un-handled exception situations in the plugin which are addressed in following commits in https://github.com/archanpaul/flutter-plugins/commits/fix-firebase_admob

commit 7dfeeac4f43e781cff1f7d5c548cce166a5f32f9 (HEAD -> fix-firebase_admob, origin/fix-firebase_admob)
Author: Archan Paul <archan.paul@arputer.com>
Date:   Sat Jul 6 18:36:29 2019 +0530

    Handle exception in _invokeBooleanMethod(...) and return false in case of exception.

commit a2b1eac1408b9457ae8889629f0f15c5a60ff0ea
Author: Archan Paul <archan.paul@arputer.com>
Date:   Sat Jul 6 17:51:34 2019 +0530

    Interim fix for dangling BannerAd during dispose call.

## Related code and log references

```
  BannerAd _adbmobBanner = BannerAd(
      adUnitId: BannerAd.testAdUnitId,
      size: AdSize.smartBanner,
      targetingInfo: _targetingInfo,
      listener: (MobileAdEvent event) {
        _subjectAdEvent.sink.add(event);
        switch (event) {
          case MobileAdEvent.clicked:
            break;
          case MobileAdEvent.closed:
            break;
          case MobileAdEvent.failedToLoad:
            break;
          case MobileAdEvent.impression:
            break;
          case MobileAdEvent.leftApplication:
            break;
          case MobileAdEvent.loaded:
            break;
          case MobileAdEvent.opened:
            break;
          default:
            break;
        }
      },
    );
  ...
  Future dispose() async {
    await _adbmobBanner?.dispose();
  }

```

```
E/flutter (29705): [ERROR:flutter/lib/ui/ui_dart_state.cc(148)] Unhandled Exception: PlatformException(no_ad_for_id, dispose failed, no add exists for id=663181399, null)
E/flutter (29705): #0      StandardMethodCodec.decodeEnvelope 
package:flutter/…/services/message_codecs.dart:564
E/flutter (29705): #1      MethodChannel.invokeMethod 
package:flutter/…/services/platform_channel.dart:316
E/flutter (29705): <asynchronous suspension>
E/flutter (29705): #2      _invokeBooleanMethod 
package:firebase_admob/firebase_admob.dart:524
E/flutter (29705): <asynchronous suspension>
E/flutter (29705): #3      MobileAd.dispose 
package:firebase_admob/firebase_admob.dart:242
E/flutter (29705): #4      AdmobProvider.dispose 
package:fitzy_20190415/…/controllers/AdmobProvider.dart:39
E/flutter (29705): <asynchronous suspension>
E/flutter (29705): #5      _WidgetPromoState.dispose 
package:fitzy_20190415/…/widgets/widgetPromo.dart:36
E/flutter (29705): <asynchronous suspension>
E/flutter (29705): #6      StatefulElement.unmount 
package:flutter/…/widgets/framework.dart:4107
E/flutter (29705): #7      _InactiveElements._unmount 
package:flutter/…/widgets/framework.dart:1737
E/flutter (29705): #8      _InactiveElements._unmount.<anonymous closure> 
package:flutter/…/widgets/framework.dart:1735
E/flutter (29705): #9      MultiChildRenderObjectElement.visitChildren 
package:flutter/…/widgets/framework.dart:5181
E/flutter (29705): #10     _InactiveElements._unmount 
package:flutter/…/widgets/framework.dart:1733
E/flutter (29705): #11     ListIterable.forEach  (dart:_internal/iterable.dart:39:13)
E/flutter (29705): #12     _InactiveElements._unmountAll 
package:flutter/…/widgets/framework.dart:1746
E/flutter (29705): #13     BuildOwner.finalizeTree.<anonymous closure> 
package:flutter/…/widgets/framework.dart:2426
E/flutter (29705): #14     BuildOwner.lockState 
package:flutter/…/widgets/framework.dart:2258
E/flutter (29705): #15     BuildOwner.finalizeTree 
package:flutter/…/widgets/framework.dart:2425
E/flutter (29705): #16     _WidgetsFlutterBinding&BindingBase&GestureBinding&ServicesBinding&SchedulerBinding&PaintingBinding&SemanticsBinding&RendererBinding&WidgetsBinding.drawFrame 
package:flutter/…/widgets/binding.dart:702
E/flutter (29705): #17     _WidgetsFlutterBinding&BindingBase&GestureBinding&ServicesBinding&SchedulerBinding&PaintingBinding&SemanticsBinding&RendererBinding._handlePersistentFrameCallback 
package:flutter/…/rendering/binding.dart:285
E/flutter (29705): #18     _WidgetsFlutterBinding&BindingBase&GestureBinding&ServicesBinding&SchedulerBinding._invokeFrameCallback 
package:flutter/…/scheduler/binding.dart:1016
E/flutter (29705): #19     _WidgetsFlutterBinding&BindingBase&GestureBinding&ServicesBinding&SchedulerBinding.handleDrawFrame 
package:flutter/…/scheduler/binding.dart:958
E/flutter (29705): #20     _WidgetsFlutterBinding&BindingBase&GestureBinding&ServicesBinding&SchedulerBinding._handleDrawFrame 
package:flutter/…/scheduler/binding.dart:874
E/flutter (29705): #21     _rootRun  (dart:async/zone.dart:1124:13)
E/flutter (29705): #22     _CustomZone.run  (dart:async/zone.dart:1021:19)
E/flutter (29705): #23     _CustomZone.runGuarded  (dart:async/zone.dart:923:7)
E/flutter (29705): #24     _invoke  (dart:ui/hooks.dart:236:10)
E/flutter (29705): #25     _drawFrame  (dart:ui/hooks.dart:194:3)
E/flutter (29705):
```

### DIFF
--- a/packages/firebase_admob/lib/firebase_admob.dart
+++ b/packages/firebase_admob/lib/firebase_admob.dart
@@ -233,7 +233,11 @@ abstract class MobileAd {
   /// Disposing a banner ad that's been shown removes it from the screen.
   /// Interstitial ads can't be programmatically removed from view.
   Future<bool> dispose() {
-    assert(_allAds[id] != null);
+    // When dispose is called, following assert prevents graceful cleanup of resources
+    // and occationally causes exception in the plugin which results in dangling
+    // BannerAd in the app. Hence commenting it for now till better way of handling is
+    // added by the plugin maintainer.
+    // assert(_allAds[id] != null);
     _allAds[id] = null;
     return _invokeBooleanMethod("disposeAd", <String, dynamic>{'id': id});
   }
--- a/packages/firebase_admob/lib/firebase_admob.dart
+++ b/packages/firebase_admob/lib/firebase_admob.dart
@@ -518,10 +518,26 @@ class FirebaseAdMob {
   }
 }
 
+// Future<bool> _invokeBooleanMethod(String method, [dynamic arguments]) async {
+//   final bool result = await FirebaseAdMob.instance._channel.invokeMethod<bool>(
+//     method,
+//     arguments,
+//   );
+//   return result;
+// }
 Future<bool> _invokeBooleanMethod(String method, [dynamic arguments]) async {
-  final bool result = await FirebaseAdMob.instance._channel.invokeMethod<bool>(
-    method,
-    arguments,
-  );
+  bool result = false;
+  try {
+    result = await FirebaseAdMob.instance._channel.invokeMethod<bool>(
+      method,
+      arguments,
+    );
+  } catch (e) {
+    // In case of "no_ad_for_id" error return result as false. Don't expect app code
+    // to deal with the exception condition because app code does not know about
+    // platform implementation.
+    result = false;
+  }
+
   return result;
 }


archanpaul:
I signed CLA.

packages/firebase_admob/lib/firebase_admob.dart
```diff
@@ -233,7 +233,11 @@ abstract class MobileAd {
   /// Disposing a banner ad that's been shown removes it from the screen.
   /// Interstitial ads can't be programmatically removed from view.
   Future<bool> dispose() {
-    assert(_allAds[id] != null);
+    // When dispose is called, following assert prevents graceful cleanup of resources
+    // and occationally causes exception in the plugin which results in dangling
+    // BannerAd in the app. Hence commenting it for now till better way of handling is
+    // added by the plugin maintainer.
+    // assert(_allAds[id] != null);
```
bparrishMines: I think you could replace this line with something like

```
if (_allAds[id] == null) return Future<bool>.value(true);

Comment (archanpaul) [reply packages/firebase_admob/lib/firebase_admob.dart]: Thanks @bparrishMines for suggestion. I have commit changes at e0a82e1.

**Comment by archanpaul** [comment]
Thanks @collinjackson for your feedback. 

I think it is not easy to add a test case to simulate the exact scenario, which prompted me to added the patch. Nevertheless an unit test can be added in which a supposed to be non existing BannerAd.id (the Banner which is successfully disposed already) can be sent to disposeAd MethodCall to encounter "no_ad_for_id" error and return False in turn. Do you have any other idea?

> Can you please update pubspec.yaml and CHANGELOG for release?

Do you want me to create 0.9.0+2 ?

**Comment by archanpaul** [comment]
Thanks @collinjackson for your feedback.

I think it is not easy to add a test case to simulate the exact scenario, which prompted me to added the patch. In the unit test we need to create a scenario in which, while Platform is working on `loadBannerAd`/`showAd` ( `_bannerAd..load()..show()` .... not listening and acting based on event callbacks) real Ad (over a non LTE network), the banner object is released without waiting (eg: timeout scenario in real app). This might result into a scenario of `no_ad_for_id` error, which should return false. Do you have better idea?

For the benefit of plugin users, you may consider merging the current patch now and later add a comprehensive unit-test related to "`dispose()` returns false".

> Can you please update pubspec.yaml and CHANGELOG for release?

Do you want me to create 0.9.0+2 ?

**Comment by collinjackson** [comment]
> Thanks @collinjackson for your feedback.
> 
> I think it is not easy to add a test case to simulate the exact scenario, which prompted me to added the patch. In the unit test we need to create a scenario in which, while Platform is working on `loadBannerAd`/`showAd` ( `_bannerAd..load()..show()` .... not listening and acting based on event callbacks) real Ad (over a non LTE network), the banner object is released without waiting (eg: timeout scenario in real app). This might result into a scenario of `no_ad_for_id` error, which should return false. Do you have better idea?

This is a unit test, not an integration test; you should be able to simulate the no_ad_for_id error in the [`setMockMethodCallHandler`](https://github.com/flutter/plugins/blob/master/packages/firebase_admob/test/firebase_admob_test.dart#L20).

You could either throw a PlatformException in the unit test or (my preference) update the native side so that `disposeAd` returns false instead of a "no_ad_for_id" `PlatformException`.

> > Can you please update pubspec.yaml and CHANGELOG for release?
> 
> Do you want me to create 0.9.0+2 ?

Yes.

**Comment by archanpaul** [comment]
@collinjackson 

> You could either throw a PlatformException in the unit test or (my preference) update the native side so that disposeAd returns false instead of a "no_ad_for_id" PlatformException. Then you won't need to worry about catching and handling an exception on the Dart side, and there's no need to test the special handing of the "no_ad_for_id" on the Dart side.

I have added unit test case in 24cdbe6 where I throw `PlatformException` with corresponding error code.

I prefer Platform to make Dart side of plugin aware that `no_ad_for_id` happened and let Dart plugin make a note and take care of it (return false to upper layer).

> Can you please update pubspec.yaml and CHANGELOG for release?

Done.

File: packages/firebase_admob/test/firebase_admob_test.dart [hunk 2]
```diff
@@ -132,5 +142,24 @@ void main() {
...
+
+    test('noAdForId', () async {
+      log.clear();
+
+      final BannerAd banner = BannerAd(
+        adUnitId: BannerAd.testAdUnitId,
+        size: AdSize.banner,
+      );
+      final int id = banner.id;
+      invalidAdId = banner.id;
```
collinjackson: I think what you're doing here is fragile and might cause the other disposeAd tests to fail depending on the order that the tests run, because they're re-using the same id `testAdUnitId`.

You might want to reset `invalidAdId` to `null` in `setUp()`

archanpaul: @collinjackson thanks for pointing out. I have updated in f409fc4.

archanpaul:
@collinjackson 

I do not have any use-case where BannerAd.dispose() is required to be called multiple times. The dispose() should return immediately in case of non-existent adID .. there is no need to wait for PlatformException to happen which is anyway inevitable in this case.

Generally speaking, most of the app developers only care to know the bool return of dispose() and ensure that bannerAd = null after successful dispose.

if you want to enforce the app developer to know that dispose is called on a non-existent adID, instead of assert (which actually does not give any meaningful reason in log), you can either throw a Dart exception informing the developer the reason explicitly or throw the PlatformException as it is (though not intuitive).

I personally like to have hide() api call which should immediately hide the Ad instead of dispose(). As a Flutter app developer I will not like to await on dispose() to hide the Ad from UI. It is always preferable to have features like load() once (and sometimes await refresh() with corresponding event listner) the Ad and do show()  & hide() as an when required to make Admob and this plugin intuitive/useful for the developer. This is my wish, not related to this PR.

PR #1826, opened 2019-07-10
[firebase_auth] Register for iOS notifications to support phone auth

collinjackson:
Apparently the automatic method swizzling stopped working somewhere between v1.5.4-hotfix.2 and v1.7.8+hotfix.2 (it would be great to narrow that down further)

See https://firebase.google.com/docs/auth/ios/phone-auth#enable-app-verification

Fixes flutter/flutter#35267

Unfortunately I don't have a good solution to test for this yet, but have manually tested.

The disadvantage of this change as written is that it may result in firebase_auth users getting warnings about "missing Push Notification Entitlement" even after https://github.com/flutter/flutter/issues/9984 is fixed. But since that has affected every iOS Flutter submission up to this point, it seems like a low priority.

### DIFF
--- a/packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m
+++ b/packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m
@@ -81,11 +81,7 @@ int nextHandle = 0;
 
 - (void)application:(UIApplication *)application
     didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken {
-#ifdef DEBUG
-  [[FIRAuth auth] setAPNSToken:deviceToken type:FIRAuthAPNSTokenTypeSandbox];
-#else
   [[FIRAuth auth] setAPNSToken:deviceToken type:FIRAuthAPNSTokenTypeProd];
-#endif
 }
 
 - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary *)options {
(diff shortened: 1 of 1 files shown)


packages/firebase_auth/ios/Classes/FirebaseAuthPlugin.m
```diff
@@ -68,6 +69,29 @@ - (FIRAuth *_Nullable)getAuth:(NSDictionary *)args {
   return [FIRAuth authWithApp:[FIRApp appNamed:appName]];
 }
 
+- (bool)application:(UIApplication *)application
+    didReceiveRemoteNotification:(NSDictionary *)notification
+          fetchCompletionHandler:(void (^)(UIBackgroundFetchResult result))completionHandler {
+  if ([[FIRAuth auth] canHandleNotification:notification]) {
+    completionHandler(UIBackgroundFetchResultNoData);
```
cyanglaz: if completionHandler is nullable, `completionHandler(UIBackgroundFetchResultNoData);` might crash if completionHandler is null. 
Ignore this comment if completionHandler is nonnull.

collinjackson: This code sample comes straight from Firebase so I think it's probably safe.

collinjackson: https://firebase.google.com/docs/auth/ios/phone-auth

PR #844, opened 2018-10-12
[google_sign_in] Added NonNull annotations, reduce Guava usage

ened:

PR #1828, opened 2019-07-10
Revert "[google_sign_in] Added NonNull annotations, reduce Guava usage (#844)

cyanglaz:
This reverts commit 19bdccee76eb1063063fa5300b885d3cc70c9019.

PR https://github.com/flutter/plugins/pull/844

@ened The previous PR broke some example apps in the repo because they have not been migrated to AndroidX yet. We have to keep this commit reverted until we migrate example apps to androidX

### DIFF
--- a/packages/google_sign_in/CHANGELOG.md
+++ b/packages/google_sign_in/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.0.4
+
+* Revert changes in 4.0.3.
+
 ## 4.0.3	
 
 * Update guava to `27.0.1-android`.	
--- a/packages/google_sign_in/CHANGELOG.md
+++ b/packages/google_sign_in/CHANGELOG.md
@@ -1,8 +1,3 @@
-## 4.0.3
-
-* Update guava to `27.0.1-android`.
-* Add correct @NonNull annotations to reduce compiler warnings.
-
 ## 4.0.2
 
 * Add missing template type parameter to `invokeMethod` calls.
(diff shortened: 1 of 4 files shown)


mklim:
Landed on red because the breakage is because of the version in pub breaking the example app, so CI would definitely be red until the fix is published. I tested this with the example app locally edited to use the filesystem instead of pub versions, and saw the revert fix the compile failure.

ened:
Ok no problem to revert.

I see two ways to fix it again:
- depend on the legacy annotation library (AndroidX projects can use jetifier)
- depend on AndroidX annotation library 

Both times the actual google sign in library project would depend on it instead of the example.

A bit strange this landed on green and turned it red though, btw. Will prepare an updated PR with the first solution - @cyanglaz ok for you?

mklim:
Thanks for understanding @ened and offering to help this re-land.

>depend on the legacy annotation library (AndroidX projects can use jetifier)

Unfortunately Jetifier isn't working for Flutter plugins. It's incompatible with how Flutter plugins are built into Flutter Android apps. flutter/flutter#29328 is probably the most concise source of info on that issue right now. @blasten is working on fixing it, but until then the incompatibilities are unavoidably breaking.

Because of that I think it's better that we hold off on this until we get the plugins into a state where Jetifier is working. Right now there's no compile time support dependency in the plugin at all. Introducing either a legacy support or a new AndroidX dependency would cause incompatibilities in new cases where it's not strictly needed right now. Once Jetifier is working again I think this patch good to have, it does make sense for apps to move to a state where they're on AndroidX and using Jetifier so that they aren't hitting compatibility issues like this.

> A bit strange this landed on green and turned it red though, btw. 

Yeah, this is a side effect of how our CI is set up for these example apps. The example apps depend on the published versions of any secondary plugins they use. So in this case our CI didn't break until the new version was published and an example app that used it as a secondary dependency pulled the latest version in.

PR #1831, opened 2019-07-10
[firebase_crashlytics] Handle case where function isn't in class for stack

rmtmckenzie:
## Description

Currently, the stacktrace parser can't handle it when there is a method without a class in dart - it instead throws its own exception and loses the original one!

This PR handles the case where there is no class name. 

There is a slight caveat to the fix on Android - firebase reporting uses a StackTraceElement which requires a class (which makes sense in Java since you can't have Class-less methods), so on Android class-less methods are reported as `.<functionName>` whereas on iOS they are simply reported as `<className>`.

I don't think this is a big problem as there seem to be other differences between how the stacktraces are reported on Android and iOS anyways.

i.e. iOS: 
image

Android:
image

## Related Issues

* flutter/flutter#32178

## Checklist

- [X] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_crashlytics/android/src/main/java/io/flutter/plugins/firebase/crashlytics/firebasecrashlytics/FirebaseCrashlyticsPlugin.java
+++ b/packages/firebase_crashlytics/android/src/main/java/io/flutter/plugins/firebase/crashlytics/firebasecrashlytics/FirebaseCrashlyticsPlugin.java
@@ -104,7 +104,7 @@ public class FirebaseCrashlyticsPlugin implements MethodCallHandler {
       String methodName = errorElement.get("method");
 
       return new StackTraceElement(
-          className == null ? "" : className, methodName, fileName, Integer.parseInt(lineNumber));
+          className ?? "", methodName, fileName, Integer.parseInt(lineNumber));
     } catch (Exception e) {
       Log.e(TAG, "Unable to generate stack trace element from Dart side error.");
       return null;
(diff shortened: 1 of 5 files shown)


packages/firebase_crashlytics/android/src/main/java/io/flutter/plugins/firebase/crashlytics/firebasecrashlytics/FirebaseCrashlyticsPlugin.java
```diff
@@ -103,7 +103,8 @@ private StackTraceElement generateStackTraceElement(Map<String, String> errorEle
       String className = errorElement.get("class");
       String methodName = errorElement.get("method");
 
-      return new StackTraceElement(className == null ? "" : className, methodName, fileName, Integer.parseInt(lineNumber));
+      return new StackTraceElement(
+          className == null ? "" : className, methodName, fileName, Integer.parseInt(lineNumber));
```
collinjackson: I think `className ?? ""` would be more readable
```suggestion
          className ?? "", methodName, fileName, Integer.parseInt(lineNumber));
```

collinjackson: Oops, never mind, this is Java not Dart :)

rmtmckenzie: Unless you know something I don't, I don't think that Java has a null coalescing operator yet...

rmtmckenzie: But I agree that's definitely more readable 😄

rmtmckenzie:
stjet

PR #1832, opened 2019-07-10
[camera] Dart Interface for camera refactor

bparrishMines:
## Description

Beginning of refactor of the camera plugin. This only implements the camera interface and shared classes (e.g. camera mixins, `CameraChannel` & `CameraTesting` )

This also adds the `publish_to: none` to the `pubspec.yaml` until the refactor is finished.

nomnoml

## Related Issues

https://github.com/flutter/flutter/issues/31225

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/lib/new/src/camera_controller.dart
+++ b/packages/camera/lib/new/src/camera_controller.dart
@@ -1,0 +1,110 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:flutter/foundation.dart';
+
+import 'common/camera_interface.dart';
+
+/// Controls a device camera.
+///
+/// Use [CameraController.availableCameras] to get a list of available cameras.
+///
+/// This class is used as a simple interface that works for Android and iOS.
+///
+/// When using iOS, simultaneously calling [start] on two [CameraController]s
+/// will throw a [PlatformException].
+///
+/// When using Android, simultaneously calling [start] on two
+/// [CameraController]s may throw a [PlatformException] depending on the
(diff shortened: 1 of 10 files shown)


packages/camera/lib/new/src/camera_controller.dart
```diff
@@ -0,0 +1,110 @@
...
+
+import 'common/camera_interface.dart';
+
+/// Controls a device camera.
+///
+/// Use [CameraController.availableCameras] to get a list of available cameras.
+///
+/// This class is used as a simple interface that works for Android and iOS.
+///
+/// When using iOS, simultaneously calling [start] on two [CameraController]s
```
mklim: Thanks for clarifying!

Discussed offline, I'm thinking this is kind of a foot gun. Sounds like the current plugin behavior is to guard against this by disposing of multiple instances under the hood. I think we should probably maintain it here, but that doesn't need to block this PR.

packages/camera/lib/new/src/camera_controller.dart
```diff
@@ -0,0 +1,110 @@
...
+    switch (api) {
+      case CameraApi.android:
+        throw UnimplementedError();
+      case CameraApi.iOS:
+        throw UnimplementedError();
+      case CameraApi.supportAndroid:
+        throw UnimplementedError();
+    }
+
+    return null;
```
mklim: subjective, optional: Sorry, missed this. In general I think silently returning null in cases where we're not sure what to do is hard to deal with, since either developers need to always code against it or it causes NPEs somewhere murky down the line. I think it probably makes sense to log and error and throw here if we're not getting the API, WDYT?

bparrishMines: Discussed offline and added a comment

PR #1835, opened 2019-07-11
[firebase_dynamic_links] Allow FDL plugin to be registered without an activity

mehmetf:
Some apps do not have an activity available when they register plugins. We should still register the plugin and leave it up to the intent listener to grab the latest intent.


mehmetf:
@bparrishMines Ping for a quick review. This change is already made internally since it was breaking a team. We need to submit it upstream asap. Thanks!

PR #1838, opened 2019-07-11
[camera] Implement initialize and add documentation

bparrishMines:
## Description

To keep [backwards compatibility](https://github.com/flutter/plugins/blob/master/packages/camera/lib/camera.dart#L227), this implements the initialize method and CameraController maintains a singular static instance in the new interface. As per comment: https://github.com/flutter/plugins/pull/1832#discussion_r302288544

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/lib/new/src/camera_controller.dart
+++ b/packages/camera/lib/new/src/camera_controller.dart
@@ -93,13 +93,21 @@ class CameraController {
   /// call [initialize] on a [CameraController] while another is active, the old
   /// controller will be disposed before initializing the new controller.
   Future<void> initialize() {
+    final Completer<void> completer = Completer<void>();
+
     if (_instance == this) {
       return Future<void>.value();
     }
 
-    _instance?.dispose();
+    if (_instance != null) {
+      _instance
+          .dispose()
+          .then((_) => configurator.initialize())
+          .then((_) => completer.complete());
+    }
     _instance = this;
-    return configurator.initialize();
+
+    return completer.future;
   }
 
   /// Begins the flow of data between the inputs and outputs connected to the camera instance.
--- a/packages/camera/lib/new/src/camera_controller.dart
+++ b/packages/camera/lib/new/src/camera_controller.dart
@@ -65,8 +65,13 @@ class CameraController {
     );
   }
 
+  static const String _isNotInitializedMessage = 'Initialize was not called.';
+  static const String _isDisposedMessage = 'This controller has been disposed.';
+
   // Keep only one active instance of CameraController.
   static CameraController _instance;
+
+  bool _isDisposed = false;
 
   /// Details for the camera this controller accesses.
   final CameraDescription description;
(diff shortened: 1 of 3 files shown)


packages/camera/lib/new/src/camera_controller.dart
```diff
@@ -28,7 +35,6 @@ class CameraController {
   ///
   /// This will choose the best [CameraConfigurator] for the current device.
   factory CameraController({@required CameraDescription description}) {
-    assert(description != null);
```
bparrishMines: Unnecessary since its in the constructor.

packages/camera/lib/new/src/camera_controller.dart
```diff
@@ -75,14 +84,35 @@ class CameraController {
...
+  ///
+  /// Only one instance of [CameraController] can be active at a time. If you
+  /// call [initialize] on a [CameraController] while another is active, the old
+  /// controller will be disposed before initializing the new controller.
+  Future<void> initialize() {
+    if (_instance == this) {
+      return Future<void>.value();
+    }
+
+    _instance?.dispose();
```
mklim: Should this `await` the dispose?

packages/camera/lib/new/src/camera_controller.dart
```diff
@@ -59,6 +65,9 @@ class CameraController {
     );
   }
 
+  // Keep only one active instance of CameraController.
+  static CameraController _instance;
```
BugsBunnyBR: May I offer a different view about this approach? How about not having it as a static field? If the client developer unwisely calls `initialize` without disposing the previous CameraController, this may be a error, but also that may be that he is trying to access more than one camera at the same time (I don't know if this is possible in the hardware layer), but seems feasible to have more than one camera open and this to be alright.

Even if there is no chance of the developer accessing more than one camera at the same time, I believe that an exception should be raised explaining that the device is busy.

packages/camera/lib/new/src/camera_controller.dart
bparrishMines: Yea it does. But, I want this method to also be usable synchronously, so I don't want to make it `async`. I added a `Completer` to the method.

Also, I added tests to make sure the first controller is disposed.

packages/camera/lib/new/src/camera_controller.dart
bparrishMines: Thanks for the review!

Our reasoning to keep this functionality is because the current implementation of the plugin only allows one controller to be active at a time. It isn't documented in the plugin, but that is how it works underneath. While we refactor the platform specific code, we want to keep the interface the same and not cause many breaking changes.

The other reason for this functionality is because the `CameraController` will only serve as a simple to use and narrowly defined interface usable on any platform. For more complex features (like accessing multiple cameras), we will also expose access to this through another interface or require platform specific code.

PR #1844, opened 2019-07-12
[image_picker] Fix image_picker Hanging After Attempting to Open Unavailable Camera on iOS Simulator

chgibb:
## Description

This PR resolves the description given in the title. image_picker hangs on iOS simulator after being rejected for the camera being unavailable.

## Related Issues
[flutter/flutter #18095](https://github.com/flutter/flutter/issues/18095) describes very similar behaviour, except with respect to the gallery and video picker.
[flutter/plugins #595](https://github.com/flutter/plugins/pull/595) introduced the `FLTImagePickerPlugin#imagePickerControllerDidCancel` method, however it is never actually called anywhere in the code base. This PR simply makes use of it.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/image_picker/CHANGELOG.md
+++ b/packages/image_picker/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.6.0+16
+
+* iOS Simulator: fix hang after trying to take an image from the non-existent camera.
+
 ## 0.6.0+15
 
 * Android: throws an exception when permissions denied instead of ignoring.
--- a/packages/image_picker/ios/Classes/ImagePickerPlugin.m
+++ b/packages/image_picker/ios/Classes/ImagePickerPlugin.m
@@ -128,7 +128,9 @@ static const int SOURCE_GALLERY = 1;
                                delegate:nil
                       cancelButtonTitle:@"OK"
                       otherButtonTitles:nil] show];
-    [self imagePickerControllerDidCancel:_imagePickerController];              
+    self.result(nil);
+    self.result = nil;
+    _arguments = nil;          
   }
 }
 
--- a/packages/image_picker/ios/Classes/ImagePickerPlugin.m
+++ b/packages/image_picker/ios/Classes/ImagePickerPlugin.m
@@ -130,7 +130,7 @@ static const int SOURCE_GALLERY = 1;
                       otherButtonTitles:nil] show];
     self.result(nil);
     self.result = nil;
-    _arguments = nil;          
+    _arguments = nil;
   }
 }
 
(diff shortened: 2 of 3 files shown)


cyanglaz:
I believe the imagePickerControllerDidCancel is a delegate method in the `UIImagePickerControllerDelegate` so it should be called by the system when the image picker is cancelled.
I feel like this issue can be fixed by bringing the below code outside of the camera authorization check.
 ```objc
  if ([UIImagePickerController isSourceTypeAvailable:UIImagePickerControllerSourceTypeCamera]) {
    [_viewController presentViewController:_imagePickerController animated:YES completion:nil];
  } else {
    [[[UIAlertView alloc] initWithTitle:@"Error"
                                message:@"Camera not available."
                               delegate:nil
                      cancelButtonTitle:@"OK"
                      otherButtonTitles:nil] show];
  }
```

**Comment by chgibb** [comment]
A full search for the name returns no results except the declaration. As for factoring out the code you showed, what exactly do you propose? I am only passingly familiar with Objective-C.

**Comment by cyanglaz** [comment]
@chgibb no worries. So OBJC has a delegation pattern. Basically our ImagePickerPlugin conform to the UIImagePickerControllerDelegate protocol. One of the method in the protocol is  imagePickerControllerDidCancel. So the iOS framework is calling `[UIImagePickerControllerDelegate imagePickerControllerDidCancel]` somewhere. Delegate methods are not usually meant to be called manually.
That said, my hunch is that the issue is that we didn't check the camera availability before checking the camera authorization. We were checking the availability after authorization. See `showCamera`.
So ideally, if we move this logic that i posted outside and check the camera availability first, then check the authorization, we might be able to fix the issue.

**Comment by chgibb** [comment]
@cyanglaz I believe that fundamentally, a result is not being returned in the event that the camera is not available. The `else` branch after the check for availability displays a dialog alerting the use that the camera is not available but does not return a result back to Flutter. Adding
```
self.result(nil);
self.result = nil;
_arguments = nil;
```
immediately following the "Camera not available" dialog also appears to resolve the issue. I realize that is simply the method body of the didCancel delegate, but it was my understanding that lines `122-124` already checked for camera availability.

**Comment by cyanglaz** [comment]
Adding
```objc
self.result(nil);
self.result = nil;
_arguments = nil;
``` 
makes much more sense! Do you mind to update the PR with this change?

chgibb:
@cyanglaz absolutely! I will make the change on Monday and ping you again for review and approval. I appreciate the timely and kind discussion and feedback!

chgibb:
Made the change @cyanglaz .

cyanglaz:
@chadidi CI seems not happy about some formatting, do you mind reformat?

PR #1845, opened 2019-07-12
[webview_flutter] Basic fix for input pre N

mklim:
## Description

This patch follows the same hacks that WebView does on Android<N to set
up an InputConnection in onCreateInputConnection on the IME thread.
Previously WebView was asserting that the InputConnection code was being
run on the IME thread and then failing.

Of some extra and specific concern to this is that our webviews are
rendered inside of virtual displays. There is some extra logic in the
plugin to handle restarting the input connection and using the correct
cached connection on resize events because of this. It relies on
engine changes contained in flutter/flutter@b1a8637 (currently only on
master) in order to be called for the resize logic. When used on dev
this still brings up the keyboard when the webview isn't resized by the
keyboard pre N, but fails to fix the issue when bringing up the keyboard
would resize the webview UI.

This patch is forked from the proof of concept at amirh/plugins@64fe811,
plus refactoring and bug fixes to handle the resizing.

This still has an ongoing issue with switching between text inputs
inside of the webview UI and Flutter UI.

## Related Issues

flutter/flutter#19718

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.9+3
+
+* Add partial fix for webview keyboard input on Android versions prior to N.
+
 ## 0.3.9+2
 
 * Update Dart code to conform to current Dart formatter.
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,6 +1,11 @@
 ## 0.3.10
 
-* Add partial WebView keyboard support for Android versions prior to N.
+* Add partial WebView keyboard support for Android versions prior to N. Support
+  for UIs that also have Flutter `TextInput` fields is still pending. This basic
+  support currently only works with Flutter `master`. The keyboard will still
+  appear when it previously did not when run with older versions of Flutter. But
+  if the WebView is resized while showing the keyboard the text field will need
+  to be focused multiple times for any input to be registered.
 
 ## 0.3.9+2
 
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,6 +1,6 @@
-## 0.3.9+3
+## 0.3.10
 
-* Add partial fix for webview keyboard input on Android versions prior to N.
+* Add partial WebView keyboard support for Android versions prior to N.
 
 ## 0.3.9+2
 
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -4,19 +4,12 @@
 
 package io.flutter.plugins.webviewflutter;
 
-import static android.content.Context.INPUT_METHOD_SERVICE;
-
 import android.annotation.TargetApi;
 import android.content.Context;
 import android.os.Build;
 import android.os.Handler;
-import android.util.Log;
 import android.view.View;
-import android.view.inputmethod.EditorInfo;
-import android.view.inputmethod.InputConnection;
-import android.view.inputmethod.InputMethodManager;
 import android.webkit.WebStorage;
-import android.webkit.WebView;
 import android.webkit.WebViewClient;
 import io.flutter.plugin.common.BinaryMessenger;
 import io.flutter.plugin.common.MethodCall;
@@ -34,93 +27,6 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
   private final MethodChannel methodChannel;
   private final FlutterWebViewClient flutterWebViewClient;
   private final Handler platformThreadHandler;
-  private final View flutterView;
-
-  private static class InputAwareWebView extends WebView {
-    private static final String TAG = "InputAwareWebView";
-    private final View flutterView;
-
-    InputAwareWebView(Context context, View flutterView) {
-      super(context);
-      this.flutterView = flutterView;
-    }
-
-    private View threadedInputConnectionProxyView;
-
-    private ThreadedInputConnectionProxyAdapterView proxyAdapterView;
-
-    @Override
-    public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
-      if (proxyAdapterView == null) {
-        // No proxy adapter set, so we use the default implementation.
-        return super.onCreateInputConnection(outAttrs);
-      }
-
-      if (!proxyAdapterView.isTriggerDelayed()) {
-        // Currently on the IME thread. Delegate to the superclass.
-        return super.onCreateInputConnection(outAttrs);
-      }
-
-      InputConnection result = super.onCreateInputConnection(outAttrs);
-      if (result != null) {
-        // This should never happen.
-        Log.wtf(TAG, "Failed unexpectedly creating a webview input connection.");
-      }
-
-      final InputMethodManager imm =
-          (InputMethodManager) getContext().getSystemService(INPUT_METHOD_SERVICE);
-      proxyAdapterView.requestFocus();
-      final View containerView = this;
-
-      // This is the crucial trick that gets the InputConnection creation to happen on the correct
-      // thread. https://cs.chromium.org/chromium/src/content/public/android/java/src/org/chromium/content/browser/input/ThreadedInputConnectionFactory.java?l=169.
-      this.post(
-          new Runnable() {
-            @Override
-            public void run() {
-              // This is a hack to make InputMethodManager believe that the proxy view now has a focus.
-              // As a result, InputMethodManager will think that mProxyView is focused, and will call
-              // getHandler() of the view when creating input connection.
-
-              // Step 1: Set proxyAdapterView as InputMethodManager#mNextServedView. This does not
-              // affect the real window focus.
-              proxyAdapterView.onWindowFocusChanged(true);
-
-              // Step 2: Have InputMethodManager focus in on containerView. As a result, IMM will call
-              // onCreateInputConnection() on proxyAdapterView on the same thread as
-              // proxyAdapterView.getHandler(). It will also call subsequent InputConnection methods on
-              // this IME thread.
-              imm.isActive(containerView);
-            }
-          });
-      return null;
-    }
-
-    @Override
-    public boolean checkInputConnectionProxy(final View view) {
-      View previousProxy = threadedInputConnectionProxyView;
-      threadedInputConnectionProxyView = view;
-      if (previousProxy != view) {
-        proxyAdapterView =
-            new ThreadedInputConnectionProxyAdapterView(
-                /*containerView=*/ flutterView,
-                /*targetView=*/ view,
-                /*imeHandler=*/ view.getHandler());
-        final View container = this;
-        proxyAdapterView.requestFocus();
-        post(
-            new Runnable() {
-              @Override
-              public void run() {
-                InputMethodManager imm =
-                    (InputMethodManager) getContext().getSystemService(INPUT_METHOD_SERVICE);
-                imm.restartInput(container);
-              }
-            });
-      }
-      return super.checkInputConnectionProxy(view);
-    }
-  }
 
   @SuppressWarnings("unchecked")
   FlutterWebView(
@@ -128,9 +34,8 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
       BinaryMessenger messenger,
       int id,
       Map<String, Object> params,
-      final View flutterView) {
-    this.flutterView = flutterView;
-    webView = new InputAwareWebView(context, flutterView);
+      final View containerView) {
+    webView = new InputAwareWebView(context, containerView);
 
     platformThreadHandler = new Handler(context.getMainLooper());
     // Allow local storage.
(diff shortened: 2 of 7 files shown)


packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -24,14 +30,108 @@
...
 public class FlutterWebView implements PlatformView, MethodCallHandler {
   private static final String JS_CHANNEL_NAMES_FIELD = "javascriptChannelNames";
-  private final WebView webView;
+  private final InputAwareWebView webView;
   private final MethodChannel methodChannel;
   private final FlutterWebViewClient flutterWebViewClient;
   private final Handler platformThreadHandler;
+  private final View flutterView;
+
+  private static class InputAwareWebView extends WebView {
```
amirh: What do you think about moving this to its own file? I feel like someone coming here most likely doesn't care about this logic and it's more of a distraction.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -24,14 +30,108 @@
...
+        Log.wtf(TAG, "Failed unexpectedly creating a webview input connection.");
+      }
+
+      final InputMethodManager imm =
+          (InputMethodManager) getContext().getSystemService(INPUT_METHOD_SERVICE);
+      proxyAdapterView.requestFocus();
+      final View containerView = this;
+
+      // This is the crucial trick that gets the InputConnection creation to happen on the correct
+      // thread. https://cs.chromium.org/chromium/src/content/public/android/java/src/org/chromium/content/browser/input/ThreadedInputConnectionFactory.java?l=169.
```
amirh: nit, let's pin the link to a specific version in case it moves e.g: https://cs.chromium.org/chromium/src/content/public/android/java/src/org/chromium/content/browser/input/ThreadedInputConnectionFactory.java?l=169&rcl=f0698ee3e4483fad5b0c34159276f71cfaf81f3a

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -24,14 +30,108 @@
...
+      this.post(
+          new Runnable() {
+            @Override
+            public void run() {
+              // This is a hack to make InputMethodManager believe that the proxy view now has a focus.
```
amirh: nit: extra 'a'

amirh: I feel like we need to explain that WebView is pulling that trick, which breaks our input connection proxying, so we are pulling the same trick to make it work. If we're moving all of this to a new file it will probably fit as a general explanation there.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -24,14 +30,108 @@
...
+              // As a result, InputMethodManager will think that mProxyView is focused, and will call
+              // getHandler() of the view when creating input connection.
+
+              // Step 1: Set proxyAdapterView as InputMethodManager#mNextServedView. This does not
+              // affect the real window focus.
+              proxyAdapterView.onWindowFocusChanged(true);
+
+              // Step 2: Have InputMethodManager focus in on containerView. As a result, IMM will call
```
amirh: I'm not sure that's correct (won't it focus on mNextServedView which is the proxy view?)

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -24,14 +30,108 @@
...
+      threadedInputConnectionProxyView = view;
+      if (previousProxy != view) {
+        proxyAdapterView =
+            new ThreadedInputConnectionProxyAdapterView(
+                /*containerView=*/ flutterView,
+                /*targetView=*/ view,
+                /*imeHandler=*/ view.getHandler());
+        final View container = this;
+        proxyAdapterView.requestFocus();
+        post(
```
amirh: nit: add a comment explaining that checkInputConnectionProxy for the proxy view happened after WebView#onCreateInputConnection, so we're triggering another onCreateInputConnectionProxy to start the dance to set up a proxy view.

Actually - I wonder if we can just start the focus/isActive dance directly here without calling `imm.restartInput`.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -24,14 +30,108 @@
...
 
   @SuppressWarnings("unchecked")
-  FlutterWebView(Context context, BinaryMessenger messenger, int id, Map<String, Object> params) {
-    webView = new WebView(context);
+  FlutterWebView(
+      Context context,
+      BinaryMessenger messenger,
+      int id,
+      Map<String, Object> params,
+      final View flutterView) {
```
amirh: With the new Android embedding the FlutterView may come and go, maybe better to prepare for that.
Actually I'm not sure how can a plugin get a reference to the current FlutterView with the new embedding. @matthew-carroll do you know if that's currently supported?

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -57,6 +157,37 @@ public View getView() {
     return webView;
   }
 
+  // @Override
```
amirh: awesome that we can do that!

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -57,6 +157,37 @@ public View getView() {
...
+  // TODO(mklim): Add the @Override annotation once flutter/engine#9727 rolls to stable.
+  public void onInputConnectionUnlocked() {
+    if (webView.proxyAdapterView == null) {
+      return;
+    }
+
+    webView.proxyAdapterView.setLocked(false);
+    InputMethodManager imm =
+        (InputMethodManager) flutterView.getContext().getSystemService(INPUT_METHOD_SERVICE);
+    imm.restartInput(flutterView);
```
amirh: nit: mention why we're doing this, it might not be obvious in the future.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
matthew-carroll: The new embedding does not allow access to the FlutterView for plugins. As you mentioned, this is because FlutterEngines can move between FlutterViews, and even execute without a FlutterView. I haven't looked much at this PR, but odds are that either the things this code needs from FlutterView should be exposed as stable public APIs from the FlutterEngine, or we may want to restructure this code. 

At first blush, I'm noticing input connection management in this code, which could easily conflict with FlutterView's input connection management. We had the recent PR that seemed to suggest we may have the wrong responsibility boundaries across the embedding objects. Is this PR making an even stronger case for that? CC @xster

amirh: This can definitely conflict with Flutter's input management, though I do not see any clean way to do it. If you have a better suggestion that will be great. I'm sharing a doc with more context on why these workarounds are needed.

matthew-carroll: Even though this PR and related changes are workarounds for unusual conditions, it may be the case that we're accumulating enough of them that we should engineer the embedding to accommodate such things. I'll look forward to reading your doc to better understand the requirements and then I'll be happy to offer ideas for how we might clean up some architecture boundaries, even if the literal code remains unsightly.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/InputAwareWebView.java
```diff
@@ -0,0 +1,102 @@
...
+
+import static android.content.Context.INPUT_METHOD_SERVICE;
+
+import android.content.Context;
+import android.view.View;
+import android.view.inputmethod.InputMethodManager;
+import android.webkit.WebView;
+
+/**
+ * A WebView subclass that mirrors the same implementation hacks that the system WebView does in
```
amirh: I will be a little more specific here, mention that these hacks happen only prior to Android N, and this view is here to support these Android versions. I'd mention that the point of the workaround is to create an input connection on a dedicated thread.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/InputAwareWebView.java
```diff
@@ -0,0 +1,102 @@
...
+
+    proxyAdapterView.setLocked(false);
+
+    // Restart the input connection to avoid ViewRootImpl assuming an incorrect window state.
+    InputMethodManager imm =
+        (InputMethodManager) containerView.getContext().getSystemService(INPUT_METHOD_SERVICE);
+    imm.restartInput(containerView);
+  }
+
+  /** Creates an InputConnection from the IME thread when needed. */
```
amirh: We should explain that we're abusing the fact that the system calls this method to get a handle to the ThreadedInputConnectionProxyView (and once we get it we trigger an input connection on the webview's IME thread).

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/InputAwareWebView.java
```diff
@@ -0,0 +1,102 @@
...
+  @Override
```
amirh: We should also explain that the engine guarantees to only call this for views created by the relevant webview instance, and that we're relying/assuming that the only view created with webview that's going to interact with IMM is the ThreadedInputConnectionProxyView.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/ThreadedInputConnectionProxyAdapterView.java
```diff
@@ -0,0 +1,100 @@
...
+package io.flutter.plugins.webviewflutter;
+
+import android.os.Handler;
+import android.os.IBinder;
+import android.view.View;
+import android.view.inputmethod.EditorInfo;
+import android.view.inputmethod.InputConnection;
+
+/**
+ * A fake View only exposed to InputMethodManager.
```
amirh: We should explain the role of this view in the game (the input connection is created for it on the IMM thread and it forwards the connection creation to ThreadedInputConnectionProxyView)

packages/webview_flutter/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: webview_flutter
 description: A Flutter plugin that provides a WebView widget on Android and iOS.
-version: 0.3.9+2
+version: 0.3.9+3
```
amirh: minor:

I guess we can look at it either as a bug fix (fix webview keyboard prior to Android N), or as a feature(support webview keyboard prior to Android N) as we never supported it before.
I'm assuming people read into patch number bumps as: I can keep doing everything I did before, but some bugs I've hit are fixed and minor bumps as: there's new stuff I can do now. In this case I'd opt for a minor version bump given that people couldn't use keyboard before, and how expected this feature is. (Also rephrase the changelog update accordingly)

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/ThreadedInputConnectionProxyAdapterView.java
mklim: Done.

packages/webview_flutter/pubspec.yaml
mklim: Makes sense. Done.

packages/webview_flutter/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
-## 0.3.9+3
+## 0.3.10
 
-* Add partial fix for webview keyboard input on Android versions prior to N.
+* Add partial WebView keyboard support for Android versions prior to N.
```
amirh: Explain what's incomplete.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/InputAwareWebView.java
```diff
@@ -9,7 +9,7 @@
 
 /**
  * A WebView subclass that mirrors the same implementation hacks that the system WebView does in
- * order to correctly create an InputConnection.
+ * Android versions below N in order to correctly create an InputConnection on the IME thread.
```
amirh: nit: The IME thread is an Android WebView term, maybe we should say "...the IME thread (the WebView's dedicated input thread)".

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/InputAwareWebView.java
```diff
@@ -53,15 +53,32 @@ void unlockInputConnection() {
...
+   * system WebView in order to know whether or not the system WebView expects an InputConnection on
+   * the IME thread.
+   */
   @Override
   public boolean checkInputConnectionProxy(final View view) {
+    // Check to see if the view param is WebView's ThreadedInputConnectionProxyView.
     View previousProxy = threadedInputConnectionProxyView;
     threadedInputConnectionProxyView = view;
     if (previousProxy == view) {
+      // This isn't ThreadedInputConnectionProxyView. Ignore it.
```
amirh: It might still be the ThreadedInputConnectionProxyView, only that it's the same one that we've already set up the proxying for.

PR #1848, opened 2019-07-13
[url_launcher] add support for android headers

BugsBunnyBR:
## Description
This PR enables the developer to send headers to the Android browser.
This enables the developer to keep the app session in the browser when the same authentication header is used in the app http requests and the in the web site.

## Checklist

- [ X I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/url_launcher/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncherPlugin.java
+++ b/packages/url_launcher/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncherPlugin.java
@@ -18,13 +18,13 @@ import android.view.KeyEvent;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
-import java.util.HashMap;
-import java.util.Map;
 import io.flutter.plugin.common.MethodCall;
 import io.flutter.plugin.common.MethodChannel;
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler;
 import io.flutter.plugin.common.MethodChannel.Result;
 import io.flutter.plugin.common.PluginRegistry.Registrar;
+import java.util.HashMap;
+import java.util.Map;
 
 /** UrlLauncherPlugin */
 public class UrlLauncherPlugin implements MethodCallHandler {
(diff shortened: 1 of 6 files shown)


collinjackson:
I've updated the version to a minor (rather than patch) update because you're introducing a new optional argument.

packages/url_launcher/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncherPlugin.java
```diff
@@ -104,23 +123,37 @@ public void onCreate(Bundle savedInstanceState) {
...
+      webview.getSettings().setDomStorageEnabled(enableDomStorage);
+
       // Open new urls inside the webview itself.
       webview.setWebViewClient(
           new WebViewClient() {
+
+            @Override
+            @SuppressWarnings("deprecation")
+            public boolean shouldOverrideUrlLoading(WebView view, String url) {
+              if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
```
amirh, line 145: I'm not sure I understand why are we overriding this to call loadUrl on versions older than Lolipop?

BugsBunnyBR: You can verify why running the code in different emulators APIs, but the gist of it is that the first method is deprecated and should be used only in older versions of android, at the same time second method only exist in new API versions and they will not call the first one.

amirh: Thanks for following up!

I'm still not sure I understand - what breaks if we don't override `shouldOverrideUrlLoading` at all?

BugsBunnyBR: I don't quite remember, but I guess that is so the user can navigate using hyperlinks.
I didn't included the call to `setWebViewClient`, I just fixed the deprecation warning I was getting in Android Studio.

PR #1857, opened 2019-07-16
[webview_flutter] Don't log unknown setting key for debuggingEnabled …

amirh:
…on iOS

## Description

The debuggingEnabled setting key is not used by the iOS implementation, and was logging an error.
Updated the iOS code to explicitly ignore that key.

## Related Issues

https://github.com/flutter/flutter/issues/34620

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/ios/Classes/FlutterWebView.m
+++ b/packages/webview_flutter/ios/Classes/FlutterWebView.m
@@ -74,6 +74,8 @@
     }];
     NSDictionary<NSString*, id>* settings = args[@"settings"];
     [self applySettings:settings];
+    // TODO(amirh): return an error if apply settings failed once it's possible to do so.
+    // https://github.com/flutter/flutter/issues/36228
 
     NSString* initialUrl = args[@"initialUrl"];
     if ([initialUrl isKindOfClass:[NSString class]]) {
@@ -118,8 +120,14 @@
 }
 
 - (void)onUpdateSettings:(FlutterMethodCall*)call result:(FlutterResult)result {
-  [self applySettings:[call arguments]];
-  result(nil);
+  NSString* error = [self applySettings:[call arguments]];
+  if (error == nil) {
+    result(nil);
+    return;
+  }
+  result([FlutterError errorWithCode:@"updateSettings_failed"
+                             message:error
+                             details:nil]);
 }
 
 - (void)onLoadUrl:(FlutterMethodCall*)call result:(FlutterResult)result {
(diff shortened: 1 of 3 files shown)


packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -228,18 +234,27 @@ - (void)clearCache:(FlutterResult)result {
...
+- (NSString*)applySettings:(NSDictionary<NSString*, id>*)settings {
+  NSMutableArray<NSString*>* unknownKeys = [[NSMutableArray alloc] init];
   for (NSString* key in settings) {
     if ([key isEqualToString:@"jsMode"]) {
       NSNumber* mode = settings[key];
       [self updateJsMode:mode];
     } else if ([key isEqualToString:@"hasNavigationDelegate"]) {
       NSNumber* hasDartNavigationDelegate = settings[key];
       _navigationDelegate.hasDartNavigationDelegate = [hasDartNavigationDelegate boolValue];
+    } else if ([key isEqualToString:@"debuggingEnabled"]) {
```
cyanglaz: nits: Prefer to escape early e.g.
```objc
if ([key isEqualToString:@"debuggingEnabled"]) {
  continue;
}
```

amirh: Feels a little inconsistent to only continue here, so I guess if we do this we should add `continue` in all if cases above, in which case all the `else`s are redundant. So we'll end up with:

```objc
if ([key isEqualToString:@"jsMode"]) {
      NSNumber* mode = settings[key];
      [self updateJsMode:mode];
      continue;
    }
    if ([key isEqualToString:@"hasNavigationDelegate"]) {
      NSNumber* hasDartNavigationDelegate = settings[key];
      _navigationDelegate.hasDartNavigationDelegate = [hasDartNavigationDelegate boolValue];
      continue;
    }
    if ([key isEqualToString:@"debuggingEnabled"]) {
      // no-op debugging is always enabled on iOS.
      continue;
    }
    [unknownKeys addObject:key];
```

Alternatively we can keep it as it currently is. Let me know if you have a preference here, one thing that comes to mind is that relying on continue is more error-prone when copy pasting to add a new key(my feeling is that there's higher chance to miss a `continue` than to miss out an `else`).

cyanglaz: Sure. Let's keep it as it is!

PR #1863, opened 2019-07-16
[webview_flutter] fix typo in comment

amirh:
Didn't bump a version and update change as I haven't published the version where this typo was introduced yet.

### DIFF
--- a/packages/webview_flutter/ios/Classes/FlutterWebView.m
+++ b/packages/webview_flutter/ios/Classes/FlutterWebView.m
@@ -234,7 +234,7 @@
   }
 }
 
-// Returns nil when succesfull, or an error message when one or more keys are unknown.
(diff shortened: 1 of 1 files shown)


packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -234,7 +234,7 @@ - (void)clearCache:(FlutterResult)result {
   }
 }
 
-// Returns nil when succesfull, or an error message when one or more keys are unknown.
+// Returns nil when successfull, or an error message when one or more keys are unknown.
```
mklim:
```suggestion
// Returns nil when successful, or an error message when one or more keys are unknown.
```

amirh: 😶 thanks

PR #1313, opened 2019-03-05
[ci] Switch to macOS VM with Flutter pre-installed

fkorotkov:

PR #1864, opened 2019-07-16
Revert "[ci] Switch to macOS VM with Flutter pre-installed"

amirh:
Reverts flutter/plugins#1313

We've been started to get the following error on the Cirrus tasks:
```
./script/incremental_build.sh build-examples --ipa
fatal: No such ref: 'FETCH_HEAD'
fatal: Not a valid object name FETCH_HEAD
```

I suspect the reverted change is where it regressed.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -57,21 +57,26 @@ task:
 task:
   use_compute_credits: $CIRRUS_USER_COLLABORATOR == 'true'
   osx_instance:
-    image: mojave-xcode-10.2-flutter
+    image: mojave-xcode-10.1
+  install_cocoapods_script:
+    - sudo gem install cocoapods
   setup_script:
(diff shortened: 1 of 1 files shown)


fkorotkov:
@amirh do you have a link to the build where you saw it?

amirh:
It happens on any macOS task since this change landed, I think what happens is that `./script/incremental_build.sh` expects the current directory to be one where the plugins repo is cloned but it isn't(though let's first see if we're not getting the error with the revert).

e.g see the log for build here: https://cirrus-ci.com/task/4977285139005440

fkorotkov:
I think it's because `git fetch origin master` got removed. My bad! See #1865

collinjackson:
Closing this in favor of #1865

PR #1870, opened 2019-07-17
Cirrus should report errors on failures of incremental_build.sh

collinjackson:
It looks like we're testing whether the assignment of an environment variable succeeds instead of whether the environment variable actually is set to something.

Example of a masked failure: https://cirrus-ci.com/task/4977285139005440

See also https://github.com/flutter/plugins/pull/1864

### DIFF
--- a/script/common.sh
+++ b/script/common.sh
@@ -14,7 +14,7 @@ function check_changed_packages() {
   # We need this check because some CIs can do a single branch clones with a limited history of commits.
   local packages
   local branch_base_sha="$(get_branch_base_sha)"
-  if [[ "$?" == 0 ]]; then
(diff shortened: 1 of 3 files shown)


script/common.sh
```diff
@@ -14,7 +14,7 @@ function check_changed_packages() {
   # We need this check because some CIs can do a single branch clones with a limited history of commits.
   local packages
   local branch_base_sha="$(get_branch_base_sha)"
-  if [[ "$?" == 0 ]]; then
+  if [[ "$branch_base_sha" != "" ]]; then
```
fkorotkov: Nice! 😅👌

PR #1871, opened 2019-07-17
[firebase_auth] Add missing method getIdTokenResult()

lukaspili:
## Description

Firebase auth is missing the method `getIdTokenResult()` which is available on native SDKs. It allows to retrieve the current token + other useful info like the token custom claims.

The PR includes the changes:

- Add new public API on dart side: `Future<IdTokenResult> getIdTokenResult(bool refresh)`
- Add new object on dart side: `IdTokenResult`
- Add new method call handler on iOS, which calls the native firebase `getIdTokenResult(BOOL refresh)`
- Update the method call handler `getIdToken` on Android, in order to handle both `getIdToken()` and `getIdTokenResult()`. The reason is the firebase Android SDK only provides one method `getIdToken`, which returns the id token result. The method now returns only the token or the whole token result, depending on the dart call.

What is missing: tests, changelog, version bump.
Will add these later.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/firebase_auth/android/src/main/java/io/flutter/plugins/firebaseauth/FirebaseAuthPlugin.java
```diff
@@ -530,6 +531,9 @@ private void handleGetToken(MethodCall call, final Result result, FirebaseAuth f
       return;
     }
 
+    // Contrary to iOS, android has a single method to handle getIdToken() and getIdTokenResult()
```
collinjackson:
```suggestion
    // Contrary to iOS, Android has a single method to handle `getIdToken()` and `getIdTokenResult()`
```

collinjackson:
After looking at this more carefully, I think that getIdToken() should probably return a result object that can be used to get the token. This is a breaking API change, so I'll get some internal feedback about naming.

lukaspili:
I understand, in my opinion there should be only but 1 method. Javascript and iOS Firebase SDKs provide 2 methods, while Android SDK provides only 1.

What about deprecating `getIdToken()` (which currently returns the token string) and adding the new method `getIdTokenResult()` (which will return the result object)? 
It would allow to push without breaking change, and removing the deprecated method in the next major version.

collinjackson:
Haven't forgotten about this one, I should have a naming decision soon and then we can move forward. Thanks for your patience.

lukaspili:
No problem. I'll wait on your decision before implementing the tests.

lukaspili:
Implemented changes for renaming, updated the tests and the changelog.
PTAL.

collinjackson:
I think we should call it `authTime` rather than `authDate` to be a bit more idiomatic: this lines up with the naming conventions of the web SDK is doing, matches React Native Firebase, and to reflect that the granularity of the data includes time info. This lines up with what's happening in #1919 where I want to change `Timestamp` to `Time` in the Dart API.

lukaspili:
Makes sense. Let me know if there is anything else I can do.

lookfirst:
@collinjackson lol, I was just looking for this and couldn't find it in 0.11! I guess I'll have to try your new version asap. =)

PR #1881, opened 2019-07-19
[share] Specify explicit type for invokeMethod & const for Rect

creativecreatorormaybenot:
I was also wondering if there are plans to upgrade to Dart `2.2.2`. That would allow for putting the `sharePositionOrigin` parameters inside the map literal.

### DIFF
--- a/packages/share/CHANGELOG.md
+++ b/packages/share/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.6.2+1
 
 * Specify explicit type for `invokeMethod`.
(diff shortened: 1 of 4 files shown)


collinjackson:
This requires Flutter 1.5.0 for the `invokeMethod` change and 1.6.0 for the `const Rect` change.

creativecreatorormaybenot:
@collinjackson Right, I forgot to change it. I only payed attention to what Flutter version `stable` is at.

PR #1882, opened 2019-07-19
Update examples to adhere to new StreamBuilder API in Flutter 1.8.3+

tvolkert:


amirh:
It seems like the breaking change is potentially forcing packages to decide whether they want to support Flutter master or stable? left a comment on https://github.com/flutter/flutter/pull/34626#issuecomment-513403324

collinjackson:
Once https://github.com/flutter/flutter/pull/36577/files is landed I think this PR can be closed without merging

tvolkert:
Note the discussion in https://github.com/flutter/flutter/issues/34545#issuecomment-513446461 and below.  tldr: I still might not land https://github.com/flutter/flutter/pull/36577.  If it doesn't land, the way callers can be safe on both master and stable at the same time would be to explicitly pass `initialData: null` to the default `StreamBuilder` constructor.

collinjackson:
Marking this PR as WIP since the new `StreamBuilder` API is still under discussion on https://github.com/flutter/flutter/issues/34545, feel free to close or let us when it's ready to be reviewed.

collinjackson:
We're closing this PR as part of the migration to a new repository but feel free to re-open it there once the new `StreamBuilder` API is finalized.

PR #1885, opened 2019-07-20
Recommend 4.3.0 version of google-services in READMEs/examples

collinjackson:
Per discussion https://github.com/flutter/flutter/issues/28770#issuecomment-512519916

We can prevent these deprecation warnings by updating to a newer version of google-services gradle plugin.

```
flutter run                                                                                                     11:32:05
Launching lib/main.dart on Pixel 2 in debug mode...
Initializing gradle...                                              1.9s
Resolving dependencies...                                           5.2s
registerResGeneratingTask is deprecated, use registerGeneratedResFolders(FileCollection)
registerResGeneratingTask is deprecated, use registerGeneratedResFolders(FileCollection)
registerResGeneratingTask is deprecated, use registerGeneratedResFolders(FileCollection)
Note: /Users/deg/.pub-cache/hosted/pub.dartlang.org/firebase_core-0.4.0+6/android/src/main/java/io/flutter/plugins/firebase/core/FirebaseCorePlugin.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.                      
Running Gradle task 'assembleDebug'...                                  
Running Gradle task 'assembleDebug'... Done                        45.0s
Built build/app/outputs/apk/debug/app-debug.apk.
...
```

This change will cause warnings to be emitted to the console the first time building after upgrading an app's google-services.

```
ERROR: [TAG] Failed to resolve variable '${junit.version}'              
ERROR: [TAG] Failed to resolve variable '${animal.sniffer.version}'     
```

Subsequent builds will not have the error.

Fixes fluttere/flutter#28982

### DIFF
--- a/packages/firebase_messaging/example/pubspec.yaml
+++ b/packages/firebase_messaging/example/pubspec.yaml
@@ -2,6 +2,7 @@ name: firebase_messaging_example
 description: Demonstrates how to use the firebase_messaging plugin.
 
 dependencies:
+  firebase_analytics: any
   flutter:
     sdk: flutter
   firebase_messaging:
--- a/packages/android_alarm_manager/CHANGELOG.md
+++ b/packages/android_alarm_manager/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.1+8
+
+* Remove dependency on google-services in the Android example.
+
 ## 0.4.1+7
 
 * Fix possible crash on Android devices with APIs below 19.
--- a/packages/android_alarm_manager/example/android/build.gradle
+++ b/packages/android_alarm_manager/example/android/build.gradle
@@ -6,7 +6,6 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:3.3.0'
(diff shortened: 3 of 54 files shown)


collinjackson:
Landing this on red because the analyzer failures are a known issue (see https://github.com/flutter/plugins/pull/1882) and iOS build timeouts cannot be caused by Android-only changes.

repentsinner:
Thanks for this @collinjackson!

Is this type of thing something that `flutter doctor` should be checking for?

Coming up to speed with Flutter feels like I spend the same amount of effort hunting down Magic Boilerplate issues as I do figuring out what's wrong with my own code.

packages/firebase_messaging/example/pubspec.yaml
```diff
@@ -2,6 +2,7 @@ name: firebase_messaging_example
 description: Demonstrates how to use the firebase_messaging plugin.
 
 dependencies:
+  firebase_analytics: any
```
cremor: @collinjackson Why was this added? Is it recommended to also depend on `firebase_analytics` when you use `firebase_messaging`?

PR #1898, opened 2019-07-24
[firebase_messaging] Android: Receiving Data messages if app terminated

ivk1800:
## Description

Adding support receiving data messages if app terminated on Android.

## Related Issues

https://github.com/flutter/flutter/issues/22072

https://github.com/flutter/flutter/issues/31365
https://github.com/flutter/flutter/issues/32751

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #1900, opened 2019-07-24
[WIP][firebase_messaging]Add support for handling messages in background

kroikie:
## Description

When Flutter app is in the background or terminated allow it to handle incoming FCM messages.

## Related Issues

https://github.com/flutter/flutter/issues/32372

## Related PR

https://github.com/flutter/plugins/pull/1898

## Breaking Change

This is not a breaking change, but a version bump and change log entry will be needed.


collinjackson:
Thanks. @ivk1800 I'd love to get your thoughts on how this PR compares to #1898 since it seems like they're both tackling the same problem.

The-Redhat:
Wouldn't it be better to have per default the same behaviour like on iOS ?

kroikie:
@The-Redhat Background message handling here is optional here. Storing the messages till the app is launched would be a departure from how the native SDKs generally operate. The desired functionality is for messages (non user visible) to be handled when they are received, otherwise the app could just perform required operations at next launch with no need for messaging.

Here a background isolate is started (if enabled by the developer) to handle background messages as they arrive. I'd be happy to do this on iOS as well once I figure out background isolates for iOS.

kroikie:
@ivk1800 If you can please take a look to ensure that this PR covers the use case you solved in #1898

packages/firebase_messaging/README.md
```diff
@@ -54,7 +54,74 @@ Note: When you are debugging on Android, use a device or AVD with Google Play se
...
+    FlutterFirebaseMessagingService.setPluginRegistrant(this);
+  }
+
+  @Override
+  public void registerWith(PluginRegistry registry) {
+    GeneratedPluginRegistrant.registerWith(registry);
+  }
+}
+```
+1. Set name property of application in `AndroidManifest.xml`
```
collinjackson: This isn't showing up as an auto-numbered list (probably because of the intervening content) so you'll have to use 2, 3, etc.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
```diff
@@ -99,7 +110,28 @@ public void onReceive(Context context, Intent intent) {
 
   @Override
   public void onMethodCall(final MethodCall call, final Result result) {
-    if ("configure".equals(call.method)) {
+    if ("FcmDartService.start".equals(call.method)) {
```
collinjackson: For consistency with other plugins that use method names to represent both a class and a method (e.g. [storage](https://github.com/flutter/plugins/blob/master/packages/firebase_storage/android/src/main/java/io/flutter/plugins/firebase/storage/FirebaseStoragePlugin.java#L67)), consider using `#` here.

Also if the class you're calling is `FlutterFirebaseMessagingService` would it make sense to put that here as well?

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -18,16 +39,81 @@
...
+  private static MethodChannel sBackgroundChannel;
+
+  private static Long sBackgroundMessageHandle;
+
+  private static List<RemoteMessage> sBackgroundMessageQueue =
+      Collections.synchronizedList(new LinkedList<RemoteMessage>());
+
+  private static PluginRegistry.PluginRegistrantCallback sPluginRegistrantCallback;
+
+  private static final String TAG = "FlutterFcmService";
```
collinjackson: Nit: I think this should generally match the class you're in, i.e. `FlutterFirebaseMessagingService`

packages/firebase_messaging/lib/firebase_messaging.dart
```diff
@@ -3,13 +3,41 @@
...
+        CallbackHandle.fromRawHandle(call.arguments['handle']);
+    final Function handlerFunction =
+        PluginUtilities.getCallbackFromHandle(handle);
+    await handlerFunction(call.arguments['message']);
+    return Future<void>.value();
+  });
+
+  // Once we've finished initializing, let the native portion of the plugin
+  // know that it can start scheduling handling messages.
+  _channel.invokeMethod<void>('FcmDartService.initialized');
```
collinjackson: Should we `await` here? Can we add a unit test for this behavior?

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -18,16 +39,81 @@
   public static final String ACTION_TOKEN = "io.flutter.plugins.firebasemessaging.TOKEN";
   public static final String EXTRA_TOKEN = "token";
 
+  private static final String SHARED_PREFERENCES_KEY = "io.flutter.android_fcm_plugin";
+  private static final String BACKGROUND_SETUP_CALLBACK_HANDLE_KEY = "background_setup_callback";
+  private static final String BACKGROUND_MESSAGE_CALLBACK_HANDLE_KEY =
+      "background_message_callback";
+
+  // TODO(kroikie): make sIsIsolateRunning per-instance, not static.
```
collinjackson: Instance data would be preferable. If there's something holding you back from addressing this TODO now, maybe you could mention that in the comment, or we could address the TODO now if it's easy.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -42,4 +128,148 @@ public void onNewToken(String token) {
     intent.putExtra(EXTRA_TOKEN, token);
     LocalBroadcastManager.getInstance(this).sendBroadcast(intent);
   }
+
+  public static void startBackgroundIsolate(Context context, long callbackHandle) {
+    FlutterMain.ensureInitializationComplete(context, null);
+    String mAppBundlePath = FlutterMain.findAppBundlePath(context);
```
collinjackson: I wouldn't use `m` here since it isn't a member.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -42,4 +128,148 @@ public void onNewToken(String token) {
...
+    if (remoteMessage.getData() != null) {
+      messageData.put("data", remoteMessage.getData());
+    }
+    if (remoteMessage.getNotification() != null) {
+      messageData.put("notification", remoteMessage.getNotification());
+    }
+
+    args.put("message", messageData);
+
+    sBackgroundChannel.invokeMethod("", args, result);
```
collinjackson: It strange to use an empty method name here

packages/firebase_messaging/lib/firebase_messaging.dart
```diff
@@ -3,13 +3,41 @@
...
+void _fcmSetupCallback() {
+  const MethodChannel _channel =
+      MethodChannel(_backgroundName, JSONMethodCodec());
+
+  // Setup Flutter state needed for MethodChannels.
+  WidgetsFlutterBinding.ensureInitialized();
+
+  // This is where the magic happens and we handle background events from the
+  // native portion of the plugin.
+  _channel.setMethodCallHandler((MethodCall call) async {
```
collinjackson: It seems like it would be more readable/consistent with other plugins to read a method off the `MethodCall` here.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
```diff
@@ -41,9 +44,17 @@
   public static void registerWith(Registrar registrar) {
     final MethodChannel channel =
         new MethodChannel(registrar.messenger(), "plugins.flutter.io/firebase_messaging");
+    final MethodChannel backgroundCallbackChannel =
+        new MethodChannel(
+            registrar.messenger(),
+            "plugins.flutter.io/android_fcm_background",
+            JSONMethodCodec.INSTANCE);
```
collinjackson: Just curious, why do we use `JSONMethodCodec` here instead of the default?

packages/firebase_messaging/lib/firebase_messaging.dart
```diff
@@ -3,13 +3,41 @@
...
+import 'dart:ui';
 
 import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
 import 'package:meta/meta.dart';
 import 'package:platform/platform.dart';
 
 typedef Future<dynamic> MessageHandler(Map<String, dynamic> message);
 
+const String _backgroundName = 'plugins.flutter.io/android_fcm_background';
```
collinjackson: Not sure it makes sense to put this here as it's only used in one place, but should probably have channel in the method name

packages/firebase_messaging/lib/firebase_messaging.dart
```diff
@@ -3,13 +3,41 @@
...
+
+void _fcmSetupCallback() {
```
collinjackson: You might want to put this code in the one place it's used, or in a static method of the class, it's unusual to have it in a function at the global scope.

packages/firebase_messaging/lib/firebase_messaging.dart
```diff
@@ -3,13 +3,41 @@
...
+  const MethodChannel _channel =
```
collinjackson: An underscore for hiding is confusing here since the variable is already scoped to the current function. I would either put this on the main class (so you can make it `@visibleForTesting`) or just call it `channel` here.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -42,4 +128,148 @@ public void onNewToken(String token) {
...
+  }
+
+  // TODO(kroikie): Find a better way to determine application state.
```
collinjackson: @matthew-carroll might have ideas for other ways to do this

packages/firebase_messaging/README.md
```diff
@@ -54,7 +54,74 @@ Note: When you are debugging on Android, use a device or AVD with Google Play se
...
+```
+<application android:name=".Application" ...>
+```
+1. Define a top level method to handle background messages
```
matthew-carroll: Can you explicitly mention that this is a Dart method so that readers do not get confused?

packages/firebase_messaging/README.md
```diff
@@ -54,7 +54,74 @@ Note: When you are debugging on Android, use a device or AVD with Google Play se
...
+```
+Future<dynamic> myBackgroundMessageHandler(Map<String, dynamic> message) {
+  if (message.containsKey('data')) {
```
matthew-carroll: Can you mention somewhere where a developer can find the protocol that is being used here? The reference to 'data' and 'notification' looks magical.

packages/firebase_messaging/README.md
```diff
@@ -54,7 +54,74 @@ Note: When you are debugging on Android, use a device or AVD with Google Play se
...
+    // Handle data message
+    dynamic data = message['data'];
+  }
+
+  if (message.containsKey('notification')) {
```
matthew-carroll: Can a message be both a "data" message and a "notification" message? If not, please consider using an if-else-if statement to make that explicit.

packages/firebase_messaging/README.md
```diff
@@ -54,7 +54,74 @@ Note: When you are debugging on Android, use a device or AVD with Google Play se
...
+    // Handle notification message
+    dynamic notification = message['notification'];
+  }
+
+  // Or do work with other plugins, eg: write to RTDB.
+  FirebaseDatabase.instance.reference().child('foo').set('bar');
```
matthew-carroll: This reference to another plugin might be confusing for the reader - it seems to come out of nowhere. Also, are you sure this works as expected? You're executing in a background isolate, which will have a different version of the FirebaseDatabase plugin than the main isolate. This line might suggest to readers that there is only 1 FirebaseDatabase plugin and that they're controlling it from this background isolate.

packages/firebase_messaging/README.md
```diff
@@ -54,7 +54,74 @@ Note: When you are debugging on Android, use a device or AVD with Google Play se
...
+  return Future<void>.value();
+}
+```
+1. Set `onBackgroundMessage` handler when calling `configure`
```
matthew-carroll: Can you mention where `configure` is expected to be invoked?

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
```diff
@@ -99,7 +110,28 @@ public void onReceive(Context context, Intent intent) {
...
-    if ("configure".equals(call.method)) {
+    if ("FcmDartService.start".equals(call.method)) {
+      long setupCallbackHandle = 0;
+      long backgroundMessageHandle = 0;
+      try {
+        List<Long> callbacks = ((ArrayList<Long>) call.arguments);
+        setupCallbackHandle = callbacks.get(0);
+        backgroundMessageHandle = callbacks.get(1);
+      } catch (Exception e) {
+        Log.e(TAG, "There was an exception when getting callback handle from dart side");
```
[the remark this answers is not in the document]
matthew-carroll: Do we know why this exception might occur? Can we offer the developer any sanity checks to find out why the exception happened?

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
```diff
@@ -99,7 +110,28 @@ public void onReceive(Context context, Intent intent) {
...
+        e.printStackTrace();
+      }
+      FlutterFirebaseMessagingService.setBackgroundSetupHandle(
```
matthew-carroll: I'm concerned about all these static calls. Ideally we wouldn't have any static access. @collinjackson it would be great if we could find a better solution for this in both this plugin and the android_alarm_manager. Maybe there is an opportunity with the new embedding...

In fact, @kroikie do you know what happens if this plugin is registered with 2 different isolates?

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
matthew-carroll: I think some thorough comments in this method might be helpful. Very few developers (including framework developers) are familiar with Dart callback registration and invocation, as well as Android Service behaviors. I would recommend err'ing on the side of an abundance of comments that clarify these details...

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -18,16 +39,81 @@
...
+  private static final String SHARED_PREFERENCES_KEY = "io.flutter.android_fcm_plugin";
+  private static final String BACKGROUND_SETUP_CALLBACK_HANDLE_KEY = "background_setup_callback";
+  private static final String BACKGROUND_MESSAGE_CALLBACK_HANDLE_KEY =
+      "background_message_callback";
+
+  // TODO(kroikie): make sIsIsolateRunning per-instance, not static.
+  private static AtomicBoolean sIsIsolateRunning = new AtomicBoolean(false);
+
+  /** Background Dart execution context. */
+  private static FlutterNativeView sBackgroundFlutterView;
```
matthew-carroll: @collinjackson @kroikie I'm guessing that most of this stuff was copied from android_alarm_manager, right? If so, what do you two think about extracting the commonalities? There are enough regrettable details in the android_alarm_manager implementation that it might be a good idea to find a way to centralize it so that we can refactor and improve it one time. I'm not sure how we'd do it, or where the correct encapsulation boundary is, but I thought it worth asking.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
matthew-carroll: In the embedding I've migrated us away from hungarian notation. @collinjackson do you think we could migrate plugins away from it, too?

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -42,4 +128,148 @@ public void onNewToken(String token) {
...
+      FlutterRunArguments args = new FlutterRunArguments();
+      args.bundlePath = mAppBundlePath;
+      args.entrypoint = flutterCallback.callbackName;
+      args.libraryPath = flutterCallback.callbackLibraryPath;
+      sBackgroundFlutterView.runFromBundle(args);
+      sPluginRegistrantCallback.registerWith(sBackgroundFlutterView.getPluginRegistry());
+    }
+  }
+
+  public static void onInitialized() {
```
matthew-carroll: Please comment all public methods. And when commenting these, it is especially important to mention when they are expected to be invoked, and by whom.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
matthew-carroll: @kroikie can you clarify what you mean technically by "foreground" in this case?

matthew-carroll:
CC @xster - This PR replicates a lot of the functionality from android_alarm_manager. I think you and I knew this was eventually coming after our conversation with @kroikie, but wanted to make sure its on your radar. We've got a very subpar solution for background execution in android_alarm_manager and we're seeing that expand here. We may want to prioritize a first-class background execution solution in the add-to-app timeline.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
collinjackson: I would support having the common functionality extracted. I'm not sure if now's the time to do it, or if we should land this functionality and with a TODO to refactor plugins as a second pass. Are there other important use cases for a background Dart isolate other than handling local and remote messages? If it's just these two plugins, then perhaps putting a lot of effort into building a common API is less critical.

matthew-carroll: All mobile app background behavior will eventually become a candidate for this kind of implementation...or preferably a much better one. This could be any OS signal: push notifications, user location changes, bluetooth messages, possibly 2FA messages, etc. Any message that you can register with the OS to receive could theoretically require background processing and full-Flutter apps would probably want to execute the application logic in Dart.

I've actually been suggesting since late last year that we consider background processing in Android to be its own project, fully researching the issue, finding relevant customers, and then producing a strong architectural solution. Obviously we can't do that here, but that's the background context (pun intended).

collinjackson: It sounds like a worthwhile project, the question in my mind is whether it should block landing this PR. Are you leaning one way or another?

matthew-carroll: We definitely don't need to hold up this PR on that much larger project. But we should all be aware that this behavior, both here and in android_alarm_manager, will likely need major refactoring once we have a better solution. 

As for centralizing this behavior, I'll leave that judgement to you and @kroikie because you two have better visibility into the desired timeline for this work. If we choose not to centralize this background behavior, then we should all be sure that any fix or change to any of this behavior in one project should also be made in the other project to avoid divergence.

collinjackson: OK, sounds like we should move forward with this PR with a TODO and file an issue for centralizing the behavior.

XSiyabonga:
@kroikie  I have implemented the branch and maybe this might help developers who don't notice at first hand which I didn't. When creating the Application.java class. The package name needs to be specific to the application path of the file. If not the app breaks on launch. 

`package "your specific package name of your app"`

I hope that helps.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -42,4 +128,148 @@ public void onNewToken(String token) {
...
     LocalBroadcastManager.getInstance(this).sendBroadcast(intent);
   }
+
+  public static void startBackgroundIsolate(Context context, long callbackHandle) {
+    FlutterMain.ensureInitializationComplete(context, null);
+    String mAppBundlePath = FlutterMain.findAppBundlePath(context);
+    FlutterCallbackInformation flutterCallback =
+        FlutterCallbackInformation.lookupCallbackInformation(callbackHandle);
+    if (flutterCallback == null) {
+      Log.e(TAG, "Fatal: failed to find callback");
```
XSiyabonga: `E/FlutterFcmService: Fatal: failed to find callback`

Logcat logs this both on App in Background and App Terminated.

It could be that the shared preference of the handle call back doesn't exist. Don't know why I'm getting null on the callback.

kroikie: That wold indicate that the Dart side did not set the handle, so the lookup is failing. That is my first feeling, I'll dig deeper.

kroikie: @XSiyabonga I will add the package information about the Application class to the README, thanks.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
kroikie: @matthew-carroll If the new embedding is a better solution I'd be happy to transition to it, I am not sure it is though could you point me.

FWIU the goal is to have the plugin registered in both isolates with each isolate having its own channel for message passing.

matthew-carroll: I suppose we probably can't use the new embedding until we call it stable later in the year, so don't worry about that.

packages/firebase_messaging/lib/firebase_messaging.dart
```diff
@@ -67,6 +97,20 @@ class FirebaseMessaging {
...
+    if (_onBackgroundMessage != null) {
+      _onBackgroundMessage = onBackgroundMessage;
+      final CallbackHandle backgroundSetupHandle =
+          PluginUtilities.getCallbackHandle(_fcmSetupCallback);
+      final CallbackHandle backgroundMessageHandle =
+          PluginUtilities.getCallbackHandle(_onBackgroundMessage);
+      _channel.invokeMethod<bool>(
+        'FcmDartService.start',
+        <dynamic>[
+          backgroundSetupHandle.toRawHandle(),
```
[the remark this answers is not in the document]
kroikie: Used a <String, dynamic> map so parameters can be named.

packages/firebase_messaging/README.md
kroikie: Had to indent the intervening content.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
kroikie: Used `#` for consistency. 

`FlutterFIrebaseMessagingService` starts without the call from the Dart side. I'm happy to change the name but it should be something that indicates the setting up of the background channel.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
kroikie: Done, I seem to remember there being a string length limit on Tags for logcat but maybe that restriction has been removed.

packages/firebase_messaging/lib/firebase_messaging.dart
kroikie: Done

xster:
Responding late. We did talk about this and FWIW, doing a thing for now rather than building out some sort of dependency injection mechanism to keep the same engine instance is ok and pragmatic for now. Though it's "unsupported" and we can't guarantee what's going to happen when this creates 2 engine instances.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
kroikie: Moved to instance data.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
kroikie: Looks like 23 is still the max so reverting to `FlutterFcmService`.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
kroikie: I'll go with `appBundlePath`, I'll change if we decide otherwise.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
kroikie: This was copied from the alarm manager plugin where the background channel only has a single method call so no name was provided. I'm fine with adding a name anyway so added a method name `handleBackgroundMessage`.

packages/firebase_messaging/lib/firebase_messaging.dart
kroikie: Done

packages/firebase_messaging/lib/firebase_messaging.dart
kroikie: Done

packages/firebase_messaging/lib/firebase_messaging.dart
kroikie: Added as a static member of FirebaseMessaging

packages/firebase_messaging/lib/firebase_messaging.dart
kroikie: This variable is no longer defined in this scope.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
kroikie: @matthew-carroll the goal here is to determine when the Flutter process is still alive, in which case the original isolate can be used to process incoming messages. However when the Flutter process is no longer running the "background" isolate is used to process incoming messages.

Here we are checking if the app is in the foreground and visible, which is not exactly what the goal is but I could not find a clean way to determine if the Flutter process is still alive, this approximation was the closest I could get.

packages/firebase_messaging/README.md
kroikie: Done

packages/firebase_messaging/README.md
kroikie: Done, it is in line with [RemoteMessage](https://firebase.google.com/docs/reference/android/com/google/firebase/messaging/RemoteMessage) fields.

packages/firebase_messaging/README.md
kroikie: RemoteMessages can contain both a data and notification part.

How the Android SDK passes the message over to the developer (and Plugin) depends on the [state of the app](https://firebase.google.com/docs/cloud-messaging/android/receive#handling_messages).

packages/firebase_messaging/README.md
kroikie: This example is to suggest that the developer is able to perform work in the background. In this case writing some data to RTDB. I assume the dev could use the shared preferences plugin to store some data locally.

I have tried this and it does work as expected from my tests, is there some reason why getting an instance of RTDB (which is a singleton) in a different isolate would be an issue?

packages/firebase_messaging/README.md
kroikie: Done

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
kroikie: Added comments describing the need for `FlutterFirebaseMessagingService` and the methods `FcmDartService#start` and `FcmDartService#initialized`

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
kroikie: Agreed!

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
kroikie: Added comments.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
matthew-carroll: There should only be 1 process, and that's the app's process. If this code is executing then that process exists, by definition...

packages/firebase_messaging/README.md
matthew-carroll: If the implementation of a plugin ensures that it shares resources across all plugin instances then things are fine. My concern here is that we're suggesting that you can interact with any plugin from here, but any plugin that does not setup its own singleton behavior would not work as expected...

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
kroikie: I think this is just a artifact from the alarm manager plugin I don't think it is necessary, the default codec should be sufficient.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -129,9 +129,18 @@ public void onNewToken(String token) {
...
+   * the app receives a message while it is inactive.
+   *
+   * @param context Registrar or FirebaseMessagingService context.
+   * @param callbackHandle Handle used to retrieve the Dart function that sets up
+   *                       background handling on the dart side.
+   */
   public static void startBackgroundIsolate(Context context, long callbackHandle) {
     FlutterMain.ensureInitializationComplete(context, null);
-    String mAppBundlePath = FlutterMain.findAppBundlePath(context);
+    String appBundlePath = FlutterMain.findAppBundlePath();
```
XSiyabonga: error: method findAppBundlePath in class FlutterMain cannot be applied to given types;

` String appBundlePath = FlutterMain.findAppBundlePath();`

required: Context
  found: no arguments
  reason: actual and formal argument lists differ in length

Not sure why the context was removed on your latest commit but it creates a compile time error with the removed context

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
pafnat: Tested commit #52a19bacbeadac4d0e989091d3ad58e4da3368b1
- run example :
got E/FlutterFcmService(10271): Fatal: failed to find callback
- press "home" android button to hide
- send with firebase console message 
- notification appears but no callback with myBackgroundMessageHandler...
seems there is a bug still.. how to fix?

pafnat: > That wold indicate that the Dart side did not set the handle, so the lookup is failing. That is my first feeling, I'll dig deeper.

how to fix? added below my comment how to reproduce

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
pafnat: > error: method findAppBundlePath in class FlutterMain cannot be applied to given types;
> 
> ` String appBundlePath = FlutterMain.findAppBundlePath();`
> 
> required: Context
> found: no arguments
> reason: actual and formal argument lists differ in length
> 
> Not sure why the context was removed on your latest commit but it creates a compile time error with the removed context

how to fix above error in background? E/FlutterFcmService: Fatal: failed to find callback 
any advice?

kroikie: Hmmm, this may be a [method that is deprecated](https://github.com/flutter/engine/commit/4c6948a88af62b5c79e93073cf974e9130e84abe#diff-599e1d64442183ead768757cca6805c3R323) on the master channel. I'll revert to the previous method since that has not come to the stable channel yet.

Thanks for flagging this.

pafnat: > Hmmm, this may be a [method that is deprecated](https://github.com/flutter/engine/commit/4c6948a88af62b5c79e93073cf974e9130e84abe#diff-599e1d64442183ead768757cca6805c3R323) on the master channel. I'll revert to the previous method since that has not come to the stable channel yet.
> 
> Thanks for flagging this.

now.. toRawHandle() error with red screen after running example app...

kroikie: @pafnat where are you getting that error, do you have a stack trace?

pafnat: > @pafnat where are you getting that error, do you have a stack trace?

suppose this error on 122 or 123 lines of firebase_messaging.dart

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
```diff
@@ -41,9 +41,14 @@
   public static void registerWith(Registrar registrar) {
     final MethodChannel channel =
         new MethodChannel(registrar.messenger(), "plugins.flutter.io/firebase_messaging");
+    final MethodChannel backgroundCallbackChannel =
+        new MethodChannel(registrar.messenger(), "plugins.flutter.io/android_fcm_background");
```
athornz: Would the future iOS implementation not use this channel as well? Perhaps it should be called `plugins.flutter.io/firebase_messaging_background`?

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
XSiyabonga: > > Hmmm, this may be a [method that is deprecated](https://github.com/flutter/engine/commit/4c6948a88af62b5c79e93073cf974e9130e84abe#diff-599e1d64442183ead768757cca6805c3R323) on the master channel. I'll revert to the previous method since that has not come to the stable channel yet.
> > Thanks for flagging this.
> 
> now.. getting this:
> NoSuchMethodError: The method 'toRawHandle' was called on null.

I'm getting the same error. At least its now on the dart side. Everything on the native seems alright. Looks like somethings wrong with the incoming MessageHandler onBackgroundMessage from our custom implementation on our dart side in the fcm.configure(). Returning null on getCallbackHandle. Lines 117 and 118 on the dart. Subsequently hitting a runtime at line 123.

pafnat: > > > Hmmm, this may be a [method that is deprecated](https://github.com/flutter/engine/commit/4c6948a88af62b5c79e93073cf974e9130e84abe#diff-599e1d64442183ead768757cca6805c3R323) on the master channel. I'll revert to the previous method since that has not come to the stable channel yet.
> > > Thanks for flagging this.
> > 
> > 
> > now.. getting this:
> > NoSuchMethodError: The method 'toRawHandle' was called on null.
> 
> I'm getting the same error. At least its now on the dart side. Everything on the native seems alright. Looks like somethings wrong with the incoming MessageHandler onBackgroundMessage from our custom implementation on our dart side in the fcm.configure(). Returning null on getCallbackHandle. Lines 117 and 118 on the dart. Subsequently hitting a runtime at line 123.

@kroikie @XSiyabonga
Does it mean, that there are no working solution for this issue right now? i was going to rapidly make an app (with learning Flutter in parallel) and this issue ic critical for me. And if anyone needs make some tasks in background (when your app closed or minimized), - for example catch push notification, or listening for incoming SMS to parse.. better to choice native code for Android platform then Flutter?

pafnat: Finished checking #1898 pull request,
It seems there (with class FlutterBackgroundMessagesReceiver extends BroadcastReceiver) we are able to catch notification after we close the app.
Is there way to do the same on Dart side here?

kroikie: @XSiyabonga The README has an example of what `onBackgroundMessage` handler would look like.

@pafnat what solution are you looking for. This PR gives the ability to receive and handle FCM messages on Android when the app is in the background. Could you give more context on what is not working for you?

pafnat: > @XSiyabonga The README has an example of what `onBackgroundMessage` handler would look like.
> 
> @pafnat what solution are you looking for. This PR gives the ability to receive and handle FCM messages on Android when the app is in the background. Could you give more context on what is not working for you?

I was learning Flutter and there is an interesting task for automation my gate at home. The app should parse incoming SMS (or better parse incoming Push notification) while running always in background (or closed completely). Then the phone should make a call automatically on concrete phone number. Seems it is very easy task. I can make it natively without Flutter on Android but i want to measure how to achieve the same with Flutter and check both sources.
So, for Flutter it would be perfect if all loic would be on Dart side for example:
1. Parse incoming push from firebase (if app runnin or closed)
2. Check secret code inside push
3. If code is right, then make a call

2 and 3 i have already finished in Dart (there is a plugin exists for calling for Flutter)
But now i want to do the 1 with Dart too(without changing the logic on Java side)

kroikie: @pafnat `toRawHandle` is called only in two places. Lines 121 and 122 of `firebase_messaging.dart`. If that is being called on null then either the backgroundSetupHandle or backgroundMessageHandle is null. Since the plugin already defines `backgroundSetupHandle` that is unlikely to be null, so maybe the `backgroundMessageHandle` is null (a stack trace would confirm).

Could you verify that you are setting the `onBackgroundMessage` parameter of the `configure` method?

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FirebaseMessagingPlugin.java
kroikie: Done

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
pafnat: > `onBackgroundMessage`

sure, i have used onBackgroundMessage  as described at README. still  the same error as above.

kroikie: @pafnat, could you confirm exactly where you are getting that error. If you provide a stack trace it would help track down the problem.

pafnat: > @pafnat, could you confirm exactly where you are getting that error. If you provide a stack trace it would help track down the problem.

I will try but in 1-2 days.
Anyway i have made my task with #1898 PR. Can confirm, as temporaly solution to achieve my task.
Even if i close completely Flutter app, firebase messages are delivered and after that, app make outgoing phone call automatically. Thats what i needed.
Can you confirm that in that #1900 PR should be the same possible, or, only true in case the app is not closed by user?

PS. In Android Q you have to manually open developer options and enable start activities from bachrounf service toggle. By default it will be failed. 
And it means, that this #1900 PR will not work on Android 10....

kroikie: @pafnat, I'm glad you have a solution. This PR covers all use cases of #1898 so if that one worked this will work as well.

The FirebaseMessagingService is a special service that is managed by Google Play Services so the Q changes will not affect the ability of messages to be handled in the background.

pafnat: > @pafnat, I'm glad you have a solution. This PR covers all use cases of #1898 so if that one worked this will work as well.
> 
> The FirebaseMessagingService is a special service that is managed by Google Play Services so the Q changes will not affect the ability of messages to be handled in the background.

The problem that in Android Q dp5 it occurred. To quick fiz tou have to enable toggle in developers options to execute code from backround. 
In Android < 10, it works out of box without going tondevelopers options.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
collinjackson: @matthew-carroll I'm not sure if this is actionable feedback for Arthur, are you ok with landing this as is?

packages/firebase_messaging/lib/firebase_messaging.dart
```diff
@@ -3,13 +3,54 @@
...
 
+/// Setup method channel to handle Firebase Cloud Messages received while
+/// the Flutter app is not active. The handle for this method is generated
+/// and passed to the Android side so that the background isolate knows where
+/// to send background messages for processing.
+///
+/// Your app should never call this method directly, this is only for use
+/// by the firebase_messaging plugin to setup background message handling.
+@visibleForTesting
+Future<void> fcmSetupBackgroundChannel(
```
collinjackson: nit: I would lean towards making this a static void on `FirebaseMessaging` and removing `fcm` for stylistic consistency with other plugins. We don't usually namespace global functions with prefixes like this.

packages/firebase_messaging/android/src/main/java/io/flutter/plugins/firebasemessaging/FlutterFirebaseMessagingService.java
```diff
@@ -18,16 +39,81 @@
   public static final String ACTION_TOKEN = "io.flutter.plugins.firebasemessaging.TOKEN";
   public static final String EXTRA_TOKEN = "token";
 
+  private static final String SHARED_PREFERENCES_KEY = "io.flutter.android_fcm_plugin";
+  private static final String BACKGROUND_SETUP_CALLBACK_HANDLE_KEY = "background_setup_callback";
+  private static final String BACKGROUND_MESSAGE_CALLBACK_HANDLE_KEY =
+      "background_message_callback";
+
+  // TODO(kroikie): make sIsIsolateRunning per-instance, not static.
+  private static AtomicBoolean sIsIsolateRunning = new AtomicBoolean(false);
```
collinjackson: Per @matthew-carroll's suggestion, let's migrate away from Hungarian notation in first-party plugins. I think his feedback from earlier in this PR is being hidden due to changes but here's a [direct link](https://github.com/flutter/plugins/pull/1900/files/dd6cbaec5f62ff7840609f4e4fa471c65bfca5bf#diff-3a1f06b57cf55fdbd20863baeaadef12R134).

packages/firebase_messaging/README.md
```diff
@@ -54,7 +54,74 @@ Note: When you are debugging on Android, use a device or AVD with Google Play se
       <category android:name="android.intent.category.DEFAULT" />
   </intent-filter>
   ```
+#### Optionally handle background messages
 
+By default background messaging is not enabled. To handle messages in the background:
+
+1. Add an Application.java class to your app
```
JillyTaboga: Where in the App we should put this Java class, with the MainActivity or inside io.flutter.plugins?

brunoripa:
So, basically, this PR makes possible to handle background messages by following the info in the README ? Adding java class etc ? Would it be possible to get pointed to a complete example, so that maybe we can contribute to a documentation ?

packages/firebase_messaging/README.md
satyen893: @JillyTaboga You should add it alongside your `MainActivity.java` file.

PR #1912, opened 2019-07-26
[firebase_crashlytics] Fix Printing, README, Changelog & log FlutterErrorDetails.informationCollector

creativecreatorormaybenot:
The initial issue I was facing was that `firebase_crashlytics` would throw an unhandled exception on e.g. `RenderFlex` exceptions without a stack trace (and the stack trace formatting was unreadable + the exceptions was missing information):

```
Error caught by Crashlytics plugin:
[ERROR:flutter/lib/ui/ui_dart_state.cc(148)] Unhandled Exception: Invalid argument(s): Cannot create a Trace from null.
#0      new Trace.from (package:stack_trace/src/trace.dart:105:7)
#1      Trace.format (package:stack_trace/src/trace.dart:71:21)
#2      Crashlytics.onError (package:firebase_crashlytics/src/firebase_crashlytics.dart:37:19)
<asynchronous suspension>
```

Along the way of fixing this behavior, I noticed that the `README.md` file was outdated and did not even contain the [**breaking change** from `0.1.0`](https://pub.dev/packages/firebase_crashlytics#010). Consequently, I fixed my original issue, added what was missing to the README and also adjusted some inconsistent formatting I found in the README.

I also noticed that `Error reported to Crashlytics.` was not even being printed and added that and also made the `Result` section more clear.

## Notes

I found myself adding `print(details)` to the [old `onError` implementation](https://pub.dev/documentation/firebase_crashlytics/0.0.4+6/#use-the-plugin) because the formatting from `Trace.format` is terrible for reading it in the console and it was also missing what the actual exception is, which I also added in this PR. This is the reason I averted from using `Trace.format` in the debug implementation.  

I noticed that `context` was not being used in the native Android implementation, so I thought that I would add it as a string as it is a very short term describing when the exception was thrown.  
Additionally, I added **`information`** and logged that using [`Crashlytics.log`](https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=android) and [`CS_LOG`](https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=ios#add_custom_log_messages). For me, this is a very crucial part for working with Flutter exceptions as it gives information about the tree, the specific widget and also some background information on why the exception occurred.

Furthermore, I noticed that the tests were printing errors when trying to access `lineParts[2]`, which was distracting. Hence, I added a check to prevent that from happening.

If the PR itself is fine but there are minor issues like comments being abundant, then please just fix them to speed things up.

I hope that my comments also help to understand my changes.

## Testing

It seems like driving examples was added for both Android and iOS now. Thus, I would assume that the tests run by CI should confirm whether the changes are implemented correctly or not as I added `informationCollector` and `context` to the test call.  
The logging part I could only confirm locally by trying it out. I am not sure how I would write a test for that.

## Question

After going through the code and reading the README, I am wondering if the error is also supposed to be printed locally in either of the following two scenarios because it was not printed before my change and I did not adjust that (only clarified in the README):

 1. App is in release mode.

 1. App is in debug mode but `enableInDevMode` is `true`.

### Changelog

I fixed the changelog typo that was pointed out in #1913 by @lsaudon in here as well.

### DIFF
--- a/packages/firebase_crashlytics/CHANGELOG.md
+++ b/packages/firebase_crashlytics/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 0.1.1
 
+* Added additional exception information from the Flutter framework to the reports.
+* Reformatted debug printing of exceptions to be human-readable.
 * Updated `README.md` to include both the breaking change from `0.1.0` and the newly added
   `recordError` function in the setup section.
 * Made the `_recordError` function safe for missing stack traces (`null`).
--- a/packages/firebase_crashlytics/CHANGELOG.md
+++ b/packages/firebase_crashlytics/CHANGELOG.md
@@ -7,10 +7,11 @@
 * Made the `_recordError` function safe for missing stack traces (`null`).
 * Added the "Error reported to Crashlytics." print statement that was previously missing.
 * Adjusted `README.md` formatting.
+* Fixed `recordFlutterError` method name in the `0.1.0` changelog entry.
 
 ## 0.1.0
 
-* **Breaking Change** Renamed `onError` to `reportFlutterError`.
+* **Breaking Change** Renamed `onError` to `recordFlutterError`.
 * Added `recordError` method for errors caught using `runZoned`'s `onError`.
 
 ## 0.0.4+12
--- a/packages/firebase_crashlytics/CHANGELOG.md
+++ b/packages/firebase_crashlytics/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 0.1.1
+
+* Updated `README.md` to include both the breaking change from `0.1.0` and the newly added
+  `recordError` function in the setup section.
+* Made the `_recordError` function safe for missing stack traces (`null`).
+* Added the "Error reported to Crashlytics." print statement that was previously missing.
+* Adjusted `README.md` formatting.
+
 ## 0.1.0
 
 * **Breaking Change** Renamed `onError` to `reportFlutterError`.
--- a/packages/firebase_crashlytics/CHANGELOG.md
+++ b/packages/firebase_crashlytics/CHANGELOG.md
@@ -1,11 +1,11 @@
-## 0.1.1
+## 0.1.0+1
 
 * Added additional exception information from the Flutter framework to the reports.
-* Reformatted debug printing of exceptions to be human-readable.
+* Refactored debug printing of exceptions to be human-readable.
+* Passing `null` stack traces is now supported.
+* Added the "Error reported to Crashlytics." print statement that was previously missing.
 * Updated `README.md` to include both the breaking change from `0.1.0` and the newly added
   `recordError` function in the setup section.
-* Made the `_recordError` function safe for missing stack traces (`null`).
-* Added the "Error reported to Crashlytics." print statement that was previously missing.
 * Adjusted `README.md` formatting.
 * Fixed `recordFlutterError` method name in the `0.1.0` changelog entry.
 
--- a/packages/firebase_crashlytics/README.md
+++ b/packages/firebase_crashlytics/README.md
@@ -110,7 +110,7 @@ If an error is caught, you should see the following messages in your logs:
 flutter: Flutter error caught by Crashlytics plugin:
 // OR if you use recordError for runZoned:
 flutter: Error caught by Crashlytics plugin <recordError>:
-// Exception and stack trace in debug mode
+// Exception, context, information and stack trace in debug mode
 // OR if not in debug mode:
 flutter: Error reported to Crashlytics.
 ```
--- a/packages/firebase_crashlytics/README.md
+++ b/packages/firebase_crashlytics/README.md
@@ -11,13 +11,13 @@ For Flutter plugins for other Firebase products, see [FlutterFire.md](https://gi
 ## Usage
 
 ### Import the firebase_crashlytics plugin
-To use the firebase_crashlytics plugin, follow the [plugin installation instructions](https://pub.dartlang.org/packages/firebase_crashlytics#pub-pkg-tab-installing).
+To use the `firebase_crashlytics` plugin, follow the [plugin installation instructions](https://pub.dartlang.org/packages/firebase_crashlytics#pub-pkg-tab-installing).
 
 ### Android integration
 
-Enable the Google services by configuring the Gradle scripts as such.
+Enable the Google services by configuring the Gradle scripts as such:
 
-1. Add Fabric repository to the `[project]/android/build.gradle` file.
+1. Add the Fabric repository to the `[project]/android/build.gradle` file.
 ```
 repositories {
   google()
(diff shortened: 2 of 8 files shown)


packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -185,29 +194,60 @@ class Crashlytics {
...
+
+      // Not using Trace.format here to stick to the default stack trace format
+      // that Flutter developers are used to seeing.
+      if (stack != null) print('\n$stack');
     } else {
-      // Report error
+      // The stack trace can be null. To avoid the following exception:
+      // Invalid argument(s): Cannot create a Trace from null.
+      // To avoid that exception, we can check for null and provide an empty stack trace.
+      if (stack == null) stack = StackTrace.fromString('');
```
collinjackson:
```suggestion
      stack ??= StackTrace.fromString('');
```

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -185,29 +194,60 @@ class Crashlytics {
     return elements;
   }
 
+  /// On top of the default exception components, [information] can be passed as well.
```
collinjackson: I think these should be `//` comments because this is a private method.

packages/firebase_crashlytics/CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
+## 0.1.1
+
+* Added additional exception information from the Flutter framework to the reports.
+* Reformatted debug printing of exceptions to be human-readable.
+* Updated `README.md` to include both the breaking change from `0.1.0` and the newly added
+  `recordError` function in the setup section.
+* Made the `_recordError` function safe for missing stack traces (`null`).
```
collinjackson:
```suggestion
* Passing `null` stack traces is now supported.
```

Since `_recordError` method is an internal implementation detail I think it's best to focus on the behavior change of the user-facing methods.

packages/firebase_crashlytics/CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
+## 0.1.1
```
collinjackson:
```suggestion
## 0.1.0+1
```

Justification: This made a bunch of improvements without changing the public API of the public.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
```diff
@@ -165,6 +168,12 @@ class Crashlytics {
           'line': lineNumber,
         };
 
+        // The next section would throw an exception in some cases if there was no stop here.
```
collinjackson: It feels like we should add a small unit test for `getStackTraceElements` since it is `@visibleForTesting`

packages/firebase_crashlytics/android/src/main/java/io/flutter/plugins/firebase/crashlytics/firebasecrashlytics/FirebaseCrashlyticsPlugin.java
```diff
@@ -70,6 +70,15 @@ public void onMethodCall(MethodCall call, Result result) {
       exception.setStackTrace(elements.toArray(new StackTraceElement[elements.size()]));
 
       Crashlytics.setString("exception", (String) call.argument("exception"));
+
+      // Set context to show when the exception was thrown.
+      final String context = call.argument("context");
+      if (context != null) Crashlytics.setString("context", "thrown " + context);
```
collinjackson: I am not sure how I feel about the "thrown " bit here. Can you explain why you went with that, as opposed to just the context?

creativecreatorormaybenot: Because the string passed as a context ["should be in a form that will make sense in English when following the word 'thrown'"](https://api.flutter.dev/flutter/foundation/FlutterErrorDetails/context.html). The context is also only passed from `FlutterErrorDetails` and not from `runZoned`. I am not sure if every user would understand `"context": "while obtaining the image from the network"`. Everyone would probably figure it out, but adding "thrown" as a prefix seems to make it understandable more easily.

packages/firebase_crashlytics/lib/src/firebase_crashlytics.dart
creativecreatorormaybenot: Do you mean for the `continue` part I added? I did it to clean up the tests.  
Otherwise, there are [three tests for the method](https://github.com/flutter/plugins/blob/7c0a3bf650e7c09a18469a43bb6f59ef186041e5/packages/firebase_crashlytics/test/firebase_crashlytics_test.dart#L147).

collinjackson: Got it. If that path is being exercised in tests and not in production then I think it's tested and we're fine.

packages/firebase_crashlytics/android/src/main/java/io/flutter/plugins/firebase/crashlytics/firebasecrashlytics/FirebaseCrashlyticsPlugin.java
collinjackson: Makes sense, but should we update iOS to match?

collinjackson:
Thank you so much for the fast turnaround on implementing my suggestions.

packages/firebase_crashlytics/android/src/main/java/io/flutter/plugins/firebase/crashlytics/firebasecrashlytics/FirebaseCrashlyticsPlugin.java
creativecreatorormaybenot: Yes, I think that `"reason": "thrown while obtaining the image from the network"` sounds good. However, I am not exactly sure how it is displayed in Crashlytics. It might make sense to name the key `reason` on Android as well, as in `setString("reason", "thrown..`. What do you think?

collinjackson: I'm fine with reason for now. Thanks.

creativecreatorormaybenot:
Thanks for taking the time.  
I added another comment explaining the current role of `context` and why the `reason` decision was made for future contributors to understand.

packages/firebase_crashlytics/README.md
```diff
@@ -85,20 +86,32 @@ void main() {
   // development.
   Crashlytics.instance.enableInDevMode = true;
 
-  // Pass all uncaught errors to Crashlytics.
-  FlutterError.onError = (FlutterErrorDetails details) {
-    Crashlytics.instance.onError(details);
-  };
+  // Pass all uncaught errors from the framework to Crashlytics.
+  FlutterError.onError = Crashlytics.instance.recordFlutterError;
+
```
collinjackson:
```suggestion
```

PR #1907, opened 2019-07-25
[cloud_firestore] Update autoId generation to match native implementation more closely

KevinTheGray:
## Description

Currently the flutter implementation of Firestore generates a new document ID in the same way it is geenrated for the RTDB.  However, this behavior has been modified for Firestore, and it no longer generates IDs in the same way.  I was motivated to make this change because:

1. The plugins should behave as close as possible to the native implementation.  It was very obvious to me when I saw the keys that were being generated that they were not mimicking expected behavior.

2.  It's probably negligible, but the new code is a lot less complex and under less constraints.  

## Related Issues

https://github.com/flutter/flutter/issues/15715

https://github.com/flutter/flutter/issues/15715 is resolved because it now matches how the Android/iOS generate document IDs.  That is, from the pool of characters from String "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789", we choose 20 random ones and generate an ID.  

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[...]

PR #1915, opened 2019-07-26
[cloud_firestore] Add metadata to QuerySnapshot

kroikie:
## Description

Bring `cloud_firestore` plugin in line with native SDKs by exposing the metadata of QuerySnapshot.

## Related Issues

Fixes: https://github.com/flutter/flutter/issues/35550

PR #1916, opened 2019-07-26
Bump cloud_firestore version for release

collinjackson:
Left out of https://github.com/flutter/plugins/pull/1907

Adding tests to prevent this in the future: https://github.com/flutter/plugin_tools/pull/49


kroikie:
Actually it may be better to keep the versions/commits/changelog in sync so lets land this and I'll update #1915

PR #968, opened 2018-12-07
[webview_flutter]Support UserAgent setting for webview_flutter

YukiOya:
It possible to get and set custom user agent to webview.

PR #1920, opened 2019-07-27
[webview_flutter] get/set a custom User Agent

LuisThein:
## Description

Based on this pull request https://github.com/flutter/plugins/pull/968 that was closed, I added the functionality to set a custom User Agent and get the current User Agent in the current WebView version.

## Related Issues

https://github.com/flutter/flutter/issues/28256

## Checklist

- [x] I signed the [CLA].

## Breaking Change


LuisThein:
Updated emails CLA.

tp:
Thank you @LuisThein, we're already making use of this as we had the same requirement to change the user agent :)

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -255,6 +256,18 @@ class WebView extends StatefulWidget {
   /// By default `debuggingEnabled` is false.
   final bool debuggingEnabled;
 
+  /// The User_Agent used for WebView requests.
+  ///
+  /// By default `userAgent` is null.
+  /// If the `userAgent` is null, the default User Agent from the Android / iOS WebView is used.
```
amirh: nit: When null the platform's webview default is used for the `User-Agent` header.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -255,6 +256,18 @@ class WebView extends StatefulWidget {
   /// By default `debuggingEnabled` is false.
   final bool debuggingEnabled;
 
+  /// The User_Agent used for WebView requests.
+  ///
+  /// By default `userAgent` is null.
```
amirh: nit: move this to the bottom on its own paragraph.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -255,6 +256,18 @@ class WebView extends StatefulWidget {
...
+  /// If the `userAgent` is null, the default User Agent from the Android / iOS WebView is used.
+  /// WebView User Agent for Android (https://developer.chrome.com/multidevice/user-agent)
+  ///
+  /// When the WebView is rebuild with a new User Agent, the page reloads and the request uses the new User Agent.
```
amirh: s/User Agent/user agent/

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -255,6 +256,18 @@ class WebView extends StatefulWidget {
   /// By default `debuggingEnabled` is false.
   final bool debuggingEnabled;
 
+  /// The User_Agent used for WebView requests.
+  ///
+  /// By default `userAgent` is null.
+  /// If the `userAgent` is null, the default User Agent from the Android / iOS WebView is used.
+  /// WebView User Agent for Android (https://developer.chrome.com/multidevice/user-agent)
```
amirh: s/WebView/[WebView]/

packages/webview_flutter/lib/webview_flutter.dart
amirh: s/rebuild/rebuilt/

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -255,6 +256,18 @@ class WebView extends StatefulWidget {
...
+  ///
+  /// When the WebView is rebuild with a new User Agent, the page reloads and the request uses the new User Agent.
+  ///
+  /// When [WebViewController.goBack] is called after changing the User Agent, the old User Agent is used.
```
amirh: Suggestion:

When [WebViewController.goBack] is called after changing `userAgent` the previous `userAgent` value is used until the page is reloaded.

packages/webview_flutter/lib/src/webview_method_channel.dart
```diff
@@ -122,6 +127,7 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
     _addIfNonNull('jsMode', settings.javascriptMode?.index);
     _addIfNonNull('hasNavigationDelegate', settings.hasNavigationDelegate);
     _addIfNonNull('debuggingEnabled', settings.debuggingEnabled);
+    _addIfNonNull('userAgent', settings.userAgent);
```
amirh: What happens if we rebuild a WebView that had an explicitly set `userAgent` and set the `userAgent` to null. IIUC the way this currently propagates will lead to preserving the previous user agent. I'd expect it to go back to the platform's default user agent.

We should add an test for this scenario as well (build with an explicitly set user agent, rebuild with user agent set to null)

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -152,6 +152,11 @@ abstract class WebViewPlatformController {
     throw UnimplementedError(
         "WebView removeJavascriptChannels is not implemented on the current platform");
   }
+
+  Future<String> getUserAgent() {
```
amirh: Add Dartdoc

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -176,16 +182,24 @@ class WebSettings {
   /// See also: [WebView.debuggingEnabled].
   final bool debuggingEnabled;
 
+  /// The User Agent used for all following requests
```
amirh: See Dartdoc comments on the userAgent field in WebView

packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -321,4 +326,29 @@ - (void)registerJavaScriptChannels:(NSSet*)channelNames
...
+           result(userAgent);
+         }
+       }];
+}
+
+- (void)updateUserAgent:(NSString*)userAgent {
+  if (@available(iOS 9.0, *)) {
+    [_webView setCustomUserAgent:userAgent];
+  } else {
+    NSLog(@"Updating UserAgent is not supported for Flutter WebViews prior to iOS 9.");
```
amirh: Mention that in the Dartdoc for WebView.userAgent.

packages/webview_flutter/lib/src/webview_method_channel.dart
LuisThein: Makes sense. I changed it so that the user agent can be set to null and the default platform user agent is used after a rebuild. Additionally I adapted the same logic for the FakePlatformWebView unit tests.

LuisThein:
Added requested changes and updated branch.

packages/webview_flutter/lib/src/webview_method_channel.dart
amirh: Thanks!
I think we should avoid updating the user agent each time the widget is rebuilt.
It seems like we're lacking the ability to express presence/absence of a setting that may be null. One way to do it would be to use some kind of an optional type.

I pushed https://github.com/flutter/plugins/pull/1920/commits/79761515b33c3e96234e7b0555651c74f971adee with a proposal on how to do that, let me know whether that makes sense to you.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -64,6 +64,11 @@
       String url = (String) params.get("initialUrl");
       webView.loadUrl(url);
     }
+
+    if (params.containsKey("userAgent")) {
```
amirh: This should be done before `loadUrl` for the initial URL.

amirh: pushed fix

amirh:
@LuisThein I pushed a few changes to your branch.
I'm ok with the current state of this PR.
Let me know whether the updates I pushed makes sense to you as well.

packages/webview_flutter/lib/src/webview_method_channel.dart
LuisThein: I think that is a nice way to handle nullable settings and I agree that updating the user agent every rebuild is unnecessary.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
LuisThein: Makes sense, thanks for fixing it.

packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -347,4 +352,29 @@ - (void)registerJavaScriptChannels:(NSSet*)channelNames
   }
 }
 
+- (void)onGetUserAgent:(FlutterMethodCall*)call result:(FlutterResult)result {
+  [_webView
+      evaluateJavaScript:@"navigator.userAgent"
```
amirh: I just realized we can't rely on JavaScript here as it may be disabled.
Sorry for missing this in previous iterations.

I guess if WkWebView doesn't support getting the user agent we should probably let our users inject js to get it as they are the ones in control of the js policy.

My suggestion for this PR would be to not include the `getUserAgent` method, and inject js to get the user agent in the e2e tests.

We should file an issue for `getUserAgent` if that turns out to be important to many users we should consider adding it as an Android only API (maybe as part of the Android controller).

LuisThein: I see, it makes sense that we shouldn't rely on injected javascript to get the user agent in the WkWebView. As the main goal of this pr was to add the ability to set the user agent, I don't mind removing the `getUserAgent` method. I removed all methods involved in getting the user agent and updated the tests accordingly.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -72,21 +76,29 @@ public View getView() {
   }
 
   // @Override
-  // This is overriding a method that hasn't rolled into stable Flutter yet. Including the
-  // annotation would cause compile time failures in versions of Flutter too old to include the new
-  // method. However leaving it raw like this means that the method will be ignored in old versions
+  // This is overriding a method that hasn't rolled into stable Flutter yet.
```
amirh: Please revert unrelated re-formatting

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -72,21 +76,29 @@ public View getView() {
...
+  // stable.
   public void onInputConnectionUnlocked() {
     webView.unlockInputConnection();
   }
 
   // @Override
-  // This is overriding a method that hasn't rolled into stable Flutter yet. Including the
-  // annotation would cause compile time failures in versions of Flutter too old to include the new
-  // method. However leaving it raw like this means that the method will be ignored in old versions
+  // This is overriding a method that hasn't rolled into stable Flutter yet.
```
amirh: Also here

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -261,7 +276,8 @@ private void updateJsMode(int mode) {
   }
 
   private void updateAutoMediaPlaybackPolicy(int mode) {
-    // This is the index of the AutoMediaPlaybackPolicy enum, index 1 is always_allow, for all
+    // This is the index of the AutoMediaPlaybackPolicy enum, index 1 is
```
amirh: Same here.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
LuisThein: Oops, should be fixed now.

PR #1928, opened 2019-07-29
[firebase_auth] Fixes for iOS auth crash on failed sign-in

collinjackson:
Adds integration testing

Fixes flutter/flutter#37160

### DIFF
--- a/packages/firebase_auth/example/test/firebase_auth.dart
+++ b/packages/firebase_auth/example/test/firebase_auth.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:math';
 
 import 'package:firebase_auth/firebase_auth.dart';
+import 'package:flutter/services.dart';
 import 'package:flutter_driver/driver_extension.dart';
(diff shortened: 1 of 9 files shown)


packages/firebase_auth/example/test/firebase_auth.dart
```diff
@@ -46,6 +50,37 @@ void main() {
           user2.metadata.lastSignInTime, equals(user2.metadata.creationTime));
     });
 
+    test('email auth', () async {
+      final Random random = Random();
+      final String testEmail = 'testuser${random.nextInt(100000000)}@gmail.com';
```
kroikie: Is there no concern of a clash here which may make the test flaky?

collinjackson: It's unlikely since the user is deleted at the end of the test.

I am reluctant to roll my own random string generator since there's already a package for that: https://pub.dev/packages/random_string

Switched to using uuid for now.

PR #1930, opened 2019-07-29
[camera] Implement SupportAndroidConfigurator and CameraController

bparrishMines:
## Description

This implements the `SupportAndroidConfigurator` and adds the `supportAndroid` api to `CameraController`.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/lib/new/src/camera_controller.dart
+++ b/packages/camera/lib/new/src/camera_controller.dart
@@ -4,9 +4,13 @@
 
 import 'dart:async';
 
+import 'package:device_info/device_info.dart';
 import 'package:flutter/foundation.dart';
 
 import 'common/camera_interface.dart';
+import 'support_android/camera.dart';
+import 'support_android/camera_info.dart';
+import 'support_android/support_android_configurator.dart';
 
 /// Controls a device camera.
 ///
@@ -88,6 +92,22 @@ class CameraController {
   ///
   /// This will choose the best [CameraAPI] for the current device.
   static Future<List<CameraDescription>> availableCameras() async {
+    final List<CameraDescription> devices = <CameraDescription>[];
+
+    final DeviceInfoPlugin infoPlugin = DeviceInfoPlugin();
+    if (defaultTargetPlatform == TargetPlatform.android) {
+      final AndroidDeviceInfo info = await infoPlugin.androidInfo;
+      if (info.version.sdkInt < 21) {
+        final int numCameras = await Camera.getNumberOfCameras();
+
+        for (int i = 0; i < numCameras; i++) {
+          devices.add(await Camera.getCameraInfo(i));
+        }
+
+        return devices;
+      }
+    }
+
     throw UnimplementedError('$defaultTargetPlatform not supported');
   }
 
@@ -150,22 +170,21 @@ class CameraController {
       case CameraApi.iOS:
         throw UnimplementedError();
       case CameraApi.supportAndroid:
-        throw UnimplementedError();
+        return SupportAndroidConfigurator(description);
     }
 
     return null; // Unreachable code
   }
 
   static CameraApi _getCameraApi(CameraDescription description) {
-    return CameraApi.iOS;
+    if (description is CameraInfo) {
+      return CameraApi.supportAndroid;
+    }
 
-    // TODO(bparrishMines): Uncomment this when platform specific code is added.
-    /*
     throw ArgumentError.value(
       description.runtimeType,
       'description.runtimeType',
       'Failed to get $CameraApi from',
     );
-    */
   }
 }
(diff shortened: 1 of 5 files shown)


packages/camera/pubspec.yaml
```diff
@@ -16,6 +16,7 @@ homepage: https://github.com/flutter/plugins/tree/master/packages/camera
 dependencies:
   flutter:
     sdk: flutter
+  device_info: ^0.4.0+2
```
bparrishMines: I'm not sure if this is the best way to include this dependency. If someone uses a different version of `device_info` and `camera`, this might break. However, I think the `all_plugins_apk` test should prevent them from being incompatible thought.

packages/camera/lib/new/src/support_android/support_android_configurator.dart
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @override
+  Future<void> stop() {
+    assert(!_isDisposed);
+    return _camera.stopPreview();
+  }
+
+  @override
+  // ignore: missing_return
```
bparrishMines: This isn't asynchronous for SupportAndroid api, so i had to add this or `return null;` at the end.

packages/camera/lib/new/src/camera_controller.dart
```diff
@@ -88,6 +92,22 @@ class CameraController {
   ///
   /// This will choose the best [CameraAPI] for the current device.
   static Future<List<CameraDescription>> availableCameras() async {
+    final List<CameraDescription> devices = <CameraDescription>[];
+
+    final DeviceInfoPlugin infoPlugin = DeviceInfoPlugin();
+    if (defaultTargetPlatform == TargetPlatform.android) {
```
mklim: subjective, optional: Consider using guard clauses and early returns to reduce nesting here.

https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

```dart
if (defaultTargetPlatform != TargetPlatform.android || await infoPlugin.androidInfo >= 21) {
   throw Unimplemented Error;
}

final int numCameras = await Camera.getNumberOfCameras();
// etc
```

packages/camera/lib/new/src/support_android/support_android_configurator.dart
```diff
@@ -0,0 +1,84 @@
...
+  Camera _camera;
+  bool _isDisposed = false;
+
+  final CameraInfo info;
+
+  Camera get camera => _camera;
+
+  @override
+  Future<int> addPreviewTexture() {
+    assert(!_isDisposed);
```
mklim: nit: I think a human readable message with the assert here would be helpful.

packages/camera/lib/new/src/support_android/support_android_configurator.dart
```diff
@@ -0,0 +1,84 @@
...
+  NativeTexture _texture;
+  Camera _camera;
+  bool _isDisposed = false;
+
+  final CameraInfo info;
+
+  Camera get camera => _camera;
+
+  @override
+  Future<int> addPreviewTexture() {
```
mklim: So just making sure I understand this: it creates a preview texture if one doesn't exist, but if one does it exist it always returns the existing one?

I think that behavior is reasonable but it's not what I would have guessed from the method name. `getOrCreatePreviewTexture` maybe? Would be good for the docs on the interface to mention the behavior too.

packages/camera/lib/new/src/camera_controller.dart
bparrishMines: I wrote it like this since it will eventually expand to:

```dart
  final AndroidDeviceInfo info = await infoPlugin.androidInfo;
  if (info.version.sdkInt < 21) {
    // Camera Implementation
  } else {
    // Camera2 Implementation
  }
} else if (defaultTargetPlatform == TargetPlatform.iOS) {
  // ios implementation
}
```

Is there another way to prevent the nesting without making another method?

packages/camera/lib/new/src/support_android/support_android_configurator.dart
bparrishMines: Added to interface documentation here: https://github.com/flutter/plugins/pull/1930/commits/08ef2931ac4b9bb269ff4f1c5c0bba3a9273ed70

packages/camera/lib/new/src/camera_controller.dart
mklim: Ah okay. Yes, actually! You can flatten nested ifs like that by calculating the final logical cases first, and then casing into each of them. [Explanation](https://softwareengineering.stackexchange.com/a/205814).

```dart
final bool isAndroid = defaultTargetPlatform == TargetPlatform.android;
final bool isAndroidSdkBelow21 = isAndroid && (await infoPlugin.androidInfo).version.sdkInt < 21;
final bool isIos = !isAndroid && defaultTargetPlatform == TargetPlatform.ios;

if (isAndroidSdkBelow21) {
  // Camera
} else if (isAndroid && !isAndroidSdkBelow21) {
  // Camera 2
} else if (isIos) {
  // ios
} else {
  // throw
}
```

I think smaller methods here would also be alright. I am personally very averse to nesting. This is just a subjective readability suggestion, feel free to ignore it.

bparrishMines:
Closing since I'm not working on it for the time being.

PR #1932, opened 2019-07-30
[firebase_auth] add custom claims to example app

long1eu:
## Description

Adds custom claims example in example app. Also it uses the claims to for the IdTokenResult fields.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/firebase_auth/lib/src/firebase_user.dart
```diff
@@ -40,7 +40,14 @@ class FirebaseUser extends UserInfo {
       'app': _app.name,
     });
 
-    return IdTokenResult(data, _app);
+    if (data != null) {
```
collinjackson: Is there a way to get the `data` to be `null` for integration testing? If not, I think a unit test of the changed behavior would be fine.

packages/firebase_auth/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.14.0+2
+
+* Update example app to show token custom claims
+* Rely on custom claims to get the field from the IdTokenResult
```
collinjackson: The CHANGELOG entry is kind of vague; it would be helpful to be as specific as possible about user-facing differences in behavior that might be expected from this change.

long1eu:
The approach take here is actually the approach taken by the SDK itself.  Why pass the data twice since we already have it there? Bellow you can see the Android implementation.

```
@PublicApi
public class GetTokenResult {
  private String zza;
  private Map<String, Object> zzb;

  @KeepForSdk
  public GetTokenResult(String var1, Map<String, Object> var2) {
    this.zza = var1;
    this.zzb = var2;
  }

  @Nullable
  @PublicApi
  public String getToken() {
    return this.zza;
  }

  @PublicApi
  public long getExpirationTimestamp() {
    return this.zza("exp");
  }

  @PublicApi
  public long getAuthTimestamp() {
    return this.zza("auth_time");
  }

  @PublicApi
  public long getIssuedAtTimestamp() {
    return this.zza("iat");
  }

  @Nullable
  @PublicApi
  public String getSignInProvider() {
    Map var1;
    return (var1 = (Map)this.zzb.get("firebase")) != null ? (String)var1.get("sign_in_provider") : null;
  }

  @PublicApi
  public Map<String, Object> getClaims() {
    return this.zzb;
  }

  private final long zza(String var1) {
    Integer var2;
    return (var2 = (Integer)this.zzb.get(var1)) == null ? 0L : var2.longValue();
  }
}
```

packages/firebase_auth/lib/src/firebase_user.dart
long1eu: You are right. I don't think there is a way to get null if the Task completed successful. The only reason the IDE suggests a null check here is because of the Task API witch allow for the result of a task to be null. But does not apply in this case.

cyanglaz:
flutterfire has been re-homed to https://github.com/FirebaseExtended/FlutterFire
Please open a PR in https://github.com/FirebaseExtended/FlutterFire
I'm closing this PR.

PR #1878, opened 2019-07-19
[Image_Picker] fixed image orientation problem on Android

BugsBunnyBR:
## Description
This PR fixes (or tries to fix) orientation problems in the displayed images.
It uses the Exif information to normalize the image rotation so the image is always displayed correctly.
It will not solve all cases, as there are images without the correct Exif information, but will solve cases where the picture was taken with the phone up side down.

## Related Issues
https://github.com/flutter/flutter/issues/35334
Solving in the framework would be the best solution, but I think it may take some time to land if we wait the framework.

## Breaking change
 It will always create a new image file on Android. The current implementation sometimes creates an image file, that the developer already need to clear up after using the image if a resize was required. With the orientation fix we need a new file every time.

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [X] All existing and new tests are passing.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [X] I read and followed the [Flutter Style Guide].
- [X] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [X] I updated CHANGELOG.md to add a description of the change.
- [X] I signed the [CLA].
- [X] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [X] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [ ] No, this is *not* a breaking change.

PR #1934, opened 2019-07-31
[image_picker] Depend on AndroidX Core instead of Legacy Support v4

ChristianBecker:
## Description

It's not recommended to depend on the `legacy-support-v4` artifact any longer as we should depend on the relevant sub-artifacts instead to reduce the transitive dependencies.

The `image_picker` package only uses classes from the `androidx.core` package, so we can safely replace the `androidx.legacy:legacy-support-v4` dependency with `androidx.core:core`.

## Related Issues

This simple change is part of the PR #1878.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/image_picker/android/build.gradle
+++ b/packages/image_picker/android/build.gradle
@@ -48,7 +48,7 @@ android {
         disable 'InvalidPackage'
     }
     dependencies {
-        api 'androidx.core:core:1.0.2'
+        implementation 'androidx.core:core:1.0.2'
         implementation 'androidx.annotation:annotation:1.0.0'
     }
 }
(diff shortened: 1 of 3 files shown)


packages/image_picker/android/build.gradle
```diff
@@ -48,7 +48,7 @@ android {
         disable 'InvalidPackage'
     }
     dependencies {
-        api 'androidx.legacy:legacy-support-v4:1.0.0'
+        api 'androidx.core:core:1.0.2'
```
mklim: Good catch, thanks!

Do you mind also changing this to an `implementation` instead of an `api` dependency? That way we also wouldn't be exposing the dep here like we are now.

```suggestion
        implementation 'androidx.core:core:1.0.2'
```

PR #1937, opened 2019-08-01
[google_sign_in] Fixes PlataformException leaking in debug mode (#37343)

feinstein:
## Description

Prevents `PlatformException` from leaking in debug mode, as Dart is not able to support an exception thrown in a future because the event loop that handles the future is not able to see the caller's stack trace.

## Related Issues

flutter/flutter#37343
flutter/flutter#26705
dart-lang/sdk#37268

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/google_sign_in/CHANGELOG.md
+++ b/packages/google_sign_in/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 4.0.5+1
+## 4.0.6
 
 * Fixed the `PlatformException` leaking from `catchError()` in debug mode.
 
--- a/packages/google_sign_in/CHANGELOG.md
+++ b/packages/google_sign_in/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.0.5+1
+
+* Fixed the `PlatformException` leaking from `catchError()` in debug mode.
+
 ## 4.0.5
 
 * Update README with solution to `APIException` errors.
@@ -6,7 +10,7 @@
 
 * Revert changes in 4.0.3.
 
-## 4.0.3	
+## 4.0.3
 
 * Update guava to `27.0.1-android`.	
 * Add correct @NonNull annotations to reduce compiler warnings.	
--- a/packages/google_sign_in/lib/google_sign_in.dart
+++ b/packages/google_sign_in/lib/google_sign_in.dart
@@ -342,9 +342,13 @@ class _MethodCompleter {
   final Completer<GoogleSignInAccount> _completer =
       Completer<GoogleSignInAccount>();
 
-  void complete(FutureOr<GoogleSignInAccount> value) {
+  Future<void> complete(FutureOr<GoogleSignInAccount> value) async {
     if (value is Future<GoogleSignInAccount>) {
-      value.then(_completer.complete, onError: _completer.completeError);
+      try {
+        _completer.complete(await value);
+      } catch (e, stacktrace) {
+        _completer.completeError(e, stacktrace);
+      }
     } else {
       _completer.complete(value);
     }
--- a/packages/google_sign_in/pubspec.yaml
+++ b/packages/google_sign_in/pubspec.yaml
@@ -3,7 +3,7 @@ description: Flutter plugin for Google Sign-In, a secure authentication system
   for signing in with a Google account on Android and iOS.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/google_sign_in
-version: 4.0.5
+version: 4.0.5+1
 
 flutter:
   plugin:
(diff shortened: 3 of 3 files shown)


packages/google_sign_in/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.0.5+1
```
cyanglaz: version should be 4.0.6

feinstein: I was in doubt because of this:

> going to 0.1.2+1 indicates a change that doesn’t affect the public API.

Since my modification doesn't change the public API, shouldn't it be a `+1`?

feinstein:
> 
> 
> Thank you for the contribution! The code looks good!
> Do you mind add a test for this. It should be easily testable.

I was thinking about how to add a test for this and I am not sure how to make it, since the `PlatformException` only leaks in debug mode and when it leaks all `try-catch` statements can't catch it, how can I simulate this? How can I simulate debug mode and how can I see if an Exception leaked if I can't catch it?

packages/google_sign_in/CHANGELOG.md
cyanglaz: that is true for pre 1.0.0, after 1.0.0, in a x.y.z version, x is the major, y is the minor and z is the patch.
Before 1.0.0, it becomes 0.x.y+z

cyanglaz:
> I was thinking about how to add a test for this and I am not sure how to make it, since the `PlatformException` only leaks in debug mode and when it leaks all `try-catch` statements can't catch it, how can I simulate this? How can I simulate debug mode and how can I see if an Exception leaked if I can't catch it?

I haven't tried, but I think you can test it by creating a mock platform class to handle the method channel.  Some examples can be found here: https://github.com/flutter/plugins/blob/374a35a0a10e30e340fac0821f74d73c8fdfe5f4/packages/in_app_purchase/test/in_app_purchase_connection/app_store_connection_test.dart#L272-L436

And in your mock platform class, just throw an exception.

Now in your test case, you can do something like
```dart
bool exceptionCaught;
try {
  googleSignIn.signIn();
} catch(e) {
  exceptionCaught = true;
}
expect(exceptionCaught, true);
```

Without your change, exceptionCaught should never be assigned to true.

packages/google_sign_in/CHANGELOG.md
feinstein: Ohh... Then I think the `+` should be better explained in that documentation, as I see a lot of post-1.0.0 packages with `+`, like [`provider 3.0.0+1`](https://pub.dev/packages/provider). 

What's the meaning of `+1` after 1.0.0 then?

cyanglaz: There shouldn't be a +1 after 1.0.0. provider 3.0.0+1 is probably a mistake.

feinstein:
The google_sign_in_test.dart has some `setUp()` that are already handling the method channel, can't I just use something like them?

cyanglaz:
@feinstein Yes, if they helps. Also feel free to edit the existing handler or create another if they don't suffice your need.

feinstein:
There's a test already there which might be the same code as you suggested:

https://github.com/flutter/plugins/blob/374a35a0a10e30e340fac0821f74d73c8fdfe5f4/packages/google_sign_in/test/google_sign_in_test.dart#L268-L273

Wouldn't this test do the same? It failed to catch this problem.

cyanglaz:
there is a suppressError  parameter in `signInSilently()` that you might need to set to false to make it throw.

feinstein:
It was throwing before I fixed it, the example app starts with a `signInSilently()`, which immidiately throws into the IDE, when I fixed it, it stopped throwing. It makes sense, as the VM is "ignoring" all `catch` blocks, and `signInSilently()` ignores the exceptions by catching them.

The whole issue was that the VM was ignoring `try-catch' blocks, so the configuration inside `signInSilently()` shouldn't matter.

cyanglaz:
> It was throwing before I fixed it, the example app starts with a `signInSilently()`, which immidiately throws into the IDE, when I fixed it, it stopped throwing. It makes sense, as the VM is "ignoring" all `catch` blocks, and `signInSilently()` ignores the exceptions by catching them.
> 
> The whole issue was that the VM was ignoring `try-catch` blocks, so the configuration inside `signInSilently()` shouldn't matter.

Makes sense, I think we can merge this now since the test is covered.

feinstein:
I think this fix was regressed as I am experiencing the same problem after an update.

maheshmnj:
@feinstein  the google sign in cancelled issue still persissts https://github.com/flutter/flutter/issues/17677#issue-323834402

feinstein:
I know, and I communicated this on other issues but no one seemed to care.

maheshmnj:
it would be really helpful for the community if someone could really fix it.

feinstein:
I don't think it will be an easy fix, the whole code was changed to a more generic one, which makes ir really difficult to fix, as the fix is to remove `then() ` for `await`.

pr-Mais:
I have this same issue, isn't there a workaround for the moment?

feinstein:
Compile in Release, thus only happens in Debug.

ahumaro:
I can confirm this issue is still present today. I am using version 4.1.1

maheshmnj:
> I can confirm this issue is still present today. I am using version 4.1.1

I can confirm it works well in release mode

brianschardt:
i can confirm this is still an issue in debug mode

PR #1939, opened 2019-08-01
[package_info] [path_provider] Add some apis

isMeiying:
Get an external path instead of an external cache path, change the original getExternalStorageDirectory to getExternalFilesDirectory and add a new getExternalStorageDirectory.

Get apk signature in flutter.

### DIFF
--- a/packages/path_provider/lib/path_provider.dart
+++ b/packages/path_provider/lib/path_provider.dart
@@ -76,6 +76,18 @@ Future<Directory> getApplicationDocumentsDirectory() async {
 /// to access outside the app's sandbox.
 ///
 /// On Android this uses the `getExternalFilesDir(null)`.
+Future<Directory> getExternalFilesDirectory() async {
+  if (Platform.isIOS)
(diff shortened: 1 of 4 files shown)


isMeiying:
> @googlebot I signed it!

packages/path_provider/lib/path_provider.dart
```diff
@@ -76,6 +76,18 @@ Future<Directory> getApplicationDocumentsDirectory() async {
...
+    throw UnsupportedError("Functionality not available on iOS");
+  final String path =
+      await _channel.invokeMethod<String>('getExternalFilesDir');
+  if (path == null) {
+    return null;
+  }
+  return Directory(path);
+}
+
+
```
collinjackson: It seems like you removed the Dartdoc from `getExternalStorageDirectory` and applied it to `getExternalFilesDirectory`. They  should both be documented.

collinjackson:
I'm going to go ahead and close this one, but I'm happy to re-review if you want to open a PR for each plugin.

PR #1949, opened 2019-08-05
[camera] Refactor Camera/Android implementation to allow reuse of Camera functionality

mehmetf:
Purpose of the PR is two-fold:
- Allow sharing Camera class across different plugins internally. Upstream reuse would take some work since it needs to be published to Maven.
- Pave way for supporting Camera1 APIs which would let us support all the way down to Android 16. That support is not in scope for this PR.


mehmetf:
@bparrishMines This is the canonical PR for my work. Please ignore the CL.

packages/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.3
+
+* Massive refactoring of Android Camera functionality to allow reuse of Camera class.
```
bparrishMines: I don't believe refactors require a version bump if no new features were added. If someone accessing this from `pub.dev` they would notice a difference on the dart side.

mehmetf: So we don't publish this at all? That's fine by me. Reverting.

packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
```diff
@@ -1,88 +1,33 @@
...
-
-    registrar.addRequestPermissionsResultListener(new CameraRequestPermissionsListener());
+    this.view = registrar.view();
+    this.imageStreamChannel =
+        new EventChannel(registrar.messenger(), "plugins.flutter.io/camera/imageStream");
   }
 
   public static void registerWith(Registrar registrar) {
-    if (registrar.activity() == null || Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
```
bparrishMines: Did you mean to remove this? I believe this check is needed when you are embedding the plugin in an Android app.

mehmetf: Yes, I meant to remove it. In general, it is a bad idea to *not* register plugins. Activity is technically not needed when this plugin is registered, so why require it?

We should probably remove the version check as well. Often apps (Dart) do not check the presence of activity or SDK version before calling a plugin. To make it as user friendly as possible, plugins should degrade gracefully and return proper errors.

If you don't register the plugin altogether, you get a catastrophic failure that says "channel name not found" which is not indicative of the real problem.

mehmetf:
What did you mean by "pending tests"? Do you mean manual tests?

bparrishMines:
I just meant the formatting and build tests.

packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
bparrishMines: The version check was added in this [PR](https://github.com/flutter/plugins/pull/1598) and was only added so a developer could still compile the plugin with sdk < 21. To do this, one would need to add a specific line in their manifest to ignore that this plugin requires 21+.

PR #1951, opened 2019-08-06
Add an initialAutoMediaPlaybackPolicy setting

amirh:
## Description

Controls whether a user action (e.g is touch event) is required in order to start media playback.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -271,6 +275,30 @@ - (void)updateJsMode:(NSNumber*)mode {
...
+    case 1:  // always_allow
+      NSLog(@"allowing auto playback");
+      if (@available(iOS 10.0, *)) {
+        configuration.mediaTypesRequiringUserActionForPlayback = WKAudiovisualMediaTypeNone;
+      } else {
+        configuration.mediaPlaybackRequiresUserAction = false;
+      }
+      break;
+    default:
+      NSLog(@"webview_flutter: unknown auto media plaback policy: %@", policy);
```
mklim: nit: The Java side will silently eat any value instead of doing nothing and logging. I think either way is fine but it would be easier to understand if they were consistent.

packages/webview_flutter/lib/src/webview_method_channel.dart
```diff
@@ -135,6 +135,7 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
       'initialUrl': creationParams.initialUrl,
       'settings': _webSettingsToMap(creationParams.webSettings),
       'javascriptChannelNames': creationParams.javascriptChannelNames.toList(),
+      'autoMediaPlaybackPolicy': creationParams.autoMediaPlaybackPolicy.index,
```
mklim: nit: Other plugins that deal with enums tend to use their stringified name instead of their ordinal. I think this is also OK, but it's a little less safe. There should at least be some kind of comment within `AutoMediaPlaybackPolicy` itself warning future devs not to re-arrange the order of enums, since it's the position and not the name that matters.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -72,6 +72,23 @@ typedef NavigationDecision NavigationDelegate(NavigationRequest navigation);
...
 
+/// Specifies possible restrictions on automatic media playback.
+///
+/// This is typically used in [WebView.initialMediaPlaybackPolicy].
+enum AutoMediaPlaybackPolicy {
+  /// Starting any kind of media playback requires a user action.
+  ///
+  /// For example: JavaScript code cannot start playing media unless the code was executed
+  /// as a result of a user action (like a touch event).
+  require_user_action_for_all_media_types,
```
mklim: nit: I think it's worth mentioning that this is the default.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -121,7 +138,10 @@ class WebView extends StatefulWidget {
     this.gestureRecognizers,
     this.onPageFinished,
     this.debuggingEnabled = false,
+    this.initialMediaPlaybackPolicy =
+        AutoMediaPlaybackPolicy.require_user_action_for_all_media_types,
```
mklim: Just checking--this is already the default, right? This should be treated as a breaking change otherwise.

packages/webview_flutter/ios/Classes/FlutterWebView.m
amirh: The iOS webview allows for a more fine grained control over this policy (e.g you can say that only audio or only video requires a user action) while the Android webview just lets you set a boolean.

The Android implementation is thus set to only allow media playback if the value is `always_allow`. The iOS implementation is built with this switch as if we decide to add more enum values we should handle each of them separately.

packages/webview_flutter/lib/src/webview_method_channel.dart
amirh: Good point, I'd say it's worth choosing one way or another and add it to the style guide, though I guess it may not be needed if we adopt type safe method channels.

Note that tests will fail if a future dev reorders the fields so I'm not super worried about this.

Added a comment on the enum.

packages/webview_flutter/lib/webview_flutter.dart
amirh: Added a comment on [WebView.initialMediaPlaybackPolicy].

packages/webview_flutter/lib/webview_flutter.dart
amirh: yes

packages/webview_flutter/ios/Classes/FlutterWebView.m
mklim: Ah okay, makes sense. Thanks.

PR #1960, opened 2019-08-08
[share] Fixed email sharing on IOS w/o filled subject

SergeShkurko:
## Description

When i share content by email on IOS app crashes and throws `NSInvalidArgumentException` due to the fact that subject cannot be null


cyanglaz:
Since this PR doesn't seem to be actively maintained. I am relentlessly closing the PR. I created another similar PR to fix the same issue https://github.com/flutter/plugins/pull/2104

PR #1964, opened 2019-08-09
[webview_flutter] Filter onChanged events for invalid displays

amirh:
## Description

Works around an Android WebView bug that was causing a crash by filtering some DisplayListener invocations.

Older Android WebView versions had assumed that when `DisplayListener#onDisplayChanged` is invoked, the display ID it is provided is of a valid display. However it turns out that when a display is removed Android may call onDisplayChanged with the ID of the removed display, in this case the Android WebView code tries to fetch and use the display with this ID and crashes with an NPE.

The issue was fixed in the Android WebView code in https://chromium-review.googlesource.com/517913 which is available starting WebView version 58.0.3029.125 however older webviews in the wild still have this issue.

Since Flutter removes virtual displays whenever a platform view is resized the webview crash is more likely to happen than other apps. And users were reporting this issue see: https://github.com/flutter/flutter/issues/30420

This change works around the webview bug by unregistering the WebView's DisplayListener, and instead registering our own DisplayListener which delegates the callbacks to the WebView's listener unless it's a onDisplayChanged for an invalid display.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/30420

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/DisplayListenerProxy.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/DisplayListenerProxy.java
@@ -1,0 +1,142 @@
+package io.flutter.plugins.webviewflutter;
+
+import android.annotation.TargetApi;
+import android.hardware.display.DisplayManager;
+import android.os.Build;
+import android.util.Log;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+
+import static android.hardware.display.DisplayManager.DisplayListener;
+
+/**
+ * Works around an Android WebView bug by filtering some DisplayListener invocations.
+ *
+ * <p>Older Android WebView versions had assumed that when {@link DisplayListener#onDisplayChanged}
+ * is invoked, the display ID it is provided is of a valid display. However it turns out that
+ * when a display is removed Android may call onDisplayChanged with the ID of the removed display,
+ * in this case the Android WebView code tries to fetch and use the display with this ID and crashes
+ * with an NPE.
+ *
+ * <p>This issue was fixed in the Android WebView code in
+ * https://chromium-review.googlesource.com/517913 which is available starting WebView version
+ * 58.0.3029.125 however older webviews in the wild still have this issue.
+ *
+ * <p>Since Flutter removes virtual displays whenever a platform view is resized the webview crash
+ * is more likely to happen than other apps. And users were reporting this issue see:
+ * https://github.com/flutter/flutter/issues/30420
+ *
+ * <p>This class works around the webview bug by unregistering the WebView's DisplayListener,
+ * and instead registering its own DisplayListener which delegates the callbacks to the WebView's
+ * listener unless it's a onDisplayChanged for an invalid display.
+ *
+ * <p>I did not find a clean way to get a handle of the WebView's DisplayListener so I'm using
+ * reflection to fetch all registered listeners before and after initializing a webview. In the
+ * first initialization of a webview within the process the difference between the lists is the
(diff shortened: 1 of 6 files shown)


packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/DisplayListenerProxy.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+@TargetApi(Build.VERSION_CODES.KITKAT)
+class DisplayListenerProxy {
+  private static final String TAG = "DisplayListenerProxy";
+
+  private ArrayList<DisplayListener> listenersBeforeWebView;
+
+  /** Should be called prior to the webview's initialization. */
+  void onPreWebViewInitialization(DisplayManager displayManager) {
+    listenersBeforeWebView = yoinkDisplayListeners(displayManager);
```
mklim: :smile:

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/DisplayListenerProxy.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  /** Should be called after the webview's initialization. */
+  void onPostWebViewInitialization(final DisplayManager displayManager) {
+    final ArrayList<DisplayListener> listeners = yoinkDisplayListeners(displayManager);
+    // We recorded
```
mklim: nit: Is this comment incomplete?

packages/webview_flutter/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -38,6 +38,7 @@
 /* End PBXCopyFilesBuildPhase section */
```
mklim: nit: The changes in this file look like they could be reverted.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -28,14 +29,21 @@
...
   @SuppressWarnings("unchecked")
   FlutterWebView(
-      Context context,
+      final Context context,
       BinaryMessenger messenger,
       int id,
       Map<String, Object> params,
       final View containerView) {
+
+    DisplayListenerProxy displayListenerProxy = new DisplayListenerProxy();
```
mklim: The code in that class is a huge workaround, and we know we only need it for WebView `58.0.3029.125` and below. Do you think it's worth checking what webview is actually running with `WebViewCompat.getWebViewClient` and only initializing this where we need it for the bugfix?

amirh: I actually rather use this workaround widely so that potential issues are easier to track. I'm currently not aware of any issues that should be caused by the workaround.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/DisplayListenerProxy.java
amirh: Learned from the best 😄

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/DisplayListenerProxy.java
amirh: oops, completed it.

PR #1968, opened 2019-08-10
Request code needs to be below 2^16 for FragmentActivity

mehmetf:
If using FragmentActivity, the requestCode used for startActivityForResult needs to be below 65536. See https://stackoverflow.com/questions/25529865/java-lang-illegalargumentexception-can-only-use-lower-16-bits-for-requestcode.

### DIFF
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.3+1
+
+* Fix too large request code for FragmentActivity users.
+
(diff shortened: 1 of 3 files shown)


mehmetf:
@mklim Not sure what submit queue is. Does it mean I don't take any action and you will merge this in the near future?

mklim:
@mehmetf yes, I was waiting for CI to finish running before merging. I added a description to the label so it's more obvious.

PR #1970, opened 2019-08-11
[webview_flutter] Webiew post url support

charafau:
## Description

Added initial post parameters - first request to the webpage will be POST request

## Related Issues

https://github.com/flutter/flutter/issues/27730
https://github.com/flutter/flutter/issues/29930

## Checklist

- [x] I signed the [CLA].

## Breaking Change


charafau:
I would like to ask e2e test but that would require adding support for POST headers to `https://flutter-header-echo.herokuapp.com/` would that be possible?

amirh:
(Haven't reviewed yet, just replying on your testing comment:)

What I'd really want to experiment with is launching a local HTTP server as part of the test, this would have unblocked other scenarios we couldn't test as well. It may as well be too much out of scope for this PR though.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -135,6 +149,20 @@ public void onMethodCall(MethodCall methodCall, Result result) {
     }
   }
 
+  private static String initialParametersToString(Map<String, Object> parameters)
```
amirh: Expecting a Map -> `application/x-www-form-urlencoded` encoding in each platform implementation sounds a little fragile in case these implementations don't exactly match.

My preference would be to take a similar approach to Android's WebView and WkWebView where the encoding is the responsibility of the app developer, which allows some more flexibility for developers.

packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -332,6 +340,40 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
   return true;
 }
 
+- (NSMutableString*)createPostParams:(NSDictionary<NSString*, NSObject*>*)params {
+  NSMutableString* resultString = [NSMutableString string];
+  for (NSString* key in [params allKeys]) {
+    if ([resultString length] > 0) [resultString appendString:@"&"];
+    [resultString appendFormat:@"%@=%@", key, [params objectForKey:key]];
```
amirh: These values would have to be url encoded as well to match the Android implementation.
Anyway as mentioned in previous comment I think we better not make this the responsibility of the platform implementation.

packages/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -332,6 +340,40 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
...
+
+  NSMutableURLRequest* request = [NSMutableURLRequest requestWithURL:nsUrl];
+
+  NSData* postData = [resultString dataUsingEncoding:NSUTF8StringEncoding allowLossyConversion:YES];
+  NSString* contentLength = [NSString stringWithFormat:@"%ld", (unsigned long)postData.length];
+
+  [request setHTTPMethod:@"POST"];
+  [request setHTTPBody:postData];
+  [request setValue:contentLength forHTTPHeaderField:@"Content-Length"];
+  [request setValue:@"application/x-www-form-urlencoded" forHTTPHeaderField:@"Content-Type"];
```
amirh: Seems like the Android webview forces this encoding, do you know if `multipart/form-data` is allowed with WkWebView?

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -134,6 +134,7 @@ class WebView extends StatefulWidget {
     Key key,
     this.onWebViewCreated,
     this.initialUrl,
+    this.initialPostParameters,
```
amirh: I think we better start with a `WebViewController.postUrl` which covers the more generic use case (posting at any point). The `initialUrl` is a convenience mechanism for simple cases, I'd be hesitant to add complexity to it.

charafau:
Thank you for review, need to prepare for presentation now, but will come back to it in 2 weeks.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
charafau: I think this is how you're suppose to do that, webview send form by post with this.

packages/webview_flutter/ios/Classes/FlutterWebView.m
charafau: I changed implementation so developer can send encoded string by himself

packages/webview_flutter/ios/Classes/FlutterWebView.m
charafau: I don't know if you can send multipart like that, but code I've seen uses form encoded only

packages/webview_flutter/lib/webview_flutter.dart
charafau: Moved method to controller as suggested :)

charafau:
It took me longer than I thought it would. Sorry for delay. Still I don't have e2e test for that, would it be possible to add endpoint to server?

stuartmorgan:

Can you clarify your last question? What server are you wanting to add an endpoint to? (I suspect the best way to unblock an end-to-end test here would be a local webserver as discussed above, but if there are other options we can certainly explore them.)

charafau:
@stuartmorgan  That was some time ago but as far as I remeber, I wanted to write test to check if webview actually make that POST request. My idea was to integration test to this server https://github.com/flutter/plugins/blob/master/packages/webview_flutter/example/integration_test/webview_flutter_test.dart#L95 and check if it returns different response for POST

stuartmorgan:
@iskakaushik Do you know what https://flutter-header-echo.herokuapp.com/ is? Do we own it?

It seems problematic to have our integration test depending on a service that we don't have a clearly documented handling of (ownership, update process, etc.). I'm curious if there is a reason we did that originally vs. using a local server.

charafau:
@stuartmorgan  I tried to start dart http server inside test but from what I find, you'd have to run it in different isolate and that's not possible to do when test run

stuartmorgan:
@iskakaushik Ping on the server question above?

stuartmorgan:
> you'd have to run it in different isolate and that's not possible to do when test run

Can you elaborate on this? I.e., do you have docs or other resources to link to? (I'm just not familiar with this restriction, and would like to investigate further.)

charafau:
@stuartmorgan  sorry for long reply, checked if I have my old code but it's lost. I checked flutter issues and other people have the same problems with isolates in flutter driver tests:

https://github.com/flutter/flutter/issues/24703
I'm fine with adding this as isolate but I guess server is there for some reason

stuartmorgan:
> @stuartmorgan sorry for long reply, checked if I have my old code but it's lost. I checked flutter issues and other people have the same problems with isolates in flutter driver tests:
> 
> [flutter/flutter#24703](https://github.com/flutter/flutter/issues/24703)
> I'm fine with adding this as isolate but I guess server is there for some reason

It sounds like it wouldn't have worked as an isolate at the time the server was added. If that's no longer the case though, and we can make this self-contained, that would be far better than having an outside server; we're seeing here the issues with the server approach, where it's non-obvious how to update the one that's there (or even whether we still can).

charafau:
makes, sense. I will check if that's still an issue (PR is quite old) and come back with the results.

PR #1972, opened 2019-08-13
[android_intent] add flags parameter

datvo0110:
## Description
we are launching to a different package we need to set the FLAG_ACTIVITY_NEW_TASK flag

Similar to react native Link.openUrl() : https://github.com/facebook/react-native/blob/3b6f6ca4d5fcee6f1bc6d6242e3e2ef136e4d546/ReactAndroid/src/main/java/com/facebook/react/modules/intent/IntentModule.java#L97

## Checklist

- [x] I signed the [CLA].

## Breaking Change


datvo0110:
> @googlebot I signed it!

ened:
@datvo0110 please bump version & add change log entry.

I'm wondering if this should be done explicitly via an extra enum flag in the AndroidIntent class.
Can you think of other flags that may be of interest in the future?

I can think of those, for example:

```
Intent.FLAG_ACTIVITY_NEW_TASK
Intent.FLAG_ACTIVITY_NO_HISTORY
Intent.FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS
Intent.FLAG_GRANT_READ_URI_PERMISSION
```

datvo0110:
@ened I have bumped version from 0.3.2 to 0.3.3 & add change log entry. Is this ok?

ened:
@datvo let’s rewrite this a bit and add the flags argument instead of having app behavior done implicitly using the package name only. L

Are you up for changing adding and implementing the ‘flags’ parameter?

datvo0110:
ok, let's me do that! I'll try to cover most case :D

datvo0110:
I added flags parameter as List<String> and prepare some popular flags as
you recommended

Intent.FLAG_ACTIVITY_NEW_TASK
Intent.FLAG_ACTIVITY_NO_HISTORY
Intent.FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS
Intent.FLAG_GRANT_READ_URI_PERMISSION

I also added an example on how to use flags option.
I'm looking for your opinions :D

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/AndroidIntentPlugin.java
```diff
@@ -51,6 +51,29 @@ private String convertAction(String action) {
     }
   }
 
+  private int convertFlag(String flag) {
```
ened: This PR converts the flags from String to integer using this method. 
A better method is to let the Dart side know about the flag names & values and simply push the flag through. You can create a new Flags.dart (or similar) and fill in the constants from https://developer.android.com/reference/android/content/Intent (All the constants from FLAG_).

Please follow the hex numbering (0x00400000) and keep the order as it is listed on the documentation page.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/AndroidIntentPlugin.java
```diff
@@ -51,6 +51,29 @@ private String convertAction(String action) {
...
+      case "exclude_from_recents":
+        return Intent.FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS;
+      case "grant_read_uri_permission":
+        return Intent.FLAG_GRANT_READ_URI_PERMISSION;
+      default:
+        return 0;
+    }
+  }
+
+  private int convertFlags(ArrayList<String> flags) {
```
ened: As said above, this logic should be done in the Dart side.

packages/android_intent/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -9,7 +9,6 @@
 /* Begin PBXBuildFile section */
 		1498D2341E8E89220040F4C2 /* GeneratedPluginRegistrant.m in Sources */ = {isa = PBXBuildFile; fileRef = 1498D2331E8E89220040F4C2 /* GeneratedPluginRegistrant.m */; };
 		3B3967161E833CAA004F5970 /* AppFrameworkInfo.plist in Resources */ = {isa = PBXBuildFile; fileRef = 3B3967151E833CAA004F5970 /* AppFrameworkInfo.plist */; };
-		2D5378261FAA1A9400D5DBA9 /* flutter_assets in Resources */ = {isa = PBXBuildFile; fileRef = 2D5378251FAA1A9400D5DBA9 /* flutter_assets */; };
```
ened: The changes in here relate to the iOS example. If possible, please remove them from the commit.

packages/android_intent/example/lib/main.dart
```diff
@@ -117,6 +117,16 @@ class ExplicitIntentsWidget extends StatelessWidget {
     intent.launch();
   }
 
+  void _openLinkInDefaultBrowserInSeperatedWindow() {
+    final List<String> flags = <String>[];
+    flags.add('new_task');
+    final AndroidIntent intent = AndroidIntent(
+        action: 'action_view',
+        data: Uri.encodeFull('https://flutter.io'),
+        flags: flags);
```
ened: - please add a `,` to before the parentheses end
- this now becomes simpler as you can do the bit operation in Dart

packages/android_intent/example/lib/main.dart
```diff
@@ -162,6 +172,11 @@ class ExplicitIntentsWidget extends StatelessWidget {
                 child: const Text('Tap here to open link in Google Chrome.'),
                 onPressed: _openLinkInGoogleChrome,
               ),
+              RaisedButton(
+                child: const Text(
+                    'Tap here to open link in browser in a seperated window.'),
+                onPressed: _openLinkInDefaultBrowserInSeperatedWindow,
```
ened: I like a close-to-reality example like this one, however I would name it more accurate ("Start Activity in New Task"). What other ways could we use in the example to demonstrate the functionality?

datvo0110:
@ened I already
-- bring logic from java to dart, add flag list
-- add test case
-- update example
sorry because I tried to remove packages/android_intent/example/ios/Runner.xcodeproj/project.pbxproj so I have to reset to the old commit, now it contained some unrelated file @@

datvo0110:
@ened 
sorry, the commit is like a mess now
i have already prepared a new clean commit https://github.com/flutter/plugins/pull/2000
let's switch to the other commit to continue

ened:
@datvo0110 ok let's close this one and move to #2000.

datvo0110:
Hi @ened, @mklim please help me review https://github.com/flutter/plugins/pull/2000

PR #1974, opened 2019-08-13
[Firebase Auth] Fix user creation example on README

ronpetit:
## Description

Properly assign user to the FirebaseUser from the Auth Result on creating a new user from email and password on README.md


collinjackson:
This was fixed in FirebaseExtended/flutterfire#25

PR #1979, opened 2019-08-14
[webview_flutter] Add a getTitle method to WebViewController

amirh:
## Description

Allows inspecting the title of the currently loaded page.

## Related Issues

https://github.com/flutter/flutter/issues/37177

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.12
+
+* Added a getTitle getter to WebViewController.
+
 ## 0.3.11+4
 
 * Removed noisy log messages on iOS.
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -130,6 +130,9 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
       case "clearCache":
         clearCache(result);
         break;
+      case "getTitle":
+        getTitle(result);
+        break;
       default:
         result.notImplemented();
     }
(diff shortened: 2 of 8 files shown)


ened:
Could this possibly be a `Stream<String>` as the title may whenever the web view loads the next page?

ened:
> Could this possibly be a `Stream<String>` as the title may change whenever the web view loads the next page?

Actually no need, there's a page listener that's called when a (or any further) page is loaded, so developers could use that. :-) Ignore my original comment.

amirh:
> Could this possibly be a `Stream<String>` as the title may change whenever the web view loads the next page?

Good point, thanks!

I think a broadcast stream makes sense for this case, though I think we would still need this getter for the "current" value (as broadcast streams drop values if there are no listeners).

WDYT?

amirh:
> > Could this possibly be a `Stream<String>` as the title may change whenever the web view loads the next page?
> 
> Actually no need, there's a page listener that's called when a (or any further) page is loaded, so developers could use that. :-) Ignore my original comment.

Your comment is still valid as the title might change without a new page being loaded (e.g javascript code can change the title).

I do think it still makes sense to to have this getter as I commented above though.

ened:
If JS code can change the title, we'd need to fiddle with the html to get it, unless `WebView` has some sort of listener we can hook into. 

From my point of view, the current PR (to get the title) provides enough of a building block to build a stream of titles later when combining this with the page listener and perhaps other methods (even a `Timer.periodic`..).

Thx @amirh.

amirh:
> If JS code can change the title, we'd need to fiddle with the html to get it, unless `WebView` has some sort of listener we can hook into.
> 
> From my point of view, the current PR (to get the title) provides enough of a building block to build a stream of titles later when combining this with the page listener and perhaps other methods (even a `Timer.periodic`..).
> 
> Thx @amirh.

Yes such listeners are available on Android and iOS.
Anyway going to land this PR as-is it seems that we all agree this getter makes sense.

packages/webview_flutter/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
+## 0.3.12
+
+* Added a getTitle getter to WebViewController.
+
 ## 0.3.11+6
 
 * Calling destroy on Android webview when flutter webview is getting disposed.
+>>>>>>> master
```
ened: @amirh saw this?

amirh: oops

PR #1981, opened 2019-08-16
Revert "[webview_flutter] Add a getTitle method to WebViewController"

amirh:
Reverts flutter/plugins#1979

The change broke Fuchsia, as [FuchsiaWebViewController](https://fuchsia.googlesource.com/topaz/+/refs/heads/master/public/dart/fuchsia_webview_flutter/lib/src/fuchsia_webview_platform_controller.dart#20) implements WebViewPlatformController so it doesn't get the no-op implementation of getTitle. We could change the FuchsiaWebViewController implementation to extend WebViewPlatformController so we can add new methods without requiring roll coordination.

I'm reverting temporarily until we resolve the roll issue.

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,7 +1,3 @@
-## 0.3.12
-
-* Added a getTitle getter to WebViewController.
-
 ## 0.3.11+6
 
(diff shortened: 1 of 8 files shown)


amirh:
Landing without waiting for CI as this is a clean revert.

PR #1983, opened 2019-08-16
[android_intent] support for setType() and setDataAndType()

shaxxx:
## Description

Added support for [setType](https://developer.android.com/reference/android/content/Intent.html#setType(java.lang.String)) and [setDataAndType](https://developer.android.com/reference/android/content/Intent.html#setDataAndType(android.net.Uri,%20java.lang.String)) parameters. 
This is needed if IE. you want to send intent to stream video. You would need to set both data (some http Uri) and mime type (ie. 'video/*'). Keep in mind, Intent documentation specifies

> To set only the data URI, call setData(). To set only the MIME type, call setType(). If necessary, you can set both explicitly with setDataAndType().
> 
> Caution: If you want to set both the URI and MIME type, do not call setData() and setType() because they each nullify the value of the other. Always use setDataAndType() to set both URI and MIME type.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


shaxxx:
@googlebot I signed it!

shaxxx:
Gladly, but there was nothing to begin with to extend, and seems logical to
test all other functions as well.
As much as I would like to help, the force is still weak inside me.
Maybe if you write some initial tests to begin.

Regards.


pet, 23. kol 2019. 23:13 Michael Klimushyn <notifications@github.com> je
napisao:

mklim:
It would also be totally fine if you just wanted to add a test for your specific new feature instead of backfilling tests for all the existing code. No need to add coverage retroactively, we'd just like to improve what we have from here on out. But we're also happy to shepherd it forward from here. Thanks again for contributing the patch!

I'm following [the initial PR review policy](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#the-review-process). Since we're shepherding it forward from here the PR isn't trivial to review, so I'm labeling it with "backlog." We'll prioritize it according to the issue's priority.

Relevant issues:

* flutter/flutter#39148

shaxxx:
I guess the real question is how do you test intent? 
What kind of test are you expecting: integration or functional?
As far as I could see in other plugins these are all functional tests, and make perfect sense. 
To write functional test for intent one would need something to handle an intent, and return result to the test. I see no way of doing this without external app/service test would rely on which is against TDD logic.  
Only test I can think of is: check if right method is being called with right parameters.
I believe there is a good reason why this plugin did not have tests in the first place.

ened:
hi @shaxxx! We had some work on the android_intent package recently, by adding support for the `flags` parameter. It also included tests. Would that be a good start for writing tests for this PR for you? Thank you!

shaxxx:
@ened There was also a bunch of refactorings inside android_intent plugin rendering my original PR pretty much useless. I re-implemented support for `setType()` and `setDataAndType`.
Also I created 2 new tests on Java side of the plugin, and updated existing one on the Dart side.
That pretty much covers everything, this should be ready to be merged.

@mklim I think this also closes https://github.com/flutter/flutter/issues/39148 since only tests were missing in the original PR.

shaxxx:
@mklim I've updated the pubspec.yaml and CHANGELOG.

PR #1984, opened 2019-08-16
Remove Flutterfire plugins (moved to FirebaseExtended)

collinjackson:
We are moving FlutterFire code and issues from the flutter/plugins repo to [FirebaseExtended/flutterfire](github.com/FirebaseExtended/flutterfire).

This will make them easier for Firebase users to discover.

The package names will remain the same.

PR #1989, opened 2019-08-20
[firebase_dynamic_links] Don't crash if activity is missing

mehmetf:
`registrar.activity` can be null with dynamic links since it is possible to open the app and execute this code before the activity is ready.

### DIFF
--- a/packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
+++ b/packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
@@ -41,7 +41,6 @@ public class FirebaseDynamicLinksPlugin implements MethodCallHandler, NewIntentL
     FirebaseDynamicLinks.getInstance()
         .getDynamicLink(intent)
         .addOnSuccessListener(
-            registrar.activity(),
             new OnSuccessListener<PendingDynamicLinkData>() {
               @Override
               public void onSuccess(PendingDynamicLinkData pendingDynamicLinkData) {
@@ -53,7 +52,6 @@ public class FirebaseDynamicLinksPlugin implements MethodCallHandler, NewIntentL
               }
             })
         .addOnFailureListener(
-            registrar.activity(),
             new OnFailureListener() {
               @Override
               public void onFailure(@NonNull Exception e) {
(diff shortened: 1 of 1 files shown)


mehmetf:
@bparrishMines I am not sure if that call is needed for registering OnSuccessListener. There are some public examples where only the listener is passed in. Is there a way to test this?

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
```diff
@@ -117,6 +115,11 @@ public void onMethodCall(MethodCall call, Result result) {
   }
 
   private void handleGetInitialDynamicLink(final Result result) {
+    if (registrar.activity() == null) {
+      result.success(null);
```
bparrishMines: I think we should use `result.error()` here. Wouldn't the user want to know that it failed because there was no activity?

bparrishMines:
@mehmetf We haven't written tests for this yet. I've been testing it manually. I can do that for you if you need me to. Although, I believe the listeners don't require an `Activity`.

packages/firebase_dynamic_links/android/src/main/java/io/flutter/plugins/firebasedynamiclinks/FirebaseDynamicLinksPlugin.java
mehmetf: If the activity has not started, then there *is* no initial link. I think it is WAI that it returns null.

mehmetf:
Thanks @bparrishMines . I would really appreciate it if you gave it a whirl. I am not sure how to test it out.

bparrishMines:
Oh! I just remembered that we are in the middle of moving flutterfire plugins to a new repo https://github.com/firebaseextended/flutterfire.

@collinjackson All changes should be added to firebaseextended now, right? Or do we plan on cherry-picking until https://github.com/flutter/plugins/pull/1984 lands.

bparrishMines:
Also, i tested your PR and the plugin still works as intended.

mehmetf:
Thanks Maurice. I am OK submitting this to the other repo as well if you want to close this. I made this change internally already but the next time we sync plugins, we need to point to the other repo. Let me know when that work is complete, @collinjackson .

mehmetf:
I am going to close this and redo it in the new repo.

PR #1009, opened 2018-12-28
[firebase_messaging]Added modern iOS notification types 

kflip:
New authorization options (formerly known as `UIUserNotificationType`) have been provided by apple with iOS10 and iOS12: 
`provisional` (aka: silent notification), `criticalAlert`, `notificationSetting` and `carPlay`

This pull request intends to be backwards compatible until iOS8. Registering of any `UIUserNotificationType` or `UNAuthorizationOptions` is possible, if the operating system supports it.

A basic implementation of `UNUserNotificationCenter` for iOS10+ was necessary to request permission for these new types.

PR #1991, opened 2019-08-20
[firebase_messaging] check iOS push notification permissions

wwwdata:
## Description

This change introduces a method call for iOS to check whether the user has previously accepted or declined the push notification setting. This is useful to check it later in the app and then remind the user to enable the push notifications manually in the control center, because the initial request permissions dialog is only displayed once.

## Related Issues

I stuck to the old API, if #1009 is merged in, this should also be updated to use the new [UNUserNotificationCenter](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter?language=objc)

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/firebase_messaging/CHANGELOG.md
+++ b/packages/firebase_messaging/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.1.4
+
+* Add checkIosNotificationSettings method to check whether the user has previously accepted or declined the push notifications on iOS devices.
+
 ## 5.1.3
 
 * Update google-services Android gradle plugin to 4.3.0 in documentation and examples.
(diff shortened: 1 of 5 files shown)


wwwdata:
oops, just figured out that `onIosSettingsRegistered` is called after the app is launched and the settings were previously accepted/declined and only in that case. So this is unnecessary.

wwwdata:
Hm, I tested a bit more and am confused. I think when you don't explicitly make the call to request permissions under iOS the `onIosSettingsRegistered` is never called. Also when going out of the app, changing settings in the Settings app, and jumping back into the app, this was not called (probably because the app was in background during that time?)

So in order to really know what the current settings are, I would have to manually request the settings, like I implemented in this PR, right? So this is needed after all?

wwwdata:
Ah, and now the source code was moved.. OK going to migrate stuff to the other repo then 😆

PR #1990, opened 2019-08-20
[path_provider] Add missing tests

ened:
## Description

This simply activates a integration test for a newly added method (getApplicationSupportDirectory) on the Android side.

Also completes the test suite for all path_provider methods.

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

## Note

This integration test will fail until this PR is merged. I'm surprised this didn't block anywhere?

PR #1993, opened 2019-08-20
[pathprovider] Fix fall through bug

mehmetf:


ened:
@mehmetf @amirh how could we possibly write a test for that?
I was _just_ working on adding test for path_provider and this did fall through! See #1990.

```
  test('getApplicationSupportDirectory', () async {
    final Directory result = await getApplicationSupportDirectory();
    final String uuid = Uuid().v1();
    final File file = File('${result.path}/$uuid.txt');
    file.writeAsStringSync('Hello world!');
    expect(file.readAsStringSync(), 'Hello world!');
    expect(result.listSync(), isNotEmpty);
    file.deleteSync();
  });
```

rockingdice:
Any ETA about v1.2.1 of path_provider on pub?
I met this issue:
java.lang.IllegalStateException: Reply already submitted which is probably this fix can make it work.

PR #2003, opened 2019-08-22
[video_player] Added `formatHint` to to override video format on Android

jtmcdole:
## Description

Passes down the suggested video format for use when URLs do not include a file path.

https://github.com/flutter/flutter/issues/39076

### DIFF
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.2
+
+* Adds optional VideoFormat used to signal what format the plugin should try.
+
 ## 0.10.1+7
 
 * Fix tests by ignoring deprecated member use.
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.10.2
 
-* Adds optional VideoFormat used to signal what format the plugin should try.
+* **Android Only** Adds optional VideoFormat used to signal what format the plugin should try.
 
 ## 0.10.1+7
 
--- a/packages/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
+++ b/packages/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
@@ -53,6 +53,10 @@ import java.util.Map;
 public class VideoPlayerPlugin implements MethodCallHandler {
 
   private static class VideoPlayer {
+    private static final String FORMAT_SS = "ss";
+    private static final String FORMAT_DASH = "dash";
+    private static final String FORMAT_HLS = "hls";
+    private static final String FORMAT_OTHER = "other";
 
     private SimpleExoPlayer exoPlayer;
 
@@ -107,11 +111,6 @@ public class VideoPlayerPlugin implements MethodCallHandler {
       String scheme = uri.getScheme();
       return scheme.equals("file") || scheme.equals("asset");
     }
-
-    private static final FORMAT_SS = "ss";
-    private static final FORMAT_DASH = "dash";
-    private static final FORMAT_HLS = "hls";
-    private static final FORMAT_OTHER = "other";
 
     private MediaSource buildMediaSource(
         Uri uri, DataSource.Factory mediaDataSourceFactory, String formatHint, Context context) {
@@ -330,7 +329,8 @@ public class VideoPlayerPlugin implements MethodCallHandler {
   }
 
   private void onDestroy() {
-    // The whole FlutterView is being destroyed. Here we release resources acquired for all instances
+    // The whole FlutterView is being destroyed. Here we release resources acquired for all
+    // instances
     // of VideoPlayer. Once https://github.com/flutter/flutter/issues/19358 is resolved this may
     // be replaced with just asserting that videoPlayers.isEmpty().
     // https://github.com/flutter/flutter/issues/20989 tracks this.
--- a/packages/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
+++ b/packages/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
@@ -71,7 +71,8 @@ public class VideoPlayerPlugin implements MethodCallHandler {
         EventChannel eventChannel,
         TextureRegistry.SurfaceTextureEntry textureEntry,
         String dataSource,
-        Result result) {
+        Result result,
+        String formatHint) {
       this.eventChannel = eventChannel;
       this.textureEntry = textureEntry;
 
@@ -93,7 +94,7 @@ public class VideoPlayerPlugin implements MethodCallHandler {
                 true);
       }
 
-      MediaSource mediaSource = buildMediaSource(uri, dataSourceFactory, context);
+      MediaSource mediaSource = buildMediaSource(uri, dataSourceFactory, formatHint, context);
       exoPlayer.prepare(mediaSource);
 
       setupVideoPlayer(eventChannel, textureEntry, result);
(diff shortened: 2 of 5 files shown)


packages/video_player/lib/video_player.dart
```diff
@@ -148,14 +150,15 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+        formatHint = null,
         super(VideoPlayerValue(duration: null));
 
   /// Constructs a [VideoPlayerController] playing a video from obtained from
   /// the network.
   ///
   /// The URI for the video is given by the [dataSource] argument and must not be
   /// null.
-  VideoPlayerController.network(this.dataSource)
+  VideoPlayerController.network(this.dataSource, {this.formatHint})
```
cyanglaz: Add doc to describe formatHint. Also indicate it only works for android

packages/video_player/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.2
+
+* Adds optional VideoFormat used to signal what format the plugin should try.
```
cyanglaz: Note it is only on Android

packages/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
```diff
@@ -108,29 +109,27 @@ private static boolean isFileOrAsset(Uri uri) {
...
-          return new HlsMediaSource.Factory(mediaDataSourceFactory).createMediaSource(uri);
-        case C.TYPE_OTHER:
-          return new ExtractorMediaSource.Factory(mediaDataSourceFactory)
-              .setExtractorsFactory(new DefaultExtractorsFactory())
-              .createMediaSource(uri);
-        default:
-          {
-            throw new IllegalStateException("Unsupported type: " + type);
-          }
+      if (type == C.TYPE_SS || "ss".equals(formatHint)) {
```
cyanglaz: what happens if the type is C.TYPE_DASH and the formatHint is "ss"? 
If the formatHint is a back up when type is not found, we should only use formatHint after finding the type is not supported or found.
If the formatHint is an override, then we should use formatHint when it is present in the arguments and ignore the type.

packages/video_player/lib/video_player.dart
```diff
@@ -397,6 +405,21 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     value = value.copyWith(volume: volume.clamp(0.0, 1.0));
     await _applyVolume();
   }
+
+  static String _nameOf(VideoFormat format) {
+    switch (format) {
+      case VideoFormat.ss:
+        return 'ss';
```
cyanglaz: nits: Make these string constant, same on the android side.
e.g. kFormatSS, kFormatHLS...

packages/video_player/lib/video_player.dart
```diff
@@ -397,6 +405,21 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     value = value.copyWith(volume: volume.clamp(0.0, 1.0));
     await _applyVolume();
   }
+
+  static String _nameOf(VideoFormat format) {
```
cyanglaz: nits: _formatToString
Could we also add dart unit test for this conversion?

packages/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
jtmcdole: Had to wait on my coworker to get back in to ask him what he wanted. We agreed overriding is probably best.

packages/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
```diff
@@ -107,9 +108,35 @@ private static boolean isFileOrAsset(Uri uri) {
...
+    private static final FORMAT_SS = "ss";
+    private static final FORMAT_DASH = "dash";
+    private static final FORMAT_HLS = "hls";
+    private static final FORMAT_OTHER = "other";
+
     private MediaSource buildMediaSource(
-        Uri uri, DataSource.Factory mediaDataSourceFactory, Context context) {
-      int type = Util.inferContentType(uri.getLastPathSegment());
+        Uri uri, DataSource.Factory mediaDataSourceFactory, String formatHint, Context context) {
+      int type;
```
cyanglaz: nits maybe:
```java
      if (formatHint != null) {
        switch (formatHint) {
          case FORMAT_SS:
            type = C.TYPE_SS;
            break;
          case FORMAT_DASH:
            type = C.TYPE_DASH;
            break;
          case FORMAT_HLS:
            type = C.TYPE_HLS;
            break;
          case FORMAT_OTHER:
            type = C.TYPE_OTHER;
            break;
          default:
            type = -1;
            break;
        }
      } else {
        int type = Util.inferContentType(uri.getLastPathSegment());
        switch (type) {...
     }
```

packages/video_player/lib/video_player.dart
```diff
@@ -148,14 +150,16 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+        formatHint = null,
         super(VideoPlayerValue(duration: null));
 
   /// Constructs a [VideoPlayerController] playing a video from obtained from
   /// the network.
   ///
   /// The URI for the video is given by the [dataSource] argument and must not be
-  /// null.
-  VideoPlayerController.network(this.dataSource)
+  /// null. The [formatHint] option allows the caller to specify the video
```
cyanglaz: nits:
Start `The [formatHint]...` on a new line.
Also indicate it is Android only.

packages/video_player/lib/video_player.dart
```diff
@@ -148,14 +150,16 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+  /// format in the case where it cannot be detected via url (Android only).
```
cyanglaz: I think it is the opposite based on the JAVA implementation. It seems that the formatHint will override the url detection. Maybe:
`Android only: The [formatHint] option allows the caller to override the video format from url`?

packages/video_player/test/video_player_test.dart
```diff
@@ -41,6 +41,9 @@ class FakeController extends ValueNotifier<VideoPlayerValue>
   Future<void> play() async {}
   @override
   Future<void> setLooping(bool looping) async {}
+
+  @override
+  VideoFormat get formatHint => null;
```
cyanglaz: I just realized that we don't have any tests on the `initialize`
Do you mind to add one here? Just to test the formatHint is pass to the method channel correctly.

jtmcdole: I'd like to put that in another CL as we're trying to integrate with this at the moment.

packages/video_player/lib/video_player.dart
jtmcdole: I did have (Android only), but I've moved to a new line and added `Android only` to the start.

packages/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
jtmcdole: GitHub swallowed my comment:

You've moved the swtich() statement into the else. This would mean formathint wouldn't return the MediaSource and instead return null.

packages/video_player/lib/video_player.dart
jtmcdole: Not sure I've ever seen unit tests on private internal methods.

packages/video_player/CHANGELOG.md
jtmcdole: The file was updated already; outdatd comment?

packages/video_player/test/video_player_test.dart
cyanglaz: Sounds good! Thanks.

jtmcdole:
Thank you!

PR #2011, opened 2019-08-24
[shared_preferences] Unhardcode Flutter prefix in platform code

nuzelac:
## Description

This is a first step towards resolving the linked issue and being able to access all shared prefs stored in the system. I'm not sure what's the best way to proceed after removing the hardcoded prefix in ios/android. Since prefs is a singleton, maybe adding a new method in Dart that returns all unfiltered prefs, e.g. by calling `getAll("")`?

## Related Issues

flutter/flutter#38769

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,3 +43,4 @@ Audrius Karosevicius <audrius.karosevicius@gmail.com>
 Lukasz Piliszczuk <lukasz@intheloup.io>
 SoundReply Solutions GmbH <ch@soundreply.com>
 Rafal Wachol <rwachol@gmail.com>
+Nino Uzelac <uzelac.nino@gmail.com>
--- a/packages/shared_preferences/CHANGELOG.md
+++ b/packages/shared_preferences/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.4
+
+* Remove hardcoded prefix from internal platform calls
+
 ## 0.5.3+4
 
 * Copy `List` instances when reading and writing values to prevent mutations from propagating.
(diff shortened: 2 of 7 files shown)


mehmetf:
We cannot accept this this PR. This plugin assumes a specific serialization format for certain object types stored in preferences (https://github.com/flutter/plugins/blob/master/packages/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.java#L34-L36). That is why it is namespaced to "flutter."

If you have previously stored preferences in your system that you would like access to, you cannot do so from this plugin reliably. I recommend one of two things:

- Write a new plugin that is specific to your app (because only you know what type of preferences you stored in there).
- Write some custom code on the native side to *convert* the preferences to the format that this plugin accepts.

If #2 sounds reasonable, we should provide an API to register a "migrator". If such a migrator is registered, upon initialization, we can walk through all non-Flutter preferences, call the `migrate` API and store the returned value in the "flutter." namespace based on its type.

The migrator could look something like this:

```java
enum Type {
  INT,
  FLOAT,
  STRING,
  LIST,
  MAP,
  ...
}

public class FlutterPreferenceValue {
  final Type type;

  // You set one of the following values as determined by [type].
  Integer intValue;
  Float floatValue;
  BigInteger bigIntegerValue;
  String stringValue;
  List listValue;
  Map mapValue;
  ...
}

// You implement this interface and give it to SharedPreferences plugin.
public interface PreferencesMigrator {
  // Only you know how to convert from Object to one of the supported types.
  // If you don't store any List, Map or BigInteger types, this function would be trivial.
  FlutterPreference migrate(String preferencesKey, Object preferencesValue);
}
```

nuzelac:
Thank you for the explanation, I understand why this is not a good approach for the current plugin. 

I have only stored strings in the old app so I did not think about the big picture :)
Will do it in a new plugin (and just for the String type) as this is not a common requirement I guess.

PR #2025, opened 2019-08-28
Enable Firebase Test Lab testing of package_info package

collinjackson:
This test updates the package_info package to use the instrumentation_adapter plugin for Android testing.

Also changed the Cirrus configuration to upload the test to Firebase Test Lab.

TODOs:
- [X] #2050 needs to land first so that we can point to the released version of instrumentation_adapter.


.ci/Dockerfile
```diff
@@ -1,6 +1,20 @@
...
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | \
+    sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+
+RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
+    sudo apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+
+RUN sudo apt-get update && sudo apt-get install -y google-cloud-sdk && \
+    gcloud config set core/disable_usage_reporting true && \
+    gcloud config set component_manager/disable_update_check true
+
```
digiter: In general, fewer RUNs could reduce Docker image size.

collinjackson: I combined a couple of the run commands.

If we really wanted to be efficient we'd probably be building the Docker and putting it in the cloud like the flutter/flutter repo does: https://github.com/flutter/flutter/blob/master/.cirrus.yml#L2

collinjackson:
A more updated version of the migrated plugin is in https://github.com/flutter/plugins/pull/2160.

I'm closing this PR in favor of that one.

PR #2029, opened 2019-08-29
fix android crash when pausing or resuming video on APIs lower than 24.

BugsBunnyBR:
## Description

This PR fixes a crash when pause/resume methods are called in APIs lower than 24 on Android.

## Checklist

- [X] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.4+1
+
+* Fix Android pause and resume video crash when executing in APIs below 24.
+
 ## 0.5.4
 
 * Add feature to pause and resume video recording.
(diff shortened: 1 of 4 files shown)


BugsBunnyBR:
> Thanks for the fix! LGTM. It's unfortunate but I don't think we could add a meaningful test for this as-is since it depends on a specific Android version and we don't have working JUnit tests in the plugins.
> 
> I filed [flutter/flutter#39499](https://github.com/flutter/flutter/issues/39499) to try and avoid errors like this in the future. @bparrishMines this is something to keep in mind with the refactor, I think ideally we'd probably want better behavior here instead of just failing to pause and resume.

I think we can add a test, but it will require to encapsulate the call to `Build.VERSION.SDK_INT >= Build.VERSION_CODES.N` in a wrapper class and the test needs to run in Robolectric or in a device.

PR #2039, opened 2019-09-04
[local_auth] Added value none to BiometricsType

Menelphor:
## Description

Added the value none to the enum BiometricType.
Sometimes it is useful to have a null type in the enum BiometricsType.
It is definetly no breaking change.

## Checklist

- [ x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ x] All existing and new tests are passing.
- [ x] I updated/added relevant documentation (doc comments with `///`).
- [ x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [ x] I read and followed the [Flutter Style Guide].
- [ x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [ x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [ x] I updated CHANGELOG.md to add a description of the change.
- [ x] I signed the [CLA].
- [ x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

- [ x] No, this is *not* a breaking change.

### DIFF
--- a/packages/local_auth/lib/local_auth.dart
+++ b/packages/local_auth/lib/local_auth.dart
@@ -11,7 +11,7 @@ import 'package:meta/meta.dart';
 import 'auth_strings.dart';
 import 'error_codes.dart';
 
-enum BiometricType { face, fingerprint, iris }
+enum BiometricType { face, fingerprint, iris, none }
 
 const MethodChannel _channel = MethodChannel('plugins.flutter.io/local_auth');
 


Menelphor:
My usage of this would be:  I'm having a settings screen, where the user can choose if the app can be unlocked via a fingerprint or face id. (The text in this screen depends on the type of available authentification methods, currently 1 for all devices)

If the user chooses to not use any of the biometric authentification, i want to pass the value BiometricsType.none to a stream. Later in the app Lock Screen, a part of the screen listens on this stream and depending on the BiometricsType either no image, or a fingerprint or a face is shown.

Furthermore i have to distinguish between the stream has no data yet and no biometric authentification is chosen.

I know i could define my own object to handle the none case, but i thought perhaps one entry more to the enum is not much. 

Nevertheless Thank you for your response :)

mehmetf:
Thanks for explaining. I agree that you should build your own object for that use case since you seem to want to create this value from your own code (e.g. when the user chooses not to use any biometric authentication).

This enum is meant to be generated only by the local_auth plugin and it is closely coupled to the native implementation. Adding more functionality on top of it that does not have anything to do with the plugin itself will create confusing API.

PR #2040, opened 2019-09-04
[shared_preferences] Handling multiple files on Android

axel-op:
## Description

This PR gives the possibility to use multiple files to save the SharedPreferences. 
This for now only works on Android. On iOs, the previous behaviour is preserved. 
See https://github.com/flutter/flutter/issues/14337#issuecomment-525947648. 

## Related Issues

Fix flutter/flutter#14337

## Checklist

-> one test seems to be unfinished or/and broken. 
- [x] I signed the [CLA].

## Breaking Change


mehmetf:
This PR is fine.

I believe `SharedPreferences` instances are cached by name: https://developer.android.com/reference/android/content/Context#getSharedPreferences(java.lang.String,%20int) . So this should not be a noticeable change. It would be good to validate that there's no performance regression by calling putString() N times before and after.

packages/shared_preferences/lib/shared_preferences.dart
```diff
@@ -177,16 +206,16 @@ class SharedPreferences {
   }
 
   /// Initializes the shared preferences with mock values for testing.
-  ///
-  /// If the singleton instance has been initialized already, it is nullified.
   @visibleForTesting
-  static void setMockInitialValues(Map<String, dynamic> values) {
-    _kChannel.setMockMethodCallHandler((MethodCall methodCall) async {
+  void setMockInitialValues(Map<String, dynamic> values) {
+    throw UnimplementedError("Mocking function seems unfinished");
```
collinjackson: I'm not sure why you commented this out. Would it make sense to bring back? It is useful for unit testing apps that rely on the SharedPreferences plugin (although now that we have e2e testing that might be a reasonable alternative).

axel-op: @collinjackson I don't understand the purpose of this function. As you can see, its implementation doesn't seem complete. What do you want me to do with this exactly?

axel-op:
> This PR is fine.
> I believe SharedPreferences instances are cached by name: https://developer.android.com/reference/android/content/Context#getSharedPreferences(java.lang.String,%20int) . So this should not be a noticeable change. It would be good to validate that there's no performance regression by calling putString() N times before and after.

@mehmetf Do you want me to do this myself?

mehmetf:
@axel-op Yes. We would not want to rollback the change because it introduced a regression.

collinjackson:
@axel-op are you still working on this? I'm happy to help land this one @mehmetf's feedback is addressed.

packages/shared_preferences/lib/shared_preferences.dart
collinjackson: This method is used by apps that want to set mock SharedPreferences values for testing. It shouldn't be commented out.

axel-op:
> I believe SharedPreferences instances are cached by name: https://developer.android.com/reference/android/content/Context#getSharedPreferences(java.lang.String,%20int) . So this should not be a noticeable change. It would be good to validate that there's no performance regression by calling putString() N times before and after.

@mehmetf Here are my results with N = 10^5 (time is in microseconds):

. | Previous impl / write | Previous impl / read | New impl / write | New impl / read
--- | --- | --- | --- | ---
min | 28 | 0 | 36 | 0
max | 149136 | 347 | 849831 | 452
mean | 115.6 | 0.1 | 152.5 | 0.1
std dev | 1098.9 | 1.6 | 4431.5 | 1.9

axel-op:
> I believe SharedPreferences instances are cached by name: https://developer.android.com/reference/android/content/Context#getSharedPreferences(java.lang.String,%20int) . So this should not be a noticeable change. It would be good to validate that there's no performance regression by calling putString() N times before and after.

@mehmetf Here are my results with N = 10^5 (time is in microseconds):

. | Previous impl / write | Previous impl / read | New impl / write | New impl / read
--- | --- | --- | --- | ---
min | 28 | 0 | 36 | 0
max | 149136 | 347 | 849831 | 452
mean | 115.6 | 0.1 | 152.5 | 0.1
std dev | 1098.9 | 1.6 | 4431.5 | 1.9

Is this what you wanted?

axel-op:
After enabling some sort of cache on the Java side with a HashMap, here's what I get:

. | New impl with caching / write |
--- | ---
min | 35
max | 928371
mean | 153.35
std dev | 5428.29

axel-op:
After creating some sort of cache on the Java side with a HashMap, here's what I get:

. | New impl with caching / write
--- | ---
min | 35
max | 124522
mean | 84.27
std dev | 782.10

However I'm not sure these differences are significant

axel-op:
Furthermore, I'm struggling with the `setMockInitialValues` method with this new implementation. Could you help me on this one?

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -15,34 +15,47 @@ const MethodChannel _kChannel =
...
   ///
   /// Because this is reading from disk, it shouldn't be awaited in
   /// performance-sensitive blocks.
-  static Future<SharedPreferences> getInstance() async {
-    if (_completer == null) {
-      _completer = Completer<SharedPreferences>();
-      try {
+  ///
+  /// WARNING: [filename] argument for now only works on Android.
+  /// On iOs, the default name will always be used, even with different value in parameter.
```
mehmetf: On *iOS*

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -15,34 +15,47 @@ const MethodChannel _kChannel =
...
+  /// The values in [SharedPreferences] are cached.
+  /// A new instance is actually created only the first time this method is called with the specified [filename].
+  ///
+  /// If a file with the specified [filename] doesn't already exist, it will automatically be created.
+  /// The [filename] cannot be null ; otherwise an [ArgumentError] will be thrown.
+  /// The default value of [filename] is the name of the file used in the previous version of this plugin.
+  ///
+  /// For Android, see https://developer.android.com/training/data-storage/shared-preferences.html for more details on the platform implementation.
+  static Future<SharedPreferences> getInstance(
+      {String filename = "FlutterSharedPreferences"}) async {
```
mehmetf: I would leave this API alone and define a new one for Android specifically:

`getAndroidPreferencesForFile(String preferencesFile)`

Otherwise, it is misleading. Your application code will potentially contain multiple shared preferences instances and on iOS, these will clobber each other.

axel-op: I created the method `getInstanceForFile`, which will throw an `AssertionError` if the current platform isn't Android. What do you think?

axel-op:
@yjbanov I will definitely need your help here since you recently made some important changes.

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
mehmetf: Thanks Axel.

The more I think about it, the more I am worried about this API. What is your primary motivation?
- Do you intend to create multiple files backing different shared preferences? 
- or are you in a migration scenario where you have an existing app that is using a different file name and you want to hook shared preferences plugin to it? (and you have no intention to use multiple files?)

The main problem with #1 is that multiple files is not supported on iOS so your app needs to be structured and behave differently on iOS when it comes to preferences. That's a tall order because this is not just a simple switch. For instance, on iOS, if you wanted to separate your preferences, you would need to namespace them. Let's say you were separating preferences for each user.  You would do this on iOS:

`user_id1:preference1`
`user_id1:preference2`
`user_id2:preference1`
`user_id2:preference2`

but then you would write to separate files on Android? Why? You could do the same on Android too and you would not have to worry about differing behavior on different platforms.

mehmetf: I apologize for not spotting this earlier. This PR has been open for a while and we did not give proper attention to it.

mehmetf:
I am going to close this PR. Please see my update above date Nov 28 about my concerns with the API. If you want to do a migration from a non-conforming file, you can do so entirely in native code and then use this plugin.

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
axel-op: My intention was to use multiple files in Flutter to separate the data. It seemed to me that it could be easier to manage because then you can use functions like `clear` and `getAll` on an entire file without the need to filter the preferences. 
The reason why I didn't code the iOs-native part is because I don't have the knowledge to do so, but with your comment I now understand that it cannot be done at all. So it's probably better to not change the API indeed.

PR #1953, opened 2019-08-07
[path_provider] add getApplicationLibraryDirectory

jerryzhoujw:
## Description

For iOS, `NSLibraryDirectory` are more used for application data like sqlite.db or images which would like to be capacity to persistent, backed up and not visible in iTunes File Sharing when `UIFileSharingEnabled` is `TRUE` in app's plist file.

From StackOverFlow discuss: [Path Directory Usable in iOS](https://stackoverflow.com/questions/7268299/path-directory-usable-in-ios)

- `NSDocumentDirectory` is `Documents/`     (persistent, backed up, may be visible in iTunes)
- `NSLibraryDirectory` is `Library/`     (persistent, backed up, not visibe to the user)
- `NSCachesDirectory` is `Library/Caches/`     (not backed up, may be cleared by system)

From iOS framework:

- `NSLibraryDirectory`,   // various documentation, support, and configuration files, resources (Library)
- `NSApplicationSupportDirectory` = 14, // location of application support files (plug-ins, etc) (Library/Application Support)

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[...]

PR #966, opened 2018-12-07
[path_provider] Android: Support multiple external storage options.

ened:
Most Android phones have, despite a single hardware storage, an internal/external partition layout.
Additionally, some phones support external SD cards for storing files.

Apps may store files onto the internal and external storage folders. For external storage, permissions can not be strongly enforced (hence Android's requirement for the WRITE_EXTERNAL_STORAGE permission) - yet if an App stores files into the app-specific folder, that folder can be cleaned up when an App is uninstalled.

The path_provider was missing 3 parts to work this out which have been addressed by the PR:
 1. Devices with multiple external storage options (like a Sony XZ performance w/ SD card) are now supported and developers can choose to store files on either storage option. 
 1. A direct link into the apps specific files & cache folders was missing, thus requiring developers to build it themselves. E.g. `/storage/emulated/0` is different than `/storage/emulated/0/Android/data/...`. Developers should by default use the latter because it should be cleaned up automatically by the Android system when an App is uninstalled
 1. The App specific folder actually supports standard folder types like `Music`, `Photos`, `Podcasts`. Developers can now request this type in the API.

Screenshot on Pixel 2:

Screenshot on Sony XZ performance w/ additional SD card inserted:

PR #2049, opened 2019-09-09
[path_provider] Android: Support multiple external storage options

ened:
## Description

Some Android phones support multiple external storage options to store files, e.g. SD cards.
This PR adds functionality to `path_provider` so that developers can inquire about those folders.

The Android API defines a set of standard folder name constants in `android.os.Environment` which are now bridged to Dart using this PR.

## Checklist

 => An existing test does not pass currently, this was introduced in #1953. Will include the fix in this PR.

- [x] I signed the [CLA].

## Breaking Change

This is a followup PR of #966

/cc @collinjackson


packages/path_provider/lib/path_provider.dart
```diff
@@ -106,3 +111,64 @@ Future<Directory> getExternalStorageDirectory() async {
...
+  final List<String> paths =
+      await _channel.invokeListMethod<String>('getExternalCacheDirectories');
+
+  return paths.map((String path) => Directory(path)).toList();
+}
+
+/// Shadows directory values from Androids `android.os.Environment` class.
+///
+/// https://developer.android.com/reference/android/os/Environment.html#fields_1
+class AndroidEnvironment {
```
ened: - Users are advised to use a string to determine those specific folders and I am wondering where to store the string.
 - In `android_intent`, we have a list of actions in Dart (using their actual Android intent values) - see https://github.com/flutter/plugins/blob/master/packages/android_intent/lib/flag.dart

It does make sense but developers could use any other string (and may not easily discover the AndroidEnvironment class for the constants).

If we turn it into a enum (and let path provider map those values to String values internally), it would increase the complexity quite a bit.

@amirh do you think the case here is the same as in `android_intent`?

cyanglaz: I spoke to @amirh  offline and we do prefer to have it as enum in the public api. And we can have a method convert the enum to String in dart before sending it to the Android Platform.

ened:
@cyanglaz Thank you for reviewing and will update the PR tomorrow.

packages/path_provider/example/lib/main.dart
```diff
@@ -140,7 +171,39 @@ class _MyHomePageState extends State<MyHomePage> {
...
+                            StorageDirectory.music,
+                          );
+                        },
+                ),
+              ),
+            ]),
+            FutureBuilder<List<Directory>>(
+                future: _externalStorageDirectories,
+                builder: _buildDirectories),
+            Column(children: <Widget>[
```
ened: @cyanglaz I have followed the style in the current example code, but think the indentation could be shown better with a few additional "," here & there. Perhaps better to be added in a future PR?

ened:
@amirh @cyanglaz there may a problem that's not catch with either approach (enum, String etc):
When you look at the documentation here: https://developer.android.com/reference/android/os/Environment.html#fields_1 , then you find that the `DIRECTORY_*` variables are marked as *Fields* and do not show their actual constant value.

I would therefore assume that the values are not stable and thus, we should not rely on arbitrary string values.

Perhaps that's what you meant with public-facing enum & string internally, but I think there needs to be double-mapping. So on dart side enum to a transfer type (like integer) and on native side from integer to the actual constant. WDYT?

packages/path_provider/lib/path_provider.dart
```diff
@@ -61,7 +61,13 @@ Future<Directory> getApplicationSupportDirectory() async {
 
 /// Path to the directory where application can store files that are persistent,
 /// backed up, and not visible to the user, such as sqlite.db.
+///
+/// On Android, this function throws an [UnsupportedError] as no equivalent
+/// folder exists.
```
cyanglaz: nits: `path exists`

ened:
@cyanglaz back to you, when you get a chance

packages/path_provider/android/src/main/java/io/flutter/plugins/pathprovider/StorageDirectoryMapper.java
```diff
@@ -0,0 +1,49 @@
...
+      case 6:
+        return Environment.DIRECTORY_MOVIES;
+      case 7:
+        return Environment.DIRECTORY_DOWNLOADS;
+      case 8:
+        return Environment.DIRECTORY_DCIM;
+      case 9:
+        if (VERSION.SDK_INT >= VERSION_CODES.KITKAT) {
+          return Environment.DIRECTORY_DOCUMENTS;
+        }
```
cyanglaz: seems missing a return statement here.

ened: There's actually a fall-through to the `default` branch, however that could cause issues for future extension. Will amend.

packages/path_provider/android/src/main/java/io/flutter/plugins/pathprovider/StorageDirectoryMapper.java
```diff
@@ -0,0 +1,51 @@
+package io.flutter.plugins.pathprovider;
+
+import android.os.Build.VERSION;
+import android.os.Build.VERSION_CODES;
+import android.os.Environment;
+
+/** Helps to map the Dart `StorageDirectory` enum to a Android system constant. */
+class StorageDirectoryMapper {
```
cyanglaz: I don't think we want this to be public.

packages/path_provider/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.3.2
```
cyanglaz: should be 1.4.0 since we added new public APIs in this PR

packages/path_provider/android/src/main/java/io/flutter/plugins/pathprovider/StorageDirectoryMapper.java
ened: Without a specifier, this would be "package-private". Would you prefer if I move it into the plugin class?

cyanglaz: My bad it is good. For some reason I thought this is a dart file.

ened:
@cyanglaz sorry to re-request - Github shows more changes required but I can not find them anywhere.

PR #2050, opened 2019-09-09
[instrumentation_adapter] Add support for running tests with Flutter driver

collinjackson:
The instrumentation_adapter plugin doesn't support iOS yet, but we'd like to start sharing integration test code between iOS and Android.

Our existing plugin integration tests are using `test()` from `package:test` rather than `testWidgets()`. This won't work with the instrumentation_adapter plugin -- they need to be converted over to use `testWidgets` to ensure that the tests results are reported properly. But `testWidgets()` only works when the `TestWidgetsFlutterBinding` is initialized, and the Flutter driver extension uses a binding that doesn't extend `TestWidgetsFlutterBinding`.

By emulating the Flutter driver extension in the instrumentation_adapter, we can ensure that existing tests can be converted to Flutter plugins CI using the `flutter_plugin_tools drive-examples` command by replacing the call to `enableFlutterDriverExtension` with `InstrumentationAdapterFlutterBinding.ensureInitialized()`

### DIFF
--- a/packages/instrumentation_adapter/lib/instrumentation_adapter.dart
+++ b/packages/instrumentation_adapter/lib/instrumentation_adapter.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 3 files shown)


packages/instrumentation_adapter/lib/instrumentation_adapter.dart
```diff
@@ -32,14 +36,46 @@ class InstrumentationAdapterFlutterBinding
...
+        case 'get_health':
+          response = <String, String>{'status': 'ok'};
+          break;
+        default:
+          throw UnimplementedError('$command is not implemented');
+      }
+      return <String, dynamic>{
+        'isError': false,
+        'response': response,
+      };
```
digiter: When would this function return "isError": true?

collinjackson: We have to attach `'isError': false` to the return value or the response will be rejected by `FlutterDriver`. I thought about returning `'isError': true` in the event of test failure.

PR #2061, opened 2019-09-13
[image_picker] Fix ios scaling in portrait mode

Xavjer:
## Description

Fixes the issue, that within image_picker when taking a picture on iOS in Portrait mode, the scaling did not work as intended (AspectRatio was flipped)

## Related Issues

https://github.com/flutter/flutter/issues/40334 => Will be fixed with this PR

## Checklist

=> No new unit tests added, as the existing should already cover this, but the image did not have the values that were returned from the channel.
- [X] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/image_picker/ios/Classes/FLTImagePickerImageUtil.m
+++ b/packages/image_picker/ios/Classes/FLTImagePickerImageUtil.m
@@ -31,57 +31,18 @@
 + (UIImage *)scaledImage:(UIImage *)image
                 maxWidth:(NSNumber *)maxWidth
                maxHeight:(NSNumber *)maxHeight {
-  double originalWidth = image.size.width;
-  double originalHeight = image.size.height;
+    NSData *imageData = UIImagePNGRepresentation(image);
+    CGImageSourceRef src = CGImageSourceCreateWithData((__bridge CFDataRef)imageData, NULL);
+    CFDictionaryRef options = (__bridge CFDictionaryRef) @{
+                                                           (id) kCGImageSourceCreateThumbnailWithTransform : @YES,
+                                                           (id) kCGImageSourceCreateThumbnailFromImageAlways : @YES,
+                                                           (id) kCGImageSourceThumbnailMaxPixelSize : MAX(maxWidth, maxHeight)
+                                                           };
 
-  bool hasMaxWidth = maxWidth != (id)[NSNull null];
-  bool hasMaxHeight = maxHeight != (id)[NSNull null];
-
-  double width = hasMaxWidth ? MIN([maxWidth doubleValue], originalWidth) : originalWidth;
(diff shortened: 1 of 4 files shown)


cyanglaz:
@xavjer Can we just flip the width and height for portrait pictures instead of using a completely different API? This change may be a bit risky until we set up a CI and write tests for this.

cyanglaz:
I created this PR for an alternative fix. Do you want to take a look? @Xavjer

Xavjer:
@cyanglaz I was thinking about that too, but from my understanding this is the better way to create a thumbnail, as the resulting image has a better resolution (and memory/performance improvement, especially during the case that we have to rotate the image) and it get's rid of the "issue" to rotate the image first.

This is apple's recommended way, see: https://medium.com/@prafullkumar77/image-usage-memory-comparison-and-best-practices-in-ios-wwdc2018-4a8919019ae9
Or the WWDC2018 video referenced there https://developer.apple.com/videos/play/wwdc2018/219/

Your fix should work too, but I was trying to improve the code itself not only fix the issue.

packages/image_picker/ios/Classes/FLTImagePickerImageUtil.m
```diff
@@ -31,57 +31,18 @@ @implementation FLTImagePickerImageUtil : NSObject
...
-
-  UIImage *scaledImage = UIGraphicsGetImageFromCurrentImageContext();
-  UIGraphicsEndImageContext();
-  return scaledImage;
+  NSData *imageData = UIImagePNGRepresentation(image);
+  CGImageSourceRef src = CGImageSourceCreateWithData((__bridge CFDataRef)imageData, NULL);
+  CFDictionaryRef options = (__bridge CFDictionaryRef) @{
+    (id)kCGImageSourceCreateThumbnailWithTransform : @YES,
+    (id)kCGImageSourceCreateThumbnailFromImageAlways : @YES,
+    (id)kCGImageSourceThumbnailMaxPixelSize : MAX(maxWidth, maxHeight)
```
cyanglaz: It seems that the behavior would be different here.
Imagine we have an image that we want to specify maxWidth to be 100 and height to be 400. The original image is very big say 2000 by 4000.
In our old implementation, the final image would be 100 * 200.
In this implementation, the final image would be 400 * 800.

Xavjer: You are right, that would happen in this solution. There is only a maxDimension which can be set with this, therefore your recommended way will probably be the correct solution. For my usecase, I want to reduce the imagesize for the larger side to be 1024, therefore that worked for my case.

cyanglaz:
Closing this PR and we will solve the issue with https://github.com/flutter/plugins/pull/2070
@Xavjer If think your solution is more for https://github.com/flutter/flutter/issues/26191. If you are interested, you could create a PR to add a new method which generates a thumbnail for the image

PR #2069, opened 2019-09-19
swap width and height when source image orientation is left or right

owops:
fix issue:
https://github.com/flutter/flutter/issues/40831


owops:
@googlebot I signed it!

PR #2084, opened 2019-09-24
[update] local_auth - intl version

rodydavis:
Upgrade local auth `intl` -> ^0.16.0


amirh:
Do you know whether the breaking change to the intl package is breaking for the users of local_auth? (if so this should be a major bump for local_auth as well).

cc @collinjackson

AppleEducate:
I just bumped it for a need in my app, and I required no changes!

amirh:
> I just bumped it for a need in my app, and I required no changes!

We should understand whether this is the case for all use cases.

AppleEducate:
Ok what can we do to test that?

amirh:
We should understand what is the breaking change, and whether/how it affect this plugin.

AppleEducate:
The Flutter on the master channel uses intl 0.16.0 and so do many plugins, so regardless, having a version that supports the latest can be an option for those requiring it.

As for upgrade path, they can continue to use 0.6.0 since there are no changes.

amirh:
Agreed that we need to support intl 0.16.0, I want to figure versioning correctly though, if we do an unnecessary major version bump it means that future bugfixes may not arrive to all apps that can get them.

AppleEducate:
Gotcha, I understand. What about a minor version bump 0.6.1?

amirh:
That's why we need to understand if this is a breaking change for the users of this plugin, if it is then this must be a major version bump, otherwise this shouldn't be a major version bump.

AppleEducate:
Ok sounds good. I made it a minor version bump. Unless we find a breaking change.

amirh:
What we should do is understand the nature of the breaking change to intl and whether/how it affects the users of this plugin.

rodydavis:
OK sounds good. Here are the changes to intl since 0.15.1 -> 0.16.0

## 0.16.0
 * Fix 'k' formatting (1 to 24 hours) which incorrectly showed 0 to 23.
 * Tighten up types in a couple of places.
 * Add dart2js pragmas for inlining to help remove descriptions and other
   compile-time information from the output.

## 0.15.8
 * Add return type to some internal methods to improve dart2js output.
 * Change parameter types in some public methods from dynamic (implicit or
   explicit) to Object. In particular, the examples and args parameters on
   Intl.message, Intl.plural, Intl.gender, and Intl.select, as well as the args
   parameter on MessageLookup.
 * Allow Dart enums in an Intl.select call. The map of cases can either take
   enums directly, or the short string name of the enum. Requires
   intl_translation 0.17.4 to take advantage of this.

## 0.15.7
 * Upate to require Dart 2.0. Remove deprecated calls,
 * Compensate for rare cases where a parsed Date in date-only format gets a
   1:00am time. This is presumably because of DST time-shifts. We may not be
   able to correct these dates, because midnight may not exist at a transition
   date, but we can cause the strict parsing to not fail for these dates.
 * Update tests to split VM and web number tests, since larger integers now fail
   to compile with dart2js.

## 0.15.6
 * More upper case constant removal.

## 0.15.5
 * Add type parameters on numberFormatSymbols for Dart 2 compatibility. Note
   that it only adds them on the right-hand side because adding them to the
   static type can cause unnecessary cast warnings.
 * Replace uses of JSON constant for Dart 2 compatibility.

## 0.15.4
 * A couple of minor Dart 2 fixes.

## 0.15.3
 * Add a customPattern parameter to the currency constructor. This can be used
   to provide a custom pattern if you have one, e.g. for accounting formats.
 * Update data to CLDR 32.0.1
 * Update for Dart 2.0 fixed-size integers.
 * Add missing support for specifying decimalDigits in compactSimpleCurrency.
 * Fix doc comments for DateFormat (Pull request dart-lang/intl#156)
 * Added a skip argument to not output the message in the extract step.
 * Compensate for parsing a Date that happens at a DST transition, particularly
   in Brazil, where the transition happens at midnight. This can result in a
   time of 11:00pm the previous day, or of 1:00am the next day. Make sure that
   the the 11:00pm case does not cause us to get the wrong date.

## 0.15.2
 * Group the padding digits to the left of the number, if present. e.g. 00,001.
 * Tweak lookup code to support translated messages as JSON rather than code.
 * Update data to CLDR 31.0.1
 * Adds locales en_MY, fr_CH, it_CH, and ps.
 * Use locale digits for printing DateTime. This can also be disabled for a
   particular locale use useNativeDigitsByDefaultFor or for a particular
   DateFormat instance use useNativeDigits.
 * Provide a library for custom-initialized DateTime and number formatting. This
   allows easier custom locales and synchronous initialization.

AppleEducate:
looks like it was mostly bug fixes but ` Remove deprecated calls,` could be an issue.

amirh:
It's not completely clear from the changelog what is the breaking change in 0.16.0, filed https://github.com/dart-lang/i18n/issues/169 to ask for a clarification.

amirh:
Seems like the major version bump for the intl package was a mistake, we should use `>=0.15.1 <0.17.0` as the intl constraint here and make this a patch version bump.

rodydavis:
Ok sounds great 👍🏼

AppleEducate:
OK i just updated it to use `>=0.15.1 <0.17.0`

packages/local_auth/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Flutter plugin for Android and iOS device authentication sensors
   such as Fingerprint Reader and Touch ID.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/local_auth
-version: 0.6.0
+version: 0.6.1
```
amirh: Should be 0.6.0+1

amirh: I pushed an update

amirh (approved):
LGTM

PR #2090, opened 2019-09-25
[camera] Update and migrate iOS example project

jmagman:
## Description

1. Let https://github.com/flutter/flutter/pull/26630 flutter_assets migrator run
1. Run `pod install` so flutter_assets is removed from the asset copy build phase
1. Migrate deprecated "English" to "en" language
1. Allow Xcode to remove extraneous xcconfigs, see https://github.com/flutter/flutter/pull/38724
1. Let Xcode 11 update build settings
1. Remove DEVELOPMENT_TEAM references since having those provisioning profiles should not be required to run the examples (most examples don't have one)
1. Looks like this was last run with `use_frameworks!`?  Let CocoaPods build as libraries instead of frameworks.
1. Remove ARCHS, which was causing a compilation error

## Related Issues

Get to a buildable baseline so I can start working on https://github.com/flutter/flutter/issues/41007.

## Checklist
- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.5
+
+* Update and migrate iOS example project.
+
 ## 0.5.4+2
 
 * Fix Android NullPointerException on devices with only front-facing camera.
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.5.5
+## 0.5.4+3
 
 * Update and migrate iOS example project.
 
--- a/packages/camera/example/ios/Runner.xcodeproj/project.pbxproj
+++ b/packages/camera/example/ios/Runner.xcodeproj/project.pbxproj
@@ -426,7 +426,6 @@
 			baseConfigurationReference = 9740EEB21CF90195004384FC /* Debug.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
-				DEVELOPMENT_TEAM = "";
 				ENABLE_BITCODE = NO;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
@@ -448,7 +447,6 @@
 			baseConfigurationReference = 7AFA3C8E1D35360C0083082E /* Release.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
-				DEVELOPMENT_TEAM = "";
 				ENABLE_BITCODE = NO;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
--- a/packages/camera/example/ios/Runner.xcodeproj/project.pbxproj
+++ b/packages/camera/example/ios/Runner.xcodeproj/project.pbxproj
@@ -8,15 +8,12 @@
 
 /* Begin PBXBuildFile section */
 		1498D2341E8E89220040F4C2 /* GeneratedPluginRegistrant.m in Sources */ = {isa = PBXBuildFile; fileRef = 1498D2331E8E89220040F4C2 /* GeneratedPluginRegistrant.m */; };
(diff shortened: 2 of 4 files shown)


jmagman:
Does a change to the example app require a build suffix version bump?

packages/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.5
```
mklim: nit: This should be `0.5.4+3`. Pub changes it's semver scheme for when the plugin is above or below 1.0.0. When it's above 1.0.0 it's `major.minor.patch` but when it's below it's `0.major.minor+patch`.

jmagman: Thanks for the catch, sorry about that!

jmagman:
> @Hixie can we get an exemption for this one and the other build refactoring patches @jmagman is landing?

These changes get us to a baseline where the iOS example apps are migrated and build, so I can start more easily adding tests without including all these diff lines in a PR for every plugin, mostly to make those future reviews smaller and more focused.

mklim:
> @Hixie can we get an exemption for this one and the other build refactoring patches @jmagman is landing?

I talked with @Hixie offline. The `DEVELOPMENT_TEAM` removal is something that we could/should test for here. We should add something to our CI that greps our examples and makes sure that it's an empty string.

I still think we can't land that exactly in this PR because other plugins are failing so CI would go red. But @jmagman either you or I can land it as a followup to these fixes to prevent this from regressing. I can add it in myself once these are landed.

jmagman:
> I talked with @Hixie offline. The `DEVELOPMENT_TEAM` removal is something that we could/should test for here. We should add something to our CI that greps our examples and makes sure that it's an empty string.

Better yet, it shouldn't be present at all.  Let me fix that in this change.

jmagman:
> our infra make sure that the example app still builds on iOS. 

I don't see how anything could be testing this builds.  Here's what it looks like on master:
```
$ cd packages/camera/example/ios 
$ xcodebuild CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=- EXPANDED_CODE_SIGN_IDENTITY=- -workspace Runner.xcworkspace -scheme Runner -configuration Debug
...

** BUILD FAILED **


The following build commands failed:
	CpResource /Users/m/Projects/plugins/packages/camera/example/ios/Flutter/flutter_assets /Users/m/Library/Developer/Xcode/DerivedData/Runner-awfyzlddjninmfeqadavdewbrcar/Build/Products/Debug-iphoneos/Runner.app/flutter_assets
(1 failure)

**Comment by mklim** [comment]
> > our infra make sure that the example app still builds on iOS.
> 
> I don't see how anything could be testing this builds. Here's what it looks like on master:

The Cirrus task is here: https://github.com/flutter/plugins/blob/master/.cirrus.yml#L76-L78. That goes through a couple layers, but it ends up running `flutter build ios --no-codesign` https://github.com/flutter/plugin_tools/blob/194fcac6a8e4da55153b35765a4de59c9148d061/lib/src/build_examples_command.dart#L40-L48. I think the `--no-codesign` is probably why it's green in CI.

**Comment by jmagman** [comment]
> > > our infra make sure that the example app still builds on iOS.
> > 
> > 
> > I don't see how anything could be testing this builds. Here's what it looks like on master:
> 
> The Cirrus task is here: https://github.com/flutter/plugins/blob/master/.cirrus.yml#L76-L78. That goes through a couple layers, but it ends up running `flutter build ios --no-codesign` https://github.com/flutter/plugin_tools/blob/194fcac6a8e4da55153b35765a4de59c9148d061/lib/src/build_examples_command.dart#L40-L48. I think the `--no-codesign` is probably why it's green in CI.

Ah, it's because it runs the migrator every time. Note the "Removing obsolete reference to flutter_assets from Runner.app" line.

```
$ flutter build ios --no-codesign
Warning: Building for device with codesigning disabled. You will have to manually codesign before deploying to device.
Building io.flutter.plugins.cameraExample for device (ios-release)...
Removing obsolete reference to flutter_assets from Runner.app
```

PR #2108, opened 2019-09-25
[google_maps_flutter] Add Projection methods to google_maps

iskakaushik:
This doesn't yet expose the full projection functionality but
addresses the most common usecases to translate screen coordinates
to latlngs.

Issue: https://github.com/flutter/flutter/issues/37959


packages/google_maps_flutter/lib/src/controller.dart
```diff
@@ -208,4 +208,25 @@ class GoogleMapController {
 
     return LatLngBounds(northeast: northeast, southwest: southwest);
   }
+
+  /// Return [ScreenCoordinate] of the [LatLng] in the current map view.
+  ///
+  /// A projection is used to translate between on screen location and geographic coordinates.
+  /// Screen location is in screen pixels (not display pixels) with respect to the top left corner
```
mklim: I think this would be useful information on `ScreenInformation`'s DartDoc, too. I think it would be helpful to also say on the `x` and `y` members how the coordinate system works, eg is 0,0 top left or center?

iskakaushik: Updated this.

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -226,6 +227,32 @@ public void onMethodCall(MethodCall call, MethodChannel.Result result) {
           }
           break;
         }
+      case "map#getScreenCoordinate":
+        {
+          if (googleMap != null) {
```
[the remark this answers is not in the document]
iskakaushik: Interesting approach. I like the idea. I will revisit this in a follow-up PR.

PR #2111, opened 2019-09-26
[local_auth] Api to stop authentication

karan-rawal:
## Description

This Api Allows the user to cancel the authentication. Helpful when there's in display fingerprint scanner.

## Related Issues

No related issues. It's an enhancement.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


karan-rawal:
@cyanglaz Can you please review this PR, I need it in my project.

packages/local_auth/lib/local_auth.dart
```diff
@@ -90,6 +90,19 @@ class LocalAuthentication {
...
 
+  /// Returns true if auth was cancelled successfully.
+  /// Returns false if there was some error or no auth in progress.
+  ///
+  /// Returns [Future] bool true or false:
+  Future<bool> stopAuthentication() {
+    if (_platform.isAndroid) {
+      return _channel.invokeMethod<bool>('stopAuthentication');
+    }
+    final Future<bool> future = Future<bool>(() => true);
```
mehmetf: You can replace these three lines with:

```dart
return Future.sync(() => true);
```

packages/local_auth/lib/local_auth.dart
```diff
@@ -90,6 +90,19 @@ class LocalAuthentication {
         'authenticateWithBiometrics', args);
   }
 
+  /// Returns true if auth was cancelled successfully.
```
mehmetf: Mention in the comments that this only works for Android.

packages/local_auth/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Flutter plugin for Android and iOS device authentication sensors
   such as Fingerprint Reader and Touch ID.
 author: Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/local_auth
-version: 0.6.0+2
+version: 0.6.0+3
```
mehmetf: I believe for API changes, we up the minor version. So this should be 0.6.1

packages/local_auth/lib/local_auth.dart
```diff
@@ -90,6 +90,19 @@ class LocalAuthentication {
...
+  /// Returns false if there was some error or no auth in progress.
+  ///
+  /// Returns [Future] bool true or false:
+  Future<bool> stopAuthentication() {
```
mehmetf: Out of curiosity, how would the Dart app decide to call this?

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
```diff
@@ -59,7 +60,7 @@ public void onMethodCall(MethodCall call, final Result result) {
             null);
```
mehmetf: There's a bug here for `authInProgress`. We might set it to true and never actually call `authenticate`.

Could you chage line 41 to: `if (authInProgress.get())`
then add to line 62: `authInProgress.set(true);`.

Since this method always runs in the `UIThread`, we technically don't need to use AtomicBoolean. You can also optionally change all of this to a normal `boolean` primitive.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
```diff
@@ -112,8 +113,27 @@ public void onError(String code, String error) {
...
     }
   }
+
+  /*
+   Stops the authentication if in progress.
+  */
+  private void stopAuthentication(Result result) {
+    try {
+      if (authenticationHelper != null && authInProgress.get()) {
+        authenticationHelper.stopAuthentication();
```
mehmetf: Doing this means `.authenticate` method will never return? Will one of the `AuthCompletionHandler` methods be called when you stop authentication?

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
karan-rawal: Resolved.

packages/local_auth/lib/local_auth.dart
karan-rawal: Resolved.

packages/local_auth/lib/local_auth.dart
karan-rawal: For devices with in-display fingerprint sensor, when authentication is in progress and if user clicks on any other button on the screen the fingerprint scanning will still be in progress. Hence, we needed an API to stop the authentication in such scenarios.

Otherwise, the fingerprint scanner stops working and the user has to restart the application.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
karan-rawal: @mehmetf 
1. The ```.authenticate``` will return ``` false``` when we stop authentication.
2. Yes, the ```onFailure``` method will be called.

packages/local_auth/lib/local_auth.dart
karan-rawal: Resolved.

packages/local_auth/lib/local_auth.dart
karan-rawal: Resolved.

packages/local_auth/pubspec.yaml
karan-rawal: Resolved.

packages/local_auth/lib/local_auth.dart
karan-rawal: For devices with in-display fingerprint sensor(assuming that authentication is in progress), if user clicks on any other button on the screen, the fingerprint scanning will still be in progress. Hence, we needed an API to stop the authentication in such scenarios.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
karan-rawal: Resolved. Not changing it to boolean primitive. Since, I don't want things to break. :)

packages/local_auth/lib/local_auth.dart
mehmetf: Thanks for the response!

> If user clicks on any other button on the screen

This is the part that I don't get. Why would I allow user to click anything in my app if there's authentication in progress? The user is *not* authenticated so should not be interacting with the app at all. I was assuming that on-screen auth will show exactly the same dialog. The only difference would be where you touch.

Could you please point me to a video or an article of some sort which shows cancelAuthentication in action?

karan-rawal: @mehmetf For in-display fingerprint scanners, we don't get any popup. We just get the fingerprint scanner icon.

http://tiny.cc/e345ez

*Assuming current behaviour of the plugin*
In the link, you can see that we only get the fingerprint scanner icon and no popup. The user can still type the passcode, and click done. When user clicks done, and we navigate to another screen the authentication is still going on. The plugin doesn't provide any mechanism to cancel this authentication.

I hope I was clear in explaining. :)

mehmetf: Very interesting.

Something is broken in this flow though.

- The image shows FP and a PIN and user can choose to authenticate either way.
- If both of these are *not* appearing due to this plugin then who is displaying the PIN entry method? That seems wrong. So, I assume both of these are appearing due to the plugin calling `biometric.authenticate`.
- So, assuming biometric API is showing the PIN entry, if user enters PIN instead of FP, why would `biometric.authenticate` not return "success"? Why would we still expect a fingerprint scan?

I am OK adding this API because it is supported on Android. However the use case you are describing seems very very wrong to me. The app should not be responsible for cancelling authentication *if the user already authenticated*.

If you agree with me, could you take some time to research how the Android API is supposed to be used? In particular why would PIN entry not run success/failure callbacks? (https://github.com/flutter/plugins/blob/master/packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java#L142)

karan-rawal: @mehmetf Thanks for your quick response.

1. The screenshot is not from my app but explains the exact scenario we need in our app. The screenshot is from oneplus's app locker screen.
2. The PIN entry is not due to the plugin. It's the app's screen having print entry, and ```biometric.authenticate``` just overlays the fingerprint icon on top of it for scanning (The user can still interact with other part of the screen though).
3. Because the biometric API is not responsible for showing the PIN entry, the biometric.authenticate doesn't know anything about it. There's no relation between them. :)

I'm not sure if the use case is wrong, but I have seen this flow in a lot of apps.

mehmetf: I see!

OK that makes more sense. So basically you app is offering two auth methods simultaneously and you want to cancel one if the user chooses the other one.

karan-rawal: @mehmetf Exactly. :)

karan-rawal:
@mehmetf Could you please merge this PR? I mean it's been a while. :|

mehmetf:
I normally don't merge PRs :-) However, I don't see anyone else reviewing this so I think it is safe to say that it has been approved. Thanks for the contribution!

karan-rawal:
> I normally don't merge PRs :-) However, I don't see anyone else reviewing this so I think it is safe to say that it has been approved. Thanks for the contribution!

Thanks @mehmetf :) Any idea about when it'll be on pub.
I will definitely contribute more to flutter. :)

mehmetf:
@collinjackson I can't publish this package for whatever reason. Could you please do the honors?

PR #2114, opened 2019-09-26
[camera] Migrate to new embedding

matthew-carroll:
Migrates the Camera plugin to the new Android embedding.

A doc exists internally that corresponds to this PR. If you'd like to review this PR, I would recommend reading that doc as well, to better understand the intentions.

This change serves the following purposes:

- Migrate to the new embedding
- Demonstrate how a plugin migrates to the new embedding
- Demonstrate an improved architecture that separates Flutter comms, from logical behavior, from Android details. This is done for testing purposes and for developer comprehension
- Add tests. At the time of putting up the PR, unit test coverage has gone from 0% to 50% lines of code for the plugin. This includes nearly 100% coverage for 2 of the most significant files.

For demonstration purposes, I have not yet setup the new and old plugins to share implementation details. This is for developer clarity when using this PR as a guide. Once the plugin migration effort is underway, I plan to delete as much of the old plugin as possible, pointing the old plugin to the new plugin's implementation. By pointing the old plugin to the new implementation, instead of the other way around, we can keep all the new tests that I've written.

The architecture of the Camera plugin is not 100% where I want it to be. There are still some references that prevent ideal segmentation points. However, this is probably 90% to the finish line and a reasonable place to pause for the purpose of helping others migrate 1P plugins.


packages/camera/android/src/main/java/dev/flutter/plugins/camera/CameraPlugin.java
```diff
@@ -0,0 +1,207 @@
...
+        new EventChannel(
+            pluginBinding.getFlutterEngine().getDartExecutor(),
+            "flutter.io/cameraPlugin/cameraEvents" + camera.getFlutterTexture().id());
+    camera.setupCameraEventChannel(cameraEventChannel);
+
+    camera.open(result);
+  }
+
+  @Override
+  public void onMethodCall(@NonNull MethodCall call, @NonNull final Result result) {
```
amirh: Can we find a way to not duplicate all this code?

matthew-carroll: Sure, but it looks like this PR probably won't merge because @bparrishMines has other plans for the plugin. I would recommend that each plugin dev do whatever is most appropriate in terms of replication as they migrate.

amirh: Would be useful to have a sample PR/commit that shows only the diff that's relevant to properly extending a plugin to the new embedder.

The camera refactoring plans are longer term, we will want to extend the existing plugin to support the new embedder much sooner, and I won't tie doing that to the big refactoring planned.

matthew-carroll: Ok. I'll put up a PR that does the minimum possible work and shares code.

mklim:
Closing since the camera plugin has been migrated.

PR #2115, opened 2019-09-27
[camera] Define clang modules in for iOS

jmagman:
## Description

- Limit the supported podspec platform to iOS so tests don't run (and fail) for macOS.
- Define the module by setting `DEFINES_MODULE` in the podspec.  See [CocoaPod modular headers docs](http://blog.cocoapods.org/CocoaPods-1.5.0/).
- Explicitly set `VALID_ARCHS` to architectures included in the Flutter universal binary.  This is the CocoaPods-suggested workaround to prevent tests from running on the i386 simulator.  See https://github.com/CocoaPods/CocoaPods/pull/8159.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/41439.
See also https://github.com/flutter/flutter/issues/41007.

## Tests

- Add a test_spec to the podspec so the CocoaPods linter will test them.  These tests fail before the `s.platform` and `s.pod_target_xcconfig` changes.
- Add Cirrus script to lint and test plugins with iOS platform code. This uses CocoaPods's built-in [test_specs](https://guides.cocoapods.org/using/test-specs.html) and lints and tests via `pod lib lint`.
- Exclude every iOS plugin except for camera from this script until they can be migrated.  Linter errors/warnings for all plugins:
```shell
find . -name "*\.podspec" -exec pod lib lint {} --analyze --verbose  \;
```

## Checklist
- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/script/build_all_plugins_app.sh
+++ b/script/build_all_plugins_app.sh
@@ -4,8 +4,8 @@
 # sure all first party plugins can be compiled together.
 
 # So that users can run this script from anywhere and it will work as expected.
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
-REPO_DIR="$(dirname "$SCRIPT_DIR")"
+readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
+readonly REPO_DIR="$(dirname "$SCRIPT_DIR")"
 
 source "$SCRIPT_DIR/common.sh"
 check_changed_packages > /dev/null
--- a/script/lint_darwin_plugins.sh
+++ b/script/lint_darwin_plugins.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+# This script lints and tests iOS and macOS platform code.
+
+# So that users can run this script from anywhere and it will work as expected.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+REPO_DIR="$(dirname "$SCRIPT_DIR")"
+
+source "$SCRIPT_DIR/common.sh"
+
+function lint_package() {
+  local package_name=$1
+  local platform=$2
+  local podspec_dir="$REPO_DIR/packages/$package_name/$platform/$package_name.podspec"
+  local failure_count=0
+
+  if [[ -f "$podspec_dir" ]]; then
+    echo "Linting $platform $package_name.podspec"
+
+    # Build as frameworks.
+    # This will also run any tests set up as a test_spec. See https://blog.cocoapods.org/CocoaPods-1.3.0.
+    pod lib lint "$podspec_dir" --allow-warnings --fail-fast --silent
+    if [[ $? -ne 0 ]]; then
+      error "Package $package_name has framework issues. Run \"pod lib lint $podspec_dir\" to inspect."
+      failure_count+=1
+    fi
+
+    # Build as libraries.
+    pod lib lint "$podspec_dir" --allow-warnings --use-libraries --fail-fast --silent
+    if [[ $? -ne 0 ]]; then
+      error "Package $package_name has library issues. Run \"pod lib lint $podspec_dir --use-libraries\" to inspect."
+      failure_count+=1
+    fi
+  fi
+
+  return $failure_count
+}
+
+function lint_packages() {
+  if [[ ! $(which pod) ]]; then 
+    echo "pod not installed. Skipping."
+    return
+  fi
+
+  # TODO: These packages have linter errors. Remove plugins from this list as linter issues are fixed.
+  local skipped_packages=(
+    android_alarm_manager
+    android_intent
+    battery
+    connectivity
+    device_info
+    google_maps_flutter
+    google_sign_in
+    image_picker
+    in_app_purchase
+    instrumentation_adapter
+    local_auth
+    package_info
+    path_provider
+    quick_actions
+    sensors
+    share
+    shared_preferences
+    url_launcher
+    video_player
+    webview_flutter
+  )
+
+  local failure_count=0
+  for package_name in "$@"; do
+    if [[ ${skipped_packages[*]} =~ $package_name ]]; then
+      continue
+    fi      
+    lint_package $package_name "ios"
+    failure_count+=$?
+    lint_package $package_name "macos"
+    failure_count+=$?
+  done
+
+  return $failure_count
+}
+
+# Sets CHANGED_PACKAGE_LIST
+check_changed_packages
+
+if [[ "${#CHANGED_PACKAGE_LIST[@]}" != 0 ]]; then
+  lint_packages "${CHANGED_PACKAGE_LIST[@]}"
+fi
--- a/script/lint_darwin_plugins.sh
+++ b/script/lint_darwin_plugins.sh
@@ -10,12 +10,11 @@ source "$SCRIPT_DIR/common.sh"
 
 function lint_package() {
   local package_name=$1
-  local platform=$2
-  local podspec="$REPO_DIR/packages/$package_name/$platform/$package_name.podspec"
+  local package_dir="$REPO_DIR/packages/$package_name/"
   local failure_count=0
 
-  if [[ -f "$podspec" ]]; then
-    echo "Linting $platform $package_name.podspec"
+  for podspec in $(find "$package_dir" -name "*\.podspec"); do
+    echo "Linting $package_name.podspec"
 
     # Build as frameworks.
     # This will also run any tests set up as a test_spec. See https://blog.cocoapods.org/CocoaPods-1.3.0.
@@ -33,7 +32,7 @@ function lint_package() {
       error "Package $package_name has library issues. Run \"pod lib lint $podspec --use-libraries\" to inspect."
       failure_count+=1
     fi
-  fi
+  done
 
   return $failure_count
 }
@@ -73,9 +72,7 @@ function lint_packages() {
     if [[ ${skipped_packages[*]} =~ $package_name ]]; then
       continue
     fi      
-    lint_package $package_name "ios"
-    failure_count+=$?
-    lint_package $package_name "macos"
+    lint_package $package_name
     failure_count+=$?
   done
 
--- a/script/lint_darwin_plugins.sh
+++ b/script/lint_darwin_plugins.sh
@@ -3,32 +3,34 @@
 # This script lints and tests iOS and macOS platform code.
 
 # So that users can run this script from anywhere and it will work as expected.
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
-REPO_DIR="$(dirname "$SCRIPT_DIR")"
+readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+readonly REPO_DIR="$(dirname "$SCRIPT_DIR")"
 
 source "$SCRIPT_DIR/common.sh"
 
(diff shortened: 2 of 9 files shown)


jmagman:
> * Add Cirrus script to lint and test plugins with iOS platform code. This uses CocoaPods's built-in [test_specs](https://guides.cocoapods.org/using/test-specs.html) and lints and tests via `pod lib lint`.

Here's the first run:
https://cirrus-ci.com/task/5308706226438144?command=main#L5

script/lint_darwin_plugins.sh
```diff
@@ -0,0 +1,88 @@
...
+  local platform=$2
+  local podspec_dir="$REPO_DIR/packages/$package_name/$platform/$package_name.podspec"
+  local failure_count=0
+
+  if [[ -f "$podspec_dir" ]]; then
+    echo "Linting $platform $package_name.podspec"
+
+    # Build as frameworks.
+    # This will also run any tests set up as a test_spec. See https://blog.cocoapods.org/CocoaPods-1.3.0.
+    pod lib lint "$podspec_dir" --allow-warnings --fail-fast --silent
```
jmagman: I would like to get rid of `--allow-warnings` and add `--analyze` but there are a lot of issues that can be addressed in another PR.  Here are the camera warnings, including the error analyzer warning:

```
    - WARN  | license: Missing license type.
    - WARN  | description: The description is equal to the summary.
    - WARN  | [iOS] keys: Missing primary key for `source` attribute. The acceptable ones are: `git, hg, http, svn`.
    - ERROR | [iOS] build_pod: Static Analysis failed. You can use `--verbose` for more information. You can use `--no-clean` to save a reproducible buid environment.
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:66:24: warning: 'AVCapturePhotoOutput' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:69:43: warning: 'AVCaptureResolvedPhotoSettings' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:78:7: warning: 'JPEGPhotoDataRepresentationForJPEGSampleBuffer:previewPhotoSampleBuffer:' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:77:19: warning: 'AVCapturePhotoOutput' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:163:52: warning: this block declaration is not a prototype [-Wstrict-prototypes]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:170:32: warning: 'AVCapturePhotoOutput' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:257:26: warning: 'AVCapturePhotoOutput' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:276:3: warning: 'AVCapturePhotoSettings' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:276:39: warning: 'AVCapturePhotoSettings' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:299:48: warning: 'AVCaptureSessionPreset3840x2160' is only available on iOS 9.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:300:41: warning: 'AVCaptureSessionPreset3840x2160' is only available on iOS 9.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:850:10: warning: block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Wimplicit-retain-self]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:804:5: warning: 'AVCaptureDeviceDiscoverySession' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:804:58: warning: 'AVCaptureDeviceDiscoverySession' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | [iOS] xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:805:44: warning: 'AVCaptureDeviceTypeBuiltInWideAngleCamera' is only available on iOS 10.0 or newer [-Wunguarded-availability]
    - WARN  | [iOS] xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:494:5: warning: Incorrect decrement of the reference count of an object that is not owned at this point by the caller
    - WARN  | [iOS] xcodebuild:  /Users/m/Projects/plugins/packages/camera/ios/Classes/CameraPlugin.m:805:44: warning: 
```

script/lint_darwin_plugins.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
```
stuartmorgan: Is there a reason this script can't be Dart? It seems like anything non-trivial should be Dart if it all possible, as it's easier to maintain.

script/lint_darwin_plugins.sh
```diff
@@ -0,0 +1,88 @@
...
+
+# This script lints and tests iOS and macOS platform code.
+
+# So that users can run this script from anywhere and it will work as expected.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+REPO_DIR="$(dirname "$SCRIPT_DIR")"
```
stuartmorgan: These should be `readonly`

script/lint_darwin_plugins.sh
```diff
@@ -0,0 +1,88 @@
...
+
+source "$SCRIPT_DIR/common.sh"
+
+function lint_package() {
+  local package_name=$1
+  local platform=$2
+  local podspec_dir="$REPO_DIR/packages/$package_name/$platform/$package_name.podspec"
```
stuartmorgan: Why `_dir`, given that it's a file?

script/lint_darwin_plugins.sh
stuartmorgan: Presumably turning on modules for everything will happen before adding this, since many plugins have such problems, so this will be a long-term project. Given that, I think it would be good to capture this in a bug and add a comment about it here ("# Add --analyze and remove --allow-warnings; see https://github.com/flutter/flutter/issues/#####")

script/lint_darwin_plugins.sh
stuartmorgan: This structure is about to start changing, per https://docs.google.com/document/d/1PeS-QSAFydHXhjZrqkub-GRPMs83roh8VwhbUScWuGQ/edit#

It would be good to support the new structure as well (the plan is option 1, as I understand it) from the start so we don't accidentally start losing coverage as plugins migrate.

script/lint_darwin_plugins.sh
jmagman: All the Cirrus tests are run from shell scripts, and this was closely cribbed from [check_publish.sh](https://github.com/flutter/plugins/blob/master/script/check_publish.sh).  I will defer to the plugins team as to whatever they find most maintainable (I'd prefer to not to be the sole owner of this effort).

script/lint_darwin_plugins.sh
jmagman: Good idea.
https://github.com/flutter/flutter/issues/41443
https://github.com/flutter/flutter/issues/41444

script/lint_darwin_plugins.sh
stuartmorgan: I'm fine deferring final decision to the plugins team. I would strongly encourage them to view adding a new non-trivial script as a good opportunity to begin a transition to Dart though :)

(Also worth considering is that at some point we're going to have Windows plugins in this repo, and probably will want to run scripts for them, and I cannot stress enough how much we want to those scripts to be in Dart rather than Batch.)

script/build_all_plugins_app.sh
```diff
@@ -4,8 +4,8 @@
 # sure all first party plugins can be compiled together.
 
 # So that users can run this script from anywhere and it will work as expected.
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
-REPO_DIR="$(dirname "$SCRIPT_DIR")"
+readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
```
mklim: Neat, I didn't know bash had this.

script/lint_darwin_plugins.sh
mklim: So it's kind of scattered. A lot of our CI scripts start off in bash here, but then ultimately end up deferred to a completely separate Dart tool maintained in the [flutter/plugin_tools](https://github.com/flutter/plugin_tools) repo. It's kind of hard to follow but the flow is that CI calls a bash script, and the bash script downloads and runs random commands from that repo by calling the package via `flutter_plugin_tools`, see [these lines](https://github.com/flutter/plugins/blob/master/script/incremental_build.sh#L26-L28). 

I think it probably makes the most sense for this functionality to live as a command in that tool instead of in bash given how much is going on here. The meat of this is easier to do in shell since it needs to run `pod lint`, but I'm pretty sure that repo does expose process run commands that could be called instead. That said the bash is relatively straightforward here so I wouldn't block the PR on moving it either, I don't think maintaining it in this form would be a huge burden.

script/lint_darwin_plugins.sh
jmagman: I changed it to `find` all podspecs in each top-level package directory.

script/lint_darwin_plugins.sh
jmagman: I created https://github.com/flutter/flutter/issues/41511 and assigned it to myself.  I'll work on it once I convert a few more plugins to modules and get the plugin template done in flutter tools.

script/lint_darwin_plugins.sh
```diff
@@ -0,0 +1,88 @@
...
+  local failure_count=0
+
+  if [[ -f "$podspec_dir" ]]; then
+    echo "Linting $platform $package_name.podspec"
+
+    # Build as frameworks.
+    # This will also run any tests set up as a test_spec. See https://blog.cocoapods.org/CocoaPods-1.3.0.
+    pod lib lint "$podspec_dir" --allow-warnings --fail-fast --silent
+    if [[ $? -ne 0 ]]; then
```
mklim: nit: Please quote the variables here and throughout.

https://google.github.io/styleguide/shell.xml?showone=Quoting#Quoting

jmagman: Thanks, I forgot to check that.  Hopefully I got them all.

PR #2119, opened 2019-09-27
Add web url launcher

hterkelsen:
## Description

Adds the `url_launcher_web` plugin, the Web platform implementation of `url_launcher`.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/url_launcher/url_launcher_web/test/url_launcher_web_test.dart
```diff
@@ -0,0 +1,34 @@
...
+    });
+
+    test('can launch "https" URLs', () {
+      expect(canLaunch('https://google.com'), completion(isTrue));
+    });
+
+    test('cannot launch "tel" URLs', () {
+      expect(canLaunch('tel:5551234567'), completion(isFalse));
+    });
+  });
```
amirh: is there a way to test launch ?

packages/url_launcher/url_launcher/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: url_launcher
 description: Flutter plugin for launching a URL on Android and iOS. Supports
   web, phone, SMS, and email schemes.
 author: Flutter Team <flutter-dev@googlegroups.com>
-homepage: https://github.com/flutter/plugins/tree/master/packages/url_launcher
+homepage: https://github.com/flutter/plugins/tree/master/packages/url_launcher/url_launcher
 version: 5.1.4
```
amirh: we should bump the minor version so we can publish the updated link to pub

amirh: sorry - the patch version

stuartmorgan:
> `url_launcher_<platform>` or `<platform>_url_launcher` ? my tendency is to go with the former as that what we already used for [fuchsia webview](https://fuchsia.googlesource.com/topaz/+/refs/heads/master/public/dart/fuchsia_webview_flutter/), 

I think you mean the latter?

> (especially once it get endorsed we don't expect people to find this plugin in pub and use it...)

Except that the doc explicitly describes cases (API extension) where people will directly use the implementation package; in that case, having `foo_plugin` and `foo_plugin_bar` in their pubspec.yaml is clearer than `bar_foo_plugin`.

But in the directory structure we picked having the platform first is probably nicer in terms of navigating the repository, so I think both have relatively small pros and cons, and either is fine with me.

stuartmorgan:
The web implementation package doesn't include the dummy iOS and Android files, so adding this to a project that targets iOS and/or Android will break them. Do you want to add the workaround, or are we just waiting for the real fixes to the Flutter plugin tooling on that?

amirh:
> > `url_launcher_<platform>` or `<platform>_url_launcher` ? my tendency is to go with the former as that what we already used for [fuchsia webview](https://fuchsia.googlesource.com/topaz/+/refs/heads/master/public/dart/fuchsia_webview_flutter/),
> 
> I think you mean the latter?

Oops, yes the latter.

hterkelsen:
I have a slight preference for `url_launcher_<platform>` because "Web_____" may lead to confusion (for example, a plugin name starting with "view" or "assembly"), but I'm okay either way

packages/url_launcher/url_launcher_web/test/url_launcher_web_test.dart
hterkelsen: It would require a lot of extra code. Rather than running the test using `flutter test --platform chrome` we would have to spin up our own instance of Chrome, somehow build the JS getting the Flutter Web SDK from somewhere, call `launch`, and determine that a new tab was opened with the correct API.

packages/url_launcher/url_launcher/pubspec.yaml
hterkelsen: Done

hterkelsen:
FYI, I changed the `incremental_build.sh` script to not ignore plugin folders without a `pubspec.yaml` (since this would also cause us to ignore all federated plugins).

hterkelsen:
I updated the PR so the tests run, and the publish check works properly, PTAL

packages/url_launcher/url_launcher/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.1.4+1
```
amirh: 5.1.5 (the numbers after the + for versions >= than 1.0.0 aren't part of semver)

hterkelsen: Done.

amirh:
> I have a slight preference for `url_launcher_<platform>` because "Web_____" may lead to confusion (for example, a plugin name starting with "view" or "assembly"), but I'm okay either way

Ok, lets keep the current name,

script/common.sh
```diff
@@ -23,16 +23,11 @@ function check_changed_packages() {
...
-  # been deleted in this PR. Also filter out `location_background` since it
-  # should be removed soon.
   CHANGED_PACKAGES=""
   CHANGED_PACKAGE_LIST=()
   for package in "${packages[@]}"; do
-    if [ -f "$REPO_DIR/packages/$package/pubspec.yaml" ] && [ $package != "location_background" ]; then
-      CHANGED_PACKAGES="${CHANGED_PACKAGES},$package"
-      CHANGED_PACKAGE_LIST=("${CHANGED_PACKAGE_LIST[@]}" "$package")
-    fi
+    CHANGED_PACKAGES="${CHANGED_PACKAGES},$package"
```
amirh: Seems like the pubspec.yaml logic was introduced to prevent a deleted package to be running on CI, maybe we can just check that the directory exists instead?

cc @gspencergoog who originally introduce this logic.

script/check_publish.sh
```diff
@@ -12,8 +12,8 @@ source "$SCRIPT_DIR/common.sh"
 
 function check_publish() {
   local failures=()
-  for package_name in "$@"; do
-    local dir="$REPO_DIR/packages/$package_name"
+  for dir in $(pub global run flutter_plugin_tools list --plugins="$1"); do
```
amirh: Seems wrong to only look at the first argument (prior to this change it would check the packages for all arguments)

packages/url_launcher/url_launcher_web/test/url_launcher_web_test.dart
hterkelsen: I added some basic tests for `launch`

script/check_publish.sh
hterkelsen: That was because previously, `check_publish` was called with the dereferenced `CHANGED_PACKAGE_LIST` array, which passed the array items as separate arguments. Now there is only one argument, the packages list in a comma-separated list.

script/common.sh
hterkelsen: Changed this to check if the directory exists

PR #2122, opened 2019-09-30
[shared_preferences] Updated to use platform interface

csells:
## Description
Allows replacing the platform-specific implementation of SharedPreferences in Dart in addition to the existing ability to do so in native code by implementing the method channel contract.

This is a required step to allowing other platform implementations that don't need to implement the same method channel protocol implemented for Android and iOS, e.g. a Windows implementation could use Dart:FFI instead of method channels.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

## Notes
- I'm not sure I exported the new SharedPreferencesPlatform type correctly for implementation outside of the plugin

- I'm not sure I added my tests correctly -- I created a new test file so I'd get a new instance of the test runner, as the implementation (properly) only allows setting the ```platform``` once; the original tests depended on a method channel implementation whereas the new tests needed to test a custom implementation.

### DIFF
--- a/packages/shared_preferences/lib/method_channel_shared_preferences.dart
+++ b/packages/shared_preferences/lib/method_channel_shared_preferences.dart
@@ -1,0 +1,44 @@
+import 'dart:async';
+import 'package:flutter/services.dart';
+import 'shared_preferences_platform_interface.dart';
+
+const MethodChannel _kChannel =
+    MethodChannel('plugins.flutter.io/shared_preferences');
+
+/// Wraps NSUserDefaults (on iOS) and SharedPreferences (on Android), providing
+/// a persistent store for simple data.
+class MethodChannelSharedPreferences extends SharedPreferencesPlatform {
+  @override
+  Future<Map<String, Object>> getAll() =>
+      _kChannel.invokeMapMethod<String, Object>('getAll');
+
+  @override
+  Future<bool> remove(String key) =>
+      _kChannel.invokeMethod<bool>('remove', <String, dynamic>{'key': key});
+
+  @override
+  Future<bool> clear() => _kChannel.invokeMethod<bool>('clear');
+
+  @override
+  Future<bool> setBool(String key, bool value) => _kChannel.invokeMethod<bool>(
+      'setBool', <String, dynamic>{'key': key, 'value': value});
+
+  @override
+  Future<bool> setInt(String key, int value) => _kChannel.invokeMethod<bool>(
+      'setInt', <String, dynamic>{'key': key, 'value': value});
+
+  @override
+  Future<bool> setDouble(String key, double value) =>
+      _kChannel.invokeMethod<bool>(
+          'setDouble', <String, dynamic>{'key': key, 'value': value});
+
+  @override
+  Future<bool> setString(String key, String value) =>
+      _kChannel.invokeMethod<bool>(
+          'setString', <String, dynamic>{'key': key, 'value': value});
+
+  @override
+  Future<bool> setStringList(String key, List<String> value) =>
+      _kChannel.invokeMethod<bool>(
+          'setStringList', <String, dynamic>{'key': key, 'value': value});
+}
--- a/packages/shared_preferences/lib/shared_preferences.dart
+++ b/packages/shared_preferences/lib/shared_preferences.dart
@@ -52,15 +52,7 @@ class SharedPreferences {
   /// The default value is [MethodChannelSharedPreferences] on Android or iOS.
   static SharedPreferencesPlatform get platform {
     if (_platform == null) {
-      switch (defaultTargetPlatform) {
-        case TargetPlatform.android:
-        case TargetPlatform.iOS:
           _platform = MethodChannelSharedPreferences();
-          break;
-        default:
-          throw UnsupportedError(
-              "Trying to use the default shared preferences implementation for $defaultTargetPlatform but there isn't a default one");
-      }
     }
     return _platform;
   }
--- a/packages/shared_preferences/lib/shared_preferences.dart
+++ b/packages/shared_preferences/lib/shared_preferences.dart
@@ -3,21 +3,20 @@
 // found in the LICENSE file.
 
 import 'dart:async';
-
-import 'package:flutter/services.dart';
+import 'package:flutter/foundation.dart';
 import 'package:meta/meta.dart';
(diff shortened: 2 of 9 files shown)


packages/shared_preferences/lib/shared_preferences.dart
```diff
@@ -38,6 +37,34 @@ class SharedPreferences {
...
+    assert(_platform == null);
+    _platform = platform;
+  }
+
+  /// The SharedPreferences platform that's used by the plugin.
+  ///
+  /// The default value is [MethodChannelSharedPreferences] on Android or iOS.
+  static SharedPreferencesPlatform get platform {
+    if (_platform == null) {
+      switch (defaultTargetPlatform) {
```
stuartmorgan: I don't think the target platform is actually what we should be using in these switches. The implementation that's built into the app is going to be a function of the host build, not the target.

For instance, imagine someone building on Windows, but wanting to get a sort of "iOS Simulator lite" experience for testing, sets defaultTargetPlatform to iOS to get iOS look and feel. Having that affect the backend it tries to load here doesn't seem like it makes sense. (Yes, it could be overridden, and if/when the Dart generation is added it presumably would be automatically, but in that case why have this code at all?)

packages/shared_preferences/lib/shared_preferences.dart
```diff
@@ -38,6 +37,34 @@ class SharedPreferences {
...
+        case TargetPlatform.android:
+        case TargetPlatform.iOS:
+          _platform = MethodChannelSharedPreferences();
+          break;
+        default:
+          throw UnsupportedError(
```
stuartmorgan: It's worth discussing what the default should be for new projects, but for existing plugins I strongly recommend we default to the old behavior, which is the method channel implementation. This change would break at least two unofficial federated implementations for people who set their desktop projects' TargetPlatform to Fuchsia (which FDE's examples use a sort of "neutral" default).

stuartmorgan: (And in general, anything whitelist-style should be avoided in plugins going forward IMO, since it works against federation. I had issues with path_provider's Dart code allowing some code paths only for iOS, for instance, when in fact they would apply to macOS as well.)

csells: I like your ideas but I'm not sure what you're recommending about the code. What do you think the implementation should look like?

@amirh This conversation will be relevant to you, too, as this is the basic architecture that you proposed for the Battery plugin.

stuartmorgan: As an existing plugin, I think the whole implementation of this function should just be:
```
if (_platform == null) {
  MethodChannelSharedPreferences();
}
return _platform;
```

That preserves behavior for all unofficial implementations, and requires implementations that want a new and different behavior to opt into it.

csells: updated to use MethodChannelSharedPreferences if no other platform-specific instance is provided.

amirh:
The platform interface should go in a separate package to avoid a dependency cycle (app-facing plugin <--> platform implementation) for endorsed implementations, and to version the platform and app-facing APIs separately.

packages/shared_preferences/lib/shared_preferences.dart
amirh: Stuart's point makes sense to me for existing plugins, the only drawback is that in cases where there's no platform implementation the error message is less clear but I'm willing to live with that if we think we're going to break folks otherwise.

Ideally we'd want existing platform implementations to migrate to the new approach, so maybe worth considering logging an error when we're falling back to a method channel implementation.

packages/shared_preferences/lib/shared_preferences.dart
```diff
@@ -38,6 +37,26 @@ class SharedPreferences {
...
+  /// This property can be set to use a custom platform implementation.
+  static set platform(SharedPreferencesPlatform platform) {
+    assert(_platform == null);
+    _platform = platform;
+  }
+
+  /// The SharedPreferences platform that's used by the plugin.
+  ///
+  /// The default value is [MethodChannelSharedPreferences] on Android or iOS.
+  static SharedPreferencesPlatform get platform {
```
amirh: I think that we should keep this static field in `SharedPreferencesPlatform` so that platform implementations can depend solely on the platform interface package.

packages/shared_preferences/lib/shared_preferences_platform_interface.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+abstract class SharedPreferencesPlatform {
```
amirh: We should add Dartdocs here

packages/shared_preferences/lib/method_channel_shared_preferences.dart
```diff
@@ -0,0 +1,44 @@
...
+import 'shared_preferences_platform_interface.dart';
+
+const MethodChannel _kChannel =
+    MethodChannel('plugins.flutter.io/shared_preferences');
+
+/// Wraps NSUserDefaults (on iOS) and SharedPreferences (on Android), providing
+/// a persistent store for simple data.
+class MethodChannelSharedPreferences extends SharedPreferencesPlatform {
+  @override
+  Future<Map<String, Object>> getAll() =>
```
amirh: Flutter Style is to only use `=>` if everything including the function declaration fits in one line: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#consider-using--for-short-functions-and-methods

packages/shared_preferences/lib/shared_preferences.dart
```diff
@@ -88,68 +107,67 @@ class SharedPreferences {
   /// Saves a boolean [value] to persistent storage in the background.
   ///
   /// If [value] is null, this is equivalent to calling [remove()] on the [key].
-  Future<bool> setBool(String key, bool value) => _setValue('Bool', key, value);
+  Future<bool> setBool(String key, bool value) =>
```
amirh: ditto (`=>` Flutter style) in this file

yjbanov:
I'm taking over this work as part of adding Web support for this plugin. I'll close this PR and start a new one. Thanks @csells for bootstrapping this work!

PR #2123, opened 2019-09-30
[camera] Fix event type check

Zazo032:
## Description

Trivial change to properly check if the type of an event is an error, so a description can be added.

## Related Issues

Fixes flutter/flutter#41601

## Checklist

  - Trivial change
- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.4+4
+
+* Fix event type check
+
 ## 0.5.4+3
 
 * Update and migrate iOS example project.
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,6 +1,10 @@
-## 0.5.4+4
+## 0.5.5+1
 
 * Fix event type check
+
+## 0.5.5
+
+* Define clang modules for iOS.
 
 ## 0.5.4+3
 
--- a/packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
+++ b/packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
@@ -28,7 +28,6 @@ import android.view.Surface;
 import androidx.annotation.NonNull;
 import io.flutter.plugin.common.EventChannel;
 import io.flutter.plugin.common.MethodChannel.Result;
-import io.flutter.view.FlutterView;
 import io.flutter.view.TextureRegistry.SurfaceTextureEntry;
 import java.io.File;
 import java.io.FileOutputStream;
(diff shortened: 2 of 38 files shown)


packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -496,7 +496,7 @@ private void sendEvent(EventType eventType, String description) {
       Map<String, String> event = new HashMap<>();
       event.put("eventType", eventType.toString().toLowerCase());
       // Only errors have description
-      if (eventType != EventType.ERROR) {
+      if (eventType == EventType.ERROR) {
```
mklim: Whoops. Thank you for fixing this.

packages/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.4+4
```
mklim: nit: This should be updated to `0.5.5+1` now.

mklim:
Thank you for the fix!

We've recently re-emphasized that all PRs need tests, [following the style guide](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#write-test-find-bug).

This is difficult for plugins PRs since we don't have an integrated testing solution, but I don't think it's impossible to do here. Normally I would recommend writing a flutter driver test that tries to do something invalid and then make sure the error has a description, but tests for this plugin need to somehow automatically accept the camera permission to run and we don't have a good cross platform solution for that yet.

We do have the ability to write JUnit tests though, and I think that would also work. Would you be able to create a CameraTests.java class with a unit test making sure messages are sent with a description? I think the easiest way to do this would be to open `example/app/android` in Android studio, and then use it to automatically make and run a new unit test class for `Camera.java`.

If you're not able to carry this forward, please let us know and we'll be happy to shepherd it on from here. Thanks again for the contribution.

Zazo032:
I have been trying to generate a test class for `Camera.java` but Android Studio is ignorning me.

image

Pressing "OK" does nothing, tried checking/unchecking different methods, using Groovy/JUnit 4/JUnit 5 but Android Studio just ignores me, not sure if I'm missing something else.

mklim:
Thanks for trying! I opened up Android Studio myself to add in a "quick test" skeleton for you to pick up and then realized that actually properly unit testing this fix is actually _really_ difficult. The class wasn't set up to be testable. I didn't see a way to do it without refactoring the `send()` logic out into a new class. Even instantiating `Camera` depends on having hardware mocked out at a level that was really difficult to do. I ended up at a point where I was trying to make custom Robolectric shadows for the camera hardware and still couldn't cover everything.

So instead I ended up making [two commits](https://github.com/mklim/plugins/compare/4c082f4...489867e), first to refactor the class and then to actually test this method. What do you think about cherry picking them into this PR and then we can land it all together?

Zazo032:
Not sure if cherry picked or just merged (a bit new to this tools, excuse me if this isn't the correct process), but tests are passing correctly so this is probably ready to merge (waiting for review approvals)

PR #2125, opened 2019-09-30
[in_app_purchase] Define clang module for iOS

jmagman:
## Description

- Limit the supported podspec platform to iOS so tests don't run (and fail) for macOS.
- Define the module by setting `DEFINES_MODULE` in the podspec.  See [CocoaPod modular headers docs](http://blog.cocoapods.org/CocoaPods-1.5.0/).
- Explicitly set `VALID_ARCHS` to architectures included in the Flutter universal binary.  This is the CocoaPods-suggested workaround to prevent tests from running on the i386 simulator.  See https://github.com/CocoaPods/CocoaPods/pull/8159.
- Delete PaymentQueueTest.m.  It doesn't compile and it wasn't actually included in the example Xcode project.
- Fix a few analyzer warnings:
  - nullability warning in FIAPReceiptManager
  - NSNumber warning to check against nil or `-boolValue`

## Related Issues

Fixes https://github.com/flutter/flutter/issues/41648.
See also https://github.com/flutter/flutter/issues/41007.

## Tests

- Add a test_spec to the podspec so the CocoaPods linter will test them.  These tests fail before the `s.platform` and `s.pod_target_xcconfig` changes.
- **Move existing tests to new Tests directory so they are run on pre-submit!**
- Fix availability in tests so they pass on older simulators.

## Checklist
- [x] I signed the [CLA].

## Breaking Change


packages/in_app_purchase/ios/Tests/Stubs.h
```diff
@@ -0,0 +1,63 @@
...
+// found in the LICENSE file.
+
+#import <Foundation/Foundation.h>
+#import <StoreKit/StoreKit.h>
+#import "FIAPReceiptManager.h"
+#import "FIAPRequestHandler.h"
+#import "InAppPurchasePlugin.h"
+
+NS_ASSUME_NONNULL_BEGIN
+API_AVAILABLE(ios(11.2), macos(10.13.2))
```
jmagman: Added availability.

packages/in_app_purchase/ios/Tests/TranslatorTest.m
```diff
@@ -0,0 +1,146 @@
...
+  };
+
+  self.productMap = [[NSMutableDictionary alloc] initWithDictionary:@{
+    @"price" : @"1",
+    @"priceLocale" : [FIAObjectTranslator getMapFromNSLocale:NSLocale.systemLocale],
+    @"productIdentifier" : @"123",
+    @"localizedTitle" : @"title",
+    @"localizedDescription" : @"des",
+  }];
+  if (@available(iOS 11.2, *)) {
```
jmagman: Added availability so these tests passed with older SDKs.

packages/in_app_purchase/ios/Tests/InAppPurchasePluginTest.m
```diff
@@ -0,0 +1,206 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
jmagman: These test files were moved from the example/ project to the new Tests/ directory.  Added symlinks from the old location so the tests still pass in the example app.

packages/in_app_purchase/example/ios/in_app_purchase_pluginTests/PaymentQueueTest.m
```diff
@@ -1,183 +0,0 @@
-// Copyright 2019 The Chromium Authors. All rights reserved.
```
jmagman: @cyanglaz This test doesn't compile and wasn't included in the Xcode project, so I deleted it.  I can file an issue for it to be re-added and fixed to pass, if you still value in it.

cyanglaz: Yes, please add this back and remove the .testing part. I am not sure why it would be like this in the first place.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -154,7 +154,7 @@ - (void)addPayment:(FlutterMethodCall *)call result:(FlutterResult)result {
     SKMutablePayment *payment = [SKMutablePayment paymentWithProduct:product];
     payment.applicationUsername = [paymentMap objectForKey:@"applicationUsername"];
     NSNumber *quantity = [paymentMap objectForKey:@"quantity"];
-    payment.quantity = quantity ? quantity.integerValue : 1;
+    payment.quantity = (quantity != nil) ? quantity.integerValue : 1;
```
cyanglaz: what is the different or benefit we gain with this change?

packages/in_app_purchase/ios/Tests/Stubs.h
```diff
@@ -0,0 +1,63 @@
...
+@end
+
+@interface NSErrorStub : NSError
+- (instancetype)initWithMap:(NSDictionary *)map;
+@end
+
+@interface FIAPReceiptManagerStub : FIAPReceiptManager
+@end
+
+API_AVAILABLE(ios(7.0), macos(10.9))
```
cyanglaz: Flutter supports 8 and up so this might not be necessary. I'm ok leave it though.

packages/in_app_purchase/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: in_app_purchase
 description: A Flutter plugin for in-app purchases. Exposes APIs for making in-app purchases through the App Store and Google Play.
 author:  Flutter Team <flutter-dev@googlegroups.com>
 homepage: https://github.com/flutter/plugins/tree/master/packages/in_app_purchase
-version: 0.2.1+4
+version: 0.2.2
```
cyanglaz: If there is no public API change, we just need to update the patch, thus 0.2.1+5
I might have missed something though. Please point it out if so.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
jmagman: It fixes a static analyzer warning:
```
    157:24: warning: Converting a pointer value of type 'NSNumber *' to a primitive boolean value; instead, either compare the pointer to nil or call -boolValue
        payment.quantity = quantity ? quantity.integerValue : 1;
```

packages/in_app_purchase/ios/Tests/Stubs.h
jmagman: I'll remove it, most of the others should also have `API_AVAILABLE(ios(7.0), macos(10.9))` so this isn't consistent.

packages/in_app_purchase/pubspec.yaml
jmagman: The only public API change was adding the nullability return value.  The podspec change means this will be imported as a module, so modulemaps get generated, and CocoaPods does a lot of other magic.  I'm happy to change this to 0.2.1+5 though!

packages/in_app_purchase/ios/Tests/InAppPurchasePluginTest.m
jmagman: Turns out Xcode really didn't link the symlinking.  I changed the Xcode project to just point at the new location instead of following the symlinks.

packages/in_app_purchase/example/ios/in_app_purchase_pluginTests/PaymentQueueTest.m
jmagman: Done.

PR #2126, opened 2019-09-30
WIP Remove pod repo update from Cirrus script

jmagman:
## Description

CocoaPods 1.8 uses a CDN instead of a local spec repository.  `pod repo update` adds a full minute to Cirrus tests.  Remove.


jmagman:
This actually makes the upgrade take longer.  Closing.

PR #2131, opened 2019-10-01
[share]fix iOS crash when setting the subject to null

cyanglaz:
## Description

When subject is set to null, iOS will set the subject to NSNull which results a crash.
This PR defaults NSNull to nil before setting the subject.
It also avoid the undocumented KVC implementation in the legacy code and instead, use the UIActivityItemSource.
Also fixed a compiler warning with giving originRect an initial value while I'm at it.

## Related Issues

https://github.com/flutter/flutter/issues/39740

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/video_player/example/test_driver/video_player.dart
```diff
@@ -0,0 +1,19 @@
+import 'dart:async';
```
amirh: revert

packages/share/ios/Classes/SharePlugin.m
```diff
@@ -6,6 +6,49 @@
...
+  self = [super init];
+  if (self) {
+    _subject = subject;
+    _text = text;
+  }
+  return self;
+}
+
+- (id)activityViewControllerPlaceholderItem:(UIActivityViewController *)activityViewController {
+  return @"";
```
amirh: curious - should this be null?

cyanglaz: Per https://developer.apple.com/documentation/uikit/uiactivityitemsource/1620458-activityviewcontrollerplaceholde?language=objc

` It should be one that the activity can handle otherwise you may get an activity with empty content. `

So setting it to null/nil will result the activityController showing an empty content (manually tested)

packages/video_player/example/test_driver/video_player.dart
cyanglaz: oops! How funny that this is included in this PR

PR #2134, opened 2019-10-02
[video_player] Fixes video initialization future stall.

slightfoot:
## Description

Fixes issue where `initialize()` `Future` stalls when failing to load source data and does not throw error.

For example if the network is offline and you try to load a video. The video will attempt to load the source on Android through ExoPlayer. This error of failing to load is reported through the event stream to the client. However, this does not complete the initialization `Completer` leaving the `Future stalled waiting for the completer to complete which it never will.


slightfoot:
@amirh Added tests as requested on discord sometime ago. I also have another fix to go into release on top of this. I'll link the PR to this one and update the CHANGELOG and pubspec in that PR.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -133,6 +133,21 @@ void main() {
...
+          'http://testing.com/invalid_url',
+        );
+        try {
+          dynamic error;
+          fakeVideoPlayerPlatform.forceInitError = true;
+          await controller.initialize().catchError((dynamic e) => error = e);
+          final PlatformException platformEx = error;
+          expect(platformEx.code, equals('VideoError'));
+        } finally {
+          fakeVideoPlayerPlatform.forceInitError = false;
```
mklim: nit: Why the `try/finally` here? Is the `catchError` not really catching the error?

slightfoot: Just used to reset the forceInitError flag to ensure if any other error is thrown during testing the flag is reset so other tests can complete without issue.

slightfoot:
@mklim do you want me to update the CHANGELOG and pubspec in this PR or do you want to handle that yourself?

My other work that depends on this needs test cases and other things that I do not have time to complete on. https://github.com/slightfoot/plugins/commit/599b6b0c1ba28637070f0cef5e15318355425419

So you might want to merge this ASAP before it gets out of date.

mklim:
@slightfoot ideally this patch would be ready to land as-is. :) Since you don't have the bandwidth to work on this though I'll get it up to date.

mklim:
Thanks again for the contribution!

PR #2141, opened 2019-10-03
BugFix: `formatHint` was meant for network streams.

jtmcdole:
This hint is only valid in the network path.


jtmcdole:
> Fix looks good, we need a test before we can land

You have no testing infrastructure for that entire class.

packages/video_player/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.2+3
+
+* Fix bug where formatHint was not being pass down to network sources.
```
Tywholland: pass -> passed?

PR #2142, opened 2019-10-03
[Connectivity] migrate to the new android embedding

cyanglaz:
## Description

Migrate connectivity to the new android embedding.

## Related Issues

https://github.com/flutter/flutter/issues/41835

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #2143, opened 2019-10-03
[android_intent] Migrate to the new embedding

mklim:
## Description

Migrate android intent to the new embedding.

- Refactors the existing plugin logic into two new classes.
- Adds an implementation of the plugin for the new embedding.
- Adds a unit test.

## Related Issues

Fixes flutter/flutter#41831.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/AndroidIntentPlugin.java
```diff
@@ -0,0 +1,66 @@
...
+   * Initialize this within the {@code #configureFlutterEngine} of a Flutter activity or fragment.
+   *
+   * <p>See {@code dev.flutter.plugins.androidintentexample.MainActivity} for an example.
+   */
+  public AndroidIntentPlugin() {
+    sender = new IntentSender(/*activity=*/ null, /*context=*/ null);
+  }
+
+  /** This exists for legacy compatibility purposes. */
+  public static void registerWith(Registrar registrar) {
```
cyanglaz: Doesn't seem to be called in this PR.
We might not need this. The io MainAcitivity can be left unchanged.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -0,0 +1,116 @@
...
+    this.activity = activity;
+  }
+
+  /**
+   * Caches the given {@code applicationContext} to use for {@link #send}.
+   *
+   * <p>Also resets the cached {@code activity} to null, since the given activity should no longer
+   * be valid after applicationContext changes.
+   */
+  void setApplicationContext(@Nullable Context applicationContext) {
```
cyanglaz: How do we force people to set the ApplicationContext first and then Activity?
For example:
I want to set a new application context and a new activity. I have to know to set the application context first before calling set activity.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/AndroidIntentPlugin.java
```diff
@@ -0,0 +1,66 @@
...
+
+import androidx.annotation.NonNull;
+import io.flutter.embedding.engine.plugins.FlutterPlugin;
+import io.flutter.embedding.engine.plugins.activity.ActivityAware;
+import io.flutter.embedding.engine.plugins.activity.ActivityPluginBinding;
+import io.flutter.plugin.common.MethodChannel;
+import io.flutter.plugin.common.PluginRegistry.Registrar;
+
+/**
+ * The new embedding implementation of the plugin.
```
matthew-carroll: I'm not sure if this is temporary or not, but is it a good idea to include temporal references in docs?

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/AndroidIntentPlugin.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /** This exists for legacy compatibility purposes. */
+  public static void registerWith(Registrar registrar) {
+    io.flutter.plugins.androidintent.AndroidIntentPlugin.registerWith(registrar);
+  }
+
+  @Override
+  public void onAttachedToEngine(@NonNull FlutterPluginBinding binding) {
+    sender.setApplicationContext(binding.getApplicationContext());
```
matthew-carroll: To clarify, do we use the application context when it's available because devs might be sending Intents with data that don't launch an `Activity`? Because launching an `Activity` from an application context requires special window settings.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -0,0 +1,116 @@
...
+import android.os.Bundle;
+import android.text.TextUtils;
+import android.util.Log;
+import androidx.annotation.Nullable;
+
+/** Forms and launches intents. */
+public class IntentSender {
+  private static final String TAG = "IntentSender";
+
+  private @Nullable Activity activity;
```
matthew-carroll: Should `activity` be a `WeakReference` as precaution? No opinion on my end, just a thought.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/IntentSender.java
matthew-carroll: I might be missing some context (no pun intended), but why would we not be able to cache both? The app context is always there. It outlives all Activitys. Then an Activity comes and goes.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/MethodCallHandlerImpl.java
```diff
@@ -0,0 +1,132 @@
...
+import android.provider.Settings;
+import android.text.TextUtils;
+import io.flutter.plugin.common.MethodCall;
+import io.flutter.plugin.common.MethodChannel.MethodCallHandler;
+import io.flutter.plugin.common.MethodChannel.Result;
+import java.util.ArrayList;
+import java.util.Map;
+
+/** Forwards incoming {@link MethodCall}s to {@link IntentSender#send}. */
+public class MethodCallHandlerImpl implements MethodCallHandler {
```
matthew-carroll: Given that this is public, would it make sense to specify what kind of method calls this is handling? e.g., `IntentMethodCallHandlerImpl`?

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/MethodCallHandlerImpl.java
```diff
@@ -0,0 +1,132 @@
...
+            ? new ComponentName(packageName, (String) call.argument("component"))
+            : null;
+
+    sender.send(action, flags, category, data, arguments, packageName, componentName);
+
+    result.success(null);
+  }
+
+  private static String convertAction(String action) {
+    switch (action) {
```
matthew-carroll: BTW, not related to the migration, but I've always wondered why we don't make it easy to launch a new `Activity` the way that any Android dev would normally do. It might be nice to at least have a method on the Dart side that simply takes the qualified name of an `Activity` to launch.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/MethodCallHandlerImpl.java
```diff
@@ -0,0 +1,132 @@
...
+    this.sender = sender;
+  }
+
+  /**
+   * Parses the incoming call and forwards it to the cached {@link IntentSender}.
+   *
+   * <p>Always calls {@code result#success}.
+   */
+  @Override
+  public void onMethodCall(MethodCall call, Result result) {
```
matthew-carroll: I'm not sure if annotations are missing by accident, or on purpose, but FYI.

packages/android_intent/android/src/test/java/dev/flutter/plugins/androidintent/MethodCallHandlerImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+  @Before
+  public void setUp() {
+    context = ApplicationProvider.getApplicationContext();
+    sender = new IntentSender(null, null);
+    methodCallHandler = new MethodCallHandlerImpl(sender);
+  }
+
+  @Test
+  public void onMethodCall_doesNothingWhenContextIsNull() {
+    sendLaunchMethodCall("foo", null, null, null, null);
```
matthew-carroll: Opinion: When the details that are relevant to a particular test case are hidden in other methods, it can make it more difficult for developers to understand what is relevant under test vs what is orthogonal to the behavior under test.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/AndroidIntentPlugin.java
mklim: I tried to future proof it a little.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/AndroidIntentPlugin.java
mklim: Ah, right. I was thinking there was some magic here but since I didn't delete the `io.` class this isn't necessary. Deleted.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/AndroidIntentPlugin.java
mklim: We do set `FLAG_ACTIVITY_NEW_TASK` when launching from just the application context, is there something else we should be doing? I'm following the existing logic in the plugin here, which may be buggy.

I'm not 100% sure of the original intent. I'm guessing that it's more supposed to be a safeguard in cases where there is no activity available for whatever reason than a deliberate option for developers.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/IntentSender.java
mklim: I don't think so, but I may be missing something. This shouldn't be leaking because we're always updating the state from the `onDetatched...` callbacks, right? If those handlers aren't totally trustworthy I feel like that's a separate bug that we should find and fix rather than trying to guard against here.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/IntentSender.java
mklim: I was thinking that if the application context changed for some reason we know that any previous activity reference we've set is invalid, because the application context should outlive the activities. But @cyanglaz has a good point that a dev may try to call both and happen to call this first so this is kind of a footgun. I removed the `setActivity(null)` logic from here.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/MethodCallHandlerImpl.java
mklim: I think that's already implied from the package name, but I don't feel super strongly on this if you'd still prefer a rename.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/MethodCallHandlerImpl.java
mklim: Done.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/MethodCallHandlerImpl.java
mklim: Reasonable, looks like flutter/flutter#34296 is the existing feature request for it.

It struck me as weird while I was refactoring this that we don't have any way to get the result of the launch either.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/IntentSender.java
matthew-carroll: Can an application context change? I believe the application context is the `Application` class, which I believe is instantiated exactly one time at process launch. Do you know if that's right?

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/AndroidIntentPlugin.java
matthew-carroll: Ok. Probably needs to be there then. But generally we'd like devs to understand the consequences of launching a new task.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/IntentSender.java
matthew-carroll: Yeah this should be good.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/MethodCallHandlerImpl.java
matthew-carroll: Ok. You may want to unify the approach on the team. The connectivity plugin chose to prefix:
https://github.com/flutter/plugins/pull/2142

Either seems acceptable, but it also seems like one of those details that doesn't have any cause to randomly change between 1P plugins.

packages/android_intent/android/src/main/java/dev/flutter/plugins/androidintent/IntentSender.java
mklim: That sounds roughly right to me, but I'm not confident enough to totally rule it changing out. I'm suspicious of what would happen in a multiprocess app.

mklim:
I updated this PR with the two extra changes we discussed offline: [diff](https://github.com/flutter/plugins/pull/2143/commits/945fd728b05643ed458f5e5a36ba359b13799d11).

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/AndroidIntentPlugin.java
```diff
@@ -1,161 +1,60 @@
...
-        return Settings.ACTION_LOCATION_SOURCE_SETTINGS;
-      case "action_application_details_settings":
-        return Settings.ACTION_APPLICATION_DETAILS_SETTINGS;
-      default:
-        return action;
-    }
+  @Override
+  public void onAttachedToEngine(@NonNull FlutterPluginBinding binding) {
+    sender.setApplicationContext(binding.getApplicationContext());
+    sender.setActivity(null);
```
matthew-carroll: Would it make more senes to set activity to null in the appropriate `ActivityAware` methods? The execution of this method doesn't really imply anything about the Activity...

mklim:
I think this is also ready to merge. There isn't an e2e test, but it would be hard to write one here since the plugin works by launching native Android views. This patch already added some unit tests so it's not untested.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/AndroidIntentPlugin.java
mklim: Would a user potentially be changing the plugin's engine, but expecting it to hold on to an existing activity reference? I was assuming that if the engine changes, the activity itself is also invalid. I also figured that the Activity callbacks should be called anyway so I am updating the Activity there too, but I wanted to add an extra check here just to guarantee that they never got out of state based on some callback issue.

matthew-carroll: You can add an extra call, but if that extra call ever has an effect, it indicates a bug in the embedding. The Activity calls are supposed to be strictly between engine calls:

attach engine
attach activity
detach activity
detach engine

PR #2160, opened 2019-10-07
[package_info] Support the v2 Android embedder (with e2e tests)

collinjackson:
## Description

- Migrate the package_info plugin to use the new embedder.
- Upgraded to AndroidX.
- Added a unit test.
- Added Firebase Test Lab compatible e2e testing.

This PR is intended to serve as an example of how to do same-isolate e2e testing of plugins.

To e2e test low-level plugin APIs using Flutter driver:

```
cd example
flutter drive ../test/<package_name>_e2e.dart
```

To e2e test the example app itself:

```
cd example
flutter drive test/<package_name>_e2e.dart
```

When using flutter driver, to change the binding update the intent filter in AndroidManifest.xml.

To e2e test on a local emulator using Android instrumentation:

```
./gradlew app:connectedAndroidTest -Ptarget=`pwd`/../test_driver/<package_name>_e2e.dart 
```

To e2e test using Firebase test lab, follow the instructions on the [e2e README](https://github.com/flutter/plugins/tree/master/packages/e2e) or let CI [plugin_tools](https://github.com/flutter/plugin_tools) do it for you.

## Known issues

- [ ] Although the test results appear in logcat, the instrumentation adapter isn't exiting at the end of the test)
- [X] instrumentation_adapter package should be renamed e2e (see https://github.com/flutter/plugins/pull/2161)
- [x] Tests that aren't specific to the example app should be moved to the parent folder's test dir

## Related Issues

Fixes https://github.com/flutter/flutter/issues/41843

For a full list of migrations, see https://github.com/flutter/flutter/projects/59


packages/package_info/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,17 @@
                   android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|layoutDirection"
                   android:hardwareAccelerated="true"
                   android:windowSoftInputMode="adjustResize">
+            <!--<intent-filter>-->
```
amirh: nit: I'd just delete this commented out section

packages/package_info/example/pubspec.yaml
```diff
@@ -10,6 +10,7 @@ dependencies:
 dev_dependencies:
   flutter_driver:
     sdk: flutter
+  instrumentation_adapter: "^0.1.3"
```
amirh: We should figure out if this change needs a bump of the minimally required Flutter SDK

packages/package_info/android/gradle.properties
```diff
@@ -1 +1,2 @@
 org.gradle.jvmargs=-Xmx1536M
+android.enableJetifier=true
```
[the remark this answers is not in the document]
collinjackson: Based on the name, I'm hoping that `enableJetifier` does not imply `enableAndroidX`, so it should reduce compilation errors without causing compilation errors.

packages/package_info/example/test/package_info_e2e.dart
```diff
@@ -0,0 +1,46 @@
+import 'dart:io';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:instrumentation_adapter/instrumentation_adapter.dart';
+import 'package:package_info/package_info.dart';
+import 'package:package_info_example/main.dart';
+
+void main() {
+  InstrumentationAdapterFlutterBinding.ensureInitialized();
+
+  testWidgets('fromPlatform', (WidgetTester tester) async {
```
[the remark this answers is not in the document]
collinjackson: Currently the e2e adapter doesn't hook at a low enough level to catch calls to test(), but that could be improved in the future.

packages/package_info/android/gradle.properties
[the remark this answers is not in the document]
mklim: I don't think this has any meaning without `enableAndroidX`, and if it does I think it would also implicitly turn AndroidX on. If AndroidX is enabled and this flag is true, Gradle will attempt to run Jetifier to convert any support references to AndroidX ones. I'm not sure how Gradle handles just this flag with `enableAndroidX` being undefined.

I'm pretty sure that putting this here means that this subproject and only this subproject would be affected, but there may be some weirdness here on account of our build system.

packages/package_info/example/android/app/src/androidTest/java/dev/flutter/plugins/packageinfoexample/MainActivityTest.java
```diff
@@ -0,0 +1,15 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package dev.flutter.plugins.packageinfoexample;
+
+import androidx.test.rule.ActivityTestRule;
+import dev.flutter.plugins.instrumentationadapter.FlutterRunner;
```
amirh: [reminder from offline discussion]: update to the correct package name

packages/package_info/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,13 @@
                   android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|layoutDirection"
                   android:hardwareAccelerated="true"
                   android:windowSoftInputMode="adjustResize">
+        </activity>
+        <activity android:name="dev.flutter.plugins.packageinfoexample.MainActivity"
+                  android:launchMode="singleTop"
```
matthew-carroll: FYI, I made a comment about `singleTop` in the sensor PR.

packages/package_info/example/test_driver/package_info_e2e.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright 2019, the Chromium project authors.  Please see the AUTHORS file
```
cyanglaz: @collinjackson Was it necessary to create these xxx_e2e.dart files for the new e2e test or we can just update the old xxx.dart to the new e2e test style?

collinjackson: CI will only run the tests on Android if they follow the new convention.

packages/package_info/example/android/app/src/main/AndroidManifest.xml
collinjackson: Fixed

packages/package_info/example/android/app/src/androidTest/java/dev/flutter/plugins/packageinfoexample/MainActivityTest.java
collinjackson: Done

packages/package_info/example/pubspec.yaml
collinjackson: Updated it to `flutter: ">=1.6.7 <2.0.0"`

packages/package_info/example/android/app/src/main/java/io/flutter/plugins/packageinfoexample/MainActivity.java
```diff
@@ -4,14 +4,15 @@
...
+import io.flutter.embedding.android.FlutterActivity;
+import io.flutter.embedding.engine.FlutterEngine;
 
 public class MainActivity extends FlutterActivity {
   @Override
-  protected void onCreate(Bundle savedInstanceState) {
-    super.onCreate(savedInstanceState);
-    GeneratedPluginRegistrant.registerWith(this);
+  public void configureFlutterEngine(FlutterEngine flutterEngine) {
+    flutterEngine.getPlugins().add(new PackageInfoPlugin());
```
amirh: nit: add a TODO to remove this once the generated registrant for the v2 embedder is on stable

collinjackson: done, we should probably make this change in the battery plugin also

https://github.com/flutter/plugins/blob/master/packages/battery/example/android/app/src/main/java/io/flutter/plugins/batteryexample/MainActivity.java

PR #2165, opened 2019-10-08
[camera] Migrate to the new embedding

mklim:
## Description

Migrates the camera plugin to the new embedding and adds an example
activity exercising it.

Unlike some of the previous PRs on this, this is the minimal migration
that would land this without adding any tech debt. From previous offline
discussions we'd like to scope these migrations to the bare minimum
required to keep the patches small. See #2114 for a migration combined
with a refactoring.

That said there are still some minor changes to prevent this from adding
additional tech debt, going a little beyond #2118:

- `#onMethodCall` has been split up into its own `MethodCallHandlerImpl`
class to try and keep `CameraPlugin` from mixing concerns. Before the
migration, _all_ it was doing was responding to method calls. Now
there's a significant amount of logic just based around responding to
lifecycle events. It still seemed better to split that into a separate
file than to try and manage that logic all in one place.

Other than the above refactorings, the original logic is untouched.

## Related Issues

Fixes flutter/flutter#41834.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
```diff
@@ -4,172 +4,107 @@
...
-          result.success(null);
-          break;
-        }
-      default:
-        result.notImplemented();
-        break;
+  @Override
+  public void onDetachedFromActivity() {
+    if (methodCallHandler == null) {
+      Log.wtf(TAG, "Detached before initialzed.");
```
cyanglaz: offline discussion:
This is not necessary if assuming `onDetachedFromActivity` is always called after `onAttachedToActivity`
We leave the check here just in case since the v2 embedding is untested.

packages/camera/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,30 +1,39 @@
...
+      android:name=".EmbeddingV1Activity"
+      android:theme="@style/LaunchTheme"
+      android:windowSoftInputMode="adjustResize">
+      <meta-data
+        android:name="io.flutter.app.android.SplashScreenUntilFirstFrame"
+        android:value="true"/>
+    </activity>
+    <activity
+      android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|layoutDirection"
+      android:exported="true"
```
cyanglaz: i think it is supposed to be in the `EmbeddingV1Activity`

packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
```diff
@@ -4,172 +4,107 @@
...
+  /**
+   * Registers a plugin implementation that uses the stable {@code io.flutter.plugin.common}
+   * package.
+   *
+   * <p>Calling this automatically initializes the plugin. However plugins initialized this way
+   * won't react to changes in activity or context, unlike {@link CameraPlugin}.
+   */
   public static void registerWith(Registrar registrar) {
-    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+    if (registrar.activity() == null || Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
```
mehmetf: We should talk about this. We recommend against this pattern because background vs foreground views are not distinguishable in most cases (a background view can switch to foreground).

mklim: Got it, I'll remove this and update the below comment. I saw the comment and noticed that the code didn't match it and assumed that there was a mistake. The code must have been patched at some point and the comment wasn't updated.

packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
```diff
@@ -4,172 +4,107 @@
...
 import io.flutter.plugin.common.PluginRegistry.Registrar;
-import io.flutter.view.FlutterView;
+import io.flutter.plugins.camera.CameraPermissions.PermissionsRegistry;
 import io.flutter.view.TextureRegistry;
 
-public class CameraPlugin implements MethodCallHandler {
+/**
+ * Plugin implementation that uses the {@code io.flutter.embedding} package.
+ *
+ * <p>Instantiate this in an add to app scenario to gracefully handle activity and context changes.
```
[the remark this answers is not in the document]
mklim: Reworded this.

packages/camera/example/android/app/src/main/AndroidManifest.xml
mklim: Done.

packages/camera/android/src/main/java/io/flutter/plugins/camera/CameraPlugin.java
mklim: Done.

mklim: I thought about this a bit more and we also want to keep the logic the same on both the old and new plugin registration path. So I moved this guard into `startListening` and renamed that to `maybeStartListening`.

/cc @cyanglaz this means that now sometimes it is expected that when we call `onDetached` `methodCallHandler` is null after all. So I removed the `Log.wtf` from the guard clause there and replaced it with a comment explaining the known case where that could happen.

packages/camera/example/android/app/src/main/java/io/flutter/plugins/cameraexample/MainActivity.java
```diff
@@ -1,13 +1,22 @@
...
 
 public class MainActivity extends FlutterActivity {
   @Override
-  protected void onCreate(Bundle savedInstanceState) {
-    super.onCreate(savedInstanceState);
-    GeneratedPluginRegistrant.registerWith(this);
+  public void configureFlutterEngine(@NonNull FlutterEngine flutterEngine) {
+    flutterEngine.getPlugins().add(new CameraPlugin());
+
+    ShimPluginRegistry shimPluginRegistry = new ShimPluginRegistry(flutterEngine);
```
matthew-carroll: It might be a good idea to register these first. Given that they're required by CameraPlugin, it is conceivable that CameraPlugin could immediately attempt to utilize them, in which case this would fail. So it looks like an implementation detail that this happens to work. But up to you.

mklim: That makes sense to me, but `GeneratedPluginRegistrant` registers them in this order. I thought it was better to match that.

PerLycke:
@mklim After this merge using the camera plugin on a Note 10+ (and other devices) to stream images is useless. Everything freezes after a couple of seconds moving the camera around. Log shows this:

```
E/CameraCaptureSession( 5749): Session 1: Exception while stopping repeating: 
E/CameraCaptureSession( 5749): android.hardware.camera2.CameraAccessException: CAMERA_ERROR (3): The camera device has encountered a serious error
E/CameraCaptureSession( 5749): 	at android.hardware.camera2.impl.CameraDeviceImpl.checkIfCameraClosedOrInError(CameraDeviceImpl.java:2435)
E/CameraCaptureSession( 5749): 	at android.hardware.camera2.impl.CameraDeviceImpl.stopRepeating(CameraDeviceImpl.java:1126)
E/CameraCaptureSession( 5749): 	at android.hardware.camera2.impl.CameraCaptureSessionImpl.close(CameraCaptureSessionImpl.java:526)
E/CameraCaptureSession( 5749): 	at io.flutter.plugins.camera.Camera.closeCaptureSession(Camera.java:539)
E/CameraCaptureSession( 5749): 	at io.flutter.plugins.camera.Camera.close(Camera.java:545)
E/CameraCaptureSession( 5749): 	at io.flutter.plugins.camera.Camera$2.onError(Camera.java:203)
E/CameraCaptureSession( 5749): 	at android.hardware.camera2.impl.CameraDeviceImpl$CameraDeviceCallbacks.notifyError(CameraDeviceImpl.java:1929)
E/CameraCaptureSession( 5749): 	at android.hardware.camera2.impl.CameraDeviceImpl$CameraDeviceCallbacks.lambda$Sm85frAzwGZVMAK-NE_gwckYXVQ(Unknown Source:0)
E/CameraCaptureSession( 5749): 	at android.hardware.camera2.impl.-$$Lambda$CameraDeviceImpl$CameraDeviceCallbacks$Sm85frAzwGZVMAK-NE_gwckYXVQ.accept(Unknown Source:8)
E/CameraCaptureSession( 5749): 	at com.android.internal.util.function.pooled.PooledLambdaImpl.doInvoke(PooledLambdaImpl.java:258)
E/CameraCaptureSession( 5749): 	at com.android.internal.util.function.pooled.PooledLambdaImpl.invoke(PooledLambdaImpl.java:182)
E/CameraCaptureSession( 5749): 	at com.android.internal.util.function.pooled.OmniFunction.run(OmniFunction.java:77)
E/CameraCaptureSession( 5749): 	at android.os.Handler.handleCallback(Handler.java:873)
E/CameraCaptureSession( 5749): 	at android.os.Handler.dispatchMessage(Handler.java:99)
E/CameraCaptureSession( 5749): 	at android.os.Looper.loop(Looper.java:216)
E/CameraCaptureSession( 5749): 	at android.app.ActivityThread.main(ActivityThread.java:7263)
E/CameraCaptureSession( 5749): 	at java.lang.reflect.Method.invoke(Native Method)
E/CameraCaptureSession( 5749): 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
E/CameraCaptureSession( 5749): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:975)
E/BufferQueueProducer( 5749): [ImageReader-1280x720f23m2-5749-1] dequeueBuffer: attempting to exceed the max dequeued buffer count (9)
```

mklim:
Thanks for reporting the issue and the stack trace @PerLycke, sorry about the regression. Do you mind also posting steps on how to reproduce the crash? A [small test case](https://github.com/flutter/flutter/wiki/Issue-hygiene#provide-reduced-test-cases) would help too if the example app isn't enough to cause it.

PerLycke:
@mklim Never mind my report, after some additional debugging it seems like it's device specific to the Note 10+, and not connected to this commit in particular. Sorry for the confusion, and keep up the good work.

mklim:
@PerLycke no problem, thanks for debugging this further too. Looks like the Note 10+ issue is now being tracked at flutter/flutter#43566.

PR #2168, opened 2019-10-08
Add plugin for Android lifecycle in embedding

amirh:
*Based off @matthew-carroll's https://github.com/flutter/plugins/pull/2129*

Changes in this PR:
- Change the iOS plugin to ObjC.
- Add e2e tests.
- Don't crash the sample app if the lifecycle API isn't available (so it passes on master)
- Cleanup unneeded files.

## Original description
Add plugin for Android lifecycle in embedding.

Embedding side:
flutter/engine#12715


packages/flutter_android_lifecycle/android/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+    }
+    lintOptions {
+        disable 'InvalidPackage'
+    }
+
+    dependencies {
+        implementation "androidx.annotation:annotation:1.1.0"
+        implementation "androidx.lifecycle:lifecycle-runtime:2.0.0-rc01"
+        api 'android.arch.lifecycle:runtime:1.1.1'
```
mklim: These should be changed to the `androidx.` versions.

packages/flutter_android_lifecycle/android/src/main/java/io/flutter/embedding/engine/plugins/lifecycle/FlutterLifecycleAdapter.java
```diff
@@ -0,0 +1,46 @@
...
+
+package io.flutter.embedding.engine.plugins.lifecycle;
+
+import androidx.annotation.NonNull;
+import androidx.lifecycle.Lifecycle;
+import io.flutter.Log;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class FlutterLifecycleAdapter {
```
mklim: nit: Needs documentation.

https://google.github.io/styleguide/javaguide.html#s7.3-javadoc-where-required

packages/flutter_android_lifecycle/example/README.md
```diff
@@ -0,0 +1,16 @@
+# flutter_android_lifecycle_example
```
mklim: optional: I think it's better to delete this completely if we're not going to fill it out.

packages/flutter_android_lifecycle/example/android/app/src/main/java/io/flutter/plugins/flutter_android_lifecycle_example/MainActivity.java
```diff
@@ -0,0 +1,43 @@
...
+      Lifecycle lifecycle = adapter.getLifecycle();
+
+      if (lifecycle == null) {
+        Log.d(TAG, "Couldn't obtained Lifecycle!");
+        // TODO(amirh): make this throw once the lifecycle API is available on stable.
+        // throw new RuntimeException(
+        //     "The FlutterLifecycleAdapter did not correctly provide a Lifecycle instance. Source reference: "
+        //         + flutterPluginBinding.getLifecycle());
+      }
+      Log.d(TAG, "Successfully obtained Lifecycle: " + lifecycle);
```
mklim: nit: This should be behind an `else` if the `throw` is commented out.

packages/flutter_android_lifecycle/example/ios/Runner/AppDelegate.h
```diff
@@ -0,0 +1,6 @@
+#import <Flutter/Flutter.h>
```
mklim: Do we want to commit example app iOS code? Would it be worth ripping it out? It includes some binary files in our checkout that it would be nice to avoid adding.

packages/flutter_android_lifecycle/lib/flutter_android_lifecycle.dart
```diff
@@ -0,0 +1 @@
+
```
mklim: nit: I think it would be worth putting a comment here saying that this is intentionally blank and that this package is meant to distribute Java code.

packages/flutter_android_lifecycle/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+*Note*: This plugin is still under development, and some APIs might not be available yet.
+
+## Installation
+
+Add `flutter_android_lifecycle` as a [dependency in your pubspec.yaml file](https://flutter.io/using-packages/).
+
+## Example
+
+Use a `FlutterLifecycleAdapter` within another Flutter plugin's Android implementation, as shown
```
mklim: I think it's worth explaining here why we want to do this instead of using the engine's lifecycle directly.

packages/flutter_android_lifecycle/android/src/main/java/io/flutter/embedding/engine/plugins/lifecycle/FlutterLifecycleAdapter.java
```diff
@@ -0,0 +1,46 @@
...
+                + reference);
+      }
+
+      Method getLifecycle = reference.getClass().getMethod("getLifecycle");
+      this.lifecycle = (Lifecycle) getLifecycle.invoke(reference);
+    } catch (ClassNotFoundException
+        | NoSuchMethodException
+        | IllegalAccessException
+        | InvocationTargetException e) {
+      Log.w(
```
mklim: If we hit this case, `getLifecycle()` will return null. Should we maybe allow all those exceptions to actually throw instead of catching them?

packages/flutter_android_lifecycle/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+## 1.0.0
```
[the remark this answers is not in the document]
amirh: I think so. We don't expect any significant changes (or added test coverage) here. I'd rather avoid doing a major version bump in the near future just so the version is 1.0.0

packages/flutter_android_lifecycle/LICENSE
```diff
@@ -0,0 +1 @@
+TODO: Add your license here.
```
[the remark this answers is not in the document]
amirh: Thanks for the catch, added the chromium license.

packages/flutter_android_lifecycle/android/build.gradle
amirh: done

packages/flutter_android_lifecycle/example/README.md
amirh: done

packages/flutter_android_lifecycle/example/android/app/src/main/java/io/flutter/plugins/flutter_android_lifecycle_example/MainActivity.java
amirh: done

packages/flutter_android_lifecycle/example/ios/Runner/AppDelegate.h
amirh: Yes, it's important to verify that iOS apps build with this plugin as a dependency.

packages/flutter_android_lifecycle/lib/flutter_android_lifecycle.dart
amirh: done

packages/flutter_android_lifecycle/README.md
amirh: done

packages/flutter_android_lifecycle/android/src/main/java/io/flutter/embedding/engine/plugins/lifecycle/FlutterLifecycleAdapter.java
amirh: I think null returning null is reasonable, especially as we expect that to be the case on stable for awhile.

I added some comments on getLifecycle and changed it to Nullable.

packages/flutter_android_lifecycle/README.md
```diff
@@ -0,0 +1,43 @@
+# Flutter Android Lifecycle Plugin
+
+[![pub package](https://img.shields.io/pub/v/flutter_android_lifecycle.svg)](https://pub.dartlang.org/packages/flutter_android_lifecycle)
+
+A Flutter plugin for Android to allow other Flutter plugins to access an Android `Lifecycle` object
+in the plugin's binding.
+
+The purpose of having this plugin instead of exposing an Android `Lifecycle` object in the engine's
+Android embedder plugins API is to force plugins to have a pub constraint that signifies the
```
matthew-carroll: FYI, the term "embedder" refers to the C++ code. "Embedding" is the name given to the Java/Obj-C layers.
https://github.com/flutter/flutter/wiki/Glossary

packages/flutter_android_lifecycle/android/src/main/java/io/flutter/embedding/engine/plugins/lifecycle/FlutterLifecycleAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+import io.flutter.Log;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class FlutterLifecycleAdapter {
+  private static final String TAG = "FlutterLifecycleAdapter";
+
+  @Nullable private Lifecycle lifecycle = null;
+
+  public FlutterLifecycleAdapter(@NonNull Object reference) {
```
amirh: @mklim @matthew-carroll 

Currently the usage of this API looks like:

```java
      FlutterLifecycleAdapter adapter =
          new FlutterLifecycleAdapter(flutterPluginBinding.getLifecycle());
      Lifecycle lifecycle = adapter.getLifecycle();
```

What do you think about making the `getLifecycle` method static instead, I see how usually we want to avoid static methods for testability, but in this case I don't see how this is particularly useful especially without us adding some `FlutterLifecycleAdapterFactory` method which will potentially allow tests to replace the concrete implementation of FlutterLifecycleAdapter.

Second question - what do yo think about making the adapter be the one that calls `getLifecycle`?

If we do both of these we may end up with the call site looking like:

```java
  Lifecycle lifecycle = FlutterLifecycleAdapter.getLifecycle(flutterPluginBinding);
```

What do you think?

matthew-carroll: I avoid any static call that I don't absolutely need, but it's up to you. Seems like you can also achieve the second goal whether you go static or not.

amirh: > I avoid any static call that I don't absolutely need, but it's up to you.
I'm looking for feedback 😄 do you see any issue with using static here?

> Seems like you can also achieve the second goal whether you go static or not.
Yes, that's why I separated them to discuss separately, what is your opinion on that?

matthew-carroll: I think the second suggestion is fine.

My feedback on the static suggestion is that I never make something static if I don't have to. I would personally not make it static. But I'm also not going to block this PR if you choose to do so. I don't see anything that will break by making it static. Then again, I could say the same thing for many instance methods that aren't static.

mklim: This is totally subjective so I think it's up to you.

My personal opinion on `static` is that it's preferable if the method is truly immutable, doesn't require any kind of state or precondition, and _truly_ doesn't have any kind of side effects. Using `static` in that way is a soft indicator that a method is purely functional. Good examples are `Arrays#asList` or `TextUtils#isEmpty`. A "bad" counter example of `static` is our own FlutterJNI, which has a bunch of "static" methods that don't depend on any member variables but do depend on there being this magic global c library loaded in the program's global runtime. The danger in methods like that is first that it's extremely easy to call them out of sync, and then second that they're really hard to work around when testing.

The generic problem with any static method is that it can't be mocked in unit tests. I generally prefer testing real behavior to using mocks whenever possible so that's not a _huge_ concern to me but it can be a gigantic headache in the moment when the method really does need to be avoided in a test context (again, like `FlutterJNI`). Static also prevents subclasses from overriding the method, but that's a non-issue for me because I generally prefer composition to inheritance.

I think in this specific case it would be fine to make this static, because it's basically just returning the object that's passed into it. In a test context if you really needed to mock anything you would be returning a mock from the input parameter. I again prefer static interfaces because they're a soft signal that the function doesn't have silent side effects so I'd be happy to use one here, but it's totally subjective.

packages/flutter_android_lifecycle/README.md
amirh: done

packages/flutter_android_lifecycle/android/src/main/java/io/flutter/embedding/engine/plugins/lifecycle/FlutterLifecycleAdapter.java
amirh: done

packages/flutter_android_lifecycle/example/android/app/src/main/java/io/flutter/plugins/flutter_android_lifecycle_example/MainActivity.java
amirh: (the change was returning early)

mklim: Whoops, my mistake.

packages/flutter_android_lifecycle/android/src/main/java/io/flutter/embedding/engine/plugins/lifecycle/FlutterLifecycleAdapter.java
amirh: updated to static, I think there's an upside (less cluttered call site) and no downside.

amirh:
Will land on green

amirh:
Actually I'll wait for another LGTM from @matthew-carroll and @mklim as the public API changed

matthew-carroll:
LGTM

PR #2190, opened 2019-10-15
[e2e] Update to support new embedder

collinjackson:
Required in order to end-to-end test migrated plugins.


collinjackson:
Yes, I've updated the example app, this plugin can actually e2e test itself.

PR #2192, opened 2019-10-15
[in_app_purchase] Fix iOS build warning

jmagman:
## Description

- Fix mutable collection being stored immutably.

Slight cleanup of InAppPurchasePlugin since I was editing it anyway:
- Decorated designated initializer to initialize mutable collection ivars on init, then remove (unthreadsafe) getters in favor of synthesized property getters.
- Set ivars instead of properties in initializers.
- Make category properties readonly.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/42679.

## Checklist
- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
+++ b/packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
@@ -54,26 +54,25 @@
   _messenger = [registrar messenger];
 
   __weak typeof(self) weakSelf = self;
-  _paymentQueueHandler =
-      [[FIAPaymentQueueHandler alloc] initWithQueue:[SKPaymentQueue defaultQueue]
-          transactionsUpdated:^(NSArray<SKPaymentTransaction *> *_Nonnull transactions) {
-            [weakSelf handleTransactionsUpdated:transactions];
-          }
-          transactionRemoved:^(NSArray<SKPaymentTransaction *> *_Nonnull transactions) {
-            [weakSelf handleTransactionsRemoved:transactions];
-          }
-          restoreTransactionFailed:^(NSError *_Nonnull error) {
-            [weakSelf handleTransactionRestoreFailed:error];
-          }
-          restoreCompletedTransactionsFinished:^{
-            [weakSelf restoreCompletedTransactionsFinished];
-          }
-          shouldAddStorePayment:^BOOL(SKPayment *payment, SKProduct *product) {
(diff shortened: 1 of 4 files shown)


packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -17,7 +17,7 @@ @interface InAppPurchasePlugin ()
 
 // After querying the product, the available products will be saved in the map to be used
 // for purchase.
-@property(copy, nonatomic) NSMutableDictionary *productsCache;
+@property(strong, nonatomic) NSMutableDictionary *productsCache;
```
jmagman: This is the actual fixed warning, everything else in this file is cleanup.
```
warning: Property of mutable type 'NSMutableDictionary' has 'copy' attribute; an immutable object
      will be stored instead
```

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -40,19 +40,21 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
 }
 
 - (instancetype)initWithReceiptManager:(FIAPReceiptManager *)receiptManager {
-  self = [self init];
-  self.receiptManager = receiptManager;
+  self = [super init];
+  _receiptManager = receiptManager;
+  _requestHandlers = [NSMutableSet new];
+  _productsCache = [NSMutableDictionary new];
```
cyanglaz: curious, what is the benefit of initialize them in at initialization v.s. the lazy init that we used to have?

jmagman: In this particular case the lazy getter wasn't threadsafe.  Also this class assumes these collections are instantiated, but there's nothing stopping a subclass from overriding the getter to do something wonky that doesn't call super or set up the ivar (probably unlikely).  To me it looked like it was lazily instantiated because there was no designated initializer decoration on the init methods, so there was no one canonical spot where these collections could be relied on to be instantiated.

Assuming this object is queue confined (I don't know if that's actually enforced or just convention) it's fine to keep it lazily instantiated, and I can change it back if you prefer!

cyanglaz: Thanks for the explanation. What you have now is good! LGTM again.

PR #2193, opened 2019-10-15
[android_alarm_manager] migrate to the V2 Android embedding

bkonyi:
## Description

Minimum work to migrate to the Android v2 embedding.

## Related Issues

flutter/flutter#41830

## Checklist

- [x] I signed the [CLA].

## Breaking Change


bkonyi:
@matthew-carroll is there anything else here that I missed that's specific to background execution? Also, do we still think `onViewDestroyed` is necessary or is there a new equivalent for `FlutterEngine`?

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
```diff
@@ -47,54 +54,71 @@
    * connected to, and running against, the associated Flutter execution context.
    */
   public static void registerWith(Registrar registrar) {
-    // alarmManagerPluginChannel is the channel responsible for receiving the following messages
+    final AndroidAlarmManagerPlugin plugin = new AndroidAlarmManagerPlugin();
+    // Listen for FlutterView destruction so that this plugin can move itself
+    // to background mode.
+    registrar.addViewDestroyListener(plugin);
```
matthew-carroll: @bkonyi do you know what the analogous behavior of this plugin is when using the new embedding? I assume that what this plugin is doing when the view is destroyed is important, right? If so, we won't be able to act on view destruction, but plugins will now have the opportunity to take action when the activity disappears, which is probably what this view destruction really signified in the first place...

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
```diff
@@ -38,7 +40,12 @@
  *       Dart is ready to execute tasks.
  * </ol>
  */
-public class AndroidAlarmManagerPlugin implements MethodCallHandler, ViewDestroyListener {
+public class AndroidAlarmManagerPlugin
+    implements FlutterPlugin, MethodCallHandler, ViewDestroyListener {
+  private Context mContext;
+  private MethodChannel mAlarmManagerPluginChannel;
```
mklim: Readability nit: Google Java style doesn't use `m` prefixes on members. https://google.github.io/styleguide/javaguide.html#s5.2.5-non-constant-field-names

(I think Android Style does, or at least did.)

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
mklim: +1

bkonyi: Actually, looking at this code now I'm uncertain as to what it even accomplishes. It appears that on view destruction we try and store the view in `AlarmService`... but we also create a view in `AlarmService.startBackgroundIsolate`. I _think_ we're safe to remove this... I'll try and see if it causes any issues.

bkonyi:
> Right now this plugin has no unit tests or integration tests. I think that this android_alarm_manager should be straightforward to integration test using the e2e plugin. Would you be open to adding that?
> 
> I wrote some example tests that I think should work as a starting point. Feel free to copy these into the PR.
> 
> https://github.com/collinjackson/plugins/tree/android_alarm_manager_test/packages/android_alarm_manager/example/test_driver

I'm definitely open to adding tests if we can show they're not going to be extremely flaky. The `AlarmManager` alarms aren't as predictable as I'd like, but I'll see what I can do.

> Note: these tests do not pass for me when I run them under `flutter drive` (the alarm manager doesn't seem to trigger), and I'm not sure why. Maybe the alarm manager only works when the app is in the background?

These tests run on an actual device, correct? There might be an issue with the phone being in doze mode which results in background events being delayed until the device is in a more "active" state to conserve battery life.

packages/android_alarm_manager/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.4.5
+* Add support for Flutter Android embedding V2
+* Called WidgetsFlutterBinding.ensureInitialized() in
```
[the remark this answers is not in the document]
bkonyi: Fair point. I can remove this from the initialize method, but is there any reason why we're leaving this up to the plugin user?

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
bkonyi: Done.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
```diff
@@ -152,15 +152,6 @@ public static void setCallbackDispatcher(Context context, long callbackHandle) {
     prefs.edit().putLong(CALLBACK_HANDLE_KEY, callbackHandle).apply();
   }
 
-  public static boolean setBackgroundFlutterView(FlutterNativeView view) {
```
matthew-carroll: Notice that this isn't really setting a `View`, it's setting a `FlutterNativeView`, which is now essentially represented by `FlutterEngine`. Are you sure this method can be deleted? Where is this behavior handled now?

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
```diff
@@ -147,21 +166,6 @@ public void onMethodCall(MethodCall call, Result result) {
...
-   * background execution.
-   *
-   * <p>Invoked when the {@link FlutterView} connected to the given {@link FlutterNativeView} is
-   * destroyed.
-   *
-   * <p>Returns true if the given {@code nativeView} was successfully stored by this plugin, or
-   * false if a different {@link FlutterNativeView} was already registered with this plugin.
-   */
-  @Override
-  public boolean onViewDestroy(FlutterNativeView nativeView) {
```
matthew-carroll: This still seems like it could be relevant. I think that view destruction is actually activity detachment. The question is, should the plugin hand over its FlutterEngine to the AlarmService when the Activity goes away? I'm not sure. But this plugin probably requires some careful consideration to its use-cases with the new embedding.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
bkonyi: I think this is an artifact from the original implementation of this plugin when we were running all callbacks on the main isolate, which meant we had to be able to handle the case when the application was moved from foreground to background. This plugin now instead starts a background isolate immediately when `AndroidAlarmManager.initialize` is invoked or when the `AlarmService` is started via an `Intent`, and this isolate is kept alive until the service shuts down.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AndroidAlarmManagerPlugin.java
bkonyi: It may have been relevant early on in this plugin's life, but I don't believe it is anymore. See my response to your other comment.

collinjackson:
> @collinjackson it looks like driver tests don't work correctly when additional isolates are spawned (see [flutter/flutter#24703](https://github.com/flutter/flutter/issues/24703)). It looks like isolates are being paused on start and the driver only resumes the main isolate, regardless of how the test is launched (both `flutter drive` and the Gradle e2e) so the background isolate for the alarm manager never actually runs.

I'm confused -- we're not using Flutter driver, so it seems like the Gradle e2e method should work. I wonder if @vishna's workaround in https://github.com/flutter/flutter/issues/24703#issuecomment-531277291 would work here.

bkonyi:
> > @collinjackson it looks like driver tests don't work correctly when additional isolates are spawned (see [flutter/flutter#24703](https://github.com/flutter/flutter/issues/24703)). It looks like isolates are being paused on start and the driver only resumes the main isolate, regardless of how the test is launched (both `flutter drive` and the Gradle e2e) so the background isolate for the alarm manager never actually runs.
> 
> The Gradle e2e method doesn't use Flutter driver, so I'm surprised it doesn't work. I wonder if @vishna's workaround in [flutter/flutter#24703 (comment)](https://github.com/flutter/flutter/issues/24703#issuecomment-531277291) would be helpful here.

Disregard, for some reason the background callback dispatcher wasn't calling `WidgetsFlutterBinding.ensureInitialized` so the isolate was started but was unable to communicate that back to the plugin. Also, that workaround didn't seem to work before, but I'll give it another shot to see if I can get this working with `flutter drive` as well.

However, I am still running into some issues with plugins seemingly not being recognized when running tests. It looks like the alarms are firing correctly but we're failing to write the count to the temporary file. We were swallowing the exception before, but I'm seeing this failure:

`MissingPluginException(No implementation found for method getTemporaryDirectory on channel plugins.flutter.io/path_provider)`

I'm also seeing this warning in the logs: `Warning: E2E test plugin was not detected.`

I've tried cleaning the project and re-running and haven't had any luck. The `GeneratedPluginRegistrant` looks like it's setup correctly, so I'm uncertain as to what's going on here...

packages/android_alarm_manager/example/android/app/src/main/java/io/flutter/plugins/androidalarmmanagerexample/MainActivity.java
```diff
@@ -1,15 +1,16 @@
...
 
 public class MainActivity extends FlutterActivity {
-  public static final String TAG = "AlarmExampleMainActivity";
-
   @Override
-  protected void onCreate(Bundle savedInstanceState) {
-    super.onCreate(savedInstanceState);
-    GeneratedPluginRegistrant.registerWith(this);
+  public void configureFlutterEngine(FlutterEngine flutterEngine) {
+    flutterEngine.getPlugins().add(new AndroidAlarmManagerPlugin());
```
collinjackson: As discussed you'll want to register the E2E plugin and path provider here.

bkonyi: Done.

xster:
@bkonyi, @collinjackson, anything blocking this PR currently?

bkonyi:
> @bkonyi, @collinjackson, anything blocking this PR currently?

Ah sorry for the delay. I think the migration is mostly done, but I'll need to upload my most recent changes. However, as discussed with @matthew-carroll and @mklim we're going to delay publishing this until the next stable is released to handle registering plugins with the background isolate automatically.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/BackgroundExecutionContext.java
```diff
@@ -0,0 +1,204 @@
...
+   *       PluginRegistrantException} will be thrown.
+   * </ul>
+   */
+  public void startBackgroundIsolate(Context context, long callbackHandle) {
+    if (backgroundFlutterEngine != null) {
+      Log.e(TAG, "Background isolate already started");
+      return;
+    }
+
+    FlutterMain.ensureInitializationComplete(context, null);
```
matthew-carroll: On master this call should no longer be needed.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
```diff
@@ -27,192 +22,64 @@
...
-      sPluginRegistrantCallback.registerWith(sBackgroundFlutterView.getPluginRegistry());
-    }
+    assert (backgroundExecutionContext == null);
+    backgroundExecutionContext = new BackgroundExecutionContext();
+    backgroundExecutionContext.startBackgroundIsolate(context, callbackHandle);
   }
 
   /**
-   * Called once the Dart isolate ({@code sBackgroundFlutterView}) has finished initializing.
+   * Called once the Dart isolate ({@code backgroundFlutterView}) has finished initializing.
```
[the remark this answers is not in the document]
bkonyi: No, this was an oversight. Thought I had caught all the instances of `backgroundFlutterView` in the code base... will remove.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
```diff
@@ -27,192 +22,64 @@
...
    *
    * <p>Invoked by {@link AndroidAlarmManagerPlugin} when it receives the {@code
    * AlarmService.initialized} message. Processes all alarm events that came in while the isolate
    * was starting.
    */
-  // TODO(mattcarroll): consider making this method package private
-  public static void onInitialized() {
+  static void onInitialized() {
```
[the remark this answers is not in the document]
bkonyi: Ah I wasn't aware. I'll go ahead and do that.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
```diff
@@ -27,192 +22,64 @@
...
-    sIsIsolateRunning.set(true);
-    synchronized (sAlarmQueue) {
+    synchronized (alarmQueue) {
       // Handle all the alarm events received before the Dart isolate was
       // initialized, then clear the queue.
-      Iterator<Intent> i = sAlarmQueue.iterator();
+      Iterator<Intent> i = alarmQueue.iterator();
       while (i.hasNext()) {
-        executeDartCallbackInBackgroundIsolate(i.next(), null);
+        backgroundExecutionContext.executeDartCallbackInBackgroundIsolate(i.next(), null);
```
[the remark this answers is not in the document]
bkonyi: This does actually allow for invoking multiple callbacks on an isolate, but that's only due to the fact that the isolate's entrypoint is a "callback dispatcher". The isolate only enters at a single entry point which sets up a method channel which this method communicates over. That method call handler invokes callbacks based on a provided callback, allowing for multiple callbacks to be handled on a single background isolate.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
```diff
@@ -27,192 +22,64 @@
...
-  }
-
-  public static boolean setBackgroundFlutterView(FlutterNativeView view) {
-    if (sBackgroundFlutterView != null && sBackgroundFlutterView != view) {
-      Log.i(TAG, "setBackgroundFlutterView tried to overwrite an existing FlutterNativeView");
-      return false;
-    }
-    sBackgroundFlutterView = view;
-    return true;
+    BackgroundExecutionContext.setCallbackDispatcher(context, callbackHandle);
```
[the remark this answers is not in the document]
bkonyi: For this plugin there's only one valid callback dispatcher, so it doesn't really make sense to make this per-instance state.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/BackgroundExecutionContext.java
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * The {@link MethodChannel} that connects the Android side of this plugin with the background
+   * Dart isolate that was created by this plugin.
+   */
+  private MethodChannel backgroundChannel;
+
+  private FlutterEngine backgroundFlutterEngine;
+
+  private AtomicBoolean isIsolateRunning = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
bkonyi: Yes, this is to ensure that the method channel in the background isolate has been initialized. If we were to use the logic you've suggested we'd run into races where we try to send callback invocations over the method channel before it's been registered.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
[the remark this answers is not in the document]
matthew-carroll: What if this plugin is being used by 2+ isolates/`FlutterEngine`s? Is there still just 1 dispatcher across multiple isolates?

bkonyi:
> LGTM with possibly one remaining question that I left.
> 
> This plugin is relatively complicated compared to others, and delves into areas of Android that are very difficult to reason about, so I'll just add that beyond an LGTM, I encourage us to exercise this plugin wherever, whenever, and however we can. If any bugs exist, we're likely to only realize it with a stacktrace.

FYI, I've got changes addressing your earlier comments that are a WIP but I won't get to them until I'm physically in the office next week (I'm not setup for Android dev on my MacBook at the moment). I'll do my best to fix up documentation to make this plugin's behavior as clear as possible.

I'm wondering what would be the best way to exercise this plugin other than the rather simplistic tests this PR adds. If we had more official plugins supporting background execution, I'd imagine that looking into having a dedicated device running applications triggering background execution events would be something worth doing, but other than that I'm not sure what we can do.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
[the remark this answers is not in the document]
bkonyi: Correct, there's one dispatcher across multiple isolates. All callbacks will execute on the single isolate spawned by this plugin, regardless of which isolate they were scheduled on.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/BackgroundExecutionContext.java
bkonyi: Is this still required for apps using the V1 embedding?

matthew-carroll: Yeah I think so. In short, if a Flutter experience is based on a `FlutterEngine` then that call is not needed. But if it's based on `FlutterNativeView` then it is.

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
```diff
@@ -449,15 +316,11 @@ public static void reschedulePersistentAlarms(Context context) {
...
-
-    Context context = getApplicationContext();
-    FlutterMain.ensureInitializationComplete(context, null);
-
-    if (!sIsIsolateRunning.get()) {
-      SharedPreferences p = context.getSharedPreferences(SHARED_PREFERENCES_KEY, 0);
-      long callbackHandle = p.getLong(CALLBACK_HANDLE_KEY, 0);
-      startBackgroundIsolate(context, callbackHandle);
+    if (backgroundExecutionContext == null) {
+      backgroundExecutionContext = new BackgroundExecutionContext();
```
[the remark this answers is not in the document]
bkonyi: Instantiation in `onCreate` is to cover the case where the application is woken up in the background via an intent to process an alarm event. In this case, we follow a slightly different initialization path that loads the callback handle from `SharedPreferences`

packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/AlarmService.java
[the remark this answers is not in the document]
matthew-carroll: Interesting. Could that cause unintended consequences? @xster

bkonyi:
@collinjackson it looks like the tests are failing, but I'm not sure why... is the test being run with `flutter test` on Cirrus?

xster:
Seems like the error on firebase test lab is 

```
Permission Denial: starting instrumentation ComponentInfo{io.flutter.plugins.androidalarmmanagerexample.test/androidx.test.runner.AndroidJUnitRunner} from pid=18303, uid=18303 not allowed because package io.flutter.plugins.androidalarmmanagerexample.test does not have a signature matching the target io.flutter.plugins.androidalarmmanagerexample
```

bkonyi:
So it looks like the timeout on Firebase is too aggressive for the e2e test to pass on API 28+ and we timeout. If you look at the logcat output, we do see an `incrementCounter: 1` in the logs after the test has been cancelled.

@collinjackson @amirh, any suggestions?

collinjackson:
> So it looks like the timeout on Firebase is too aggressive for the e2e test to pass on API 28+ and we timeout. If you look at the logcat output, we do see an `incrementCounter: 1` in the logs after the test has been cancelled.
> 
> @collinjackson @amirh, any suggestions?

I'm in favor of longer (5m?) timeouts to reduce flakiness if that actually causes the test to pass. We could also make the timeouts configurable per plugin but maybe that's overthinking it.

bkonyi:
Finally got all the checks passing... Any objections to me landing this?

vkammerer:
Really looking forward to the publication of this update!

PR #2197, opened 2019-10-15
Enable testing on both master and stable

collinjackson:
## Description

This PR runs the Cirrus tests on both master and stable for most types of tests, with the exception of formatting and analyzer. The "build all plugins" test only being run on master, but that should be fixed in the future (flutter/flutter#42810).

## Related Issues

Fixes flutter/flutter#42377


collinjackson:
This change doesn't affect CI time since the tasks run in parallel. It will however double our budget.

PR #2203, opened 2019-10-16
Re-land testing of plugins on stable

collinjackson:
Web plugins are now suppressed when testing on stable.

Filed https://github.com/flutter/flutter/issues/42864 for the eventual fix

Re-lands flutter/plugins#2197 that was reverted in flutter/plugins#2199

Fixes https://github.com/flutter/flutter/issues/42810
Fixes flutter/flutter#42377


.cirrus.yml
```diff
@@ -74,7 +101,14 @@ task:
...
           PLUGIN_SHARDING: "--shardIndex 2 --shardCount 4"
           PLUGIN_SHARDING: "--shardIndex 3 --shardCount 4"
+        matrix:
+          CHANNEL: "master"
+          CHANNEL: "stable"
         SIMCTL_CHILD_MAPS_API_KEY: ENCRYPTED[596a9f6bca436694625ac50851dc5da6b4d34cba8025f7db5bc9465142e8cd44e15f69e3507787753accebfc4910d550]
       build_script:
+        # TODO(jackson): Allow web plugins once supported on stable
+        # https://github.com/flutter/flutter/issues/42864
+        - if [ "CHANNEL" -eq "stable" ]; do find . | grep _web$ | xargs rm -rf
```
amirh: Is the matrix ordering guaranteed?

collinjackson: it's a shard, so it won't run at all on master

PR #2214, opened 2019-10-18
Revert "[Connectivity] migrate to the new android embedding"

cyanglaz:
Reverts flutter/plugins#2142

Will create a new PR to support stable instead of 1.6.7


cyanglaz:
Closing since the version is not the issue. Fixed the real issue in https://github.com/flutter/plugins/pull/2212

PR #2233, opened 2019-10-25
[e2e] update README

collinjackson:
Updates README with some documentation updates.

### DIFF
--- a/packages/e2e/README.md
+++ b/packages/e2e/README.md
@@ -1,7 +1,7 @@
 # e2e
 
 This package enables self-driving testing of Flutter code on devices and emulators.
-It can adapt the test results in a format that is compatible with `flutter drive`
+It adapts flutter_test results into a format that is compatible with `flutter drive`
 and native Android instrumentation testing.
 
 iOS support is not available yet, but is planned in the future.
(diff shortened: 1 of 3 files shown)


packages/e2e/README.md
```diff
@@ -27,12 +27,19 @@ void main() {
...
+`example/test/`. It is also acceptable to put e2e tests in `test_driver/` folder
+so that they're alongside the runner app (see below).
+
 ## Using Flutter driver to run tests
 
 `E2EWidgetsTestBinding` supports launching the on-device tests with `flutter drive`.
 Note that the tests don't use the `FlutterDriver` API, they use `testWidgets` instead.
 
-Put the a file named `<package_name>_test.dart` in the app' `test_driver` directory:
+Put the a file named `<package_name>_e2e_test.dart` in the app' `test_driver` directory:
```
digiter: Does `flutter test` scans `_test.dart` in a directory and execute them?

packages/e2e/README.md
```diff
@@ -106,22 +113,27 @@ dependencies {
 To e2e test on a local Android device (emulated or physical):
 
 ```
-./gradlew connectedAndroidTest -Ptarget=`pwd`/../test_driver/<package_name>_e2e.dart
+./gradlew app:connectedAndroidTest -Ptarget=`pwd`/../test_driver/<package_name>_e2e.dart
```
digiter: The suffix should be `_e2e.dart` or `_e2e_test.dart` as above?

packages/e2e/README.md
collinjackson: Yes. It only looks in the test/ directory by default.

packages/e2e/README.md
collinjackson: The suffix here should be _e2e.dart. The _e2e_test.dart suffix is reserved for the driver runner (which I hope will be unnecessary in the future).

packages/e2e/README.md
digiter: Then when executing `flutter test` without an attached device, the `_e2e_test.dart` will execute. Is it expected?

collinjackson: `flutter test` only looks for tests in the test/ folder. test_driver/ is ignored.

collinjackson:
Merging this since you LGTM'd it but if you'd like to make more edits let me know.

PR #2236, opened 2019-10-25
Use package import to import files inside lib/ directory.

chingjun:
This is causing issues intenally.

This is also the recommended way in Dart guidelines: https://dart.dev/guides/libraries/create-library-packages#importing-library-files

### DIFF
--- a/packages/url_launcher/url_launcher_platform_interface/test/method_channel_url_launcher_test.dart
+++ b/packages/url_launcher/url_launcher_platform_interface/test/method_channel_url_launcher_test.dart
@@ -6,8 +6,8 @@ import 'package:mockito/mockito.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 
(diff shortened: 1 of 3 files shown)


packages/url_launcher/url_launcher_platform_interface/test/method_channel_url_launcher_test.dart
```diff
@@ -6,8 +6,8 @@ import 'package:mockito/mockito.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 
-import '../lib/method_channel_url_launcher.dart';
-import '../lib/url_launcher_platform_interface.dart';
+import 'package:url_launcher_platform_interface/method_channel_url_launcher.dart';
```
amirh: interesting, Android Studio kept replacing the package path with '../' when I was editing this file.

chingjun:
Done, PTAL. Thanks.

PR #2248, opened 2019-11-01
[in_app_purchase] Added serviceTimeout code for google iab

NikitaZhelonkin:
Fix for issue https://github.com/flutter/flutter/issues/43969


cyanglaz:
@NikitaZhelonkin Thanks for spotting this. Could you reflect the change in the generated code as well? The steps to do so is mentioned in the comments above your change as well as in README.
Also we would need some tests before we can landing this.
Finally we would also need to update the CHANGELOG and pubspec with a new version.

NikitaZhelonkin:
@cyanglaz I had already added changes to generated code

cyanglaz:
> @cyanglaz I had already added changes to generated code

Thanks a lot. We would also need tests and version update for this to be landed. Please  let me know if you need any help :)

NikitaZhelonkin:
Am I need add test after adding only one enum value?

cyanglaz:
@NikitaZhelonkin Technically, we require all new PRs (If adding features or fixing bugs) to have tests. And I think a test for the enum case is pretty straight forward to add? Maybe we can just test if the json value for `serviceTimeout` is -3?

cyanglaz:
Closing this as https://github.com/flutter/plugins/pull/3287 has been landed and issue resolved.

PR #2250, opened 2019-11-01
Run the publish with the pub version from flutter stable

amirh:


.cirrus.yml
```diff
@@ -13,7 +13,9 @@ task:
   activate_script: pub global activate flutter_plugin_tools
   matrix:
     - name: publishable
-      script: ./script/check_publish.sh
+      script:
+        - flutter channel stable
```
mklim: I think it makes more sense to alter `check_publish` to bail if it detects that it's being run on the master channel, or skip the script here if it's possible for us to check it. This will run the test twice and also add in an extra step where we downgrade flutter and re-download all its dependencies. Could be missing something but I don't think there's useful information out of running this twice on the same channel.

mklim: Never mind, I can't read whitespace indentation. :) The MATRIX below doesn't apply to this script.

cyanglaz:
@amirh Why did we decide not to run publish with master?

mklim:
@cyanglaz at the time this PR was merged master had publish warnings that we were failing, if I remember correctly because of the new federated plugin format. Also if I remember correctly there wasn't a way to satisfy the warnings and also stay backwards compatible with stable. Since there's been a new stable roll relatively recently I don't think this still applies, but not 100% sure.

cyanglaz:
@mklim Thanks!I would like to confirm that and bring back the publish check on master. I guess you are right, I have a vague memory of this too.

PR #2251, opened 2019-11-01
Shared Preferences Plugin support for App Group (iOS only feature)

luismarquesfreire:
## Description

This PR will allow setting a domain name / group (e.g. group.com.example.myap) so it is possible to write user defaults to the App Group instead of just the main app scope. This is particularly useful for implementing extensions, like the share extension, where you need o share certain information with the extension (e.g. uid / token etc).

## Related Issues

As of today, I found no related issues.


luismarquesfreire:
@googlebot I signed it!

packages/shared_preferences/lib/shared_preferences.dart
```diff
@@ -152,6 +152,15 @@ class SharedPreferences {
     return await _kChannel.invokeMethod<bool>('clear');
   }
 
+  /// Set App Domain / Group
+  /// This method only works for iOS and does nothing in Android
+  Future<void> setDomain(String domain) async {
```
Zazo032: If this only works on iOS, should we throw when using it in Android with `throw UnsupportedError('Functionality not available on Android');`? (just like other plugins do).

```suggestion
  Future<void> setDomain(String domain) async {
    if (Platform.isAndroid) {
      throw UnsupportedError('Functionality not available on Android');
    }
    final Map<String, dynamic> params = <String, dynamic>{
      'domain': domain,
    };
    _kChannel.invokeMethod<void>('setDomain', params);
  }
```

luismarquesfreire: I wouldn't do that because its not really an error, it just gets ignored by Android but you can still use setDomain without need to check if you're on iOS.

packages/shared_preferences/shared_preferences/ios/Classes/FLTSharedPreferencesPlugin.m
```diff
@@ -8,6 +8,8 @@
 
 @implementation FLTSharedPreferencesPlugin
 
+NSString *customDomain;
```
gaaclarke, line 11: This is a global variable, those won't be be accepted, you need to move the to be an instance variable.

packages/shared_preferences/shared_preferences_macos/pubspec.yaml
```diff
@@ -14,7 +14,8 @@ environment:
   flutter: ">=1.12.8 <2.0.0"
 
 dependencies:
-  shared_preferences_platform_interface: ^1.0.0
+  shared_preferences_platform_interface:
```
gaaclarke, line 17: FYI, you aren't going to be able to commit with relative path dependencies.

stuartmorgan:
Closing, as this seems to be a WIP (e.g., merge conflict markers in the diffs, unpublishable dependencies), and there hasn't been a response to the last review. If you are still interested in moving forward with this change, feel free to re-open the PR once it's updated.

PR #2255, opened 2019-11-05
[video_player] fix: #40429

p30arena:
resolves this issue: https://github.com/flutter/flutter/issues/40429

## Description

on iOS:
1. The local file path is not encoded correctly, so I had to replace spaces with %20
2. The local file path must have an extension, say ".mp4"

## Related Issues

*Replace this paragraph with a list of issues related to this PR from the [issue database](https://github.com/flutter/flutter/issues). Indicate, which of these issues are resolved or fixed by this PR. Note that you'll have to prefix the issue numbers with flutter/flutter#.*

## Checklist

- [x] I signed the [CLA].

## Breaking Change


ened:
Hello @p30arena - thank you for the PR. Can you think of a way to test it in the example/test_driver (as it seems to be iOS platform specific)? That would surely help to get it landed.

p30arena:
> Hello @p30arena - thank you for the PR. Can you think of a way to test it in the example/test_driver (as it seems to be iOS platform specific)? That would surely help to get it landed.

@mklim @ened 

is it necessary to add tests?
it seems that android handles the encoding of the Uri itself, but iOS leaves it to the programmer.
I rathered to encode it on the dart side, so the debug process will be easier in the future.

ened:
hi @p30arena , there is no problem in encoding the URI on the Dart side. If anything, it will be helpful to client programmers to see what a Flutter plugin is doing.

However, we do need tests. No matter how small the change is, it changes plugin behaviour.
We can easily ramp up a lot of untested behaviour changes if we leave such things untested. Therefore, let's stay on the safe side and add a test. There should also be a line of documentation to explain the URL encoding behaviour.

It may be enough to write a widget test (and listen to the URLs send to the video player method channel).

Can you think of any case where the URL encoding may not be desired?

p30arena:
> hi @p30arena , there is no problem in encoding the URI on the Dart side. If anything, it will be helpful to client programmers to see what a Flutter plugin is doing.
> 
> However, we do need tests. No matter how small the change is, it changes plugin behaviour.
> We can easily ramp up a lot of untested behaviour changes if we leave such things untested. Therefore, let's stay on the safe side and add a test. There should also be a line of documentation to explain the URL encoding behaviour.
> 
> It may be enough to write a widget test (and listen to the URLs send to the video player method channel).
> 
> Can you think of any case where the URL encoding may not be desired?

@ened 
File Paths don't require Uri Encoding, so The user is passing the correct path.
File Path must end with an extension, so I've included an assert.

I can't think of a proper way to write a test for The File Path Passed By The User/Developer.

We just have to make sure that on the native side, we don't encode the path.
Can I write a test for that?

ened:
> @ened
> File Paths don't require Uri Encoding, so The user is passing the correct path.

Based on the code submitted, the URI encoding happens for iOS App.

> File Path must end with an extension, so I've included an assert.

I think the assert is good. 

> I can't think of a proper way to write a test for The File Path Passed By The User/Developer.
>
> We just have to make sure that on the native side, we don't encode the path.
> Can I write a test for that?

I think the key test is to check what is transferred from Dart to Native.

So, your tests would try to use a MockPlatform (path_provider should have a example, or see https://github.com/flutter/plugins/blob/b76fe5d58024f479df4057920289c04277c30632/packages/android_intent/test/android_intent_test.dart) for `FakePlatform`, then check if the value is checked on Dart (Assetions) and also the transformation (URI encoding) when it is sent from Dart to Native.

The tests should cover:
 - on iOS: 'requires the path to contain a file extension'
 - on iOS: 'requires the path to be URI encoded'

And you can write a test (in dart only, no need for a driver implementation I think) to check those 2 points.

As a side note: Could we also encode the URI for the Android side? What happens if we encode the URI on Dart side and pass to Android? Will Android "double encode"? If not, could you remove the if/else for that and just encode every time?

cbenhagen:
@p30arena thank you for your contribution! Unfortunately the structure of the plugin has changed lately to support implementations on more platforms. Mainly the plugin was moved to a new subdirectory and now uses a common platform interface. Adapting your PR to the new structure should be straight forward. If you have questions please ask!

E2E tests are currently skipped on iOS but I think we could at a test for this already to https://github.com/flutter/plugins/blob/master/packages/video_player/video_player/example/test_driver/video_player_e2e.dart.

cbenhagen:
Oh, do you mind changing the title to something more descriptive?

stuartmorgan:
Thanks for the contribution. Since there hasn't been any update on adding the tests described above in over a year, I'm going to close this to get it off our our review queue, but if you are still interested in moving forward with it please re-open it once the PR is updated.

tomekit:
After year and a half I can confirm this is still an issue. Newest `video_player` so far is 2.4.7 and if you have spaces in the local file name or remote URL the video won't load on iOS.

PR #2260, opened 2019-11-08
Make setMockInitialValues handle non-prefixed keys

mehmetf:


packages/shared_preferences/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.4+4
+
+* setMockInitialValues need to handle non-prefixed keys since that's an implementation detail.
```
collinjackson:
```suggestion
* `setMockInitialValues` needs to handle non-prefixed keys since that's an implementation detail.
```

PR #2283, opened 2019-11-16
[android_alarm_manager] Testing and documentation

mklim:
## Related Issues

Fixes flutter/flutter#44385.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


mklim:
I don't like exposing the internal state with `@visibleForTesting`, but I don't know how else to handle dealing with the native `getCallbackHandle` or the `DateTime.now` call. Same with mocking the `MethodChannel`.

packages/android_alarm_manager/lib/android_alarm_manager.dart
```diff
@@ -54,6 +54,31 @@ void _alarmManagerCallbackDispatcher() {
   _channel.invokeMethod<void>('AlarmService.initialized');
 }
 
+/// This is a wrapper around a few static functions so that unit tests can mock
+/// them out in the test environment. This is only visible for the unit tests and
+/// should not be accessed directly by users of the plugin.
+@visibleForTesting
+class InstanceWrappers {
```
bkonyi: Will this appear in the docs for the plugin?

packages/android_alarm_manager/test/android_alarm_manager_test.dart
```diff
@@ -0,0 +1,203 @@
...
+  group('${AndroidAlarmManager.oneShotAt}', () {
+    test('validates input', () async {
+      final DateTime validTime = DateTime.utc(1993);
+      final int validId = 1;
+      final Function validCallback = (int id) => null;
+
+      // Callback should take a single int param.
+      expect(
+          () => AndroidAlarmManager.oneShotAt(
+              validTime, validId, ((String foo) => foo)),
```
bkonyi: I wouldn't pass a closure to this method, even if it's valid for this test. Callbacks can only be tear-offs for static methods (top level or class level) so this gives a false impression that anonymous closures are supported.

packages/android_alarm_manager/test/android_alarm_manager_test.dart
```diff
@@ -0,0 +1,203 @@
...
+          throwsAssertionError);
+
+      // ID should be less than 32 bits.
```
bkonyi: Is this an actual restriction for AlarmManager?

packages/android_alarm_manager/test/android_alarm_manager_test.dart
```diff
@@ -0,0 +1,203 @@
...
+        expect(call.arguments[3], exact);
+        expect(call.arguments[4], wakeup);
+        expect(call.arguments[5], alarm.millisecondsSinceEpoch);
+        expect(call.arguments[6], rescheduleOnReboot);
+        expect(call.arguments[7], rawHandle);
+        return true;
+      });
+
+      final bool result = await AndroidAlarmManager.oneShotAt(
+          alarm, id, (int id) => id,
```
bkonyi: Same as above: avoid using anonymous closures in the tests as they're not supported in general. There's a few more instances where this is done below.

packages/android_alarm_manager/lib/android_alarm_manager.dart
```diff
@@ -188,7 +218,7 @@ class AndroidAlarmManager {
     assert(callback is Function() || callback is Function(int));
     assert(id.bitLength < 32);
     final int startMillis = time.millisecondsSinceEpoch;
-    final CallbackHandle handle = PluginUtilities.getCallbackHandle(callback);
+    final CallbackHandle handle = InstanceWrappers.getCallbackHandle(callback);
```
cyanglaz: Is it possible to factor this to an instance method that all it does is to get the `CallbackHandle`. And then in the test, mock the implementation of that instance method? Instead of creating this new static class.

packages/android_alarm_manager/lib/android_alarm_manager.dart
mklim: Yes. Technically this was a "missing doc" before since it's public and visible to users of the plugin. I figured a warning saying not to touch this made the most sense as the only user-facing documentation on it.

packages/android_alarm_manager/test/android_alarm_manager_test.dart
mklim: I'm not sure if this is enforced in the Android SDK, but the Dart API for this plugin is already enforcing it. See https://github.com/flutter/plugins/blob/master/packages/android_alarm_manager/lib/android_alarm_manager.dart#L189. I could change that if it's wrong but I wanted this to completely avoid changing any real functionality and just test what was already available and document it as-is.

packages/android_alarm_manager/lib/android_alarm_manager.dart
mklim: Yeah, I can bundle these into `AndroidAlarmManager` itself with static `@visibleForTesting` setters. Initially I avoided that since I thought it would clutter its interface but this isn't great either, that probably makes some more amount of sense.

packages/android_alarm_manager/test/android_alarm_manager_test.dart
bkonyi: Ah I didn't realize that. We should probably just leave it as-is for now then.

packages/android_alarm_manager/lib/android_alarm_manager.dart
bkonyi: I'm really not a fan of polluting the API surface with code only used for testing. If this wasn't going to appear in the docs that would be a different story, but it would be great if we could hide this somewhere.

mklim: Never mind, I'm going to delete this anyway as @cyanglaz suggested in https://github.com/flutter/plugins/pull/2283#discussion_r349196050.

mklim: Sorry, Github hadn't loaded the earlier comment at the time I posted my last one.

Ideally I'd like to avoid all `@visibleForTesting` identifiers too but I'm not sure how to completely do it here. `DateTime.now` and `PluginUtilities.getCallbackHandle` are both static and referenced statically internally. `AndroidAlarmManager` itself is static so I can't use constructor injection. The unit tests can only touch the class through its public API surfaces. Right now I am cutting down on the visibility a little bit from what's currently in this patch, but I'm still adding this method to `AndroidAlarmManager` itself:

```dart
  /// This is exposed for the unit tests. It should not be accessed by users of
  /// the plugin.
  @visibleForTesting
  static void setTestOverides({Now now, GetCallbackHandle getCallbackHandle}) {
    _now = now;
    _getCallbackHandle = getCallbackHandle;
  }
```

And then the class internally is using `_now` and `_getCallbackHandle` instead of the static functions.

If you have any ideas on how to avoid it totally I would be happy to hear it, I'd love to avoid this too. I'm also exposing the MethodChannel so I can test it too.

packages/android_alarm_manager/test/android_alarm_manager_test.dart
```diff
@@ -0,0 +1,201 @@
...
+      expect(call.arguments[4], wakeup);
+      expect(
+          call.arguments[5], now.millisecondsSinceEpoch + alarm.inMilliseconds);
+      expect(call.arguments[6], rescheduleOnReboot);
+      expect(call.arguments[7], rawHandle);
+      return true;
+    });
+
+    final bool result = await AndroidAlarmManager.oneShot(
+        alarm, id, (int id) => id,
```
bkonyi: This isn't a valid callback as it's not a static named function.

packages/android_alarm_manager/test/android_alarm_manager_test.dart
```diff
@@ -0,0 +1,201 @@
...
+        expect(call.arguments[4], period.inMilliseconds);
+        expect(call.arguments[5], rescheduleOnReboot);
+        expect(call.arguments[6], rawHandle);
+        return true;
+      });
+
+      final bool result = await AndroidAlarmManager.periodic(
+        period,
+        id,
+        (int id) => null,
```
bkonyi: Also an invalid callback.

packages/android_alarm_manager/test/android_alarm_manager_test.dart
```diff
@@ -0,0 +1,201 @@
...
+    expect(initialized, isTrue);
+  });
+
+  group('${AndroidAlarmManager.oneShotAt}', () {
+    test('validates input', () async {
+      final DateTime validTime = DateTime.utc(1993);
+      final int validId = 1;
+
+      // Callback should take a single int param.
+      expect(
```
bkonyi: You might want to update these test cases to use `expectLater` since these methods you're testing are `async`. This works at this point since we're hitting the asserts before any asynchronous code is run (async methods run synchronously until their first awaited operation), but if anything is added that performs an `await` things might not behave as you'd expect.

I typically use the following pattern for testing asynchronous code:

```dart
await expectLater(() async => await BadAssertAsynchronousCode(), throwsAssertionError);
```

packages/android_alarm_manager/test/android_alarm_manager_test.dart
mklim: Whoops, didn't catch all of the lambdas. Thanks.

packages/android_alarm_manager/test/android_alarm_manager_test.dart
mklim: Good catch, fixed here and throughout.

PR #2286, opened 2019-11-19
[video_player] Improve DartDocs and test coverage

mklim:
## Description

Also adds a lint to prevent undocumented APIs going forward.

## Related Issues

Fixes flutter/flutter#44405.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/video_player/analysis_options.yaml
+++ b/packages/video_player/analysis_options.yaml
@@ -1,0 +1,11 @@
+# This exists to add a lint for missing API docs just on this specific package,
+# since not all packages have coverage for all their public members yet and
+# adding it in would be non-trivial. `public_member_api_docs` should be applied
+# to new packages going forward, and ideally the main `analysis_options.yaml`
+# file as soon as possible.
+
+include: ../../analysis_options.yaml
+
+linter:
+  rules:
+    - public_member_api_docs
--- a/packages/video_player/video_player/CHANGELOG.md
+++ b/packages/video_player/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.3+3
+
+* Add DartDocs and unit tests.
+
 ## 0.10.3+2
 
 * Update the homepage to point to the new plugin location
--- a/packages/video_player/video_player/analysis_options.yaml
+++ b/packages/video_player/video_player/analysis_options.yaml
@@ -4,7 +4,7 @@
 # to new packages going forward, and ideally the main `analysis_options.yaml`
 # file as soon as possible.
 
-include: ../../analysis_options.yaml
+include: ../../../analysis_options.yaml
 
 linter:
   rules:
--- a/packages/video_player/video_player/example/lib/main.dart
+++ b/packages/video_player/video_player/example/lib/main.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+// ignore_for_file: public_member_api_docs
+
 /// An example of using the plugin, controlling lifecycle and playback of the
 /// video.
 
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -14,16 +14,46 @@ final MethodChannel _channel = const MethodChannel('flutter.io/videoPlayer')
   // performed.
   ..invokeMethod<void>('init');
 
+/// Describes a discrete segment of time within a video using a [start] and
+/// [end] [Duration].
 class DurationRange {
+  /// Trusts that the given [start] and [end] are actually in order. They should
+  /// both be non-null.
   DurationRange(this.start, this.end);
 
+  /// The beginning of the segment described relative to the beginning of the
+  /// entire video. Should be shorter than or equal to [end].
+  ///
+  /// For example, if the entire video is 4 minutes long and the range is from
+  /// 1:00-2:00, this should be a `Duration` of one minute.
   final Duration start;
+
+  /// The end of the segment described as a duration relative to the beginning of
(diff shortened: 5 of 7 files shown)


packages/video_player/lib/video_player.dart
```diff
@@ -14,16 +14,46 @@ final MethodChannel _channel = const MethodChannel('flutter.io/videoPlayer')
...
     return start.inMilliseconds / duration.inMilliseconds;
   }
 
+  /// Assumes that [duration] is the total length of the video that this
+  /// DurationRange is a segment form. It returns the percentage that [start] is
+  /// through the entire video.
+  ///
+  /// For example, assume that the entire video is 4 minutes long. If [end] has a
+  /// duration of two minutes, this will return `0.5` since the DurationRange
+  /// ends 50% of the way through the video's total length.
```
cyanglaz: Maybe use a similar example with the `startFraction` so it's easier to understand. 0.5 is kind of ambitious. 
maybe 1 min -> 0.75 -> 75%

packages/video_player/lib/video_player.dart
```diff
@@ -87,10 +135,18 @@ class VideoPlayerValue {
   /// Is null when [initialized] is false.
   final Size size;
 
+  /// False when [duration] is null, true otherwise.
```
cyanglaz: I feel like this is kind of explaining he implementation detail. If the implementation changes we will have to update the doc as well which is kinda of "high maintenance"?
Maybe something like `Indicates whether the video player is finished initialization and can be played`.

packages/video_player/lib/video_player.dart
```diff
@@ -87,10 +135,18 @@ class VideoPlayerValue {
...
   bool get initialized => duration != null;
+
+  /// Whether or not [errorDescription] is null.
```
cyanglaz: ditto as `initialized`

packages/video_player/lib/video_player.dart
```diff
@@ -87,10 +135,18 @@ class VideoPlayerValue {
...
   bool get hasError => errorDescription != null;
+
+  /// Returns [size.width] / [size.height] when size is non-null, or `1.0.` when
```
cyanglaz: ditto as `initialized`

packages/video_player/lib/video_player.dart
```diff
@@ -305,16 +383,23 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     super.dispose();
   }
 
+  /// Starts playing the video.
+  ///
+  /// This finishes as soon as the "play" command has been sent to the platform,
+  /// not when playback is complete.
```
cyanglaz: maybe: `This method returns as soon as the playback has started, not when it is complete.`

packages/video_player/lib/video_player.dart
```diff
@@ -503,15 +596,44 @@ class _VideoPlayerState extends State<VideoPlayer> {
   }
 }
 
+/// Used to configure the [VideoProgressIndicator] widget's colors for how it
+/// describes the video's status.
+///
+/// Normally the widget uses certain default colors, but this is customizeable
```
cyanglaz: Maybe remove the 'Normally'?

packages/video_player/lib/video_player.dart
```diff
@@ -584,16 +706,38 @@ class _VideoScrubberState extends State<_VideoScrubber> {
 /// [padding] allows to specify some extra padding around the progress indicator
 /// that will also detect the gestures.
 class VideoProgressIndicator extends StatefulWidget {
+  /// Construct an instance that displays the play/buffering status of the video
+  /// controlled by [controller].
+  ///
+  /// Defaults will be used for everything except [controller] if they're not
+  /// provided. [allowScrubbing] defaults to false, and [padding] will default
```
cyanglaz: nits: `The [allowingScrubing]...`
I think we don't start sentences with small letter characters

packages/video_player/lib/video_player.dart
mklim: I think in this specific case (unlike the other two) this is actually explaining what the getter does in a way that's hard to implify. An aspect ratio really is either width / height or height / width, and I think it's useful to know what happens in strange edge cases where there aren't any dimensions loaded yet.

packages/video_player/lib/video_player.dart
mklim: This is referring specifically to the future. It completes once `"play"` itself starts, not when playback is complete. I thought that was potentially confusing so I mentioned it. I'll clear this up to say "The future that is returned from this method finishes as soon as..."

packages/video_player/lib/video_player.dart
mklim: It's weird to start off a sentence with "[a", but I think this is actually correct here for a few reasons.

Using  "the" would imply that `allowScrubbing` is significant somehow--that it's a specific `allowScrubbing` out of multiple possible `allowScrubbing`s out there, or maybe somehow very important. Similarly if I were to use "a/an" I would be implying a random `allowScrubbing` out of the various `allowScrubbing`s available. But In this case there is only one parameter called `allowScrubbing`, so it doesn't make sense to use either article. I'm referring to the one and only `allowScrubbing` throughout, and it's just a regular parameter like any other one.

And then after that normally sentences do start with upper case letters. This is where it gets into extremely confusing territory because prescriptive grammarians haven't really caught on to coding documentation yet, but here I'm basically using `[allowScrubbing]` as a name that is also lower cased. That's unusual for English, but when it does happen it is correct to just use the proper name as it's normally capitalized instead of upper casing it, the same if I were talking about iPhones. See https://english.stackexchange.com/a/68665 for some discussion on it.

That said, English doesn't really have real grammar rules so this is all very subjective.

PR #2291, opened 2019-11-20
add shared_preferences_platform_interface

yjbanov:
## Description

- Move `shared_preferences` one level down to make room for new packages.
- Add `package:shared_preferences_platform_interface` containing the core interface.
- Copy method channel implementation and tests into the new package.

There are no changes in `package:shared_preferences` yet other than moving the files.

## Related Issues

First step towards https://github.com/flutter/flutter/issues/35116


packages/shared_preferences/shared_preferences_platform_interface/lib/shared_preferences_platform_interface.dart
```diff
@@ -0,0 +1,80 @@
...
+// found in the LICENSE file.
+
+import 'dart:async';
+
+abstract class SharedPreferencesPlatform {
+  /// Returns all keys in the persistent storage.
+  Set<String> getKeys();
+
+  /// Reads a value of any type from persistent storage.
+  dynamic get(String key);
```
ditman: `_getSharedPreferencesMap` above returns a (Future) `Map<String, Object>`. Should this getter return `Object` instead of `dynamic`?

yjbanov: I think That would be a breaking change. Today you can do: `preferences.get("some_string").substring(0, 3)`. If I change it to `Object` you'd get a compiler error complaining that `Object` doesn't have the `substring` method.

hterkelsen:
As we discussed offline, please make sure that you add the static `instance` setters and associated checks and comments. See [here for an example](https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher_platform_interface/lib/url_launcher_platform_interface.dart).

packages/shared_preferences/shared_preferences_platform_interface/lib/shared_preferences_platform_interface.dart
ditman: Ouchhhh

yjbanov:
@hterkelsen 

> As we discussed offline, please make sure that you add the static instance setters and associated checks and comments.

Done. PTAL.

yjbanov:
@hterkelsen 

> You also need to update the homepage: in the pubspec.yaml for the original shared_preferences plugin, since you moved it. And you'll also need to bump the version of the shared_preferences so the new homepage is picked up in pub.

Done.

yjbanov:
@hterkelsen I think I found a more ergonomic way for overriding statics, such as `getInstance` and `setMockInitialValues`. I added an indirection via `SharedPreferencesBinding` that controls creation and mocking of `SharedPreferencesPlatform` implementations. This indirection allows keeping all interfaces, including `static` methods, intact.

https://github.com/flutter/plugins/pull/2291/commits/1a920f1d349cb085685b715b2a5b59068c5a3ef0

packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
```diff
@@ -0,0 +1,225 @@
...
+
+import 'shared_preferences_platform_interface.dart';
+
+const MethodChannel _kChannel =
+    MethodChannel('plugins.flutter.io/shared_preferences');
+
+const String _prefix = 'flutter.';
+
+class MethodChannelSharedPreferencesBinding
+    implements SharedPreferencesBinding {
```
ditman: So the way that others are going to use this is by doing:

```dart
class WebSharedPreferencesBinding implements SharedPreferencesBinding {

@override
static SharedPreferencesBinding instance =
      WebSharedPreferencesBinding();

...
```

Unless the interface of the `SharedPreferencesBinding` never changes, doesn't this need to be `extends` so changes in the interface don't cause compile-time errors in other packages?

IIRC extends is used in the PlatformInterface packages so they act similarly to Java "interfaces with default implementations".

yjbanov:
Ok, so after chatting with @hterkelsen we agreed to redesign it. The new design only abstracts away the method channel rather than all of `SharedPreferences`. The new interface is `SharedPreferencesStorePlatform` and it is only concerned about persisting key/value pairs. The caching and prefixing logic stays in `SharedPreferences`.

PTAL

packages/shared_preferences/shared_preferences_platform_interface/lib/shared_preferences_platform_interface.dart
```diff
@@ -0,0 +1,116 @@
...
+  //
+  // This private method is called by the instance setter, which fails if the class is
+  // implemented with `implements`.
+  void _verifyProvidesDefaultImplementations() {}
+}
+
+/// Stores data in memory.
+///
+/// Data does not persist across application restarts. This is useful in unit-tests.
+class InMemorySharedPreferencesStore extends SharedPreferencesStorePlatform {
```
ditman: Nice!

packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
```diff
@@ -0,0 +1,61 @@
...
+import 'shared_preferences_platform_interface.dart';
+
+const MethodChannel _kChannel =
+    MethodChannel('plugins.flutter.io/shared_preferences');
+
+/// Wraps NSUserDefaults (on iOS) and SharedPreferences (on Android), providing
+/// a persistent store for simple data.
+///
+/// Data is persisted to disk asynchronously.
+class MethodChannelSharedPreferencesStore
```
ditman: Did you just refactor this file and made it only implement the things that are .invoked in the old plugin?

ditman: Indeed he did :P https://github.com/flutter/plugins/pull/2291#issuecomment-557754517

ditman:
The new version seems to reduce significantly the amount of code that other implementors need to write, I like it! (Note that the analyzer is screaming at a couple of minor things)

packages/shared_preferences/shared_preferences_platform_interface/lib/shared_preferences_platform_interface.dart
```diff
@@ -0,0 +1,118 @@
...
+  /// * Value type "Bool" must be passed if the value is of type `bool`.
+  /// * Value type "Double" must be passed if the value is of type `double`.
+  /// * Value type "Int" must be passed if the value is of type `int`.
+  /// * Value type "String" must be passed if the value is of type `String`.
+  /// * Value type "StringList" must be passed if the value is of type `List<String>`.
+  Future<bool> setValue(String valueType, String key, Object value);
+
+  /// This does not do anything.
+  ///
+  /// It is only here for backwards-compatibility.
```
collinjackson: I found this comment confusing. This is the 1.0 version of the platform interface, so there's no need to be backwards compatible. Can we just remove it instead of baking this flaw into the platform API forever? Deprecating/removing `commit` from the app-facing API can be done later.

yjbanov: Currently the plugin sends a `commit` message over to the Java/Objective-C side via a platform channel. I don't know how breaking it would be to stop sending this message, but I was thinking of adding Web support without breaking changes. However, if you are OK with it I'll remove it.

hterkelsen: I just checked the iOS and Android implementations, and they just return `true` when they get a `commit` call. I think we should remove it from the platform interface, and when you migrate the plugin, just have `commit` return `true` directly.

yjbanov: Done.

ditman:
Yay, thanks @yjbanov!!

PR #2294, opened 2019-11-22
WIP maps plugin migration attempt as embedding sanity check.

matthew-carroll:
This is a draft PR to demonstrate the progress I've been able to make in technically migrating the maps plugin to the new Android embedding.

The approach taken is probably not ideal. I was going for the most direct changes possible.

I left the state restoration for the `MapView` unresolved in this PR because, while we should be able to hook up the appropriate callbacks, it is unclear to me how such state restoration could be made to work with the current lifespan of objects in the maps plugin.

The `GoogleMapController` appears to treat its `Activity` reference as permanent. But the nature of state restoration is that such an `Activity` has been destroyed, the state saved, and then upon recreation of that `Activity`, that state is restored. Since `GoogleMapController` expects its `Activity` to never change, it implies that a new `GoogleMapController` must be created to deal with a new `Activity`. The foundational object structure for the maps plugin looks like it probably needs to change to accommodate the possibility of a cached `FlutterEngine` that does not get destroyed with its `Activity`. I'm not sure the best refactor for that purpose, nor am I sure it's worth the time right now.

There is, however, one possible point of concern in the embedding implementation. The timing of state restoration is a tricky issue.

In `FlutterActivity`, a call to its delegate's `onAttach()` method is responsible for setting up the `FlutterEngine` and attaching plugins vis the `GeneratedPluginRegistrant`. The very next line invokes `onActivityCreated()` where the delegate can restore the previous state.

https://github.com/flutter/engine/blob/master/shell/platform/android/io/flutter/embedding/android/FlutterActivity.java#L418

Assuming all relevant plugins are registered in `onConfigureFlutterEngine()` or before, they should receive an opportunity to restore instance state. However, if a dev registers any other plugins at any point after that, those plugins will miss the opportunity. So, even if a developer, for some reason, needs to wait for `onStart()` or `onResume()` to register a plugin, they'll miss that opportunity.

I'm not sure that there is a good way around this or not, but I can imagine that any dev that runs into this situation could find it very difficult to deal with the inability to restore plugin state. I'm mentioning this for all PR reviewers to be aware and have an opportunity to offer thoughts. I don't have any particular recommendation.

### DIFF
--- a/packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapBuilder.java
+++ b/packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapBuilder.java
@@ -6,10 +6,11 @@ package io.flutter.plugins.googlemaps;
 
 import android.content.Context;
 import android.graphics.Rect;
+import androidx.lifecycle.Lifecycle;
 import com.google.android.gms.maps.GoogleMapOptions;
 import com.google.android.gms.maps.model.CameraPosition;
 import com.google.android.gms.maps.model.LatLngBounds;
-import io.flutter.plugin.common.PluginRegistry;
+import io.flutter.plugin.common.BinaryMessenger;
 import java.util.concurrent.atomic.AtomicInteger;
 
 class GoogleMapBuilder implements GoogleMapOptionsSink {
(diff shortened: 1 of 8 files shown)


stuartmorgan:
Closing as obsolete.

PR #2299, opened 2019-11-23
[local_auth] support v2 embedding

bparrishMines:
Fixes https://github.com/flutter/flutter/issues/41841

~~Tests fail on stable until `FlutterLifecycleAdapter` lands in stable.~~


xster:
LGTM, thanks!

script/build_all_plugins_app.sh
```diff
@@ -11,6 +11,7 @@ source "$SCRIPT_DIR/common.sh"
 check_changed_packages > /dev/null
 
 readonly EXCLUDED_PLUGINS_LIST=(
+  "flutter_plugin_android_lifecycle"
```
bparrishMines: Adding this to the exclude list because the all plugins app creates a local dependency while this plugin creates a hosted dependency. This caused errors at compilation.

mehmetf:
I am confused about this PR. The lifecycle adapter does not exist in master either: https://github.com/flutter/engine/tree/master/shell/platform/android/io/flutter/embedding/engine/plugins/lifecycle.

PR #2302, opened 2019-11-24
[google_maps_flutter] iOS: add a cache for defaultMarker(WithHue) and fromBytes icons...

sperochon:
... to avoid following errors :
> ((null)) was false: Reached the max number of texture atlases, can not allocate more.
> ((null)) was false: Failed to allocate texture space for marker

Markers are created but the icon images are missing into the map.

## Description
iOS only: This PR adds a cache system to keep references on the icons images created for each marker.
When there are more than about 1250 markers using the same icon = BitmapDescriptor.defaultMarker(WithHue) or BitmapDescriptor.fromBytes, then the plugin creates 1250 UIImage in memory.
It writes some error messages while debugging and crashes on a release app.

## Related Issues
This PR fixes this issue:
https://github.com/flutter/flutter/issues/45508

I don't know how an integration test could test this... I've started one but I'm stuck on it...

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
+++ b/packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
@@ -168,12 +168,24 @@ static UIImage* scaleImage(UIImage* image, NSNumber* scaleParam) {
 
 static UIImage* ExtractIcon(NSObject<FlutterPluginRegistrar>* registrar, NSArray* iconData) {
   UIImage* image;
+  static NSMutableDictionary* iconsCache;
+
+  if (iconsCache == nil) {
+    iconsCache = [NSMutableDictionary dictionary];
+  }
+
+  NSString *key;
   if ([iconData.firstObject isEqualToString:@"defaultMarker"]) {
     CGFloat hue = (iconData.count == 1) ? 0.0f : ToDouble(iconData[1]);
+    key = [NSString stringWithFormat:@"%@/%f", iconData.firstObject, hue];
+    if (iconsCache[key] != nil) {
+      return iconsCache[key];
+    }
     image = [GMSMarker markerImageWithColor:[UIColor colorWithHue:hue / 360.0
-                                                       saturation:1.0
-                                                       brightness:0.7
-                                                            alpha:1.0]];
+                                                    saturation:1.0
+                                                    brightness:0.7
+                                                    alpha:1.0]];
+    iconsCache[key] = image;
   } else if ([iconData.firstObject isEqualToString:@"fromAsset"]) {
     if (iconData.count == 2) {
       image = [UIImage imageNamed:[registrar lookupKeyForAsset:iconData[1]]];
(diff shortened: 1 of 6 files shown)


packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -196,19 +209,29 @@ static void InterpretInfoWindow(id<FLTGoogleMapMarkerOptionsSink> sink, NSDictio
       @throw exception;
     }
   } else if ([iconData[0] isEqualToString:@"fromBytes"]) {
-    if (iconData.count == 2) {
+    if ((iconData.count == 2) || (iconData.count == 3)) {
+      key = nil;
+      if (iconData.count == 3) {
+          key = [NSString stringWithFormat:@"%@/%f", iconData[0], iconData[2]];
```
otopba: I think it is not good idea to make `fromBytes`:
- This cache can be both for Android and iOS.
- It is more likely that the user should cache the BitmapDescriptor, if necessary.
- If you add something by key, you should be able to update / reset the cache

packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -168,12 +168,24 @@ static void InterpretInfoWindow(id<FLTGoogleMapMarkerOptionsSink> sink, NSDictio
 
 static UIImage* ExtractIcon(NSObject<FlutterPluginRegistrar>* registrar, NSArray* iconData) {
   UIImage* image;
+  static NSMutableDictionary* iconsCache;
+
+  if (iconsCache == nil) {
+    iconsCache = [NSMutableDictionary dictionary];
```
otopba: May be this cache should have some configuration? User can have a lot of icons with different `hue` values. I think you should limit cache size

packages/google_maps_flutter/ios/Classes/GoogleMapMarkerController.m
amirh: +1 to that, I think we need to have an identifier per bitmap and make sure we only transmit the bytes over the method channel once.

In a scenario with a large number of `fromBytes` markers, in addition to the high memory usage here we also have much more bytes going over the method channel than needed.

amirh: Sorry I spoke too early :) looks like the tag is that ID I'm asking for, though we're still sending the same bytes over the method channel multiple times. What do you think about making it so that each BitmapDescriptor has a unique ID? then I think we can make it so that we're only transmitting the bytes over the method channel once per bitmap?

It will require some more work, but it will have gains for all existing users (on all platforms) even without having them add a tag per marker. If we envision this as the better solution we'll want to eventually have I'd be very hesitant to introduce the tag parameter right now which will be a breaking change to remove later.

sanekyy:
Hi @sperochon. Can you update your PR?

sperochon:
Hi @sanekyy Sorry but I have no time right now to do what is asked... You can do it if you've got time.

stuartmorgan:

PR #2306, opened 2019-11-26
[connectivity] Adds Macos support

franciscojma86:
## Description

Adds a macOS implementation to the Connectivity plugin.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/41711

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/connectivity/connectivity_macos/.gitignore
```diff
@@ -0,0 +1,2 @@
+.packages
+pubspec.lock
```
mklim: nit: These are already in the root `.gitignore`, I don't think there's a reason to create a file with duplicate entries here.

packages/connectivity/connectivity_macos/ios/.gitignore
```diff
@@ -0,0 +1,36 @@
+.idea/
```
mklim: Same here, most of these are already in the root `.gitignore` under `flutter/plugins`. If there's anything missing I would rather just add it there since this is going to apply to all macOS plugins.

packages/connectivity/connectivity_macos/macos/Classes/ConnectivityPlugin.swift
```diff
@@ -0,0 +1,104 @@
+// Copyright 2019 Google LLC
```
mklim: Should we be writing these in Swift or Objective C? I don't mind Swift, but so far we've been avoiding it to avoiding adding an N+1 language burden to the repo. /cc @amirh 

If we're writing the macOS plugins in Swift I feel like that's more of a reason to also write the mobile ones in Swift, too.

packages/connectivity/connectivity_macos/macos/Classes/ConnectivityPlugin.swift
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /// Returns a string describing connection type
+  ///
+  /// - Parameters:
+  ///   - reachability: an instance of reachability
+  /// - Returns: connection type string
+  private func statusFromReachability(reachability: Reachability?) -> String {
+    // checks any non-WWAN connection
+    if reachability?.isReachableViaWiFi() ?? false {
```
mklim: What is the `?? false` accomplishing here? (I'm not very familiar with Swift's syntax. I'm guessing it's saying false if anything in the previous statement evaluates to null, but if that's the case my guess would be that it's not needed since the null aware operators would evaluate to false anyway. Swift may be more picky about types than that though.)

packages/connectivity/connectivity_macos/macos/Classes/ConnectivityPlugin.swift
```diff
@@ -0,0 +1,104 @@
...
+import Cocoa
+import CoreWLAN
+import FlutterMacOS
+import Reachability
+import SystemConfiguration.CaptiveNetwork
+
+public class ConnectivityPlugin: NSObject, FlutterPlugin, FlutterStreamHandler {
+  var reach: Reachability?
+  var eventSink: FlutterEventSink?
+  var cwinterface: CWInterface?
```
mklim: Can this be set immutably? It's a little cleaner to be able to do that than have to use null aware operators everywhere. I assume there may be some performance issue that means this needs to live in `init()`, but wanted to double check.

packages/connectivity/connectivity_macos/macos/Classes/IPHelper.h
```diff
@@ -0,0 +1,51 @@
...
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+#import "SystemConfiguration/CaptiveNetwork.h"
+
+#include <arpa/inet.h>
+#include <ifaddrs.h>
+
+NSString* getWifiIP() {
```
mklim: Would it make sense to split this into an implementation file or no?

packages/connectivity/connectivity_macos/macos/Classes/IPHelper.h
```diff
@@ -0,0 +1,51 @@
...
+  if (success == 0) {
+    // Loop through linked list of interfaces.
+    temp_addr = interfaces;
+    while (temp_addr != NULL) {
+      if (temp_addr->ifa_addr->sa_family == AF_INET) {
+        // Check if interface is en0 which is the wifi connection on the iPhone.
+        if ([[NSString stringWithUTF8String:temp_addr->ifa_name] isEqualToString:@"en0"]) {
+          // Get NSString from C String
+          address = [NSString
+              stringWithUTF8String:inet_ntoa(((struct sockaddr_in*)temp_addr->ifa_addr)->sin_addr)];
```
mklim: Are all of these guaranteed to be non-null? Same question about the if block above.

packages/connectivity/connectivity_macos/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
+name: connectivity_macos
+description: Temporary desktop implmentation of connectivity from flutter/plugins
```
mklim: Why is this temporary? I think this needs a README to explain what this is in more detail.

packages/connectivity/example/android/settings.gradle
```diff
@@ -10,6 +10,8 @@ if (pluginsFile.exists()) {
 
 plugins.each { name, path ->
     def pluginDirectory = flutterProjectRoot.resolve(path).resolve('android').toFile()
-    include ":$name"
-    project(":$name").projectDir = pluginDirectory
+    if (pluginDirectory.exists()) {
+        include ":$name"
+        project(":$name").projectDir = pluginDirectory
+    }
```
mklim: If this is something every single app including this plugin needs or else it will fail to compile, this probably shouldn't merge because it will break existing apps. Apps currently using `connectivity` don't have this block and won't know to update this when their build breaks.

packages/connectivity/example/macos/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Flutter-related
```
mklim: Same comment here about removing the .gitignore.

packages/connectivity/pubspec.yaml
```diff
@@ -26,4 +26,4 @@ dev_dependencies:
 
 environment:
   sdk: ">=2.0.0-dev.28.0 <3.0.0"
-  flutter: ">=1.6.7 <2.0.0"
+  flutter: ">=1.6.7 <2.0.0"
```
mklim: Does this really work with Flutter 1.6.7 or does it need a higher minimum bound? If I'm remembering right a lot of the federated plugin work is newer, right?

packages/connectivity/connectivity_macos/.gitignore
franciscojma86: Done

packages/connectivity/connectivity_macos/ios/.gitignore
franciscojma86: Done

packages/connectivity/connectivity_macos/macos/Classes/ConnectivityPlugin.swift
franciscojma86: I believe the preferred language for macOS plugins is Swift. /cc @stuartmorgan 

I don't think it would be worth rewriting the iOS ones since they interoperate and would be a lot of work with small benefit.

packages/connectivity/connectivity_macos/macos/Classes/ConnectivityPlugin.swift
franciscojma86: You are mostly correct on how that works except on the optional object part :). The optional has to be unwrapped or it will raise  a compiler error. With this syntax we don't have to unwrap it, but just supply a default false value. It's less verbose than unwrapping it above and still supply the default value.

packages/connectivity/connectivity_macos/macos/Classes/ConnectivityPlugin.swift
franciscojma86: The problem is that all these are set by methods executed later in the class lifecycle. Swift requires that immutable properties are set from the initialization of the class. In this case optionals are ok since force unwrapping is discouraged, and setting empty values at init would still require lots of checking

packages/connectivity/connectivity_macos/macos/Classes/IPHelper.h
franciscojma86: No, but in this case it's not an issue. 

In the block above, calling `isEqualToString` on nil is a no-op, so the if statement would be false.
Passing nil as the return value is also acceptable.

packages/connectivity/connectivity_macos/pubspec.yaml
franciscojma86: It's not, I forgot to change that message :) Done

packages/connectivity/example/macos/.gitignore
franciscojma86: Done

packages/connectivity/pubspec.yaml
franciscojma86: AFAIK this change is not using any new federated plugin feature. /cc @amirh  is that right?

packages/connectivity/connectivity_macos/macos/Classes/ConnectivityPlugin.swift
stuartmorgan: Swift is the preferred language for macOS plugins; there isn't even an ObjC template in the tool. It's possible to write ObjC plugins, but we should be doing what we are recommending to users.

Whether new iOS plugins, if any, should use Swift is a separate decision. On macOS there's no legacy argument, so the decision is much clearer.

packages/connectivity/example/android/settings.gradle
franciscojma86: Yeah, I reverted this change and changed the connectivity_macos/pubspec.yaml to require flutter >= 1.13.0 where this is not an issue anymore. Will have to wait until https://github.com/flutter/plugins/pull/2334 is landed to land this as well.

packages/connectivity/pubspec.yaml
stuartmorgan: The macOS plugin uses the new pubspec declaration format, but if the dependency on the macOS plugin isn't being added yet the version requirement of the main plugin doesn't change AFAIK.

packages/connectivity/connectivity_macos/macos/Classes/ConnectivityPlugin.swift
mklim: Right, I was wondering if we could set ` cwinterface = CWWiFiClient.shared().interface()` immediately instead of in `init()`, specifically. I'm guessing there's a reason we want that call to be delayed though?

packages/connectivity/connectivity_macos/macos/Classes/ConnectivityPlugin.swift
mklim: Got it, thanks!

packages/connectivity/connectivity_macos/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+flutter:
+  plugin:
+    platforms:
+      macos:
+        pluginClass: ConnectivityPlugin
+
+environment:
+  sdk: ">=2.1.0 <3.0.0"
+  flutter: ">=1.13.0 <2.0.0"
```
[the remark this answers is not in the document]
stuartmorgan: How did the web federated implementations land then?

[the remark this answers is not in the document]
mklim: Uh, good question. I didn't realize that they were requiring above stable and I would also expect them to break CI in the same way this PR is currently red. I'll do some digging.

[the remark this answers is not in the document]
mklim: @franciscojma86 explained this to me offline. So our [`.cirrus.yml` actually _deletes_ the web files off of the local CI checkout](https://github.com/flutter/plugins/blob/master/.cirrus.yml#L34) before CI runs to avoid this issue. We have an outstanding issue to remove that hack once the next stable rolls. flutter/flutter#42864 We can't use that approach here easily but Francisco is going to just add in a dummy Android project to the macOS implementation for now in order to make this compatible with the current stable instead. Once v1.13 rolls to the next stable he'll take it out of the macOS implementation and up the version constraint.

packages/connectivity/connectivity_macos/android/.gitignore
```diff
@@ -0,0 +1,8 @@
+*.iml
```
mklim: nit: This .gitignore should be deleted.

packages/connectivity/connectivity_macos/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
...
+    platforms:
+      macos:
+        pluginClass: ConnectivityPlugin
+
+environment:
+  sdk: ">=2.1.0 <3.0.0"
+  
+dependencies:
+  flutter:
+    sdk: flutter
```
mklim: Should the Flutter SDK have bounds defined here?

packages/connectivity/connectivity_macos/android/.gitignore
franciscojma86: Done

packages/connectivity/connectivity_macos/pubspec.yaml
franciscojma86: Done

amirh:
Please use the following folder structure used in other federated plugins:
```
connectivity/connectivity
connectivity/connectivity_macos
```

packages/connectivity/connectivity_macos/android/build.gradle
```diff
@@ -0,0 +1,37 @@
+group 'com.example.connectivity'
```
amirh: nit: use io.flutter as the plugin's org

PR #2308, opened 2019-11-26
[android_intent] Added 'action_app' action and getIntentExtras, getIntentData

eligt:
**Note**: re-implemented version of https://github.com/flutter/plugins/pull/1101

The "action_app" special action allows to create an Intent that launches the app's launcher activity (which for most Flutter apps, it's the Activity handling the Flutter app itself).

Also added methods `getIntentExtras` and `getIntentData` which respectively let you retrieve the current Intent's extras [arguments] and data.

These two features combined are useful when your app is divided between a foreground component (the main app running the Flutter view) and a background component (a service that runs in the background).

With these features the background component can easily trigger the main app to open (whether it was running or not) and pass extra arguments to the Intent. The main app can then load up and use these extra arguments to do some custom processing.

Please note that if your app is already opened and you launch a "action_app" intent, you need a bit of extra logic in your MainActivity to ensure that arguments are refreshed from the new Intent, something like this:

```java
  @Override
  protected void onNewIntent(Intent intent) {
      super.onNewIntent(intent);
      Bundle bundle = intent.getExtras();
      getIntent().putExtras(bundle);
  }
```

### DIFF
--- a/packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
+++ b/packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
@@ -64,7 +64,27 @@ public final class IntentSender {
       return;
     }
 
-    Intent intent = new Intent(action);
+    Intent intent = null;
+
+    if (action.equals("action_app")) {
+      try {
+        intent =
+            applicationContext
+                .getPackageManager()
+                .getLeanbackLaunchIntentForPackage(applicationContext.getPackageName());
+      } catch (java.lang.NoSuchMethodError e) {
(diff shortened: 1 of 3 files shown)


stuartmorgan:
Thanks for the submission! However, we’re currently not accepting non-critical PRs for this plugin, as we hope that in time we’re able to transition users to the corresponding plugin in [the Flutter Community Plus Plugins repository](https://plus.fluttercommunity.dev/). We encourage you to submit your PR there.

(We apologize for the long delay in triaging this PR. We’re in the process of overhauling our PR triage system to respond much more quickly, as well as working through the backlog.)

PR #2230, opened 2019-10-23
Forbid `... implements UrlLauncherPlatform`

amirh:
## Description

Fail setting `UrlLauncherPlatform.instance` to something that `implements` (rather than `extends`) `UrlLauncherPlatform`.

I'm not bumping the major version as the breaking change policy for platform interface is that we're only guaranteeing to not break classes that `extend` it.

We leave a backdoor for mockito mocks which can do:
```dart
class MockUrlLauncherPlatform extends Mock
    implements UrlLauncherPlatform {}

test('', () {
  MockUrlLauncherPlatform mock = MockUrlLauncherPlatform();
  when(mock.isMock).thenReturn(true);
  UrlLauncherPlatform.instance = mock;
}
```

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [ ] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

PR #2326, opened 2019-11-27
[url_launcher] Alternative to `isMock`

collinjackson:
This is an attempt to address what I understand to be the underlying motivation of @Hixie's [comment](https://github.com/flutter/plugins/pull/2230#issuecomment-558856299) on #2230.

## Problem

The `isMock` API is currently in the public docs (despite being `@visibleForTesting`) and we're not thrilled about it.

The `@visibleForTesting` annotation on `isMock` results in a warning that can be easily ignored by the plugin consumer. This could create a maintenance burden if platform implementers create packages that use the `isMock` backdoor to implement rather than extend the platform interface.

Furthermore, it is currently possible to bypass `isMock` trivially with `noSuchMethod`.

## Solution

This PR proposes a base class `PlatformInterface` that plugins can extend. It provides a static validation method that is used to ensure that subclasses use `extends` rather than `implements`. The verification is performed using an `identical` comparison on a private `Object` token.

I believe that this approach makes it impossible to bypass the validation check using `noSuchMethod`.

The base classes should be extracted into a common package, see https://github.com/flutter/flutter/issues/43368

To avoid introducing new API surface area that we later have to remove we should probably move the `PlatformInterface` class into either the Flutter SDK or a package.

/cc @amirh @goderbauer

### DIFF
--- a/packages/url_launcher/url_launcher/test/url_launcher_test.dart
+++ b/packages/url_launcher/url_launcher/test/url_launcher_test.dart
@@ -13,7 +13,6 @@ import 'package:flutter/services.dart' show PlatformException;
 
 void main() {
   final MockUrlLauncher mock = MockUrlLauncher();
-  when(mock.isMock).thenReturn(true);
   UrlLauncherPlatform.instance = mock;
 
   test('closeWebView default behavior', () async {
@@ -208,4 +207,4 @@ void main() {
   });
 }
 
-class MockUrlLauncher extends Mock implements UrlLauncherPlatform {}
+class MockUrlLauncher extends Mock with MockPlatformInterface implements UrlLauncherPlatform {}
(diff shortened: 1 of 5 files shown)


collinjackson:
@Hixie I've updated this PR based on our discussion, PTAL

collinjackson:
After discussing with Amir I think we'll want to copy these new base classes into a new first-party package (verify_extends? platform_interface?) and also add the base classes to the Flutter SDK. We'll need to wait until after next Flutter stable release to switch over to using the Flutter SDK versions of the classes in plugins. 

 `url_launcher_platform_interface` could export the `MockXXX` base class, with a comment that it is likely to be removed from the exported API and replaced with the Flutter SDK class in the future; this would not be considered a semantically breaking change to the platform interface since it is a test-only dependency.

I'm closing this PR since it's not able to land in the current state, @amirh can you take it from here?

PR #2327, opened 2019-11-27
[android_alarm_manager] Update minimum Flutter version to 1.12.0

bkonyi:
This plugin relies partially on functionality which will be available in the next Flutter release.

### DIFF
--- a/packages/android_alarm_manager/pubspec.yaml
+++ b/packages/android_alarm_manager/pubspec.yaml
@@ -22,5 +22,4 @@ flutter:
 
 environment:
   sdk: ">=2.0.0-dev.28.0 <3.0.0"
-  # TODO(bkonyi): set minimum Flutter version to next stable release
-  flutter: ">=1.2.0 <2.0.0"
+  flutter: ">=1.12.0 <2.0.0"


collinjackson:
Can you update the CHANGELOG (referencing the new functionality used?) and bump the version?

bkonyi:
> Can you update the CHANGELOG (referencing the new functionality used?) and bump the version?

That was done in the previous PR here: https://github.com/flutter/plugins/pull/2193 I seemed to have missed this TODO before merging that.

collinjackson:
You have my approval to merge and publish this as a late-breaking fix to the previous unpublished version.

mklim:
@bkonyi @collinjackson this is causing our CI to go red because `1.12.0` is past the current flutter stable (`1.9.1`), so when the tests are ran with stable it's unable to do anything with this plugin. Can this and any changes that need the newer version be reverted? [We try to support flutter stable as our minimum bound for plugins](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#plugin-compatibility).

I'm kind of surprised this PR passed CI initially, there may be a bug in what tests initially ran on this PR.

mklim:
Discussed offline, the bound here was introduced just to guard against v2 embedding code. Unfortunately we don't have a way to version the embedding that but so far we've been relying on documentation on the higher minimum bound for when developers use that path (flutter/flutter#42792). @bkonyi is going to put up a PR lowering this minimum bound back to stable.

PR #2328, opened 2019-11-28
Update documentation for connectivity

mehmetf:
We recently discovered that connectivity broadcasts no longer work in the background post Android O (API 26). In the long term Android wants to crack down on apps that awaken unnecessarily.

Apps should update their knowledge of connectivity when they are resumed if they are storing it. The plugin itself does not store current connectivity so the Dart API is not the appropriate place to do this.


ened:
@mehmetf is this also true for Flutter isolates running in foreground services? (I think the answer may be in my question already :)). If so, should we mention that as well?

mehmetf:
@cyanglaz Please take a look at this PR.

mehmetf:
@ened What do you mean by "this"? The connectivity plugin native code is not running in a foreground service.

To be clear, this issue does not arise because the isolate is killed. In this case, the isolate is alive and running, however the background service that delivers connectivity updates no longer does so post Android O.

mehmetf:
@ened Thanks for clarifying.

At its current state,
- The plugin has no clue where your isolate is being run (activity? service? foreground service?).
- Apps cannot change where the plugin is being inited. This plugin, for instance, uses `registrar.context()` to register the broadcast receiver. That's the application context.

Perhaps you are proposing to change the way the plugin is written to register itself in a foreground service? and wondering whether any plugin using a broadcast receiver can work around this by registering itself in a foreground service?

I would advise against that because every foreground service needs to have a sticky notification as long as it is running. Foreground services are reserved for actual critical work such as playing music. It is actually *pretty rare* that your app would need a foreground service. Android strongly encourages you to use background services or work manager jobs to do work. You should not be registering a foreground service just to listen to network changes.

collinjackson:
Note that (per @amirh) we prefer to do CHANGELOG entries and pubspec.yaml updates even for purely documentation-specific updates.

PR #2329, opened 2019-11-29
[in_app_purchase]Change a comment.

HiIamAlanOu:
## Description

Change a comment.


HiIamAlanOu:
@googlebot I signed it!

mklim:
Thanks again for the fix!

droidery:
@mklim Not sure the best place to ask, but can the new versions be published (0.2.2+5/0.2.2+6) to pub.dev (or what's the release cycle) as I need the fix from 0.2.2+5

mklim:
@droidery sorry about the delay there, thanks for the ping. The release process is totally manual and it looks like we missed publishing those PRs (https://github.com/flutter/flutter/issues/27258). They should be available now.

PR #2332, opened 2019-12-02
Web implementation of shared_preferences

yjbanov:
## Description

Implement Web support for `package:shared_preferences` on top of `window.localStorage`.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/35116

### DIFF
--- a/packages/shared_preferences/shared_preferences_web/CHANGELOG.md
+++ b/packages/shared_preferences/shared_preferences_web/CHANGELOG.md
@@ -1,0 +1,3 @@
+# 0.0.1
+
+- Initial release.
--- a/packages/shared_preferences/shared_preferences_web/CHANGELOG.md
+++ b/packages/shared_preferences/shared_preferences_web/CHANGELOG.md
@@ -1,3 +1,3 @@
-# 0.0.1
+# 0.1.0
 
 - Initial release.
--- a/packages/shared_preferences/shared_preferences_web/README.md
+++ b/packages/shared_preferences/shared_preferences_web/README.md
@@ -1,0 +1,30 @@
+# url_launcher_web
+
+The web implementation of [`url_launcher`][1].
+
+## Usage
+
+### Import the package
+To use this plugin in your Flutter Web app, simply add it as a dependency in
+your pubspec alongside the base `url_launcher` plugin.
+
+_(This is only temporary: in the future we hope to make this package an
+"endorsed" implementation of `url_launcher`, so that it is automatically
+included in your Flutter Web app when you depend on `package:url_launcher`.)_
+
+This is what the above means to your `pubspec.yaml`:
+
+```yaml
+...
+dependencies:
+  ...
+  url_launcher: ^5.1.4
+  url_launcher_web: ^0.1.0
+  ...
+```
+
+### Use the plugin
+Once you have the `url_launcher_web` dependency in your pubspec, you should
+be able to use `package:url_launcher` as normal.
+
+[1]: ../url_launcher/url_launcher
--- a/packages/shared_preferences/shared_preferences_web/README.md
+++ b/packages/shared_preferences/shared_preferences_web/README.md
@@ -20,7 +20,7 @@ This is what the above means to your `pubspec.yaml`:
 dependencies:
   ...
   shared_preferences: ^0.5.4+8
-  shared_preferences_web: ^0.0.1
+  shared_preferences_web: ^0.1.0
   ...
 ```
 
--- a/packages/shared_preferences/shared_preferences_web/README.md
+++ b/packages/shared_preferences/shared_preferences_web/README.md
@@ -1,16 +1,17 @@
-# url_launcher_web
+# shared_preferences_web
 
-The web implementation of [`url_launcher`][1].
+The web implementation of [`shared_preferences`][1].
 
 ## Usage
 
 ### Import the package
+
 To use this plugin in your Flutter Web app, simply add it as a dependency in
-your pubspec alongside the base `url_launcher` plugin.
+your `pubspec.yaml` alongside the base `shared_preferences` plugin.
 
 _(This is only temporary: in the future we hope to make this package an
-"endorsed" implementation of `url_launcher`, so that it is automatically
-included in your Flutter Web app when you depend on `package:url_launcher`.)_
+"endorsed" implementation of `shared_preferences`, so that it is automatically
+included in your Flutter Web app when you depend on `package:shared_preferences`.)_
 
 This is what the above means to your `pubspec.yaml`:
 
@@ -18,13 +19,14 @@ This is what the above means to your `pubspec.yaml`:
 ...
 dependencies:
   ...
-  url_launcher: ^5.1.4
-  url_launcher_web: ^0.1.0
+  shared_preferences: ^0.5.4+8
+  shared_preferences_web: ^0.0.1
   ...
 ```
 
 ### Use the plugin
-Once you have the `url_launcher_web` dependency in your pubspec, you should
-be able to use `package:url_launcher` as normal.
 
-[1]: ../url_launcher/url_launcher
+Once you have the `shared_preferences_web` dependency in your pubspec, you should
+be able to use `package:shared_preferences` as normal.
+
+[1]: ../shared_preferences/shared_preferences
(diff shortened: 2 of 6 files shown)


packages/shared_preferences/shared_preferences_web/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# 0.0.1
```
ditman: pub.dev is going to ding you for having such a low release number, I'd recommend bumping to at least `0.1.0` before publishing.

packages/shared_preferences/shared_preferences_web/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
+name: shared_preferences_web
+description: Web platform implementation of shared_preferences
+author: Flutter Team <flutter-dev@googlegroups.com>
+homepage: https://github.com/flutter/plugins/tree/master/packages/shared_preferences/shared_preferences_web
+version: 0.0.1
```
hterkelsen: Nit: prefer the initial release to be `0.1.0` to get a better score on pub

packages/shared_preferences/shared_preferences_web/lib/shared_preferences_web.dart
```diff
@@ -0,0 +1,91 @@
...
+  Future<bool> remove(String key) async {
+    _checkPrefix(key);
+    html.window.localStorage.remove(key);
+    return true;
+  }
+
+  @override
+  Future<bool> setValue(String valueType, String key, Object value) async {
+    _checkPrefix(key);
+    html.window.localStorage[key] = _encodeValue(value);
```
ditman: Does it make sense to prepend something to the user-supplied `key`, to minimize collisions with other things that might be using the shared local storage? `shared_preferences` or similar?

We can make this transparent on all add/remove operations so users don't even see that we're doing this (unless they go to the dev tools and peek :P)

packages/shared_preferences/shared_preferences_web/lib/shared_preferences_web.dart
```diff
@@ -0,0 +1,91 @@
...
+
+  Object _decodeValue(String encodedValue) {
+    final Object decodedValue = json.decode(encodedValue);
+
+    if (decodedValue is List) {
+      // JSON does not preserve generics. The encode/decode roundtrip is
+      // `List<String>` => JSON => `List<dynamic>`. We have to explicitly
+      // restore the RTTI.
+      return decodedValue.cast<String>();
+    }
```
ditman: Since the _decodeValue method returns `Object`, isn't this cast erased anyway? 
Wouldn't the user need to cast again on their end anyways, to move from `Object` to `List<String>`?

packages/shared_preferences/shared_preferences_web/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
...
+    sdk: flutter
+  meta: ^1.1.7
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+
+environment:
+  sdk: ">=2.0.0-dev.28.0 <3.0.0"
+  flutter: ">=1.5.0 <2.0.0"
```
ditman: We haven't been too strict with this so far, but I think that you need a newer version of flutter to support the plugin.platforms syntax of the pubspec.yaml.

I'm not sure what's a good value for the minimal version of flutter yet, though :)

packages/shared_preferences/shared_preferences_web/test/shared_preferences_web_test.dart
```diff
@@ -0,0 +1,89 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@TestOn('chrome') // Uses web-only Flutter SDK
```
ditman: This looks great, thanks for the test!

packages/shared_preferences/shared_preferences_web/pubspec.yaml
yjbanov: Done.

packages/shared_preferences/shared_preferences_web/CHANGELOG.md
yjbanov: Done.

packages/shared_preferences/shared_preferences_web/lib/shared_preferences_web.dart
yjbanov: The `SharedPreferences` class from `package:shared_preferences` already prepends "flutter.". So here I'm just verifying that it's prefixed in `_checkPrefix`. On top of that, for security reasons, `localStorage` is scoped to the domain, so you can't step on another site's data.

We could allow customizing the prefix, but currently the value is hard-coded across Dart, Java, and Objective-C code, so it would be a much bigger change.

packages/shared_preferences/shared_preferences_web/lib/shared_preferences_web.dart
yjbanov: In Dart generics are reified (unlike Java, where generics are erased at runtime). So casting from `List<String>` to `Object` does not lose any type information at all. It's still `List<String>` and so you can only cast it back to `List<String>`. Attempting to cast it to, say, `List<int>` will result in a cast error.

packages/shared_preferences/shared_preferences_web/pubspec.yaml
yjbanov: It's a copy'n'paste from `url_launcher`. LMK what I should use here instead.

packages/shared_preferences/shared_preferences_web/test/shared_preferences_web_test.dart
yjbanov: :+1:

packages/shared_preferences/shared_preferences_web/test/shared_preferences_web_test.dart
```diff
@@ -0,0 +1,89 @@
...
+        final dynamic value = kTestValues[key];
+        expect(await store.setValue(key.split('.').last, key, value), true);
+      }
+      expect(html.window.localStorage.keys, hasLength(kTestValues.length));
+      for (String key in html.window.localStorage.keys) {
+        expect(html.window.localStorage[key], json.encode(kTestValues[key]));
+      }
+
+      // Check that generics are preserved.
+      expect((await store.getAll())['flutter.StringList'], isA<List<String>>());
```
yjbanov: @ditman this test here shows how the generics are preserved. If you change this to `isA<List<int>>()`, the test will fail.

packages/shared_preferences/shared_preferences_web/README.md
```diff
@@ -0,0 +1,32 @@
...
+included in your Flutter Web app when you depend on `package:shared_preferences`.)_
+
+This is what the above means to your `pubspec.yaml`:
+
+```yaml
+...
+dependencies:
+  ...
+  shared_preferences: ^0.5.4+8
+  shared_preferences_web: ^0.0.1
```
hterkelsen: update this line

yjbanov: Good catch! Done.

packages/shared_preferences/shared_preferences_web/lib/shared_preferences_web.dart
ditman: Ah ok, I thought this relied on the user prepending "flutter." to the key themselves. If it's the shared_prefs plugin, I guess it's OK.

I know you can't step on other's site data; the scenario I was envisioning was multiple plugins writing to localStorage on the same site, and overwriting each other's data.

packages/shared_preferences/shared_preferences_web/lib/shared_preferences_web.dart
ditman: > In Dart generics are reified (unlike Java, where generics are erased at runtime).

Oooooh man, my sloppy java days showing up :P

packages/shared_preferences/shared_preferences_web/test/shared_preferences_web_test.dart
ditman: This is a functionality from the language that I wasn't expecting at all!

packages/shared_preferences/shared_preferences_web/pubspec.yaml
ditman: I'm not entirely sure, I have a feeling we'll end up coming and updating all of these to the proper version later on, don't worry too much! :P

PR #2339, opened 2019-12-04
[webview_flutter] add gesture navigation for iOS

wwwdata:
## Description

This introduces a boolean for the allowsBackForwardNavigationGestures setting of the WKWebView in iOS. It has no effect on Android.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.19
+
+* Add setting for iOS to allow gesture based navigation.
+
 ## 0.3.18+1
 
 * Be explicit that keyboard is not ready for production in README.md.
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -275,6 +275,8 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
         case "userAgent":
           updateUserAgent((String) settings.get(key));
           break;
+        case "allowsBackForwardNavigationGestures":
+          break;
         default:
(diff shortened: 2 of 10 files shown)


ened:
Hi @wwwdata - thank you for the PR!
It seems like useful addition to the web view plugin.
Could you please rebase & also add a e2e test. You can find the existing e2e test suite in `example/test_driver/webview_flutter_e2e.dart`.
Once both done, we can proceed to the next step. Thank you once again!

wwwdata:
@ened thanks for the fast response. I am a bit confused about the e2e tests. As far as I see it currently, they are basically running some quick unit tests but using the real platform for that, correct? What could I in my case add as a test? In order to test whether it works, I would really have to launch a WebView in iOS then find a link to tap on, which navigates forward, after that use a swipe gesture to navigate back again.

For now, I just added a test that is launching the WebView with my new feature to see that nothing crashes

ened:
@wwwdata That will do it for the moment. Among other behaviour, the key issue is to check whether the native component will load properly and received the right configuration.

Something I'm not sure yet is whether it's ok to no-op the test on Android. Understood it's a non-supported feature on Android, but if it ever will be in the future, the test may be missed.

Besides, the underlying platform implementation should not ignore parameters it doesn't know about.

wwwdata:
> Something I'm not sure yet is whether it's ok to no-op the test on Android. Understood it's a non-supported feature on Android, but if it ever will be in the future, the test may be missed.

You're right will remove the no-op. It should go through and really test that nothing bad happens.

ened:
@wwwdata After reading this a few more times, would you agree that `gestureNavigationEnabled` is more concise & clear to the developer that will use the webview widget?
It also seems to be in line with `debuggingEnabled`.

collinjackson:
I agree with @ened that `gestureNavigationEnabled` works better as a cross-platform name.

wwwdata:
Ok. New name makes sense! Thanks for already making the change.

PR #2343, opened 2019-12-04
[webview_flutter] Fix iOS WebView ignoring SafeArea widget and  iOS 13 WebView scroll indicator showing at random location.

cyanglaz:
## Description

The WKWebView defaults to have `contentInsetAdjustmentBehavior` as automatic. Which means it adjusts its content inset while there is a safe area(notch) on the screen. In Flutter, we use SafeArea widget to handle the safe area. With this default behavior, even without the SafeArea widget, the content is pushed down (portrait mode), which results a bad UI looks like https://github.com/flutter/flutter/issues/25274. This creates an inconsistency with WebView widget and other flutter widgets, the developers might have a hard time figuring out what caused the issue.
By setting `contentInsetAdjustmentBehavior` false, we always ask flutter to handle the content inset and makes the WebView more consistent with flutter's behavior.
Note: This change might result your WebView content inset pull up if when running your app on an iOS device with a notch (Which is intended). Wrapping around the WebView with a SafeArea widget will resolve the issue.

On iOS 13, the `automaticallyAdjustsScrollIndicatorInsets` is introduced. To match with our setting for `contentInsetAdjustmentBehavior, we will set `automaticallyAdjustsScrollIndicatorInsets` to false. This will also fix https://github.com/flutter/flutter/issues/41592 

## Related Issues

https://github.com/flutter/flutter/issues/41592

## Checklist

- [x] I signed the [CLA].

## Breaking Change


sky126:
good

amirh:
Can we test this somehow?

cyanglaz:
@amirh There are 2 tests we can do. 
1. We might be able to add some test to make sure the contentInset and scrollIndicatorInset are always (0,0) for the matter. (This will need to have some `visibleForTesting` method to expose the value of `contentInset` and `scrollIndicatorInset` on iOS)
2. However, for the bug that the scroll indicator always showing at the random location, we can't. Because the value of the ScrollIndicatorInset is correct even when it shows at the wrong location. We might be able to add some screenshot test if we have some infra set up to do so.

cyanglaz:
@amirh I added the tests that we discussed offline. Also waiting for https://github.com/flutter/plugins/pull/2346 to be landed before merging this in so we are sure the tests are passing in CI.

cyanglaz:
The issue seems to be because of https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/rendering/platform_view.dart#L354-L356

cyanglaz:
@bparrishMines This fixes a nasty UI bug on iOS 13. Do you mind taking a look?

PR #2357, opened 2019-12-05
WIP Reduce CI testing time, add caching

jmagman:
## Description

- CocoaPods 1.8 uses a CDN instead of a local spec repository.  `pod repo update` is no longer necessary.  Remove.
- Add pub cache
- Add stable Flutter SDK cache


jmagman:
Attempt #2 of https://github.com/flutter/plugins/pull/2126

jmagman:
> These templates are great -- I didn't know that was available.
> 
> I'd love to find a way to share more of this with Flutterfire somehow, but I'm not sure if there's an easy way to do that.

Sorry you're getting spammed--the only way to test it to keep pushing changes...

PR #2293, opened 2019-11-21
[image_picker]fix a crash when a non-image file is picked.

cyanglaz:
## Description

This PR fixes 2 things:
The scaling is performed even when the quality/maxWidth/maxHeight are null.
When a non-image file is picked, scaling process is performed on the file.

## Related Issues

https://github.com/flutter/flutter/issues/38775
https://github.com/flutter/flutter/issues/40233
b/142680906
b/143851555

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #2362, opened 2019-12-06
[Image_picker] fix rotation regression on Android.

cyanglaz:
## Description

A regression was introduced in https://github.com/flutter/plugins/pull/2293 where the image file seems to lose rotation when not scaled. Reverting to previous approach: directly return the original image file path solves the issue.
The root cause of why image file would lose rotation after compressing is unknown.

## Related Issues

https://github.com/flutter/flutter/issues/45905

## Checklist

- [x] I signed the [CLA].

## Breaking Change


cyanglaz:
@mklim Could you take a look at this one? Should be pretty straight forward.

cyanglaz:
> The previous PR was fixing a crash though, wasn't it? Do we want to completely revert it to fix the rotation issue and then re-introduce the crash? I think it may be better to try to root cause the issue and forward fix it in this case, especially since this has been landed for awhile now.

We didn't completely revert it in this PR. We only revert the part that caused the rotation issue. Which also won't reintroduce the crash.
However, the root cause of rotation issue is that in the last PR we copied the file even if we don't scale. I'm not sure why it caused the rotation issue through.

This PR remove the copy file part if we don't have to do anything with the image (scaling, compression, etc).

ened:
@cyanglaz how could we test this PR? There are no test_driver calls (I guess it's not feasible to pick image in a automated way) plus the example has no fields / configuration for `maxWidth`, `maxHeight` or `imageQuality`.

As a first start, could you possibly add some sort of configuration UI to the example so that manual tests could be done?

cyanglaz:
@mklim I couldn't find a good way to add an integration test for this. I think what @ened mentioned about adding a manual test makes sense. It would help us test the plugin easier manually in the future. I'm going to add one in the example app and ping you once it's done.

ened:
In my mind, the pragmatic way would be to implement a Android instrumentation (not unit) test, that copies a set of pre-defined assets onto a device. The test then runs a ImageResizer implementation against those and checks the output file (like a golden test, really). 
There are some risks like different Android phones (& their JPG encoders) may have a tiny bit of different output, but that should be a solved problem in the testing world.

Secondly, the e2e test should have a way to assert the correct ImageResizer implementation is called. 

All in all quite complicated so perhaps easier to solve the problem first and add a way for developer-users to manually test the function. That is still a lot better than anything we have now. 
Anything we can give people to verify the correct function of a plugin will be helpful.

cyanglaz:
@ened @mklim  Added a dialog to specify maxWidth, maxHeight and quality when picking image in the example app. That should be able to help manual test.

packages/image_picker/example/lib/main.dart
```diff
@@ -249,8 +265,67 @@ class _MyHomePageState extends State<MyHomePage> {
...
     return null;
   }
+
+  Future<void> _displayPickImageDialog(
+      BuildContext context, OnPickImageCallback onPick) async {
+    return showDialog(
+        context: context,
+        builder: (context) {
+          return AlertDialog(
+            title: Text('TextField in Dialog'),
```
mklim: nit: This title isn't really descriptive. "Set image size" maybe?

cyanglaz: good find! Thanks. I updated it `Add optional parameters`

PR #2368, opened 2019-12-06
[url_launcher] Migrate to new pubspec plugins schema and endorse url_launcher_web

amirh:
Migrates url_launcher to use the new pubspec plugins schema, and endorses url_launcher_web as the default web implementation.

https://github.com/flutter/flutter/issues/46253


packages/url_launcher/url_launcher/pubspec.yaml
```diff
@@ -7,9 +7,12 @@ version: 5.2.7
...
-    pluginClass: UrlLauncherPlugin
+    platforms:
+      android:
+        package: io.flutter.plugins.urllauncher
+        pluginClass: UrlLauncherPlugin
+      ios:
+        pluginClass: FLTUrlLauncherPlugin
 
 dependencies:
   flutter:
```
ditman: No new dependency on `url_launcher_web` yet?

amirh: Good point, I guess we might as well do it in this PR as it is already gated on a new enough Flutter

packages/url_launcher/url_launcher/pubspec.yaml
```diff
@@ -3,18 +3,24 @@ description: Flutter plugin for launching a URL on Android and iOS. Supports
...
   plugin:
-    androidPackage: io.flutter.plugins.urllauncher
-    iosPrefix: FLT
-    pluginClass: UrlLauncherPlugin
+    platforms:
+      android:
+        package: io.flutter.plugins.urllauncher
+        pluginClass: UrlLauncherPlugin
+      ios:
+        pluginClass: FLTUrlLauncherPlugin
```
stuartmorgan: Nit: FLTURLLauncherPlugin (same for the filename). The old capitalization was wrong for ObjC, but there was no way to have it be different for ObjC and Java. Now that there is, it should be fixed.

packages/url_launcher/url_launcher/pubspec.yaml
```diff
@@ -13,11 +13,14 @@ flutter:
...
       ios:
         pluginClass: FLTUrlLauncherPlugin
+      web:
+        default_package: url_launcher_web
 
 dependencies:
   flutter:
     sdk: flutter
   url_launcher_platform_interface: ^1.0.1
+  url_launcher_web: any
```
ditman: I'd be specific on this one, but we'll see how 'any' goes for a while!

amirh: See discussion on https://flutter.dev/go/federated-plugins we want this edge to be any

packages/url_launcher/url_launcher/pubspec.yaml
amirh: done

PR #2367, opened 2019-12-06
Temporarily disable CI on stable Flutter

amirh:
To enable using new Flutter features closely before a new release we're temporarily disabling CI on stable.

https://github.com/flutter/flutter/issues/46258 for re-enabling.

cc @jmagman this conflicts with https://github.com/flutter/plugins/pull/2357, we need to disable stable ASAP to unblock endorsing web plugins,  not sure how close is the other PR to landing I'm ok with waiting for the other PR first (or @jmagman if you can land i#2357 with stable disabled already 😄 ).

cc @collinjackson

PR #2373, opened 2019-12-07
Disable stable for check publish

franciscojma86:
Continuation of https://github.com/flutter/plugins/pull/2367 to enable publishing macos plugins with the new pubspec formats.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -14,6 +14,9 @@ task:
   matrix:
     - name: publishable
       script:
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -14,7 +14,7 @@ task:
   matrix:
     - name: publishable
       script:
-        - flutter channel stable
+        - flutter channel master
```
amirh: Can you add the same comment with a TODO to undo this?

PR #2378, opened 2019-12-08
[url_launcher_web] bump the minimal flutter requirement to 1.10.0

amirh:
This is the minimal Flutter version required for the plugin platforms manifest.


amirh:
Presubmit is failing because the version checker isn't happy with a changelog update without a version bump, though the current version wasn't published yet so it should be ok. Landing even though presubmits are red.

PR #2401, opened 2019-12-11
[url_launcher] remove temporary dependency override

amirh:
## Description

Follow-up to https://github.com/flutter/plugins/pull/2399 removing the dependency override.


amirh:
version solver is failing as it expects a version bump.
([it's failing multiple tasks :( ](https://github.com/flutter/flutter/issues/46799))

I'm looking into the build_all_plugins_apk failure

PR #2405, opened 2019-12-12
[connectivity_macos] update title to connectivity_macos

JigneshPatel23:


franciscojma86:
@JigneshPatel23  thanks for your PR! I actually included this change in https://github.com/flutter/plugins/pull/2407, so we can close this now. Turns out more README s needed some love :)

PR #2408, opened 2019-12-12
[video_player_web] add a no-op android folder

amirh:
## Description

Adding an `android/` folder as a temporary mitigation for:

https://github.com/flutter/flutter/issues/46304
https://github.com/flutter/flutter/issues/46790

# Related Issues

https://github.com/flutter/flutter/issues/46898

### DIFF
--- a/packages/video_player/video_player_web/android/.gitignore
+++ b/packages/video_player/video_player_web/android/.gitignore
@@ -1,0 +1,8 @@
+*.iml
+.gradle
+/local.properties
+/.idea/workspace.xml
+/.idea/libraries
+.DS_Store
+/build
+/captures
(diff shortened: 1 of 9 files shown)


packages/video_player/video_player_web/android/.gitignore
```diff
@@ -0,0 +1,8 @@
+*.iml
```
mklim: nit: This can be deleted.

amirh: I'm keeping the template list as the easy choice, FWIW someone may decide to open this project with Android Studio...

mklim: I think our root .gitignore would still be applied here in that case, since AS should be deferring to git which would be able to tell that this is part of a larger repo. But either way it's fine to keep this if you'd prefer, it's not doing any harm and matches the template.

PR #2417, opened 2019-12-13
[plugin_platform_interface] Don't use const Object as a token

amirh:
## Description

Platform interface tokens must be unique, we've mistakenly used const Objects as a token (and all const Objects are identical).

Thanks @buntagonalprism for noticing and investigating!

We should probably follow-up with the PlatformInterface forcing tokens that are not const Objects (though this will be a breaking change).

I'm not sure whether we can test the failing scenario as it will require running tests in release mode.

## Related Issues

https://github.com/flutter/flutter/issues/46941

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/plugin_platform_interface/CHANGELOG.md
+++ b/packages/plugin_platform_interface/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.0.1
 
-* Fixed a bug that made all platform interface apeear as mock on release builds (https://github.com/flutter/flutter/issues/46941).
+* Fixed a bug that made all platform interfaces appear as mocks in release builds (https://github.com/flutter/flutter/issues/46941).
 
 ## 1.0.0 - Initial release.
 
--- a/packages/plugin_platform_interface/CHANGELOG.md
+++ b/packages/plugin_platform_interface/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.1
+
+* Fixed a bug that made all platform interface apeear as mock on release builds (https://github.com/flutter/flutter/issues/46941).
+
 ## 1.0.0 - Initial release.
 
 * Provides `PlatformInterface` with common mechanism for enforcing that a platform interface
--- a/packages/plugin_platform_interface/lib/plugin_platform_interface.dart
+++ b/packages/plugin_platform_interface/lib/plugin_platform_interface.dart
@@ -91,7 +91,7 @@ abstract class PlatformInterface {
 /// ```
 @visibleForTesting
 abstract class MockPlatformInterfaceMixin implements PlatformInterface {
-  static const Object _token = Object();
+  static final Object _token = Object();
 
   @override
   Object get _instanceToken => _token;
--- a/packages/plugin_platform_interface/lib/plugin_platform_interface.dart
+++ b/packages/plugin_platform_interface/lib/plugin_platform_interface.dart
@@ -70,7 +70,8 @@ abstract class PlatformInterface {
       return;
     }
     if (!identical(token, instance._instanceToken)) {
-      throw AssertionError('Platform interfaces must not be implemented with `implements`');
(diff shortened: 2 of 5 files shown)


packages/url_launcher/url_launcher_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.5
+
+* Make the PlatformInterface _token non const (as const Object's are not unique).
```
hterkelsen: fix formatting

amirh:
> You can also check and throw an exception if `token == const Object()` since this could be a common mistake.

Yes, I mentioned this in the PR description, I'll followup in a separate PR as we should make sure existing plugins are already updated (I think it's early enough to assume the currently only first party plugins are using this)

amirh:
Actually this doesn't work as now the token doesn't test equal for the different instances of MockPlatformInterface, fixing...

packages/plugin_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.1
+
+* Fixed a bug that made all platform interface apeear as mock on release builds (https://github.com/flutter/flutter/issues/46941).
```
collinjackson:
```suggestion
* Fixed a bug that made all platform interfaces appear as mocks in release builds (https://github.com/flutter/flutter/issues/46941).
```

packages/url_launcher/url_launcher_platform_interface/CHANGELOG.md
collinjackson:
```suggestion
* Make the `PlatformInterface` `_token` non `const` (as `const` `Object`s are not unique).
```

packages/url_launcher/url_launcher_platform_interface/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: A common platform interface for the url_launcher plugin.
 homepage: https://github.com/flutter/plugins/tree/master/packages/url_launcher/url_launcher_platform_interface
 # NOTE: We strongly prefer non-breaking changes, even at the expense of a
 # less-clean API. See https://flutter.dev/go/platform-interface-breaking-changes
-version: 1.0.4
+version: 1.0.5
```
collinjackson: You'll also want to update the dependency on plugin_platform_interface in this file.

packages/plugin_platform_interface/lib/plugin_platform_interface.dart
```diff
@@ -91,8 +91,6 @@ abstract class PlatformInterface {
 /// ```
 @visibleForTesting
 abstract class MockPlatformInterfaceMixin implements PlatformInterface {
-  static final Object _token = Object();
-
   @override
-  Object get _instanceToken => _token;
+  Object get _instanceToken => null;
```
collinjackson: Probably can just delete this

packages/url_launcher/url_launcher_platform_interface/pubspec.yaml
amirh: Good point, I'll have to either split this PR into 2, or follow-up with another PR just to bump after publishing the plugin_platform_interface. I'm going with the later as it's the faster path right now.

packages/plugin_platform_interface/lib/plugin_platform_interface.dart
amirh: done

packages/url_launcher/url_launcher_platform_interface/pubspec.yaml
amirh: Actually ended up splitting the PRs

PR #1389, opened 2019-03-23
[webview_flutter] add events: onPageStarted, onReceivedError (with some simple error handling) and action: stopLoading.

hpoul:
## Description

* Adds new events to WebView for `onPageStarted` and `onReceivedError`
   * Tries to combine error handling between iOS/Android
* add `stopLoading` of webview.

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #2420, opened 2019-12-15
[webview_flutter] Support basic auth requests

wwwdata:
## Description

This enables the webview to visit http basic auth protected websites.

## Related Issues

Originally, it was attempted to implement this here: https://github.com/flutter/plugins/pull/1283
But the author then decided to split up the features into smaller pull requests. However, the one for auth requests was never opened, only for the other features: https://github.com/flutter/plugins/pull/1389

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/webview_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.11
+
+* Add support for basic auth protected websites.
+
 ## 2.0.10
 
 * Fix keyboard issues link in the README.
@@ -5,7 +9,6 @@
 ## 2.0.9
 
 * Add iOS UI integration test target.
-* Suppress deprecation warning for iOS APIs deprecated in iOS 9.
 
 ## 2.0.8
 
--- a/packages/webview_flutter/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
+++ b/packages/webview_flutter/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
@@ -10,6 +10,7 @@ import android.graphics.Bitmap;
 import android.os.Build;
 import android.util.Log;
 import android.view.KeyEvent;
+import android.webkit.HttpAuthHandler;
 import android.webkit.WebResourceError;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebView;
@@ -157,6 +158,40 @@ class FlutterWebViewClient {
     }
   }
 
+  private void onReceivedHttpAuthRequest(final HttpAuthHandler handler, String host, String realm) {
+    HashMap<String, String> arguments = new HashMap<>();
+    arguments.put("host", host);
+    arguments.put("realm", realm);
+    methodChannel.invokeMethod(
+        "onReceivedHttpAuthRequest",
+        arguments,
+        new MethodChannel.Result() {
+          @Override
+          public void success(Object args) {
+            if (args instanceof Map) {
+              Map<?, ?> argsMap = (Map<?, ?>) args;
+              Object username = argsMap.get("username");
+              Object password = argsMap.get("password");
+              if (username != null && password != null) {
+                handler.proceed(username.toString(), password.toString());
+                return;
+              }
+            }
+            handler.cancel();
+          }
+
+          @Override
+          public void error(String error, String detailedException, Object args) {
+            handler.cancel();
+          }
+
+          @Override
+          public void notImplemented() {
+            handler.cancel();
+          }
+        });
+  }
+
   // This method attempts to avoid using WebViewClientCompat due to bug
   // https://bugs.chromium.org/p/chromium/issues/detail?id=925887. Also, see
   // https://github.com/flutter/flutter/issues/29446.
@@ -208,6 +243,12 @@ class FlutterWebViewClient {
         // handled even though they were handled. We don't want to propagate those as they're not
         // truly lost.
       }
+
+      @Override
+      public void onReceivedHttpAuthRequest(
+          WebView view, HttpAuthHandler handler, String host, String realm) {
+        FlutterWebViewClient.this.onReceivedHttpAuthRequest(handler, host, realm);
+      }
     };
   }
 
@@ -248,6 +289,12 @@ class FlutterWebViewClient {
       public void onReceivedError(
           WebView view, int errorCode, String description, String failingUrl) {
         FlutterWebViewClient.this.onWebResourceError(errorCode, description, failingUrl);
+      }
+
+      @Override
+      public void onReceivedHttpAuthRequest(
+          WebView view, HttpAuthHandler handler, String host, String realm) {
+        FlutterWebViewClient.this.onReceivedHttpAuthRequest(handler, host, realm);
       }
 
       @Override
--- a/packages/webview_flutter/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
+++ b/packages/webview_flutter/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
@@ -172,10 +172,8 @@ class FlutterWebViewClient {
               Map<?, ?> argsMap = (Map<?, ?>) args;
               Object username = argsMap.get("username");
               Object password = argsMap.get("password");
-              if (username != null && password != null) {
-                handler.proceed(username.toString(), password.toString());
-                return;
-              }
+              handler.proceed(username.toString(), password.toString());
+              return;
             }
             handler.cancel();
           }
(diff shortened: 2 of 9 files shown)


FuturistLabs:
Any idea when this feature will be available?
Thanks in advance!

wwwdata:
> Any idea when this feature will be available?
> Thanks in advance!

As soon as someone from Google presses the green merge button here... Which unfortunately takes a very long time.

JJibreel:
@wwwdata thank you for working on this feature. It seems that its nearing completion, cant wait!

stuartmorgan:

It looks like this should be pretty straightforward to review, and has good test coverage. If you're still interested in moving forward with this, once you've updated this to apply to the current version of the plugin I can review it.

wwwdata:
Hi, I will rebase this then and make it up to date when I have time.

wwwdata:
@stuartmorgan ok I just rebased it and fixed all the conflicts.

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -36,6 +36,9 @@ abstract class WebViewPlatformCallbacksHandler {
 
   /// Report web resource loading error to the host application.
   void onWebResourceError(WebResourceError error);
+
+  /// Invoked by [WebViewPlatformController] when a page requests authorization.
+  AuthInfo? onReceivedHttpAuthRequest(String host, String realm);
```
stuartmorgan, line 41: This class is public API, so adding a new method without an implementation is a breaking change unless I'm missing something. This should probably have a default implementation that returns null.

packages/webview_flutter/lib/src/webview_method_channel.dart
```diff
@@ -25,7 +26,7 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
   static const MethodChannel _cookieManagerChannel =
       MethodChannel('plugins.flutter.io/cookie_manager');
 
-  Future<bool?> _onMethodCall(MethodCall call) async {
+  Future<dynamic> _onMethodCall(MethodCall call) async {
```
stuartmorgan, line 29: `Object?`

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -833,3 +850,15 @@ void _validateUrlString(String url) {
     throw ArgumentError(e);
   }
 }
+
+/// Simple credentials wrapper for authorization requests
+class AuthInfo {
```
stuartmorgan, line 855: As a top-level public class, this should probably have a less collision-prone name, like `WebViewAuthInfo`.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -833,3 +850,15 @@ void _validateUrlString(String url) {
...
+  /// Creates a new AuthInfo
+  AuthInfo(this.username, this.password);
```
stuartmorgan, line 857: const

stuartmorgan, line 857: Since these have the same type, it would be safer to use `required` named arguments rather than positional arguments.

stuartmorgan-g (changes requested):
Small comments below, but more importantly: now that I'm reviewing this in detail I see that the test coverage isn't actually complete; it only has Dart unit tests, so none of the native code is being exercised. This should have an integration test if at all possible (and failing that, native unit tests to complement the Dart unit test).

stuartmorgan:
(FYI the mass of CI failures is from an infrastructure issue we hit; the next time you rebase it'll clear up.)

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
```diff
@@ -157,6 +158,40 @@ private void notifyOnNavigationRequest(
...
+        "onReceivedHttpAuthRequest",
+        arguments,
+        new MethodChannel.Result() {
+          @Override
+          public void success(Object o) {
+            if (o instanceof Map) {
+              Map<?, ?> map = (Map<?, ?>) o;
+              Object username = map.get("username");
+              Object password = map.get("password");
+              if (username != null && password != null) {
```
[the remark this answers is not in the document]
wwwdata: I don't know. How would I report that? This is basically handling the response from dart where it should definitely always be the case that `username` and `password` were sent. I am just checking it because the types are always optional when you send them over the MethodChannel.

packages/webview_flutter/lib/platform_interface.dart
wwwdata: There is only one implementation in the `_PlatformCallbacksHandler` of the webview flutter widget. Where the new method is implemented. So it's not breaking anything, or?

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -151,6 +151,10 @@ typedef void PageLoadingCallback(int progress);
 /// Signature for when a [WebView] has failed to load a resource.
 typedef void WebResourceErrorCallback(WebResourceError error);
 
+/// Signature for when a [WebView] is requesting authentication.
+typedef WebViewAuthInfo OnReceivedHttpAuthRequestCallback(
```
hpoul, line 155: Is there any reason this has to be synchronous? Wouldn't it make more sense to have this `async` (ie. returning `Future<WebViewAuthInfo>`) I could imagine that retrieving credentials (or asking the user) will not always be possible synchronously.

wwwdata: Yes, that's what I also thought. I will change it to be async.

hpoul: 👍️ just discovered that it should probably be `Future<WebViewAuthInfo?>` return type (ie. nullable), right now the callback has no way of indicating that it has no auth info available?

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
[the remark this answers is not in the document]
stuartmorgan: If the Dart side cannot send null, you can just assert that they aren't null rather than checking.

packages/webview_flutter/lib/platform_interface.dart
stuartmorgan: > There is only one implementation

If an interface is part of a public API surface, then you have no way of knowing how many implementations there are.

wwwdata: ok added a null default

wwwdata:
just pushed my changes again. I hope it's ok now. Sorry for the delay I was on vacation

hpoul:
@stuartmorgan is there still something missing in this PR? 🤔️ (except fixing the changelog conflict) thanks.

stuartmorgan-g (commented):
I'm sorry this review fell through the cracks again 🙁 

Overall this looks good; the complication now is that the plugin was converted to fully federated mode in the intervening time so this will now need to be [split into multiple PRs](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins). There's also work underway at the moment to change the communication channels, which would create a second round of conflicts.

/cc @bparrishMines How would you prefer to move forward here relative to your channel work? I think the options are:
1) we split the PR ASAP, get it landed, and then rebase your work with the changes, or
2) you finish your method channel changes, then we update this PR to adapt to both sets of changes at once.

Either way @wwwdata, I can take care of the updating and splitting of this PR if you'd prefer, since all the added work here is due to my losing track of the review.

stuartmorgan:
Per discussion with @bparrishMines I'm going to hold off on updating this until the in-progress conversion is complete.

hpoul:
Is there still a conversion in progress? I guess the original federated plugin changes are done? @stuartmorgan  I have seen the iOS implementation has still some refactoring in progress? https://github.com/flutter/plugins/pull/4626
I guess working on this still doesn't make sense while that is still going on? 🤔️

stuartmorgan:
Triage update: iOS webview conversion is still in progress, so this still blocked for the moment.

Hixie:
Latest status is that we're still waiting on https://github.com/flutter/flutter/issues/93732.
Before this can land even after that is fixed, the conflicts will need to be resolved.

stuartmorgan:
The Dart conversion is complete, but in the meantime we found https://github.com/flutter/flutter/issues/94051 which blocks this, since we currently implement WebViewPlatformCallbacksHandler in our implementations, making changes to it breaking changes. We're actively working on that issue, and hope to have it resolved soon.

stuartmorgan:
Status update from triage: this is still blocked, but we hope to have it unblocked soon.

stuartmorgan:
This is unblocked now that v4 has landed. @bparrishMines This is one we should look at updating.

stuartmorgan:
Update from triage: this is still on our list to update; we're currently working through updating a couple of higher-priority webview features, but we'll get to this soon.

stuartmorgan-g:

Per discussion above about this PR having fallen through the cracks, either I or @bparrishMines will be opening an updated version of this PR in flutter/packages, so there's no need for any action here. Per my comment above, we expect to get to this fairly soon.

PR #2430, opened 2019-12-18
[image_picker] support android V2 embedding

cyanglaz:
https://github.com/flutter/flutter/issues/41839

aside from migration, I also moved the code in onActivitySaveInstanceState to onStop to match with the `DefaultLifecycleObserver`.

I think onStop is enough for our needs in this plugin.


packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerPlugin.java
```diff
@@ -10,13 +10,86 @@
...
+    public void onCreate(@NonNull LifecycleOwner owner) {}
+
+    @Override
+    public void onStart(@NonNull LifecycleOwner owner) {}
+
+    @Override
+    public void onResume(@NonNull LifecycleOwner owner) {}
+
+    @Override
+    public void onPause(@NonNull LifecycleOwner owner) {}
```
mklim: Why override these empty methods?

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerPlugin.java
```diff
@@ -10,13 +10,86 @@
...
+
+    @Override
+    public void onStop(@NonNull LifecycleOwner owner) {
+      onActivityStopped(thisActivity);
```
mklim: nit: `onStop` isn't really guaranteed to be called before the app is closed. If it's important that `delegate.saveStateBeforeResult` every time the activity is closed, `onPause` is a little safer. Same with `onActivityDestroyed`. That would mean re-registering the callbacks in `onResume` though.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerPlugin.java
```diff
@@ -27,82 +100,132 @@
   private static final int SOURCE_CAMERA = 0;
   private static final int SOURCE_GALLERY = 1;
 
-  private final PluginRegistry.Registrar registrar;
+  private MethodChannel channel;
   private ImagePickerDelegate delegate;
-  private Application.ActivityLifecycleCallbacks activityLifecycleCallbacks;
+  private FlutterPluginBinding pluginBinding;
+  private ActivityPluginBinding activityBinding;
+  private Application applicationContext;
```
mklim: nit: An `Application` and an `ApplicationContext` are two separate things, so this naming is a little confusing. Do you really need this to be an `Application`, or are you just trying to get an `ApplicationContext`? If it's the latter I would just stick with a `Context` for the type here.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerPlugin.java
```diff
@@ -27,82 +100,132 @@
...
+    this.activity = activity;
+    this.applicationContext = applicationContext;
+    this.delegate = constructDelegate(activity);
+    channel = new MethodChannel(messenger, CHANNEL);
+    channel.setMethodCallHandler(this);
+    if (registrar != null) {
+      // V1 embedding setup for activity listeners.
+      observer = new LifeCycleObserver(activity);
+      applicationContext.registerActivityLifecycleCallbacks(
+          observer); // Use getApplicationContext() to avoid casting failures.
```
mklim: nit: extra comment here

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerPlugin.java
```diff
@@ -27,82 +100,132 @@
...
+      final Application applicationContext,
+      final Activity activity,
+      final PluginRegistry.Registrar registrar,
+      final ActivityPluginBinding activityBinding) {
+    this.activity = activity;
+    this.applicationContext = applicationContext;
+    this.delegate = constructDelegate(activity);
+    channel = new MethodChannel(messenger, CHANNEL);
+    channel.setMethodCallHandler(this);
+    if (registrar != null) {
```
mklim: nit: some of the statements in the below if/then blocks are the same, like `observer =`. It would be a little clearer if they were pulled out and the only things in this block were the ones that needed to be different.

packages/image_picker/example/android/app/src/main/java/io/flutter/plugins/imagepickerexample/MainActivity.java
```diff
@@ -1,18 +1,23 @@
-// Copyright 2017 The Chromium Authors. All rights reserved.
+
```
mklim: With the new Flutter stable we no longer need this. We can just refer to `FlutterActivity` in the manifest.

packages/image_picker/example/test_driver/test/image_picker_e2e_test.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright 2019, the Chromium project authors.  Please see the AUTHORS file
```
mklim: This is the test harness, but I'm not actually seeing specific e2e test cases in this PR. Would it make sense to try and test this more?

noborder:
Thank you cyanglaz. I tested it. It will solve the problem

packages/image_picker/example/test_driver/test/image_picker_e2e_test.dart
cyanglaz: All the image picker methods are static and the Future won;'t return until some sort of user interaction. So it might be hard to test at the current state.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerPlugin.java
mklim: :man_facepalming: because it's `implements`, not `extends`. My mistake. This class does need all these empty methods defined still, sorry.

packages/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerPlugin.java
mklim: If it's not critical that this fires in every single case I think `onStop` makes the most sense and this is good as-is. That's the recommended callback for relatively CPU-intensive work like saving data. https://developer.android.com/guide/components/activities/activity-lifecycle#onstop

packages/image_picker/example/test_driver/test/image_picker_e2e_test.dart
mklim: Ah, right. In that case having it at least test the binding is better than nothing and makes sense, thanks.

xster:
\o/

PR #2439, opened 2019-12-25
Fix app crash when biometrics lockout on ios 13

duynp117:
When biometrics lockout on ios 13. It's not show native dialog, it's crash application.


duynp117:
@googlebot I signed it!

duynp117:
@googlebot I fixed it.

duynp117:
@googlebot I signed it!

PR #2441, opened 2019-12-25
[local_auth]Fix app crash when biometrics lockout on ios 13

duynp117:
Fix app crash when biometrics lockout on ios 13


cyanglaz:
@duynp117 Thanks for contributing. Is there a github issue that we can link?

GenixPL:
There seems to be another issue for that [(https://github.com/flutter/plugins/pull/2760)](https://github.com/flutter/plugins/pull/2760).

Why hasn't it been merged by now?

fransboumans:
Any update on this? Still crashing after 5th attempt. I'm trying to use this on iOS 14.2.

stuartmorgan:
Removing error handling is definitely not the desired fix here; closing in favor of 2760 which is much closer to correct.

PR #2442, opened 2019-12-25
[local_auth] Fix Get Available Biometrics Android

duynp117:
Issue get available biometrics function:
- From iOS: Biometrics type return from enroll of device.
- From Android: Biometrics type return from system of device.

Solved: change get available biometrics type of Android from system to enroll. It's same with iOS and working ok.

https://github.com/flutter/flutter/issues/45497


cyanglaz:
@duynp117 Thanks for contributing. Is there a github issue that we can link?

MartimTSilva:
When is this going to be avaible? Without this, all Android devices with fingerprint scanner on Android 23 doesn't return fingerprint on getAvaibleBiometrics...

stuartmorgan:
Thanks for the contribution. I'm going to close this in favor of https://github.com/flutter/plugins/pull/3763, which also includes code to address the newer biometric types that have been added.

MartimTSilva:
> Thanks for the contribution. I'm going to close this in favor of #3763, which also includes code to address the newer biometric types that have been added.

Hello, is this issue fixed? I've tried it and has the same issue. On Android 6 doesn't return fingerprint on getAvaibleBiometrics. Can you help? Thanks

PR #2453, opened 2020-01-08
[android_alarm_manager] Fixed issue where callback lookup was failing in the background

bkonyi:
If callback lookup was attempted before Flutter was initialized, the callback cache would not yet be populated resulting in a failed callback lookup.

Fixes https://github.com/flutter/flutter/issues/47406

### DIFF
--- a/packages/android_alarm_manager/CHANGELOG.md
+++ b/packages/android_alarm_manager/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.5+3
+
+* Fixed issue where callback lookup would fail while running in the background.
+
 ## 0.4.5+2
 
 * Remove the deprecated `author:` field from pubspec.yaml
--- a/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/FlutterBackgroundExecutor.java
+++ b/packages/android_alarm_manager/android/src/main/java/io/flutter/plugins/androidalarmmanager/FlutterBackgroundExecutor.java
@@ -151,17 +151,22 @@ public class FlutterBackgroundExecutor implements MethodCallHandler {
       return;
     }
 
-    FlutterCallbackInformation flutterCallback =
-        FlutterCallbackInformation.lookupCallbackInformation(callbackHandle);
-    if (flutterCallback == null) {
-      Log.e(TAG, "Fatal: failed to find callback");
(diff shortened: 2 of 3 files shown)


bkonyi:
> Any way to test this? I get that this is inherently kind of tricky to try and e2e test since it needs to be backgrounded at some point. Would it be possible to unit test `FlutterBAckgroundExecutor#startBackgroundIsolate`? All PRs need to have automated tests before we can merge them into the tree, see [Tree Hygiene #4](https://github.com/flutter/flutter/wiki/Tree-hygiene#overview).

I honestly wouldn't know how to go about doing that since my Android testing experience is non-existent. We'd need to be able to run the application in the foreground once to initialize state, kill it, and then wait for it to come up.

mklim:
@bkonyi I typed up a long comment explaining how to write Java unit tests in our repo, looked at the method more closely, and then realized that it's doing way too much work with JNI and global static state to be really unit tested anyway. Disregard that suggestion. :)

> We'd need to be able to run the application in the foreground once to initialize state, kill it, and then wait for it to come up.

I think we could potentially do this in a Dart e2e test. Check out [camera_e2e_test.dart](https://github.com/flutter/plugins/blob/master/packages/camera/example/test_driver/camera_e2e_test.dart), where we use adb from the test runner to grant the test app permissions before it runs. What if we killed the test app with `adb shell force-stop` from the test runner and then asserted that it came back after some amount of time? I'm thinking that should work but I could be missing something here, WDYT?

bkonyi:
> I think we could potentially do this in a Dart e2e test. Check out [camera_e2e_test.dart](https://github.com/flutter/plugins/blob/master/packages/camera/example/test_driver/camera_e2e_test.dart), where we use adb from the test runner to grant the test app permissions before it runs. What if we killed the test app with `adb shell force-stop` from the test runner and then asserted that it came back after some amount of time? I'm thinking that should work but I could be missing something here, WDYT?

Hm... I think that would work. Let me hack something together and give it a shot.

Hixie:
You know, the thing that would be really good to test is whether it comes back after a reboot. I don't suppose we can reboot the phone as well during the test? :-)

bkonyi:
Filed an issue to add tests when possible: https://github.com/flutter/flutter/issues/48439

PR #2456, opened 2020-01-09
[url_launcher] Handle invalid URLs on iOS

rdev-software:
## Description
Add guard  in case of invalid url address format, and return correct value.

## Breaking Change

### DIFF
--- a/packages/url_launcher/url_launcher/ios/Classes/FLTURLLauncherPlugin.m
+++ b/packages/url_launcher/url_launcher/ios/Classes/FLTURLLauncherPlugin.m
@@ -102,6 +102,10 @@ API_AVAILABLE(ios(9.0))
 
 - (BOOL)canLaunchURL:(NSString *)urlString {
   NSURL *url = [NSURL URLWithString:urlString];
+  if (url == NULL)
+  {
+    return NO;
+  }
   UIApplication *application = [UIApplication sharedApplication];
   return [application canOpenURL:url];
 }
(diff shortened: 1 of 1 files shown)


rdev-software:
@googlebot I signed it!

packages/url_launcher/url_launcher/ios/Classes/FLTURLLauncherPlugin.m
```diff
@@ -102,6 +102,9 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
 
 - (BOOL)canLaunchURL:(NSString *)urlString {
   NSURL *url = [NSURL URLWithString:urlString];
+  if (url == NULL) {
```
stuartmorgan, line 105: An ObjC object type should be compared to `nil`, not `NULL`.

packages/url_launcher/url_launcher/ios/Classes/FLTURLLauncherPlugin.m
```diff
@@ -102,6 +102,9 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
+    return NO;
```
stuartmorgan, line 106: An invalid URL should return an error rather than `NO`, as it indicates programmer error rather than just not having a handler present.

stuartmorgan (commented):
Thanks for the contribution! We apologize for the long delay in triaging this PR. We’re in the process of overhauling our PR triage system to respond much more quickly, as well as working through the backlog.

I've left a couple of specific comments below; in addition, this will need test coverage of the issue being fixed, per Flutter policy, as well as a version and CHANGELOG update so that it can be published.

stuartmorgan:

PR #2462, opened 2020-01-11
Custom Cursor for [mac] and [web]

rodydavis:
Example:

```dart
import 'dart:io';

import 'package:flutter/foundation.dart';

import 'package:custom_cursor/custom_cursor.dart';

void main() => runApp(MyApp());

class MyApp extends StatefulWidget {
  @override
  _MyAppState createState() => _MyAppState();
}

  @override
    return MaterialApp(
      home: Scaffold(
        appBar: AppBar(
          title: const Text('Plugin example app'),
        ),
        body: Center(
          child: Container(
            width: 400,
            child: _buildBody(),
          ),
        ),
      ),
    );
  }

  Widget _buildBody() {
    if (kIsWeb) {
      return _buildWebBody();
    }
    if (Platform.isMacOS) {
      return _buildMacOSBody();
    }
    return Wrap(
      children: CursorType.values.map((t) {
        return Padding(
          padding: const EdgeInsets.all(8.0),
          child: MouseRegion(
            onEnter: (_) => CustomCursorPlugin().setCursor(t),
            onExit: (_) => CustomCursorPlugin().resetCursor(),
            child: Text(describeEnum(t)),
          ),
        );
      }).toList(),
    );
  }

  Widget _buildWebBody() {
    return Wrap(
      children: WebCursor.values.map((t) {
        return Padding(
          padding: const EdgeInsets.all(8.0),
          child: MouseRegion(
            onEnter: (_) =>
                CustomCursorPlugin().setWebCursor(WebCursor.custom(t)),
            onExit: (_) => CustomCursorPlugin().resetCursor(),
            child: Text(t),
          ),
        );
      }).toList(),
    );
  }

  Widget _buildMacOSBody() {
    return Wrap(
      children: MacOSCursor.values.map((t) {
        return Padding(
          padding: const EdgeInsets.all(8.0),
          child: MouseRegion(
            onEnter: (_) =>
                CustomCursorPlugin().setMacOSCursor(MacOSCursor.custom(t)),
            onExit: (_) => CustomCursorPlugin().resetCursor(),
            child: Text(t),
          ),
        );
      }).toList(),
    );
  }
}

```

Supports all web cursors: 

```
       'all-scroll',
        'auto',
        'cell',
        'context-menu',
        'col-resize',
        'copy',
        'crosshair',
        'default',
        'e-resize',
        'ew-resize',
        'grab',
        'grabbing',
        'help',
        'move',
        'n-resize',
        'ne-resize',
        'nesw-resize',
        'ns-resize',
        'nw-resize',
        'nwse-resize',
        'no-drop',
        'none',
        'not-allowed',
        'pointer',
        'progress',
        'row-resize',
        's-resize',
        'se-resize',
        'sw-resize',
        'text',
        'w-resize',
        'wait',
        'zoom-in',
        'zoom-out',
```

Support all MacOS system cursors:

```
       'arrow',
        'beam-vertical',
        'cross-hair',
        'closed-hand',
        'open-hand',
        'pointing-hand',
        'resize-left',
        'resize-right',
        'resize-down',
        'resize-up',
        'resize-left-right',
        'resize-up-down',
        'beam-horizontial',
        'disappearing-item',
        'not-allowed',
        'drag-link',
        'drag-copy',
        'context-menu',
```

And includes an enum with all shared cursors across platforms:

```dart
enum CursorType {
  arrow,
  cross,
  hand,
  resizeLeft,
  resizeRight,
  resizeDown,
  resizeUp,
  resizeLeftRight,
  resizeUpDown,
  custom,
}
```

In the example I include how to do a mouse region and set and unset the cursor.


rodydavis:
@googlebot I fixed it

rodydavis:
@googlebot I fixed it.

stuartmorgan:
Custom cursor support is being built into the framework; it doesn't make sense to make a new plugin that we would remove again in the near future. If you want to contribute to the implementation the way forward would be to coordinate with @dkwingsmt on the design and/or implementation of the framework feature.

rodydavis:
I needs platform channels though? Where would it live? The engine, or material?

stuartmorgan:
I'm not sure what you're asking exactly. It would use a system channel presumably, like the other framework features that require native embedding code.

rodydavis:
One nice thing about it being a plugin is that it can be customized, easily extended and also the source separate from the engine so that others can add custom cursors that are beyond the system ones.

rodydavis:
@googlebot I fixed it.

stuartmorgan:
> One nice thing about it being a plugin is that it can be customized, easily extended

Since plugins allow extending the capabilities of the engine+framework, I'm not understanding why building the functionality into Flutter would prevent that. Can you provide concrete examples of things that would be possible if it's a plugin, but not if it's built in?

> and also the source separate from the engine so that others can add custom cursors that are beyond the system ones.

Again, I'm not clear on how these things are coupled. Why do you believe it will be impossible to add non-system cursors if it's not a plugin?

rodydavis:
I don't think it is impossible to add if it is in the engine, just if it is a plugin you can have the calls in the same place. Otherwise if you do a custom cursor you will have to set up the platform channel anyway. 

One example of something you can do if it is a plugin:

- Cursor Loaded from Image or SVG

But I think if it is apart of the framework that is correct too. Also it could prevent conflicts of different plugins updating the cursor at the same time.

I just think if someone wants to update a cursor with their own branding or images then they will need a plugin regardless. Maybe if it is in the engine it could just expose loading it from an asset then you are right there would be no need for a plugin.

orestesgaolin:
Just to let everybody know - there's a new PR in flutter repo that adds custom cursor support https://github.com/flutter/flutter/pull/54171

cyanglaz:

PR #2474, opened 2020-01-16
[in_app_purchase] Remove unused method from example app

amirh:
This triggers an analysis warning that makes CI red.


amirh:
Landing on red, as this is one of the causes the tree is red. (it's not the only cause so it's expected that the tree will be red after this is merged as well)

PR #2475, opened 2020-01-16
Bump roboelectric version

dnfield:
Fixes CI

Fixes https://github.com/flutter/flutter/issues/48943

Maven has changed their access policies as of today: https://support.sonatype.com/hc/en-us/articles/360041287334

The old version of roboelectric is refering to http resources in its POM.  This one is good.

@amirh

### DIFF
--- a/packages/image_picker/CHANGELOG.md
+++ b/packages/image_picker/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.6.3+2
+
+* Bump RoboElectric dependency to 4.3.1 and update resource usage.
+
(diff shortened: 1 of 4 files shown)


amirh:
Thanks!

amirh:
Though I hope that major version bump doesn't imply a breaking change that breaks our tests...

dnfield:
It it does we can try to find another version.  This seemed to work locally but I'm not sure I actually ran the tests :)

amirh:
In presubmit we trust... today

PR #2479, opened 2020-01-21
[google_maps_flutter] Dummy map initialization

otopba:
## Description

`MapView` initialization takes a lot of time. This is well know issue. Cause before `getMapAsync` you should call `onCreate` and it is takes a lot of time.

I made time measurements:

```java
private static int attempt = 0;

  GoogleMapController build(
      int id, Context context, AtomicInteger state, PluginRegistry.Registrar registrar) {
    final GoogleMapController controller =
        new GoogleMapController(id, context, state, registrar, options);
    long startTime = System.currentTimeMillis();
    controller.init();
    Log.e("!!!","Attempt " + attempt + ". Init takes " + (System.currentTimeMillis() - startTime) + "ms");
    attempt++;
 ...
```

On my OnePlus 6T I have this results:

```
Attempt 0. Init takes 374ms
Attempt 1. Init takes 54ms
Attempt 2. Init takes 60ms
Attempt 3. Init takes 49ms
Attempt 4. Init takes 52ms

Attempt 0. Init takes 400ms
Attempt 1. Init takes 51ms
Attempt 2. Init takes 57ms
Attempt 3. Init takes 50ms
Attempt 4. Init takes 66ms

Attempt 0. Init takes 337ms
Attempt 1. Init takes 63ms
Attempt 2. Init takes 51ms
Attempt 3. Init takes 41ms
Attempt 4. Init takes 47ms
```

So, first initialization takes 300-400ms on UI thread.

I made code that creates dummy map fragment. This fragment is initialized asynchronously. Thus, upon the first call of the map, all services will be already initialized (unless of course the map is the application’s start screen)

In the future, it’s worth switching to `MapFragment` instead of `MapViev`. Or wait for map library updates

## Related Issues

https://github.com/flutter/flutter/issues/28493

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.22
+
+* Add dummy map initialization
```
sanekyy: May be worth clarifying that only for Android.

Like this:
image

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
```diff
@@ -94,5 +100,34 @@ public void onActivityDestroyed(Activity activity) {
 
   private GoogleMapsPlugin(Registrar registrar) {
     this.registrarActivityHashCode = registrar.activity().hashCode();
+    if (dummyMapInitialized) {
+      return;
+    }
+    Activity activity = registrar.activity();
+    if (activity == null) {
```
sanekyy: Looks like after first line of this method activity can't be null

packages/google_maps_flutter/CHANGELOG.md
otopba: Done

packages/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
otopba: image

sanekyy: Yes. But if it will be null, crash occurs on line 102.

otopba: You are right. Fixed

amirh:
Won't this will result in a performance hit on startup of any application that's using the google maps plugin? e.g if my app is only using a map in some internal page which is not commonly visited by users I may not prefer to pay the cost at startup. If there's a reason to allow "pre-warming" a map - I'd rather give the app developer control over when to do so.

As for how to do the prewarm itself - would it be reasonable to actually load the actual map that we want to show, only without showing it? (possibly in an [Offstage](https://master-api.flutter.dev/flutter/widgets/Offstage-class.html))

otopba:
It’s a good idea to make a separate method for warming up.

But I think it makes no sense to render a real map in Offstage, because anyway, work will be done on a UI thread

otopba:
> Won't this result in a performance hit on startup of any application that's using the google maps plugin? e.g if my app is only using a map in some internal page which is not commonly visited by users I may not prefer to pay the cost at startup. If there's a reason to allow "pre-warming" a map - I'd rather give the app developer control over when to do so.
> 
> As for how to do the prewarm itself - would it be reasonable to actually load the actual map that we want to show, only without showing it? (possibly in an [Offstage](https://master-api.flutter.dev/flutter/widgets/Offstage-class.html))


@amirh 
It’s a good idea to make a separate method for warming up.

But I think it makes no sense to render a real map in Offstage, because anyway, work will be done on a UI thread

amirh:
> > Won't this result in a performance hit on startup of any application that's using the google maps plugin? e.g if my app is only using a map in some internal page which is not commonly visited by users I may not prefer to pay the cost at startup. If there's a reason to allow "pre-warming" a map - I'd rather give the app developer control over when to do so.
> > As for how to do the prewarm itself - would it be reasonable to actually load the actual map that we want to show, only without showing it? (possibly in an [Offstage](https://master-api.flutter.dev/flutter/widgets/Offstage-class.html))
> 
> @amirh
> It’s a good idea to make a separate method for warming up.
> 
> But I think it makes no sense to render a real map in Offstage, because anyway, work will be done on a UI thread

UI thread is not doing much in this case, the platform thread is. Anyway sounds like we agree we don't want to opt-in all maps users to prewarm maps on app startup. I'm totally open to discussion on how to do the prewarm and agree Offstage may not be optimal (depending on the use case).

otopba:
> UI thread is not doing much in this case, the platform thread is. Anyway sounds like we agree we don't want to opt-in all maps users to prewarm maps on app startup. I'm totally open to discussion on how to do the prewarm and agree Offstage may not be optimal (depending on the use case).

@amirh 
MapView initialization takes place on the Android main thread. This is required by the sdk implementation of GooogleMaps. Because main thread loads at about 500ms - Flutter rendering stops.

It doesn’t depend on how we draw the map, in Offstage or not, initialization still occurs in the Android main thread.

cyanglaz:
@otopba I wonder what would people do if they are working on a pure Android App. Do they also call certain method to "warm up" the google map service?

otopba:
> @otopba I wonder what would people do if they are working on a pure Android App. Do they also call certain method to "warm up" the google map service?

@cyanglaz 
Yeah. This is well know issue.
https://stackoverflow.com/questions/26178212/first-launch-of-activity-with-google-maps-is-very-slow
https://stackoverflow.com/questions/26265526/what-makes-my-map-fragment-loading-slow
And many others questions on stackoverflow

giorgio79:
Ok, so this is a showstopper for Google Maps with Flutter. What is Google doing about it?

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
```diff
@@ -228,8 +228,12 @@ public void onMethodCall(MethodCall call, final MethodChannel.Result result) {
             new OnMapReadyCallback() {
               @Override
               public void onMapReady(GoogleMap googleMap) {
-                fragmentManager.beginTransaction().remove(mapFragment).commit();
-                result.success(null);
+                try {
+                  fragmentManager.beginTransaction().remove(mapFragment).commit();
```
sanekyy: Maybe check state of fragment managaer by isStateSaved method?

shlima:
Waiting for this fix a lot

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -4,6 +4,19 @@
...
+    MethodChannel('flutter.io/googleMapsPluginUtils');
+
+/// Utils for GoogleMaps plugin
+class GoogleMapsPluginUtils {
+  ///This method creates dummy map. This call will initialize all services needed by GoogleMaps This
+  ///will speed up next GoogleMap view initialization. Android only.
+  static Future<void> warmUp() {
+    if (!Platform.isAndroid) throw Exception("warmUp is Android only");
+    return _channel.invokeMethod<void>('warmUp');
+  }
```
ditman, line 17: This should become a method in the `platform_interface` of the map, [like this one here](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_flutter_platform.dart#L235-L240) (and its [implementation](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart#L432-L438)).

ditman:
The refactor has [landed](https://github.com/flutter/plugins/commit/1a9d51d344f71b0f8af657053b04783e274fd1ca).

otopba:
@ditman 
Fixed

otopba:
@amirh @ditman 
friendly ping

cyanglaz:

mtc-jed:
Waiting on this to go through, such a massive initialization time is a real problem for a lot of people.

PR #2488, opened 2020-01-29
[google_maps_flutter] supports v2 android embedding.

cyanglaz:
https://github.com/flutter/flutter/issues/41837

Modified some e2e tests as they were actually failing prior to this PR:
1. The android getMaxZoomLevel API in google map since unstable, updated the `updateMinMaxZoomLevels` test to use a different approach: set min-max preference and try to zoom to see if the camera can zoom beyond the boundary.

2. Disabled all the location based tests. On Android, location apis are not accessible until granting permission. I am not sure if we have a way to grand permission on CI. /cc @collinjackson @amirh 

3. Updated `testSetMapStyle invalid JSSON String` to only check if it is failing rather than a specific error message. The error message returned from iOS and Android are different.


cyanglaz:
> Maybe we should only skip the failing tests on Android? LGTM otherwise.

Yes! Will update.

PR #2489, opened 2020-01-30
[local_auth] Allow device authentication (pin/pattern/passcode)

kennethj:
## Description

Added `LocalAuthentication().authenticate()` method that allows users to authenticate using biometric (fingerprint, face) or device authentication (pin, pattern, passcode).

## Screenshots
ios-auth
Screen Shot 2020-01-31 at 8 29 51 AM
## Related Issues

This resolves the feature request on https://github.com/flutter/flutter/issues/49703

## Checklist

- [x] I signed the [CLA].

## Breaking Change


jitenders859:
on Android Can see the use password button but when clicking on it it closes the dialog and shows this message in console

FingerprintDialogFrag( 4198): Failed to check device credential. Parent handler not found.

kennethj:
@jitenders859 Is that on a physical device or with on an emulator?
I was having issues running this on android emulators, but it appears to work on physical devices. 
After a little research, I found there may be `KeyguardManager#isDeviceSecure()` check that could be added to avoid the error.

jitenders859:
@kennethj I am using real android device and 
also can you show how to implement KeyguardManager#isDeviceSecure()

also in file

https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-recyclerview-release/biometric/src/main/java/androidx/biometric/Utils.java
From the source code we can see at line 82 that if the activity type is not of DeviceCredentialHandlerActivity it will fail

jitenders859:
@kennethj 

ezgif-3-b71f99e54d14

In the above gif when pressed on authenticated the popup opens two times and when pressed on use password it closes on reopening can't see fingerprint icon and on authenticating with fingerprint didn't authenticate. after that opening second time it authenticates using fingerprint.

also is there any problem with sdk version below 28

kennethj:
@jitenders859 
This does appear to be an issue with API 28 and below. This feature only seems to work as intended on API 29 (Android 10). I am going to try to find time today to look into it more.

apgapg:
@kennethj In Android 6.0 with non-fingerprint device it doesnt allow to change verification method.
Device: **Samsung J2 Pro, SM-J210F android 6.0.1  without fingerprint**

Here's my implementation:
```
  local_auth:
    git:
      url: https://github.com/kennethj/flutter-plugins
      ref: feature/local_auth
      path: packages/local_auth

```
my expectation is to have pattern verification also which works fine in higher android version. Only in 6.0 it fails.

IMG-20200603-WA0005

Screenshot_2020-06-03-15-45-54-822_com workozy android demo

apgapg:
Any progress in this PR?
We really need pattern lock verification because some devices dont have fingerprint.

SergejSachs:
yea thats true, device password auth would be so good as a fallback

kdhanji87:
Any progress in this PR?
We really need pattern lock verification because some devices dont have fingerprint would be a useful fallback

duncwillz:
@kennethj LGTM. @softmarshmallow @mehmetf Could you please help get this PR merged?. It wouldn't be worth reinventing the wheel locally. Thank you.

Zelosys:
Hello
Any ETA on this PR?

It's a necessary feature for our App, and we don't want to reinvent the Wheel and do it manual, when the plugin can do this soon by its own :)

softmarshmallow (approved):

SergejSachs:
Any update on this?

Zelosys:
It looks like after three months it's time to annoy the Google staff again.😀
Any updates on this? @mehmetf 
Is there anything we can do to speed this up and get it in?

mehmetf:
Is this PR actively maintained? It needs to be rebased. Please ping me and I will do a pass after someone claims ownership.

@Zelosys @kennethj

kennethj:
@mehmetf I will probably have some time in the next two weeks or so.

kennethj:
@mehmetf I think this is ready for review.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
```diff
@@ -75,24 +73,38 @@
...
-            .build();
+
+    if (allowCredentials) {
+      this.promptInfo =
+          new BiometricPrompt.PromptInfo.Builder()
+              .setDescription((String) call.argument("localizedReason"))
+              .setTitle((String) call.argument("signInTitle"))
+              .setSubtitle((String) call.argument("biometricHint"))
+              .setConfirmationRequired((Boolean) call.argument("sensitiveTransaction"))
+              .setDeviceCredentialAllowed(true)
```
mehmetf, line 92: You can pass  `allowCredentials` to  `setDeviceCredentialAllowed` call and get rid  of the conditional.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
```diff
@@ -125,21 +137,27 @@ private void stop() {
   public void onAuthenticationError(int errorCode, CharSequence errString) {
     switch (errorCode) {
       case BiometricPrompt.ERROR_NO_DEVICE_CREDENTIAL:
-        completionHandler.onError(
-            "PasscodeNotSet",
-            "Phone not secured by PIN, pattern or password, or SIM is currently locked.");
-        break;
+        if (call.argument("useErrorDialogs")) {
+          showGoToSettingsDialog(
+              "Device credentials required",
```
mehmetf, line 142: These  strings should come from Dart just like the rest of the status strings used below.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
```diff
@@ -213,19 +231,23 @@ public void onResume(@NonNull LifecycleOwner owner) {
     onActivityResumed(null);
   }
 
+  // hack: dialog opens twice on Android 8
```
mehmetf, line 234: Do we know why? Do the handlers below fire twice for some reason?

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
```diff
@@ -65,138 +85,275 @@ private LocalAuthPlugin(Activity activity) {
...
   @Override
   public void onAttachedToEngine(FlutterPluginBinding binding) {
-    channel = new MethodChannel(binding.getBinaryMessenger(), CHANNEL_NAME);
+    channel = new MethodChannel(binding.getFlutterEngine().getDartExecutor(), CHANNEL_NAME);
   }
 
   @Override
   public void onDetachedFromEngine(FlutterPluginBinding binding) {}
 
+  final PluginRegistry.ActivityResultListener resultListener =
```
mehmetf, line 318: Can you please move this definition up, before it is used for the first time?

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
```diff
@@ -65,138 +85,275 @@ private LocalAuthPlugin(Activity activity) {
    */
   public LocalAuthPlugin() {}
 
+  private LocalAuthPlugin(Registrar registrar) {
+    this.registrar = registrar;
+    this.registrar.addActivityResultListener(resultListener);
```
mehmetf, line 90: There's no need for this constructor.  You can do all this in the registerWith() call.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
```diff
@@ -29,14 +39,28 @@
...
+  private Registrar registrar;
+
+  private Activity getActivity() {
+    if (activityPluginBinding != null) {
+      return activityPluginBinding.getActivity();
+    }
+    if (registrar != null) {
+      return registrar.activity();
+    }
+    return null;
```
mehmetf, line 54: Should  you ever be here? I don't see you checking  for null in everywhere getActivity is called  below.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
```diff
@@ -65,138 +85,275 @@ private LocalAuthPlugin(Activity activity) {
...
+        biometrics.add("iris");
+      }
+    }
+
+    return biometrics;
+  }
+
+  private boolean isDeviceSupported() {
+    KeyguardManager keyguardManager =
+        (KeyguardManager) getActivity().getBaseContext().getSystemService(KEYGUARD_SERVICE);
```
mehmetf, line 292: If these services can be grabbed via ApplicationContext, prefer doing that and save  them as instance members to the plugin class. That way you don't have to get them with every call and eliminate the possibility that getActivity could be null.

packages/local_auth/example/lib/main.dart
```diff
@@ -21,11 +21,21 @@ class MyApp extends StatefulWidget {
 
 class _MyAppState extends State<MyApp> {
   final LocalAuthentication auth = LocalAuthentication();
-  late bool _canCheckBiometrics;
-  late List<BiometricType> _availableBiometrics;
+
+  bool? _isSupported;
```
mehmetf, line 25: Prefer  initializing variables rather than leaving them nullable.

packages/local_auth/lib/auth_strings.dart
```diff
@@ -15,35 +15,35 @@ import 'package:intl/intl.dart';
 /// Provides default values for all messages.
 class AndroidAuthMessages {
   const AndroidAuthMessages({
-    this.fingerprintHint,
-    this.fingerprintNotRecognized,
-    this.fingerprintSuccess,
+    this.biometricHint,
```
mehmetf, line 18: This is a breaking change. Could you please add it to the CHANGELOG?

packages/local_auth/lib/local_auth.dart
```diff
@@ -66,7 +66,7 @@ class LocalAuthentication {
   /// authentication (e.g. lack of relevant hardware). This might throw
   /// [PlatformException] with error code [otherOperatingSystem] on the iOS
   /// simulator.
-  Future<bool> authenticateWithBiometrics({
+  Future<bool?> authenticateWithBiometrics({
```
mehmetf, line 69: This should not return a nullable type.

packages/local_auth/lib/local_auth.dart
```diff
@@ -92,9 +92,77 @@ class LocalAuthentication {
...
+  /// Setting [sensitiveTransaction] to true enables platform specific
+  /// precautions. For instance, on face unlock, Android opens a confirmation
+  /// dialog after the face is recognized to make sure the user meant to unlock
+  /// their phone.
+  ///
+  /// Throws an [PlatformException] if there were technical problems with local
+  /// authentication (e.g. lack of relevant hardware). This might throw
+  /// [PlatformException] with error code [otherOperatingSystem] on the iOS
+  /// simulator.
+  Future<bool?> authenticate({
```
mehmetf, line 138: Since you are making a breaking change already due to message strings, why not rename the method `authenticateWithBiometrics` as `authenticate` instead of having a near duplicate method?

That would leave the code much simpler.

mehmetf (commented):
Thanks for the contribution (and especially tests). There's additional work required before we can merge this into the repo.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
kennethj: There was an issue with setting both `setNegativeButtonText` and `setDeviceCredentialAllowed`. 


From android docs:
> Note that enabling this option replaces the negative button on the prompt with one that allows the user to authenticate with their device credential, making it an error to call setNegativeButton(java.lang.CharSequence, java.util.concurrent.Executor, android.content.DialogInterface.OnClickListener).

I will refactor to try to make it more obvious what is happening.

packages/local_auth/example/lib/main.dart
kennethj: I'll add an enum to avoid the `null`

packages/local_auth/lib/local_auth.dart
kennethj: I wanted to keep the original behavior intact of using biometrics only with `authenticateWithBiometrics` and have `authenticate` use any means available, then perhaps a `authenticateWithCredentials` at some point.

If we remove `authenticateWithBiometrics`, I would want to add a way to keep the original biometric only behavior. Maybe a `bool biometricOnly = false,` parameter on `authenticate`?

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
kennethj: updated

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
kennethj: Added to `AndroidAuthMessages` class

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
kennethj: moved up

packages/local_auth/example/lib/main.dart
kennethj: Updated to use the `enum` `_SupportState` instead of a nullable `bool`

packages/local_auth/lib/local_auth.dart
kennethj: Updated to return `Future<bool>` instead of `Future<bool?>`

kennethj:
@mehmetf, thanks for reviewing that so quickly!

I will test these changes on a few emulators and finish updates based on your review tomorrow.

packages/local_auth/lib/local_auth.dart
mehmetf: Yes, a `biometricOnly` parameter sounds good.

I like your original idea  as well (having separate methods) however the implementation (and API) is largely shared between these methods and you would need to make sure a feature added to `authenticate` is also added to all of its 
 variants (`authenticateWithBiometrics` etc).

You can reduce the implementation of `authenticateWithBiometrics` to just calling `authenticate` with `biometricOnly` set to true. However, I would rather keep the API clean and simple.

packages/local_auth/lib/auth_strings.dart
kennethj: Noted to changelog.

packages/local_auth/lib/local_auth.dart
kennethj: Added a `biometricOnly` parameter to `authenticate`. `authenticateWithBiometrics` now calls `authenticate` with `biometricOnly: true`

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
kennethj: I am not sure. This happened a while ago and don't remember the specifics and I have not been able to reproduce the issue in emulators today.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
kennethj: Moved code to `registerWith()` method and deleted extra constructor.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
mehmetf: Let's remove it if it cannot be reproduced.

mehmetf:
Thanks for making all these changes. Let me know when it is ready for  the second round!

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
kennethj: removed

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
kennethj: Removed `getActivity()`. Setting services and activity as instances variables when plugin registered and on attached to activity.

kennethj:
@mehmetf 
I believe I addressed all the comments in the first review and ready for a second look.

Also, I am not sure what's up with the failing CI build `build-apks+java-test+firebase-test-lab CHANNEL:stable PLUGIN_SHARDING:--shardIndex 0 --shardCount 4`. I started updating gradle versions, but that started to become a rabbit hole.

zigapovhe:
> @mehmetf
> I believe I addressed all the comments in the first review and ready for a second look.
> 
> Also, I am not sure what's up with the failing CI build `build-apks+java-test+firebase-test-lab CHANNEL:stable PLUGIN_SHARDING:--shardIndex 0 --shardCount 4`. I started updating gradle versions, but that started to become a rabbit hole.

@kennethj Did you enable  'android.useAndroidX' property to true? Set this property to true in the gradle.properties file and retry.

packages/local_auth/CHANGELOG.md
```diff
@@ -28,8 +37,10 @@
 
 ## 0.6.3+1
 
+* Add `authenticate` method that uses biometric authentication, but allows users to also use device authentication - pin, pattern, passcode.
```
mehmetf, line 40: We are not going to publish this as all changes are now done in the null-safety version. Please move to the CHANGELOG description above.

packages/local_auth/CHANGELOG.md
```diff
@@ -61,10 +72,12 @@
 
 * Replace deprecated `getFlutterEngine` call on Android.
 
+
```
mehmetf, line 75: Remove these extra newlines. It might be due to formatting done in your editor?

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
```diff
@@ -67,136 +95,255 @@ public LocalAuthPlugin() {}
...
-    if (call.method.equals("authenticateWithBiometrics")) {
-      if (authInProgress.get()) {
-        // Apps should not invoke another authentication request while one is in progress,
-        // so we classify this as an error condition. If we ever find a legitimate use case for
-        // this, we can try to cancel the ongoing auth and start a new one but for now, not worth
-        // the complexity.
-        result.error("auth_in_progress", "Authentication in progress", null);
-        return;
-      }
+    this.result = result;
```
mehmetf, line 98: This is dangerous. I assume you are saving result because some operation below can respond asynchronously via `onPermissionResult`, `onActivityResult` or some such. You should save the result object on a per-method basis right before you invoke the `startActivityForResult`. Otherwise, another method call can override this result object and you end up calling `result` methods more than once.

This also gives you the chance to reject a second call to the async method. e.g. if you are still awaiting the result of some `foo` method and user calls `foo` again, you can ignore it and instead wait for the `Result  fooResult` to be fulfilled.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
```diff
@@ -67,136 +95,255 @@ public LocalAuthPlugin() {}
...
+    switch (call.method) {
+      case "authenticate":
+        this.authenticate(call, result);
```
mehmetf, line 101: `this.` qualifier is unnecessary (ditto below).

packages/local_auth/lib/local_auth.dart
```diff
@@ -73,28 +73,87 @@ class LocalAuthentication {
     AndroidAuthMessages androidAuthStrings = const AndroidAuthMessages(),
     IOSAuthMessages iOSAuthStrings = const IOSAuthMessages(),
     bool sensitiveTransaction = true,
+  }) =>
```
mehmetf, line 76: I would strongly recommend deprecating this method and then removing it sometime in the future.

Please put a `@deprecated` qualifier to it, remove its dartdoc replacing it with "This method is deprecated, please use `authenticate` with `biometricOnly: true`."

packages/integration_test/android/build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.3'
+        classpath 'com.android.tools.build:gradle:4.1.1'
```
mehmetf, line 11: I am not entirely sure how version resolving for gradle works for this testbed. I will get someone from Flutter core team to review these particular changes. That's likely what's causing your build failure.

mehmetf: @cyanglaz Could you please take a look at the build file changes in this PR? What is the proper way to up the gradle version?

packages/local_auth/CHANGELOG.md
kennethj: Moved.

packages/local_auth/CHANGELOG.md
kennethj: extra newlines removed

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
kennethj: Updated

packages/local_auth/lib/local_auth.dart
kennethj: Added `@Deprecated` annotation and comment docs

packages/integration_test/android/build.gradle
kennethj: I revert the changes to the `integration_test` package in this PR. This appears to prevent those failed builds.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
mehmetf: I think this is now the only blocker before I LGTM this PR.

kennethj: Removed class level `result` field. Added`lockRequestResult` field that is only set when lock request activity started. This field `lockRequestResult` is only used in activity result delegate function.

mehmetf (approved):

mehmetf:
Published as 1.1.0-nullsafety.

packages/local_auth/ios/Classes/FLTLocalAuthPlugin.m
```diff
@@ -130,6 +136,48 @@ - (void)authenticateWithBiometrics:(NSDictionary *)arguments
   }
 }
 
+- (void)authenticate:(NSDictionary *)arguments withFlutterResult:(FlutterResult)result {
+  LAContext *context = [[LAContext alloc] init];
+  NSError *authError = nil;
+  _lastCallArgs = nil;
+  _lastResult = nil;
+  context.localizedFallbackTitle = @"";
```
stuartmorgan, line 144: Why is fallback being disabled here? That seems undesirable for the non-biometric-only case.

mtskf: Please escalate this!

PR #2503, opened 2020-02-05
[google_maps_flutter] Add pointForMeters to plugin in iOS

klaes-ashford:
## Description

This PR will implement the pointForMeters projection api method in the plugin for iOS. This method helps you get the distance in meters to content size from the specified LatLng in the map. 

iOS maps sdk doesn't support drawing dotted polylines out of the box. Instead we make use of drawing circles between two LatLngs. We have to make use of pointForMeters api method to draw circles without specifying a constant radius so that the circle scales regardless of zoom.

## Related Issues

https://github.com/flutter/flutter/issues/37959
https://github.com/flutter/plugins/pull/2108

## Checklist

- [x] I signed the [CLA].

## Breaking Change


aswinkviswanath:
The android part of pointForMeters is not implemented in this PR. Although I am willing to look into that and figure out that. Any help on this will be appreciated.

aswinkviswanath:
@cyanglaz drawing dotted polylines in iOS makes use of this api since SDK doesn't support it out of the box. I can't find android equivalent for the same as the map sdk for android doesn't seem to expose this method. What should be done in this case ?

aswinkviswanath:
@cyanglaz is there something that I am missing, that should be addressed. No clue as to why there is no response on this one ?

packages/google_maps_flutter/example/test_driver/google_maps_e2e.dart
```diff
@@ -718,6 +718,35 @@ void main() {
...
+    // We suspected a bug in the iOS Google Maps SDK caused the camera is not properly positioned at
+    // initialization. https://github.com/flutter/flutter/issues/24806
+    // This temporary workaround fix is provided while the actual fix in the Google Maps SDK is
+    // still being investigated.
+    // TODO(cyanglaz): Remove this temporary fix once the Maps SDK issue is resolved.
+    // https://github.com/flutter/flutter/issues/27550
+    await tester.pumpAndSettle(const Duration(seconds: 3));
+
+    final double distance = await controller.getDistance(1, _kInitialMapCenter);
+    expect(distance, 0.00020464534463826567);
```
cyanglaz: Is there a way that we can test this without using some magic number?

packages/google_maps_flutter/lib/src/controller.dart
```diff
@@ -236,4 +236,15 @@ class GoogleMapController {
         'map#getLatLng', screenCoordinate._toJson());
     return LatLng(latLng[0], latLng[1]);
   }
+
+  /// Returns [double] corresponding to the distance in meters of the [LatLng] in the current map view.
+  ///
+  /// Return NaN (Not-a-Number) if unable to get distance in meters.
+  /// This is only accurate for small Earth distances.
+  Future<double> getDistance(double meters, LatLng latLng) async {
```
cyanglaz: The iOS Google Map SDK says "Converts a distance in meters to content size."
Maybe we should say something similar in our dart doc to avoid confusion?

What happens if meters is null, or if latlng is null?

packages/google_maps_flutter/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: google_maps_flutter
 description: A Flutter plugin for integrating Google Maps in iOS and Android applications.
 homepage: https://github.com/flutter/plugins/tree/master/packages/google_maps_flutter
-version: 0.5.22+1
+version: 0.5.22+2
```
cyanglaz: Since we are adding a new API, we should bump the minor version: 0.5.23 

aswinkviswanath:
@cyanglaz thanks for the review, I will look into the issues and try to come up with the solutions as soon as I get some free time. thanks.

packages/google_maps_flutter/lib/src/controller.dart
```diff
@@ -236,4 +236,15 @@ class GoogleMapController {
...
+    final double distance = await channel.invokeMethod<double>(
+        'map#getDistance',
+        <String, dynamic>{"meter": meters, "location": latLng._toJson()});
+    return distance;
```
ditman, line 248: This should become a method in the `platform_interface` of the map, [here](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_flutter_platform.dart#L235-L240) (and its [implementation](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart#L432-L438)).

ditman:
The refactor has [landed](https://github.com/flutter/plugins/commit/1a9d51d344f71b0f8af657053b04783e274fd1ca).

packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
```diff
@@ -1,13 +1,13 @@
...
 homepage: https://github.com/flutter/plugins/tree/master/packages/google_maps_flutter/google_maps_flutter
-version: 0.5.27
+version: 0.5.28
 
 dependencies:
   flutter:
     sdk: flutter
   flutter_plugin_android_lifecycle: ^1.0.0
-  google_maps_flutter_platform_interface: ^1.0.1
+  google_maps_flutter_platform_interface: ^1.0.2
```
aswinkviswanath, line 10: @ditman, I've added an api to the google_maps_flutter_platform_interface and bumped the version number according to sem versioning. But since this is not hosted anywhere I am getting this issue. `google_maps_flutter depends on google_maps_flutter_platform_interface ^1.0.2 which doesn't match any versions, version solving failed.` To solve this issue, I have to use relative path here, rather than hosted one. And have done the same to run test at my end. Any suggestions regarding this would be helpful.

packages/google_maps_flutter/lib/src/controller.dart
aswinkviswanath: The wording for dart docs have been rephrased as you suggested above, also we have added assertions for meters and latlng to be non null values [here](packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart)

packages/google_maps_flutter/pubspec.yaml
aswinkviswanath: This has been fixed considering the changes for federated implementation style.

aswinkviswanath:
@googlebot I fixed it.

packages/google_maps_flutter/example/test_driver/google_maps_e2e.dart
aswinkviswanath: I assume the math specified in [Google Groups forums](https://groups.google.com/forum/#!msg/google-maps-js-api-v3/hDRO4oHVSeM/osOYQYXg2oUJ)
`meters/px = 156543.03392 * Math.cos(latitude * Math.PI / 180) / Math.pow(2, zoom)
` can be used to test this and **add android implementation**, But using this method, there was a precision mismatch between the value generated from projection api. This magic number specified here is the value we get from projection api for 1 meter at latitude 0 and longitude -0. I will check if I can get the math correct. help here is appreciated.

cyanglaz: maybe we can try https://api.flutter.dev/flutter/flutter_test/moreOrLessEquals.html

aswinkviswanath:
@cyanglaz  Also, since google_maps_flutter plugin is now using the federated plugin implementation style and will support upto 6 different platforms in the future, can we not implement the same for android now, as the underlying google maps sdk for Android itself doesn't support the capability right now ? 

Also, `getDistance` for method name doesn't convey proper meaning now that I think about it. should it be `getPoint / getContentSize`. Do you have any advice here ?

packages/google_maps_flutter/example/test_driver/google_maps_e2e.dart
aswinkviswanath:
```
final double distance = await inspector.getDistance(1, _kInitialCameraPosition.target);
// Formula is from Chris Broadfoot's comment : https://groups.google.com/d/msg/google-maps-js-api-v3/hDRO4oHVSeM/osOYQYXg2oUJ
final double meters_per_pixel = 156543.03392 * cos(_kInitialCameraPosition.target.latitude * pi / 180) / pow(2, _kInitialZoomLevel);
final double points_per_meter = 1 / meters_per_pixel;
expect(distance, moreOrLessEquals(points_per_meter));
```
When I try the above code, I am getting a test failure.
```
flutter: The following TestFailure object was thrown running a test:
flutter:   Expected: 0.00020441663355236448 (±1e-10)
flutter:   Actual: <0.00020464534463826567>
flutter:    Which: 0.00020464534463826567 is not in the range of 0.00020441663355236448 (±1e-10).
```
for different zoom levels, the actual value and expected value starts to diverge from 6th scale.

aswinkviswanath:
@ditman @cyanglaz ,  I am integrating Federated plugin implementations changes to my open PR , and is facing some issues. Help resolving this is much appreciated.

I've added an api to the google_maps_flutter_platform_interface and bumped the version number according to sem versioning. But since this is not hosted anywhere I am getting this issue. 
`google_maps_flutter depends on google_maps_flutter_platform_interface ^1.0.2 which doesn't match any versions, version solving failed.`

 To solve this issue, I have to use relative path in my local project, rather than hosted one. But doing so will make my changes unpublishable in github tests.
`Don't depend on "google_maps_flutter_platform_interface" from the path source. Use the hosted source instead.`

and if I stick with the google_maps_flutter_platform_interface ^1.0.1, the new changes are not available for the CI and tests would fail.
`error - The method 'getPoint' isn't defined for the type 'GoogleMapsFlutterPlatform' at lib/src/controller.dart:259:39 - (undefined_method)`

Now, the tests are failing due to this. How should I solve this problem ?

ditman:
@aswinkviswanath you're correct. In order for you to push this change, you'll need to first get the changes in the platform_interface package reviewed and published, then the changes in the core plugin. **This requires two separate PRs.**

It is normal to have a version of the core plugin that depends on the platform_interface via path, that is unpublishable until the platform_interface lands.

Let me know if you have more questions, and thanks for adopting the new style so quickly!

ditman:
I've introduced a change to this PR that should be reverted once the platform interface changes land, and are published to pub.dev. This is to make sure that all tests pass (except the "publishable" one, of course)

ditman:
(It seems there's ~38 tests broken in Android, can you take a look at those @aswinkviswanath?)

```
+---------+------------------------+--------------------------------+
| OUTCOME |    TEST_AXIS_VALUE     |          TEST_DETAILS          |
+---------+------------------------+--------------------------------+
| Failed  | flame-29-en-portrait   | 38 test cases failed, 8 passed |
| Failed  | walleye-26-en-portrait | 38 test cases failed, 8 passed |
+---------+------------------------+--------------------------------+
```

Some traces (don't look very useful):

```
05-04 12:36:16.807: I/TestRunner(10192): started: testTraffic(io.flutter.plugins.googlemaps.EmbeddingV1ActivityTest)
05-04 12:36:16.808: I/MonitoringInstr(10192): Activities that are still in CREATED to STOPPED: 0
05-04 12:36:16.809: E/TestRunner(10192): failed: testTraffic(io.flutter.plugins.googlemaps.EmbeddingV1ActivityTest)
05-04 12:36:16.809: E/TestRunner(10192): ----- begin exception -----
05-04 12:36:16.809: E/TestRunner(10192): java.lang.Exception: failed
05-04 12:36:16.809: E/TestRunner(10192): 	at dev.flutter.plugins.e2e.FlutterRunner.run(FlutterRunner.java:62)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runners.Suite.runChild(Suite.java:128)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runners.Suite.runChild(Suite.java:27)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
05-04 12:36:16.809: E/TestRunner(10192): 	at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
05-04 12:36:16.809: E/TestRunner(10192): 	at androidx.test.internal.runner.TestExecutor.execute(TestExecutor.java:56)
05-04 12:36:16.809: E/TestRunner(10192): 	at androidx.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:392)
05-04 12:36:16.809: E/TestRunner(10192): 	at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2074)
05-04 12:36:16.809: E/TestRunner(10192): ----- end exception -----
05-04 12:36:16.810: I/TestRunner(10192): finished: testTraffic(io.flutter.plugins.googlemaps.EmbeddingV1ActivityTest)
```

aswinkviswanath:
@ditman `flutter driver --target test_driver/google_maps_e2e.dart` for android would fail due to `MissingPluginException(No implementation found for method projection#getPointsForMeters on channel` as getPointsForMeters is not implemented for android as the android google maps sdk itself doesn't expose the same. I was unable to see the getPointsForMeters in android projection api's [documentation](https://developers.google.com/android/reference/com/google/android/gms/maps/Projection). Correct me if I am wrong, I assume this is the failure with integration tests. what should be done in this case ? 

We could use the logic used in the driver test to verify getPointsForMeters to make an implementation for andriod, but I am not sure how accurate that is / if its the way to go.

and master branch contains all the changes in `aswinkviswanath:feature/GetPoint`.

ditman:
@aswinkviswanath this is the list of failing tests in Android:

```
testRotateGesturesEnabled
testZoomGesturesEnabled
testSetMapStyle valid Json String
testSetMapStyle invalid Json String
testGetVisibleRegion
testScreenCoordinate
testInitialCenterLocationAtCenter
testTiltGesturesEnabled
testGetPointsForMeters
testTraffic
testGetLatLng
testSetMapStyle null string
testZoomControlsEnabled
testMapToolbarToggle
updateMinMaxZoomLevels
testCompassToggle
testScrollGesturesEnabled
testGetZoomLevel
testBuildings
testRotateGesturesEnabled
testZoomGesturesEnabled
testSetMapStyle valid Json String
testSetMapStyle invalid Json String
testGetVisibleRegion
testScreenCoordinate
testInitialCenterLocationAtCenter
testTiltGesturesEnabled
testGetPointsForMeters
testTraffic
testGetLatLng
testSetMapStyle null string
testZoomControlsEnabled
testMapToolbarToggle
updateMinMaxZoomLevels
testCompassToggle
testScrollGesturesEnabled
testGetZoomLevel
testBuildings
```

I'd understand `testGetPointsForMeters` fails, but I don't understand why the others do.

BTW, can you make `testGetPointsForMeters` test to be skipped when not run on iOS?

ditman:
See an [example of a skipped test](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter/example/test_driver/google_maps_e2e.dart#L635).

aswinkviswanath:
@ditman skipped testGetPointsForMeters in non iOS platforms, and the integration tests seems to pass now.

ditman:
> @ditman skipped testGetPointsForMeters in non iOS platforms, and the integration tests seems to pass now.

Yes, this is great, thanks for taking care of this @aswinkviswanath!!

cyanglaz:
Can we make this API available for android as well?

PR #2506, opened 2020-02-06
[image_picker] update flutter_plugin_android_lifecycle version

PiN73:
## Description

`flutter_plugin_android_lifecycle: 1.0.3` was causing a crash in Android. It's fixed in `1.0.4`. But the crash is still reproduced if `version: "1.0.3"` is saved in pubspec.lock.

This PR updates the dependency to `^1.0.4`

## Related Issues

https://github.com/flutter/flutter/issues/48622 (original issue)
https://github.com/flutter/flutter/issues/48906 (more detailed reproduction steps)

## Breaking Change

### DIFF
--- a/packages/image_picker/CHANGELOG.md
+++ b/packages/image_picker/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.6.3+2
 
 * Bump RoboElectric dependency to 4.3.1 and update resource usage.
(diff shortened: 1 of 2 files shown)


Hixie:
Actually revving dependencies doesn't need a test. However, the crash should have one.

PiN73:
Closing as lower version constraint is updated by other commits

PR #2511, opened 2020-02-10
nit: Add a short comment explain path dependencies

jonasfj:
Adds a comment `# ^x.y.z in your application` where examples have path dependencies.

This was motivated by https://github.com/dart-lang/pub/issues/1850#issuecomment-581117550 where a user is running into problems by copying the dependency from an `example/pubspec.yaml`.

Maybe we shouldn't make it all one-line... and maybe we could phrase the comment better.
Or maybe we shouldn't even do this -- feel free to close this if you think this is too obvious :)


packages/android_alarm_manager/example/pubspec.yaml
```diff
@@ -4,8 +4,7 @@ description: Demonstrates how to use the android_alarm_manager plugin.
 dependencies:
   flutter:
     sdk: flutter
-  android_alarm_manager:
-    path: ../
+  android_alarm_manager: {path: ../} # ^x.y.z in your application
```
amirh: I really don't care if this is one line or 2 lines, though I don't want these to go back-and-forth between the 2 styles.
This is the kind of things we should write in a style guide, the closest thing to a style guide I found are the examples here: https://dart.dev/tools/pub/dependencies (which are using 2 lines). If we believe this style is preferred we should probably update that page as well.

amirh: nit: I'd make the comment more verbose so it's easier to understand why it's a path dependency here, maybe something like:

```
The example app is bundled with the plugin so we use a path dependency on the parent directory to use the current plugin's version. When depending on this package from a real application you should use : plugin_name: ^x.y.z
See https://dart.dev/tools/pub/dependencies#version-constraints for more information.
```

jonasfj: I only used the one-liner style because it put the comment on the same line.

jonasfj: Good point, that's better than using one-line :)

jonasfj:
@amirh, I just updated the PR to include `publish_to: 'none'` for all example plugins.

They aren't meant to be published I assume, and the template already has these:
https://github.com/flutter/flutter/blob/5f4658973be1ad10a59a0a278b6cdf41b10fe7f6/packages/flutter_tools/templates/app/pubspec.yaml.tmpl#L4

jonasfj:
@amirh, I've updated this and https://github.com/flutter/flutter/pull/50978, if you're still happy with this let's ship it :)

jonasfj:
Please merge if you're happy with this.

ened:
hi @jonasfj could you please rebase & take a look at the CI errors?

jonasfj:
Not sure why CI would fail.

stuartmorgan:
This still seems like a useful thing to land. @jonasfj do you have cycles to resolve the conflicts and then we can submit it?

jonasfj:
Okay, I'm going to close this... Not sure what mess I just made of the history :see_no_evil:

stuartmorgan:
@jonasfj If you want to do a clean version I can review it ASAP so it doesn't bitrot.

PR #2465, opened 2020-01-13
[e2e] Fixed bug in e2e for ios where it doesn't exit.

gaaclarke:
## Description

Fixed problem with XCTest in XCode 11.3 where the testing bundles were getting opened multiple times which interfered with the singleton logic for E2EPlugin.

## Related Issues

- https://github.com/flutter/flutter/issues/48601

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #2514, opened 2020-02-11
[e2e] Test multiple failure behavior

collinjackson:
## Description

Adds a skipped integration test for flutter/flutter#48601.

Will be fixed by https://github.com/flutter/plugins/pull/2465

Note: Right now the plugins CI only runs e2e driver tests on iOS. However, this should be sufficient to exercise the fix for flutter/flutter#48601.

## Related Issues

https://github.com/flutter/flutter/issues/48601

### DIFF
--- a/packages/e2e/example/test_driver/failure_e2e_test.dart
+++ b/packages/e2e/example/test_driver/failure_e2e_test.dart
@@ -1,0 +1,19 @@
+import 'dart:async';
+import 'dart:io';
+
+import 'package:flutter_driver/flutter_driver.dart';
+import 'package:test/test.dart';
+
+Future<void> main() async {
+  test('fails gracefully', () async {
+    final FlutterDriver driver = await FlutterDriver.connect();
+    final String result =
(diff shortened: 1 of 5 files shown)


packages/e2e/example/test_driver/failure_e2e_test.dart
```diff
@@ -0,0 +1,18 @@
...
+Future<void> main() async {
+  test('fails gracefully', () async {
+    final FlutterDriver driver = await FlutterDriver.connect();
+    final String result =
+        await driver.requestData(null, timeout: const Duration(minutes: 1));
+    await driver.close();
+    expect(
+      result,
+      'failure',
+      skip: true, // https://github.com/flutter/flutter/issues/48601
```
nturgut: question: why is skip used?

gaaclarke: i'm about to issue a fix for it is the idea.

PR #2522, opened 2020-02-12
[shared_preferences] Fix README broken link

franciscojma86:


packages/shared_preferences/shared_preferences_macos/README.md
```diff
@@ -31,4 +31,4 @@ dependencies:
   ...
 ```
 
-[1]: ../shared_preferences/shared_preferences
+[1]: ../
```
mklim: This links to a directory of folders on Github, if somebody didn't already know what it was it would be confusing. I think it would make more sense to link to the interface package on pub, WDYT?

packages/shared_preferences/shared_preferences_macos/LICENSE
```diff
@@ -23,4 +23,4 @@ LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
 DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
 THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
\ No newline at end of file
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
mklim: This change should probably be reverted.

packages/shared_preferences/shared_preferences_macos/README.md
franciscojma86: I don't have a strong preference, but I think it's better to be consistent and other plugins do the same. Maybe at some point we should change them all?

mklim: Sure, I think that makes more sense.

PR #2527, opened 2020-02-14
[connectivity] Use platform_interface in core plugin

franciscojma86:
## Description

Change the core plugin to make use of the platform_interface

## Related Issues
https://github.com/flutter/flutter/issues/46307

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.24
+
+* Exposed `getZoomLevel` in `GoogleMapController`.
+
 ## 0.5.23+1
 
 * Move core plugin to its own subdirectory, to prepare for federation.
--- a/packages/google_maps_flutter/google_maps_flutter/example/test_driver/google_maps_e2e.dart
+++ b/packages/google_maps_flutter/google_maps_flutter/example/test_driver/google_maps_e2e.dart
@@ -683,6 +683,41 @@ void main() {
     expect(topLeft, northWest);
   });
 
+  testWidgets('testGetZoomLevel', (WidgetTester tester) async {
+    final Key key = GlobalKey();
+    final Completer<GoogleMapController> controllerCompleter =
+        Completer<GoogleMapController>();
+
+    await tester.pumpWidget(Directionality(
+      textDirection: TextDirection.ltr,
+      child: GoogleMap(
+        key: key,
+        initialCameraPosition: _kInitialCameraPosition,
+        onMapCreated: (GoogleMapController controller) {
+          controllerCompleter.complete(controller);
+        },
+      ),
+    ));
+
+    final GoogleMapController controller = await controllerCompleter.future;
+
+    // We suspected a bug in the iOS Google Maps SDK caused the camera is not properly positioned at
+    // initialization. https://github.com/flutter/flutter/issues/24806
+    // This temporary workaround fix is provided while the actual fix in the Google Maps SDK is
+    // still being investigated.
+    // TODO(cyanglaz): Remove this temporary fix once the Maps SDK issue is resolved.
+    // https://github.com/flutter/flutter/issues/27550
+    await tester.pumpAndSettle(const Duration(seconds: 3));
+
+    double zoom = await controller.getZoomLevel();
+    expect(zoom, _kInitialZoomLevel);
+
+    await controller.moveCamera(CameraUpdate.zoomTo(7));
+    await tester.pumpAndSettle();
+    zoom = await controller.getZoomLevel();
+    expect(zoom, equals(7));
+  });
+
   testWidgets('testScreenCoordinate', (WidgetTester tester) async {
     final Key key = GlobalKey();
     final Completer<GoogleMapController> controllerCompleter =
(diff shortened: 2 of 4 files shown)


packages/connectivity/connectivity/pubspec.yaml
```diff
@@ -17,6 +17,8 @@ dependencies:
   flutter:
     sdk: flutter
   meta: "^1.0.5"
+  connectivity_platform_interface:
+    path: ./connectivity_platform_interface
```
ditman: Please, depend on the published pub ^ version:

`connectivity_platform_interface: ^1.0.2`

packages/connectivity/connectivity/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: connectivity
 description: Flutter plugin for discovering the state of the network (WiFi &
   mobile/cellular) connectivity on Android and iOS.
 homepage: https://github.com/flutter/plugins/tree/master/packages/connectivity
```
ditman: homepage is now: https://github.com/flutter/plugins/tree/master/packages/connectivity/connectivity

packages/connectivity/connectivity/example/pubspec.yaml
```diff
@@ -5,9 +5,9 @@ dependencies:
   flutter:
     sdk: flutter
   connectivity:
-    path: ../
+    path: ../../connectivity
```
ditman: You don't need this change here, the `connectivity` plugin is still in `../`

packages/connectivity/connectivity/example/pubspec.yaml
```diff
@@ -5,9 +5,9 @@ dependencies:
...
   connectivity_macos:
-    path: ../connectivity_macos
+    path: ../../connectivity_macos
```
ditman: I would depend on the published version of this one (once it is published), so you don't have to pierce outside of your own package to access it.

packages/connectivity/connectivity/pubspec.yaml
```diff
@@ -17,6 +17,8 @@ dependencies:
   flutter:
     sdk: flutter
   meta: "^1.0.5"
+  connectivity_platform_interface:
+    path: ^1.0.2
```
ditman: Does this work? I think you need to remove `path`. The version number goes right after the plugin name.

franciscojma86: hahah you are fast. I notice right away but you beat me to it

packages/connectivity/connectivity/example/pubspec.yaml
ditman: Ah, now that I mention this. You don't need this dependency *at all* once you make connectivity_macos the "endorsed" implementation of the plugin for the macos platform (which you should do for sure :P)

packages/connectivity/connectivity/pubspec.yaml
ditman: I was in the review page and it refreshed, thanks for fixing! :P

ditman:
`Because all_plugins depends on connectivity from path which depends on connectivity_platform_interface ^1.0.2, connectivity_platform_interface from hosted is required.`

Ah, IIRC you need to modify `build_all_plugins_app.sh` script to exclude the `connectivity_platform_interface` (and possibly `connectivity_macos` and `connectivity_web`) from the all_plugins app. See [`EXCLUDED_PLUGINS_LIST`](https://github.com/flutter/plugins/blob/master/script/build_all_plugins_app.sh#L13-L27).

packages/connectivity/connectivity/example/pubspec.yaml
franciscojma86: I actually can't do that yet, without making some changes to the macos package. I think those changes should be done in a separate PR, I think for now we should leave it in. WDYT?

ditman: SGTM!

packages/connectivity/connectivity/test/connectivity_test.dart
```diff
@@ -0,0 +1,92 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:connectivity/connectivity.dart';
+import 'package:connectivity_platform_interface/connectivity_platform_interface.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:plugin_platform_interface/plugin_platform_interface.dart';
+import 'package:mockito/mockito.dart';
```
ditman: I don't think you need mockito here at all, since you're not really mocking anything?

franciscojma86: Yeah I'm mocking the platform, at the bottom :)

ditman: Yes, you're mocking the platform at the bottom, but it's returning hardcoded stuff, so you can remove the Mock class and mockito from the mix, I think :P

ditman: (What I'm saying is you're never doing: `mockito.when(myMock.getBlah()).thenReturn("xyz")`, so no need for mockito for what you're doing now)

franciscojma86: But I do need it to do this: https://github.com/flutter/plugins/pull/2527/files#diff-e20d4ed7787f3ec62efa41db12110cadR64
If I don't include it, it fails

ditman: You can't `extend Mock`, but you should be able to do `class MockConnectivityPlatform extends ConnectivityPlatform {`, and just return mocked data from all the methods there. (You can't extend Mock, but you can create a fake class that returns hardcoded data through its overridden methods)

ditman: Discussed offline, and we'll leave this as is for now :)

PR #2528, opened 2020-02-14
[video_player] Made the video_player clean up after itself on iOS.

gaaclarke:
## Description

For add-to-app users of video_player, the plugin doesn't clean up its resources, like it's display link which causes a crash.  This handles the event that we are detaching the plugin and cleans everything up as a result.

## Related Issues

https://github.com/flutter/flutter/issues/50761

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
+++ b/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
@@ -390,6 +390,9 @@ static inline CGFloat radiansToDegrees(CGFloat radians) {
   return nil;
 }
 
+/// This method allows you to dispose without touching the event channel.  This
+/// is useful for the case where the Engine is in the process of deconstruction
+/// so the channel is going to die or is already dead.
 - (void)disposeSansEventChannel {
   _disposed = true;
   [_displayLink invalidate];
@@ -422,7 +425,6 @@ static inline CGFloat radiansToDegrees(CGFloat radians) {
 @property(readonly, weak, nonatomic) NSObject<FlutterBinaryMessenger>* messenger;
 @property(readonly, nonatomic) NSMutableDictionary* players;
 @property(readonly, weak, nonatomic) NSObject<FlutterPluginRegistrar>* registrar;
-
 @end
 
 @implementation FLTVideoPlayerPlugin
--- a/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
+++ b/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
@@ -6,6 +6,10 @@
 #import <AVFoundation/AVFoundation.h>
 #import <GLKit/GLKit.h>
 
+#if !__has_feature(objc_arc)
+#error Code Requires ARC.
+#endif
+
 int64_t FLTCMTimeToMillis(CMTime time) {
   if (time.timescale == 0) return 0;
   return time.value * 1000 / time.timescale;
@@ -13,7 +17,7 @@ int64_t FLTCMTimeToMillis(CMTime time) {
 
 @interface FLTFrameUpdater : NSObject
 @property(nonatomic) int64_t textureId;
-@property(nonatomic, readonly) NSObject<FlutterTextureRegistry>* registry;
+@property(nonatomic, weak, readonly) NSObject<FlutterTextureRegistry>* registry;
 - (void)onDisplayLink:(CADisplayLink*)link;
 @end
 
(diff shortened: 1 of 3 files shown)


gaaclarke:
Note: This PR won't function until https://github.com/flutter/engine/commit/76b291a65973685bb9d92a99810e27b3de9afc6c is on stable which seemed to have happened as part of this roll: https://github.com/flutter/flutter/commit/fc2c0af48d9d1fb7da2be1a51fcadf27ec0ae2fc which is tagged v1.15.3

packages/video_player/video_player/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: video_player
 description: Flutter plugin for displaying inline video with other Flutter
   widgets on Android and iOS.
-version: 0.10.7
+version: 0.10.8
 homepage: https://github.com/flutter/plugins/tree/master/packages/video_player/video_player
 
 flutter:
```
xster: Can we add restrictions on line 36? It crashes on compile if users accidentally flutter pub upgrade otherwise if they had ^0.10.7 right?

gaaclarke: Nope, my change to the plugins api is backwards compatible with respect to compiling, it just doesn't function unless you have v1.15.3.  Would we prefer to force the upgrade and increment the minor version to 0.11?

packages/video_player/video_player/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: video_player
 description: Flutter plugin for displaying inline video with other Flutter
   widgets on Android and iOS.
-version: 0.10.7
+version: 0.10.8
```
cyanglaz: I think we can keep this as 0.10.7+1 since there are no public api changes.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
```diff
@@ -404,17 +411,20 @@ - (void)dispose {
...
 
 @interface FLTVideoPlayerPlugin ()
-@property(readonly, nonatomic) NSObject<FlutterTextureRegistry>* registry;
-@property(readonly, nonatomic) NSObject<FlutterBinaryMessenger>* messenger;
+@property(readonly, weak, nonatomic) NSObject<FlutterTextureRegistry>* registry;
+@property(readonly, weak, nonatomic) NSObject<FlutterBinaryMessenger>* messenger;
 @property(readonly, nonatomic) NSMutableDictionary* players;
-@property(readonly, nonatomic) NSObject<FlutterPluginRegistrar>* registrar;
-
+@property(readonly, weak, nonatomic) NSObject<FlutterPluginRegistrar>* registrar;
```
cyanglaz: who retains `registrar`?  in https://github.com/flutter/engine/blob/master/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm#L666-L673
It seems like we autorelease the registrar.

packages/video_player/video_player/pubspec.yaml
gaaclarke: I guess this is a grey area but this is a significant enough change and represents a new feature (support for add-to-app) so I think an increment of the third number is better.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
gaaclarke: If you have the latest version of the engine, FlutterEngine holds on to the registrar.  If this is a strong property it would result in a cycle.  On old engines it wouldn't work.  We can either bump the required engine or adjust the FlutterEngine to not retain the registrars.

cyanglaz: Oh, never mind. I didn't see the line that engine holds the registrar. This is good! Thanks.

packages/video_player/video_player/pubspec.yaml
cyanglaz: SG!

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
gaaclarke: Switched to a strong retain to be backwards compatible with older version of flutter.  The registrar has a weak link to the engine so it shouldn't block the deletion of the engine:
https://github.com/flutter/engine/blob/fdabcaddf45deaf93a5238c99c2acbeb25a3a027/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm#L717

gaaclarke:
The presubmit passed (https://cirrus-ci.com/task/5095163986968576) but for some reason triggered itself again and hung.

PR #2531, opened 2020-02-15
[webview_flutter] Add Android support for scaling a web page

bparrishMines:
## Description

Adds Android support for scaling a web page. 

I have not been able to find a simple equivalent method for [WKWebview](https://developer.apple.com/documentation/webkit/wkwebview?language=objc) on iOS.

Some explored options:
* [setMagnification:centeredAtPoint:](https://developer.apple.com/documentation/webkit/wkwebview/1414996-setmagnification?language=objc) is only supported on Mac 10+.
* [UIScrollView.zoomScale](https://developer.apple.com/documentation/uikit/uiscrollview/1619419-zoomscale?language=objc). This only sets the zoom of the Webview.
* [scalePageToFit](https://developer.apple.com/documentation/uikit/uiwebview/1617950-scalespagetofit?language=objc) in conjunction with zoom. Only supported for [UIWebView](https://developer.apple.com/documentation/uikit/uiwebview?language=objc), while we use `WKWebView`. `UIWebView` is now deprecated.
* Adding width `viewport` meta tag: https://stackoverflow.com/a/26583062/9711864. I believe this could work, but I haven't tested it.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/50837

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart
+++ b/packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart
@@ -563,7 +563,7 @@ void main() {
       () => controller.setInitialScale(150),
       throwsA(isA<AssertionError>()),
     );
-  });
+  }, skip: defaultTargetPlatform != TargetPlatform.android);
 
   group('NavigationDelegate', () {
     final String blankPage = "<!DOCTYPE html><head></head><body></body></html>";
--- a/packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart
+++ b/packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart
@@ -536,6 +536,35 @@ void main() {
     expect(title, 'Some title');
   });
 
+  testWidgets('setInitialScale', (WidgetTester tester) async {
+    final Completer<WebViewController> controllerCompleter =
+        Completer<WebViewController>();
+
+    await tester.pumpWidget(
+      Directionality(
+        textDirection: TextDirection.ltr,
+        child: WebView(
+          key: GlobalKey(),
+          initialUrl: 'https://flutter.dev/',
+          onWebViewCreated: (WebViewController controller) {
+            controllerCompleter.complete(controller);
+          },
+        ),
+      ),
+    );
+
+    final WebViewController controller = await controllerCompleter.future;
+    expect(controller.setInitialScale(50), completes);
+    expect(
+      () => controller.setInitialScale(-1),
+      throwsA(isA<AssertionError>()),
+    );
+    expect(
+      () => controller.setInitialScale(150),
+      throwsA(isA<AssertionError>()),
+    );
+  });
+
   group('NavigationDelegate', () {
     final String blankPage = "<!DOCTYPE html><head></head><body></body></html>";
     final String blankPageEncoded = 'data:text/html;charset=utf-8;base64,' +
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -168,6 +168,25 @@ abstract class WebViewPlatformController {
         "WebView getTitle is not implemented on the current platform");
   }
 
+  /// Sets the initial scale of page content.
+  ///
+  /// Only supported on Android.
+  ///
+  /// [scaleInPercent] must be between 0 and 100 inclusive.
+  ///
+  /// The behavior for the default initial scale depends on whether the WebView
+  /// supports the "viewport" HTML meta tag, whether the WebView is set to use a
+  /// wide viewport, and whether the WebView loads pages in overview mode, that
+  /// is, zooms out the content to fit on screen by width.
+  ///
+  /// If the content fits into the WebView control by width, then the zoom is
+  /// set to 100%. When loading wide content and the WebView is supporting wide
+  /// content, the content will be zoomed out to be fit by width into the
+  /// WebView control, otherwise not. If initial scale is greater than 0,
+  /// WebView starts with this value as initial scale.
+  ///
+  /// Please note that unlike the scale properties in the viewport meta tag,
+  /// this method doesn't take the screen density into account.
   Future<void> setInitialScale(int scaleInPercent) {
     throw UnimplementedError(
         "WebView setInitialScale is not implemented on the current platform");
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -167,6 +167,11 @@ abstract class WebViewPlatformController {
     throw UnimplementedError(
         "WebView getTitle is not implemented on the current platform");
   }
+
+  Future<void> setInitialScale(int scaleInPercent) {
+    throw UnimplementedError(
+        "WebView setInitialScale is not implemented on the current platform");
+  }
 }
 
 /// A single setting for configuring a WebViewPlatform which may be absent.
--- a/packages/webview_flutter/lib/platform_interface.dart
+++ b/packages/webview_flutter/lib/platform_interface.dart
@@ -167,30 +167,6 @@ abstract class WebViewPlatformController {
     throw UnimplementedError(
         "WebView getTitle is not implemented on the current platform");
   }
-
-  /// Sets the initial scale of page content.
-  ///
-  /// Only supported on Android.
-  ///
-  /// [scaleInPercent] must be between 0 and 100 inclusive.
-  ///
-  /// The behavior for the default initial scale depends on whether the WebView
-  /// supports the "viewport" HTML meta tag, whether the WebView is set to use a
-  /// wide viewport, and whether the WebView loads pages in overview mode, that
-  /// is, zooms out the content to fit on screen by width.
-  ///
-  /// If the content fits into the WebView control by width, then the zoom is
-  /// set to 100%. When loading wide content and the WebView is supporting wide
-  /// content, the content will be zoomed out to be fit by width into the
-  /// WebView control, otherwise not. If initial scale is greater than 0,
-  /// WebView starts with this value as initial scale.
-  ///
-  /// Please note that unlike the scale properties in the viewport meta tag,
-  /// this method doesn't take the screen density into account.
-  Future<void> setInitialScale(int scaleInPercent) {
-    throw UnimplementedError(
-        "WebView setInitialScale is not implemented on the current platform");
-  }
 }
 
 /// A single setting for configuring a WebViewPlatform which may be absent.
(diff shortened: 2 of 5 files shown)


packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -167,6 +167,30 @@ abstract class WebViewPlatformController {
...
+  ///
+  /// If the content fits into the WebView control by width, then the zoom is
+  /// set to 100%. When loading wide content and the WebView is supporting wide
+  /// content, the content will be zoomed out to be fit by width into the
+  /// WebView control, otherwise not. If initial scale is greater than 0,
+  /// WebView starts with this value as initial scale.
+  ///
+  /// Please note that unlike the scale properties in the viewport meta tag,
+  /// this method doesn't take the screen density into account.
+  Future<void> setInitialScale(int scaleInPercent) {
```
amirh: If we ever want to support setting the scale to a fraction of a percent we would need a breaking change. Maybe we should start this with a double to be future compatible?

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -167,6 +167,30 @@ abstract class WebViewPlatformController {
...
   }
+
+  /// Sets the initial scale of page content.
+  ///
+  /// Only supported on Android.
+  ///
+  /// [scaleInPercent] must be between 0 and 100 inclusive.
+  ///
+  /// The behavior for the default initial scale depends on whether the WebView
+  /// supports the "viewport" HTML meta tag, whether the WebView is set to use a
```
amirh: How do I know if my webview supports the viewport meta tag? how do I know if I'm set to use a wide viewport? what's controlling overview mode?

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -167,6 +167,30 @@ abstract class WebViewPlatformController {
...
+  /// wide viewport, and whether the WebView loads pages in overview mode, that
+  /// is, zooms out the content to fit on screen by width.
+  ///
+  /// If the content fits into the WebView control by width, then the zoom is
```
amirh: I'm not sure I understand this.
Does it mean to say that setting the initial scale to 100% will make the page fit the webview's width?

amirh: What's WebView control?

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -659,6 +659,32 @@ class WebViewController {
...
+  ///
+  /// If the content fits into the WebView control by width, then the zoom is
+  /// set to 100%. When loading wide content and the WebView is supporting wide
+  /// content, the content will be zoomed out to be fit by width into the
+  /// WebView control, otherwise not. If initial scale is greater than 0,
+  /// WebView starts with this value as initial scale.
+  ///
+  /// Please note that unlike the scale properties in the viewport meta tag,
+  /// this method doesn't take the screen density into account.
+  Future<void> setInitialScale(int scaleInPercent) {
```
amirh: Should this be a widget parameter?

packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart
```diff
@@ -536,6 +536,35 @@ void main() {
...
+    final WebViewController controller = await controllerCompleter.future;
+    expect(controller.setInitialScale(50), completes);
+    expect(
+      () => controller.setInitialScale(-1),
+      throwsA(isA<AssertionError>()),
+    );
+    expect(
+      () => controller.setInitialScale(150),
+      throwsA(isA<AssertionError>()),
+    );
```
amirh: Can we test that the value was actually set? Would it affect the value returned by [onScaleChanged](https://developer.android.com/reference/android/webkit/WebViewClient.html#onScaleChanged(android.webkit.WebView,%20float,%20float)) ?

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -167,6 +167,30 @@ abstract class WebViewPlatformController {
...
+  /// set to 100%. When loading wide content and the WebView is supporting wide
+  /// content, the content will be zoomed out to be fit by width into the
+  /// WebView control, otherwise not. If initial scale is greater than 0,
+  /// WebView starts with this value as initial scale.
+  ///
+  /// Please note that unlike the scale properties in the viewport meta tag,
```
amirh: What happens when there's a viewport meta tag and this is also set?

packages/webview_flutter/lib/platform_interface.dart
amirh: Does this apply to all following page loads?

bparrishMines: tldr: I'll change it to a `double`.

The Android method only uses an `int`. Using a `double` would give the impression that changing from `5.1` to `5.4` would be different values.

Also, wouldn't this be true for almost any method that uses an `int`? Should we only use `int`s in situations where it's impossible to switch to a `double`. (e.g. passing a count value)? If so, we should have that written down in a plugin style guide.

In an instance where the Android method changes, wouldn't it make since to have it as a breaking change since the underlying platform (Java) method is different. 

If you're referring to iOS requiring a `double`, iOS doesn't use a percent, they set a value between a min and max. (e.g.  [minZoomScale](https://developer.apple.com/documentation/uikit/uiscrollview/1619428-minimumzoomscale?language=objc)). I suppose using a `double` would be more accurate than an `int`, so I guess would make sense to use a `double` despite it causing slight confusion for Android.

packages/webview_flutter/lib/platform_interface.dart
bparrishMines: These values are a part of Android [WebSettings](https://developer.android.com/reference/android/webkit/WebSettings). 
https://developer.android.com/reference/android/webkit/WebSettings.html#getUseWideViewPort()
https://developer.android.com/reference/android/webkit/WebSettings.html#getLoadWithOverviewMode()

I wanted to explain what the default behavior was without needing to add multiple other methods or refer to Android classes.

I decided against adding `boolean useWideViewPort` and `boolean getLoadWithOverviewMode` as optional parameters to `setInitialScale`, because I didn't want to expand the scope of this method in a critical issue PR.

packages/webview_flutter/lib/platform_interface.dart
bparrishMines: I think it depends on the webpage and what the settings for 
https://developer.android.com/reference/android/webkit/WebSettings.html#getUseWideViewPort()
https://developer.android.com/reference/android/webkit/WebSettings.html#getLoadWithOverviewMode() are.

For example if you set it to 20:
Screen Shot 2020-02-18 at 4 52 54 PM

If you set it to 100:
Screen Shot 2020-02-18 at 4 54 55 PM

packages/webview_flutter/lib/platform_interface.dart
bparrishMines: Yes it does. If we go with the meta tag on iOS, then we will have to maintain this parameter.

packages/webview_flutter/lib/webview_flutter.dart
bparrishMines: done

bparrishMines:
> > * Adding width `viewport` meta tag: https://stackoverflow.com/a/26583062/9711864. I believe this could work, but I haven't tested it.
> 
> I feel like we should try this, we'd probably need something less naive (so we could deal with an already existing viewport tag in a consistent way across platforms)

I'm not sure it's a good idea to include this on iOS. We are essentially adding a hacky solution to a feature that noone has asked for. I think it would be more trouble than it's worth based on priorities.

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -261,9 +264,30 @@ class WebSettings {
   /// See also: [WebView.gestureNavigationEnabled]
   final bool gestureNavigationEnabled;
 
+  /// Sets the initial scale of page content.
+  ///
+  /// Only supported on Android.
```
amirh: The platform interface shouldn't care which platform supports it.

packages/webview_flutter/lib/platform_interface.dart
amirh: Please explain this in the doc comment

packages/webview_flutter/lib/platform_interface.dart
amirh: These are just the questions I had in mind when I read this documentation, we should answer these in the doc to the best of our ability.

packages/webview_flutter/lib/platform_interface.dart
amirh: If we think it's reasonable that a value would need to be double in the future (which I think is the case here) we should make it a double as it doesn't cost much and gives us better future compatibility.

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -261,9 +264,30 @@ class WebSettings {
...
+  ///
+  /// Must be between 0 and 100 inclusive.
+  ///
+  /// The behavior for the default initial scale depends on whether the WebView
```
amirh: I'm still having a hard time understanding this... 😕

packages/webview_flutter/lib/src/webview_method_channel.dart
```diff
@@ -148,7 +151,8 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
   /// This is used for the `creationParams` argument of the platform views created by
   /// [AndroidWebViewBuilder] and [CupertinoWebViewBuilder].
   static Map<String, dynamic> creationParamsToMap(
-      CreationParams creationParams) {
+    CreationParams creationParams,
```
amirh: Is this dartfmt?

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -319,6 +320,27 @@ class WebView extends StatefulWidget {
   /// The default policy is [AutoMediaPlaybackPolicy.require_user_action_for_all_media_types].
   final AutoMediaPlaybackPolicy initialMediaPlaybackPolicy;
 
+  /// Sets the initial scale of page content.
+  ///
+  /// Only supported on Android.
```
amirh: What should a user expect on unsupported platforms?

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -319,6 +320,27 @@ class WebView extends StatefulWidget {
...
+  ///
```
amirh: What's the behavior when this is null?

bparrishMines:
Closing this in favor of changing `<meta>` tags in html.

PR #2544, opened 2020-02-20
[video_player] upgraded video_player to use pigeon

gaaclarke:
## Description

Upgraded to use pigeon, parent commit: https://github.com/flutter/plugins/pull/2617

## Related Issues

Related: https://github.com/flutter/flutter/issues/32930

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
```diff
@@ -0,0 +1,338 @@
...
+
+    private String package;
+    public String getPackage() { return package; }
+    public void setPackage(String setterArg) { this.package = setterArg; }
+
+    private String formatHint;
+    public String getFormatHint() { return formatHint; }
+    public void setFormatHint(String setterArg) { this.formatHint = setterArg; }
+
+    HashMap toMap() {
```
xster: I'm thinking, since the APIs are directional in the IDL, this should theoretically never be needed right? Could we omit this by default unless we annotate it somehow?

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
```diff
@@ -0,0 +1,338 @@
...
+      PositionMessage fromMapResult = new PositionMessage();
+      fromMapResult.textureId = (int)map.get("textureId");
+      fromMapResult.position = (int)map.get("position");
+      return fromMapResult;
+    }
+  }
+
+  /** Generated interface from Pigeon that represents a handler of messages from Flutter.*/
+  public interface VideoPlayerApi {
+    VoidMessage initialize(VoidMessage arg);
```
xster: ultra nit, the argument name should match the IDL's argument name :)

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
```diff
@@ -0,0 +1,338 @@
...
+    PositionMessage position(TextureMessage arg);
+    VoidMessage seekTo(PositionMessage arg);
+    VoidMessage pause(TextureMessage arg);
+
+    /** Sets up an instance of `VideoPlayerApi` to handle messages through the `binaryMessenger` */
+    public static void setup(BinaryMessenger binaryMessenger, VideoPlayerApi api) {
+      {
+        BasicMessageChannel<Object> channel =
+            new BasicMessageChannel<Object>(binaryMessenger, "dev.flutter.pigeon.VideoPlayerApi.initialize", new StandardMessageCodec());
+        channel.setMessageHandler(new BasicMessageChannel.MessageHandler<Object>() {
```
xster: I suspect we might have to do some state keeping at some point to let the use remove these handlers when time comes to clean up. I think you fixed some bugs recently on letting plugins unregister on iOS too.

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
```diff
@@ -85,109 +85,86 @@ private void onDestroy() {
...
-      case "dispose":
-        player.dispose();
-        videoPlayers.remove(textureId);
-        result.success(null);
-        break;
-      default:
-        result.notImplemented();
-        break;
+  TextureMessage create(CreateMessage arg) {
+    TextureRegistry.SurfaceTextureEntry handle =
```
xster: I wish GitHub was fancy enough to have critique style detect-moves. I'll just assume these are simple moves.

packages/video_player/video_player/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -426,6 +415,7 @@
 			baseConfigurationReference = 9740EEB21CF90195004384FC /* Debug.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				DEVELOPMENT_TEAM = S8QB4VV633;
```
xster: don't check these in

packages/video_player/video_player/ios/Classes/messages.h
```diff
@@ -0,0 +1,57 @@
...
+@end
+
+@interface FLTPositionMessage : NSObject 
+@property(nonatomic, strong) NSNumber * textureId;
+@property(nonatomic, strong) NSNumber * position;
+@end
+
+@protocol FLTVideoPlayerApi
+-(FLTVoidMessage *)initialize:(FLTVoidMessage*)input error:(FlutterError **)error;
+-(FLTTextureMessage *)create:(FLTCreateMessage*)input error:(FlutterError **)error;
```
xster: Ditto here, can this just return an NSNumber?

packages/video_player/video_player/ios/Classes/messages.m
```diff
@@ -0,0 +1,307 @@
+// Autogenerated from Dartle.
+#import "messages.h"
+#import <Flutter/Flutter.h>
+
+#if !__has_feature(objc_arc)
+#error File requires ARC to be enabled.
+#endif
+
+static NSDictionary* wrapResult(NSDictionary *result, FlutterError *error) {
```
xster: uber ultra nit: be consistent on where in the file you add this static between java/objc

packages/video_player/video_player/ios/Classes/messages.m
```diff
@@ -0,0 +1,307 @@
...
+        nil];
+  }
+  return [NSDictionary dictionaryWithObjectsAndKeys:
+      (result ? result : [NSNull null]), @"result",
+      errorDict, @"error",
+      nil];
+}
+
+@interface FLTVoidMessage ()
++(FLTVoidMessage*)fromMap:(NSDictionary*)dict;
```
xster: Ditto here. If the APIs are directional, we don't need both methods.

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
```diff
@@ -0,0 +1,338 @@
...
+    static PositionMessage fromMap(HashMap map) {
+      PositionMessage fromMapResult = new PositionMessage();
+      fromMapResult.textureId = (int)map.get("textureId");
+      fromMapResult.position = (int)map.get("position");
+      return fromMapResult;
+    }
+  }
+
+  /** Generated interface from Pigeon that represents a handler of messages from Flutter.*/
+  public interface VideoPlayerApi {
```
xster: I have a feeling we don't have to do any proguarding of any sort currently since the user has to implement this interface and actually use all the generated classes at some point. Which is super awesome. Do you think that's right?

packages/video_player/video_player/ios/Classes/messages.m
```diff
@@ -0,0 +1,307 @@
...
+  return result;
+}
+-(NSDictionary*)toMap {
+  return [NSDictionary dictionaryWithObjectsAndKeys:(self.textureId ? self.textureId : [NSNull null]), @"textureId", (self.position ? self.position : [NSNull null]), @"position", nil];
+}
+@end
+
+void FLTVideoPlayerApiSetup(id<FlutterBinaryMessenger> binaryMessenger, id<FLTVideoPlayerApi> api) {
+  {
+    FlutterBasicMessageChannel *channel =
```
xster: ditto here. I think we'll have to have a way to unregister at some point.

xster: ditto here. I think we'll have to have a way to unregister at some point.

packages/video_player/video_player/pigeons/messages.dart
```diff
@@ -0,0 +1,51 @@
...
+  VoidMessage dispose(TextureMessage msg);
+  VoidMessage setLooping(LoopingMessage msg);
+  VoidMessage setVolume(VolumeMessage msg);
+  VoidMessage play(TextureMessage msg);
+  PositionMessage position(TextureMessage msg);
+  VoidMessage seekTo(PositionMessage msg);
+  VoidMessage pause(TextureMessage msg);
+}
+
+void setupPigeon(PigeonOptions opts) {
```
xster: So in this whole file, only the import line and this function are required, everything else is freeform right? I wish we can somehow make this distinction clearer. Maybe we need a pub run to generate the dart template and this is pre-filled with a big dartdoc saying you need to fill this.

packages/video_player/video_player/pubspec.yaml
```diff
@@ -17,13 +17,15 @@ flutter:
 
 dependencies:
   meta: "^1.0.5"
-  video_player_platform_interface: ^1.0.1
+  video_player_platform_interface:
+    path: ../video_player_platform_interface
   # The design on https://flutter.dev/go/federated-plugins was to leave
```
xster: your change invalidates this comment right?

packages/video_player/video_player/pubspec.yaml
```diff
@@ -32,6 +34,8 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   pedantic: ^1.8.0
+  pigeon:
+    path: ../../../../packages/packages/pigeon
```
xster: We can probably just reference a version here right? If we were to actually check it in.

packages/video_player/video_player_platform_interface/lib/messages.dart
```diff
@@ -0,0 +1,271 @@
...
+    if (replyMap['error'] != null) {
+      Map error = replyMap['error'];
+      throw PlatformException(
+          code: error['code'],
+          message: error['message'],
+          details: error['details']);
+    } else {
+      return VoidMessage._fromMap(replyMap['result']);
+    }
+
```
xster: nit: some whitespace unbalance here

packages/video_player/video_player_platform_interface/lib/messages.dart
```diff
@@ -0,0 +1,271 @@
...
+    result.textureId = dartleMap["textureId"];
+    result.position = dartleMap["position"];
+    return result;
+  }
+}
+
+class VideoPlayerApi {
+  Future<VoidMessage> initialize(VoidMessage arg) async {
+    Map requestMap = arg._toMap();
+    BasicMessageChannel channel =
```
xster: I guess this part is a bit different from the receiving side. I get the ergonomics of it but I wonder what's the cost (if a function is invoked on each frame). Is it reasonable to save it somewhere?

gaaclarke:
@xster the Android and iOS example apps both work now, yay.

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
gaaclarke: Yep, should be the case.  There are instances where "toMap" or "fromMap" might not be used for a class if the communication is one directional.

packages/video_player/video_player/pubspec.yaml
gaaclarke: I'll change that when i switch from being a draft.

packages/video_player/video_player/pigeons/messages.dart
gaaclarke: This function isn't required.  All of these flags can be supplied at command-line.  I added this function so people can save arguments.

xster: right, understood. 

1- I just meant how can we provide a clearer affordance to this thing? Maybe just a doc in the readme is ok. But since we're telling people to create this file from scratch, it would be good to have some template available. 

2- kinda unrelated, can we tweak the name to not have `setup` in there? The generated java/objc code has the word `setup` in the static functions too and they mean pretty different things than this `setup`. Could be confusing if people cognitively mixed those things. Maybe just configurePigeon or some such.

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
```diff
@@ -0,0 +1,446 @@
...
+
+  /** Generated class from Pigeon that represents data sent in messages. */
+  public static class CreateMessage {
+    private String asset;
+
+    public String getAsset() {
+      return asset;
+    }
+
+    public void setAsset(String setterArg) {
```
xster, line 50: forgot if we talked about this before. Perhaps for a v1.1. Can't all these be construction arguments rather than setters? Since the "receiver" can theoretically take this object as is and just pass it around deeper into their application as immutable data. It's harder to do if it's mutable.

packages/video_player/video_player/pigeons/messages.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:pigeon/pigeon_lib.dart';
```
xster, line 1: nit, the convention is to have the main dart file be the same name as the package, e.g. package:pigeon/pigeon.dart

packages/video_player/video_player/ios/Classes/messages.m
```diff
@@ -0,0 +1,289 @@
...
+}
+- (NSDictionary *)toMap {
+  return [NSDictionary
+      dictionaryWithObjectsAndKeys:(self.textureId ? self.textureId : [NSNull null]), @"textureId",
+                                   (self.volume ? self.volume : [NSNull null]), @"volume", nil];
+}
+@end
+
+@implementation FLTPositionMessage
++ (FLTPositionMessage *)fromMap:(NSDictionary *)dict {
```
cyanglaz, line 134: maybe indicate the return object is "nonnull".

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
```diff
@@ -429,12 +430,9 @@ @interface FLTVideoPlayerPlugin ()
 
 @implementation FLTVideoPlayerPlugin
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar>*)registrar {
-  FlutterMethodChannel* channel =
-      [FlutterMethodChannel methodChannelWithName:@"flutter.io/videoPlayer"
-                                  binaryMessenger:[registrar messenger]];
   FLTVideoPlayerPlugin* instance = [[FLTVideoPlayerPlugin alloc] initWithRegistrar:registrar];
-  [registrar addMethodCallDelegate:instance channel:channel];
   [registrar publish:instance];
+  FLTVideoPlayerApiSetup(registrar.messenger, instance);
```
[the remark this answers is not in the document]
gaaclarke: It's the same error then if someone sent a message to a channel that hasn't had a callback set for it.  It doesn't say anything specifically about Pigeon.  Thats good feedback.  I'll look into that for future versions.

packages/video_player/video_player/ios/Classes/messages.m
gaaclarke: It's defined in an NS_ASSUME_NONNULL_BEGIN, so it is `nonnull`.

packages/video_player/video_player/ios/Classes/messages.m
```diff
@@ -0,0 +1,289 @@
...
+  FLTTextureMessage *result = [[FLTTextureMessage alloc] init];
+  result.textureId = dict[@"textureId"];
+  if ((NSNull *)result.textureId == [NSNull null]) {
+    result.textureId = nil;
+  }
+  return result;
+}
+- (NSDictionary *)toMap {
+  return
+      [NSDictionary dictionaryWithObjectsAndKeys:(self.textureId ? self.textureId : [NSNull null]),
```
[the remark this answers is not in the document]
gaaclarke: Good call.  I would update this but since it is part of the generated code it is really an issue with pigeon, not this PR.  That requires republishing pigeon, since it's a nit I'll address it later if you don't mind.

packages/video_player/video_player/pigeons/messages.dart
gaaclarke: https://github.com/flutter/flutter/issues/56495

packages/video_player/video_player/ios/Classes/messages.m
cyanglaz: Ah, yeah! I always forgot bout the NS_ASSUME_NONNULL thingy.

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
gaaclarke: Good idea I've considered it i n the past, just haven't had the time: https://github.com/flutter/flutter/issues/56496

Since java doesn't have named parameters i've never prioritized it to keep the generated code  similar across platforms.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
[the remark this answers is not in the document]
cyanglaz: Sure. I guess documentations on the method would also help in some patch release.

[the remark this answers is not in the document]
gaaclarke: https://github.com/flutter/flutter/issues/56497

packages/video_player/video_player/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: video_player
 description: Flutter plugin for displaying inline video with other Flutter
   widgets on Android and iOS.
-version: 0.10.9+1
+version: 0.10.10
```
cyanglaz, line 4: Does this PR change the public api of the plugin?
We usually only update minors when there are non-breaking public api changes.

gaaclarke: It doesn't change the API.  It isn't a breaking change in that sense.  It adds no new functionality from the users perspective.

cyanglaz: if it doesn't add anything from users perspective, we should change it to 0.10.9+2 so users 
who depends on `video_player: ^0.10.9^` can get this update.

gaaclarke: My understanding is that the "+N" nomenclature is for changes that don't change the binary.  This doesn't change the intention of the code, but the binary is definitely changed.

gaaclarke:
Landing on red, the only breaking CI is submit-queue, this should fix it.  @cyanglaz I had to make a few changes after your LGTM, but they were all fixing tests in case you want to review them.

antholeole:
What is the command used to invoke pigeon here? running `flutter pub run pigeon --input pigeons/messages.dart` fails with
 
`Error: This requires the 'non-nullable' experiment to be enabled.
Try enabling this experiment by adding it to the command line when compiling and running`

I've attempted to add the flag `--enable-experiment:non-nullable` but the same error message pops up. What is the proper command to invoke pigeon on this project?

PR #2547, opened 2020-02-21
[google_sign_in] Add new Oauth scope methods to google_sign_in_platform_interface.

emerssso:
## Description

Adds two new methods, `hasGrantedScope` and `requestScope`, to the `GoogleSginInPlatform` class and the method channel implementation.

I have a second [branch](https://github.com/flutter/plugins/compare/master...emerssso:scope-apis-mobile) queued up for the implementations, but my understanding of the federated plugin design is that this has to be merged and pushed to pub first?

## Related Issues

* flutter/flutter#50208

## Checklist

- [X] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
+++ b/packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
@@ -124,7 +124,7 @@ abstract class GoogleSignInPlatform {
 
   /// Checks to see if [scopes] have been granted, returns those which have not.
   Future<List<String>> listMissingScopes(List<String> scopes) async {
-    throw UnimplementedError('hasGrantedScope() has not been implmented.');
+    throw UnimplementedError('listMissingScopes() has not been implmented.');
   }
 
   /// Requests the user grant an additional Oauth [scope].
--- a/packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
+++ b/packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
@@ -121,4 +121,14 @@ abstract class GoogleSignInPlatform {
   Future<void> clearAuthCache({@required String token}) async {
     throw UnimplementedError('clearAuthCache() has not been implemented.');
   }
+
+  /// Indicates whether this Oauth [scope] has been granted.
+  Future<bool> hasGrantedScope(String scope) async {
+    throw UnimplementedError('hasGrantedScope() has not been implmented.');
+  }
+
+  /// Requests the user grant an additional Oauth [scope].
+  Future<bool> requestScope(String scope) async {
+    throw UnimplementedError('requestScope() has not been implmented.');
+  }
 }
--- a/packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
+++ b/packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
@@ -123,7 +123,7 @@ abstract class GoogleSignInPlatform {
   }
 
   /// Requests the user grants additional Oauth [scopes].
-  /// 
+  ///
   /// Scopes should come from the full  list
   /// [here](https://developers.google.com/identity/protocols/googlescopes).
   Future<bool> requestScopes(List<String> scopes) async {
--- a/packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
+++ b/packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
@@ -122,12 +122,7 @@ abstract class GoogleSignInPlatform {
     throw UnimplementedError('clearAuthCache() has not been implemented.');
   }
 
-  /// Checks to see if [scopes] have been granted, returns those which have not.
-  Future<List<String>> listMissingScopes(List<String> scopes) async {
-    throw UnimplementedError('listMissingScopes() has not been implmented.');
-  }
-
-  /// Requests the user grant an additional Oauth [scope].
+  /// Requests the user grants additional Oauth [scopes].
   Future<bool> requestScopes(List<String> scopes) async {
     throw UnimplementedError('requestScopes() has not been implmented.');
   }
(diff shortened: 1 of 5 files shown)


emerssso:
Re-created PR since I messed up a rebase. Sorry about that.

mehmetf:
@ditman This is an internal request. I am not sure how to review it since I am used to reviewing end to end implementations of plugin code. This is just a change on interface. Could you please advise?

ditman:
@mehmetf This looks good, the only thing "missing" (which I'm assuming is on a separate PR for the `google_sign_in` package) are the changes to the native code to implement the new MethodChannel messages (`hasGrantedScope` and `requestScope `), in case they're new?

My only comment may be that this methods seem to be called `hasPermissions` and `requestPermissions` on the [android side](https://developers.google.com/identity/sign-in/android/additional-scopes), and maybe you want to keep that name on the method channel call as well.

(This also seems implementable on web through the [hasGrantedScopes](https://developers.google.com/identity/sign-in/web/reference#googleuserhasgrantedscopesscopes) and [grant](https://developers.google.com/identity/sign-in/web/reference#googleusergrantoptions) methods?)

emerssso:
@ditman, the implementations for Android/iOS/web are in #2562. I split them out since it seemed like the build failed due to the 1.1.0 not being in pub?

Happy to swap out the method names, I went with 'scope' since that seemed to be what was used in the plugin (i.e. constructor).

ditman:
@emerssso re: naming: whatever makes more sense when using the plugin is fine, no need to rename.

Re: splitting the PR: yes, it is needed to publish the packages in the correct order. Thanks for the link!

emerssso:
Suggested offline that this would be improved by allowing multiple new scopes to be requested at once so the user doesn't potentially have to be prompted multiple times in that case. 

I'll work on that and add the changes needed to each PR.

emerssso:
Updated both methods to allow the caller to pass multiple scopes. The flow I'm envisioning is
```
final missingScopes = await googleSignIn.listMissingScopes(newScopes);
// handle missing scopes, i.e. showing justifications,etc

final granted = await googleSignIn.requestScopes(missingScopes);
if(granted) doStuff();
```

Still working on updating the implementations; wanted to get feedback on this change to the API.

packages/google_sign_in/google_sign_in_platform_interface/lib/src/method_channel_google_sign_in.dart
```diff
@@ -81,10 +81,10 @@ class MethodChannelGoogleSignIn extends GoogleSignInPlatform {
 
   @override
   Future<List<String>> listMissingScopes(List<String> scopes) {
-    return channel.invokeMethod<List<String>>(
+    return channel.invokeMethod<List<dynamic>>(
       'listMissingScopes',
       <String, List<String>>{'scopes': scopes},
-    );
+    ).then((result) => List<String>.from(result));
```
ditman: Does this work?

`(result) => result.cast<String>()`;

ditman: (I wonder why you needed to change from List<String> to List<dynamic>?)

ditman:
> wanted to get feedback on this change to the API.

@emerssso Can we do the "missingScopes" check internally, so the user doesn't have to care about the ones that are already granted? I think that'd make the API slightly easier to use.

packages/google_sign_in/google_sign_in_platform_interface/lib/src/method_channel_google_sign_in.dart
emerssso: I think that would work, Effective Dart says to prefer List.from over cast: https://dart.dev/guides/language/effective-dart/usage#dont-use-cast-when-a-nearby-operation-will-do

invokeMethod apparently doesn't support generic type parameters other than dynamic? With the initial approach, the future never completed. This change seemed to resolve the issue I was seeing in manual testing, but I've never seen dart act this way before.
https://api.flutter.dev/flutter/services/MethodChannel/invokeMethod.html

emerssso:
> Can we do the "missingScopes" check internally, so the user doesn't have to care about the ones that are already granted? I think that'd make the API slightly easier to use.

That would meet my immediate needs. I've seen uses of the "missingScopes" flow, but it's not necessary.

ditman:
@emerssso just trying to keep the API surface of the plugin as small as possible (and as close as possible to the Google Sign In API itself; the closest thing I've seen there is `getGrantedScopes`, but nothing similar to `getMissingScopes`)

emerssso:
@ditman Makes sense. I'll consolidate them to just `requestScopes`.

ditman:
This makes sense to me, but please seek approval from @cyanglaz since this affects all platforms, not only web!

_(PS: Thanks for the web implementation!)_

packages/google_sign_in/google_sign_in_platform_interface/lib/src/method_channel_google_sign_in.dart
ditman: So weird, I've seen this done in other places, but I didn't understand why. It might be a bug in the MethodChannel :/

packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
```diff
@@ -121,4 +121,9 @@ abstract class GoogleSignInPlatform {
   Future<void> clearAuthCache({@required String token}) async {
     throw UnimplementedError('clearAuthCache() has not been implemented.');
   }
+
+  /// Requests the user grants additional Oauth [scopes].
+  Future<bool> requestScopes(List<String> scopes) async {
```
cyanglaz: what can be the values of the `scopes`? Maybe we should either make an enum or have some concrete list of what the values can be.

ditman: @cyanglaz There's a TON of auth scopes, here's the list: https://developers.google.com/identity/protocols/googlescopes (each API that you want to use has its own set of scopes)

ditman: Yes, that's a good point! It can probably be added to the README.md.

emerssso: Added a link in the dartdoc. Looks like this is already mentioned in the README.md [here](https://github.com/flutter/plugins/blob/master/packages/google_sign_in/google_sign_in/README.md#use-the-plugin).

PR #2552, opened 2020-02-24
[url_launcher] fix: url_launcher - updated _launchUniversalLinkIos and Added Tests

jddeep:
## Description

updated _launchUniversalLinkIos  to launch the url passed. It could also be done that we remove the url argument passed and open the hardcoded url. Either of the case needs to be done.
## Checklist

- [ ] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/url_launcher/url_launcher/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.4.3
+
+* Fixed the launchUniversalLinkIos method.
+
 ## 5.4.2
 
 * Make the pedantic dev_dependency explicit.
--- a/packages/url_launcher/url_launcher/example/lib/main.dart
+++ b/packages/url_launcher/url_launcher/example/lib/main.dart
@@ -91,15 +91,15 @@ class _MyHomePageState extends State<MyHomePage> {
   }
 
   Future<void> _launchUniversalLinkIos(String url) async {
-    if (await canLaunch('https://youtube.com')) {
+    if (await canLaunch(url)) {
       final bool nativeAppLaunchSucceeded = await launch(
-        'https://youtube.com',
+        url,
         forceSafariVC: false,
         universalLinksOnly: true,
       );
       if (!nativeAppLaunchSucceeded) {
         await launch(
-          'https://youtube.com',
+          url,
           forceSafariVC: true,
         );
       }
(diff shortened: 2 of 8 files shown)


franciscojma86:
LGTM but not sure if there might be an initial reason for having that hardcoded link. @cyanglaz added that snippet so might be the best person to approve. 

If approved, please also update https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher_macos/example/lib/main.dart, which is a copy of that example.

jddeep:
@mklim There is already a test written for universal links here - https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher/test/url_launcher_test.dart#L105 and it passes
And sure, I will update the `pubspec.yaml` and CHANGELOG.md version for publishing.

franciscojma86:
@jddeep please change the macOS example app as well. I know it's another package, and a duplicated example is not optimal, but until we have a better solution we should try to keep them as equal as possible.
https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher_macos/example/lib/main.dart.

jddeep:
@franciscojma86 Done.

jddeep:
@mklim So here is a widget test I was writing, but there were some doubts. How would we test whether the url is launchable or not once the button is tapped and is directed to a browser of OS? Also but the line `await tester.tap(browserlaunchBtn);` throws an error while running the test `Failed assertion: boolean expression must not be null`. What do you think?

` testWidgets('Tesing widgets', (WidgetTester tester) async {
    await tester.pumpWidget(MaterialApp(home:MyHomePage(
      title: 'Url launcher Widget Test',
    )));
      Finder browserlaunchBtn = find.widgetWithText(RaisedButton, 'Launch in browser');
      expect(browserlaunchBtn, findsOneWidget);
      await tester.tap(browserlaunchBtn);
  }); `

mklim:
@jddeep that's pretty close. So the missing piece there is that when you click on the button, it's still trying to actually call through the plugin to the platform code to launch the link. That error message you're seeing is the Flutter app expecting the plugin to return a bool and then panicking when it doesn't. For our  purposes here a widget test is basically the same as a unit test. There is no underlying platform, so the plugin can't "really" work and really launch any links. So the thing to do here is to test that the button works is to mock out the Dart part of the plugin that would normally send the calls down to Java/ObjC to actually launch the link, and then verify that the mock Dart plugin interface received a call from the example app asking it to launch the link. I uploaded a working test doing this to mklim/plugins@6a9d4387cb200943043a67e7fce46d862e364650, feel free to patch it into your PR.

jddeep:
@mklim Thanks for the help. I see now. I have updated my test as well. Do we want tests for each button here or testing for one button to check if the url is launchable is cool? Do let me know or we can merge this up.

PR #2553, opened 2020-02-25
[path_provider] Create platform interface

franciscojma86:
## Description

Creates a platform_interface for path_provider. Second step for federation outlined in https://github.com/flutter/flutter/issues/46307#issuecomment-589301237

## Related Issues

https://github.com/flutter/flutter/issues/46307

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/path_provider/path_provider_platform_interface/lib/path_provider_platform_interface.dart
+++ b/packages/path_provider/path_provider_platform_interface/lib/path_provider_platform_interface.dart
@@ -1,0 +1,95 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io' show Directory;
+
+import 'src/enums.dart';
+import 'src/method_channel_path_provider.dart';
+
+import 'package:plugin_platform_interface/plugin_platform_interface.dart';
+
+/// The interface that implementations of path_provider must implement.
+///
+/// Platform implementations should extend this class rather than implement it as `PathProvider`
+/// does not consider newly added methods to be breaking changes. Extending this class
+/// (using `extends`) ensures that the subclass will get the default implementation, while
+/// platform implementations that `implements` this interface will be broken by newly added
+/// [PathProviderPlatform] methods.
+abstract class PathProviderPlatform extends PlatformInterface {
+  /// Constructs a PathProviderPlatform.
+  PathProviderPlatform() : super(token: _token);
+
+  static final Object _token = Object();
+
+  static PathProviderPlatform _instance = MethodChannelPathProvider();
+
+  /// The default instance of [PathProviderPlatform] to use.
+  ///
+  /// Defaults to [MethodChannelPathProvider].
+  static PathProviderPlatform get instance => _instance;
+
+  /// Platform-specific plugins should set this with their own platform-specific
+  /// class that extends [PathProviderPlatform] when they register themselves.
+  static set instance(PathProviderPlatform instance) {
+    PlatformInterface.verifyToken(instance, _token);
+    _instance = instance;
+  }
+
+  /// Path to the temporary directory on the device that is not backed up and is
+  /// suitable for storing caches of downloaded files.
+  Future<Directory> getTemporaryDirectory() {
+    throw UnimplementedError('getTemporaryDirectory() has not been implemented.');
+  }
+  /// Path to a directory where the application may place application support
+  /// files.
+  Future<Directory> getApplicationSupportDirectory() {
+    throw UnimplementedError('getApplicationSupportDirectory() has not been implemented.');
+  }
+
+  /// Path to the directory where application can store files that are persistent,
+  /// backed up, and not visible to the user, such as sqlite.db.
+  Future<Directory> getLibraryDirectory() {
+    throw UnimplementedError('getLibraryDirectory() has not been implemented.');
+  }
+
+  /// Path to a directory where the application may place data that is
+  /// user-generated, or that cannot otherwise be recreated by your application.
+  Future<Directory> getApplicationDocumentsDirectory() {
+    throw UnimplementedError('getApplicationDocumentsDirectory() has not been implemented.');
+  }
+
+  /// Path to a directory where the application may access top level storage.
+  /// The current operating system should be determined before issuing this
+  /// function call, as this functionality is only available on Android.
+  Future<Directory> getExternalStorageDirectory() {
+    throw UnimplementedError('getExternalStorageDirectory() has not been implemented.');
+  }
+
+  /// Paths to directories where application specific external cache data can be
+  /// stored. These paths typically reside on external storage like separate
+  /// partitions or SD cards. Phones may have multiple storage directories
+  /// available.
+  Future<List<Directory>> getExternalCacheDirectories() {
+    throw UnimplementedError('getExternalCacheDirectories() has not been implemented.');
+  }
+
+  /// Paths to directories where application specific data can be stored.
+  /// These paths typically reside on external storage like separate partitions
+  /// or SD cards. Phones may have multiple storage directories available.
+  Future<List<Directory>> getExternalStorageDirectories({
+    /// Optional parameter. See [AndroidStorageDirectory] for more informations on
+    /// how this type translates to Android storage directories.
+    AndroidStorageDirectory type,
+  }) {
+    throw UnimplementedError('getExternalStorageDirectories() has not been implemented.');
+  }
+
+
+  /// Path to the directory where downloaded files can be stored.
+  /// This is typically only relevant on desktop operating systems.
+  Future<Directory> getDownloadsDirectory() {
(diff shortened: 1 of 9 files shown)


packages/path_provider/path_provider_platform_interface/lib/path_provider_platform_interface.dart
```diff
@@ -0,0 +1,102 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io' show Directory;
```
ditman: I'm torn about making 'dart:io' part of the dependencies of the platform_interface. This makes it 100% non-implementable in web at all.

Any chance this platform interface can return paths (String) instead of fully-formed Directories?

franciscojma86: How about I rename those to getXPath instead of getXDirectory, and return the strings. The core plugin would actually make the `Directory` object?

franciscojma86: I just did that. I think it makes sense because it's now actually returning paths, not directories, and the plugin is called PATH_provider :) WDYT?

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
```diff
@@ -60,19 +59,19 @@ class MethodChannelPathProvider extends PathProviderPlatform {
     if (path == null) {
       return null;
     }
-    return Directory(path);
+    return path;
```
ditman: No need for all these null checks, just `return path`

packages/path_provider/path_provider_platform_interface/lib/path_provider_platform_interface.dart
ditman: Yep, the current version is way more portable! That way if we ever need to refactor this, we only need to go to the core plugin and change stuff there.

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
franciscojma86: Nice

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
```diff
@@ -0,0 +1,90 @@
...
+  Platform _platform = const LocalPlatform();
+
+  /// This API is only exposed for the unit tests. It should not be used by
+  /// any code outside of the plugin itself.
+  @visibleForTesting
+  void setMockPathProviderPlatform(Platform platform) {
+    _platform = platform;
+  }
+
+  Future<String> getTemporaryPath() async {
```
ditman: None of this methods need to be `async`, since you're not `await`ing for anything now!

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
```diff
@@ -0,0 +1,90 @@
...
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'enums.dart';
+
+import 'package:flutter/services.dart';
+import 'package:meta/meta.dart';
+import 'package:path_provider_platform_interface/path_provider_platform_interface.dart';
+import 'package:platform/platform.dart';
```
ditman: Strong applause for not using dart:io Platform

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  Future<List<String>> getExternalCachePaths() async {
+    if (!_platform.isAndroid) {
+      throw UnsupportedError('Functionality only available on Android');
+    }
+    final List<String> paths = await methodChannel
+        .invokeListMethod<String>('getExternalCacheDirectories');
+
+    return paths.map((String path) => path).toList();
```
ditman: Why do you need to do this conversion? Doesn't `invokeListMethod<String>` return a `List<String>`?

invokeListMethod already finishes with this:

`return result?.cast<T>();`

If you really, really need this, take a page from the implementation above. You should be able to just do: `return paths.cast<String>();`, right? ([List.cast](https://api.dart.dev/stable/2.7.1/dart-core/List/cast.html) documentation)

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
```diff
@@ -0,0 +1,90 @@
...
+    }
+    return methodChannel.invokeMethod<String>('getStorageDirectory');
+  }
+
+  Future<List<String>> getExternalCachePaths() async {
+    if (!_platform.isAndroid) {
+      throw UnsupportedError('Functionality only available on Android');
+    }
+    final List<String> paths = await methodChannel
+        .invokeListMethod<String>('getExternalCacheDirectories');
```
ditman: You don't need to await here to apply the transformation, you can just return the future returned by doing `.then` on the result of `invokeListMethod`:

```
return methodChannel
    .invokeListMethod<String>('blah')
    .then((List<String> paths) => paths.map((String p) => p).toList());
```

The user of this API needs to await for the `Future` that you're returning here, so there's no need for this method to await. You can just return something that will eventually generate a `List<String>` and be done quickly.

Then, this method doesn't need to be `async` anymore ;)

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
```diff
@@ -0,0 +1,90 @@
...
+      <String, dynamic>{'type': type?.index},
+    );
+
+    return paths.map((String path) => path).toList();
+  }
+
+  Future<String> getDownloadsPath() async {
+    if (!_platform.isMacOS) {
+      throw UnsupportedError('Functionality only available on macOS');
+    }
```
ditman: Same comment I did in your previous platform interface migration. IMO this should be thrown by the Java side (I don't know what to do with the 'getDownloadsDirectory' call, and not in the MethodChannel implementation).

If the Fuchsia version is able to fulfill this, this'd be throwing an exception before it even reaches native code.

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
```diff
@@ -0,0 +1,90 @@
...
+  }) async {
+    if (!_platform.isAndroid) {
+      throw UnsupportedError('Functionality only available on Android');
+    }
+    final List<String> paths = await methodChannel.invokeListMethod<String>(
+      'getExternalStorageDirectories',
+      <String, dynamic>{'type': type?.index},
+    );
+
+    return paths.map((String path) => path).toList();
```
ditman: (Same comments as above)

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
franciscojma86: Not sure I follow.  I'm keeping it here to make sure it's as similar to the current implementation as possible. I think you are right about the fact that the native side would just return `MissingImplementation`. I think if we want that exception as opposed to the `UnsupportedError` that might be a bigger discussion to be consistent with the rest of the plugins.

What do you mean by the fuchsia version example? In theory, nothing other than macOS should reach the native code. Maybe that's why the platform checks were added? To save some calls if we know they are going to fail? Also, maybe for testing :)

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
franciscojma86: I don't. That's remaining from the code that converted the paths to Directory. Removed it and now using the invoke call directly :)

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
franciscojma86: See response above :)

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
franciscojma86: Done

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
ditman: Let's agree to disagree :P

franciscojma86: Oh I don't have a strong posture on this at all :) It's mostly a "Want to change as little as possible from the way it already was". I think @amirh or @cyanglaz  might have a better opinion on this.

PR #2554, opened 2020-02-25
[web] adding a test for e2e web testing.

nturgut:
## Description

(1) Adding a timeout to e2e so any test that doesn't use the MethodChannel won't timeout.
(2) Adding a test for web use case, altering the existing tests.
(3) Altering the previous code not to use dart:io on web
(4) Changing the documentation for future users.
(5) Adding the test to CI

## Related Issues

Fixes: https://github.com/flutter/flutter/issues/50212
Fixes: https://github.com/flutter/flutter/issues/50654

## Checklist

- [x] I signed the [CLA].

## Breaking Change


.cirrus.yml
```diff
@@ -45,6 +45,18 @@ task:
...
+    - name: e2e_web_smoke_test
+      # Tests e2e example test in web.
+      only_if: "changesInclude('.cirrus.yml', 'packages/e2e/**') || $CIRRUS_PR == ''"
+      script:
+        - flutter config --enable-web
+        - git clone https://github.com/flutter/web_installers.git
+        - cd web_installers/packages/web_drivers/
+        - pub get
+        - dart lib/web_driver_installer.dart &
+        - sleep 20
```
collinjackson: Can we make this more deterministic? Maybe look at the output of web_driver_installer and use that to decide when to call flutter drive?

https://superuser.com/questions/375223/watch-the-output-of-a-command-until-a-particular-string-is-observed-and-then-e

packages/e2e/lib/e2e.dart
```diff
@@ -19,10 +19,16 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
     // TODO(jackson): Report test results as they arrive
     tearDownAll(() async {
       try {
-        await _channel.invokeMethod<void>(
-            'allTestsFinished', <String, dynamic>{'results': _results});
+        // For web integration tests we are not using the
```
collinjackson: Maybe we should just check `!kIsWeb` and call `allTestsFinished` if it returns true? This plugin isn't federated yet so there isn't really a need to worry about future platforms, and it seems confusing to have a warning printed on every test run.

packages/e2e/example/test_driver/failure.dart
```diff
@@ -21,7 +21,10 @@ void main() {
 
   testWidgets('failure 1', (WidgetTester tester) async {
     // Build our app and trigger a frame.
-    await tester.pumpWidget(MyApp());
+    app.main();
+
+    // Trigger a frame.
+    await tester.pumpAndSettle();
```
collinjackson: Would it be possible to `await expectLater` below instead of calling this?

I worry about littering tests with `tester.pumpAndSettle()`, it seems kind of mysterious whether it's needed or not. In any case, I hope that `WidgetTester` synchronization will help.

packages/e2e/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.2.4+4
+
+* Add timeout to invokeMethod call, for tests do not use MethodChannel.
```
collinjackson:
```suggestion
* Fixed a hang that occurred on platforms that don't have a `MethodChannel` listener registered.
```

.cirrus.yml
ditman: Hmmm... Why is the previous command run in the background, though?

Wouldn't you be able to remove the `sleep` if the `&` at the end of `dart lib/web_driver_installer.dart` were removed?

nturgut: No it should always run on the background. The code is installing (downloading over the web and extracting) the chrome driver. It also starts running it on port 4444

ditman: Ahhh... maybe splitting the code in "the part that downloads and installs" (being synchronous) and the part that "starts the driver on port 444 in the background" could be split in two separate commands, that way you wouldn't need the sleep (maybe)

nturgut: What I can do is, I can only use web_driver installers for installing. and for launching I can use
chromedriver --port=4444 &

This can remove the need for sleep since it is  starts instantly.

ditman: If that works, I think it's more deterministic than having the sleep there (imagine that installing takes 21 seconds for whatever reason)

nturgut:
The chrome version flutter/plugings CI is suing is 78. It was not supported in web_installers. I'll update web_installers and continue on the PR.

Thanks for the comments.

.cirrus.yml
nturgut: @ditman I send you a PR to fix the webdriver command (https://github.com/flutter/web_installers/pull/4)

ditman: @nturgut just approved it!

packages/e2e/lib/e2e.dart
nturgut: done.

packages/e2e/example/test_driver/failure.dart
nturgut: done.

.cirrus.yml
nturgut: done.

packages/e2e/CHANGELOG.md
nturgut: done.

packages/e2e/example/test_driver/example_e2e_web_test.dart
```diff
@@ -0,0 +1,15 @@
+@TestOn('browser')
```
collinjackson: This isn't a test file, you can remove this

packages/e2e/example/test_driver/example_e2e_web_test.dart
```diff
@@ -0,0 +1,15 @@
...
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:flutter_driver/flutter_driver.dart';
+import 'package:test/test.dart';
```
collinjackson: This isn't a test file, you can remove this

ditman:
> I think this approach will solve the build issue: collinjackson@7fa057d

That's a very nice trick @collinjackson

nturgut:
> I think this approach will solve the build issue: [collinjackson@7fa057d](https://github.com/collinjackson/plugins/commit/7fa057d643db496efabb8caa08c5d38d5f2fcc8a)

it didn't work unfortunately: https://cirrus-ci.com/task/6188264717549568 It still wants to compile each file and in case of example_e2e_web_test it is not able to.

collinjackson:
> It still wants to compile each file and in case of example_e2e_web_test it is not able to.

Actually it looks like the conditional import is working and the tests are running successfully on mobile. Your commit differs from my suggestion in that you left example_e2e_web_test.dart in place and that's where things are not working. You should delete it. 

See https://github.com/flutter/plugin_tools/blob/master/lib/src/drive_examples_command.dart#L52

nturgut:
> example_e2e_web_test.dart

May be I am missing something, but as far as I know if the file  <test_name>_test.dart, does not exist we won't be able to use the following smoke test command:

flutter drive -v --target=test_driver/example_e2e_web.dart -d web-server --release --browser-name=chrome

My suggestions are:
(1) web smoke test will be in another directory
(2) I can change the plugin_tools to run only examples does not contain web

Let me know if these makes sense.

collinjackson:
> > example_e2e_web_test.dart
> 
> May be I am missing something, but as far as I know if the file <test_name>_test.dart, does not exist we won't be able to use the following smoke test command:
> 
> flutter drive -v --target=test_driver/example_e2e_web.dart -d web-server --release --browser-name=chrome
> 
> My suggestions are:
> (1) web smoke test will be in another directory
> (2) I can change the plugin_tools to run only examples does not contain web
> 
> Let me know if these makes sense.

My current proposal is to remove example_e2e_web_test.dart and you'd have two options for running the web integration test:

1. flutter drive -v --target=test_driver/example_e2e.dart -d web-server --release --browser-name=chrome
1. flutter drive -v --target=test_driver/example_e2e_web.dart -d web-server --release --browser-name=chrome --driver=example_e2e_test.dart

I'd recommend the former, because it won't require any change to our existing CI, and you're following existing conventions for how to find the driver boilerplate file.

As for moving the web tests to their own folder -- I am niot sure about using folder names to determine the platform of tests. It's one thing for us to do this for our own CI, but plugin developers are going to mimic whatever we're doing here, and having a solution based on code instead of directory structure seems more flexible to me.

nturgut:
> > > example_e2e_web_test.dart
> > 
> > 
> > May be I am missing something, but as far as I know if the file <test_name>_test.dart, does not exist we won't be able to use the following smoke test command:
> > flutter drive -v --target=test_driver/example_e2e_web.dart -d web-server --release --browser-name=chrome
> > My suggestions are:
> > (1) web smoke test will be in another directory
> > (2) I can change the plugin_tools to run only examples does not contain web
> > Let me know if these makes sense.
> 
> My current proposal is to remove example_e2e_web_test.dart and you'd have two options for running the web integration test: 
> 
> 1. `flutter drive -v --target=test_driver/example_e2e.dart -d web-server --release --browser-name=chrome` would run the main tests entry point and import the web tests. it won't require much of a change to our existing CI, and you're following existing conventions for how to find the driver boilerplate file from the e2e test.
> 2. `flutter drive -v --target=test_driver/example_e2e_web.dart -d web-server --release --browser-name=chrome --driver=example_e2e_test.dart` would run the web test directly. If you go this route you wouldn't need the conditional import, but you'd be introducing more conventions around filenames that might be brittle.
> 
> As for moving the web tests to their own folder -- I am not sure about using folder names to determine the platform of tests. It's one thing for us to do this for our own CI, but plugin developers are going to mimic whatever we're doing here, and an approach based on directory structure baked into all our tools seems less flexible to me than a code solution.

The code runs the tests only if there are under example folder, so if I add a new folder such as web_example or web_smoke it should be ok.

The reason I don't want to remove example_e2e_web_test.dart is soon it would be very different than example_e2e_test.dart by content due to screenshots taken by the driver for web. In the original PR we added more functionality to the driver test: https://github.com/flutter/flutter/pull/49833

For now I can go with the options you suggested but for the future we still need to reconsider differences between driver files.

packages/e2e/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.2.4+4
+
+* Fixed a hang that occurred on platforms that don't have a `MethodChannel` listener registered..
```
collinjackson:
```suggestion
* Fixed a hang that occurred on platforms that don't have a `MethodChannel` listener registered.
```

PR #2561, opened 2020-02-27
Revert "[connectivity_platform_interface] Add `ConnectivityResult.unknown`."

ditman:
Reverts flutter/plugins#2560

### DIFF
--- a/packages/connectivity/connectivity_platform_interface/CHANGELOG.md
+++ b/packages/connectivity/connectivity_platform_interface/CHANGELOG.md
@@ -1,8 +1,3 @@
-## 2.0.0
-
-* Added `ConnectivityResult.unknown`, for the cases where the plugin is unable to determine the
-connectivity status of the device. _(This happens mostly in the `web` platform.)_
-
 ## 1.0.3
 
 * Make the pedantic dev_dependency explicit.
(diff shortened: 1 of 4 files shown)


ditman:
(This might need to be rolled back to "1.0.3+1" if we want to merge this fully green, even after "2.0.0" is removed from pub.dev)

jonasfj:
`2.0.0` should now be removed from pub.dev

ditman:
This change will always have a red pre-submit, since the version script checks against `master` and not what's published in pub.dev. Post-submit, the repo should go green again.

collinjackson:
OK, sounds good.

packages/connectivity/connectivity_platform_interface/CHANGELOG.md
```diff
@@ -1,7 +1,6 @@
-## 2.0.0
+## 1.0.3+1
```
amirh: We don't need to publish a patch bump

ditman: Done

ditman:
Merging on red, since most of the failures are related to versioning, which is what we're trying to straighten out here.

```
Running version check for changed packages
packages/connectivity/connectivity_platform_interface/pubspec.yaml incorrectly updated version.
HEAD: 1.0.3, master: 2.0.0.
Allowed versions: {3.0.0: NextVersionType.BREAKING_MAJOR, 2.1.0: NextVersionType.MINOR, 2.0.1: NextVersionType.PATCH}
```

PR #2563, opened 2020-02-27
fix(video_player): buffering state events missing on Android & Web (fixes flutter/flutter#28494)

chengyuhui:
## Description

Currently the `isBuffering` field on Android does not chage even if the video is buffering, this PR fixed this by sending `bufferingStart` and `bufferingEnd` events in the event handler of ExoPlayer.

## Related Issues

Fixes flutter/flutter#28494

## Checklist

- [X] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]). * No Java tests are available for this part *
- [X] All existing and new tests are passing. * No Java tests are available for this part *
- [x] I signed the [CLA].

## Breaking Change


chengyuhui:
@iskakaushik @cyanglaz Is it possible to have a look at this?

Not a big deal, but quite annoying when the network condition is bad and we can't provide the users with correct message.

obshksw:
Not sure if it is appropriate to add this here since this issue is specifically for Android, but it has the same problem for flutter web... It'd be great to have this fixed there also!

chengyuhui:
> Not sure if it is appropriate to add this here since this issue is specifically for Android, but it has the same problem for flutter web... It'd be great to have this fixed there also!

Just had a quick glance, the issue seems similar (events not triggered). However, I personally have no experience with directly controlling `<video>`s, so I am not sure if I could fix it :(

obshksw:
@chengyuhui Yeah that's what it seemed like. Thanks anyway for taking the time to check it out! :)

p30arena:
This is actually working :+1:

chengyuhui:
I was considering adding some tests, but have no idea how to do that, since it seems nothing in the native part has test associated with it, and this is a platform-specific bug.

cyanglaz:
@chengyuhui You might be able to add something similar to https://github.com/flutter/plugins/blob/master/packages/video_player/video_player/example/test_driver/video_player_e2e.dart
Note that video player e2e tests were not working on iOS so you might want to skip iOS like the other tests.

chengyuhui:
I got the idea, but still wondering how this could be done (controlling the underlying buffering status?). I originally discovered this bug entirely by accident (I am in China and the Apple HLS test stream is really slow, but the buffering indicator is not showing up).

The logic of applying changes to `VideoPlayerValue` based on events from native code is already tested.

cyanglaz:
> The logic of applying changes to `VideoPlayerValue` based on events from native code is already tested.

@chengyuhui Which tests did you mean? Could you link me to the code?

chengyuhui:
> @chengyuhui Which tests did you mean? Could you link me to the code?

It's here: https://github.com/flutter/plugins/blob/master/packages/video_player/video_player/test/video_player_test.dart#L392

cyanglaz:
@chengyuhui This is only testing the dart apis with a mock controller. We might want to test the android implementation. Maybe an e2e test would help us here?

chengyuhui:
The problem is to simulate a stream that requires buffering (ExoPlayer does not seem to have mocking ability). I see some 127.0.0.1 references in the test files, how does that work?

veloosooo:
This is actually working

BeomJobs:
I want this great solution combined 🥇

Ehesp (approved):
LGTM

chengyuhui:
Is there someting I still need to do?

Ehesp:
@amirh this one works, not sure we can test since it doesn't run in emulator.

Salakar:
@googlebot I consent.

Salakar:
Hey everyone;

I've pushed a few changes to this PR;

 - Added an integration test.
 - Fixed the issue for Web also.
 - Fixed an issue where the `bufferingEnd` event wasn't been sent.

### Local Testing


#### Web - Integration Testing

image

#### Android - Integration Testing

image

Ehesp (approved):

Salakar:
cc @amirh @cyanglaz for Review and merging

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -54,6 +54,8 @@
 
   private boolean isInitialized = false;
 
+  private boolean isBuffering = false;
```
cyanglaz, line 57: I think `isBuffering` should be a state of `EventListener` instead of `VideoPlayer`. Let's move this ivar to the `EventListener`.

cyanglaz:
cc @ditman for additional reviewing on web implementation.

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -176,10 +176,38 @@ class _VideoPlayer {
         isInitialized = true;
         sendInitialized();
       }
+      if (isBuffering) {
+        isBuffering = false;
+        eventController.add(VideoEvent(eventType: VideoEventType.bufferingEnd));
+      }
```
cyanglaz, line 182: Looks like this part is repeated several times, can we refactor this into a function? Maybe something like:
`sendBufferingEndEventIfNeeded()`

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -176,10 +176,38 @@ class _VideoPlayer {
...
+    });
+
+    videoElement.onWaiting.listen((dynamic _) {
+      if (!isBuffering) {
+        isBuffering = true;
+        eventController
+            .add(VideoEvent(eventType: VideoEventType.bufferingStart));
+        sendBufferingUpdate();
```
cyanglaz, line 204: Should this be outside of `if (!isBuffering)`?

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -176,10 +176,38 @@ class _VideoPlayer {
...
+      }
     });
 
     // The error event fires when some form of error occurs while attempting to load or perform the media.
     videoElement.onError.listen((Event _) {
+      isBuffering = false;
```
cyanglaz, line 210: Why do we not need to send the "bufferingEnd" event in this case?

packages/video_player/video_player/example/integration_test/video_player_test.dart
```diff
@@ -29,10 +34,53 @@ void main() {
...
+        await networkController.play();
+        await networkController.seekTo(const Duration(seconds: 5));
+        await tester.pumpAndSettle(_playDuration);
+        await networkController.pause();
+
+        expect(networkController.value.isPlaying, false);
+        expect(networkController.value.position,
+            (Duration position) => position > const Duration(seconds: 0));
+
+        expect(
```
cyanglaz, line 62: Can we use `networkController.addListener()` to test if `networkController.bufferingStart` has been set? We can even check the order of these event types too. 

This way, we don't have to expose `videoPlayerPlatform`. 

Something like:
```dart 
        final Completer<void> started = Completer();
        final Completer<void> ended = Completer();
        bool startedBuffering = false;
        networkController.addListener(() {
          if (networkController.value.isBuffering) {
            startedBuffering = true;
            started.complete();
          }
          if (startedBuffering && !networkController.value.isBuffering) {
            ended.complete();
          }
        });
```

Then we can check the state at desired places, such as
```dart
        await started;
        expect(startedBuffering, true);
```

```dart
        await ended;
        expect(startedBuffering, false);
```

(The above code is not tested)

cyanglaz (commented):
Thanks for the PR! I left some reviews comments!

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.2
```
cyanglaz, line 1: Note: if we ended up needing to expose `videoPlayerPlatform`, a minor version bump is required, so 1.1.0.

cyanglaz (changes requested):

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -131,7 +131,6 @@ class VideoPlayerPlugin extends VideoPlayerPlatform {
 
   @override
   Future<Duration> getPosition(int textureId) async {
-    _videoPlayers[textureId].sendBufferingUpdate();
```
ditman, line 134: Nit: This change makes sense to me, but I'd add a note to the CHANGELOG about it. "The videoplayer won't fire a buffering update event immediately when getting the playhead position" or something like that. This might be breaking for some (even if the buffering is sent immediately afterwards, we may be changing the order in which events are sent to the app). This is probably not a breaking change that's going to affect many people, but in event-driven apps, it might confuse users.

packages/video_player/video_player_web/lib/video_player_web.dart
ditman: This looks like a good opportunity for a isBuffering setter, actually:

```dart
void setBuffering(bool buffering) {
  if (isBuffering != buffering) {
    isBuffering = buffering;
    eventController.add(VideoEvent(eventType: isBuffering ? VideoEventType.bufferingStarted : VideoEventType.bufferingEnd));
  }
}
```

(However there's some special cases like the onError that break the nice pattern :P)

packages/video_player/video_player_web/lib/video_player_web.dart
ditman: I checked, and it seems the android version doesn't send it either, it just fires the error event. So I guess this is good for feature parity.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -17,7 +17,10 @@ export 'package:video_player_platform_interface/video_player_platform_interface.
 import 'src/closed_caption_file.dart';
 export 'src/closed_caption_file.dart';
 
-final VideoPlayerPlatform _videoPlayerPlatform = VideoPlayerPlatform.instance
+/// This is just exposed for testing. It shouldn't be used by anyone depending
+/// on the plugin.
+@visibleForTesting
+final VideoPlayerPlatform videoPlayerPlatform = VideoPlayerPlatform.instance
```
ditman, line 23: Instead of making this visibleForTesting, can we just access `VideoPlayerPlatform.instance` on the tests? It's already public anyway!

ditman (commented):
The web part LGTM, I've only added a few comments/nits. Once @cyanglaz is happy, 🚀

packages/video_player/video_player_web/lib/video_player_web.dart
Salakar: Both good ideas, I chose the `setBuffering` approach in the end that @ditman mentioned (on both Android & Web) - and error events on both these platforms will now also send the buffering end event also (only if was buffering previously) - made sense to send this as otherwise there could be a potential for an invalid buffering state when the player errors.

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+* Fix issue where `isBuffering` is not updating on Android.
```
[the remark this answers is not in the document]
Salakar: Resolved, thanks!

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
Salakar: Makes sense, have moved to inside `EventListener`, thanks

packages/video_player/video_player_web/lib/video_player_web.dart
Salakar: Error events on both these platforms will now also send the buffering end event also (only if was buffering previously) - made sense to send this as otherwise there could be a potential for an invalid buffering state when the player errors.

packages/video_player/video_player_web/lib/video_player_web.dart
Salakar: Some context on why I removed this; setting the player position would result in a buffering update event being sent immediately, this would happen before any buffering start event and didn't really make sense to be like that, especially since internally the event listeners already send buffering update events.

I could add this back in (same on Android) but the events would be `bufferingUpdate`, `bufferingStart`, `bufferingUpdate` (as many times as applicable), `bufferingEnd` - which doesn't seem right to me 🤔 

I'm on the fence on whether I should just revert this change and not worry about those events in the scope of this PR vs having a breaking change.

Thoughts, should I just revert or are we ok with having a breaking change here?

Salakar: Reverted this change for now, could address it in another PR if necessary.

packages/video_player/video_player/lib/video_player.dart
Salakar: Removed this change as no longer required

packages/video_player/video_player/example/integration_test/video_player_test.dart
Salakar: I did do this initially before exposing the platform but had issues, tried it again for sanity and seems to be working now 🙈 updated PR to test using this approach and remove the `videoPlayerPlatform` visibility change.

packages/video_player/video_player/CHANGELOG.md
Salakar: No longer exposing `videoPlayerPlatform`

Salakar:
@cyanglaz @ditman thanks for the review! I've updated based on review points and also resolved merge conflicts again, please could you take a look. Thanks

cyanglaz:
CC @blasten to confirm nullsafety version

cyanglaz (approved):
LGTM!

blasten (approved):
LGTM!

csells:
@cyanglaz thoughts about Mike's question on versioning?

cyanglaz:
@csells blasten is the expert on null-safety so I tagged him for confirmation. He just gave LGTM so I think it's good.

glosama:
Not working for me.
Working fine in iOS, Emulator of Android but not in Android Physical device.

PR #2567, opened 2020-02-28
[device_info] add PackageManager's SystemFeatures to AndroidDeviceInfo

EricEnslen:
## Description

There's a variety of scenarios where checking at runtime which system
features are available is useful.  System features are device
capabilities that do not change at runtime (for example, FEATURE_BLUETOOTH
is always present if the device has a bluetooth radio, even if Bluetooth
is presently disabled), so DeviceInfo seems like the right place to put
this.

## Related Issues

N/A

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/device_info/lib/device_info.dart
+++ b/packages/device_info/lib/device_info.dart
@@ -62,9 +62,11 @@ class AndroidDeviceInfo {
     this.type,
     this.isPhysicalDevice,
     this.androidId,
+    List<String> systemFeatures,
   })  : supported32BitAbis = List<String>.unmodifiable(supported32BitAbis),
         supported64BitAbis = List<String>.unmodifiable(supported64BitAbis),
-        supportedAbis = List<String>.unmodifiable(supportedAbis);
+        supportedAbis = List<String>.unmodifiable(supportedAbis),
+        systemFeatures = List<String>.unmodifiable(systemFeatures);
 
   /// Android operating system version values derived from `android.os.Build.VERSION`.
   final AndroidBuildVersion version;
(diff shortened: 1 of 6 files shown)


packages/device_info/lib/device_info.dart
```diff
@@ -126,6 +128,9 @@ class AndroidDeviceInfo {
   /// The Android hardware device ID that is unique between the device + user and app signing.
   final String androidId;
 
+  /// A list of available features, from PackageManager.getSystemAvailableFeatures.
```
mehmetf: This would require Flutter apps to know Android specific system feature strings. Would it be possible to expose this as an enum? It would be OK for it to be less generic in the name of making API easier to grok.

EricEnslen: I think that would create a maintenance headache, and it would restrict the flexibility of the API (unless we also surfaced the strings that don't match to an enum, which would be a bit redundant).

The Android API that this maps to is inherently based on strings.  Some of the more common feature strings are surfaced as public constants in PackageManager (see https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_ACTIVITIES_ON_SECONDARY_DISPLAYS and below) but that list is both regularly updated, and always incomplete.  For example, there's manufacturer features like "com.google.android.feature.GOOGLE_EXPERIENCE" that would never be added to the list of string constants in PackageManager, but some app developers may care to check for.  For any developer wanting to use this API, it should be a simple effort to figure out which strings they care about, and hardcode them in a simple wrapper around AndroidDeviceInfo.

mehmetf: > The Android API that this maps to is inherently based on strings.

Good to know. In that case, I am OK with this API.

> common feature strings are surfaced as public constants

Can we do the same and expose `const` that the apps can use?

Flutter developers are not typically familiar with Android and how it works. They might not even know what to google. Ideally, everything they need to know is in the Dart API. If that's not the case, we should improve the API.

EricEnslen: Talked a bit offline.  It would be a lot of constants to add (PackageManager has ~140 public FEATURE_ constants), and they're already available through PackageManager's documentation, so we agreed there's not much value in adding the consts in Dart.  Updated the documentation on the dart side to reference PackageManager's documentation where the common strings are available, as well as to caution people against using this API except for in specific circumstances where they have no other option.

lidongze91:
We have a breaking change for this PR. https://sponge2.corp.google.com/1b5a148c-8e14-4265-961d-b8cab9746303/targets/%2F%2Fnbu%2Fpaisa%2Fgpay%2Fapp%2Fauthenticated_service%3Atests_test_device_service_test.host. It happens when the receiver of the function fromList is null. Could you provide proper checks please?

PR #2568, opened 2020-03-02
Fix for connectivity deprecation warning.

collinjackson:
Fixes https://github.com/flutter/flutter/issues/39842


collinjackson:
If we want this to land with a test it seems like the possibilities are

- second example app
- land it as a manual test
- wait for #51781 first

@amirh do you have a preference?

amirh:
I'd rather avoid the complexity of keeping 2 example apps for a single package, especially if we believe that within a few months we'll have both kinds of tests running with a single example app.

I'm ok with the 2 other options.

stuartmorgan:
Closing as we’re no longer accepting non-critical PRs for this plugin, as we hope that in time we’re able to transition users to the corresponding plugin in [the Flutter Community Plus Plugins repository](https://plus.fluttercommunity.dev/).

PR #2571, opened 2020-03-02
[espresso] Remove file that was causing compile errors on iOS

collinjackson:
The EspressoFlutter project has a stub iOS app. There were both Objective-C and Swift files. I've removed the Swift references.

Enables landing test for flutter/flutter#48439


stuartmorgan:
Obsoleted by https://github.com/flutter/plugins/pull/2580; closing.

PR #2576, opened 2020-03-04
[e2e] Updates to support simultaneous E2E and Espresso tests

collinjackson:
Right now we're using a mock implementation of the driver extension protocol. This change switches to using the real one for now. It allows EarlGrey and Espresso tests to talk to an app that is using `E2EWidgetsFlutterBinding`

Fixes https://github.com/flutter/flutter/issues/51781


collinjackson:
Closing this in favor of https://github.com/flutter/plugins/pull/2584, although I'll probably come back to it soon.

PR #2584, opened 2020-03-06
Make FlutterRunner launch deterministic when running other tests.

collinjackson:
Fixes a flake where `FlutterRunner` didn't always launch properly.

Instead we are opting out of the Android test rule's auto-launch and having the runner do it manually.

Also renamed `FlutterRunner` to `FlutterTestRunner` to avoid conflict with Fuchsia.


collinjackson:
Flake is unrelated (https://github.com/flutter/flutter/issues/51074), I'll tackle that next

PR #2585, opened 2020-03-06
[in_app_purchase] Android: fix potential crash when casting in v1 embedding. (Trivial)

cyanglaz:
## Description

Fix a potential crash when we cast context() to Application in the v1 embedding code.

## Related Issues

https://github.com/flutter/flutter/issues/50861

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
+++ b/packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
@@ -50,7 +50,8 @@ public class InAppPurchasePlugin implements FlutterPlugin, ActivityAware {
   public static void registerWith(Registrar registrar) {
     InAppPurchasePlugin plugin = new InAppPurchasePlugin();
     plugin.setupMethodChannel(registrar.activity(), registrar.messenger(), registrar.context());
-    if (registrar.context().getApplicationContext() instanceof Application) {
+    if (registrar.context().getApplicationContext() != null
+        && registrar.context().getApplicationContext() instanceof Application) {
       ((Application) registrar.context().getApplicationContext())
           .registerActivityLifecycleCallbacks(plugin.methodCallHandler);
     }
--- a/packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
+++ b/packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
@@ -50,8 +50,10 @@ public class InAppPurchasePlugin implements FlutterPlugin, ActivityAware {
   public static void registerWith(Registrar registrar) {
     InAppPurchasePlugin plugin = new InAppPurchasePlugin();
     plugin.setupMethodChannel(registrar.activity(), registrar.messenger(), registrar.context());
-    ((Application) registrar.context())
-        .registerActivityLifecycleCallbacks(plugin.methodCallHandler);
+    if (registrar.context().getApplicationContext() instanceof Application) {
+      ((Application) registrar.context().getApplicationContext())
+          .registerActivityLifecycleCallbacks(plugin.methodCallHandler);
+    }
   }
 
   @Override
--- a/packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
+++ b/packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
@@ -50,11 +50,8 @@ public class InAppPurchasePlugin implements FlutterPlugin, ActivityAware {
   public static void registerWith(Registrar registrar) {
     InAppPurchasePlugin plugin = new InAppPurchasePlugin();
     plugin.setupMethodChannel(registrar.activity(), registrar.messenger(), registrar.context());
-    if (registrar.context().getApplicationContext() != null
-        && registrar.context().getApplicationContext() instanceof Application) {
-      ((Application) registrar.context().getApplicationContext())
-          .registerActivityLifecycleCallbacks(plugin.methodCallHandler);
-    }
+    ((Application) registrar.context().getApplicationContext())
+        .registerActivityLifecycleCallbacks(plugin.methodCallHandler);
   }
 
   @Override
--- a/packages/in_app_purchase/example/android/app/src/test/java/io/flutter/plugins/inapppurchase/InAppPurchasePluginTest.java
+++ b/packages/in_app_purchase/example/android/app/src/test/java/io/flutter/plugins/inapppurchase/InAppPurchasePluginTest.java
@@ -1,3 +1,7 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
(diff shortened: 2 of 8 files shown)


packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
```diff
@@ -50,8 +50,11 @@
   public static void registerWith(Registrar registrar) {
     InAppPurchasePlugin plugin = new InAppPurchasePlugin();
     plugin.setupMethodChannel(registrar.activity(), registrar.messenger(), registrar.context());
-    ((Application) registrar.context())
-        .registerActivityLifecycleCallbacks(plugin.methodCallHandler);
+    if (registrar.context().getApplicationContext() != null
+        && registrar.context().getApplicationContext() instanceof Application) {
```
jmagman: Is it bad if it's null or not an Application?  Does that need to be handled?

cyanglaz: if either of that situation happens, we are not able to get the application, which makes us no way to register life cycle callbacks. We kind of just bailing here.
I'm not sure if there would be a way to handle that situation in V1 embedding, but we definitely don't want to crash here.

jmagman: Do we know how people are getting in this state?

cyanglaz: I am not entirely sure. @mklim or @matthew-carroll might have more ideas on this.

matthew-carroll: I believe that you can just say `registrar.context().getApplicationContext()`. I don't think that `context()` can return null, and I don't think that `getApplicationContext()` can return null

cyanglaz: @matthew-carroll Thanks, will getApplicationContext always be able to cast to Application?

matthew-carroll: I believe so. As long as the `context` on which you're invoking `getApplicationContext()` is non-null, I think the line will always be fine.

matthew-carroll: BTW, do you know why you need an `Application`? What are the `Application` methods that are being called that are not defined on a `Context`?

cyanglaz: ok. I just did some search and it seems to be the case. I'll update the fix.

cyanglaz: I believe `.registerActivityLifecycleCallbacks` is only in `Application`?

matthew-carroll: That looks right.

packages/in_app_purchase/example/android/app/src/test/java/io/flutter/plugins/inapppurchase/InAppPurchasePluginTest.java
```diff
@@ -0,0 +1,39 @@
...
+  @Before
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
+    when(mockRegistrar.activity()).thenReturn(activity);
+    when(mockRegistrar.messenger()).thenReturn(mockMessenger);
+    when(mockRegistrar.context()).thenReturn(context);
+  }
+
+  @Test
+  public void registerWith_doNotCrashWhenApplicationContextIsTypeApplication() {
```
matthew-carroll: Is this test necessary? We should probably avoid testing situations that can't happen, because it might indicate to readers that it can happen...

packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
mklim: To explain this a little more, there is a difference between an `ApplicationContext` (can be cast to `Application`) and an `ActivityContext` (can be cast to `Activity`) even though they're both `Context`s. Generally it's hard to tell the difference but occasionally it's critical. The Activity Context is tied to the Activity lifecycle so can be invalidated much easier.

packages/in_app_purchase/example/android/app/src/test/java/io/flutter/plugins/inapppurchase/InAppPurchasePluginTest.java
mklim: Isn't `context.getApplicationContext()` always going to return an application? This looks like the standard case to me.

packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
```diff
@@ -50,7 +50,7 @@
   public static void registerWith(Registrar registrar) {
     InAppPurchasePlugin plugin = new InAppPurchasePlugin();
     plugin.setupMethodChannel(registrar.activity(), registrar.messenger(), registrar.context());
-    ((Application) registrar.context())
```
mklim: Just for a little more info in case this doesn't make sense, the bug here is that sometimes `registrar.context()` will return an Activity context instead of an Application context if it's available and this case will fail. My understanding is that it's the typical for it to do this so I'm actually surprised that this was routinely working before.

packages/in_app_purchase/example/android/app/src/test/java/io/flutter/plugins/inapppurchase/InAppPurchasePluginTest.java
mklim: Though on a second look, maybe a more meaningful test here would be to return an `Activity` from `registrar.context()` and make sure that that doesn't crash?

packages/in_app_purchase/android/src/main/java/io/flutter/plugins/inapppurchase/InAppPurchasePlugin.java
cyanglaz: this was newly added in my new patch so it might be that my patch broke it. However, when I tested locally, it was working. I wonder why activity was not available when i was testing it.

packages/in_app_purchase/example/android/app/src/test/java/io/flutter/plugins/inapppurchase/InAppPurchasePluginTest.java
cyanglaz: @mklim Sounds good, I will update the test.

cyanglaz:
Updated the tests based on discussion. Will merge on green.

cyanglaz:
Landing on tree red as it is going to fix a ci error.

PR #2586, opened 2020-03-07
[android_intent] Bump to Flutter stable, remove deprecation warnings

ened:
## Description

Bumps dart & flutter version & adds a little bit of house keeping so that the next PR is distraction free.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/android_intent/pubspec.yaml
```diff
@@ -23,5 +23,5 @@ dev_dependencies:
   pedantic: ^1.8.0
 
 environment:
-  sdk: ">=2.0.0-dev.28.0 <3.0.0"
-  flutter: ">=1.10.0 <2.0.0"
+  sdk: ">=2.3.0 <3.0.0"
+  flutter: ">=1.12.13+hotfix.5 <2.0.0"
```
matthew-carroll: Is there a reason that we're bumping the min supported Dart and Flutter versions?

ened: The Dart version was increased to make the spread operator available.

Flutter version was increased to the latest stable version to be inline with other plugins like `google_maps`, `connectivity` etc.

packages/android_intent/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: android_intent
 description: Flutter plugin for launching Android Intents. Not supported on iOS.
 homepage: https://github.com/flutter/plugins/tree/master/packages/android_intent
-version: 0.3.6
+version: 0.3.7
```
mklim: nit: I think this should actually be a patch bump, semver considers transitive dependency upgrades a small change since they don't upgrade the public API.

https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

ened: OK

packages/android_intent/pubspec.yaml
matthew-carroll: Should we be forcing all consumers of this plugin to upgrade their Dart version just so that this plugin can use the spread operator? I'm not sure how many people are on various versions of Dart, but my initial thought is that using the spread operator is a pretty trivial detail and we probably shouldn't force a language change on people just do that. Thoughts?

ened: Glad you asked. I had similar concerns but ended up with:
- [Current stable](https://duckduckgo.com/?q=flutter+1.12+stable&t=osx) is referring to Dart 2.7, while this PR only bumps to 2.3 (when spread was introduced, roughly)
  - Every time I bump a Flutter version on my machine, the bundled Dart version is synced
  - The [previous stable](https://developers.googleblog.com/2019/09/flutter-news-from-gdd-china-flutter1.9.html) version (1.9) shipped with Dart 2.5, so I think we're ok
- IMO Flutters official plugins should showcase new language (stable) features so that users adopt them quicker, thus making the the platform more attractive
  - Flutters style guide mentions how to use the spread operator, so I assumed it's already accepted
- Reducing LOC is always a compelling argument although not too strong one here

mklim: I think current stable including a higher version of Dart already makes this a non-issue.

matthew-carroll: As @mklim mentioned, your first bullet point is probably the right answer. If stable has bumped Dart then we're probably good.

I would caution the 2nd and 3rd bullet points. I actually don't think using new language features or showcasing capabilities would be seen as an appropriate reason to break versions. I'll defer to @amirh and @Hixie in that regard, but that would be my guess.

In any case, bullet point 1 makes sense and works for me.

ened:
@mklim @matthew-carroll thank you for reviewing this PR. How could the updated packaged be pushed to pub?

PR #674, opened 2018-07-18
[image_picker]Add argument maxDuration to video record

vagrantrobbie:
Can pass a double of seconds to maxDuration to limit the duration of the video record

PR #2589, opened 2020-03-08
[path_provider] path provider linux implementation in dart

TimWhiting:
## Description

This PR contributes a Linux implementation of the path provider plugin. It is a work in progress, since it seems that pure dart implementations aren't automatically detected / registered. (Flutter generates a makefile that expects a C++ class / header file).

## Related Issues

flutter/flutter#41716 would be resolved by this PR
google/flutter-desktop-embedding PR #674 contains discussion related / leading to this implementation
flutter/packages PR #117 contains work leading up to this

## Checklist

- [] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].

## Breaking Change


TimWhiting:
@stuartmorgan 

Here is the dart implementation of path_provider for Linux we discussed. I'm having trouble figuring out how to get the example app to run. It seems that currently there is a lack of support for automatically registering pure dart plugins for platforms other than web. Can you confirm or point me in the right direction? Currently the Makefile expects there to be a C++ header file by the name of path_provider.

TimWhiting:
I took a look at the place where the plugin registry is built:
(here)[https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/plugins.dart]
as well as the web plugin registrar: (here)[https://github.com/flutter/flutter/blob/master/packages/flutter_web_plugins/lib/src/plugin_registry.dart]

My proposal would be to:
1. Rename the `flutter_web_plugins` package to `flutter_dart_plugins`, 
2. Have a `language` key/value in the yaml map in the pubspec that specifies if you are using dart to implement the plugin. (Defaults to dart for web). Then use that to generate the dart plugin registrar. 
3. Provide some documentation that you can use dart to implement a plugin on desktop, just like you can on web, but you have to specify dart for the language parameter in the pubspec

By adding this language parameter this we could eventually allow using ffi (through dart parameter in pubspec), C# for windows, etc...

stuartmorgan:
> It seems that currently there is a lack of support for automatically registering pure dart plugins for platforms other than web. Can you confirm or point me in the right direction?

Yes, unfortunately that's the case; a couple of us were looking at that a bit at the end of last week and discovered that the planned solution hasn't actually been implemented yet.

> 1. Rename the `flutter_web_plugins` package to `flutter_dart_plugins`,

We shouldn't need that; the goal is to implement the platform interface, bypassing platform messages entirely. That's one of the things the new platform interface abstraction was designed to address.

> 2. Have a `language` key/value in the yaml map in the pubspec that specifies if you are using dart to implement the plugin. (Defaults to dart for web). Then use that to generate the dart plugin registrar.

The existing plan for allowing Dart registration on any platform is described [here](https://docs.google.com/document/d/1LD7QjmzJZLCopUrFAAE98wOUQpjmguyGTN2wd_89Srs/edit#heading=h.xox07196p8c9). There's no plan to add something as generic as "language" for all plugin platforms as the potential for confusion is high.

The complexity is not in generating the Dart plugin registrar, but in finding a way to invoke it in a way that doesn't have problematic edge cases. It's something that will need more investigation and discussion in the coming weeks.

> By adding this language parameter this we could eventually allow using ffi (through dart parameter in pubspec), C# for windows, etc...

In general, we want to be extremely thoughtful about when/how we add native language options for plugins; allowing lots of choice at that layer is not necessarily a feature.

TimWhiting:
@stuartmorgan Thanks for pointing me to that document. I haven't forgotten this plugin. 

Is there an issue I can follow to know when the flutter_tools support of dart plugins for non-web clients is available?

stuartmorgan:
https://github.com/flutter/flutter/issues/52267 is tracking dartPluginClass plumbing. However, based on recent discussions I think while we're waiting for that we can do manual registration in the front-end plugin. So we would have path_provider endorse path_provider_linux, and then for `Platform.isLinux` path_provider would set `PathProviderPlatform.instance` to the Linux Dart implementation.

stuartmorgan:
@TimWhiting I'm going to be working on setting up the testing infrastructure for Linux (which is a blocker for landing Linux plugins) next week; if you're interested in updating this PR we can use it to continue blazing the trail of landing an FFI-based Linux plugin.

I think it's pretty close, the main things are to add the registration to the front-end package (to work around the lack of support for automatic registration), and removing the `linux/` directories from the example apps. See https://github.com/flutter/plugins/pull/2631, which is the first attempt to do this same thing for Windows, for an example and relevant discussion.

TimWhiting:
Okay, I'll look into it tomorrow, thanks!

stuartmorgan:
> get rid of android and ios plugin stubs

The podspec stub is still required for iOS. I believe some kind of Android stub is still required on stable as well for the moment.

TimWhiting:
> > get rid of android and ios plugin stubs
> 
> The podspec stub is still required for iOS. I believe some kind of Android stub is still required on stable as well for the moment.

Okay, I'll revert that last commit

stuartmorgan:
On the testing infrastructure front, the `plugin_tools` support is in, and once https://github.com/flutter/plugins/pull/2717 lands we should be able to merge master into this PR and see it run e2e tests on this plugin 🤞

packages/path_provider/path_provider/pubspec.yaml
```diff
@@ -14,13 +14,17 @@ flutter:
...
+      linux:
+        default_package: path_provider_linux
 
 dependencies:
   flutter:
     sdk: flutter
   path_provider_platform_interface: ^1.0.1
   path_provider_macos: ^0.0.4
+  path_provider_linux:
+    path: ../path_provider_linux
```
stuartmorgan, line 26: This should not be using a path-based dependency. I think you'll need to publish path_provider_linux first, then endorse it in a follow-up.

packages/path_provider/path_provider_linux/ios/Classes/PathProviderLinuxPlugin.h
```diff
@@ -0,0 +1,4 @@
+#import <Flutter/Flutter.h>
```
stuartmorgan, line 1: The iOS stub should only be the podspec, none of the code.

packages/path_provider/path_provider_linux/example/linux/Makefile
```diff
@@ -0,0 +1,146 @@
+include app_configuration.mk
```
stuartmorgan, line 1: The `example/linux/` folder must not be checked in; it will break. The test driver will create it on the fly.

packages/path_provider/path_provider/pubspec.yaml
TimWhiting: Yeah I know, I was just following the windows example link you sent me. I'll go ahead and revert the changes to the path_provider, and get the linux plugin in first in this pull request before endorsing it in another.

TimWhiting:
Looks like there is an issue with the Linux e2e test, is there an environment variable I need to set? Here is the relevant part of the error output:
```
Starting application: test_driver/path_provider_e2e.dart
Building Linux application...                                   
GLFW error 65544: X11: The DISPLAY environment variable is missing
Could not create window; FlutterDesktopInit failed.
```

Also, as far as the android stub. I'm not sure what I did wrong, it seems to think the project is in the wrong format (i.e. pre-2017 flutter tooling). However, I'm on flutter master when I do `flutter create`, and the same tests fail when I copy the android folder from path_provider_macos, which should be the same no-op implementation that we want here.

stuartmorgan:
> Looks like there is an issue with the Linux e2e test, is there an environment variable I need to set? Here is the relevant part of the error output:

It turns out my attempted test run before landing it only covered build, not build+drive, to a mistake on my part. I'll need to figure out how to handle the lack of X server; hopefully I can get that figured out today.

stuartmorgan:
After a long detour in what turned out to be the wrong direction, https://github.com/flutter/plugins/pull/2750 should fix the flutter-driver issue. With that PR I was able to run driver on this branch and only missing plugin exceptions (presumably something didn't get fully unwound from removing the endorsement?), no graphics errors.

stuartmorgan:
@TimWhiting The Android and iOS errors look like they are from trying to build the `path_provider_linux/example`, which doesn't have ios/ and android/ directories. I expect re-adding the template-created `example/ios` and `example/android` directories would fix that. (Longer term `plugin_tools` will need reworking to not expect iOS and Android to always be there).

apgapg:
@TimWhiting I am really forward to land this PR in flutter repo as my development would become fast. Currently my ubuntu pc doesnt support running emulator. Just because of path_provider plugin i need to use my android phone for debugging. 
This PR will definitely help me to run linux desktop thereby making my development fast

TimWhiting:
@apgapg 
I've been caught up with other things at the moment. I've tested the plugin and it works great locally, I just need to get the android and ios tests to pass before it can be merged in and officially endorsed, and I'm meeting more problems there than anticipated. I'll try to fix the issues over the weekend, but I can't guarantee anything, as I have other commitments.
Until then you can use my fork and branch in your pubspec.yaml.

In your dependencies section (something like this, I'm not sure if I got the syntax 100%):
```yaml
   path_provider: **choose latest version**
   path_provider_linux:
     git:
       url: https://github.com/TimWhiting/plugins
       ref: path_provider_linux
       path: packages/path_provider/path_provider_linux
```

TimWhiting:
@stuartmorgan 
The only issues with the apks right now seems to be on the stable channel since linux plugins aren't supported on stable. Here is the error:
```
Invalid plugin specification path_provider_linux.
Invalid "linux" plugin specification.
```

As far as the ipas are concerned. I tried both creating an ios folder from the tool (`flutter create .`), as well as copying the `path_provider_macos/example` ios folder, and am still seeing failure to build because of ios signing, and build / version number missing in pubspec.yaml (required for publishing the app to the ios App Store). But `path_provider_macos/example`, as well as various other plugin examples have neither of those fields either, is that a recent requirement for examples?

Is this something that should be blocking the merge of this pull request, since users are unlikely to be trying to build/publish the ios apps in the linux example plugin? I'm unsure what to do from here.

Other than that it looks like most of the builds are succeeding and the only files I touched are in the `path_provider_linux` directory.

stuartmorgan:
> The only issues with the apks right now seems to be on the stable channel since linux plugins aren't supported on stable. Here is the error:
> 
> ```
> Invalid plugin specification path_provider_linux.
> Invalid "linux" plugin specification.
> ```

`linux` has been in the schema for a long time so it should be supported on stable. I believe the actual problem is that having it only contain `dartPluginClass` is invalid on stable since that's a much newer change.

I didn't realize that would be a fatal error at plugin parse time across platforms. If that is indeed what's going on, it'll block endorsing any Dart-only desktop plugins until that change reaches stable, which would be unfortunate. I'll need to investigate further next week.

> and am still seeing failure to build because of ios signing, and build / version number missing in pubspec.yaml (required for publishing the app to the ios App Store).

I think you're attributing the build failures to non-fatal warning messages. Earlier in the output I see a fatal error about missing xcconfig files.

> Is this something that should be blocking the merge of this pull request, since users are unlikely to be trying to build/publish the ios apps in the linux example plugin?

All test failures block merging of pull requests; in cases where the test isn't testing what we want, the solution is to fix the testing, not have the CI be red forever.


I have a partial PR to fix the fact that `plugin_tools` requires iOS and Android examples. I'll try to finish that next week, which I think will be enough to unblock landing this. Endorsing it will still be blocked until/unless I can figure out a solution to the plugin platform schema issue.

TimWhiting:
@stuartmorgan 

Yeah that makes sense. I'll figure out the xcconfig issue hopefully beginning of next week so that the only blocker would be the dartPluginClass. If the dartPluginClass turns out to be the only blocker and can't be merged into flutter stable in the near future would it be possible to change the testing to only test desktop plugin examples on master? 

I'm new to CI testing so thanks for clarifying that we can't merge until all tests are green

stuartmorgan:
> If the dartPluginClass turns out to be the only blocker and can't be merged into flutter stable in the near future would it be possible to change the testing to only test desktop plugin examples on master?

Desktop examples are only tested on master; the issue is that all Android and iOS examples are tested on stable. The PR I mentioned above will solve that by not having those examples (pointlessly) include Android and iOS.

But the endorsement part isn't just a testing issue. If stable chokes on that pubspec, then adding an endorsement from `path_provider` to `path_provider_linux` would break every single Flutter application build on stable, for every platform, which is clearly not viable.

apgapg:
@TimWhiting 
I tried the same thing earlier but it seems it tries to find platform channel code. The path_provider_linux directory contains code in lib and not in linux/ as the package uses xdg lib.
```
flutter: ⛔ MissingPluginException(No implementation found for method getApplicationSupportDirectory on channel plugins.flutter.io/path_provider)
flutter: #0   MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:154:7)
flutter: <asynchronous suspension>
flutter: #1   MethodChannel.invokeMethod (package:flutter/src/services/platform_channel.dart:329:12)
flutter: #2   MethodChannelPathProvider.getApplicationSupportPath (package:path_provider_platform_interface/src/method_channel_path_provider.dart:38:26)
flutter: #3   getApplicationSupportDirectory (package:path_provider/path_provider.dart:45:39)
```
I might be going off topic here. Certainly i can wait for the PR to merge.

TimWhiting:
@apgapg 
Yes, you will also have to specifically set the implementation of the Linux plugin when you are running on Linux, (at least until this gets endorsed). 

Add the following lines to the beginning of your main function
```dart
// TODO: Remove the following four lines once path provider endorses the linux plugin
  if (Platform.isLinux) {
    await WidgetsFlutterBinding.ensureInitialized();
    PathProviderPlatform.instance = PathProviderLinux();
  }
```

stuartmorgan:
I just landed https://github.com/flutter/flutter/pull/57498, which should allow working around the "Invalid "linux" plugin specification" error by adding `pluginClass: none` to the `linux:` section of the Linux plugin's `pubspec.yaml`. That will pass validation on stable, but be ignored on the latest master when actually building the Linux plugin.

TimWhiting:
@stuartmorgan 
Looks like the new CMake build system for Linux doesn't have a minimum CMake version set, and the CI doesn't have a compatible version? The TRANSFORM command that it is failing on looks like it was introduced in CMake 3.12.

stuartmorgan:
> @stuartmorgan
> Looks like the new CMake build system for Linux doesn't have a minimum CMake version set, and the CI doesn't have a compatible version? The TRANSFORM command that it is failing on looks like it was introduced in CMake 3.12.

Sigh. Unfortunately setting a minimum version doesn't do any auditing/enforcement, so while in theory the minimum version is 3.10, in practice it doesn't warn if we use something newer. I tried to be careful about it while writing it, but apparently missed some things, since I didn't have a setup to actually run it under 3.10.

I need to get that set up and fully audited it the next day or two.

stuartmorgan:
https://github.com/flutter/flutter/pull/57515 will fix it. Tested that generation works with a manually installed 3.10 this time, so hopefully that will be the last of the CMake issues.

stuartmorgan:
FYI the linux build is currently broken because I accidentally broke the template, and the Flutter tree has been closed so I haven't been able to submit [the fix](https://github.com/flutter/flutter/pull/57538) yet. I'll trigger a re-run of build-linux+drive-examples after I'm able to land that.

stuartmorgan:
The lint errors look like there is a tighter set of requirements being enforced on new/changed packages. https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter/ios/google_maps_flutter.podspec looks like one that's recent enough to address the things flagged by the linter.

stuartmorgan:
> ```
> CMake Error at /usr/share/cmake-3.10/Modules/BundleUtilities.cmake:966 (message):
>   error: fixup_bundle: not a valid bundle
> ```

I'm not sure what's going on there; my local 3.10 test didn't have that issue. I'll need to try to figure out how to get verbose build output enabled for a bot run (which will be tricky since the tools that drive the build are from a published package).

stuartmorgan:
> I'm not sure what's going on there

https://github.com/flutter/plugins/pull/2774

stuartmorgan:
Fix landed; if you merge in master, I think there's a good chance this will finally be green across the board! 🤞

TimWhiting:
@stuartmorgan 

It looks like a lot of plugins are now removing the no-op android folders. Should I do the same here?
https://github.com/flutter/flutter/issues/46304

stuartmorgan:
> It looks like a lot of plugins are now removing the no-op android folders. Should I do the same here?]

Yes, I hadn't followed which release that bug was waiting on, but apparently it was 1.17. No need to land the Android no-op if it's no longer needed.

TimWhiting:
@stuartmorgan 
Is this ready to merge now?

stuartmorgan:
I was planning on landing and publishing it this morning. I was just doing a final once-over and noticed that `path_provider_linux/README.md` is still just the placeholder. Can you update it with an actual README? See `path_provider_macos/README.md` for an example of a federated plugin readme.

TimWhiting:
@stuartmorgan 
Sorry about that, I thought I had done that earlier. Just made the update to the `README.md` and merged in the latest master.

stuartmorgan:
The readme is somewhat misleading since there is a step other than adding it to your pubspec, but since we should be doing the follow-up to endorse it shortly I'll land it as-is.

stuartmorgan:
https://pub.dev/packages/path_provider_linux 🎉 

Congratulations on getting the first Linux flutter/plugins plugin, and the first pure-Dart desktop plugin, landed and published 🙂  Thanks again for bearing with all the difficulties of blazing this trail.

You should be able to do the endorsement PR now, which I expect to go much more smoothly.

apgapg:
@TimWhiting In unit test(s) `getApplicationDocumentsDirectory` generally fails due to `MissingPuginException` but when running `flutter test` in Linux environment, the test doesn't fail, because in linux path_provider doesnt call method channel instead use xdg lib directly.

Am i clear in my understanding?

TimWhiting:
@apgapg 
I guess I'm not understanding the question.
At one point you say that it fails in unit tests, and then you say that it doesn't fail in `flutter test`. Shouldn't you be using `flutter test` for all of your unit tests? What exactly is failing & what environment / setup, do you have a simple reproducible example?

You are correct that `path_provider_linux does` not use a method channel.
However, now you should be able to use `path_provider` regularly, and not import the linux package separately. The plugin has been endorsed. If you want to see how to set your own `PathProviderPlatform` for overriding during tests, see `path_provider` ^1.6.10's [documentation](https://pub.dev/documentation/path_provider/latest/path_provider/path_provider-library.html) about this new setter: `disablePathProviderPlatformOverride`. (It will disable the registering of `path_provider_linux`, and use the method channel instead. (Which you can then mock like your example). However, the preferred way of mocking federated plugins like this one is to register a MockPathProviderPlatform that conforms to the PathProviderPlatform interface. See the documentation in [this package}(https://pub.dev/packages/path_provider_platform_interface)

Or see the discussion on this pull request about `disablePathProviderPlatformOverride`:
https://github.com/flutter/plugins/pull/2789

apgapg:
@TimWhiting for first question:
I was trying `flutter test` on my mac air. Here all test related to path_provider failed as it was throwing MissingPlatformException.

But when i run the same command in my CI/CD on github action where the 
```  build:
    runs-on: ubuntu-latest```
test run on ubuntu, they all passed.

So the flutter test in my case is behaving differently on different platforms.

For second para:
Yes i am now using path_provider only. Thanks for clarification

hpoul:
@apgapg i'm not sure you are testing what you think you are testing.. but you probably want to just mock plugins in **unit** tests anyway, mocking path_provider method channel is actually an example on the flutter test page: https://flutter.dev/docs/cookbook/persistence/reading-writing-files#testing

TimWhiting:
@apgapg 
Was `flutter test` on your mac working before? I'm guessing it was.

The macos plugin uses method channels so it needs to be mocked in the same way as your link you wrote about earlier. (Unless you mock the `PathProviderPlatform`, like I mentioned). 

What is it that you are trying to test? Tests for `path_provider` pass, so you shouldn't be testing `path_provider`.
You should probably be mocking `PathProviderPlatform` to get a consistent result across platforms.

apgapg:
@TimWhiting  @hpoul Yes its my ignorance about testing code. I didn't mock path_provider and just messing around with nothing. 

Thanks both of you for clarifying and helping me to learn it in more better way.
Thanks again

PR #2591, opened 2020-03-09
 [url_launcher] Fixed when there are line break (CRLF,LF) in the urlString

wasabeef:
## Description

If the URL contains a line break (LF, CRLF), the iOS app will fail with canLaunch.
And if there is a line break, there is not working.

## Related Issues

- No issues

## Checklist

Before you create this PR confirms that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/url_launcher/url_launcher/lib/url_launcher.dart
+++ b/packages/url_launcher/url_launcher/lib/url_launcher.dart
@@ -84,7 +84,7 @@ Future<bool> launch(
         : SystemUiOverlayStyle.light);
   }
   final bool result = await UrlLauncherPlatform.instance.launch(
-    urlString,
+    url.toString(),
     useSafariVC: forceSafariVC ?? isWebURL,
     useWebView: forceWebView ?? false,
     enableJavaScript: enableJavaScript ?? false,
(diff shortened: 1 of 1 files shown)


tp:
@wasabeef We also encountered a crasher right now, because the native code would crash if the URL starts with a whitespace.

If you stick with trimming in the library, your code should be symmetric though and use `trimLeft` (or better just `trim`) on both `canLaunch` and `launch`.

I would suggest though to rework `canLaunch` return `false` on invalid inputs (possibly on all platforms), so the library doesn't "fix" invalid inputs implicitly.

An exception on the Flutter side every app should easily be able to handle, as opposed to the current crasher on native.

stuartmorgan:
Thanks for the contribution!

However, the API expects a valid URL string; having line breaks in a URL isn't valid. Going down the road of trying to fix up a small subset of invalid URLs doesn't seem like the right approach, so this is not something we will move forward with. The more robust solution is to move the plugin to a `Url`-based API: https://github.com/flutter/flutter/issues/79043

PR #2597, opened 2020-03-10
[battery] Announce API stability and compatibility with 1.0.0

amirh:
## Description

Announce that the package is going to be compatible with 1.0.0 and encourage users to set their constraints accordingly.

@collinjackson @csells after we're happy with the wording here I'll apply it to more plugins we want to upgrade to 1.0.0


csells:
lgtm

PR #2598, opened 2020-03-11
[android_intent] Adds canResolveActivity method

ened:
## Description

As outlined in https://github.com/flutter/flutter/issues/51942, I needed a method to check whether a particular activity is reachable on a target device.

I have marked the PR WIP for the following reasons:
- `canResolve` currently only checks for Activities, using this method [resolveActivity(android.content.Intent,%20int)](https://developer.android.com/reference/android/content/pm/PackageManager#resolveActivity(android.content.Intent, int)) . It also sets a default flag here.
  - there is also `resolveService` & `resolveContentProvider` which can be helpful respectively. I don't see the need for them just yet
  - the passed `flag` is fixed to `PackageManager.MATCHDEFAULTONLY`, which somewhat makes sense for activities. It could be copied to the dart side so that developers can specify it
  - the `resolve*` methods return information on, well, how to resolve a component. We can't do much with it on flutter side, so I would assume we don't need to pass that up
- the name `canResolve` could be replaced with `canLaunch` (to follow `url_launcher`).

## Related Issues

Fixes flutter/flutter#51942.

## Checklist

- [ ] My PR includes unit or integration tests for all changed/updated/fixed behaviors (See [Contributor Guide]).
- [] All existing and new tests are passing.
- [x] I signed the [CLA].

## Breaking Change

- [x] No, this is not a breaking change.


packages/android_intent/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.4.0
```
mklim: nit: At this level pub would consider this a breaking change, should this be `0.3.7`? When plugins are pre 1.0.0 the schema is `0.major.minor+patch`.

packages/android_intent/example/test_driver/android_intent_e2e.dart
```diff
@@ -49,4 +49,13 @@ void main() {
           e.message.contains('No Activity found to handle Intent');
     }));
   }, skip: !Platform.isAndroid);
+
+  testWidgets('#canResolve throws when no Activity is found',
```
mklim: Could you add a success test here? Maybe you could check for the main activity of the example app? It should be installed every e2e run.

packages/android_intent/lib/android_intent.dart
```diff
@@ -138,6 +138,17 @@ class AndroidIntent {
     await _channel.invokeMethod<void>('launch', _buildArguments());
   }
 
+  /// Check whether the intent can be resolved to an activity.
+  ///
+  /// This works only on Android platforms.
+  Future<bool> canResolve() async {
```
mklim: nit: Should this be `canResolveActivity`? Especially since there are other resolve types. I agree that we don't need to add them now, but I think it would make sense to name this just based off the specific method it's adding today instead of being more general.

ened: Yes, and the next change, if anyone ever needs it, could be `canResolveService` or `canResolveContentProvider`. But not for now. :)

ened:
@mklim thank you for reviewing. Have added a few more tests & addressed your feedback. Do you know why the "submit-queue" check failed?

ened:
@mklim still happy after the extended unit test? If so I shall merge soon. :)

ened:
@mklim could you please help to get this new version (0.3.7) uploaded to pub ?

mklim:
@ened sorry about the delay, done.

PR #2618, opened 2020-03-23
[package_info] add support for macos to package_info plugin

hpoul:
## Description

1. adds macos support to package_info plugin. (similar to [firebase_core](https://github.com/FirebaseExtended/flutterfire/tree/master/packages/firebase_core/firebase_core)) moves the objective c implementation file to `darwin/Classes` and symlinks to `ios/Classes` and `macos/Classes`.
2. add macos example (basically just a `flutter create`) - i've also added an explicit `version` code to `pubspec.yaml` of the example app.

## Related Issues

https://github.com/flutter/flutter/issues/41727

## Checklist

- [x] I signed the [CLA].

## Breaking Change


stuartmorgan:
I believe the failures here are that you've added a `version` to the example, and the package version check script is trying to compare your new version to the (non-existent) old version. You'll need to either revert that change to the example's pubspec, or you or someone else will need to fix [the script](https://github.com/flutter/plugin_tools/blob/master/lib/src/version_check_command.dart) (filtering out example/ directories would probably be the best option)

hpoul:
@stuartmorgan thanks for taking a look. I have removed the version and updated the tests.

The `version` property actually got me a bit confused.. because on iOS it is a warning if omitted:

```
Warning: Missing build name (CFBundleShortVersionString).
Warning: Missing build number (CFBundleVersion).
Action Required: You must set a build name and number in the pubspec.yaml file version field before submitting to the App Store.
```

and even though the [pubspec.yaml documentation](https://dart.dev/tools/pub/pubspec#version) says "If you omit it, your package is implicitly versioned 0.0.0." - but it seems in the app templates for android it falls back to `1.0`. 🤔️

anyway, it seems the package_info/example/ios is not even set up to use the flutter version strings (the iOS Info.plist file specifies it directly instead of referencing FLUTTER_BUILD_NAME/FLUTTER_BUILD_NUMBER) so my test updates weren't even correct 🤦️

stuartmorgan:
I suspect this example app's creation predates both the warning, and the iOS version wire-up. If you feel inspired to re-create the example with a current template as a separate PR, that probably wouldn't be a bad idea (assuming the version check script is fixed as well), but it definitely doesn't need to block this.

packages/package_info/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.4.1
+
+* Add support for macos.
```
stuartmorgan: Nit: macOS

packages/package_info/example/test_driver/package_info_e2e.dart
```diff
@@ -24,6 +24,11 @@ void main() {
       expect(info.buildNumber, '1');
       expect(info.packageName, 'io.flutter.plugins.packageInfoExample');
       expect(info.version, '1.0');
+    } else if (Platform.isMacOS) {
```
stuartmorgan: Given that these are all identical, instead of adding another copy here and below, how about collapsing these to just `if (Platform.isAndroid || Platform.isIOS || Platform.isMacOS)`.

hpoul: Well, it's not really the identical, iOS uses version name "1.0" because this is defined in the `Info.plist` and macOS uses "1.0.0" because it uses the generated `FLUTTER_BUILD_NAME` [as that's the default value](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/ios/xcodeproj.dart#L204) if i understand it correctly, for `Flutter/ephemeral/Flutter-Generated.xcconfig`

also android uses lower case package name `packageinfoexample` vs. ios (and now macos) `packageInfoExample` - I guess it wouldn't be too hard to make it identical though, if that's better?

stuartmorgan: Ah, I missed that. Sounds like this is just something that will need to be revisited if/wher the example version is cleaned up.

packages/package_info/CHANGELOG.md
hpoul: 👍️changed to macOS

packages/package_info/example/macos/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Flutter-related
```
cyanglaz, line 1: Do we need this file?

packages/package_info/example/macos/Runner/AppDelegate.swift
```diff
@@ -0,0 +1,9 @@
+import Cocoa
```
cyanglaz, line 1: Are we defaulting to swift for macos?

packages/package_info/example/macos/.gitignore
stuartmorgan: It's part of the template. Having a monolithic .gitignore for the whole project doesn't scale well to N (optional) platforms, so all platform-specific ignores now live in their respective platform directories.

packages/package_info/example/macos/Runner/AppDelegate.swift
stuartmorgan: It's the only template for macOS.

packages/package_info/example/macos/.gitignore
cyanglaz: SG thx

packages/package_info/example/macos/Runner/AppDelegate.swift
cyanglaz: SG

stuartmorgan:
@cyanglaz Any idea what's going on with the test failures here? Was there an issue with running tests against stable recently?

cyanglaz:
@stuartmorgan Not sure, im triggering a re-run and let's see what happens

hpoul:
@stuartmorgan @cyanglaz any chance to take another look at this? no idea what the test failures are all about.. thanks..

stuartmorgan:
@hpoul Sorry, getting back to this was buried in my todo list. Can you try reverting the `ios/` changes in this PR and see if that fixes the issue? I know that was tool-generated rather than something you did manually, but they are the most obvious candidate for having broken iOS tests.

hpoul:
@stuartmorgan thanks, looks like you were right.. :) seems like the master channel made the xcodeproj incompatible with stable.. sounds a bit like https://github.com/flutter/flutter/pull/51453 .. no idea why i actually committed that file

stuartmorgan:
@cyanglaz PTAL; you said LGTM above but didn't actually approve, so I'm not sure if you wanted to take a final look once everything was green.

stuartmorgan:
@cyanglaz Ping again on this (easy) review.

stuartmorgan:
@hpoul If you can sync with master and resolve the conflicts, I can go ahead and land this :)

hpoul:
@stuartmorgan 👍️ merged with master, let's hope the tests are still happy 🤞️

stuartmorgan:
And they are! Landing, thanks for doing this!

hpoul:
Any chance to publish this update to pub? :-) https://pub.dev/packages/package_info is still 0.4.0+18

stuartmorgan:
@hpoul Sorry about that, I sound have published when I landed it. Thanks for pinging. 0.4.1 is live.

PR #2613, opened 2020-03-19
[connectivity, battery, android_alarm_manager] v2 support

GaryQian:
## Description

Embedder V2 support connectivity, battery, and android alarm manager plugins

## Related Issues

https://github.com/flutter/flutter/issues/47153

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ ] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [ ] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #2622, opened 2020-03-26
[android_alarm_manager] remove MainActivity references 

cyanglaz:
## Description

Following up https://github.com/flutter/plugins/commit/899744f38216ad4f8fed1bbd8126139d7d054c4c
The `MainActivity` was removed but there were still references left. 

## Related Issues

fix ci.

### DIFF
--- a/packages/android_alarm_manager/CHANGELOG.md
+++ b/packages/android_alarm_manager/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.5+8
+
+* Remove `MainActivity` references in android example app and tests.
+
 ## 0.4.5+7
 
 * Update minimum Flutter version to 1.12.13+hotfix.5
--- a/packages/android_alarm_manager/example/android/app/src/androidTest/java/io/plugins/androidalarmmanager/DriverExtensionActivity.java
+++ b/packages/android_alarm_manager/example/android/app/src/androidTest/java/io/plugins/androidalarmmanager/DriverExtensionActivity.java
@@ -3,8 +3,8 @@
 // found in the LICENSE file.
 
 package io.flutter.plugins.androidalarmmanagerexample;
-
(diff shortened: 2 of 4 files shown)


cyanglaz:
Landing to fix CI.

vkammerer:
@cyanglaz @GaryQian : is there still additional cleanup to be done in this plugin following https://github.com/flutter/plugins/commit/899744f38216ad4f8fed1bbd8126139d7d054c4c ?

It looks like we could:
- set the application name to the default FlutterApplication in [AndroidManifest.xml](https://github.com/flutter/plugins/blob/master/packages/android_alarm_manager/example/android/app/src/main/AndroidManifest.xml)
```xml
<application android:name="io.flutter.app.FlutterApplication" >
```
- delete [Application.java](https://github.com/flutter/plugins/blob/master/packages/android_alarm_manager/example/android/app/src/main/java/io/flutter/plugins/androidalarmmanagerexample/Application.java)
- delete [EmbeddingV1Activity.java](https://github.com/flutter/plugins/blob/master/packages/android_alarm_manager/example/android/app/src/main/java/io/flutter/plugins/androidalarmmanagerexample/EmbeddingV1Activity.java)
- Replace the whole section at https://github.com/flutter/plugins/blob/master/packages/android_alarm_manager/README.md#using-other-plugins-in-alarm-callbacks with a section explaining that this plugin only supports Flutter >= 1.12.13+hotfix.5

This would simplify a lot the usage of this plugin, which can still be quite confusing for new users.

If you think these tasks can be done I'd be happy to open a PR.
I may be missing other cleanup tasks that would also need to be done, if so please let me know.

vkammerer:
I shouldn't have posted this question in this thread. Moving my question to https://github.com/flutter/plugins/pull/2613

PR #2631, opened 2020-04-03
[shared_preferences] Adds support for Windows

franciscojma86:
## Description

Dart implementation of the `shared_preferences` plugin using ffi for Windows.

## Related Issues

https://github.com/flutter/flutter/issues/41719

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -0,0 +1,119 @@
...
+  /// The path to the file in disk were the preferences are stored.
+  @visibleForTesting
+  String get getLocalDataFilePath {
+    if (_localDataFilePath != null) {
+      return _localDataFilePath;
+    }
+    _localDataFilePath = win32Wrapper.getLocalDataPath();
+    String appName = _getAppName();
+    // Append app-specific identifiers.
+    _localDataFilePath += '\\$fileDirectory\\$appName.ini';
```
csells, line 53: should use a json file extension for json data, not an ini file extension, which has its own data format: https://en.wikipedia.org/wiki/INI_file

stuartmorgan, line 53: Use `path.join` to construct paths from components.

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -0,0 +1,119 @@
...
+
+  /// Calls the win32 function for [getModuleFileName].
+  String getModuleFileName() {
+    return win32.getModuleFileName();
+  }
+}
+
+/// The Windows implementation of [SharedPreferencesStorePlatform].
+///
+/// This class implements the `package:shared_preferences` functionality for the web.
```
stuartmorgan, line 30: s/the web/Windows/

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -0,0 +1,119 @@
...
+      _cachedPreferences = stringMap.isEmpty
+          ? {}
+          : json.decode(stringMap) as Map<String, Object>;
+    }
+    return _cachedPreferences;
+  }
+
+  String _getAppName() {
+    String appPath = win32Wrapper.getModuleFileName();
+    appPath = appPath.replaceAll(extension(appPath), '');
```
stuartmorgan, line 76: `basenameWithoutExtension`?

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -0,0 +1,119 @@
...
+    return _localDataFilePath;
+  }
+
+  Map<String, Object> _cachedPreferences;
+  /// The in-memory representation of the map saved to a file in disk;
+  @visibleForTesting
+  Map<String, Object> get getCachedPreferences {
+    if (_cachedPreferences == null) {
+      File localDataFile = fs.file(getLocalDataFilePath);
+      if (!localDataFile.existsSync()) {
```
stuartmorgan, line 63: Why does reading create the file?

packages/shared_preferences/shared_preferences_windows/lib/src/win32.dart
```diff
@@ -0,0 +1,86 @@
...
+final getModuleFileNameW =
+    kernel32.lookupFunction<GetModuleFileNameC, GetModuleFileNameDart>(
+        'GetModuleFileNameW');
+
+/// Convenience method to get the AppData Local folder path.
+String getLocalDataPath() {
+  final CSIDL_LOCAL_APPDATA = 0x001c;
+  final CSIDL_FLAG_CREATE = 0x8000;
+  Pointer<Uint16> path = allocate<Uint16>(count: kMaxPath);
+  shGetFolderPath(
```
stuartmorgan, line 72: Maybe add a note about the deprecated API and why it's being used?

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
franciscojma86: There is a chance that the user calls `getAll()` before writing anything in which case it would throw an exception. As I understand it, the file should be created by an installer anyway to ensure that it will be removed once the app is uninstalled. Do you see a potential problem with this?

csells: As we discussed, the installer will almost always *not* create the file, although it may create the folder where the file will go. Further, the developer is *much* more likely to use reasonable defaults for missing values when writing the code because they won't be using and installer. Further, it's just as likely that the app doesn't even have an installer. We should *not* assume there will be an installer.

stuartmorgan: > in which case it would throw an exception

Only if you try to read it. For instance, this won't throw:

```
_cachedPreferences = {};
if (localDataFile.existsSync()) {
  String stringMap = localDataFile.readAsStringSync();
  if (!stringMap.isEmpty) {
    _cachedPreferences = json.decode(stringMap) as Map<String, Object>;
  }
}
return _cachedPreferences;
...
```

franciscojma86: I just tried that, but calling `readAsStringSync` with no file throws a `OSError: The system cannot find the file specified`.

stuartmorgan: The code I suggested only reads the file when it's there.

stuartmorgan: Maybe it wasn't clear which parts I was suggesting you replace ; here's the entire suggested function:
```
  Map<String, Object> get getCachedPreferences {
    if (_cachedPreferences == null) {
      _cachedPreferences = {};
      File localDataFile = fs.file(getLocalDataFilePath);
      if (localDataFile.existsSync()) {
        String stringMap = localDataFile.readAsStringSync();
        if (!stringMap.isEmpty) {
          _cachedPreferences = json.decode(stringMap) as Map<String, Object>;
        }
      }
    }
    return _cachedPreferences;
  }
```

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -18,8 +20,12 @@ class SharedPreferences {
   static const String _prefix = 'flutter.';
   static Completer<SharedPreferences> _completer;
 
-  static SharedPreferencesStorePlatform get _store =>
-      SharedPreferencesStorePlatform.instance;
+  static SharedPreferencesStorePlatform get _store {
+    if (LocalPlatform().isWindows) {
```
stuartmorgan, line 24: Let's add a comment here referencing https://github.com/flutter/flutter/issues/52267 so it's clear what this is a workaround for.

stuartmorgan, line 24: I think we first need to check that `kIsWeb` isn't true?

stuartmorgan, line 24: I forgot that the registration of the default implementation is in the platform interface rather than this layer; this override has the unfortunate effect that setting a different instance explicitly (e.g., the in-memory implementation, or an unofficial alternate version) won't work on Windows.

I wonder if we should temporarily add a static property to disable this, so that in those cases people can get the dynamic behavior with, e.g., `SharedPreferencesStorePlatform.disablePlatformOverride = true;` before calling `getInstance()`. Removing it later when we don't need it would technically be a breaking change, but it would be a very easy one for people to deal with (just remove the line of code). @amirh, thoughts?

packages/shared_preferences/shared_preferences_windows/LICENSE
```diff
@@ -0,0 +1,27 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
stuartmorgan, line 1: Does flutter/plugins not use an identical license everywhere like flutter/engine?

packages/shared_preferences/shared_preferences_windows/android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
stuartmorgan, line 1: I thought we only needed the gradle file to make Android builds happy?

packages/shared_preferences/shared_preferences_windows/example/LICENSE
```diff
@@ -0,0 +1 @@
+TODO: Add your license here.
```
stuartmorgan, line 1: We should either have a license, or remove the file; it looks like both have been done. I'm not sure if one or the other is the preferred practice going forward so someone who knows should weigh in.

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -0,0 +1,117 @@
...
+  /// Writes the cached preferences to disk. Returns [true] if the operation
+  /// succeeded.
+  bool _writePreferencesToFile() {
+    try {
+      File localDataFile = fs.file(getLocalDataFilePath);
+      if (!localDataFile.existsSync()) {
+        localDataFile.createSync(recursive: true);
+      }
+      String stringMap = json.encode(getCachedPreferences);
+      localDataFile.writeAsStringSync(stringMap);
```
stuartmorgan, line 87: Since all the callers are async, we should probably make this whole function async and change this to `await localDataFile.writeAsString(stringMap);`

stuartmorgan: Same for reading, now that I think about it.

packages/shared_preferences/shared_preferences_windows/LICENSE
franciscojma86: This license is the same as the one in `shared_preferences`, other than the year. Is that what you are referring to?

packages/shared_preferences/shared_preferences_windows/example/LICENSE
franciscojma86: I just copied the license from the front-end package

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
franciscojma86: What's the benefit on this? I would still have to call `await` on all the async functions. Then calls like `getCachedPreferences.clear();` would have to change to `await getCachedPreferences.clear();` which is less clean IMO.

franciscojma86: Discussed offline. There's no strong opinion on using `sync`. Will use the `*AsSync` methods.

franciscojma86:
All comments have been addressed. Now waiting on `plugin_tools` and Cirrus CI to be updated.
plugin_tools PR: https://github.com/flutter/plugin_tools/pull/95 

Also waiting on @amirh comments for the class registration.

packages/shared_preferences/shared_preferences_windows/LICENSE
stuartmorgan: Specifically I was wondering if we are using a set year. flutter/engine does that, to avoid having a ton of license variations (which adds size to Flutter applications, since they include all the licenses).

packages/shared_preferences/shared_preferences_windows/example/LICENSE
```diff
@@ -0,0 +1,27 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
```
stuartmorgan, line 1: Regardless of the policy, these two license should have the same year (since either they should both be the set year, or both be current).

packages/shared_preferences/shared_preferences_windows/example/test_driver/shared_preferences_e2e.dart
```diff
@@ -0,0 +1,89 @@
+import 'dart:async';
```
stuartmorgan, line 1: Should there be a license on this?

packages/shared_preferences/shared_preferences_windows/test/shared_preferences_windows_test.dart
```diff
@@ -0,0 +1,86 @@
+import 'package:flutter_test/flutter_test.dart';
```
stuartmorgan, line 1: License?

franciscojma86:
I changed the licenses to 2017 to match the rest of the repo. I have seen licenses with 2020 and 2019 in other packages, so I'm not really sure what's the policy. @amirh or @collinjackson might know?

GroovinChip:
https://github.com/flutter/plugin_tools/pull/95 got merged, is anything else standing in the way of this PR being merged? I'm really excited for this to be official (I'm using it as a local path dependency)

stuartmorgan:
Filed https://github.com/flutter/flutter/issues/56143 to track what's still blocking this PR.

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -0,0 +1,117 @@
...
+
+  /// Calls the win32 function for [getModuleFileName].
+  String getModuleFileName() {
+    return win32.getModuleFileName();
+  }
+}
+
+/// The Windows implementation of [SharedPreferencesStorePlatform].
+///
+/// This class implements the `package:shared_preferences` functionality for Windowds.
```
robert-ancell, line 30: Typo: "Windowds"

packages/shared_preferences/shared_preferences_windows/pubspec.yaml
```diff
@@ -0,0 +1,27 @@
+name: shared_preferences_windows
+description: Web platform implementation of shared_preferences
```
robert-ancell, line 2: "Web platform" -> "Windows"

packages/shared_preferences/shared_preferences_windows/pubspec.yaml
```diff
@@ -0,0 +1,27 @@
...
+      windows:
+        dartPluginClass: SharedPreferencesWindows
+
+dependencies:
+  shared_preferences_platform_interface: ^1.0.0
+  flutter:
+    sdk: flutter
+  meta: ^1.1.7
+  ffi: ^0.1.3
+  file: ^5.1.0
```
robert-ancell, line 18: Missing a dependency on 'path'?

skyred:
Can someone please provide instructions on how to use/test this PR with your app? I imagine add a specific reference/patch to this MR in `pubspec.yaml`

liudonghua123:
I hope this pr would be merged soon. 😄

skyred:
I tried to use the PR, by specifying below in `pubspec.yaml`

```
  shared_preferences:
    git:
      url: git@github.com:franciscojma86/plugins.git
      path: ./packages/shared_preferences/shared_preferences/
      ref: shared
```

But, got this error:

```
Error on line 34, column 11: Invalid description in the "shared_preferences" pubspec on the "shared_preferences_windows" dependency: "../shared_preferences_windows" is a relative path, but this isn't a local pubspec.

   ╷

34 │     path: ../shared_preferences_windows

   │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

   ╵
pub get failed (65;    ╵)
exit code 65
```

Haven't found a solution yet, can someone please help?

skyred:
I had use this to make it working:

Step 1: `git clone --branch shared git@github.com:franciscojma86/plugins.git --depth 1`

Step 2: Update `pubspec.yaml`

```
dependencies:
  shared_preferences: '>=0.5.8 <2.0.0'
dependency_overrides:
  shared_preferences:
    path: ../plugins/packages/shared_preferences/shared_preferences/
```

Anyway, the functionality works in this PR.

gerryhigh:
> 
> 
> I had use this to make it working:
> 
> Step 1: `git clone --branch shared git@github.com:franciscojma86/plugins.git --depth 1`
> 
> Step 2: Update `pubspec.yaml`
> 
> ```
> dependencies:
>   shared_preferences: '>=0.5.8 <2.0.0'
> dependency_overrides:
>   shared_preferences:
>     path: ../plugins/packages/shared_preferences/shared_preferences/
> ```
> 
> Anyway, the functionality works in this PR.

I tried this but get an issue when trying to run it in my app:

```
/plugins/packages/shared_preferences/shared_preferences/lib/shared_preferences.dart(38,22): error G3824A48C: Getter not found: 'isLinux'.
/plugins/packages/shared_preferences/shared_preferences/lib/shared_preferences.dart(40,29): error G3824A48C: Getter not found: 'isWindows'

```
Flutter Doctor:
[✓] Flutter (Channel master, 1.22.0-10.0.pre.153, on Microsoft Windows [Version 10.0.19041.508], locale en-US)
    • Flutter version 1.22.0-10.0.pre.153 at c:...
    • Framework revision 2e643651a9 (2 days ago), 2020-09-11 23:07:03 -0400
    • Engine revision 16b900b63e
    • Dart version 2.10.0 (build 2.10.0-117.0.dev)
[✓] Visual Studio - develop for Windows (Visual Studio Community 2019 16.7.2)
    • Visual Studio at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community
    • Visual Studio Community 2019 version 16.7.30413.136
    • Windows 10 SDK version 10.0.18362.0
[✓] Connected device (1 available)
    • Windows (desktop) • windows • windows-x64 • Microsoft Windows [Version 10.0.19041.508]

stuartmorgan:
Closing as obsolete. Most of the implementation landed in https://github.com/flutter/plugins/pull/2988, it will be published when https://github.com/flutter/plugins/pull/3058 lands, and then https://github.com/flutter/plugins/pull/3059 will add the endorsement.

PR #2639, opened 2020-04-07
[webview_flutter] Add onReceivedError callback 

bparrishMines:
## Description

This adds the feature where the webview can report an error when loading page resources. 

After looking for error reporting in the `Webview` implementation of each platform I found Android uses [onReceivedError](https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedError(android.webkit.WebView,%20android.webkit.WebResourceRequest,%20android.webkit.WebResourceError)) in its `WebviewClient` and iOS uses [webView:didFailNavigation:withError:](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455623-webview?language=objc) and [webView:didFailProvisionalNavigation:withError:](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455637-webview?language=objc) in its `WKNavigationDelegate`. 

However, each platform handles error codes differently. Android provides a `WebResourceError` which has an `int` code that is defined as a constant in `WebViewClient`. iOS uses a generic `NSError` and doesn't specify the possible values for the error code.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/53023

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
@@ -175,14 +175,18 @@ class FlutterWebViewClient {
         FlutterWebViewClient.this.onPageFinished(view, url);
       }
 
-      // We don't need to support the deprecated version since this WebViewClient is only used on
-      // API versions >= N. See `FlutterWebViewClient.createWebViewClient`.
       @TargetApi(Build.VERSION_CODES.M)
       @Override
       public void onReceivedError(
           WebView view, WebResourceRequest request, WebResourceError error) {
         FlutterWebViewClient.this.onReceivedError(
             error.getErrorCode(), error.getDescription().toString());
+      }
+
+      @Override
+      public void onReceivedError(
+          WebView view, int errorCode, String description, String failingUrl) {
+        FlutterWebViewClient.this.onReceivedError(errorCode, description);
       }
 
       @Override
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
@@ -123,9 +123,9 @@ class FlutterWebViewClient {
     methodChannel.invokeMethod("onPageFinished", args);
   }
 
-  private void onReceivedError(final String errorCode, final String description) {
+  private void onReceivedError(final int errorCode, final String description) {
     final Map<String, Object> args = new HashMap<>();
-    args.put("errorCode", errorCode);
+    args.put("errorCode", FlutterWebViewClient.errorCodeToString(errorCode));
     args.put("description", description);
     methodChannel.invokeMethod("onReceivedError", args);
   }
@@ -179,8 +179,7 @@ class FlutterWebViewClient {
       @TargetApi(Build.VERSION_CODES.M)
       @Override
       public void onReceivedError(WebView view, WebResourceRequest request, WebResourceError error) {
-        final String errorCode = FlutterWebViewClient.errorCodeToString(error.getErrorCode());
-        FlutterWebViewClient.this.onReceivedError(errorCode, error.getDescription().toString());
+        FlutterWebViewClient.this.onReceivedError(error.getErrorCode(), error.getDescription().toString());
       }
 
       @Override
@@ -219,14 +218,12 @@ class FlutterWebViewClient {
       @SuppressLint("RequiresFeature")
       @Override
       public void onReceivedError(WebView view, WebResourceRequest request, WebResourceErrorCompat error) {
-        final String errorCode = FlutterWebViewClient.errorCodeToString(error.getErrorCode());
-        FlutterWebViewClient.this.onReceivedError(errorCode, error.getDescription().toString());
+        FlutterWebViewClient.this.onReceivedError(error.getErrorCode(), error.getDescription().toString());
       }
 
       @Override
       public void onReceivedError(WebView view, int errorCode, String description, String failingUrl) {
-        final String errorCodeString = FlutterWebViewClient.errorCodeToString(errorCode);
-        FlutterWebViewClient.this.onReceivedError(errorCodeString, description);
+        FlutterWebViewClient.this.onReceivedError(errorCode, description);
       }
 
       @Override
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
@@ -37,37 +37,37 @@ class FlutterWebViewClient {
   private static String errorCodeToString(int errorCode) {
     switch (errorCode) {
       case WebViewClient.ERROR_AUTHENTICATION:
-        return "AUTHENTICATION";
+        return "authentication";
       case WebViewClient.ERROR_BAD_URL:
-        return "BAD_URL";
+        return "badUrl";
       case WebViewClient.ERROR_CONNECT:
-        return "CONNECT";
+        return "connect";
       case WebViewClient.ERROR_FAILED_SSL_HANDSHAKE:
-        return "FAILED_SSL_HANDSHAKE";
+        return "failedSslHandshake";
       case WebViewClient.ERROR_FILE:
-        return "FILE";
+        return "file";
       case WebViewClient.ERROR_FILE_NOT_FOUND:
-        return "FILE_NOT_FOUND";
+        return "fileNotFound";
       case WebViewClient.ERROR_HOST_LOOKUP:
-        return "HOST_LOOKUP";
+        return "hostLookup";
       case WebViewClient.ERROR_IO:
-        return "IO";
+        return "io";
       case WebViewClient.ERROR_PROXY_AUTHENTICATION:
-        return "PROXY_AUTHENTICATION";
+        return "proxyAuthentication";
       case WebViewClient.ERROR_REDIRECT_LOOP:
-        return "REDIRECT_LOOP";
+        return "redirectLoop";
       case WebViewClient.ERROR_TIMEOUT:
-        return "TIMEOUT";
+        return "timeout";
       case WebViewClient.ERROR_TOO_MANY_REQUESTS:
-        return "TOO_MANY_REQUESTS";
+        return "tooManyRequests";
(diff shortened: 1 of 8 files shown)


packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart
```diff
@@ -576,6 +576,31 @@ void main() {
...
+              });
+            },
+          ),
+        ),
+      );
+
+      final Map<String, String> errorResult = await errorCompleter.future;
+      expect(errorResult['errorCode'], isNotNull);
+      expect(errorResult['description'], isNotNull);
+    });
```
amirh, line 602: Let's add a test with a valid url that verifies the error callback is not invoked on a valid url. I'd use a data url like: 

https://github.com/flutter/plugins/blob/c03166bc12791bebc172b63259bd7200c052187e/packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart#L116

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
```diff
@@ -30,6 +34,47 @@
...
+      case WebViewClient.ERROR_UNSAFE_RESOURCE:
+        return "UNSAFE_RESOURCE";
+      case WebViewClient.ERROR_UNSUPPORTED_AUTH_SCHEME:
+        return "UNSUPPORTED_AUTH_SCHEME";
+      case WebViewClient.ERROR_UNSUPPORTED_SCHEME:
+        return "UNSUPPORTED_SCHEME";
+    }
+
+    final String message =
+        String.format(Locale.getDefault(), "Could not find a string for errorCode: %d", errorCode);
```
amirh, line 74: nit:

Is `%d` formatting affected by the locale?
Also seems like you can omit specifying `Locale.getDefault()` as this is the default for the `String.format` version that doesn't take a locale.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -277,6 +281,22 @@ class WebView extends StatefulWidget {
...
+  ///
+  /// This can be called for any resource (iframe, image, etc.), not just for
+  /// the main page.
+  ///
+  /// On Android, the error code will be a constant from
+  /// [WebViewClient](https://developer.android.com/reference/android/webkit/WebViewClient#summary).
+  /// (e.g. The error code of `WebViewClient.ERROR_AUTHENTICATION` on Android
+  /// translates to `AUTHENTICATION` in Dart.
+  ///
+  /// On iOS, the error code will represent the "domain" and the "code" for an
```
amirh, line 294: My preference would be to avoid having different semantic meaning to the same field on different platforms.
Instead I'd have different fields with a single semantic meaning which may not be set on all platforms.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -277,6 +281,22 @@ class WebView extends StatefulWidget {
...
+  /// `NSError` in Objective-C. The format will be `$domain: $code`. See
```
amirh, line 295: Is $code the http status code?

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -79,6 +79,9 @@ typedef void PageStartedCallback(String url);
 /// Signature for when a [WebView] has finished loading a page.
 typedef void PageFinishedCallback(String url);
 
+/// Signature for when a [WebView] has failed to load a resource.
+typedef void ReceivedErrorCallback(String errorCode, String description);
```
amirh, line 83: When reading this my gut was that we should have an error class though I'm not finding good arguments any way 😄 (slight pro of an object is that it's easier for users to pass the error information around within their app)

I wonder if you had a reason to prefer not introducing an error class.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -147,6 +150,7 @@ class WebView extends StatefulWidget {
     this.gestureRecognizers,
     this.onPageStarted,
     this.onPageFinished,
+    this.onReceivedError,
```
amirh, line 153: another nit: why onReceivedError and not onError/onResourceLoadError?

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
bparrishMines: I only added the `Locale.getDefault()` to get rid of the lint warning. It does change the digit to whatever locale the device is set to. I don't think it would change the interpretation of the digit tho.

packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart
bparrishMines: done

packages/webview_flutter/lib/webview_flutter.dart
bparrishMines: done. I added a `domain` only for iOS and a `errorType` only for Android.

packages/webview_flutter/lib/webview_flutter.dart
bparrishMines: I think it's possible for it to be an http status code, but the documentation doesn't say anything about it:
https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455623-webview?language=objc

However, their https://developer.apple.com/documentation/webkit/wkerrorcode?language=objc states that these are possible values. I can't tell if these are the ONLY possible values or not.

packages/webview_flutter/lib/webview_flutter.dart
bparrishMines: I was in favor of a data class as well. I chose not to create one to be consistent with the style of the `WebViewPlatformCallbacksHandler`.  Method
```
void onJavaScriptChannelMessage(String channel, String message);
```
and
```
FutureOr<bool> onNavigationRequest({String url, bool isForMainFrame});
```
don't use them either.

However, since I need more than 2 parameters now, I created a `WebResourceError` class.

packages/webview_flutter/lib/webview_flutter.dart
bparrishMines: It was the name of the method in `WebViewClient`. Same as `onPageStarted` and `onPageFinished`. It was mostly to keep the pattern of method name origins, but I think `WebResourceError` is more expressive. (or something similar)

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
amirh: Interesting, my read of https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html#syntax is that `%d` formatting is not affected by the locale.

packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart
```diff
@@ -576,6 +578,52 @@ void main() {
       expect(currentUrl, 'https://www.google.com/');
     });
 
+    testWidgets('onReceivedError', (WidgetTester tester) async {
```
amirh, line 581: nit: onWebResourceError

packages/webview_flutter/example/test_driver/webview_flutter_e2e.dart
```diff
@@ -576,6 +578,52 @@ void main() {
...
+      );
+
+      final WebResourceError error = await errorCompleter.future;
+      expect(error, isNotNull);
+
+      if (Platform.isIOS) expect(error.domain, isNotNull);
+      if (Platform.isAndroid) expect(error.errorType, isNotNull);
+    });
+
+    testWidgets('onReceivedError is not called with valid url',
```
amirh, line 605: ditto

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -28,6 +28,123 @@ abstract class WebViewPlatformCallbacksHandler {
...
+  webViewInvalidated,
+
+  /// A JavaScript exception occurred.
+  javaScriptExceptionOccurred,
+
+  /// The result of JavaScript execution could not be returned.
+  javaScriptResultTypeIsUnsupported,
+}
+
+/// On Android, the error code will be a constant from
```
amirh, line 99: This class comment probably needs an update as well?

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -28,6 +28,123 @@ abstract class WebViewPlatformCallbacksHandler {
...
+  /// The domain of where to find the error code.
+  ///
+  /// This field is only available on iOS and represents a "domain" from where
+  /// the [errorCode] is from. This value is taken directly from an `NSError`
+  /// in Objective-C. See
+  /// https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/ErrorObjectsDomains/ErrorObjectsDomains.html
+  /// for more information on error handling on iOS.
+  final String domain;
+
+  /// Describes the error.
```
amirh, line 141: Style guide says [avoid useless documentation](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-useless-documentation)

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -28,6 +28,123 @@ abstract class WebViewPlatformCallbacksHandler {
...
+  final String description;
+
+  /// The type of error.
+  ///
+  /// This is will never be null on Android, but can be null on iOS.
```
amirh, line 146: nit: s/is//

packages/webview_flutter/lib/src/webview_method_channel.dart
```diff
@@ -43,9 +43,31 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
...
+        _platformCallbacksHandler.onWebResourceError(
+          WebResourceError(
+            errorCode: call.arguments['errorCode'],
+            description: call.arguments['description'],
+            domain: call.arguments['domain'],
+            errorType: call.arguments['errorType'] == null
+                ? null
+                : WebResourceErrorType.values.firstWhere(
+                    (WebResourceErrorType type) {
+                      final String enumName = type.toString().replaceFirst(
```
amirh, line 56: nit, may be easier to follow as:

```dart
return type.toString() == '$WebResourceErrorType.${call.arguments['errorType']}';
```

amirh:
Don't forget to add a version bump and a changelog entry

PR #2645, opened 2020-04-11
[android_intent] Ignoring packages in intent.

NikulinEB:
## Description

This update allows to specify a list of packages that should not be used to resolve the intent. For example, this is useful with links to the Steam store. When click on such a link, the Steam application appears in the default selection dialog, but the application crashes when opening a link. This happens due to the lack of a component name in the intent. Therefore, it would be more convenient to exclude the Steam app from this list to avoid crashes when opening a link. And offer the user a choice: open in the Steam app (using the intent with the specified package and component names) or open in a browser (by default or from the list of available browsers without the Steam app). And in this update, a parameter is added to force the default selection dialog to be displayed, if necessary.

## Related Issues

I did not find or create an issue about this. But in the previous paragraph I described the situation in which this update is useful.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -106,6 +111,9 @@ void setApplicationContext(@Nullable Context applicationContext) {
    * @param packageName forwarded to {@link Intent#setPackage(String)} if non-null. This is forced
    *     to null if it can't be resolved.
    * @param componentName forwarded to {@link Intent#setComponent(ComponentName)} if non-null.
+   * @param ignoredPackages list of package names that should not by displayed in the chooser and
```
matthew-carroll, line 114: type: "by" -> "be"

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -106,6 +111,9 @@ void setApplicationContext(@Nullable Context applicationContext) {
...
+   *     should not be used to resolve this intent.
+   * @param showChooser forces to show the default selection dialog.
```
matthew-carroll, line 116: What happens if there's no UI/`Activity` available and `showChooser` is set to `true`?

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -158,6 +169,45 @@ Intent buildIntent(
         intent.setPackage(null);
       }
     }
+    Boolean needIgnore = ignoredPackages != null && ignoredPackages.size() > 0;
```
matthew-carroll, line 172: Can we clean up this code a bit for readability? It's difficult to quickly see the relevant behavior vs the low level execution. Based on what I see, this could probably be reduced to something like:

```
boolean isPackageFilterRequired = ignoredPackages != null && ignoredPackages.size() > 0;

if (isPackageFilterRequired || isChooserDesired) {
  List<Intent> targetIntents = calculateTargetIntents(...);
  
  if (isChooserDesired) {
    String choosePackageToLaunch(targetIntents);
    intent.setPackage(packageToLaunch);
  } else {
    //...The code that would go here has other issues that I'll comment on
    // but this example should give you an idea of the functional decomposition
    // that I'm thinking of.
  }
}

return intent;
```

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -158,6 +169,45 @@ Intent buildIntent(
...
+          } else {
+            packageToLaunch = activities.get(0).activityInfo.packageName;
+          }
+          intent.setPackage(packageToLaunch);
+          Log.v(TAG, "Resolve this intent with package: " + packageToLaunch);
+        } catch (IndexOutOfBoundsException exception) {
+          Log.v(TAG, "Not found packages to resolve this intent: " + exception);
+        }
+      } else {
+        intent = Intent.createChooser(targetIntents.remove(0), "");
```
matthew-carroll, line 207: I have 2 concerns on this line. First, I don't thin we should be replacing the `intent` value. It is very easy for someone to miss this line and believe that `intent` always refers to the same instance throughout this method.

Second, is there a reason that `remove(0)` is being called? If not, please consider changing it to `get(0)`. If there is a reason, then please consider moving that call to its own line and commenting as to why we need `remove()` instead of `get()`.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -158,6 +169,45 @@ Intent buildIntent(
...
+        String packageToLaunch;
+        try {
+          if (ignoredPackages != null) {
+            packageToLaunch = targetIntents.get(0).getPackage();
+          } else {
+            packageToLaunch = activities.get(0).activityInfo.packageName;
+          }
+          intent.setPackage(packageToLaunch);
+          Log.v(TAG, "Resolve this intent with package: " + packageToLaunch);
+        } catch (IndexOutOfBoundsException exception) {
```
matthew-carroll, line 203: I think I would suggest checking the size explicitly before accessing an item and then voiding the try/catch altogether.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -158,6 +169,45 @@ Intent buildIntent(
...
+          Log.v(TAG, "Not found packages to resolve this intent: " + exception);
```
matthew-carroll, line 204: This log should either be removed or improved.

If this log refers to an expectation that the developer has, and that expectation is not met, then I think this should probably be a "warning" at minimum. 

Also the `IndexOutOfBoundsException` is unlikely to help a developer. Instead, this log could say something like:
`"Intent has no available targets. Some targets were explicitly ignored: " + Arrays.toString(ignoredPackages)`

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -158,6 +169,45 @@ Intent buildIntent(
...
+      if (!showChooser) {
+        String packageToLaunch;
+        try {
+          if (ignoredPackages != null) {
+            packageToLaunch = targetIntents.get(0).getPackage();
+          } else {
+            packageToLaunch = activities.get(0).activityInfo.packageName;
+          }
+          intent.setPackage(packageToLaunch);
+          Log.v(TAG, "Resolve this intent with package: " + packageToLaunch);
```
matthew-carroll, line 202: Please remove this log.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
NikulinEB: If there is no activity available, then `targetIntents` will be empty and the сhooser will not be created. The intent will be created as if `showChooser` would be `false`.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
NikulinEB: I call `remove(0)` because I then call the `putExtra` method with the entire array of packages. If I do not remove the first intent, then the corresponding package will be in the chooser twice. I could call `createChooser` with `new Intent()` as the first argument, but I'm not sure that is good.

NikulinEB:
Thanks for the comments! I updated the pull request.

ened:
@NikulinEB @matthew-carroll the launchIntent method seems already a bit complex. Do you think it's more approachable to have a new "showChooser" method which handles the appropriate use case only?

NikulinEB:
@ened Good suggestion, thanks!
I added the `showChooser` method, several tests, and updated the pull request.

NikulinEB:
Hi @matthew-carroll @mklim 
Could you review this PR, please?
Thanks in advance!

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
mklim: It also took me a bit of reading to figure out why we were popping the head from the list here and if it was a mistake. I think the decision makes sense, but an implementation comment would be helpful.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -158,7 +185,46 @@ Intent buildIntent(
         intent.setPackage(null);
       }
     }
+    Boolean isPackageFilterRequired = ignoredPackages != null && ignoredPackages.size() > 0;
```
mklim, line 188: nit: the primitive `boolean` would be preferable here, confusingly `Boolean` is an object and could theoretically be null.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -158,7 +185,46 @@ Intent buildIntent(
...
+    if (isPackageFilterRequired) {
+      String packageToLaunch = choosePackageToLaunch(intent, ignoredPackages);
+      intent.setPackage(packageToLaunch);
+    }
 
     return intent;
   }
+
+  String choosePackageToLaunch(Intent intent, List<String> ignoredPackages) {
```
mklim, line 197: This and the below helper methods should be `private`, and I think also `static`?

packages/android_intent/lib/android_intent.dart
```diff
@@ -27,6 +27,8 @@ class AndroidIntent {
   /// [package] refers to the package parameter of the intent, can be null.
   /// [componentName] refers to the component name of the intent, can be null.
   /// If not null, then [package] but also be provided.
+  /// [ignoredPackages] is the list of package names that should not by displayed in the selection dialog and
+  /// should not be used to resolve this intent.
```
mklim, line 31: I think it would also be worth saying this is only used in the `showChooser` method.

packages/android_intent/lib/android_intent.dart
```diff
@@ -100,6 +104,10 @@ class AndroidIntent {
   ///
   /// See https://developer.android.com/reference/android/content/Intent.html#setComponent(android.content.ComponentName).
   final String componentName;
+
+  /// List of package names that will be excluded from the list of applications to resolve the intent.
+  final List<String> ignoredPackages;
```
mklim, line 109: Same here, I think it's worth saying this is only being used for `showChooser`.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -59,6 +63,26 @@ void send(Intent intent) {
     }
   }
 
+  void showChooser(Intent intent, @Nullable List<String> ignoredPackages) {
+    Intent chooser = null;
+    intent.setPackage(null);
+    intent.setComponent(null);
+    Boolean isPackageFilterRequired = ignoredPackages != null && ignoredPackages.size() > 0;
```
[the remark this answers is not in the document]
NikulinEB: Thanks for the comments. I added `ignoredPackages` initialization in the `onMethodCall` method (MethodCallHandlerImpl.java). Tell me if this is not what you meant.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
NikulinEB: I made them only private, but not static, because they use the instance fields `activity` and `applicationContext`, which I would have to pass through several static methods.

packages/android_intent/lib/android_intent.dart
NikulinEB: The `ignoredPackages` argument is used in the `launch` method, if the package is not set.

NikulinEB:
Hi @matthew-carroll 
Could you please review the updated pull request and approve it, if everything is fine?
And tell me, please, should I do something with the check that was not successful?

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
matthew-carroll: Can you add that information to the docs?

packages/android_intent/android/src/test/java/io/flutter/plugins/androidintent/MethodCallHandlerImplTest.java
```diff
@@ -283,4 +285,38 @@ public void onMethodCall_setsDataAndType() {
     assertEquals(type, intent.getType());
     assertEquals(data, intent.getData());
   }
+
+  @Test
+  public void onMethodCall_setsIgnoredPackages() {
```
matthew-carroll, line 290: Can you describe what this is testing? As far as I can tell, this test is verifying that including ignored packages doesn't impact the result at all. But isn't it also possible that sometimes the ignored packages will impact the behavior? Shouldn't there a best for that, too?

packages/android_intent/android/src/test/java/io/flutter/plugins/androidintent/MethodCallHandlerImplTest.java
```diff
@@ -283,4 +285,38 @@ public void onMethodCall_setsDataAndType() {
...
+
+    verify(result, times(1)).success(null);
+    Intent intent = shadowOf((Application) context).getNextStartedActivity();
+    assertNotNull(intent);
+    assertEquals("foo", intent.getAction());
+    assertNotEquals("io.flutter.plugins.androidintent", intent.getPackage());
+  }
+
+  @Test
+  public void onMethodCall_showChooser() {
```
matthew-carroll, line 309: I'm not sure what this test is verifying, either. Shouldn't there be some verification that the new "show chooser" behavior did something? Maybe I'm overlooking some details in this test, but it looks like it's verifying the same kind of behavior that we'd expect if we ran the regular `Intent` behavior, right?

packages/android_intent/lib/android_intent.dart
```diff
@@ -27,6 +27,8 @@ class AndroidIntent {
   /// [package] refers to the package parameter of the intent, can be null.
   /// [componentName] refers to the component name of the intent, can be null.
   /// If not null, then [package] but also be provided.
+  /// [ignoredPackages] is the list of package names that should not by displayed in the selection dialog and
```
matthew-carroll, line 30: type: "by" -> "be"

packages/android_intent/lib/android_intent.dart
```diff
@@ -152,6 +160,26 @@ class AndroidIntent {
...
 
+  /// Show the default selection dialog to allow the user to select
+  /// an application to resolve this intent from the list of available applications.
+  ///
+  /// [action] must be specified.
+  ///
+  /// [package] and [componentName] will be ignored.
+  ///
+  /// This works only on Android platforms.
+  Future<void> showChooser() async {
```
matthew-carroll, line 171: I'll connect my earlier question to this point here. When a developer invokes this method, does that developer expect that a chooser will definitely be displayed? If so, how can the developer take action in the case that no `Activity` is available, or in the case that no target `Activity`s exist?

packages/android_intent/test/android_intent_test.dart
```diff
@@ -140,6 +140,52 @@ void main() {
...
+          await androidIntent.showChooser();
+          fail('should raise a StateError');
+        } on StateError catch (e) {
+          expect(e.message, 'action must not be null or empty string');
+        } catch (e) {
+          fail('should raise a StateError');
+        }
+      });
+
+      test('pass right params', () async {
```
matthew-carroll, line 162: I don't think `"pass right params"` is a name that future developers will be able to maintain. Can you please specify what condition you're testing so that there's no confusion in the future? In other words, a developer might wonder "what are 'right' params and what are 'wrong' params?" and "is there only 1 possible set of 'right' params?"

packages/android_intent/test/android_intent_test.dart
```diff
@@ -140,6 +140,52 @@ void main() {
...
+        verify(mockChannel.invokeMethod<void>('showChooser', <String, Object>{
+          'action': 'action_view',
+          'data': Uri.encodeFull('https://flutter.io'),
+          'flags':
+              androidIntent.convertFlags(<int>[Flag.FLAG_ACTIVITY_NEW_TASK]),
+          'type': 'video/*',
+        }));
+      });
+
+      test('call in ios platform', () async {
```
matthew-carroll, line 180: Rather than naming a test by what it does, consider naming the test based on what you're verifying. For example, this test might be called `'does nothing when called on iOS'`.

packages/android_intent/test/android_intent_test.dart
```diff
@@ -140,6 +140,52 @@ void main() {
         verifyZeroInteractions(mockChannel);
       });
     });
+
+    group('showChooser', () {
```
matthew-carroll, line 144: Related to a couple of my earlier questions, I think that tests should probably exist in this group based on a situation where no target activity is found, and any other situation that would result in a chooser not being shown.

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -59,6 +64,30 @@ void send(Intent intent) {
     }
   }
 
+  void showChooser(Intent intent, @NonNull List<String> ignoredPackages) {
+    Intent chooser = null;
+    intent.setPackage(null);
+    intent.setComponent(null);
+
+    // Exclude ignored packages from the displayed list.
+    if (ignoredPackages.size() > 0) {
```
matthew-carroll, line 73: Is this if-statement correct? Why do we need to ignore packages to show a chooser?

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -59,6 +64,30 @@ void send(Intent intent) {
...
+      if (targetIntents != null && targetIntents.size() > 0) {
+        // Create a chooser with the first element of the list,
+        // removing it from the list to avoid double displaying it in the chooser,
+        // since we add the whole list further using putExtra.
+        chooser = Intent.createChooser(targetIntents.remove(0), "");
+        chooser.putExtra(Intent.EXTRA_INITIAL_INTENTS, targetIntents.toArray(new Parcelable[] {}));
+      }
+    }
+
+    if (chooser != null) {
```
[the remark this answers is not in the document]
NikulinEB: The reason for having this method is not for some specific purpose for using it. This method appeared to simplify the `buildIntent` method by removing the corresponding `showChooser` parameter from it. This parameter appeared to save the previous method behavior. Earlier, when calling the `send` method of this class, the default chooser might appear or not, depending on whether the default application is set to resolve this intent. That is, the appearance of the chooser was not controlled in any way by the developer (unless explicitly specifying a package so as not to show the chooser).

When I implemented ignoring the specified packages, there were two options: use any available activity to resolve this intent (using the `intent.setPackage` method), if it is not in the ignore list - in this case, the chooser will never appear because the activity is already selected. Another option was to show the chooser with all available activities except ignored. Therefore, this `showChooser` method appeared, so that the developer can explicitly choose this method to call, if he prefers to show the chooser.

If the chooser is null, then no available activities were found and the same thing will happen as if we called the `send` method for this intent.

What do you think about the need for this method or its name?

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
NikulinEB: This is a mistake, thanks

packages/android_intent/android/src/main/java/io/flutter/plugins/androidintent/IntentSender.java
```diff
@@ -158,7 +189,46 @@ Intent buildIntent(
...
+      return targetIntents.get(0).getPackage();
+    } else {
+      return null;
+    }
+  }
+
+  private List<ResolveInfo> getIntentActivities(Intent intent) {
+    PackageManager packageManager;
+    if (activity != null) {
+      packageManager = activity.getPackageManager();
```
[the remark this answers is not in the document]
NikulinEB: I'm not sure, but it seems to me that the same instance is being returned

packages/android_intent/android/src/test/java/io/flutter/plugins/androidintent/MethodCallHandlerImplTest.java
NikulinEB: You're right, I will fix this test.

packages/android_intent/android/src/test/java/io/flutter/plugins/androidintent/MethodCallHandlerImplTest.java
NikulinEB: I would say that it tests that the package that was added to the `ignoredPackages` is not used to resolve the intent.
I will try to update the test to explicitly verify that the specified package to ignore was excluded.

packages/android_intent/test/android_intent_test.dart
NikulinEB: Just used the same name as in the existing test. No problem, I will fix it.

NikulinEB:
Hello @matthew-carroll! I apologize for the long absence. I updated the tests for displaying a chooser and ignoring packages, and fixed some of the problems you wrote about. Also I added comments, so I hope that now the action of the `showChooser` method is clear.
Please review again, I hope in the last time. Thank you in advance!

NikulinEB:
@matthew-carroll Hi! I am here again. Let's finish this pull request. Please review and approve if everything is ok. And who should merge these changes?

matthew-carroll:
Hi @NikulinEB, I'm swamped with other things at the moment. Perhaps a member of the team can help get this merged. CC @amirh

stuartmorgan:
We apologize for the long delay in following up on this PR. We’ve made the decision to no longer accept non-critical PRs for this plugin, as we hope that in time we’re able to transition users to the corresponding plugin in [the Flutter Community Plus Plugins repository](https://plus.fluttercommunity.dev/), so we encourage you to submit your PR there. I realize it's not ideal that this decision is happening after there has already gone through some review here, but hopefully that feedback will still be useful for a Flutter Community Plus Plugin version of the same change.

We’re in the process of overhauling our PR triage system to respond much more quickly to avoid this kind of delay going forward. (Transitioning plugins like this one to the Flutter Community group is in fact part of that effort.)

PR #2646, opened 2020-04-11
[in_app_purchase] Fix cannot fix transaction issue

LHLL:
## Description

1. Solves `complete transaction`(AKA `finishTransaction`) is not working on iOS. Current design can only finish transactions with state `purchased` and `failed`. Transactions with other states such `deferred` will fail with a wrong error.
2. Allows only one pending transaction per product identifier. This change can help prevent client app from purchasing the same products multiple times by accident. Such as auto-renewable subscriptions and deferred transactions.

## Related Issues

https://github.com/flutter/flutter/issues/51773
https://github.com/flutter/flutter/issues/53534
https://github.com/flutter/flutter/issues/32759

## Checklist

- [x] I signed the [CLA].

## Breaking Change


cyanglaz:
Thanks for the PR! Is https://github.com/flutter/flutter/issues/32759 also related?

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.h
```diff
@@ -26,11 +29,19 @@ typedef void (^UpdatedDownloads)(NSArray<SKDownload *> *downloads);
...
 // Can throw exceptions if the transaction type is purchasing, should always used in a @try block.
 - (void)finishTransaction:(nonnull SKPaymentTransaction *)transaction;
 - (void)restoreTransactions:(nullable NSString *)applicationName;
 
+// This method needs to be called before any other methods.
+- (void)startObservingPaymentQueue;
+
+// Appends a payment to the SKPaymentQueue.
+//
+// @param payment Payment object that needs to be added to the payment queue.
```
cyanglaz, line 41: nits:
"Payment object to be added to the payment queue."

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.h
```diff
@@ -26,11 +29,19 @@ typedef void (^UpdatedDownloads)(NSArray<SKDownload *> *downloads);
...
+// @return A flag that indicates whether or not this payment can be added.
```
cyanglaz, line 42: nits:
//@return whether "addPayment" was successful.

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
```diff
@@ -31,21 +32,28 @@ - (instancetype)initWithQueue:(nonnull SKPaymentQueue *)queue
...
+    _shouldAddStorePayment = shouldAddStorePayment;
+    _updatedDownloads = updatedDownloads;
+    _transactionsSetter = [NSMutableDictionary dictionary];
   }
   return self;
 }
 
-- (void)addPayment:(SKPayment *)payment {
+- (void)startObservingPaymentQueue {
+  [_queue addTransactionObserver:self];
```
cyanglaz, line 48: We should find someway to indicate if the queue is observed. And all the other methods should throw if it is not observed. 
This can make users easier to debug if they forgot to observe.
Edit: After further review, I saw that this method is only called within the plugin, then it is less important to have the check. Might still have benefits for debugging and testing purpose. You call! :)

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
```diff
@@ -31,21 +32,28 @@ - (instancetype)initWithQueue:(nonnull SKPaymentQueue *)queue
...
+}
+
+- (BOOL)addPayment:(SKPayment *)payment {
+  if (self.transactionsSetter[payment.productIdentifier]) {
```
cyanglaz, line 52: Can user buy multiple same "consumable"? This seems to prevent that to happen.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -151,25 +152,34 @@ - (void)addPayment:(FlutterMethodCall *)call result:(FlutterResult)result {
...
+    payment.simulatesAskToBuyInSandbox =
+        [[paymentMap objectForKey:@"simulatesAskToBuyInSandBox"] boolValue];
+  }
+
+  if (![self.paymentQueueHandler addPayment:payment]) {
+    error = [FlutterError
+        errorWithCode:@"storekit_duplicate_product_object"
+              message:@"There is a pending transaction for the same product identifier. Please "
+                      @"either wait"
+                      @"for it to be finished or finish it manuelly to avoid possible edge cases."
```
cyanglaz, line 179: nits:
We can add more information here for easier debugging. For example, we can tell users what methods to call to finish the transaction manually.

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
```diff
@@ -151,25 +152,34 @@ - (void)addPayment:(FlutterMethodCall *)call result:(FlutterResult)result {
...
-    return;
+  FlutterError *error;
+  if (!product) {
+    error = [FlutterError
+        errorWithCode:@"storekit_invalid_payment_object"
+              message:
+                  @"You have requested a payment for an invalid product. Either the "
+                  @"`productIdentifier` of the payment is not valid or the product has not been "
+                  @"fetched before adding the payment to the payment queue."
+              details:call.arguments];
```
cyanglaz, line 163: Can we return early here?
```
result(error);
return;
```

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
cyanglaz: I also don't see a test for this.
Another option is to do it on the dart side, which is easier testable.

LHLL:
@cyanglaz [flutter/flutter#32759](https://github.com/flutter/flutter/issues/32759) should also be fixed after this pr.

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
LHLL: Right now this method is called in the:
`- (instancetype)initWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar`
I assume there won't be a case that `startObservingPaymentQueue ` is not called?

Another thing to mention, it's normally a bigger issue when this method is called multiple times during the app running such as user switches account during the app running, but I guess it might not be a concern for Flutter apps since plugin is only configured once when app launches?

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
LHLL: Normally, a good user experience would be app showing a blocking UI when a transaction is still pending such as Amazon. Also, if we allow the user to purchase the same product more than once, it might be very hard for apps that allows multi-login to deliver products to the right user.

Let's say app uses Google Sign-in as app's account system, user signed-in with account foo, purchased product A and closed the app before transaction finishes. At this time, transaction in the SKPaymentQueue doesn't have a transaction identifier yet, thus, there is no way for the app to establish an one-to-one mapping between the user account and the transaction. (applicationUserName is nil when state changes.) Next time, user launches the app and signed-in with account bar. The product will be delivered to bar instead of foo. If app has offers, then there will be more too many edge cases to handle on the app side. Limiting one pending transaction per product identifier will not hurt user experience a lot but dramatically decrease the complicity of the issue both us and client engineer need to solve.

I will add a test case on the dart side, thanks for calling this out, good catch!

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
LHLL: Done

packages/in_app_purchase/ios/Classes/InAppPurchasePlugin.m
LHLL: Done

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
LHLL: Does `+ (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar>*)registrar;` get called every time when a new registration happens? If so, I can just stop listening in this method.

packages/in_app_purchase/ios/Classes/FIAPaymentQueueHandler.m
LHLL: Good catch! Added logic and its associated test.

PR #2647, opened 2020-04-12
Upgrade dependency google_sign_in_web: ^0.9.1

eliudio:
google_sign_in_web version 0.9.0 fixes an issue which otherwise causes compilation errors like "Error: JS interop classes do not support non-external constructors. Try annotating with `external`. GoogleAuth.fakeConstructor$();  "

More info see: https://stackoverflow.com/questions/60911568/flutter-build-for-web-failed-to-compile-application/61167824#61167824


eliudio:
@googlebot I signed it!

ditman:
The fix for this issue happened in version [0.8.4](https://pub.dev/packages/google_sign_in_web#084). Why do you need to do this?

ditman:
> The fix for this issue happened in version [0.8.4](https://pub.dev/packages/google_sign_in_web#084). Why do you need to do this?

Because pub.dev treats 0.x dependencies in a special way. See [this](https://dart.dev/tools/pub/versioning#semantic-versions).

packages/google_sign_in/google_sign_in/pubspec.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
   # validation, so we set a ^ constraint.
   # TODO(amirh): Revisit this (either update this part in the  design or the pub tool).
   # https://github.com/flutter/flutter/issues/46264
-  google_sign_in_web: ^0.8.2
+  google_sign_in_web: ^0.9.0
```
ditman, line 28: In order to publish this, you'll also need to update the version of google_sign_in in this pubspec, and add an entry to the CHANGELOG file. Can you do that?

I think the `google_sign_in` package would need to become `4.4.1+1`.

eliudio:
> > The fix for this issue happened in version [0.8.4](https://pub.dev/packages/google_sign_in_web#084). Why do you need to do this?
> 
> Because pub.dev treats 0.x dependencies in a special way. See [this](https://dart.dev/tools/pub/versioning#semantic-versions).

I'm confused... Why do I need to do this? 

Well, the dependency in google_sign_in to google_sign_in_web version 0.8.4 is specified in the pubspec.yaml. The version 0.9.0 fixes an issue which causes the compiler to blow up with "Error: JS interop classes do not support non-external constructors. Try annotating with external. GoogleAuth.fakeConstructor$(); ".

So, because google_sign_in depends on google_sign_in_web version 0.8.4, I can not make use of the fix provided in the version 0.9.0. So hence I suggest to upgrade so that google_sign_in uses the latest version of google_sign_in_web, i.e. version 0.9.0, so that this fix can be used in google_sign_in. I'm sorry if I just keep repeating. As said, I'm a bit confused with the question.

ditman:
> I'm confused... Why do I need to do this?

@eliudio I was replying to myself... Actually wondering *why* you needed to do this, since our dependency was `^0.8.0`, which I thought would pick up everything before `1.0.0`, but I was wrong. It'll only pick up things before `0.9.0`. Versions less than `1.0.0` behave differently in pub.

I have a new fix for `google_sign_in_web` that I want to publish ASAP. If you don't mind, I'll take over this PR. Thanks for bringing this to my attention, though!

PR #2656, opened 2020-04-15
[url_launcher] Open url in same window when browser is in standalone mode.

tugorez:
## Description

Users using this plugin on iOS PWAs are experiencing blank screens when a url is opened on another app and then they try to go back. It seems to be an well known issue on iOS [1] and [2] as reference. In the meantime we can overcome this limitation by modifying the target to "_top" according to [2] (solution manually verified). This PR does exactly that, verifies if the user is running a PWA in iOS and if so it will open the url with "_top" as target.
 
[1] https://stackoverflow.com/questions/55517297/mailto-doesnt-work-when-website-launched-from-ios-home-screen/56153685#56153685
[2] https://stackoverflow.com/questions/56220862/ios-pwa-open-maps-returns-to-blank-screen
## Related Issues

flutter/flutter#51461

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/url_launcher/url_launcher_web/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher_web/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 0.1.1+2
 
-- Open urls with target "_top" (instead of default "_blank") on iOS PWAs.
+- Open urls with target "_top" on iOS PWAs.
 
 # 0.1.1+1
 
--- a/packages/url_launcher/url_launcher_web/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher_web/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.1.1+2
+
+- Open urls with "_top" (instead of "_blank") as target on iOS PWAs.
+
 # 0.1.1+1
 
 - Make the pedantic dev_dependency explicit.
--- a/packages/url_launcher/url_launcher_web/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher_web/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 0.1.1+2
 
-- Open urls with "_top" (instead of "_blank") as target on iOS PWAs.
+- Open urls with target "_top" (instead of default "_blank") on iOS PWAs.
 
 # 0.1.1+1
 
--- a/packages/url_launcher/url_launcher_web/lib/navigator.dart
+++ b/packages/url_launcher/url_launcher_web/lib/navigator.dart
@@ -2,12 +2,14 @@
 library navigator.dart;
 
 import 'package:js/js.dart';
+import 'package:meta/meta.dart';
 
 @JS('window.navigator.standalone')
 external bool get _standalone;
 
-/// Utility class to access the window.navigator DOM property.
-class Navigator {
-  /// The window.navigator.standalone DOM property.
-  bool get standalone => _standalone ?? false;
-}
+/// The window.navigator.standalone DOM property.
+bool get standalone => _standalone ?? false;
+
+@visibleForTesting
+@JS('window.navigator.standalone')
+external set standalone(bool enabled);
--- a/packages/url_launcher/url_launcher_web/lib/navigator.dart
+++ b/packages/url_launcher/url_launcher_web/lib/navigator.dart
@@ -6,6 +6,8 @@ import 'package:js/js.dart';
 @JS('window.navigator.standalone')
 external bool get _standalone;
 
+/// Utility class to access the window.navigator DOM property.
 class Navigator {
+  /// The window.navigator.standalone DOM property.
   bool get standalone => _standalone ?? false;
 }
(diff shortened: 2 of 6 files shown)


packages/url_launcher/url_launcher_web/lib/navigator.dart
```diff
@@ -0,0 +1,15 @@
+@JS()
+library navigator.dart;
```
ditman, line 2: You can call this library 'navigator' so you don't have to rename it when you import it later.

Also, please move this file inside `lib/src`, so it can't be accessed by external users of the package.

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
```diff
@@ -19,7 +20,10 @@ class UrlLauncherPlugin extends UrlLauncherPlatform {
   /// Returns the newly created window.
   @visibleForTesting
   html.WindowBase openNewWindow(String url) {
-    return html.window.open(url, '');
+    // We need to open on _top in ios browsers in standalone mode.
+    // See https://github.com/flutter/flutter/issues/51461 for reference.
+    final target = navigator.standalone ? '_top' : '';
```
ditman, line 25: Can you move this logic to the `launch` method, so we can pass it to the `openNewWindow` method?

I'd propose the openNewWindow to have this new signature:

```
html.WindowBase openNewWindow(String url, { String target = ''}) {
  return html.window.open(url, target);
}
```

And you'd call it like so:

```
openNewWindow(url, target: theTargetYouComputed)
```

ditman: _(The reason for this is that some people want to be able to override the attribute manually; so that'd be a new (optional) parameter to the `launch` method. That might have to take into account the standalone computation as well!)_

packages/url_launcher/url_launcher_web/test/url_launcher_web_test.dart
```diff
@@ -54,6 +54,17 @@ void main() {
       expect(newWindow.opener, equals(html.window));
     });
 
+    test('the url is opened in the same window', () {
+      // Simulate the navigator is in standalone mode on iOS devices.
+      // https://developer.mozilla.org/en-US/docs/Web/API/Navigator
+      navigator.standalone = true;
```
ditman, line 60: Since this is modifying global state from the test, please reset it after the test is done... Something like:

```
final originalStandalone = navigator.standalone;
... your test
navigator.standalone = originalStandalone;
```

tugorez: Done!

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
tugorez: I totally agree on allow callers to specify the target but I think this should be handled as a separate thing with it's own issue/pr. What do you think?

packages/url_launcher/url_launcher_web/lib/navigator.dart
tugorez: Done

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
```diff
@@ -4,7 +4,7 @@ import 'dart:html' as html;
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:meta/meta.dart';
 import 'package:url_launcher_platform_interface/url_launcher_platform_interface.dart';
-import 'navigator.dart' as navigator;
+import 'src/navigator.dart' as navigator;
```
ditman, line 7: (No need for the `as navigator` anymore)

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
ditman: OK!

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
tugorez: If I remove the prefix declaration, the analyzer throws:

>  info • Unused import: 'src/navigator.dart' • lib/url_launcher_web.dart:7:8 • unused_import  
error • Undefined name 'navigator' • lib/url_launcher_web.dart:25:20 • undefined_identifier

ditman: Ah, you get the unprefixed getter (you could call `standalone` directly. You're right, this is better :)

packages/url_launcher/url_launcher_web/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # 0.1.1+2
 
-- Open urls with "_top" (instead of "_blank") as target on iOS PWAs.
+- Open urls with target "_top" (instead of default "_blank") on iOS PWAs.
```
ditman, line 3: But the default is `''`, right? :P

tugorez: I might be wrong but most browsers treat "no valid" names (empty string is an instance of this) as "_blank" by default. TBH I could not find a valid reference that enunciated this as standard so I think its safer to only indicate the change we are introducing.

PR #2658, opened 2020-04-16
[video_player_web] Add a custom analysis_options file to video_player_web.

ditman:
## Description

flutter/plugins is currently red because inline analysis checks are no longer honored, as described here: https://github.com/flutter/flutter/issues/52899.

`video_player_web` used a web-only API that was being flagged (correctly) by the analyzer, but suppressed inline via `// ignore` comments.

Now that `// ignore` comments don't work, the tree has gone red.

The fix for `video_player_web` will take longer than acceptable to unblock the flutter/plugins tree, so we're putting this mitigation so work can continue normally.

This patch will be removed, once the affected APIs can be accessed without triggering analyzer errors.

## Related Issues

* https://github.com/flutter/flutter/issues/52899
* https://github.com/flutter/flutter/issues/41563

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/video_player/video_player_web/CHANGELOG.md
+++ b/packages/video_player/video_player_web/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.1.2+2
 
-* Add `analysis_options.yaml` to the package, so we can ignore `undefined_prefixed_name` errors (temporarily).
+* Add `analysis_options.yaml` to the package, so we can ignore `undefined_prefixed_name` errors. Works around https://github.com/flutter/flutter/issues/41563.
 
 ## 0.1.2+1
 
--- a/packages/video_player/video_player_web/CHANGELOG.md
+++ b/packages/video_player/video_player_web/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.1.2+2
+
+* Add `analysis_options.yaml` to the package, so we can ignore `undefined_prefixed_name` errors (temporarily).
+
 ## 0.1.2+1
 
 * Make the pedantic dev_dependency explicit.
--- a/packages/video_player/video_player_web/analysis_options.yaml
+++ b/packages/video_player/video_player_web/analysis_options.yaml
@@ -1,7 +1,7 @@
-# This is a temporary file to allow us to land a new set of linter rules in a
-# series of manageable patches instead of one gigantic PR. It disables some of
-# the new lints that are already failing on this plugin, for this plugin. It
-# should be deleted and the failing lints addressed as soon as possible.
+# This is a temporary file to allow us to unblock the flutter/plugins repo CI.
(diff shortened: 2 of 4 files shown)


ditman:
This will have to be merged on red `submit-queue`. The point of this change is to fix the `analyze` check, which is the one currently broken in the pipeline.

packages/video_player/video_player_web/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.1.2+2
+
+* Add `analysis_options.yaml` to the package, so we can ignore `undefined_prefixed_name` errors (temporarily).
```
ferhatb, line 3: Please add an issue link, TODO: remove ...

packages/video_player/video_player_web/analysis_options.yaml
```diff
@@ -0,0 +1,10 @@
+# This is a temporary file to allow us to land a new set of linter rules in a
```
cyanglaz, line 1: Let's add a TODO to link to https://github.com/flutter/flutter/issues/45440

packages/video_player/video_player_web/CHANGELOG.md
cyanglaz: oh. Just saw this. We can also add the TODO in the anlaysis_options file, @ditman Your call.

cyanglaz: actually, I don't think we should add TODO in the CHANGELOG. let's add to the `analysis_options` file?

mklim: I don't think it makes sense to put a TODO in the changelog, but an issue link here would be fine.

packages/video_player/video_player_web/analysis_options.yaml
mklim: Agreed, but I think the TODO link should be for a bug that's specific to the inline `ignore`s being blanked and specific to this plugin. I think this is a separate issue than the existing bug because it's a new regression, not an existing failure from before the linter was introduced.

cyanglaz: Ah ok, I thought the issue is (at least partially) about removing all analysis_options.
We can make a different issue to fix this specific lint.

ditman: I have created https://github.com/flutter/flutter/issues/55000 for the cleanup.

ditman:
I'll merge as soon as the CI finishes. `analyze` is passing!

ditman:
Merging on red, after verifying `analyze` step (and everything else) now passes. This should unblock the tree.

PR #1044, opened 2019-01-07
[shared_preferences] Added compatibility for existing apps

bluelet13:
This pull request add compatibility of existing apps to the **Shared preferences plugin**.
This provides a refer to the preferences of existing apps.
This is needed for when the migration existing apps to Flutter apps.

I've added:

1. Change SharedPreference name (Only Andrond)
2. Change Key Prefix.

Fixes https://github.com/flutter/flutter/issues/52544

PR #2667, opened 2020-04-20
[shared_preferences] prefix setter and getter

agent3bood:
## Description

Allow users to change the shared_preference key prefix by providing a setter & a getter. 

## Related Issues
https://github.com/flutter/flutter/issues/52544

## Checklist

- [x] I signed the [CLA].

## Breaking Change


agent3bood:
@googlebot I fixed it.

cyanglaz:
Thanks for the PR. It looks good! Could you add some tests?

agent3bood:
@cyanglaz 
After writing more tests I see an issue with the design if this plugin, please help me understand the usefulness of `prefix`.

1. Right now this class is a singleton meaning changing the prefix in one place of the code (like how I did in this PR) will cause it to change on other places which may be undesired behavior.

2. Changing the signature of `getInstance()` to `getInstance({String prefix = '.flutter'})` will cause the same issue as in point 1.

3. The statement [here](https://github.com/flutter/plugins/blob/master/packages/shared_preferences/shared_preferences/lib/shared_preferences.dart#L47) is not correct, because the code will load only the keys with the prefix into memory! 

4. The current code tests by me will fail as it is a race condition, changing the prefix in one test will affect other tests(back to point 1 again).

I suggest a breaking change which is to remove the use of the prefix, if so next I can implement https://github.com/flutter/flutter/issues/35862

[code commit](https://github.com/agent3bood/flutter-plugins/commit/c651febdf09685b7952cd6d221b14a81667615f7)
[test commit](https://github.com/agent3bood/flutter-plugins/commit/4dcce02d88834d327d397fe1ad32cf72563ebc63)

agent3bood:
Code updated, it is possible to set the prefix, but once it is set it cannot be changed. 
This behavior is because this class is a singleton and changing the prefix in one place may cause problems in other places of the code.

stuartmorgan:

We apologize for the long delay in getting back to this PR. We’re in the process of overhauling our PR triage system to respond much more quickly.

stuartmorgan:
In the meantime, it appears that https://github.com/flutter/plugins/pull/1044 is solving the same issue. Perhaps you could coordinate with the author on picking one of the PRs to move forward with, and closing the other?

aleripe (approved):

stuartmorgan:
@aleripe Please see https://github.com/flutter/flutter/wiki/Tree-hygiene#getting-a-code-review ("approvals in the GitHub UI should be reserved for those with contributor access").

jimwan:
@agent3bood  when wil this pr be merged? I am waiting for your fix.

agent3bood:
> @agent3bood  when wil this pr be merged? I am waiting for your fix.

Hi, I am also waiting for this to be merged.
It has to be done by a Google developer.
If this is urgent for you find somebody here  to review it https://flutter.dev/community

pedromassango:
@stuartmorgan @cyanglaz @agent3bood I believe this is different from https://github.com/flutter/plugins/pull/1044. As this one only prefix the keys and it stores all the keys in the same preferences instance which is somewhat bad as you can't clear all the preferences related to an instance only: because it is just a single instance and the data is store in the same file.

I would prefer https://github.com/flutter/plugins/pull/1044 as it implements this feature on both native sides of Android and iOS which solves the issue mentioned above (https://github.com/flutter/plugins/pull/1044 actually write keys in its own preferences file based on the preference instance's name, see https://developer.android.com/reference/android/content/Context#getSharedPreferences(java.lang.String,%20int)).

Let's either update this or continue with https://github.com/flutter/plugins/pull/1044.

stuartmorgan:
> As this one only prefix the keys and it stores all the keys in the same preferences instance which is somewhat bad

I don't see how that's an issue. The use case in the issue is migration from existing non-prefixed preferences.

> as you can't clear all the preferences related to an instance only: because it is just a single instance and the data is store in the same file.

I don't understand what you mean by "instance".

> #1044 actually write keys in its own preferences file based on the preference instance's name

Why does the storage location matter to the developer?

It seems like you are assuming some use case that isn't what's in the issue and which you haven't described, so I'm not following what the concern is.

pedromassango:
> It seems like you are assuming some use case that isn't what's in the issue and which you haven't described, so I'm not following what the concern is.

Agreed, I misunderstood this PR, sorry about that.

> Why does the storage location matter to the developer?

Unrelated to this PR, but there are some cases where we need to store keys in separated files (supported on Android nativelly, not sure about iOS).

agent3bood:
@stuartmorgan  @pedromassango As I said before, this PR need a review from Google employee, I have fixed the conflicts again and if more test cases are needed I am happy to add it.

stuartmorgan:
Since there hasn't been any update here, I'm going to close this. If you're interested in moving forward with the approach described in the link I posted above, please open a new PR and we can get that reviewed.

PR #2669, opened 2020-04-20
[webview_flutter] Bump the Dart SDK requirement to 2.7.0

amirh:
A recent `pub publish` check warns when requiring a non release Dart SDK, bumping the requirement so we don't get the warning.

This change is a no-op as the minimally required Flutter version (`1.12.13+hotfix.5`) already requires Dart `2.7.0`.


amirh:
I'm going to ignore the version_checker complaints as I'm actually doing this to unblock publishing `0.3.20+1`.

PR #2703, opened 2020-04-22
[video_player] Update README to mention web support.

mariamhas:
updates to call out web is also supported for this plugin.
Explicitly tells users not to use the constructor when using the plugin for web.

## Related Issues
flutter/flutter#55340

## Breaking Change

- [≈ ] No, this is *not* a breaking change.


ditman:
(Merging on yellow to trigger a cirrus rebuild)

ditman:
Ooops, my bad, forgot to mention that we need to bump the version and the CHANGELOG in order for this to be publishable to pub.dev!

PR #2714, opened 2020-04-28
Add playback speed support

tancor:


phanirithvij:
@tancor why isn't this reviewed?

PR #2740, opened 2020-05-07
[url_launcher] Launches only mailto urls in same window on iOS devices

tugorez:
## Description 
Previously on https://github.com/flutter/plugins/pull/2656 we started launching all urls on the _top context (only on iOS PWA). It seems that this was a partial fix because non-PWA web apps are having a similar issue: after opening a mailto url and trying to get back to the app's context they are seeing a blank window. This behaviour is caused by the browser that opens a new tab, detects the url is one that could be handled by an email application and finally opening the email application without closing that new opened tab (leaving it with blank content). This PR assumes three things:

1. This problem only happens on iOS devices
1. This problem only happens with mailto urls
1. This problem is fixed by forcing the browser to open the mailto urls on the current window, on the _top context.

## Related Issues

flutter/flutter#51461

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
```diff
@@ -4,12 +4,12 @@ import 'dart:html' as html;
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:meta/meta.dart';
 import 'package:url_launcher_platform_interface/url_launcher_platform_interface.dart';
-import 'src/navigator.dart' as navigator;
 
 /// The web implementation of [UrlLauncherPlatform].
 ///
 /// This class implements the `package:url_launcher` functionality for the web.
 class UrlLauncherPlugin extends UrlLauncherPlatform {
+  static final _iosPlatforms = ['iPhone', 'iPad', 'iPod'];
```
tugorez, line 12: Could we do the same validation with https://api.flutter.dev/flutter/foundation/TargetPlatform-class.html?

tugorez: We can use a set

tugorez:
Please take in account that this issue seems to be happening only on iOS therefore I tried to change the behaviour only on those devices. To achieve that I had to rely on `window.navigator.platform`. My only concern is that I couldn't find an standar nor an official resource that listed all posible values. The best resource I could find was [this stackoverflow answer][1].

[1]: https://stackoverflow.com/questions/19877924/what-is-the-list-of-possible-values-for-navigator-platform-as-of-today

ditman:
(I've sent tugorez some changes in a separate PR to this branch)

PR #2743, opened 2020-05-08
reference apple sign in plugin from google sign in plugin

xster:


xster:
ah you're right, I still need to pub publish it. Done

PR #2755, opened 2020-05-13
[image_picker] fixes for iOS which doesn't present camera/albums with more complex navigation

chris-rutkowski:
## Description

Existing implementation is incorrect especially when adding project as a module with more complex presentation structure.
If window's root view controller has presented modally any other view controller, it cannot present any other any more. Only the top most controller can present other.

With this change, the image picker/camera controller is presented on the top most view controller.

## Related Issues

I found this issue raised by someone already:
https://github.com/flutter/flutter/issues/34787

## Checklist

- [x] I signed the [CLA].

## Breaking Change


chris-rutkowski:
@xster @mehmetf @cyanglaz @blasten I saw you merged some other PRs related to image picker recently, that's why I'm @-ing you. Is there anything more I can do here in this PR?

packages/image_picker/image_picker/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: image_picker
 description: Flutter plugin for selecting images from the Android and iOS image
   library, and taking new pictures with the camera.
 homepage: https://github.com/flutter/plugins/tree/master/packages/image_picker/image_picker
-version: 0.6.6+3
+version: 0.6.7
```
cyanglaz, line 5: there is no public api change, we should only update the patch -> 0.6.6+4

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -25,33 +25,29 @@ @interface FLTImagePickerPlugin () <UINavigationControllerDelegate, UIImagePicke
...
-  }
-  return self;
-}
-
 - (UIImagePickerController *)getImagePickerController {
   return _imagePickerController;
 }
 
+- (UIViewController *)viewController {
+  UIViewController *topController = [UIApplication sharedApplication].keyWindow.rootViewController;
```
cyanglaz, line 44: this is great! By the way, while we are here, can we also make it work for navigation controller? (pushed viewcontrollers)

chris-rutkowski: this solution will work in this case as well.
According to the [documentation](https://developer.apple.com/documentation/uikit/uiimagepickercontroller) you shouldn't push UIImagePickerController, but present it modally instead.

So in the situation that FlutterViewController is pushed into an existing navigation controller, e.g.:

```
Window
  rootViewController
    someOtherViewController
      navigation controller (with some native view controller and FlutterVC)
         image picker controller 

not

Window
  rootViewController
    someOtherViewController
      navigation controller (with some native view controller and FlutterVC, image picker controller)
          
```

I was worried about iPad which for a long time required popover presentation for albums, but since iOS 13 it presents modally as well without an issue.

chris-rutkowski:
@cyanglaz as requested changed version to 0.6.6+5

chris-rutkowski:
@cyanglaz any more feedback? I've already updated changelog a few times due to conflicts - other PRs being merged.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
chris-rutkowski: @cyanglaz the solution I made will in all cases you are concerned about.

UINavigationController, UITabBarController they both manage internally some view controllers, **but** they themselves inherit from UIViewControllerr and either of them can be window's root view controller, or presented on some existing View Controller of any type.

The topViewController cannot be pushed in the navigation stack. In that case, UINavigationCoontroller is topViewController and some other push view controller is managed by UINavigationController. And with that arrangement, my solution will correctly display the Image Picker on top of the navigation controller.

**Trust me** - in my company we have already very complex navigation - soome legacy Drawer Controller (in iOS), multiple Navigation controllers, TabBarControllers and Flutter V.C. is displayed on top of all of this. And my solution has been proven successful.

aboutlan:
@cyanglaz I have the same problem.
I presented flutterViewController after I use present image_picker doesn't work, Xcode error warning: Attempt to present <UIImagePickerController: 0x7fbbb1856800> on <UINavigationController: 0x7fbbb201c000> whose view is not in the window hierarchy !. 
chris-rutkowski solved my problem

diegotori:
@cyanglaz a gentle reminder for reviewing this PR, since this could be very useful to have.

@chris-rutkowski looks like your branch has conflicts that need to be resolved before this can merge. Thank you in advance for this PR.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
cyanglaz: according to https://developer.apple.com/documentation/uikit/uiviewcontroller/1621407-presentedviewcontroller?language=objc the `presentedViewController` is only the top view controller when it is presented with `presentViewController:animated:completion:`. 
What if the top view controller is presented by [pushViewController](https://developer.apple.com/documentation/uikit/uinavigationcontroller/1621887-pushviewcontroller?language=objc)? Does this still work?

chris-rutkowski: @cyanglaz I am already repeating myself, but let me rephrase.

`presentViewController` is a method of `UIViewController`
UINavigationController, UITabBarController, MySweetViewController - they are all subclasses of UIViewController, therefore they can be presented on any other view controller, or some other view controller can be presented on top of them.

The top view controller **presented** by push view controller - such a thing doesn't exist.
You can have UINavigationController either as a root of the window, or presented on some other view controller. And Navigation Controller manages the stack of view controllers. These view controllers shouldn't present any other view controller (because they are managed by NC). Instead, you either push another VC in NC (horizontal), or present VC on top of NC (vertical).

Moreover - if UINavigationController is one of the children of UITabBarController, you shouldn't present any VC on top of that NC, you should present on top of UITabBarController instead.

cyanglaz: Ok. I'm interested in this case you presented 
```
example2:
window
  navigation controller (home, details)
     image picker
```

In this case, when you go through the code, what is the top view controller? Is it the `navigation controller`? 
And the image picker is going to use the navigation controller as the "viewController", regardless which view controller is the on the top of the stack of the navigation controller? Am I understanding correctly?

chris-rutkowski: exactly, image picker will be correctly presented on navigation controller. The contents of navigation controller doesn't matter at all.

chris-rutkowski:
@cyanglaz added unit test and improved the solution a bit.

BTW I cannot find any guide to run iOS tests. I had to create a dummy project which wasn't very easy.

cyanglaz:
@chris-rutkowski 
try
```
cd ios
pod lib lint
```

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.h
```diff
@@ -7,7 +7,7 @@
 @interface FLTImagePickerPlugin : NSObject <FlutterPlugin>
 
 // For testing only.
```
cyanglaz, line 9: Should we remove this line now?

chris-rutkowski: I don't think so? Both of these methods here are exposed only for testing and not meant to be used in normal circumstances.

PR #2763, opened 2020-05-14
[webview_flutter] Add support for passing a failing url

bparrishMines:
## Related Issues

[[WKNavigationDelegate webView:didFailNavigation:withError:]](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455623-webview?language=objc) doesn't directly provide this value. It provides a `WKNavigation` which can only be used to track a call to `loadRequest`.

But [WebViewClient.onReceivedError](https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedError(android.webkit.WebView,%20int,%20java.lang.String,%20java.lang.String)) does on Android.

internal bug b/151727643

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
```diff
@@ -223,18 +226,19 @@ public void onPageFinished(WebView view, String url) {
 
       // This method is only called when the WebViewFeature.RECEIVE_WEB_RESOURCE_ERROR feature is
       // enabled. The deprecated method is called when a device doesn't support this.
+      @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
```
cyanglaz, line 229: what happens if the method is called on a lower api version?

packages/webview_flutter/lib/platform_interface.dart
```diff
@@ -139,6 +140,12 @@ class WebResourceError {
   ///
   /// This will never be `null` on Android, but can be `null` on iOS.
   final WebResourceErrorType errorType;
+
+  /// Gets the URL for which the resource request was made.
+  ///
+  /// This value is not provided on iOS. Alternatively, you can keep track of
+  /// the last values provided to [WebViewPlatformController.loadUrl].
```
cyanglaz, line 147: I wonder if we can just do that for users on iOS.

packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
bparrishMines: The comment above should explain the expected behavior. If this method isn't supported, it should call the deprecated method below. This is according to: https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedError(android.webkit.WebView,%20android.webkit.WebResourceRequest,%20android.webkit.WebResourceError)

packages/webview_flutter/lib/platform_interface.dart
bparrishMines: Hmm.. I considered doing this as well. I don't know if we want to provide a value that the native code doesn't provide just because the other platform does.

Also, Android can provide a URL for any resource, not just the main page. According to [onReceivedError]( https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedError(android.webkit.WebView,%20int,%20java.lang.String,%20java.lang.String)).

 iOS could only provide the last URL used and could return the wrong value if the user navigated to a new webpage without calling `loadUrl`.

PR #2764, opened 2020-05-14
Run the publish with the pub version from flutter master

cyanglaz:
I believe all the plugins are migrated and can not check publish on master, and the workaround in https://github.com/flutter/plugins/pull/2250 can be reverted.
This will also enable us to catch potential publish issues with newer dart version as early as possible.


cyanglaz:
Done in https://github.com/flutter/plugins/pull/2764

PR #2789, opened 2020-05-21
[path_provider] linux endorsement

TimWhiting:
## Description

This PR updates path_provider to endorse the dart Linux path provider manually, until automatic registration of dart plugins are working.

## Related Issues

flutter/flutter#41716

## Checklist

- [x] I signed the [CLA].

## Breaking Change

Only to remove manual registration if anyone does that before this gets implemented.


TimWhiting:
@stuartmorgan 
FYI Here is the PR for the endorsement of the Linux path provider. 

Note that I need to manually register the instance, at least until flutter tool support for automatically generating a plugin registrant for dart only plugins happens. If that is in the plans? (I didn't see any conclusion on which approach you are planning on taking) [Design Document: Flutter tool support for federated plugins](https://docs.google.com/document/d/1G_K5vTMWD6CBD-0DDWhpAwkcCRKqmPH1wxlXu5FfRhk/edit#heading=h.pub7jnop54q0)

stuartmorgan:
> Note that I need to manually register the instance, at least until flutter tool support for automatically generating a plugin registrant for dart only plugins happens. If that is in the plans?

Ideally yes, but it ended up having technical complexities, so we decided to go with this approach in the short-to-medium term to unblock being able to implement Dart-only plugins. https://github.com/flutter/flutter/issues/52267 is tracking a longer term solution.

TimWhiting:
@stuartmorgan 
Ready for review

packages/path_provider/path_provider/test/path_provider_test.dart
```diff
@@ -25,6 +25,9 @@ void main() {
 
     setUp(() async {
       PathProviderPlatform.instance = MockPathProviderPlatform();
+      // This is required because we manually register the Linux path provider when on the Linux platform.
+      // Will be removed when automatic registration of dart plugins is implemented.
```
stuartmorgan, line 29: Please link to https://github.com/flutter/flutter/issues/52267 in any comment about workarounds so that it's easy for someone without full context to trace back to the relevant discussion and current state.

packages/path_provider/path_provider/lib/path_provider.dart
```diff
@@ -3,14 +3,36 @@
...
+
+bool _useMethodChannel = false;
+PathProviderPlatform __platform;
+
+// This is to manually endorse the linux path provider until automatic registration of dart plugins is implemented.
+PathProviderPlatform get _platform {
+  if (__platform != null) {
+    return __platform;
+  }
+  if (Platform.isLinux && !_useMethodChannel) {
```
stuartmorgan, line 29: Do we need a `kIsWeb` check first? I'm not sure what `Platform.isLinux` returns when running Flutter for Web on Linux.

packages/path_provider/path_provider/lib/path_provider.dart
```diff
@@ -3,14 +3,36 @@
...
+// Should not be used in users' code, just made visible for testing to test the method channel implementation
+@visibleForTesting
+set testOverrideUseMethodChannel(bool override) {
+  _useMethodChannel = override;
+}
+
+bool _useMethodChannel = false;
+PathProviderPlatform __platform;
+
+// This is to manually endorse the linux path provider until automatic registration of dart plugins is implemented.
```
stuartmorgan, line 24: Per [my comment here](https://github.com/flutter/plugins/pull/2631#discussion_r406489237) I'm not sure if we want this hidden; there are legitimate reasons some could want to call this.

@tvolkert Who is the right person to weigh in on this at the moment with amirh OOO?

packages/path_provider/path_provider_linux/README.md
```diff
@@ -2,36 +2,29 @@
...
 
 ## Backward compatible 1.0.0 version is coming
 The `path_provider` plugin has reached a stable API, we guarantee that version `1.0.0` will be backward compatible with `0.0.y+z`.
-Please use `path_provider_linux: '>=0.0.y+x <2.0.0'` as your dependency constraint to allow a smoother ecosystem migration.
+Please use `path_provider: '>=0.0.y+x <2.0.0'` as your dependency constraint to allow a smoother ecosystem migration.
 For more details see: https://github.com/flutter/flutter/wiki/Package-migration-to-1.0.0
 
 ## Usage
 
 ### Import the package
```
stuartmorgan, line 14: It seems odd to have the section be called "import the package" when that's not actually necessary.

I think this whole thing could just become:
```
## Usage

This package is already included as part of the `path_provider` package dependency, and will
be included when using `path_provider` as normal. You will need to use version 1.6.10 or newer.
```

packages/path_provider/path_provider_linux/example/test/widget_test.dart
```diff
@@ -9,23 +9,20 @@ import 'dart:async';
...
       await WidgetsFlutterBinding.ensureInitialized();
-      PathProviderPlatform.instance = PathProviderLinux();
     });
 
     testWidgets('Finds tmp directory', (WidgetTester tester) async {
       // Build our app and trigger a frame.
       await tester.runAsync(() async {
         await tester.pumpWidget(MyApp());
-        await Future.delayed(Duration(milliseconds: 10));
+        await Future.delayed(Duration(milliseconds: 20));
```
stuartmorgan, line 25: Why was this necessary? Does this take more than 10 milliseconds to run?

stuartmorgan-g (changes requested):

packages/path_provider/path_provider_linux/example/test/widget_test.dart
TimWhiting: The tests were being flaky, for some reason they were failing on CI. As there are no logical flaws in the test or the app code as far as I can tell I assumed the asynchronous code in the `initState` of `MyApp` that calls out to the plugin to get the various paths had not yet been run. Increasing the timeout seemed to solve the issue. Is there a better way to make sure all asynchronous events are finished being handled in a test?

packages/path_provider/path_provider/lib/path_provider.dart
TimWhiting: Renamed the platform override and made it user visible in my latest commit, I doubt that people will be using their own registered version of `path_provider` unless they are using a different embedding i.e. `go-flutter`, but it will mean that those implementations will have to add `disablePathProviderPlatformOverride = true` until we have automatic registration.

TimWhiting: @stuartmorgan 
I've made this user visible, and renamed / reformatted it some. Should we continue waiting for a reply? The only issue would be that removing it in the future would be technically a breaking change for anyone who is using it.

stuartmorgan: I was thinking about this last night, and was going to updating this morning actually. Since getting resolution on this is proving difficult, let's return it to being test-only. That means we're not adding to the public API surface, so it won't need the same level of discussion/approval. We can always add it later if it becomes necessary, without it blocking getting the initial implementation landed.

TimWhiting:
@stuartmorgan 
Just made that setter visible for testing, and updated the dartdoc for it. Let me know if further adjustments should be made. I tried to not change the dartdoc too much, in case we want to go back to being user visible.

I think this should be ready to land, once all the tests pass.

stuartmorgan-g (approved):
LGTM!

TimWhiting:
@stuartmorgan 
Do I need to merge in master with that latest merged pull request or can you merge this in without that?

stuartmorgan:
There aren't an conflicts; no need to re-merge.

stuartmorgan:
Once main CI has cycled green I should be able to publish the update.

stuartmorgan-g:
path_provider 1.6.10 is live 🎉

stuartmorgan-g:
I filed https://github.com/flutter/website/issues/8442 for updating the cookbook. Updating the path_provider docs to call out the necessary changes sounds good. Given that it's already published, I don't think there's much more we can do to mitigate the potential unintentional test breakage.

PR #2802, opened 2020-05-29
[image_picker_for_web] Introduce image_picker_for_web package

ditman:
## Description

This is a first version of an Image Picker plugin for web. 

Since dart:io File is not available on the web, this only implements the new `PickedFile` methods from the platform interface.

Instead of storing a "normal" path value on the file, however, it stores an "object URL" (see `createObjectUrl`) that the browser keeps track of. The PickedFiles in the web platform can be accessed through a network request (not a file operation).

## Demo

https://dit-picker-tests.web.app

## Known issues

* Is there any opportunity to add `revokeObjectUrl` to free resources in the browser?
* **Safari is misbehaving** 
  * with videos, [it should work](https://bl.ocks.org/unRob/3bd07a012597aa959c92) (this might be a bug in the VideoPlayer plugin for web). (Issue: https://github.com/flutter/flutter/issues/58527)
  * ~~taking pics on an iPhone: the aspect ratio is lost, and every pic seems to be rendered in "landscape"~~ (Can't reproduce anymore)
  * In general, **only videos that can be opened directly by the browser can be previewed**. There's no client-side transcoding of any kind. Not sure if this is detectable, but it'd be a nice exception to throw! (Issue: https://github.com/flutter/flutter/issues/58532)

## Related Issues

* https://github.com/flutter/flutter/issues/36281
* https://github.com/flutter/flutter/issues/53283

## Checklist

- [x] I signed the [CLA].

## Breaking Change

- [x] No, this is an optional plugin, and users don't need to use it if they don't want to.


packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+  // DOM methods
+
+  /// Converts plugin configuration into a proper value for the `capture` attribute.
+  ///
+  /// See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#capture
+  @visibleForTesting
+  String computeCaptureAttribute(ImageSource source, CameraDevice device) {
+    String capture;
+    if (source == ImageSource.camera) {
+      capture = device == CameraDevice.front ? 'user' : 'environment';
```
yjbanov, line 80: Do you need the `capture` variable? Can you just `return device == CameraDevice.front ? 'user' : 'environment';`? Similarly `else { return null; }`.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+    }
+    return capture;
+  }
+
+  html.File _getFileFromInput(html.FileUploadInputElement input) {
+    if (_hasOverrides) {
+      return _overrides.getFileFromInput(input);
+    }
+    return input.files[0];
```
yjbanov, line 89: I believe `.first` is more idiomatic than `[0]` (provided there is such a getter).

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /// Handles the OnChange event from a FileUploadInputElement object
+  /// Returns the objectURL of the selected file.
+  String _handleOnChangeEvent(html.Event event) {
+    // load the file...
```
yjbanov, line 95: remove ...?

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+    final html.FileUploadInputElement input = event.target;
+    final html.File file = _getFileFromInput(input);
+
+    if (file != null) {
```
yjbanov, line 99: Is it possible to get null? If there are no files, you'll get `RangeError` I think.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+
+  /// Monitors an <input type="file"> and returns the selected file.
+  Future<PickedFile> _getSelectedFile(html.FileUploadInputElement input) async {
+    // Observe the input until we can return something
+    final Completer<PickedFile> _completer = Completer<PickedFile>();
+    input.onChange.listen((html.Event event) async {
+      final objectUrl = _handleOnChangeEvent(event);
+      _completer.complete(PickedFile(objectUrl));
+    });
+    input.onError // What other events signal failure?
```
yjbanov, line 113: Should the comment be formatted as a TODO?

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+    return null;
+  }
+
+  /// Monitors an <input type="file"> and returns the selected file.
+  Future<PickedFile> _getSelectedFile(html.FileUploadInputElement input) async {
+    // Observe the input until we can return something
+    final Completer<PickedFile> _completer = Completer<PickedFile>();
+    input.onChange.listen((html.Event event) async {
+      final objectUrl = _handleOnChangeEvent(event);
+      _completer.complete(PickedFile(objectUrl));
```
yjbanov, line 111: If the completer can be completed with an error, maybe check `if (_completer.isCompleted)` prior to completing it?

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+    input.onError // What other events signal failure?
+        .listen((html.Event event) {
+      _completer.completeError(event);
+    });
+
+    return _completer.future;
+  }
+
+  /// Initializes a DOM container where we can host input elements.
+  html.Element _initTarget(String id) {
```
yjbanov, line 122: Based on the logic inside this method, typically we use `ensureInitialized` nomenclature, because it does not always initialize, only the first time.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+  html.Element createInputElement(String accept, String capture) {
+    if (_hasOverrides) {
+      return _overrides.createInputElement(accept, capture);
+    }
+
+    html.Element element;
+
+    if (capture != null) {
+      // Capture is not supported by dart:html :/
+      element = html.Element.html(
```
yjbanov, line 146: I'd use JS-interop to set the capture attribute. Passing node validators results in unnecessary security reviews later on.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+  /// Injects the file input element, and clicks on it
+  void _injectAndActivate(html.Element element) {
+    _target.children.clear();
+    _target.children.add(element);
+    element.click();
+  }
+}
+
+// Some tools to override behavior for unit-testing
+typedef _OverrideCreateInputFunction = html.Element Function(
```
yjbanov, line 166: These should be public as well if they are visible to tests. It's generally an antipattern for a public interface to vent private types. `@visibleForTesting` is a fair game though.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Ah, you're right, great point. I'll use .first instead.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: AFAIK the browser prevents you from "accepting" the dialog without selecting any files, and if the user "cancels" out, you don't even get an event (that's why the DOM is rebuilt on each "click", so the state of the picker gets fully reset in case the user didn't pick a file before.)

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: It Should Never Be completed with an error *and then* a success, but yeah, I'll add an isCompleted check here.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: That's 2 writes to the DOM, instead of 1, but I'll give this a shot (since I have the html.Element, I should be able to set the capture attribute easily).

Thanks for the insight on the extra security reviews, I don't want to be involved in those unless they're really warranted :P

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Since neither the end user nor the tests need this as a left-hand assignment on their own, I left them private, but I can make them public and @visibleForTesting. At least that way they'll have documentation :)

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Ah man, I always forget about those little convenience getters :P

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Yep, can do.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
yjbanov: Right, but if I'm reading `_getFileFromInput` correctly, if `input.files` is empty, the method will throw `RangeError`. Maybe `_getFileFromInput` should be changed to not throw, and return null instead?

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
yjbanov: Yeah, I know that the analyzer leaks the private type to the call site in this case, but it still feels weird to vend private types publicly.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
...
+      return html.Url.createObjectUrl(file);
+    }
+    return null;
+  }
+
+  /// Monitors an <input type="file"> and returns the selected file.
+  Future<PickedFile> _getSelectedFile(html.FileUploadInputElement input) async {
+    // Observe the input until we can return something
+    final Completer<PickedFile> _completer = Completer<PickedFile>();
+    input.onChange.listen((html.Event event) async {
```
ferhatb, line 109: do you need to remove listener on dispose? if not add comment on why not.

packages/image_picker/image_picker_for_web/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+However, the returned `path` attribute of the `PickedFile` points to a `network` resource, and not a
+local path in your users' drive. See **Use the plugin** below for some examples on how to use this
+return value in a cross-platform way.
+
+### input file "accept"
+
+In order to filter only video/image content, some browsers offer an [`accept` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept) in their `input type="file"` form elements:
+
+![Data on support for the input-file-accept feature across the major browsers from caniuse.com](https://caniuse.bitsofco.de/image/input-file-accept.png)
```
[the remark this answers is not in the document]
ditman: What I linked is an image that self-updates as caniuse.com changes (or it should!): 

Data on support for the input-file-accept feature across the major browsers from caniuse.com

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Removed the whole comment, it was left over from the time when this implementation *did* load the bytes from the file.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Ahhh gotcha, let me add some `?`s to `_getFileFromInput`

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Ah no, this ended up being mmmmuch cleaner than it was before:

```dart
    html.Element element = html.FileUploadInputElement()..accept = accept;
    
    if (capture != null) {
      element.setAttribute('capture', capture);
    }
```

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Yes, Ferhat suggested the same. I'm going to create some issues in Github to track these TODOs.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Good point. We don't need to detach, because the input gets re-created every time the user tries to pick a file, but we also don't care about the whole lifecycle of the stream, just the first event.

I'm going to listen only to the first event, instead of subscribing to the complete stream.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Created 2 tickets for the currently known issues. Removed this comment.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -0,0 +1,182 @@
+import 'dart:async';
+import 'dart:html' as html;
+
+import 'package:flutter_web_plugins/flutter_web_plugins.dart';
+import 'package:meta/meta.dart';
+import 'package:image_picker_platform_interface/image_picker_platform_interface.dart';
+
+final String _kImagePickerInputsDomId = '__image_picker_web-file-input';
+final String _kAcceptImageMimeType = 'image/*';
+// This may not be enough for Safari.
```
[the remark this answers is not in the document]
ditman: Created https://github.com/flutter/flutter/issues/58532 and converted the above in a proper TODO.

ditman:
I think I addressed all the comments, @ferhatb, @yjbanov, PTAL if you have some time, thanks!

ditman:
> Mostly LGTM although I would prefer to return an `ImageProvider` rather than a `PickedFile`. This way, the developer could just do `Image(pickedImage)` whether you are on web or not.
> 
> See: https://api.flutter.dev/flutter/widgets/Image/Image.html

@hterkelsen I looked a little bit into this, and the issue I encountered was that only Images can be created that way; there's no concept of VideoProvider, AudioProvider, "AssetProvider" or FileProvider (AFAIK), so that solution would only cover a fraction of the use cases of the plugin (I didn't measure what's the actual breakdown of usage, it might be a large fraction of the use cases :P)

_PS: Now that web and desktop are coming, there's already feature requests of "let me pick any arbitrary file from my disk._

In any case, I think that's a very good idea for an extra API method down the line; if you're only using the plugin for Images, we can add a "PickedImageProvider" class that lets you create Image widgets directly! (It'd still require dart:io vs dart:html implementations, but they'd be doable). **Do you think we need an issue to track that task (P5/P6?), or wait until some user asks for something like that explicitly?**

hterkelsen:
Ah, that makes sense. I don't think we need to add it in that case

ditman:
Published to pub.dev:

https://pub.dev/packages/image_picker_for_web (^0.1.0)

parammittal162:
hi I am stuck in some issues regarding file validation, i.e. how can we check is user has uploaded only image extension file on web, because user can select option "All files" and upload any file on web.

```
if (kIsWeb) {
  Image.network(pickedFile.path);
} else {
  Image.file(File(pickedFile.path));
}
```

like here for web if I console the Image I am getting - 

Image(image: NetworkImage("blob:http://0.0.0.0:55555/316c9cde-1333-4294-a72d-03e45d97f5f7", scale: 1), frameBuilder: null, loadingBuilder: null, alignment: center, this.excludeFromSemantics: false, filterQuality: low)

_blob:http://0.0.0.0:55555/316c9cde-1333-4294-a72d-03e45d97f5f7_ 

does not contains any file extension info or meta data kind something ?? Please Help ...

ditman:
@parammittal162 the only reliable way of doing what you want is either doing it via server-side, or by looking at the bytes of the `PickedFile` and trying to recognize the header of the image.

You would be in a similar situation if your user renames "file.txt" to "file.jpg" and tries to pick it.

parammittal162:
ok Thanks @ditman

markgrancapal:
> @parammittal162 the only reliable way of doing what you want is either doing it via server-side, or by looking at the bytes of the `PickedFile` and trying to recognize the header of the image.

This is my problem too right now. I've used the readAsBytes() function to show the image using Image.memory(), but the problem is I want to upload the file to the server. I don't know how to get the extension of the file by reading the bytes, any help on this? Is there a way for the plugin to give this info? like a the whole filename. We can't use path extension() method also because it's in web .

andrea689:
@markgrancapal you can use [findDecoderForData()](https://pub.dev/documentation/image/latest/image/findDecoderForData.html) of package [image](https://pub.dev/packages/image)

markgrancapal:
@andrea689 thanks for the help!

ditman:
We probably need to add a file `metadata` attribute to the PickedFile object. It'll be redundant in mobile, but useful for web. Thanks for the feedback!!

m-j-g:
@ditman thanks for all your work on this.  I've been messing with 0.6.7+1 and converted my implementation.  I switched to bytes so that it would work on both web and mobile.
 final _picker = ImagePicker();
    try {
      final pickedFile = await _picker.getImage(
        source: ImageSource.gallery,
        maxWidth: 400,
        maxHeight: 400,
        imageQuality: 80,
      );
      final result = await pickedFile.readAsBytes();

      bloc.updateImageSelection(result);
    } catch (e) {
      Log.d(e.toString());
    }

However, I never make it to an image selection.  As soon as I run this code I receive >
MissingPluginException(No implementation found for method pickImage on channel plugins.flutter.io/image_picker)

I'm assuming that I missed something, and tips would be appreciated.

ditman:
**"No implementation found for method pickImage on channel plugins.flutter.io/image_picker"**

@m-j-g  The `image_picker_for_web` plugin is still *not* officially endorsed, so it will not be added to your dependencies automatically. You probably need to update your `pubspec.yaml` and depend on it, like so:

```
image_picker: ^0.6.7
image_picker_for_web: ^0.1.0
```

Once the plugin has been adopted by more people and we iron out whatever issues/features come up, we'll endorse it, so you just need to depend on a recent version of `image_picker`. For now, please add the dependency on the web version to your app!

Thanks for giving this a shot!

_PS: [check the README in the web version of the plugin](https://pub.dev/packages/image_picker_for_web), it has documentation about installation, browser support, etc..._

m-j-g:
@ditman 
ahh thanks.  I kept seeing that package, but my mind dismissed it as transitive.  Thanks for clearing that up!

ditman:
@m-j-g makes sense, almost every other plugin "hides" the web implementation so you don't have to worry about it. Note that every web plugin has (or should have) a similar README, with installation instructions and "quirks" of the web platform.

markgrancapal:
> @markgrancapal you can use [findDecoderForData()](https://pub.dev/documentation/image/latest/image/findDecoderForData.html) of package [image](https://pub.dev/packages/image)

It worked! Thanks! But now I need to know the file extension for videos, and I can't find any suitable package for this one. Any suggestion? Thank you!

markgrancapal:
> We probably need to add a file `metadata` attribute to the PickedFile object. It'll be redundant in mobile, but useful for web. Thanks for the feedback!!

a way to know the filename or even the extension would be a big help for sure. 

[image_picker_web](https://pub.dev/packages/image_picker_web) does this but the problem is it doesn't provide a network path. For videos I can't find a way to play a video using Byte data or Html.File

shujaatak:
@ditman Can you please share the link of the source code of the Demo:
https://dit-picker-tests.web.app

ditman:
@shujaatak it should be this: https://github.com/flutter/plugins/tree/master/packages/image_picker/image_picker/example

nonnyzcsrt:
@ditman I used method pickImage found that image not resizing and compressing follow maxWidth maxHeight and imageQuality. Please suggest me. Thank you
Screen Shot 2563-10-06 at 00 49 34

ditman:
@nonnyzcsrt resizing on the fly is not supported on web, for now. You might be able to load the image bytes in a canvas and then do the resizing, but this is not natively supported!

postacik:
If someone ever gets here before resizing the image is implemented in the package, you can use the following function to resize an image in the browser:

```
import 'dart:html' as html;

  static Future<String> resizeImage(String src, int maxWidth, int maxHeight, int imageQuality) {
    var completer = Completer<String>();
    var img = html.ImageElement();
    img.onLoad.listen((event) {
      var canvas = html.CanvasElement();
      var ctx = canvas.context2D;
      double ratio = 1;
      print("Original size: ${img.width} x ${img.height}");
      if (img.width > img.height) {
        if (img.width > maxWidth) ratio = maxWidth / img.width;
      } else if (img.height > maxWidth) {
        ratio = maxWidth / img.height;
      }
      if (img.width > img.height) {
        if (img.width < maxHeight) ratio = maxHeight / img.width;
      } else if (img.height < maxHeight) {
        ratio = maxHeight / img.height;
      }
      print("Resizing $ratio times");
      canvas.height = (img.height * ratio).floor();
      canvas.width = (img.width * ratio).floor();
      // Draw the image to canvas and resize
      ctx.drawImageScaled(img, 0, 0, canvas.width, canvas.height);
      var base64 = canvas.toDataUrl("image/png", imageQuality / 100);
      completer.complete(base64);
    });
    img.src = src;
    // make sure the load event fires for cached images too
    if (img.complete) {
      // Flush cache
      img.src = 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///ywAAAAAAQABAAACAUwAOw==';
      // Try again
      img.src = src;
    }
    return completer.future;
  }

```

ditman:
@postacik this is a cool example, why not add it to the plugin code itself?

postacik:
@ditman , I am not familiar with the flutter plugins structure. It would be great if someone else could add it to the plugin.

jesusrp98:
I'll do @ditman @postacik.
I'll open a PR later today containing this code!

jesusrp98:
The thing that I was thinking about is that the code above is only meant for. Would it be better to make a code that runs on any platform??

postacik:
```
    final _picker = ImagePicker();
    final PickedFile pickedFile = await _picker.getImage(source: ImageSource.gallery, maxWidth: 400, maxHeight: 400, imageQuality: 80);
```

If this code is run on iOS or Android it resizes the selected image with the given size parameters and image quality.

However on web the maxWidth, maxHeight and imageQuality parameters have no effect. The image is returned as is.

If you add this code to the web platform implementation, it will be behaving just like the mobile platforms.

@ditman , please correct me if I'm wrong.

ditman:
@jesusrp98 @postacik yes, I think the native picker is already able to resize images both on iOS and Android. This is a missing feature of Flutter web.

Please @jesusrp98, remember to create a Flutter issue for [image_picker_for_web] that explains what you're going to do :)

@postacik the code for picking images in web is here, in case you want to lend a hand to @jesusrp98 :) https://github.com/flutter/plugins/blob/master/packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart

postacik:
By the way, ImageSource.camera has no effect on desktop. It works only on mobile.

WhatsApp web uses browser WebRTC functions to capture image from webcam. 

There's a flutter library called flutter_webrtc which has web support and can capture image from webcam.

Maybe we can copy the web implementation from that library to support ImageSource.camera on desktop.

jesusrp98:
Planning to open the issue tomorrow if I have the time @ditman.

Also @postacik, I think it would be awesome to implement that feature directly into `image_picker`.

ditman:
We're soon going to work on the `camera` plugin for web, which should implement this specific feature. Then (maybe) depend on the `camera` plugin for the `ImageSource.camera` on desktop browsers?

Joseph-Nathan:
any news for provide `ImageSource.camera` on desktop browsers ?

PR #2804, opened 2020-05-29
[shared_preference_macos] Remove iOS and Android folder in example app

cyanglaz:


cyanglaz:
@stuartmorgan Sure. Need to work on something else before cycling back to this tho.

PR #2813, opened 2020-06-03
[web] Introduce the Link widget

mdebbar:
This PR introduces the Link widget and provides an implementation for the web.

The design of this widget has been discussed in:

Issues:

### DIFF
--- a/packages/url_launcher/url_launcher_web/lib/link.dart
+++ b/packages/url_launcher/url_launcher_web/lib/link.dart
@@ -1,0 +1,295 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:html' as html;
+import 'dart:js_util';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_web_plugins/flutter_web_plugins.dart';
+
+import 'package:url_launcher_platform_interface/link.dart';
+
+typedef _ClickListener = void Function(html.MouseEvent);
+
+/// The unique identifier for the view type to be used for link platform views.
+const String viewType = '__url_launcher::link';
+
+/// The name of the property used to set the viewId on the DOM element.
+const String _viewIdProperty = '__url_launcher::link::viewId';
+
+// TODO(mdebbar): Should use PlatformViewFactory.
+dynamic get webViewFactory => _LinkViewController._viewFactory;
+
+final MethodCodec _codec = const JSONMethodCodec();
+
+/// The delegate for building the [Link] widget on the web.
+///
+/// It uses a platform view to render an anchor element in the DOM.
+class WebLinkDelegate extends StatefulWidget {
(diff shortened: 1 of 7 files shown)


packages/url_launcher/url_launcher_web/lib/link.dart
```diff
@@ -0,0 +1,295 @@
...
+      final String routeName = widget.link.uri.toString();
+      // TODO(mdebbar): how do we know if this is true or false?
+      final bool isUsingRouter = false;
+      // QUESTION: does this update the history entry on the engine side?
+      final MethodCall call = isUsingRouter
+          ? MethodCall('pushRouteInformation', <dynamic, dynamic>{
+              'location': routeName,
+              'state': null,
+            })
+          : MethodCall('pushRoute', routeName);
```
mdebbar, line 85: @chunhtai is there a way currently to tell if the app is in Router mode? Maybe inspect the widget hierarchy for a Router widget?

packages/url_launcher/url_launcher_web/lib/link.dart
```diff
@@ -0,0 +1,295 @@
...
+        return '_blank';
+      default:
+        throw Exception('Unknown LinkTarget value $target.');
+    }
+  }
+
+  @override
+  Future<void> clearFocus() async {
+    // Currently this does nothing on Flutter Web.
+    // TODO(het): Implement this. See https://github.com/flutter/flutter/issues/39496
```
mdebbar, line 248: This was copied from the [similar implementation](https://github.com/flutter/flutter/blob/580a96f9576a6026ab74eb92258812d8b3010119/packages/flutter/lib/src/widgets/platform_view.dart#L406-L409) in the framework. Not sure what's the right thing to do here @hterkelsen.

packages/url_launcher/url_launcher_web/lib/link.dart
```diff
@@ -0,0 +1,295 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Stack(
+      children: <Widget>[
+        widget.link.builder(
+          context,
+          widget.link.isDisabled ? null : _followLink,
+        ),
+        Positioned.fill(
+          // child: _LinkHtmlView(uri: link.uri, target: link.target),
```
hterkelsen, line 106: delete this line and add a TODO to use a `PlatformView` once hit testing is fixed and creation parameters supported

hterkelsen (commented):
Code LGTM. Unfortunate that we have to use `PlatformViewLink` here for a pretty simple use case. This highlights how much we need to flesh out/fix `HtmlElementView`.

In order to actually land this, you will need to split it up into 2 CLs: one which lands the changes in the `_interface` package (with CHANGELOG and pubspec version bump) and one which uses those changes in `package:url_launcher`.

packages/url_launcher/url_launcher_web/lib/link.dart
chunhtai: there is no standard API to check that. Since this is a separate package, i think just do a Router.of(context, nullOk = true) != null should be fine

mdebbar: Thanks!

mdebbar:
Closing this in favor of the split PRs: https://github.com/flutter/plugins/pull/3154 and https://github.com/flutter/plugins/pull/3155.

ditman:
(Moved my review to the PRs above)

PR #2818, opened 2020-06-05
[path_provider] Add Windows support

timsneath:
## Description

Implements path provider for Windows

## Related Issues

Fixes flutter/flutter#41715.

## Checklist

- [X] I signed the [CLA].

## Breaking Change


packages/path_provider/path_provider/lib/path_provider.dart
```diff
@@ -7,6 +7,7 @@ import 'dart:io' show Directory, Platform;
 
 import 'package:flutter/foundation.dart' show kIsWeb, visibleForTesting;
 import 'package:path_provider_linux/path_provider_linux.dart';
+import 'package:path_provider_windows/path_provider_windows.dart';
```
renefloor, line 10: @timsneath It looks like this linux import breaks the support on all platforms in 1.6.10: https://pub.dev/packages/path_provider#-analysis-tab-
As a result it also breaks all platform supports for packages that depend on path_provider.
I guess both the linux import and the windows import have to be conditional imports here.

Sunbreak: [Add Linxu support](https://github.com/flutter/plugins/pull/2789) seems ok?

According to https://github.com/flutter/flutter/issues/52267, it is a workround until flutter tool supports `dartPluginClass` on native plugins

renefloor: The reason I came to the PR is actually because the Linux support is not OK. 
Since that PR on the 2nd of June path_provider doesn't support any platform anymore according to pub.dev. (That's the version 1.6.10 I'm talking about in the comment.)

It wouldn't really be a problem if it is only this package, but every package that depends on path_provider has this error when they publish a new version. A large part of pub breaks for 2 platforms that aren't even in alpha stage yet.

2 issues I found that are possibly related, but I expect more issues to be submitted because of this: 
https://github.com/flutter/flutter/issues/58954
https://github.com/dart-lang/pub-dev/issues/3680

I also found that comment that this is a workaround indeed, but IMO it is a very bad work around for a platform in technical preview when it breaks platforms that are stable.

stuartmorgan: > @timsneath It looks like this linux import breaks the support on all platforms in 1.6.10: https://pub.dev/packages/path_provider#-analysis-tab-
> As a result it also breaks all platform supports for packages that depend on path_provider.

Pana's analysis is wrong, leading to incorrect information being displayed on pub.dev; that's certainly not ideal, but is very different from "breaks all platform support". If you have examples of actual compilation or runtime failures resulting from this, please file a new bug.

> I guess both the linux import and the windows import have to be conditional imports here.

There is no conditional import that selects between individual platforms.

> flutter/flutter#58954

This is entirely unrelated; I've added a comment to that effect.

Let's continue any discussion of the workaround on https://github.com/flutter/flutter/issues/52267 (and/or https://github.com/dart-lang/pub-dev/issues/3680 for that specific side-effec), since the structure at issue here predates this PR.

renefloor: Indeed, it doesn't really break support, but as Flutter is getting more supported platforms the badges to show the platform support become more important.

timsneath: Thanks for the report, Rene!

Resolving this conversation, since as @stuartmorgan noted, it's entirely orthogonal to this pull request.

packages/path_provider/path_provider/pubspec.yaml
```diff
@@ -23,6 +22,8 @@ dependencies:
   path_provider_platform_interface: ^1.0.1
   path_provider_macos: ^0.0.4
   path_provider_linux: ^0.0.1
+  path_provider_windows:
+    path: ../path_provider_windows
```
stuartmorgan, line 26: This can't be landed this way. Normally I would say you'd need to remove this and the endorsement, and land those as a follow-up once path_provider_windows was landed and published, but since it's already been published you should be able to just changes this to ^0.0.1 now.

packages/path_provider/path_provider_windows/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+## 0.0.1+2
```
stuartmorgan, line 1: Why does this have a +2, when this is the initial landing?

packages/path_provider/path_provider_windows/README.md
```diff
@@ -0,0 +1,32 @@
+# path_provider_windows
+
+The Windows implementation of [`path_provider`][1].
+
+*This is not an official package, and is not ready for use!*
```
stuartmorgan, line 5: Remove.

packages/path_provider/path_provider_windows/README.md
```diff
@@ -0,0 +1,32 @@
...
+as a dependency in your `pubspec.yaml`. It will be automatically included in your app
+when you depend on `package:path_provider`.
+
+This is what the above means to your `pubspec.yaml`:
+
+```yaml
+...
+dependencies:
+  ...
+  path_provider: ^1.5.1
```
stuartmorgan, line 28: path_provider should get a version bump in this PR, and this should refer to that version.

packages/path_provider/path_provider_windows/example/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
+name: path_provider_example
+description: Demonstrates how to use the path_provider plugin.
+
+dependencies:
+  flutter:
+    sdk: flutter
+  path_provider: any
+  path_provider_windows:
```
stuartmorgan, line 8: IIRC this needs to be done as an override, since it'll conflict with the path_provider version of the dependency.

packages/path_provider/path_provider_windows/lib/folders.dart
```diff
@@ -0,0 +1,240 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:win32/win32.dart';
+
+/// A class containing the GUID references for each of the documented Windows
+/// known folders. A property of this class may be passed to the `getPath`
+/// method in the [PathProvidersWindows] class to retrieve a known folder from
+/// Windows.
+class WindowsKnownFolder {
```
stuartmorgan, line 11: Is using a class as a namespace a common practice in Dart? I would have expected these to just be file-level constants.

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
```diff
@@ -0,0 +1,93 @@
...
+  /// Path to a directory where the application may place data that is
+  /// user-generated, or that cannot otherwise be recreated by your application.
+  @override
+  Future<String> getApplicationDocumentsPath() =>
+      getPath(WindowsKnownFolder.Documents);
+
+  /// Path to the directory where downloaded files can be stored. This is
+  /// typically the same as %USERPROFILE%\Downloads.
+  @override
+  Future<String> getDownloadsPath() => getPath(WindowsKnownFolder.Downloads);
```
stuartmorgan, line 66: You'll need to fix https://github.com/flutter/plugins/blob/master/packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart#L81

We should really only be blocking iOS/Android as necessary for them not to explode, rather than hard-coding which platforms have support, as it's not future-proof or extensible that way.

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
```diff
@@ -0,0 +1,93 @@
...
+///
+/// This class implements the `package:path_provider` functionality for Windows
+class PathProviderWindows extends PathProviderPlatform {
+  /// Path to the temporary directory on the device that is not backed up and is
+  /// suitable for storing caches of downloaded files.
+  ///
+  /// On Windows, this the path specified by the TMP environment variable, or
+  /// the TEMP environment variable, or the USERPROFILE environment variable,
+  /// or the Windows directory, in order of preference. Windows does not
+  /// guarantee that the path exists or is writeable to.
```
stuartmorgan, line 23: I think we should handle existence here; `path_provider` users should be shielded from the platform-specific details, and they are unlikey to come read these docs. Returning a temp path that doesn't exist seems like an unnecessary foot-gun for developers.

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
```diff
@@ -0,0 +1,93 @@
...
+/// The Windows implementation of [PathProviderPlatform]
+///
+/// This class implements the `package:path_provider` functionality for Windows
+class PathProviderWindows extends PathProviderPlatform {
+  /// Path to the temporary directory on the device that is not backed up and is
+  /// suitable for storing caches of downloaded files.
+  ///
+  /// On Windows, this the path specified by the TMP environment variable, or
+  /// the TEMP environment variable, or the USERPROFILE environment variable,
+  /// or the Windows directory, in order of preference. Windows does not
```
stuartmorgan, line 22: This is all internal to the Win32 call, right? It seems odd to document implementation details that we aren't controlling here.

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
```diff
@@ -0,0 +1,93 @@
...
+      var path = buffer.unpackString(length);
+
+      // GetTempPath adds a trailing backslash, but SHGetKnownFolderPath does not.
+      // Strip off trailing backslash for consistency with other methods here.
+      if (path[path.length - 1] == '\\') {
+        path = path.substring(0, path.length - 1);
+      }
+      free(buffer);
+      return Future.value(path);
+    }
```
stuartmorgan, line 43: This would be safer as:
```
final buffer = ...;
try {
   // all the logic here, including the return
} finally {
  free(buffer);
}
```

So that there's no chance of introducing a codepath that doesn't free `buffer`.

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
```diff
@@ -0,0 +1,93 @@
...
+    if (length == 0) {
+      final error = GetLastError();
+      free(buffer);
+      throw WindowsException(error);
+    } else {
+      var path = buffer.unpackString(length);
+
+      // GetTempPath adds a trailing backslash, but SHGetKnownFolderPath does not.
+      // Strip off trailing backslash for consistency with other methods here.
+      if (path[path.length - 1] == '\\') {
```
stuartmorgan, line 38: `endsWith`?

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
```diff
@@ -0,0 +1,93 @@
...
+  Future<String> getPath(String folderID) {
+    GUID knownFolderID = GUID.fromString(folderID);
+    Pointer<IntPtr> pathPtrPtr = allocate<IntPtr>();
+
+    final hr = SHGetKnownFolderPath(
+        knownFolderID.addressOf, KF_FLAG_DEFAULT, NULL, pathPtrPtr);
+
+    if (FAILED(hr)) {
+      if (hr == E_INVALIDARG || hr == E_FAIL) {
+        throw WindowsException(hr);
```
stuartmorgan, line 81: Isn't this leaking `pathPtrPtr`? The `finally` approach would be good in this method as well.

packages/path_provider/path_provider_windows/lib/folders.dart
jonahwilliams: https://dart.dev/guides/language/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members

timsneath: Ah, I was thinking of this as matching the clause, "However, this isn’t a hard rule. With constants and enum-like types, it may be natural to group them in a class", since this is the moral equivalent of an enum. But I can change it if we don't think this is a good fit.

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
timsneath: Doh 🤦

packages/path_provider/path_provider_windows/README.md
timsneath: Ha! I added that in a vain attempt to satisfy you :) but yeah, I'll remove this.

stuartmorgan: When it's landed, it will be an official package.

packages/path_provider/path_provider_windows/lib/folders.dart
jonahwilliams: It can be nicer for autocomplete too, especially if they are related

technolion:
Could you please add support for `getDownloadsPath` as suggested in this PR?
https://github.com/google/flutter-desktop-embedding/pull/765

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
```diff
@@ -0,0 +1,93 @@
...
+      free(buffer);
+      return Future.value(path);
+    }
+  }
+
+  /// Path to a directory where the application may place application support
+  /// files.
+  @override
+  Future<String> getApplicationSupportPath() =>
+      getPath(WindowsKnownFolder.ProgramData);
```
stuartmorgan, line 50: There's a mismatch here between what the Windows API gives and what a path_provider user is likely to expect; on mobile these folders are in the app sandbox, so it's reasonable to just create files in the directory returned by `getApplicationSupportPath`; in a Win32 app (unless running in a compatibility hypervisor, IIRC) it's the top level folder that all applications share.

For the FDE plugin what I did is append a directory using the executable name (from `GetModuleFileName`). I think we need do that (or something like it if there's a better value we have access to) there.

(We may want to add an API to path_provider_windows that allows people to manually set it to a value of their choosing, overriding the default of exe name.)

packages/path_provider/path_provider_windows/test/path_provider_windows_test.dart
```diff
@@ -0,0 +1,20 @@
...
+
+  setUp(() {
+    pathProvider = PathProviderWindows();
+  });
+
+  tearDown(() {});
+
+  test('getTemporaryPath', () async {
+    expect(await pathProvider.getTemporaryPath(), contains('C:\\'));
+  });
```
stuartmorgan, line 19: - We should have more tests here, so we're covering the API surface.
- You'll either need something like https://github.com/flutter/plugins/pull/2988/files#diff-f9349795ba45b74fe94e478e177c111aR15 around the Win32 calls themselves to allow stubbing them out, or change the tests to early-return on non-Windows platforms; these tests will be run on all platforms, and will fail at library lookup on macOS and Linux (you can see this on the last set of CI runs).

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
escamoteur: Wouldn't it make sense to place all app data in the `userprofile/AppData/local/AppName` as default?

stuartmorgan: Oh, I missed that difference from FDE. I was using RoamingAppData; that (or LocalAppData) seems like the right location for application support data. ProgramData sounds like it's not user-specific.

timsneath:
> Will you have time to iterate on this PR soon? If not, should we do a handoff? I'd like to get it in ASAP after the CI is online so that we have something exercising the CI to prevent it from breaking without us noticing.

Thanks for the prod (and for the pull request review comments). It's time I picked this bit up again (got distracted with Windows Runtime API interop improvements). I'll take a look over this Labor Day weekend and see if I can patch this into good shape.

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
timsneath: Yeah, you're right -- although people might want to know where it is... I've trimmed this back to a more minimal description.

packages/path_provider/path_provider_windows/lib/path_provider_windows.dart
stuartmorgan: Resolving since there's an open PR that would do this.

stuartmorgan (commented):
I did a pass over the PR to address my comments. I'll see if there's any more it needs to make CI happy and update if so, otherwise I think it's just waiting on the new `win32` method I need.

stuartmorgan:
This is ready for review; thanks Tim for the fast turn-around on the `win32` package addition!

stuartmorgan (commented):
(Marking as "reviewed" so it doesn't continue to indicate that the PR is waiting for changes I requested.)

stuartmorgan (approved):

stuartmorgan:
@csells @cyanglaz Since I made non-trivial changes to the PR, I'll wait for one of you to review it before landing.

packages/path_provider/path_provider_windows/example/README.md
```diff
@@ -0,0 +1,8 @@
+# path_provider_windows_example
```
csells, line 1: do we need a separate sample for Windows? won't the normal path_provider sample do?

csells (commented):
all of it looks good except I'm not sure about the need for a Windows-specific sample

packages/path_provider/path_provider_windows/example/README.md
stuartmorgan: There's an open question in the longer term about how to best structure the tests and examples for federated plugins. The plugin repo tooling expects them though, and it does have value in the general case:
- It's a self-contained way to build and test the plugin, which is convenient for plugin development.
- It can test platform-specific APIs (which we don't currently have, but could in the future).

So long term, TBD, short term, this is the standard practice we've been following.

stuartmorgan:
Per https://github.com/flutter/plugins/pull/2988#discussion_r489493810 the application support path approach here isn't correct. I'll need to update the template to add VERSIONINFO, and read that here.

stuartmorgan:
@timsneath It looks like we'll need `GetFileVersionInfo`, `GetFileVersionInfoSize`, and `VerQueryValue` in `win32` now as well, to read this information.

timsneath:
Have a PR ready that will add these APIs, but I need to add some tests before I merge and publish. I'll do that by EOD. https://github.com/timsneath/win32/pull/103

timsneath:
OK, version APIs are all in win32 v1.7.1 (https://timsneath.github.io/win32/topics/version-topic.html)

stuartmorgan:
@csells Please re-review for the new AppData subdirectory logic.

csells (approved):

stuartmorgan:
> I just wasn't aware of it. Updated.

I'm going to have to revert this; it appear that it makes the flutter/plugins repo tooling think the tests have failed:

```
RUNNING path_provider/path_provider_windows tests...
Running "flutter pub get" in path_provider_windows...               1.4s
No tests ran.
SKIPPING path_provider/path_provider_windows/example - no test subdirectory



Tests for the following packages are failing (see above):
 * path_provider/path_provider_windows
```

This wasn't happening with the `skip` option 🤷

PR #2819, opened 2020-06-05
[video_player] Miscellaneous fixes to improve video player in web.

ditman:
## Description

While testing the image_picker_web plugin, I noticed that video playback was a little bit flaky. This PR addresses some of the subtle bugs in the image_picker example app, video_player plugin and video_player_web implementation.

## Related Issues

n/a

## Checklist

- [x] I signed the [CLA].

## Breaking Change

All changes are minor and independent, so no need to split this PR in several different ones. This all should be mergeable and publishable in parallel.

### DIFF
--- a/packages/image_picker/image_picker/example/lib/main.dart
+++ b/packages/image_picker/image_picker/example/lib/main.dart
@@ -42,6 +42,7 @@ class _MyHomePageState extends State<MyHomePage> {
   dynamic _pickImageError;
   bool isVideo = false;
   VideoPlayerController _controller;
+  VideoPlayerController _toBeDisposed;
   String _retrieveDataError;
 
   final ImagePicker _picker = ImagePicker();
@@ -120,10 +121,11 @@ class _MyHomePageState extends State<MyHomePage> {
   }
 
   Future<void> _disposeVideoController() async {
-    if (_controller != null) {
-      await _controller.dispose();
-      _controller = null;
+    if (_toBeDisposed != null) {
+      await _toBeDisposed.dispose();
     }
+    _toBeDisposed = _controller;
+    _controller = null;
   }
 
   Widget _previewVideo() {
--- a/packages/image_picker/image_picker/example/lib/main.dart
+++ b/packages/image_picker/image_picker/example/lib/main.dart
@@ -54,10 +54,16 @@ class _MyHomePageState extends State<MyHomePage> {
       await _disposeVideoController();
       if (kIsWeb) {
         _controller = VideoPlayerController.network(file.path);
-      } else {
+        // In web, most browsers won't honor a programmatic call to .play
+        // if the video has a sound track (and is not muted).
+        // Mute the video so it auto-plays in web!
+        // This is not needed if the call to .play is the result of user
+        // interaction (clicking on a "play" button, for example).
+        await _controller.setVolume(0.0);
+    } else {
         _controller = VideoPlayerController.file(File(file.path));
+        await _controller.setVolume(1.0);
       }
-      await _controller.setVolume(1.0);
       await _controller.initialize();
       await _controller.setLooping(true);
       await _controller.play();
--- a/packages/image_picker/image_picker/example/lib/main.dart
+++ b/packages/image_picker/image_picker/example/lib/main.dart
@@ -61,7 +61,7 @@ class _MyHomePageState extends State<MyHomePage> {
         // This is not needed if the call to .play is the result of user
         // interaction (clicking on a "play" button, for example).
         await _controller.setVolume(0.0);
-    } else {
+      } else {
         _controller = VideoPlayerController.file(File(file.path));
         await _controller.setVolume(1.0);
       }
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -97,7 +97,7 @@ class VideoPlayerValue {
   /// Returns [size.width] / [size.height] when size is non-null, or `1.0.` when
   /// size is null or the aspect ratio would be less than or equal to 0.0.
   double get aspectRatio {
-    if (size == null) {
+    if (size == null || size.width == 0 || size.height == 0) {
       return 1.0;
     }
     final double aspectRatio = size.width / size.height;
--- a/packages/video_player/video_player_web/lib/video_player_web.dart
+++ b/packages/video_player/video_player_web/lib/video_player_web.dart
@@ -66,10 +66,13 @@ class VideoPlayerPlugin extends VideoPlayerPlatform {
     final int textureId = _textureCounter;
     _textureCounter++;
 
-    Uri uri;
+    String uri;
     switch (dataSource.sourceType) {
       case DataSourceType.network:
-        uri = Uri.parse(dataSource.uri);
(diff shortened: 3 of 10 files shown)


packages/image_picker/image_picker/example/lib/main.dart
```diff
@@ -114,10 +121,11 @@ class _MyHomePageState extends State<MyHomePage> {
   }
 
   Future<void> _disposeVideoController() async {
-    if (_controller != null) {
-      await _controller.dispose();
-      _controller = null;
+    if (_toBeDisposed != null) {
```
cyanglaz, line 124: I don't get this. Are we expecting `_disposeVideoController` to be called twice and only disposing the controller the second time?

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -97,7 +97,7 @@ class VideoPlayerValue {
   /// Returns [size.width] / [size.height] when size is non-null, or `1.0.` when
   /// size is null or the aspect ratio would be less than or equal to 0.0.
   double get aspectRatio {
-    if (size == null) {
+    if (size == null || size.width == 0 || size.height == 0) {
```
cyanglaz, line 100: can we test this? Should be easy do have a unit test in dart.

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -66,10 +66,13 @@ class VideoPlayerPlugin extends VideoPlayerPlatform {
     final int textureId = _textureCounter;
     _textureCounter++;
 
-    Uri uri;
+    String uri;
     switch (dataSource.sourceType) {
       case DataSourceType.network:
-        uri = Uri.parse(dataSource.uri);
+        // Validate the incoming uri
+        Uri.parse(dataSource.uri);
```
cyanglaz, line 73: What do we do with the result of this call?

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -218,6 +224,11 @@ class _VideoPlayer {
   }
 
   void setVolume(double value) {
+    if (value > 0.0) {
```
cyanglaz, line 227: might need to introduce a small epsilon when comparing float values

cyanglaz: Or we make the `muted` property public so user can directly set mute instead of setting a 0.0 value.

packages/image_picker/image_picker/example/lib/main.dart
ditman: The issue is that we **`dispose` of the controller before we're done with it.**

The situation is the following: 

You pick a video and while it's playing, you pick a second video.

The first thing the playVideo method does is to [call `_disposeVideoController`](https://github.com/flutter/plugins/blob/master/packages/image_picker/image_picker/example/lib/main.dart#L54) which disposes of the current controller.

Then we create a [new video controller and cause a re-render](https://github.com/flutter/plugins/blob/master/packages/image_picker/image_picker/example/lib/main.dart#L55-L64), which ends up [updating the VideoPlayer widget](https://github.com/flutter/plugins/blob/master/packages/image_picker/image_picker/example/lib/main.dart#L388) with the newest controller (through the `_previewVideo` method). Then we get here:

https://github.com/flutter/plugins/blob/master/packages/video_player/video_player/lib/video_player.dart#L537

That is attempting to remove a listener from the *old* controller.

The old controller has already been disposed of, and that `removeListener` [throws an Assertion](https://api.flutter.dev/flutter/foundation/ValueNotifier/removeListener.html). 

In Safari the assertion is fatal (it interrupts all Javascript, and halts the program). In every other platform, it gets logged (but I don't think we should ignore an assertion?).

This is to dispose of the controller later, when we're really sure that nobody else is using it anymore.

(Note that the LAST controller will get taken care of by the [`dispose` method of the _MyHomePageState class](https://github.com/flutter/plugins/blob/master/packages/image_picker/image_picker/example/lib/main.dart#L109))

packages/video_player/video_player_web/lib/video_player_web.dart
ditman: I think this was only added originally to *throw* if the passed in dataSource.uri was not valid, so I left it there only to "validate" the uri, since now we pass dataSource.uri directly.

packages/video_player/video_player_web/lib/video_player_web.dart
ditman: This "hatch" is only there so web users can do: setVolume(0.0) and get a "muted" player.

About making the `muted` property public: I agree, but I am not sure what the implications are for Android and iOS of that.

There's a huge API surface for the `video` element that is being hidden by the current platform API, and I'm not sure if it's because we want a very simple API, or because the API can't be extended for other platforms :/

For this ticket I was only trying to make the selected video auto-play in the image_picker web example, not do any significant changes to the API of the video_player widget.

packages/video_player/video_player/lib/video_player.dart
ditman: Yes, let me see if there's any unit test for the aspectRatio

ditman: Added unit tests for the aspectRatio getter. There were none before, I tested all the cases.

packages/video_player/video_player_web/lib/video_player_web.dart
yjbanov: I'm not sure if it helps real-world apps, but Dart's `Uri` implementation is a spec-compliant implementation that's implemented in userspace. Last time I checked it was 18KB big. If this is for validation only, could it be wrapped in an `assert`?

ditman:
> I'm not sure if it helps real-world apps, but Dart's Uri implementation is a spec-compliant implementation that's implemented in userspace. Last time I checked it was 18KB big. If this is for validation only, could it be wrapped in an assert?

@yjbanov I'll try what happens with malformed URLs, since this may be user-generated content.

ditman:
Tested malformed URLs, and the plugin seems to be able to spit out an error that is understandable:

```
Error: PlatformException(MEDIA_ERR_SRC_NOT_SUPPORTED, MEDIA_ELEMENT_ERROR: Format error, The video
has been found to be unsuitable (missing or in a format not supported by your browser).)
```

packages/video_player/video_player_web/lib/video_player_web.dart
ditman: Tested this, and the plugin is able to report the "malformedness" of the URL to the user:

```
Error: PlatformException(MEDIA_ERR_SRC_NOT_SUPPORTED, MEDIA_ELEMENT_ERROR: Format error, The video
has been found to be unsuitable (missing or in a format not supported by your browser).)
```

I'll remove the Uri parsing to save a few kb. Thanks for bringing this up!

packages/video_player/video_player_web/lib/video_player_web.dart
ditman: I've created this issue so we can discuss this new API: 

* https://github.com/flutter/flutter/issues/60721

If we decide to add a "muted" attribute to the public API, this hatch can be removed!

PR #2835, opened 2020-06-17
Fix typo in image_picker_for_web README.md

lrhn:
## Description

Fixes typo in README file.

### DIFF
--- a/packages/image_picker/image_picker_for_web/README.md
+++ b/packages/image_picker/image_picker_for_web/README.md
@@ -5,7 +5,7 @@ A web implementation of [`image_picker`][1].
 ## Browser Support
 
 Since Web Browsers don't offer direct access to their users' file system,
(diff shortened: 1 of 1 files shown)


stuartmorgan (changes requested):
Thanks for the contribution! We apologize for the long delay in triaging this PR. We’re in the process of overhauling our PR triage system to respond much more quickly, as well as working through the backlog.

Could you add a version bump and CHANGELOG update to this? That's part of our publishing process for all changes, including READMEs (since the most important copy of the README is the one on pub.dev, and we need to publish for that to be updated.)

stuartmorgan:
Actually, we can just pick this up on the next publish that happens, rather than doing a version bump and publish just for this. Thanks again!

stuartmorgan (approved):

PR #2847, opened 2020-06-25
[url_launcher_web] Add tel URL support

Vanethos:
## Description

Adds the ability to open `tel` URLS in web

## Related Issues

* Fixes https://github.com/flutter/flutter/issues/58553
* Fixes https://github.com/flutter/flutter/issues/60293

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/url_launcher/url_launcher/pubspec.yaml
+++ b/packages/url_launcher/url_launcher/pubspec.yaml
@@ -26,7 +26,11 @@ dependencies:
   # validation, so we set a ^ constraint.
   # TODO(amirh): Revisit this (either update this part in the  design or the pub tool).
   # https://github.com/flutter/flutter/issues/46264
-  url_launcher_web: ^0.1.0+1
+  url_launcher_web: 
+    git: 
+        url: https://github.com/Vanethos/plugins
+        ref: feature/add-tel-url
+        path: packages/url_launcher/url_launcher_web
   url_launcher_macos: ^0.0.1
 
 dev_dependencies:
--- a/packages/url_launcher/url_launcher/pubspec.yaml
+++ b/packages/url_launcher/url_launcher/pubspec.yaml
@@ -26,11 +26,7 @@ dependencies:
   # validation, so we set a ^ constraint.
   # TODO(amirh): Revisit this (either update this part in the  design or the pub tool).
   # https://github.com/flutter/flutter/issues/46264
-  url_launcher_web: 
-    git: 
-        url: https://github.com/Vanethos/plugins
-        ref: feature/add-tel-url
-        path: packages/url_launcher/url_launcher_web
+  url_launcher_web: ^0.1.0+1
   url_launcher_macos: ^0.0.1
 
 dev_dependencies:
--- a/packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
+++ b/packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
@@ -8,6 +8,7 @@ import 'package:url_launcher_platform_interface/url_launcher_platform_interface.
 import 'package:platform_detect/platform_detect.dart' show browser;
 
 const _mailtoScheme = 'mailto';
+const _phoneScheme = 'tel';
 
 /// The web implementation of [UrlLauncherPlatform].
 ///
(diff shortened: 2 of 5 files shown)


packages/url_launcher/url_launcher/pubspec.yaml
```diff
@@ -26,7 +26,11 @@ dependencies:
   # validation, so we set a ^ constraint.
   # TODO(amirh): Revisit this (either update this part in the  design or the pub tool).
   # https://github.com/flutter/flutter/issues/46264
-  url_launcher_web: ^0.1.0+1
+  url_launcher_web: 
+    git: 
+        url: https://github.com/Vanethos/plugins
+        ref: feature/add-tel-url
+        path: packages/url_launcher/url_launcher_web
```
ditman, line 33: This needs to be reverted, otherwise you'll never be able to publish the plugin.

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
```diff
@@ -8,6 +8,7 @@ import 'package:url_launcher_platform_interface/url_launcher_platform_interface.
 import 'package:platform_detect/platform_detect.dart' show browser;
 
 const _mailtoScheme = 'mailto';
+const _phoneScheme = 'tel';
```
ditman, line 11: Can you create a Set containing this two URLs?

There was a bug with Safari in iOS where people would return to a "blank" page after sending an email (see https://github.com/flutter/flutter/issues/51461)

I *think* you'll need to extend the _isMailtoScheme method to also handle `tel`, and having them in a set is much easier. It'd be very easy to add `sms` at the same time as well.

You can also use Set methods to create the _supportedSchemes Set.

ditman: Here's a link with the "Set" idea:

https://github.com/tugorez/plugins/pull/1

(there's other stuff that is not needed anymore, look at the `_safariTargetTopSchemes` Set and how it's used throughout the code)

packages/url_launcher/url_launcher/pubspec.yaml
ditman: Note that `0.1.2` will be automatically picked up by url_launcher after it gets published. If you want to have this change locally for your own tests, it's fine, but you need to remove it from the PR, so the CI bots (publishable) don't complain at you.

Vanethos: OH! SNAP! I edited it locally and did not commit it! will do it now

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
Vanethos: Seems awesome! will add it with the current code structure

Vanethos: Just have one question though: 

Why is `_supportedSchemes` a private static variable of the class, but the `_safariTargetTopSchemes` a global private constant?

In the codebase that you shared this wasn't the case @ditman  🤔

ditman: My version was just a sketch meant to give an idea (twice now! :P)

I think the current version is better:

`_safariTargetTopSchemes` is a property of an external entity to the plugin (Safari), it is defined in the same file, but it could come from some configuration file or some other package.
`_supportedSchemes` is a property of the plugin. What urls does this plugin support? And it makes sense it lives within the class.

(But you could argue that it can also be considered configuration, and I would also agree with you :P)

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
```diff
@@ -7,7 +7,10 @@ import 'package:url_launcher_platform_interface/url_launcher_platform_interface.
 
 import 'package:platform_detect/platform_detect.dart' show browser;
 
-const _mailtoScheme = 'mailto';
+const _safariTargetTopSchemes = {
+  'mailto',
+  'tel',
+};
```
ditman, line 13: Can you also add `'sms'` (and a test for it)? That way we can kill two birds (issues) with one stone (this pr)! :)

Vanethos: Done!

And also added some more tests that were missing for tel and added the same for sms

ditman:
Thanks for adding the "sms" protocol! The only missing thing is to run `dartfmt -w .` to make the "format" check happy!

Vanethos:
oh my! forgot about that, sorry!

Done :D

ditman:
This looks great, let's merge and publish it! Thanks for the contribution @Vanethos!!

Vanethos:
Many many thanks for the patience and time @ditman !

It was really fun :D

ditman:
This change has passed the post-submit checks and has been published as `url_launcher_web` 0.1.2!

PR #2854, opened 2020-06-30
[e2e] Use SettableFutures instead of CompletableFuture

jiahaog:
## Description

This unblocks the import of the plugin internally.

## Related Issues

b/160208320

## Breaking Change

Breaking change if users depend on `E2EPlugin.testResults` as an instance of `CompletableFuture`.

### DIFF
--- a/packages/e2e/android/src/main/java/dev/flutter/plugins/e2e/E2EPlugin.java
+++ b/packages/e2e/android/src/main/java/dev/flutter/plugins/e2e/E2EPlugin.java
@@ -5,7 +5,6 @@
 package dev.flutter.plugins.e2e;
 
 import android.content.Context;
-import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import io.flutter.embedding.engine.plugins.FlutterPlugin;
 import io.flutter.plugin.common.BinaryMessenger;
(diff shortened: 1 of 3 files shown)


packages/e2e/android/src/main/java/dev/flutter/plugins/e2e/E2EPlugin.java
```diff
@@ -13,13 +15,14 @@
...
-import java9.util.concurrent.CompletableFuture;
 
 /** E2EPlugin */
 public class E2EPlugin implements MethodCallHandler, FlutterPlugin {
   private MethodChannel methodChannel;
 
-  public static CompletableFuture<Map<String, String>> testResults = new CompletableFuture<>();
+  private static final SettableFuture<Map<String, String>> testResultsSettable =
+      SettableFuture.create();
+  public static final ListenableFuture<Map<String, String>> testResults = testResultsSettable;
```
dnfield, line 25: Can we just make this be a `Future` instead of needing to publicly export its type?

I'm a little concerned that at some point we'll have a reason to avoid using Guava for this, and we'll have to break people again. I'm also not clear on what benefits consumers get from having this be a ListenableFuture rather than just a Future.

jiahaog: Got it, done

jiahaog:
Thanks!

jiahaog:
Sorry, I've never done this before, should I follow https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#the-release-process to release it onto pub?

dnfield:
Yes.  If you need permissions @kf6gpe can help

PR #2859, opened 2020-07-02
[e2e] Add customizable `flutter_driver` adaptor

CareF:
## Description

Add a customizable `flutter_driver` adaptor, allowing changing timeout time, flag to enable timeline tracing. 
A use case is [here](https://github.com/CareF/measure_overhead) where I'm using the timeline to estimate overhead of `E2E` and `flutter_driver` (E2E is better!) 

## Related Issues

Not a full implementation but can be considered a work around for flutter/flutter#58789

## Test

It seems that we don't yet have unit test for E2E. I changed the related example. 

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/e2e/example/test_driver/example_e2e_test.dart
+++ b/packages/e2e/example/test_driver/example_e2e_test.dart
@@ -2,4 +2,4 @@ import 'dart:async';
 
 import 'package:e2e/e2e_driver.dart' as e2e;
 
-Future<void> main() async => e2e.main();
+Future<void> main() async => e2e.e2eDriver();
--- a/packages/e2e/lib/e2e_driver.dart
+++ b/packages/e2e/lib/e2e_driver.dart
@@ -19,12 +19,20 @@ Future<void> main() async => e2eDriver();
 /// Future<void> main() async => e2e.e2eDriver();
 ///
 /// ```
+///
+/// `timeout` controls the longest time waited before the test ends, not
+/// necessarily the execution time for the test app.
+///
+/// `traceTimeline` flag controls if timeline and timeline summary should be
+/// collected.
+///
+/// `testName` is used as the file name for the output timeline files.
 Future<void> e2eDriver({
   Duration timeout = const Duration(minutes: 1),
-  String testName,
   bool traceTimeline = false,
+  String testName,
 }) async {
-  print(testName);
+  assert(!traceTimeline || testName != null);
   final FlutterDriver driver = await FlutterDriver.connect();
   String jsonResult;
   Timeline timeline;
--- a/packages/e2e/lib/e2e_driver.dart
+++ b/packages/e2e/lib/e2e_driver.dart
@@ -42,6 +42,7 @@ Future<void> e2eDriver({
 
   if (traceTimeline) {
     timeline = await driver.traceAction(runner);
+    // In the traceAction call runner is awaited.
   } else {
     await runner();
   }
--- a/packages/e2e/lib/e2e_driver.dart
+++ b/packages/e2e/lib/e2e_driver.dart
@@ -4,7 +4,7 @@ import 'dart:io';
 import 'package:e2e/common.dart' as e2e;
 import 'package:flutter_driver/flutter_driver.dart';
 
-Future<void> main() async => e2eDriver();
+Future<void> main() => e2eDriver();
 
 /// Adaptor to run E2E test using `flutter drive`.
 ///
--- a/packages/e2e/lib/e2e_driver.dart
+++ b/packages/e2e/lib/e2e_driver.dart
@@ -8,8 +8,16 @@ Future<void> main() async => e2eDriver();
 
 /// Adaptor to run E2E test using `flutter drive`.
 ///
-/// For an E2E test `<test_name>.dart`, put a file named `<test_name>_test.dart`
-/// in the app's `test_driver` directory:
+/// `timeout` controls the longest time waited before the test ends, not
+/// necessarily the execution time for the test app.
+///
+/// `traceTimeline` flag controls if timeline and timeline summary should be
+/// collected.
+///
+/// `testName` is used as the file name for the output timeline files.
+///
+/// To an E2E test `<test_name>.dart` using `flutter drive`, put a file named
+/// `<test_name>_test.dart` in the app's `test_driver` directory:
 ///
 /// ```dart
 /// import 'dart:async';
@@ -19,14 +27,6 @@ Future<void> main() async => e2eDriver();
 /// Future<void> main() async => e2e.e2eDriver();
 ///
 /// ```
-///
-/// `timeout` controls the longest time waited before the test ends, not
-/// necessarily the execution time for the test app.
-///
-/// `traceTimeline` flag controls if timeline and timeline summary should be
-/// collected.
-///
-/// `testName` is used as the file name for the output timeline files.
 Future<void> e2eDriver({
   Duration timeout = const Duration(minutes: 1),
   bool traceTimeline = false,
--- a/packages/e2e/lib/e2e_driver.dart
+++ b/packages/e2e/lib/e2e_driver.dart
@@ -24,7 +24,6 @@ Future<void> e2eDriver({
   String testName,
   bool traceTimeline = false,
 }) async {
-
   print(testName);
   final FlutterDriver driver = await FlutterDriver.connect();
   String jsonResult;
@@ -32,10 +31,10 @@ Future<void> e2eDriver({
   Future<void> runner() async {
     jsonResult = await driver.requestData(null, timeout: timeout);
   }
+
   if (traceTimeline) {
     timeline = await driver.traceAction(runner);
-  }
-  else {
+  } else {
     await runner();
   }
   final e2e.Response response = e2e.Response.fromJson(jsonResult);
--- a/packages/e2e/lib/e2e_driver.dart
+++ b/packages/e2e/lib/e2e_driver.dart
@@ -8,8 +8,9 @@ Future<void> main() async => e2eDriver();
 
 /// Adaptor to run E2E test using `flutter drive`.
 ///
-/// `timeout` controls the longest time waited before the test ends, not
-/// necessarily the execution time for the test app.
+/// `timeout` controls the longest time waited before the test ends.
+/// It is not necessarily the execution time for the test app.
+/// The test can and usually runs for shorter time if the test ends early.
 ///
 /// `traceTimeline` flag controls if timeline and timeline summary should be
 /// collected.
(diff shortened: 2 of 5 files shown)


packages/e2e/example/test_driver/example_e2e_test.dart
```diff
@@ -2,4 +2,4 @@ import 'dart:async';
 
 import 'package:e2e/e2e_driver.dart' as e2e;
 
-Future<void> main() async => e2e.main();
+Future<void> main() async => e2e.e2eDriver();
```
dnfield, line 5: Why this change?

packages/e2e/lib/e2e_driver.dart
```diff
@@ -4,15 +4,49 @@ import 'dart:io';
...
+///
+/// Future<void> main() async => e2e.e2eDriver();
+///
+/// ```
+Future<void> e2eDriver({
+  Duration timeout = const Duration(minutes: 1),
+  String testName,
+  bool traceTimeline = false,
+}) async {
+  print(testName);
```
dnfield, line 27: Remove print

packages/e2e/lib/e2e_driver.dart
```diff
@@ -4,15 +4,49 @@ import 'dart:io';
...
+  Future<void> runner() async {
+    jsonResult = await driver.requestData(null, timeout: timeout);
+  }
+
+  if (traceTimeline) {
+    timeline = await driver.traceAction(runner);
+  } else {
+    await runner();
+  }
   final e2e.Response response = e2e.Response.fromJson(jsonResult);
```
dnfield, line 40: Won't this now be null sometimes and throw?

packages/e2e/lib/e2e_driver.dart
```diff
@@ -4,15 +4,49 @@ import 'dart:io';
...
+/// in the app's `test_driver` directory:
+///
+/// ```dart
+/// import 'dart:async';
+///
+/// import 'package:e2e/e2e_driver.dart' as e2e;
+///
+/// Future<void> main() async => e2e.e2eDriver();
+///
+/// ```
```
dnfield, line 21: The doc string should explain the parameters and how/why to use them.

packages/e2e/example/test_driver/example_e2e_test.dart
CareF: It's in the example. I'm thinking with the new function `e2e.e2eDriver` is the recommended solution, but `main` is preserved to avoid breaking change.

dnfield: But why even change the method here? Why not just add a new test that covers the case you're interested in?

packages/e2e/lib/e2e_driver.dart
CareF: Why? `runner` is awaited either directly or in `traceAction`.

packages/e2e/example/test_driver/example_e2e_test.dart
CareF: I thought these examples are not test cases but demos for users about how to use e2e.

CareF: And another thing is, we currently don't have a `test` directory under e2e, I'm not sure how it's unit tested, and this part have to connect to a driver and I don't even know if we are able to run this repo on devicelab.

dnfield: We could certainlyl add tests to e2e :)

It's just not clear to me why we're adding this redirecting method and calling it directly instead of just calling main.

packages/e2e/lib/e2e_driver.dart
dnfield: Ahh I see now. 

This is a nit, but IMO it would read a bit more clearly if you just directly used start/stop trace if needed, rather than passing a closure to a convenience wrapper.

packages/e2e/example/test_driver/example_e2e_test.dart
CareF: So `main` is limited for the format of arguments, due to convention and due to if we want `e2e_driver.dart` to be used directly as the driver script. `main` is using the default argument version of `e2eDeiver` but I'm not expecting future users to use `main` anymore.

packages/e2e/lib/e2e_driver.dart
CareF: `FlutterDriver` is an abstract class. We have two implementation of it and all override `traceAction`. I would rather still use this method, but add some comments here.

packages/e2e/example/test_driver/example_e2e_test.dart
CareF: @liyuqian No it shouldn't. On the one hand, It's just an async main, there's nothing it need to await for, on the other hand, return a Future without awaiting is very common in Framework `Future<void> Foo() {return Future<void>.delayed(...)`, where whoever `await` this `Foo` would actually `await` this `Future.delayed`.

liyuqian: Oh right, it's ` => e2e.main()` instead of `{ e2e.main(); }`. I think only the latter needs `{ await e2e.main(); }`. Anyway, if we don't `await`, can we remove the `async` and just use `Future<void> main() => e2e.main();`? That way, it's very clear that no `await` is expected there.

packages/e2e/lib/e2e_driver.dart
```diff
@@ -4,15 +4,59 @@ import 'dart:io';
...
 import 'package:flutter_driver/flutter_driver.dart';
 
-Future<void> main() async {
+Future<void> main() async => e2eDriver();
+
+/// Adaptor to run E2E test using `flutter drive`.
+///
+/// `timeout` controls the longest time waited before the test ends.
+/// It is not necessarily the execution time for the test app.
+/// The test can and usually runs for shorter time if the test ends early.
```
dnfield, line 13: Maybe just `The test may finish sooner than the timeout`?  I don't think we should say anything about usual.

packages/e2e/lib/e2e_driver.dart
```diff
@@ -4,15 +4,59 @@ import 'dart:io';
...
+/// ```dart
+/// import 'dart:async';
+///
+/// import 'package:e2e/e2e_driver.dart' as e2e;
+///
+/// Future<void> main() async => e2e.e2eDriver();
+///
+/// ```
+Future<void> e2eDriver({
+  Duration timeout = const Duration(minutes: 1),
```
dnfield, line 32: Generally we try to avoid timeouts in tests, because they're a source of flakiness.

That said, I'm not sure that 1 minute is long enough here - integration tests can easily take more (for example, they may have to wait on network requests, animations to finish, etc). In the devicelab we use a default of 15 minutes for the entire test run. This should probably be higher by default, maybe something like an hour.

We should also document that these parameters must not be null.

dnfield:
Looking at this a bit more, I think it would be better if we just offered some kind of API to get the timeline at various points, and to help with starting/stopping recording.

I don't think this is fine grained enough - someone may want the timeline for just part of their test, and they may not want to write the timeline to a local file system - they might need the test itself to send the timeline somewhere else.

Can we re-think the design here to handle that kind of use case?

packages/e2e/example/test_driver/example_e2e_test.dart
CareF: > Oh right, it's ` => e2e.main()` instead of `{ e2e.main(); }`. I think only the latter needs `{ await e2e.main(); }`. Anyway, if we don't `await`, can we remove the `async` and just use `Future<void> main() => e2e.main();`? That way, it's very clear that no `await` is expected there.

Good point! I deleted the `async`.

packages/e2e/lib/e2e_driver.dart
CareF: This 1min is what used to be hard coded in `main`. Any suggestion what value I should put here?

CareF:
> Looking at this a bit more, I think it would be better if we just offered some kind of API to get the timeline at various points, and to help with starting/stopping recording.
> 
> I don't think this is fine grained enough - someone may want the timeline for just part of their test, and they may not want to write the timeline to a local file system - they might need the test itself to send the timeline somewhere else.
> 
> Can we re-think the design here to handle that kind of use case?

The hard part is, the test is driven on the device, and the timeline is collected on host. The generation of the timeline is distributed in framework and engine which we have very limited control. To achieve that there will have to be conversation between the device and the host, making the device code depending directly or indirectly on `flutter_driver`, and there will also be delay for this communication.

dnfield:
The timeline is actually connected on the device by the driver extension. It's fine to make it easier to shovel it over to the host to write, but we should expose an easier way to just collect it whether you end up using the driver extension or not.

CareF:
So if the collecting and processing of timeline is completely on device, it makes finer graining but: 
1. It introduces the overhead of parsing json on device; 
2. We need to design a new set of APIs for the timeline collecting, on the back of which we need more codes to connect E2E bindings to vmservice and flutter_driver. 

It should be do-able, but for my project I don't think I should invest more time on it because this feature is not critical for me. It will be great if this PR lands and I can use this feature for my project, otherwise I will put these codes separately in my projects. 

Our (with @liyuqian) original plan is to implement a frame build time statistics API in or out of E2E, but independent of the timeline for performance test, both for the purpose of host independent and for the purpose of using it in release mode.

packages/e2e/lib/e2e_driver.dart
CareF: And for the documents I added an `assert(timeout != null);`. I'm expecting it will be obvious since the implementation will be shown together with detailed documents anyway.

CareF:
@dnfield Would you mind letting me know if we finally decide to refuse this PR? I will need to add new features as flutter/flutter#61490 and I don't want to introducing conflicts.

dnfield:
I'd prefer to see a solution for this that does not require a host device to be attached. For example, we should support getting timeline information on Firebase Testlab.

CareF:
> I'd prefer to see a solution for this that does not require a host device to be attached. For example, we should support getting timeline information on Firebase Testlab.

@dnfield I think timeline requires vmservice, which I'm not sure how to have it enabled for an app running completely without a host.

liyuqian:
https://github.com/flutter/flutter/pull/61509/files#diff-73e66281b14c792afcdcd20acb9edd92R5 says that this is blocking that PR. I wonder what's the future plan for this PR?

CareF:
> https://github.com/flutter/flutter/pull/61509/files#diff-73e66281b14c792afcdcd20acb9edd92R5 says that this is blocking that PR. I wonder what's the future plan for this PR?

@liyuqian I didn't update here because @dnfield mentioned: 

> I'd prefer to see a solution for this that does not require a host device to be attached. For example, we should support getting timeline information on Firebase Testlab. 

I think I am not planning on implementing a host-independent timeline collection but #2873 would be an infra for that. I might make sense to modify what's here to make a new PR, with an `e2e_driver` that write the data defined in #2873 to a file, which will be used for flutter/flutter#61509.

PR #2866, opened 2020-07-08
[e2e] Fix incorrect test results when one test passes then another fails

jiahaog:
## Description

For example, the following test will result in an error reported for the first test case.

```
void main() {
  testWidgets('a test that passes', (tester) async {
    expect(true, true);
  });

  testWidgets('a test that fails', (tester) async {
    expect(true, false);
  });
}
```

We need to reset `reportTestException` back to the previous value after completion of `runTest`, or repeated failures will cause the exception handler for a previous test to be invoked, as they "stack". Instead of reseting it, however, do this once in the constructor because the test description is already provided by the function signature.

## Breaking Change

### DIFF
--- a/packages/e2e/CHANGELOG.md
+++ b/packages/e2e/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.6.1+1
+
+* Fix incorrect test results when one test passes then another fails
+
 ## 0.6.1
 
 * Added `data` in the reported json.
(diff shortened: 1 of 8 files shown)


dnfield:
Going to land this on the understanding that CI is red because of issues upstream in Flutter that have been resolved.

dnfield:
Sorry, commented on wrong PR...

jiahaog:
Done, but this approach does slow down the tests significantly as it has to start up the flutter tester for each test in a child process. One thing I'm not sure of is assuming `flutter` to be in `$PATH` which might not work internally. Getting the tests running there is on my todo list and might be something to revisit in the future.

PR #2869, opened 2020-07-10
[connectivity_for_web] Fix JS Interop in release mode.

ditman:
## Description

The old version of the plugin used a generated JS-interop file for a class that is available natively from dart:html.

This is [not allowed](https://github.com/dart-lang/sdk/issues/42200).

Since the native dart:html class is available, this PR changes the plugin to use that (I can't believe I missed that on my previous version!)

This PR also cleans up all the paraphernalia used to generate the old JS-interop file.

## Related Issues

* Fixes https://github.com/flutter/flutter/issues/60855

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/connectivity/connectivity_for_web/lib/src/utils/connectivity_result.dart
```diff
@@ -1,27 +1,22 @@
...
 import 'package:connectivity_platform_interface/connectivity_platform_interface.dart';
 
 /// Converts an incoming NetworkInformation object into the correct ConnectivityResult.
-//
-// We can't be more specific on the signature of this method because the API is odd,
-// data can come from a static value in the DOM, or as the 'target' of a DOM Event.
-//
-// If we type info as `NetworkInformation`, Dart will complain with:
-// "Uncaught Error: Expected a value of type 'NetworkInformation',
-// but got one of type 'NetworkInformation'"
```
ditman, line 10: This was probably Dart telling me that there already existed a native `NetworkInformation` class, but I misinterpreted as "coming from different parts"

packages/connectivity/connectivity_for_web/lib/src/network_information_api_connectivity_plugin.dart
```diff
@@ -38,7 +37,7 @@ class NetworkInformationApiConnectivityPlugin extends ConnectivityPlugin {
   Stream<ConnectivityResult> get onConnectivityChanged {
     if (_connectivityResult == null) {
       _connectivityResult = StreamController<ConnectivityResult>();
-      _networkInformation.onchange = allowInterop((_) {
+      _networkInformation.onChange.listen((_) {
```
mdebbar, line 40: I know this is unrelated to your PR, but I noticed it and wanted to point it out. The listener being attached here is never cleaned up. This will potentially cause problems with hot restart since the old listener will remain intact while a new listener is created.

mdebbar: Actually, before this PR there used to be only one listener that you keep overriding. Now you're using a stream that could have multiple listeners which makes it possible to leak listeners.

ditman: Interesting! I don't know of any way to detect the app is hot-restarting to clean-up the plugin!

mdebbar: This is how we do it in the [web engine](https://github.com/flutter/engine/blob/39e98d28b57e4f36de08f7dfff94550b3b2b9b95/lib/web_ui/lib/src/engine.dart#L178-L184). But that's only one example that requires clean up, there may be others.

One thing you can do is make the listening lazy. So you only attach a listener when someone subscribes to `_connectivityResult.stream`. And as soon as there are no more subscribers to `_connectivityResult.stream` you can release your listener. Doing it this way ensures that your code won't leak anything as long as the user code isn't leaking the stream subscription.

ditman: I see what you mean, I'm going to create an issue, because this is not exclusive of this plugin!

ditman: There are *indeed* listeners leaking. Two problems:

* The hot restart never calls "dispose" on the main app, so the client never detaches from the stream (this can be worked around on the app side)
* Even if I manually cancel the subscription to the outermost Stream, the innermost stream is never canceled (this is the bug that you were mentioning)

I'm going to fix the current asymmetry: subscribe to the inner Stream only when there's a new listener, and unsubscribe from it when there are none left.

Thanks for bringing this up!

ditman: I've created a Dart issue regarding the first issue: https://github.com/dart-lang/sdk/issues/42679

Without knowing when to call "dispose", we can't clear up the Stream.

I've reverted that part of the code to write directly to the "onchange" DOM property of the object, until we get a clear way of cleaning up after ourselves (from the plugin)

PR #2870, opened 2020-07-11
[camera] Support JPEG fallback for Android on ImageStream

OgulcanCelik:
## Description

Currently, ImageStream provides `YUV_420_888` on Android and `FLEX_RGBA_8888` iOS. However, YUV is a bit problematic, since it's really heavy in terms of performance cost when converting YUV to RGB plane & even default libraries like Image does not support YUV image processing.

Android's ImageReader does not support `FLEX_RGBA_8888` but it does support JPEG as ImageFormat. JPEG is already usable by default widgets like `Image.memory()` so it opens more possibilities to ImageStream on Android, compared to YUV.

So this helps with real-time processing use cases, Dart Image library already supports JPEG decoding and its much faster than trying to convert YUV planes to RGB. 

This PR adds a new optional parameter for `CameraController` to specify which ImageFormat wanted from ImageStream. Since it is optional and the default one is still `YUV_420_888`, its not a breaking change.

## Related Issues

https://github.com/flutter/flutter/issues/26348
https://github.com/flutter/flutter/issues/18968
https://github.com/flutter/flutter/issues/34743

and lots of questions on StackOverflow regarding how to use CameraImage on Android & how to convert YUV images to RGB

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/lib/camera_image.dart
+++ b/packages/camera/lib/camera_image.dart
@@ -61,7 +61,15 @@ enum ImageFormatGroup {
   ///
   /// On iOS, this is `kCVPixelFormatType_32BGRA`. See
   /// https://developer.apple.com/documentation/corevideo/1563591-pixel_format_identifiers/kcvpixelformattype_32bgra?language=objc
+  ///
   bgra8888,
+
+  /// 32-big RGB image encoded into JPEG bytes.
+  ///
+  /// On Android, this is `android.graphics.ImageFormat.JPEG`. See
+  /// https://developer.android.com/reference/android/graphics/ImageFormat#FLEX_RGBA_8888
+  ///
+  jpeg
 }
 
 /// Describes how pixels are represented in an image.
@@ -84,8 +92,10 @@ class ImageFormat {
 ImageFormatGroup _asImageFormatGroup(dynamic rawFormat) {
   if (defaultTargetPlatform == TargetPlatform.android) {
     // android.graphics.ImageFormat.YUV_420_888
-    if (rawFormat == 35) {
-      return ImageFormatGroup.yuv420;
+    if (rawFormat == 256) {
+      return ImageFormatGroup.jpeg;
+    } else if (rawFormat == 42) {
+      return ImageFormatGroup.bgra8888;
     }
   }
 
--- a/packages/camera/lib/camera_image.dart
+++ b/packages/camera/lib/camera_image.dart
@@ -67,7 +67,7 @@ enum ImageFormatGroup {
   /// 32-big RGB image encoded into JPEG bytes.
   ///
   /// On Android, this is `android.graphics.ImageFormat.JPEG`. See
-  /// https://developer.android.com/reference/android/graphics/ImageFormat#FLEX_RGBA_8888
+  /// https://developer.android.com/reference/android/graphics/ImageFormat#JPEG
   ///
   jpeg
 }
@@ -91,11 +91,13 @@ class ImageFormat {
 
 ImageFormatGroup _asImageFormatGroup(dynamic rawFormat) {
   if (defaultTargetPlatform == TargetPlatform.android) {
-    // android.graphics.ImageFormat.YUV_420_888
-    if (rawFormat == 256) {
-      return ImageFormatGroup.jpeg;
-    } else if (rawFormat == 42) {
-      return ImageFormatGroup.bgra8888;
+    switch (rawFormat) {
+      // android.graphics.ImageFormat.JPEG
+      case 256:
+        return ImageFormatGroup.jpeg;
+      // android.graphics.ImageFormat.YUV_420_888
+      case 42:
+        return ImageFormatGroup.yuv420;
     }
   }
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,3 +59,4 @@ Kazuki Yamaguchi <y.kazuki0614n@gmail.com>
 Eitan Schwartz <eshvartz@gmail.com>
 Chris Rutkowski <chrisrutkowski89@gmail.com>
 Juan Alvarez <juan.alvarez@resideo.com>
+Ogulcan Celik <ogulcancelik@gmail.com>
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,185 +1,188 @@
+## 0.5.9+4
+
+- Android: Added support for JPEG callback in image stream.
+
 ## 0.5.8+3
 
-* Fix bug in usage example in README.md 
+- Fix bug in usage example in README.md
 
 ## 0.5.8+2
 
-* Post-v2 embedding cleanups.
+- Post-v2 embedding cleanups.
 
 ## 0.5.8+1
 
-* Update lower bound of dart dependency to 2.1.0.
+- Update lower bound of dart dependency to 2.1.0.
 
 ## 0.5.8
 
-* Remove Android dependencies fallback.
-* Require Flutter SDK 1.12.13+hotfix.5 or greater.
+- Remove Android dependencies fallback.
+- Require Flutter SDK 1.12.13+hotfix.5 or greater.
 
 ## 0.5.7+5
 
-* Replace deprecated `getFlutterEngine` call on Android.
+- Replace deprecated `getFlutterEngine` call on Android.
 
 ## 0.5.7+4
 
-* Add `pedantic` to dev_dependency.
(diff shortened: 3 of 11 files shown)


OgulcanCelik:
@bparrishMines hello, sorry for mention but I believe this is somewhat an urgent issue, it's critical for lots of image process-related apps do you have time for a quick review? Also if you are not accepting feature requests via PRs, may I publish a forked version of camera as a new package, including this PR feature? -Since it will be based on your camera plugin codebase-

bparrishMines:
Hi @OgulcanCelik 

Thanks for the contribution! Unfortunately, the ecosystem team is currently focused on other priorities at the moment and we don't anyone that can focus on new camera features right now. We'll typically follow this [prioritization policy](https://github.com/flutter/flutter/wiki/Issue-hygiene#priorities). In the meantime, you can always add this branch to your the `pubspec.yaml` of a Flutter app:

```yaml
dependencies:
  camera:
    git:
      url: git@github.com:OgulcanCelik/plugins.git
      path: packages/camera
```

I would only recommend publishing your own plugin if you intend to continue to maintain it. Ultimately, it's up to you though.

mario-jerkovic:
@OgulcanCelik did you decide if you want to publish this fork as a separate plugin :) as this PR is awesome and it would be a waste not to release it 😃

OgulcanCelik:
> @OgulcanCelik did you decide if you want to publish this fork as a separate plugin :) as this PR is awesome and it would be a waste not to release it

Hey, I don't have time to publish this as a separate plugin because that would mean I have to also maintain it :) Eventually, dev team going to implement something related to this issue I hope, but until then you can use my fork for the camera I think.

Nevertheless, its shame that dev team cant find neither time nor a developer to look for their official plugin development, this isn't a big complicated issue to fix.

davidmartos96:
@OgulcanCelik Nice PR! I am no Flutter reviewer, but I think it would be nice if instead of using the magic numbers 35, 42 and 256 throughout the code, use the static constants provided by the ImageFormat class. https://developer.android.com/reference/android/graphics/ImageFormat#YUV_420_888

`35 == ImageFormat.YUV_420_888`

alexcohn:
@OgulcanCelik which device did you use to develop and test this change? While Android camera does mandate support for JPEG, there is [no such requirement to support JPEG streaming](https://source.android.com/devices/camera/stream-config#required_use_cases).

packages/camera/lib/camera_image.dart
```diff
@@ -61,7 +61,15 @@ enum ImageFormatGroup {
   ///
   /// On iOS, this is `kCVPixelFormatType_32BGRA`. See
   /// https://developer.apple.com/documentation/corevideo/1563591-pixel_format_identifiers/kcvpixelformattype_32bgra?language=objc
+  ///
   bgra8888,
+
+  /// 32-big RGB image encoded into JPEG bytes.
+  ///
+  /// On Android, this is `android.graphics.ImageFormat.JPEG`. See
+  /// https://developer.android.com/reference/android/graphics/ImageFormat#FLEX_RGBA_8888
```
alexcohn, line 70: You probably mean, `https://developer.android.com/reference/android/graphics/ImageFormat#JPEG`

packages/camera/lib/camera_image.dart
```diff
@@ -84,8 +92,10 @@ class ImageFormat {
 ImageFormatGroup _asImageFormatGroup(dynamic rawFormat) {
   if (defaultTargetPlatform == TargetPlatform.android) {
     // android.graphics.ImageFormat.YUV_420_888
-    if (rawFormat == 35) {
-      return ImageFormatGroup.yuv420;
+    if (rawFormat == 256) {
+      return ImageFormatGroup.jpeg;
+    } else if (rawFormat == 42) {
+      return ImageFormatGroup.bgra8888;
```
alexcohn, line 98: This contradicts [Camera.java:91](https://github.com/flutter/plugins/blob/cd7b678ff548c93fe377c5d16dede2210a959917/packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java#L91): there, you don't allow **bgra8888** format for Android, but only **jpeg** or **yuv420**.

joaoavf:
I wasn't able to get a jpeg stream in two different devices: Xiaomi Mi Lite 2 and Motorola G7 Play. 

The stream keeps returning `YUV_420_888 (35)` despite the CameraControler being set to `jpeg (256)`.

Has anyone been able to implement this properly?

alexcohn:
As I wrote above, it is impossible to get a JPEG stream from the camera (except that there may exist devices that support this). What's possible, is to convert the preview frames to JPEG in Java and pass them to Flutter. The problem is that I cannot imagine a use case when this kind of image stream may be helpful. Can you?

joaoavf:
My specific use case is reading a code (similar to a barcode), but as the code is somewhat hard to read (light orange on a white background), I have to make quite a few retries until getting it right, and being able to make a lot of retries in a short amount of time greatly improves the user experience.

The current bottleneck I have is converting from YUV to an [Image](https://pub.dev/packages/image) object, which is somewhat expensive in the [code I am currently using](https://github.com/Hugand/camera_tutorial).

OgulcanCelik:
> @OgulcanCelik which device did you use to develop and test this change? While Android camera does mandate support for JPEG, there is [no such requirement to support JPEG streaming](https://source.android.com/devices/camera/stream-config#required_use_cases).


> I wasn't able to get a jpeg stream in two different devices: Xiaomi Mi Lite 2 and Motorola G7 Play.
> 
> The stream keeps returning `YUV_420_888 (35)` despite the CameraControler being set to `jpeg (256)`.
> 
> Has anyone been able to implement this properly?

After the official response from the dev team, I left Flutter and went with React Native instead. Especially the official camera plugin missing so many important endpoints.

When made this PR, I tried it with Xiaomi Redmi 8, OnePlus 5T and Xiaomi Mi A2 Lite didn't notice any problem, so I'm not sure what's the issue, maybe this is the reason, but still:

> @OgulcanCelik which device did you use to develop and test this change? While Android camera does mandate support for JPEG, there is [no such requirement to support JPEG streaming](https://source.android.com/devices/camera/stream-config#required_use_cases).

https://developer.android.com/reference/android/graphics/ImageFormat#JPEG

Interesting, official documentation says `This format is always supported as an output format for the android.hardware.camera2 API, and as a picture format for the older Camera API` though, so this means all devices **_should_** feed JPEG right?

packages/camera/lib/camera_image.dart
OgulcanCelik: yeah, you are right, on first iteration, I thought about supporting bgra on both android and ios since its already on ios side, but then realized its better to feed jpeg, I missed this on the second iteration it seems. pushed new fix

packages/camera/lib/camera_image.dart
OgulcanCelik: fixed on caf092e

alexcohn:
@OgulcanCelik so why do you prefer JPEG to RGB (imagine for a moment that your device can deliver both)?

OgulcanCelik:
Actually I prefer RGB image feed with separate channels so I can manipulate the image easily but RGB didn't work properly so I went with jpeg, it worked on all 3 phones I tried. Also decoding jpeg isn't a hard task like converting YUV to RGB.

joaoavf:
Oğulcan, it is interesting to know that it is working on the Xiaomi Mi A2 Lite as it is the same model I am trying to use here.

I wonder what I might have set up wrong. The only change I have made was to add the `androidFormatCode: 256` parameter on `CameraController`. I am consuming from a stream created with `startImageStream`. Is there anything I've missed here? (It is still returning YUV instead of jpeg).

alexcohn:
@joaoavf From my measurements, the Image class is very slow. It's not surprising, because it's pure dart and manipulates byte lists. It can _decode_ Jpeg, but the overhead is more than building **Image** directly from **yuv420** in dart. At any rate, I have a [much faster way](https://github.com/alexcohn/YUV_2_RGB.git) to convert **CameraImage** to RGB (it does not build an [Image](https://github.com/brendan-duncan/image), but rather a **bmp**). I see that in your [experiment](https://github.com/joaoavf/ctt) you actually use luminance. This data you can get directly from the **CameraImage** (it's `planes.get(0)`).

mvanbeusekom:
this feature is implemented in PRs #3364 and #3359, therefore I will be closing this PR

PR #2873, opened 2020-07-14
[e2e] add support to report extra information

CareF:
## Description

Added an extra field `extraInfo` in `lib/common.dart`. Use case is to use the field to report performance. E.g. in flutter/flutter#61509, I'm migrating one existing test to an e2e based performance test. The use case is not complete because I still need to add other performance metrics but for a demo of the use case it's sufficient. 

I'm not naming it to performance because in some of our device lab test the reported value is not for performance. 

## Related Issues

Resolves flutter/flutter#61490

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/e2e/lib/common.dart
+++ b/packages/e2e/lib/common.dart
@@ -18,12 +18,12 @@ class Response {
   Map<String, dynamic> extraInfo;
 
   /// Constructor to use for positive response.
-  Response.allTestsPassed([this.extraInfo])
+  Response.allTestsPassed({this.extraInfo})
       : this._allTestsPassed = true,
         this._failureDetails = null;
 
   /// Constructor for failure response.
-  Response.someTestsFailed(this._failureDetails, [this.extraInfo])
+  Response.someTestsFailed(this._failureDetails, {this.extraInfo})
       : this._allTestsPassed = false;
 
   /// Whether the test ran successfully or not.
@@ -59,11 +59,11 @@ class Response {
     final dynamic detail = responseJson['failureDetails'];
     responseJson.remove('failureDetails');
     if (result == 'true') {
-      return Response.allTestsPassed(responseJson);
+      return Response.allTestsPassed(extraInfo: responseJson);
     } else {
       return Response.someTestsFailed(
         _failureDetailsFromJson(detail),
-        responseJson,
+        extraInfo: responseJson,
       );
     }
   }
--- a/packages/e2e/lib/common.dart
+++ b/packages/e2e/lib/common.dart
@@ -11,13 +11,20 @@ class Response {
 
   final bool _allTestsPassed;
 
+  /// The extra information to be added along side the test result.
+  ///
+  /// Specially `'result'` and `'failureDetails'` keys are occupied, and will
+  /// be ignored.
+  Map<String, dynamic> extraInfo;
+
   /// Constructor to use for positive response.
-  Response.allTestsPassed()
+  Response.allTestsPassed([this.extraInfo])
       : this._allTestsPassed = true,
         this._failureDetails = null;
 
   /// Constructor for failure response.
-  Response.someTestsFailed(this._failureDetails) : this._allTestsPassed = false;
+  Response.someTestsFailed(this._failureDetails, [this.extraInfo])
+    : this._allTestsPassed = false;
 
   /// Whether the test ran successfully or not.
   bool get allTestsPassed => _allTestsPassed;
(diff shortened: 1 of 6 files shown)


CareF:
It might be good to put some part of https://github.com/CareF/flutter/blob/f7f4471caa5f714f29e3c9e331f2d0d84b220cfc/dev/benchmarks/macrobenchmarks/test/util.dart to E2E, but I'm not sure for that.

packages/e2e/lib/common.dart
```diff
@@ -11,13 +11,20 @@ class Response {
...
 
+  /// The extra information to be added along side the test result.
+  ///
+  /// Specially `'result'` and `'failureDetails'` keys are occupied, and will
+  /// be ignored.
+  Map<String, dynamic> extraInfo;
+
   /// Constructor to use for positive response.
-  Response.allTestsPassed()
+  Response.allTestsPassed([this.extraInfo])
```
dnfield, line 21: Prefer optional named parameters to optional position parameters. They get really hard to think about as you add more :)

packages/e2e/lib/common.dart
```diff
@@ -30,19 +37,34 @@ class Response {
...
   /// Serializes this message to a JSON map.
-  String toJson() => json.encode(<String, dynamic>{
-        'result': allTestsPassed.toString(),
-        'failureDetails': _failureDetailsAsString(),
+  String toJson() {
+    final Map<String, dynamic> result = <String, dynamic>{};
+    result['result'] = allTestsPassed.toString();
+    result['failureDetails'] = _failureDetailsAsString();
+    if (extraInfo != null) {
+      extraInfo.forEach((String key, dynamic value) {
```
dnfield, line 45: Why don't we just put this in an additional key in the map, to avoid overwriting or special casing any key names?

packages/e2e/lib/common.dart
CareF: Done!

packages/e2e/lib/common.dart
CareF: I'm feeling `extraInfo` as a variable name is fine but as a key in json it's too colloquial. Maybe I should call it `report`.

dnfield: How about just 'data'?

CareF: Yeah that's simpler. Done

dnfield:
Can you add some test(s) for this?

CareF:
> Can you add some test(s) for this?

@dnfield The test would require running on a device. I think I can add something the `example` as test, though I don't think it will be tracked by CI.

dnfield:
You should be able to write some unit tests right?

CareF:
> You should be able to write some unit tests right?

Currently we don't have any unit test for E2E, and I kind of believe it's because this package is all about adapting for running tests for different ways. But yeah it's probably the good point to start unit test here.

CareF:
@dnfield Unit test done. I'm a little bit proud of coming up a way to unit test the binding.

packages/e2e/lib/e2e.dart
```diff
@@ -78,33 +78,54 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
...
+  /// The values in `data` should be json-serializable objects.
+  ///
+  /// For keys in `data` that's previously added, it will over write the
+  /// original values.
+  void addReportData(Map<String, dynamic> data) {
+    _data ??= <String, dynamic>{};
+    data.forEach((String key, dynamic value) {
+      _data[key] = value;
+    });
+  }
```
dnfield, line 94: This seems potentially confusing for users. Why not just expose the dictionary directly?

For example, how do I know if I'm overwriting a key? What if I wanted to keep the original value?

CareF: I don't like exposing the dictionary for adding data because Dart doesn't provide an API to merge Map. If the way to write the data is through the dictionary itself, users will need to write an extra for loop to add things like a timeline (which is a `Map<String, dynamic>`). For me the most common use case would be to construct a json object and pass it to some method all at once. If over writing is the concern I would rather alway erase the data and rewrite everything, and expose only a getter.

CareF: Or, to throw if there's key conflict, and add a `clearData` api to erase `_data`.

CareF: With a double think I changed my mind.

packages/e2e/lib/e2e.dart
```diff
@@ -78,33 +79,47 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
 
   static Map<String, String> _results = <String, String>{};
 
+  /// The extra data for the reported result.
+  ///
+  /// The values in `reportData` should be json-serializable objects or `null`.
```
dnfield, line 84: nit: should -> must.

CareF: Done

packages/e2e/lib/common.dart
```diff
@@ -33,16 +37,21 @@ class Response {
...
       });
 
   /// Deserializes the result from JSON.
   static Response fromJson(String source) {
-    Map<String, dynamic> responseJson = json.decode(source);
-    if (responseJson['result'] == 'true') {
-      return Response.allTestsPassed();
+    final Map<String, dynamic> responseJson = json.decode(source);
+    final String result = responseJson['result'] as String;
+    final dynamic detail = responseJson['failureDetails'];
```
liyuqian, line 47: Nit: the `detail` seems unnecessary as it's used only once below.

CareF: Done

PR #2892, opened 2020-07-23
pictureImageReader null check

emreuguryalcintr:
On some android devices,  I got an error because of Camera plugin. When we checked the error , we found it in this path : plugins\packages\camera\android\src\main\java\io\flutter\plugins\camera\camera.java

In the camera.java file startPreview caused an error because of pictureImageReader. We made null check for pictureImageReader , after that we never got that error. 

I send a commit to check null for pictureImageReader.

### DIFF
--- a/packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
+++ b/packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
@@ -421,7 +421,11 @@ public class Camera {
   }
 
   public void startPreview() throws CameraAccessException {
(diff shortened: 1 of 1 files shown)


emreuguryalcintr:
@googlebot I signed it!

emreuguryalcintr:
@googlebot I fixed it.

emreuguryalcintr:
@googlebot I fixed it.

emreuguryalcintr:
@googlebot I fixed it.

mvanbeusekom:
Closing this PR as this is a duplicate of #3127 which is merged and published.

PR #2845, opened 2020-06-24
[path_provider] Use the application ID in the application support path

robert-ancell:
## Description

Use the application ID in the application support path. The existing code uses the binary name, which is not necessarily unique. Application IDs were recently added to the Linux shell.

## Related Issues

https://github.com/flutter/flutter/issues/59234

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [x] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [ ] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #2896, opened 2020-07-24
Revert "[path_provider] Use the application ID in the application support path"

dnfield:
Reverts flutter/plugins#2845

This broke master - Linux is consistently timing out after this.

TBR @robert-ancell @stuartmorgan 

/cc @ditman

### DIFF
--- a/packages/path_provider/path_provider/CHANGELOG.md
+++ b/packages/path_provider/path_provider/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.6.12
+
+* Fixed a Java lint in a test.
+
 ## 1.6.11
 * Updated documentation to reflect the need for changes in testing for federated plugins
 
--- a/packages/path_provider/path_provider/example/android/app/src/androidTest/java/MainActivityTest.java
+++ b/packages/path_provider/path_provider/example/android/app/src/androidTest/java/MainActivityTest.java
@@ -8,7 +8,7 @@ import org.junit.Rule;
 import org.junit.runner.RunWith;
 
 @RunWith(FlutterRunner.class)
-public class FlutterActivityTest {
+public class MainActivityTest {
   @Rule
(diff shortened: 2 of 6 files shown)


dnfield:
I will wait for the test that's been timing out to pass before merging this.

ditman:
path_provider_linux 0.1.0 seems to [NOT have been published yet](https://pub.dev/packages/path_provider_linux/versions), so this looks safe to land (then reland the fixed 0.1.0)

dnfield:
Hm. I'm not sure if this will make the version check permanently angry?

ditman:
You may be able to "revert" to `0.1.0+1` (with a CHANGELOG entry), then the original code can be re-landed as 0.2.0?

dnfield:
That seems pretty wrong to me - we'd never want to publish those versions, and users are likely to get confused when reading the CHANGELOG :\

ditman:
@dnfield when we publish the 0.2.0 we can erase all mentions to 0.1.x changelog, as if this never happened :P

dnfield:
Given that this has broken the tree, going to land - we can always reland it when it's ready.

stuartmorgan:
> Go for this! Maybe get an ACK from Stuart that this is being reverted?

Noted. I'm actually not sure the PR should have landed before https://github.com/flutter/flutter/pull/60053 anyway.

robert-ancell:
> Noted. I'm actually not sure the PR should have landed before [flutter/flutter#60053](https://github.com/flutter/flutter/pull/60053) anyway.

This change is only supposed to work if an application ID is set, and reverts to the old behaviour if not, so PRs are independent.

robert-ancell:
@dnfield do you have a link to the log of the failure?

dnfield:
https://cirrus-ci.com/task/6312004510023680

PR #2897, opened 2020-07-24
File Selector Plugin

jasonpanelli:
## Description

*Replace this paragraph with a description of what this PR is doing. If you're modifying existing behavior, describe the existing behavior, how this PR is changing it, and what motivated the change.*

## Related Issues

*Replace this paragraph with a list of issues related to this PR from the [issue database](https://github.com/flutter/flutter/issues). Indicate, which of these issues are resolved or fixed by this PR. Note that you'll have to prefix the issue numbers with flutter/flutter#.*

## Checklist

- [ ] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [ ] I signed the [CLA].

## Breaking Change


jasonpanelli:
@tugorez Hello! Please take a look

packages/file_picker/file_picker_web/lib/file_picker_web.dart
```diff
@@ -0,0 +1,109 @@
...
+    String inputString = '';
+    for (String element in acceptedTypes) {
+      if (inputString.isNotEmpty) {
+        inputString += ',';
+      }
+      inputString += element;
+    }
+    
+    // Create a file input element
+    html.FileUploadInputElement element = html.FileUploadInputElement();
```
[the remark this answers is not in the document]
jasonpanelli: I updated this to final. I thought final means I cannot change the element anymore, why can I still set the value of element.accept? element.accept is considered separate from the class containing it?

jasonpanelli:
@tugorez Ready for some review.

Missing tests only now

packages/file_selector/file_selector_platform_interface/lib/src/types/x_file/html.dart
```diff
@@ -0,0 +1,131 @@
...
+    return encoding.decode(await _bytes);
+  }
+
+  @override
+  Future<Uint8List> readAsBytes() async {
+    return Future.value(await _bytes);
+  }
+
+  @override
+  Stream<Uint8List> openRead([int start, int end]) async* {
```
CDDelta, line 89: @jasonpanelli it would be awesome to see [support for streaming in the file data](https://joji.me/en-us/blog/processing-huge-files-using-filereader-readasarraybuffer-in-web-browser/) as opposed to reading it all into memory first here!

CDDelta (commented):
Maybe it's out of scope for the first PR but being able to stream in file data on web as opposed to having it all loaded into memory first would be great!

jasonpanelli:
@ditman @tugorez Now should be consistent with discussed API

tugorez:
@jasonpanelli  analyze is complaining about multiple issues. [PTAL](https://github.com/flutter/plugins/pull/2897/checks?check_run_id=1034089498).

You can get the same report locally by typing
```
flutter analyze
```

packages/file_selector/file_selector/test/file_selector_test.dart
```diff
@@ -0,0 +1,75 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter/material.dart';
+import 'package:mockito/mockito.dart';
+import 'package:plugin_platform_interface/plugin_platform_interface.dart';
+import 'package:file_selector/file_selector.dart';
+import 'package:file_selector_platform_interface/file_selector_platform_interface.dart';
+
+void main() {
+  final MockFileSelector mock = MockFileSelector();
+  FileSelectorPlatform.instance = mock;
```
tugorez, line 15: This type can be inferred by the compiler so you can just omit it.

packages/file_selector/file_selector_platform_interface/lib/src/types/x_file/html.dart
ditman: This is inherited from an old implementation I did a while back, and I agree this'd be a great change. The plan for XFile is to eventually become its own package, so we'll be able to add these improvements later!

packages/file_selector/file_selector/README.md
```diff
@@ -0,0 +1,3 @@
+# file_picker
+
+A Flutter plugin that currently doesn't do anything.
```
ditman, line 3: 🤣 👏

packages/file_selector/file_selector_platform_interface/lib/src/method_channel/method_channel_file_selector.dart
```diff
@@ -0,0 +1,65 @@
...
+/// An implementation of [FileSelectorPlatform] that uses method channels.
+class MethodChannelFileSelector extends FileSelectorPlatform {
+  /// Load a file from user's computer and return it as an XFile
+  @override
+  Future<XFile> loadFile({
+    List<XTypeGroup> acceptedTypeGroups,
+    String initialDirectory,
+    String confirmButtonText,
+  }) {
+    return _channel.invokeMethod<XFile>(
```
ditman, line 23: This will probably not work, because the method channels cannot transport "complex" responses without specifying a custom Binary messenger (which... I don't want to do :P).

Most likely this will have to do:

```dart
final path = await _channel.invokeMethod<String>('loadFile'...)
return XFile(path: path);
```

packages/file_selector/file_selector_platform_interface/lib/src/types/x_file/html.dart
```diff
@@ -0,0 +1,161 @@
...
+    var target = querySelector('#${id}');
+    if (target == null) {
+      final Element targetElement =
+      Element.tag('flt-x-file')..id = id;
+
+      querySelector('body').children.add(targetElement);
+      target = targetElement;
+    }
+    return target;
+  }
```
ditman, line 149: Can this bunch of methods live outside of the XFile itself? They look like bare helpers, since they don't need "this" for anything :)

packages/file_selector/file_selector_platform_interface/lib/src/types/x_file/html.dart
```diff
@@ -0,0 +1,161 @@
...
+  Stream<Uint8List> openRead([int start, int end]) async* {
+    final bytes = await _bytes;
+    yield bytes.sublist(start ?? 0, end ?? bytes.length);
+  }
+
+  /// Saves the data of this XFile at the location indicated by path.
+  /// For the web implementation, the path variable is ignored.
+  void saveTo(String path) async {
+    // Create a DOM container where we can host the anchor.
+    _target = _ensureInitialized('__x_file_dom_element');
```
ditman, line 104: This looks very similar to the `load` code as well, some more reason to split it (and maybe reuse it)

packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
```diff
@@ -0,0 +1,53 @@
...
+    this.webWildCards,
+  }) {
+    if (
+      this.extensions == null &&
+      this.mimeTypes == null &&
+      this.macUTIs == null &&
+      this.webWildCards == null
+    ) {
+      throw ArgumentError("At least one type must be provided for an XTypeGroup.");
+    }
```
ditman, line 35: You may be able to do this through an assertion in the initialization list of a constructor without body:

```dart
XTypeGroup({
  this.label,
  this.extensions,
  ...
}) : assert(!(extensions == null && mimeTypes == null && ...), 'At least one type must be provided for an XTypeGroup');
```

See more: https://dart.dev/guides/language/language-tour#initializer-list

packages/file_selector/file_selector_web/lib/file_selector_web.dart
```diff
@@ -0,0 +1,222 @@
...
+/// This class implements the `package:file_selector` functionality for the web.
+class FileSelectorPlugin extends FileSelectorPlatform {
+  
+  /// Open file dialog for loading files and return a XFile
+  @override
+  Future<XFile> loadFile({
+    List<XTypeGroup> acceptedTypeGroups,
+    String initialDirectory,
+    String confirmButtonText,
+  }) {
```
ditman, line 22: This function should probably be marked as `async`, so you can use `async/await` inside it, without worrying about having to use futures with `then` which is a little bit more cumbersome.

packages/file_selector/file_selector_web/lib/file_selector_web.dart
```diff
@@ -0,0 +1,222 @@
...
+    return _loadFileHelper(true, acceptedTypeGroups);
+  }
+
+  @override
+  Future<String> getSavePath({
+    List<XTypeGroup> acceptedTypeGroups,
+    String initialDirectory,
+    String suggestedName,
+    String confirmButtonText,
+  }) => Future.value();
```
ditman, line 50: If you mark this function as `async` you can just `=> null`, and the language will understand that it's a "Future<String>".

packages/file_selector/file_selector_web/lib/file_selector_web.dart
```diff
@@ -0,0 +1,222 @@
...
+
+      // Create XFile from dart:html Files
+      final xFiles = _getXFilesFromFiles(files);
+
+      _completer.complete(xFiles);
+    });
+
+    element.onError.first.then((event) {
+      if (!_completer.isCompleted) {
+        _completer.completeError(event);
```
ditman, line 175: Instead of completing with the `event`, this probably should extract information from the error, and wrap it in a `PlatformException`.

packages/file_selector/file_selector_web/test/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
...
+  file_selector_web:
+    path: ../
+  file_selector_platform_interface:
+    path: ../../file_selector_platform_interface
+  google_maps: ^3.4.4
+  flutter_driver:
+    sdk: flutter
+  flutter_test:
+    sdk: flutter
+  e2e: ^0.6.1
```
ditman, line 21: Note: this is not called e2e anymore, it's now `integration_test`, and can (should) be loaded from the path. Take a look at the latest push I did for `google_maps_flutter_web` to "migrate" to the new package:

https://github.com/flutter/plugins/commit/a9ac52f3fe10034ef7f48a98f9112bca0666d3e9

ditman:
Apologies for the spam, I wasn't supposed to do a full review today, but I got dragged into the code :P

packages/file_selector/file_selector_web/lib/file_selector_web.dart
```diff
@@ -174,9 +175,10 @@ class FileSelectorPlugin extends FileSelectorPlatform {
...
 
     element.onError.first.then((event) {
-      if (!_completer.isCompleted) {
-        _completer.completeError(event);
-      }
+      throw PlatformException(
+          code: 'file_input',
+          message: "Input element failed on event with target: " +
+              event?.target?.toString());
     });
```
jasonpanelli, line 182: @ditman Unsure what information to provide here. Does this look okay? Also not sure what to provide for code parameter

packages/file_selector/file_selector_platform_interface/lib/src/method_channel/method_channel_file_selector.dart
```diff
@@ -0,0 +1,68 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/services.dart';
+
+import '../platform_interface/file_selector_interface.dart';
```
ditman, line 7: Unneeded import in line 7? The same is being imported in line 8, isn't it?

packages/file_selector/file_selector_platform_interface/lib/src/method_channel/method_channel_file_selector.dart
```diff
@@ -0,0 +1,68 @@
...
+      <String, Object>{
+        'acceptedTypes': acceptedTypeGroups,
+        'initialDirectory': initialDirectory,
+      },
+    );
+    List<XFile> fileList = List();
+    for (String path in pathList) {
+      fileList.add(XFile(path));
+    }
+    return fileList;
```
ditman, line 49: You may be able to replace this with the `map` function:

```dart
return pathList.map((path) => XFile(path)).toList();
```

packages/file_selector/file_selector_platform_interface/lib/src/method_channel/method_channel_file_selector.dart
```diff
@@ -0,0 +1,68 @@
...
+    List<XTypeGroup> acceptedTypeGroups,
+    String initialDirectory,
+    String confirmButtonText,
+  }) async {
+    String path = await _channel.invokeMethod<String>(
+      'loadFile',
+      <String, Object>{
+        'acceptedTypes': acceptedTypeGroups,
+        'initialDirectory': initialDirectory,
+      },
```
ditman, line 26: Do you think this will require a completely separate invokeMethod call to the native side? I think this would be something like:

```dart
_channel.invokeMethod<String>('loadFiles',
  <String, Object>{
        'acceptedTypes': acceptedTypeGroups,
        'initialDirectory': initialDirectory,
        'multiple': false, // boom
  },
```

We can have more methods in the API than in the native side, especially if we're adding them as a "convenience" for users (like in this case! :P)

packages/file_selector/file_selector_platform_interface/lib/src/platform_interface/file_selector_interface.dart
```diff
@@ -0,0 +1,66 @@
...
+    throw UnimplementedError('loadFile() has not been implemented.');
+  }
+
+  /// Open file dialog for loading files and return a list of file paths
+  Future<List<XFile>> loadFiles({
+    List<XTypeGroup> acceptedTypeGroups,
+    String initialDirectory,
+    String confirmButtonText,
+  }) {
+    throw UnimplementedError('loadFile() has not been implemented.');
```
ditman, line 54: loadFile**s**() has not been implemented.

packages/file_selector/file_selector_platform_interface/lib/src/platform_interface/file_selector_interface.dart
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /// Open file dialog for saving files and return a file path at which to save
+  Future<String> getSavePath({
+    List<XTypeGroup> acceptedTypeGroups,
+    String initialDirectory,
+    String suggestedName,
+    String confirmButtonText,
+  }) {
+    throw UnimplementedError('saveFile() has not been implemented.');
```
ditman, line 64: getSavePath() has not been implemented.

packages/file_selector/file_selector_web/lib/file_selector_web.dart
```diff
@@ -0,0 +1,205 @@
...
+
+  /// Convert list of XTypeGroups to a comma-separated string
+  String _getStringFromFilterGroup(List<XTypeGroup> acceptedTypes) {
+    List<String> allTypes = List();
+    for (XTypeGroup group in acceptedTypes ?? []) {
+      _verifyXTypeGroup(group);
+
+      for (String mimeType in group.mimeTypes ?? []) {
+        allTypes.add(mimeType);
+      }
```
ditman, line 95: You may be able to use `addAll` to add everything, instead of a for loop:

```dart
allTypes.addAll(group.mimeTypes ?? []);
```

packages/file_selector/file_selector_web/lib/file_selector_web.dart
```diff
@@ -0,0 +1,205 @@
...
+
+  /// Registers this class as the default instance of [FileSelectorPlatform].
+  static void registerWith(Registrar registrar) {
+    FileSelectorPlatform.instance = FileSelectorPlugin();
+  }
+
+  void _verifyXTypeGroup(XTypeGroup group) {
+    if (group.extensions == null &&
+        group.mimeTypes == null &&
+        group.webWildCards == null) {
```
ditman, line 81: What if these are not null, but they're empty arrays?

packages/file_selector/file_selector_web/lib/file_selector_web.dart
```diff
@@ -0,0 +1,205 @@
...
+          ext = '.' + ext;
+        }
+
+        allTypes.add(ext);
+      }
+      for (String webWildCard in group.webWildCards ?? []) {
+        allTypes.add(webWildCard);
+      }
+    }
+    return allTypes?.where((e) => e.isNotEmpty)?.join(',') ?? '';
```
ditman, line 108: Instead of having a `_verifyXTypeGroup` at the top, I'd do:

```dart
assert(allTypes.isNotEmpty, 'At least one of extensions / mimeTypes / webWildCards is required!');
```

packages/file_selector/file_selector_web/lib/file_selector_web.dart
```diff
@@ -0,0 +1,205 @@
...
+  List<XFile> _getXFilesFromFiles(List<File> files) {
+    List<XFile> xFiles = List<XFile>();
+
+    final timeZoneOffset = DateTime.now().timeZoneOffset;
+
+    for (File file in files) {
+      String url = Url.createObjectUrl(file);
+      String name = file.name;
+      int length = file.size;
+      DateTime modified = file.lastModifiedDate.add(timeZoneOffset);
```
ditman, line 137: is `file.lastModifiedDate` in UTC always?

packages/file_selector/file_selector_web/lib/file_selector_web.dart
ditman: Instead of throw, do:

```dart
_completer.completeError(PlatformException(...
```

if you then:

```
await getFilesWhenReady(...);
```

Whatever Error you've completed with will be thrown thrown for you. If people use the method as a traditional Future, they'll be able to catch the error manually.

Re: info to provide: try to extract information from the error event?

https://api.dart.dev/stable/2.9.2/dart-html/ErrorEvent-class.html

```dart
code: event.type,
message: event.message,
...
```

packages/file_selector/file_selector_platform_interface/lib/src/method_channel/method_channel_file_selector.dart
jasonpanelli: Good catch :) This is actually exactly what I did in the web implementation too!

packages/file_selector/file_selector_web/lib/file_selector_web.dart
jasonpanelli: I'm picturing this

`group = XTypeGroup(extensions: []);`

😐 😐 

Should I add this to XTypeGroup constructor assertion?

ditman: I think an empty XTypeGroup doesn't make too much sense (but checking only for null is not enough, you also need to see if it's empty, right?)

packages/file_selector/file_selector_web/lib/file_selector_web.dart
jasonpanelli: Yes, it was converting my local files to UTC. The timeZoneOffset fixed that

packages/file_selector/file_selector_platform_interface/lib/src/method_channel/method_channel_file_selector.dart
jasonpanelli: to be clear, this can then be updated in the platform interface too?

In file_selector:
`
loadFile(...) {
  return FileSelectorPlatform.instance.loadFile(... multiple: false, ...)
}
loadFiles(...) {
  return FileSelectorPlatform.instance.loadFile(... multiple: true, ...)
}
`

ditman: No, I think that if we have two separate methods in the plugin, we can have them in the interface too (they might share private code, though)

packages/file_selector/file_selector_web/test/pubspec.yaml
jasonpanelli: Why is `integration_test` also in `google_maps_flutter_web` pubspec.yaml instead of just in `google_maps_flutter_web/test`?

packages/file_selector/file_selector_web/lib/file_selector_web.dart
ditman: I just realized that `lastModifiedDate` is deprecated, and we should use `lastModified`:

https://developer.mozilla.org/en-US/docs/Web/API/File/lastModified

vs

https://developer.mozilla.org/en-US/docs/Web/API/File/lastModifiedDate

jasonpanelli:
I rebased so I could get integration_test from origin, so the comments are further up now than before and may be hiding in the "Show more items" option

packages/file_selector/file_selector_web/lib/file_selector_web.dart
jasonpanelli: Yes makes sense now 👍 
We need this assertion in both places and include null and empty checks

packages/file_selector/file_selector_web/lib/file_selector_web.dart
jasonpanelli: Good news, it appears that `lastModified` doesn't convert my local files to UTC, so this is easier anyway

packages/file_selector/file_selector_web/test/pubspec.yaml
ditman: The "publishable" check fails because it doesn't recognize google_maps_flutter_web/test as a separate package (currently, these tests are put inside an "example" directory, but most of the web packages don't have a stand-alone example)

stuartmorgan:
file_selector has landed, so this draft is obsolete; closing.

PR #2899, opened 2020-07-26
[image_picker] deprecate `retrieveLostData`

JarrodCColburn:
## Description

As method `ImagePicker.retrieveLostData()` returns a deprecated class (`LostDataResponse`), the method should also be annotated `@Deprecated`.
Advise using `getLostData` method instead (`retrieveLostData` -> `getLostData`), this is consistent with `pick_video` -> `get_video`, and `pick_image` -> `get_image`.

### DIFF
--- a/packages/image_picker/image_picker/CHANGELOG.md
+++ b/packages/image_picker/image_picker/CHANGELOG.md
@@ -1,8 +1,11 @@
+## 0.6.7+6
+
+* Deprecate method `retrieveLostData`.
+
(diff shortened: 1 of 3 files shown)


JarrodCColburn:
adding this simple change to #2900

shinriyo:
Because image_picker >=0.6.7 depends on image_picker_platform_interface ^1.1.0 which depends on http ^0.12.1, image_picker >=0.6.7 requires http ^0.12.1.

PR #2903, opened 2020-07-31
[google_maps_flutter(_platform_interface)] Make the plugin friendlier to the web implementation.

ditman:
## Description

This PR adds some small modifications to the core maps plugin and its platform interface, so the web version is doable.

It allows people using the experimental branch of google_maps_flutter_web to keep using it, without having to override many packages (and will also simplify the final merge of google_maps_flutter_web, so it only touches one package!)

This PR touches two packages, but both changes should be independent, and backwards compatible, so test should pass, and they can be merged and published in any order.

### google_maps_flutter

* Pass a constant `creationMapId` to `platform.buildView`, so web can return a cached widget DOM when flutter attempts to repaint there.
* Modify some examples slightly so they're more web-friendly (remove dart:io imports, pass a size to the bitmap used as a custom Icon, so it gets rendered at the correct size in higher DPI screens)

### google_maps_flutter_platform_interface

* Pass icon width/height if present on `fromAssetImage` BitmapDescriptors (web only, because web can't determine the actual size of an asset at run-time as easily as the native side).

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
+++ b/packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
@@ -10,6 +10,8 @@ part of google_maps_flutter;
 /// map is created.
 typedef void MapCreatedCallback(GoogleMapController controller);
 
+final _random = math.Random();
+
 /// A widget which displays a map with data obtained from the Google Maps service.
 class GoogleMap extends StatefulWidget {
   /// Creates a widget displaying data from Google Maps services.
@@ -205,6 +207,8 @@ class GoogleMap extends StatefulWidget {
 }
 
 class _GoogleMapState extends State<GoogleMap> {
+  final _creationMapId = _random.nextInt(4294967296); // 1<<32
+
   final Completer<GoogleMapController> _controller =
       Completer<GoogleMapController>();
 
@@ -223,7 +227,10 @@ class _GoogleMapState extends State<GoogleMap> {
       'polygonsToAdd': serializePolygonSet(widget.polygons),
       'polylinesToAdd': serializePolylineSet(widget.polylines),
       'circlesToAdd': serializeCircleSet(widget.circles),
+      'creationMapId': _creationMapId,
     };
+    // TODO: Can this call to buildView be moved to the initState method?
+    // That way we wouldn't need the _creationMapId
     return _googleMapsFlutterPlatform.buildView(
       creationParams,
       widget.gestureRecognizers,
--- a/packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
+++ b/packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
@@ -10,7 +10,11 @@ part of google_maps_flutter;
 /// map is created.
 typedef void MapCreatedCallback(GoogleMapController controller);
 
-final _random = math.Random();
+// This counter is used to provide a stable "constant" initialization id
+// to the buildView function, so the web implementation can use it as a
+// cache key. This needs to be provided from the outside, because web
+// views seem to re-render much more often that mobile platform views.
+int _web_only_platformId = 0;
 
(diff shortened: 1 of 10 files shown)


packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -223,7 +227,10 @@ class _GoogleMapState extends State<GoogleMap> {
       'polygonsToAdd': serializePolygonSet(widget.polygons),
       'polylinesToAdd': serializePolylineSet(widget.polylines),
       'circlesToAdd': serializeCircleSet(widget.circles),
+      'creationMapId': _creationMapId,
```
cyanglaz, line 230: Can web just use the platform view id instead?

ditman: These are the parameters that are passed in to create the view; look a little bit below to the call: `_googleMapsFlutterPlatform.buildView...`, that's where the view ID would be eventually created.

I tried to keep this fix self-contained to the web plugin, but I couldn't find how. I need a constant cache key for the widget and since the web view was being "recreated" all the time, I couldn't find *where* to get this value from (tried to get `key` from the widget or the parent, but if not passed those are "null").

I couldn't find a better fix than this one, given how the Map widgets are created  :/

ditman:
@cyanglaz and I discussed some changes in the chat, concerns about the creation map ID being a random number, and having a not very clear parameter name. I've updated the creation map ID to being an auto-incrementing int, and moving the creation parameter name to something more verbose/self explanatory.

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -10,6 +10,12 @@ part of google_maps_flutter;
 /// map is created.
 typedef void MapCreatedCallback(GoogleMapController controller);
 
+// This counter is used to provide a stable "constant" initialization id
+// to the buildView function, so the web implementation can use it as a
+// cache key. This needs to be provided from the outside, because web
+// views seem to re-render much more often that mobile platform views.
+int _web_only_platformId = 0;
```
cyanglaz, line 17: maybe `_webOnlyMapId`?

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -205,6 +211,8 @@ class GoogleMap extends StatefulWidget {
 }
 
 class _GoogleMapState extends State<GoogleMap> {
+  final _web_only_mapCreationId = _web_only_platformId++;
```
cyanglaz, line 214: _mapOnlyCreationId

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -223,7 +231,9 @@ class _GoogleMapState extends State<GoogleMap> {
       'polygonsToAdd': serializePolygonSet(widget.polygons),
       'polylinesToAdd': serializePolylineSet(widget.polylines),
       'circlesToAdd': serializeCircleSet(widget.circles),
+      '_web_only_mapCreationId': _web_only_mapCreationId,
```
cyanglaz, line 234: Can we test if this ID is created properly?

ditman: I'll look into adding a test for this, I can make the creation ID visibleForTesting and instantiate this twice. AFAIK there are no unit tests for this file at all in the package.

ditman: Added a new test for this. (I think this is the first test that mocks the platform interface, instead of the MethodChannels in this package)

PR #2593, opened 2020-03-10
[e2e] Adding failure details for driver tests

nturgut:
## Description

Adding failure details for driver tests.

**Before the change the failure output was something like:**
```
[+2616 ms] result fail
[  +88 ms] Stopping application instance.
[        ] Stopping application.
[   +2 ms] "flutter drive" took 22,861ms.

```
**The updated output:**
```
[+2782 ms] Failure Details:
[        ] Failure in method: verify text
[        ] ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞═════════════════
[   +1 ms] The following TestFailure object was thrown running a test:
[        ]   Expected: <3>
[        ]   Actual: <2>
[        ] When the exception was thrown, this was the stack:
[        ]     at Object.wrapException (http://localhost:36633/main.dart.js:3646:17)
[        ]     at Object.throwExpression (http://localhost:36633/main.dart.js:3659:15)
[        ]     at Object.fail (http://localhost:36633/main.dart.js:16457:16)
[        ]     at Object._expect (http://localhost:36633/main.dart.js:16454:9)
[        ]     at Object.expect0 (http://localhost:36633/main.dart.js:16430:9)
[        ]     at Object.expect (http://localhost:36633/main.dart.js:19782:9)
[        ]     at http://localhost:36633/main.dart.js:82397:17
[        ]     at _wrapJsFunctionForAsync_closure.$protected (http://localhost:36633/main.dart.js:6615:15)
[        ]     at _wrapJsFunctionForAsync_closure.call$2 (http://localhost:36633/main.dart.js:37105:12)
[        ]     at StackZoneSpecification__registerBinaryCallback__closure.call$0 (http://localhost:36633/main.dart.js:80439:21)
[        ] The test description was:
[        ]   verify text
[        ] ═════════════════════════════════════════════════════════════════
[        ] end of failure 1
[        ] Failure in method: another failure
[        ] ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞═════════════════
[        ] The following TestFailure object was thrown running a test:
[        ]   Expected: <3>
[        ]   Actual: <2>
[        ] When the exception was thrown, this was the stack:
[        ]     at Object.wrapException (http://localhost:36633/main.dart.js:3646:17)
[        ]     at Object.throwExpression (http://localhost:36633/main.dart.js:3659:15)
[        ]     at Object.fail (http://localhost:36633/main.dart.js:16457:16)
[...]

PR #2904, opened 2020-08-01
[web] Adding capability to e2e to take screenshot for web tests.

nturgut:
## Description

Adding capability to e2e to take screenshot for web tests. This infrastructure can also be used for triggering other WebDriver actions such as resize, back, forward. (https://www.w3.org/TR/webdriver/#endpoints)

Manually tested on Chrome, Firefox, Safari and Ios-Safari.

`e2e_driver_extended.dart` is only to be used as an example. We will use Flutter Gold to do the matching goldens for Flutter for Web integration tests.

## Related Issues

https://github.com/flutter/flutter/issues/51890
https://github.com/flutter/flutter/issues/62706

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/e2e/lib/common.dart
```diff
@@ -121,3 +131,57 @@ class Failure {
     return Failure(failure['methodName'], failure['details']);
   }
 }
+
+/// E2e web tests can execute WebDriver actions such as screenshots.
+///
+/// These test will use [TestStatus] to notify e2e of their state.
+enum TestStatus {
+  /// Test is waiting for executing WebDriver actions.
+  wait_on_webdriver_command,
```
dnfield, line 140: Use camelCase for enum values.

packages/e2e/lib/e2e.dart
```diff
@@ -29,6 +30,14 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
           if (!_allTestsPassed.isCompleted) {
             _allTestsPassed.complete(true);
           }
+
+          if (!webDriverActionPipe.isCompleted) {
+            webDriverActionPipe.complete(Future.value(WebDriverAction.noop()));
```
dnfield, line 35: Is the analyzer ok with this? I thought we had the lint turned ont hat would want this to be `Future<WebDriverAction>.value`.  Ditto below.

packages/e2e/lib/e2e.dart
```diff
@@ -128,9 +137,57 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
...
+  final Mutex webDriverCommandLock = new Mutex();
+
+  /// Takes screenshot using WebDriver screenshot command.
+  ///
+  /// Only works on Web when tests are run via `flutter driver` command.
+  ///
+  /// See: https://www.w3.org/TR/webdriver/#screen-capture
+  Future<void> takeScreenshot(String screenshot_name) async {
+    if (!kIsWeb) {
+      throw StateError('Web driver commands are only avaliable on web');
```
dnfield, line 159: I think this should probably be an `UnimplementedError`. This doesn't have to be web driver specific right?

packages/e2e/lib/e2e.dart
```diff
@@ -128,9 +137,57 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
...
+    }
+    await _webDriverCommand(WebDriverAction.screenshot(screenshot_name));
+  }
+
+  Future<void> _webDriverCommand(WebDriverAction command) async {
```
dnfield, line 164: Shouldn't this go in a web-specific implementation file, and we'd have another implementation file for IO, so we just call the method (or instantiate the class) and the VM version just throws for now?

packages/e2e/lib/e2e.dart
```diff
@@ -128,9 +137,57 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
...
+        if (!awaitCommand) {
+          throw Exception('Web Driver Command failed: ${command.type}');
+        }
+      } catch (e) {
+        throw Exception(
+            'Web Driver Command failed: ${command.type} with ' 'exception $e');
+      }
+    } finally {
+      // Reset the completer and release the lock.
+      driverActionComplete = Completer<bool>();
```
dnfield, line 180: Why are we reusing this? Why can't we just make a new one in this function and avoid the whole mutex business?

packages/e2e/lib/e2e.dart
```diff
@@ -184,4 +241,88 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
...
+  /// they are finished a web driver action
+  /// ([TestStatus.webdriver_command_complete])
+  ///
+  /// The tests running on the browser sends request these WebDriver operations
+  /// from e2e tests via replying with [WebDriverAction]s.
+  Future<Map<String, dynamic>> _callbackForWebDriver(
+      Map<String, String> params) async {
+    final String command = params['command'];
+    // Test status is added as an exta message.
+    final String extraMessage = params['message'] ?? '';
```
dnfield, line 261: Do we have this documented anywhere?

Would we be better off implementing some serialized type for this that could more clearly be used on both sides of the message?

packages/e2e/lib/e2e.dart
```diff
@@ -184,4 +241,88 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
...
+    Map<String, String> response;
+    switch (command) {
```
dnfield, line 263: IOW, can we avoid having this switch statement that will continue to grow, with nested `if`s underneath that will get harder to reason about?

packages/e2e/lib/e2e.dart
```diff
@@ -184,4 +241,88 @@ class E2EWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding {
...
+                ? Response.allTestsPassed(data: reportData).toJson()
+                : Response.someTestsFailed(
+                    _failureMethodsDetails,
+                    data: reportData,
+                  ).toJson(),
+          };
+        } else {
+          // If Test status is `wait_on_webdriver_command` send the first
+          // command in the `commandPipe` to the tests.
+          if (extraMessage == '${TestStatus.wait_on_webdriver_command}') {
```
dnfield, line 279: Why would the driving script do this? Shouldn't the test itself be awaiting whatever it needs (i.e. taking the screenshot), and the driving script just waits for the test reporting to finish?

Another way of putting this concern: how does the caller know which web driver command(s) it's waiting on, that there won't be another one it's actually interested in coming right after the one it just waited on?

packages/e2e/lib/e2e.dart
```diff
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 
+import 'package:mutex/mutex.dart';
```
yjbanov, line 7: I believe a mutex is essentially the same as a pool of size 1. Since we already  have a dependency on the `pool` package (I believe `flutter_tool` depends on it), we can continue using it rather than introduce another dependency.

packages/e2e/lib/common.dart
nturgut: done.

packages/e2e/lib/e2e.dart
nturgut: done.

packages/e2e/lib/e2e.dart
nturgut: done.

packages/e2e/lib/e2e.dart
nturgut: _Why would the driving script do this? Shouldn't the test itself be awaiting whatever it needs (i.e. taking the screenshot), and the driving script just waits for the test reporting to finish?_

Only driver script can take screenshots. Almost all browsers implements the webdriver API which allows us to take screenshot. We can also extend this to do other import things such as changing screensize, contorrling cookies: https://www.w3.org/TR/webdriver1/.  Since only driver test can perform these operations, we developed this handshake mechanism between the integration_test and the driver test. 

In summary: We are sending the request from the integration_test (ex: `WebDriverActionTypes.screenshot`), waiting on the result from the driver. If there are multiple test methods which are asking for a `WebDriverAction`, they wait on the mutex, driver handles them one by one.

_Another way of putting this concern: how does the caller know which web driver command(s) it's waiting on, that there won't be another one it's actually interested in coming right after the one it just waited on?_

That is why I used a synchronization method (such as mutex) to send these requests to driver **one at a time**. I tested with many scenarios:
1) requests in within different methods.
2) few request in a few methods.
3) Failing vs passing tests in different methods.

This handshake approach worked well.

packages/e2e/lib/e2e.dart
nturgut: You mean this package right:https://pub.dev/packages/pool (The one we used in [felt](https://github.com/flutter/engine/blob/master/lib/web_ui/dev/test_runner.dart#L517) :smile: )

Pool is more suitable when we have an operation where a Delegator distributes it to multiple Workers. Such as compiling tests with dart2js in web test harness. It is not suitable for used for mutual exclusion. Our case is basic resource sharing problem: where mutex controls access to a _critical section_.

packages/e2e/lib/e2e.dart
nturgut: The mutex makes sure only one methods executes this method:
```
    await webDriverCommandLock.acquire();

    try {
      webDriverActionPipe.complete(Future.value(command));
      try {
        final bool awaitCommand = await driverActionComplete.future;
        if (!awaitCommand) {
          throw Exception('Web Driver Command failed: ${command.type}');
        }
      } catch (e) {
        throw Exception(
            'Web Driver Command failed: ${command.type} with ' 'exception $e');
      }
    } finally {
      // Reset the completer and release the lock.
      driverActionComplete = Completer<bool>();
      webDriverCommandLock.release();
    }
```

I'm trying to avoid a race condition on `webDriverActionPipe.complete(Future.value(command));` for example.

packages/e2e/lib/e2e.dart
nturgut: I realized, there is no way (at least I couldn't manage) to run the tests methods in parallel when we run these tests with `flutter drive`, therefore I removed the mutex :)

packages/e2e/lib/e2e.dart
nturgut: I removed the mutex: https://github.com/flutter/plugins/pull/2904#discussion_r475828717

packages/e2e/lib/e2e.dart
nturgut: I carried the webdriver specific logic to another file. Please have another look.

packages/e2e/lib/e2e.dart
nturgut: We can add a string to the requestData call for managing generic communication. Documentation is here: https://github.com/flutter/flutter/blob/master/packages/flutter_driver/lib/src/driver/driver.dart#L494

Are you recommending to use another command other than `RequestData`. I think that would make the handshake more complicated.

packages/e2e/lib/e2e.dart
nturgut: I refactored the methods and separated the part for request_data into 2 methods. I hope it looks more readable now. PHAL.

packages/integration_test/lib/integration_test.dart
```diff
@@ -130,35 +132,62 @@ class IntegrationTestWidgetsFlutterBinding
...
+  /// Only works on Web when tests are run via `flutter driver` command.
+  /// See [WebDriverCommandManager].
+  final DriverCommandManager driverCommandManager =
+      driver_actions.driverCommandManager;
+
+  /// Taking a screenshot.
+  ///
+  /// Called by test methods. Implementation differs for each platform.
+  Future<void> takeScreenshot(String screenshot_name) async {
+    if (kIsWeb) {
```
yjbanov, line 151: A cleaner approach would be to implement `DriverCommandManager' for the io case. Then you wouldn't need the `if (kIsWeb)`. You'd just always call `driverCommandManager.takeScreenshot` and `driverCommandManager. callbackWithDriverCommands`, and let the platform-specific implementation do the right thing.

packages/integration_test/lib/common.dart
```diff
@@ -121,3 +132,105 @@ class Failure {
...
+  /// Asking WebDriver to take a screenshot of the Web page.
+  screenshot,
+}
+
+/// Command for WebDriver to execute.
+///
+/// Only works on Web when tests are run via `flutter driver` command.
+///
+/// See: https://www.w3.org/TR/webdriver/
+class WebDriverAction {
```
yjbanov, line 169: I would recommend that we not use the term "Action" to avoid confusion.

The WebDriver spec refers to taking screenshots as a "command", not "action" (see https://www.w3.org/TR/webdriver/#screen-capture). "Action" seems to have a different meaning for WebDriver: https://www.w3.org/TR/webdriver/#actions. In fact, the dartdoc for this class also call it "Command".

packages/integration_test/lib/common.dart
```diff
@@ -121,3 +132,105 @@ class Failure {
...
+  /// Constructor for [WebDriverActionTypes.noop] screenshot.
+  WebDriverAction.screenshot(String screenshot_name)
+      : this.type = WebDriverActionTypes.screenshot,
+        this.values = {'screenshot_name': screenshot_name};
+
+  /// Util method for converting [WebDriverActionTypes] to a map entry.
+  ///
+  /// Used for converting messages to json format.
+  static Map<String, dynamic> typeToMap(WebDriverActionTypes type) => {
+        'web_driver_action': '${type}',
```
yjbanov, line 197: Are we certain that in all modes `type.toString` will produce the original name of the command? I don't believe the Dart language spec guarantees that. For example, dart2js minifies the names (e.g. "screenshot" can become "aZ").

packages/integration_test/lib/common.dart
```diff
@@ -121,3 +132,105 @@ class Failure {
...
+      };
+}
+
+/// Template methods to implement for any class which manages communication
```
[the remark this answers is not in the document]
nturgut: Template method pattern from GoF's Design Patterns.

Yet, as we discussed offline `DriverCommandManager` is not the right abstraction. I'll rename the interface.

packages/integration_test/lib/common.dart
```diff
@@ -23,6 +24,16 @@ class Response {
   Response.someTestsFailed(this._failureDetails, {this.data})
       : this._allTestsPassed = false;
 
+  /// Constructor for failure response.
+  Response.toolException({String ex})
+      : this._allTestsPassed = false,
+        this._failureDetails = [Failure('ToolException', ex)];
```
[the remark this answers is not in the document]
nturgut: No it doesn't analyze is passing for this PR.

packages/integration_test/lib/common.dart
```diff
@@ -121,3 +132,105 @@ class Failure {
...
+  /// Test executed the previously requested action.
+  webdriverCommandComplete,
+}
+
+/// Types of different WebDriver actions that can be used in web integration
+/// tests.
+///
+/// These actions are either commands that WebDriver can execute or used
+/// for the communication between `integration_test` and the driver test.
+enum WebDriverActionTypes {
```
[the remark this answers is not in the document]
nturgut: In e2e, common is shared between driver side and the e2e side. They have to be public otherwise tests like these won't see them: https://github.com/flutter/plugins/blob/master/packages/integration_test/lib/integration_test_driver.dart#L72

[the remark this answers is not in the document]
nturgut: I added a comment to the top of the document (line 8-10)

packages/integration_test/lib/integration_test.dart
```diff
@@ -130,35 +132,62 @@ class IntegrationTestWidgetsFlutterBinding
...
+      await driverCommandManager.takeScreenshot(screenshot_name);
+    } else {
+      throw UnimplementedError(
+          'Screenshots are not implemented on this platform');
```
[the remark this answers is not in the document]
nturgut: It is outside of the scope of this work.

As discussed offline, they won't be using the WebDriver. They would be implemented by calling test plugin (which would probably use https://firebase.google.com/docs/test-lab/android/test-screenshots)

packages/integration_test/lib/integration_test_driver_extended.dart
```diff
@@ -0,0 +1,84 @@
+import 'dart:async';
+import 'dart:io';
+
+import 'common.dart';
+
+import 'package:flutter_driver/flutter_driver.dart';
+
+/// Example Integration Test which can also run WebDriver action depending on
```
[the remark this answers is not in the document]
nturgut: The example adoptor's are under the lib. Have you seen the other one: https://github.com/flutter/plugins/blob/master/packages/integration_test/lib/integration_test_driver.dart#L46

packages/integration_test/lib/integration_test_driver_extended.dart
```diff
@@ -0,0 +1,84 @@
...
+/// the requests coming from the test methods.
+Future<void> integrationDriver() async {
+  final FlutterDriver driver = await FlutterDriver.connect();
+
+  // Test states that it's waiting on web driver commands.
+  String jsonResponse = await driver.requestData(
+      '${TestStatus.waitOnWebdriverCommand}',
+      timeout: const Duration(seconds: 10));
```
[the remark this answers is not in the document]
nturgut: We have to use the `request_data` command. That is what integration_test.dart is expecting/responding. Since it only carries a `message` string we can't add anything more than a "magic string".  Note that, the design review for this handshake is passed in February. We could have write e2e in a different way if this comment was requested before all those tests are started. No, changing `request_data` would be a breaking change and will effect all the other tests so I consider that refactor a big project outside of the scope of this work.

The only non-breaking change we can make is: add a json instead of a string and serialize an object. Like we did back in this PR for the response: https://github.com/flutter/plugins/pull/2593/files#diff-e8b46d9b4e394d1ff162bfb77767c911R10

packages/integration_test/lib/integration_test_driver_extended.dart
```diff
@@ -0,0 +1,84 @@
...
+
+import 'package:flutter_driver/flutter_driver.dart';
+
+/// Example Integration Test which can also run WebDriver action depending on
+/// the requests coming from the test methods.
+Future<void> integrationDriver() async {
+  final FlutterDriver driver = await FlutterDriver.connect();
+
+  // Test states that it's waiting on web driver commands.
+  String jsonResponse = await driver.requestData(
```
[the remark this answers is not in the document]
nturgut: No this is not a breaking change. All existing tests are passing. This statement is false: " API for integration test purposes, because the developer may need to use requestData for their test-specific needs. If they try to use it, either they will break integration_test functionality, or integration_test will break their code." Please read the code here: https://github.com/flutter/plugins/blob/95476a5bb20d04ac85f87315c84bdaeee3962a1c/packages/integration_test/lib/_driver_commands_web.dart#L77

"Instead, integration_test should establish its own dedicated 2-way communication channel with the app side and use it exclusively to implement this protocol." --> This is exactly what is happening today.

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+    try {
+      webDriverActionPipe.complete(Future.value(command));
+      try {
+        final bool awaitCommand = await driverActionComplete.future;
+        if (!awaitCommand) {
+          throw Exception('Web Driver Command failed: ${command.type}');
+        }
+      } catch (e) {
+        throw Exception(
+            'Web Driver Command failed: ${command.type} with ' 'exception $e');
```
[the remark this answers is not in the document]
nturgut: The developer will never see the stack trace. That is one of the long standing issues with e2e tests. Any error happening here is lost.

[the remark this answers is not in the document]
dnfield: I believe @jiahaog is working on fixing that

packages/integration_test/lib/integration_test.dart
nturgut: As discussed offline, we can't add a `DriverCommandManager` for io. But we can change abstraction from `DriverCommandManager` to `CallbackManager`. Then there would be an io implementation as well.

packages/integration_test/lib/common.dart
nturgut: Interesting I usually run the tests in all modes. Didn't had an issue for now. What is the canonical way of getting a name for an enum like WebDriverActionTypes

packages/integration_test/example/test_driver/example_integration_extended.dart
```diff
@@ -0,0 +1,18 @@
+// This is a Flutter widget test can take a screenshot.
```
jiahaog, line 1: Should this file be moved to example/test instead of test_driver?

packages/integration_test/lib/common.dart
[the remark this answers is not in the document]
jiahaog: Do you think this entire file should be moved to lib/src to try and hide the symbols though? It's a bit weird that they will show up on the [API docs](https://pub.dev/documentation/integration_test/latest/common/common-library.html).

> I added a comment to the top of the document (line 8-10)

I can't find the comment, are you referring to this file?

packages/integration_test/lib/common.dart
```diff
@@ -121,3 +132,105 @@ class Failure {
...
+  ///
+  /// Failed test method's names used as key.
+  List<Failure> get failureMethodsDetails;
+
+  /// The extra data for the reported result.
+  Map<String, dynamic> get reportData;
+
+  /// Whether all the test methods completed succesfully.
+  Completer<bool> get allTestsPassed;
+}
```
jiahaog, line 236: I'm a bit apprehensive about exporting the methods here as part of the public API of IntegrationTestWidgetsFlutterBinding. There are some [plans](https://github.com/flutter/flutter/issues/50816) to change the test reporting API and it will be good to keep such changes internal, at least for now. What about changing this interface to something the following to limit the number of public methods we expose?

```
class ResponseProvider {
  Response get reponse;
}
```

 I think the Response object above has all the methods you need here

packages/integration_test/example/test_driver/example_integration_extended.dart
jiahaog: My mistake, this is correct

packages/integration_test/lib/common.dart
```diff
@@ -121,3 +132,105 @@ class Failure {
...
+  /// Used for converting messages to json format.
+  static Map<String, dynamic> typeToMap(WebDriverActionTypes type) => {
+        'web_driver_action': '${type}',
+      };
+}
+
+/// Template methods to implement for any class which manages communication
+/// between `integration_tests` and the `driver_tests`.
+///
+/// See example [WebDriverCommandManager].
```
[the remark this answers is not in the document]
nturgut: done.

packages/integration_test/lib/common.dart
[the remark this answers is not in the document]
nturgut: @jiahaog can you see it now?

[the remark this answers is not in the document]
nturgut: done.

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+      // Reset the completer and release the lock.
+      driverActionComplete = Completer<bool>();
+    }
+  }
+
+  /// The callback function to response the driver side input.
+  ///
+  /// Provides a handshake mechanism for executing [WebDriverAction]s on the
+  /// driver side.
+  Future<Map<String, dynamic>> callbackWithDriverCommands(
```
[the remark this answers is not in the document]
nturgut: done.

packages/integration_test/lib/common.dart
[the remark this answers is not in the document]
nturgut: I changed it to CallbackManager.

packages/integration_test/lib/_driver_commands_io.dart
```diff
@@ -0,0 +1,12 @@
...
+// found in the LICENSE file.
+
+import 'common.dart';
+
+/// The dart:io implementation of [driverCommandManager].
+///
+/// See also:
+///
+///  * [_driver_commands_web.dart], which has the dart:html implementation
+DriverCommandManager get driverCommandManager => null;
```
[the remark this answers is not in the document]
nturgut: I changed the design as we discussed offline. I made DriverCommandManager-> CallbackManager. I carried the io implementation to _callback_io file.

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+class WebDriverCommandManager extends DriverCommandManager {
+  /// Tests will put the action requests from WebDriver to this pipe.
+  Completer<WebDriverAction> webDriverActionPipe = Completer<WebDriverAction>();
+
+  /// Updated when WebDriver completes the request by the test method.
+  ///
+  /// For example, a test method will ask for a screenshot by calling
+  /// `takeScreenshot`. When this screenshot is taken [driverActionComplete]
+  /// will complete.
+  Completer<bool> driverActionComplete = Completer<bool>();
```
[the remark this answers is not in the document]
nturgut: done. good catch, leftover from the implementation time (I was using them with some tests).

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+
+  Future<void> _webDriverCommand(WebDriverAction command) async {
+    try {
+      webDriverActionPipe.complete(Future.value(command));
+      try {
+        final bool awaitCommand = await driverActionComplete.future;
+        if (!awaitCommand) {
+          throw Exception('Web Driver Command failed: ${command.type}');
+        }
+      } catch (e) {
```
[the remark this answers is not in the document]
nturgut: done

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+/// Enables Web Driver commands for `integration_tests`.
+///
+/// Manages communication between `integration_tests` and the `driver_tests`.
+///
+/// Tests can execute an Web Driver actions such as `screenshot` using browsers'
+/// WebDriver APIs.
+///
+/// See: https://www.w3.org/TR/webdriver/
+class WebDriverCommandManager extends DriverCommandManager {
+  /// Tests will put the action requests from WebDriver to this pipe.
```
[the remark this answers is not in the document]
nturgut: done

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+  /// Takes screenshot using WebDriver screenshot command.
+  ///
+  /// Only works on Web when tests are run via `flutter driver` command.
+  ///
+  /// See: https://www.w3.org/TR/webdriver/#screen-capture
+  Future<void> takeScreenshot(String screenshot_name) async {
+    await _webDriverCommand(WebDriverAction.screenshot(screenshot_name));
+  }
+
+  Future<void> _webDriverCommand(WebDriverAction command) async {
```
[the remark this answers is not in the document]
nturgut: done

packages/integration_test/lib/integration_test.dart
```diff
@@ -130,35 +132,62 @@ class IntegrationTestWidgetsFlutterBinding
...
+    }
+  }
 
-  /// the callback function to response the driver side input.
+  /// The callback function to response the driver side input.
```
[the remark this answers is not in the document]
nturgut: outside of the scope of the current pr.

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+        throw Exception(
+            'Web Driver Command failed: ${command.type} with ' 'exception $e');
+      }
+    } finally {
+      // Reset the completer and release the lock.
+      driverActionComplete = Completer<bool>();
+    }
+  }
+
+  /// The callback function to response the driver side input.
```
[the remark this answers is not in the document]
nturgut: skipped. kept same with the main comment.

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+  /// `takeScreenshot`. When this screenshot is taken [driverActionComplete]
+  /// will complete.
+  Completer<bool> driverActionComplete = Completer<bool>();
+
+  /// Takes screenshot using WebDriver screenshot command.
+  ///
+  /// Only works on Web when tests are run via `flutter driver` command.
+  ///
+  /// See: https://www.w3.org/TR/webdriver/#screen-capture
+  Future<void> takeScreenshot(String screenshot_name) async {
```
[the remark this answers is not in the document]
nturgut: done.

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+      try {
+        final bool awaitCommand = await driverActionComplete.future;
+        if (!awaitCommand) {
+          throw Exception('Web Driver Command failed: ${command.type}');
+        }
+      } catch (e) {
+        throw Exception(
+            'Web Driver Command failed: ${command.type} with ' 'exception $e');
+      }
+    } finally {
```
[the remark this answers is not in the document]
nturgut: done.

packages/integration_test/lib/common.dart
nturgut: done. renamed all the files.

packages/integration_test/lib/common.dart
nturgut: Unless there is an alternative of avoiding cycling dependency between CommandManager's and the binding I keep using this. All the methods are getter so the reader won't be able to access the values of the test results.

I think this is more of a language limitation, where we are missing a package level modifier.

packages/integration_test/example/test_driver/example_integration_io_extended.dart
```diff
@@ -0,0 +1,37 @@
...
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  testWidgets('verify text', (WidgetTester tester) async {
+    // Build our app and trigger a frame.
+    app.main();
+
+    // Trigger a frame.
+    await tester.pumpAndSettle();
+
+    // TODO: Add screenshot capability for mobile platforms.
```
dnfield, line 25: nit: link this to the screenshot bug.

packages/integration_test/lib/common.dart
```diff
@@ -121,3 +132,105 @@ class Failure {
     return Failure(failure['methodName'], failure['details']);
   }
 }
+
+/// Integration web tests can execute WebDriver actions such as screenshots.
+///
+/// These test will use [TestStatus] to notify `integration_test` of their
+/// state.
+enum TestStatus {
```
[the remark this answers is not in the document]
dnfield: Maybe we could just change this to `pending` or `complete` to make it more generic?

packages/integration_test/lib/integration_test_driver_extended.dart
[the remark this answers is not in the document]
dnfield: Can we drop the timeout here? We shouldn't be recommending that people use timeouts.

nturgut:
> Please remove the .png files, and consider updating the .gitignore for that.

done. thanks for noticing :)

packages/integration_test/example/test_driver/example_integration_io_extended.dart
nturgut: is there an issue for it for mobile? or should I use the existing issue?

packages/integration_test/lib/_driver_commands_web.dart
```diff
@@ -0,0 +1,168 @@
...
+  /// For example, a test method will ask for a screenshot by calling
+  /// `takeScreenshot`. When this screenshot is taken [driverActionComplete]
+  /// will complete.
+  Completer<bool> driverActionComplete = Completer<bool>();
+
+  /// Takes screenshot using WebDriver screenshot command.
+  ///
+  /// Only works on Web when tests are run via `flutter driver` command.
+  ///
+  /// See: https://www.w3.org/TR/webdriver/#screen-capture
```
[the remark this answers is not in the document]
nturgut: Offline discussion, @yjbanov suggested using this callback instead: https://gist.github.com/yjbanov/684f757dc738a862813d37efee8ba832

packages/integration_test/example/test_driver/example_integration_io_extended.dart
dnfield: Just use the existing one.

packages/integration_test/lib/integration_test_driver_extended.dart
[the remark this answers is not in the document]
nturgut: I removed all the timeouts.

I removed TestStatus enum, I added an object instead of the string. I hope it's more readable now.

packages/integration_test/lib/common.dart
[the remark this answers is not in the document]
nturgut: I added an object. It has pending and complete factory methods, I also added another case for error.

packages/integration_test/lib/_driver_commands_web.dart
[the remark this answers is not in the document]
nturgut: I added the recommended usage from the gist.

ditman:
I'm investigating the current breakages of the CI, and they don't seem related to any changes in the flutter/plugins repo (but a change in [the engine](https://github.com/flutter/engine/commit/15f5696c4139a21e1fc54014ce17d01f6ad1737c#diff-599e1d64442183ead768757cca6805c3L154)), so if you need to merge this on red, go ahead!

nturgut:
> I'm investigating the current breakages of the CI, and they don't seem related to any changes in the flutter/plugins repo (but a change in [the engine](https://github.com/flutter/engine/commit/15f5696c4139a21e1fc54014ce17d01f6ad1737c#diff-599e1d64442183ead768757cca6805c3L154)), so if you need to merge this on red, go ahead!

thanks a lot for investigating :)

PR #2907, opened 2020-08-04
Added WAKE_LOCK permission instructions

akshay-rirev:
The package needs WAKE_LOCK permission sometimes when GoogleMap widget resizes for any reason.
Update google_maps_flutter with instructions on setting WAKE_LOCK permissions.

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/README.md
+++ b/packages/google_maps_flutter/google_maps_flutter/README.md
@@ -39,6 +39,15 @@ To use this plugin, add `google_maps_flutter` as a [dependency in your pubspec.y
 
 ### Android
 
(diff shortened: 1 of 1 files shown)


ditman:
This looks like an interesting contribution, but it's unclear to me why this is needed.

This is also missing a bunch of information, and steps, requested in the PR creation process checklist, for example bumping the package version and adding an entry to the changelog.

This is a good example of a very similar PR: https://github.com/flutter/plugins/pull/1251 (has a linked issue, bumps versions, etc...)

I'm closing this PR because I don't have the bandwidth to bring it up to speed now, however this can be reopened/restarted if made similar to the example above.

Thanks!

PR #2921, opened 2020-08-11
Add README instructions for enablePendingPurchases

matthew-carroll:
Add README instructions for enablePendingPurchases.

If a developer attempts to access this plugin on Android without invoking `enablePendingPurchases`, an exception is thrown.

Related StackOverflow post:
https://stackoverflow.com/questions/60046903/flutter-in-app-purchase-enablependingpurchases-enablependingpurchases-must


matthew-carroll:
@cyanglaz is there a reason that we would change the version number or the change log for a documentation correction?

cyanglaz:
I believe the README is only updated on pub when a new version is published

matthew-carroll:
Alright @cyanglaz I went ahead and bumped the version and updated the change log.

matthew-carroll:
Actually, hold up, I need to resolve a conflict.

matthew-carroll:
Now it's updated.

PR #2928, opened 2020-08-12
Fix unguarded availability in Camera plugin

dnfield:
~~The build is failing on this plugin in an unrelated PR currrently due to these issues.~~ This doesn't fail the build because we're specifically ignoring these warnings, so instead I'm enabling the warnings. This does cause crashes for users on older iOS versions, see linked bug.

This does make me wonder if we really should claim to support this plugin on iOS 8.0.  Currently, you can't take a picture with the camera plugin on anything lower than iOS 10 - it would cause a runtime crash (now it will just result in an exception that should be recoverable).

Fixes https://github.com/flutter/flutter/issues/20708

### DIFF
--- a/packages/camera/ios/Classes/CameraPlugin.m
+++ b/packages/camera/ios/Classes/CameraPlugin.m
@@ -837,7 +837,7 @@ FourCharCode const videoFormat = kCVPixelFormatType_32BGRA;
       }
       int64_t textureId = [_registry registerTexture:cam];
       _camera = cam;
-      __weak CameraPlugin* weakSelf = self;
+      __weak CameraPlugin *weakSelf = self;
       cam.onFrameAvailable = ^{
         [weakSelf.registry textureFrameAvailable:textureId];
       };
--- a/packages/camera/ios/Classes/CameraPlugin.m
+++ b/packages/camera/ios/Classes/CameraPlugin.m
@@ -837,8 +837,9 @@ FourCharCode const videoFormat = kCVPixelFormatType_32BGRA;
       }
       int64_t textureId = [_registry registerTexture:cam];
       _camera = cam;
+      __weak CameraPlugin* weakSelf = self;
       cam.onFrameAvailable = ^{
-        [self.registry textureFrameAvailable:textureId];
+        [weakSelf.registry textureFrameAvailable:textureId];
       };
       FlutterEventChannel *eventChannel = [FlutterEventChannel
           eventChannelWithName:[NSString
--- a/packages/camera/ios/Classes/CameraPlugin.m
+++ b/packages/camera/ios/Classes/CameraPlugin.m
@@ -204,7 +204,7 @@ static ResolutionPreset getResolutionPresetForString(NSString *preset) {
 - (void)stopVideoRecordingWithResult:(FlutterResult)result;
 - (void)startImageStreamWithMessenger:(NSObject<FlutterBinaryMessenger> *)messenger;
 - (void)stopImageStream;
-- (void)captureToFile:(NSString *)filename result:(FlutterResult)result;
+- (void)captureToFile:(NSString *)filename result:(FlutterResult)result API_AVAILABLE(ios(10));
 @end
 
 @implementation FLTCam {
@@ -288,17 +288,19 @@ FourCharCode const videoFormat = kCVPixelFormatType_32BGRA;
 - (void)setCaptureSessionPreset:(ResolutionPreset)resolutionPreset {
   switch (resolutionPreset) {
     case max:
+    case ultraHigh:
+      if (@available(iOS 9.0, *)) {
+        if ([_captureSession canSetSessionPreset:AVCaptureSessionPreset3840x2160]) {
+          _captureSession.sessionPreset = AVCaptureSessionPreset3840x2160;
+          _previewSize = CGSizeMake(3840, 2160);
+          break;
+        }
+      }
       if ([_captureSession canSetSessionPreset:AVCaptureSessionPresetHigh]) {
         _captureSession.sessionPreset = AVCaptureSessionPresetHigh;
         _previewSize =
             CGSizeMake(_captureDevice.activeFormat.highResolutionStillImageDimensions.width,
                        _captureDevice.activeFormat.highResolutionStillImageDimensions.height);
-        break;
-      }
-    case ultraHigh:
-      if ([_captureSession canSetSessionPreset:AVCaptureSessionPreset3840x2160]) {
-        _captureSession.sessionPreset = AVCaptureSessionPreset3840x2160;
-        _previewSize = CGSizeMake(3840, 2160);
         break;
       }
     case veryHigh:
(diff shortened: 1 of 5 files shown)


jmagman:
> This does make me wonder if we really should claim to support this plugin on iOS 8.0. Currently, you can't take a picture with the camera plugin on anything lower than iOS 10 - it would cause a runtime crash (now it will just result in an exception that should be recoverable).

If you changed it to 10.0 then Flutter apps at 8.0 or 9.0 minimums couldn't use the plugin.  There would be a CocoaPods error or a build time error.  The dependencies need to have a lower than equal requirement.

@dnfield can we get a link to the failure?

dnfield:
Here's an example: https://cirrus-ci.com/task/5249881340116992

I only noticed the warnings because of an unrelated error that @ditman fixed and I hadn't patched in yet.

dnfield:
As it is right now, saving a picture to a file on iOS 8.0 or 9.0 would result in a crash. Video functionality would still work. Dropping support for iOS 8.0 and 9.0 might be too severe, but we may want to implement these methods in some way friendly to the version(s) we support if we support them.

jmagman:
https://github.com/flutter/flutter/issues/20708

jmagman:
If you get rid of `--no-analyze camera` does it pass? https://github.com/flutter/plugins/blob/e26634c4ac1def13776769912d3873ec17316fbc/.cirrus.yml#L160

jmagman:
No, it doesn't pass.
CameraPlugin.m:66:24: 'AVCapturePhotoOutput' is only available on iOS 10.0 or newer
CameraPlugin.m:69:43: 'AVCaptureResolvedPhotoSettings' is only available on iOS 10.0 or newer
CameraPlugin.m:78:7: 'JPEGPhotoDataRepresentationForJPEGSampleBuffer:previewPhotoSampleBuffer:' is only available on iOS 10.0 or newer
CameraPlugin.m:77:19: 'AVCapturePhotoOutput' is only available on iOS 10.0 or newer
CameraPlugin.m:163:52: This block declaration is not a prototype
CameraPlugin.m:170:32: 'AVCapturePhotoOutput' is only available on iOS 10.0 or newer
CameraPlugin.m:257:26: 'AVCapturePhotoOutput' is only available on iOS 10.0 or newer
CameraPlugin.m:854:10: Block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior

packages/camera/ios/Classes/CameraPlugin.m
```diff
@@ -204,7 +204,7 @@ - (void)startVideoRecordingAtPath:(NSString *)path result:(FlutterResult)result;
 - (void)stopVideoRecordingWithResult:(FlutterResult)result;
 - (void)startImageStreamWithMessenger:(NSObject<FlutterBinaryMessenger> *)messenger;
 - (void)stopImageStream;
-- (void)captureToFile:(NSString *)filename result:(FlutterResult)result;
+- (void)captureToFile:(NSString *)filename result:(FlutterResult)result API_AVAILABLE(ios(10));
```
jmagman, line 207: All the method declarations can be removed from this interface, it just needs the properties.  `API_AVAILABLE(ios(10.0))` can go in the implementation:

```obj-c
- (void)captureToFile:(NSString *)path result:(FlutterResult)result API_AVAILABLE(ios(10.0)) {
```

dnfield:
I'll see if I can fix those.

packages/camera/ios/Classes/CameraPlugin.m
```diff
@@ -800,34 +802,36 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
-        case AVCaptureDevicePositionBack:
-          lensFacing = @"back";
-          break;
-        case AVCaptureDevicePositionFront:
-          lensFacing = @"front";
-          break;
-        case AVCaptureDevicePositionUnspecified:
-          lensFacing = @"external";
-          break;
+  if (@available(iOS 10.0, *)) {
```
jmagman, line 805: If call.method is `availableCameras` and `@available(iOS 10.0, *)` isn't true, should it `result(@[]);`?  Or do we want the `FlutterMethodNotImplemented` (because I guess it isn't, though I don't know the side effects of that)?

dnfield: I think FlutterMethodNotImplemented is most appropriate, since we haven't implemented the method for that platform.

Today, this is a hard crash for the end user. At least this way they can handle it in application code.

packages/camera/ios/Classes/CameraPlugin.m
```diff
@@ -847,7 +836,7 @@ - (void)handleMethodCallAsync:(FlutterMethodCall *)call result:(FlutterResult)re
       int64_t textureId = [_registry registerTexture:cam];
       _camera = cam;
       cam.onFrameAvailable = ^{
-        [_registry textureFrameAvailable:textureId];
+        [self.registry textureFrameAvailable:textureId];
```
jmagman, line 839: Is this a retain cycle?

jmagman:
One more thing to get rid of the warnings:
https://github.com/flutter/plugins/blob/15b8029d02e50b9572c293c197266d23cbdd3c35/packages/camera/ios/camera.podspec#L19
Should become:
```ruby
s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'VALID_ARCHS[sdk=iphonesimulator*]' => 'x86_64' }
```
Like all the other plugins.  Example:
https://github.com/flutter/plugins/blob/42b33ec8ed3265607df5accf45bc68aa9e4e5592/packages/device_info/device_info/ios/device_info.podspec#L21

That will fix https://github.com/flutter/flutter/issues/61927.

packages/camera/ios/Classes/CameraPlugin.m
```diff
@@ -816,6 +816,8 @@ - (void)handleMethodCallAsync:(FlutterMethodCall *)call result:(FlutterResult)re
         }];
       }
       result(reply);
+    } else {
+      result(FlutterMethodNotImplemented);
```
jmagman, line 820: I think what you had before would have fallen through to `FlutterMethodNotImplemented`, whatever you prefer though.

dnfield: Yeah but it'd make the other `else` blocks not run on iOS 10+ :\

jmagman:
Static analyzer error:
```
CameraPlugin.m:481:5: Incorrect decrement of the reference count of an object that is not owned at this point by the caller
```
Here:
https://github.com/flutter/plugins/blob/05c25052c5dca98c411da925196d7eff766e9dd7/packages/camera/ios/Classes/CameraPlugin.m#L494

Screen Shot 2020-08-12 at 5 42 51 PM

It's being released twice if it hits this codepath:
https://github.com/flutter/plugins/blob/05c25052c5dca98c411da925196d7eff766e9dd7/packages/camera/ios/Classes/CameraPlugin.m#L487

jmagman:
Looks like `-[adjustTime::]` is making a copy of the `CMSampleBufferRef`, so the code is actually right.  Just need to tell the analyzer that we know what we're doing by either adding ["copy" to the method name](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFDesignConcepts/Articles/NamingConventions.html) or a `CF_RETURNS_RETAINED`.
```objc
- (CMSampleBufferRef)adjustTime:(CMSampleBufferRef)sample by:(CMTime)offset CF_RETURNS_RETAINED {
```
or
```objc
- (CMSampleBufferRef)copyBufferAndAdjustTime:(CMSampleBufferRef)sample by:(CMTime)offset {
```

PR #2934, opened 2020-08-16
[versions] increase package:platform constraint

jonahwilliams:
## Description

Fixes https://github.com/flutter/flutter/issues/63441

package:platform required a breaking change in the behavior of FakePlatform in order to enable null safety, but there will be no other breaking changes in 3.X Since this does not affect path provider or local_auth, the pubspec constraint can be increased.

### DIFF
--- a/packages/local_auth/CHANGELOG.md
+++ b/packages/local_auth/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.6.3
+
+* Increase upper range of `package:platform` constraint to allow 3.X versions.
+
 ## 0.6.2+4
 
 * Update package:e2e reference to use the local version in the flutter/plugins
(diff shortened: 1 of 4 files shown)


jonahwilliams:
I'm not sure what the publishing procedure is, but I'm happy to do that if someone gives me permissions :)

cyanglaz:
The publish process is described here: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#the-release-process
You should already have permissions. Let me know if you don't have them. :)

jonahwilliams:
@cyanglaz :

```
Insufficient permissions to upload new versions of package `local_auth`.

pub finished with exit code 1

Publish failed. Exiting.
```

PR #2936, opened 2020-08-17
MORE SHARDS MEANS MORE CI

dnfield:
Shard some tests that have been timing out some more.

Specifically, increase the shard count of build-apks+java-test+firebase-test-lab from 2 to 4.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -63,8 +63,10 @@ task:
     - name: build-apks+java-test+firebase-test-lab
       env:
         matrix:
-          PLUGIN_SHARDING: "--shardIndex 0 --shardCount 2"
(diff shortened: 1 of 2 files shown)


dnfield:
If this passes, it will have to be landed on red, since the failure on master is a timeout that this is meant to help fix/work around.

ditman:
Currently this is being throttled by Cirrus (limit of concurrent community tasks reached)

dnfield:
The real problem was that the shards were not actually sharding the tests. I fixed the bash script, still adding a couple more shards to match what iOS is doing. We can revert that part of we think it's slowing things down for scheduling.

Going to land this on red to get the tree back to green once it passes.  TBR @ditman @cyanglaz

PR #2922, opened 2020-08-11
[video_player] Set audio mix options

wwwdata:
This is a replacement PR for #1174 because I lost access to the original repository I created the PR from.

How do I make the tests work? Unfortunately, I also had to modify the interface package.

@cyanglaz Pinging you like promised and maybe you can also help me out making the tests green and stuff :) But I have to say this new pigeon method of communicating with the native side is pretty awesome 👍 
The only "dirty" thing was that when I created the new dart files, I lost some API test files that apparently were committed to the auto-generated file. I guess that is something that should be moved into a file that is not generated 😄

PR #2937, opened 2020-08-17
Revert "[video_player] Set audio mix options"

dnfield:
Reverts flutter/plugins#2922

CI started timing out after this. Want to see if it the tests complete without it.


dnfield:
Ugh. Script checks prevent this from working properly. Nevermind.

PR #2938, opened 2020-08-18
Revert "[video_player] Set audio mix options (#2922)"

cyanglaz:
This reverts commit 77fbfcd801f7717b681c9b9981bd31c2ce10a2c4.

The original PR made CI timed out on post submit https://cirrus-ci.com/task/6297593099583488?command=main#L88

PR #2939, opened 2020-08-18
Reland "[video_player] Set audio mix options (#2922)"

cyanglaz:
Reverts flutter/plugins#2938

### DIFF
--- a/packages/video_player/video_player/CHANGELOG.md
+++ b/packages/video_player/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.12
+
+* Introduce VideoPlayerOptions to set the audio mix mode.
+
 ## 0.10.11+2
 
 * Fix aspectRatio calculation when size.width or size.height are zero.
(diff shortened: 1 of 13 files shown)


dnfield:
LGTM

steve-rait:
Hi, this new feature has a side effect : 
In my app, when I upgrade plugin, It can't play multiple videos in the same screen by default (I have a list with many items, one video player by item) (**only on Android**)
Indeed, this affects multiple players in the app itself.
I had to force **VideoPlayerOptions** **mixWithOthers** to true to let them play together again.

So it's a breaking change which need to be detailed somewhere in the doc, don't you think ?

cyanglaz:
@steve-rait This breakage is not intended. Do you mind create a github issue (https://github.com/flutter/flutter/issues/new/choose) describe the failure scenario in details? And preferably, if you could provide a sample code to reproduce the issue, that'd help us to accelerate the debugging process. :) 
Feel free to tag me if you create the github issue.

Let me know if you want me to create the issue, but I'd probably need your help to describe it in more details.

steve-rait:
Hi 
Thanks for the info, let me create it, I'll do that this weekend or next week with a complete sample code to show precisely the problem.

PR #2871, opened 2020-07-13
[camera] Fix 'getSurface() on a null object reference' Android crash

Gioxxy:
## Description

Fix 'android.view.Surface android.media.ImageReader.getSurface()' on a null object reference"
which causes the app to crash on Android devices immediately after requesting permission to use the camera if the CameraPreview is not visible (example if it is on the second page of a PageView)

## Related Issues

flutter/flutter#19595

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ ] All existing and new tests are passing.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [X] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [X] I read and followed the [Flutter Style Guide].
- [X] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [ ] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [ ] I updated CHANGELOG.md to add a description of the change.
- [X] I signed the [CLA].
- [X] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [X] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #2965, opened 2020-08-26
[camera] Adding check for null before creating capture session.

panmari:
## Description

Occasionally leads to NullPointerException (potentially due to some race
condition, hard to reproduce deterministically). This patch makes the error go away. I have an app in prod that has 0 occurences of this behavior after the patch. Example stack trace:

```
E/AndroidRuntime( 5679): java.lang.NullPointerException: Attempt to
invoke virtual method 'android.view.Surface
android.media.ImageReader.getSurface()' on a null object reference
E/AndroidRuntime( 5679): 	at
io.flutter.plugins.camera.Camera.startPreview(Camera.java:424)
E/AndroidRuntime( 5679): 	at
io.flutter.plugins.camera.Camera$2.onOpened(Camera.java:160)
E/AndroidRuntime( 5679): 	at
android.hardware.camera2.impl.CameraDeviceImpl$1.run(CameraDeviceImpl.java:145)
E/AndroidRuntime( 5679): 	at
android.os.Handler.handleCallback(Handler.java:883)
E/AndroidRuntime( 5679): 	at
android.os.Handler.dispatchMessage(Handler.java:100)
E/AndroidRuntime( 5679): 	at
android.os.Looper.loop(Looper.java:214)
E/AndroidRuntime( 5679): 	at
android.app.ActivityThread.main(ActivityThread.java:7356)
E/AndroidRuntime( 5679): 	at
java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime( 5679): 	at
com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
E/AndroidRuntime( 5679): 	at
com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

## Related Issues

* flutter/flutter#19595
* flutter/plugins#2871
* https://github.com/flutter/flutter/issues/39109

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
+++ b/packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
@@ -421,7 +421,9 @@ public class Camera {
   }
 
   public void startPreview() throws CameraAccessException {
-    createCaptureSession(CameraDevice.TEMPLATE_PREVIEW, pictureImageReader.getSurface());
+    if (pictureImageReader != null) {
+      createCaptureSession(CameraDevice.TEMPLATE_PREVIEW, pictureImageReader.getSurface());
+    }
   }
 
   public void startPreviewWithImageStream(EventChannel imageStreamChannel)
--- a/packages/camera/CHANGELOG.md
+++ b/packages/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.9
+
+* Fix rare nullptr exception on Android.
+
 ## 0.5.8+6
 
 * Avoiding uses or overrides a deprecated API in CameraPlugin.java.
--- a/packages/camera/pubspec.yaml
+++ b/packages/camera/pubspec.yaml
@@ -2,7 +2,7 @@ name: camera
 description: A Flutter plugin for getting information about and controlling the
   camera on Android and iOS. Supports previewing the camera feed, capturing images, capturing video,
   and streaming image buffers to dart.
-version: 0.5.8+6
+version: 0.5.9
 
 homepage: https://github.com/flutter/plugins/tree/master/packages/camera
 


RenanDelfanti:
This pull request resolve my problem! Thanks!

panmari:
Good to hear Renan!

@bparrishMines there is probably a nicer solution somewhere that solves the underlying issue, but how about we merge this CL for now as it fixes the symptoms?

panmari:
Friendly bump for @bparrishMines :)

panmari:
Rebased again.

bparrishMines:
cc @mvanbeusekom

packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -421,7 +421,9 @@ public void resumeVideoRecording(@NonNull final Result result) {
   }
 
   public void startPreview() throws CameraAccessException {
-    createCaptureSession(CameraDevice.TEMPLATE_PREVIEW, pictureImageReader.getSurface());
+    if (pictureImageReader != null) {
+      createCaptureSession(CameraDevice.TEMPLATE_PREVIEW, pictureImageReader.getSurface());
+    }
```
mvanbeusekom, line 426: nit: you could invert the if statement to reduce nesting and would make it a bit easier to expand the method:

```suggestion
    if (pictureImageReader == null) {
      return;
    }
    
    createCaptureSession(CameraDevice.TEMPLATE_PREVIEW, pictureImageReader.getSurface());
```

mvanbeusekom (commented):
LGTM % nit about inverting the if statement to make code more readible

packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
mvanbeusekom: Looking further into this issue, it would be good to know why `pictureImageReader` is null at this point and if that is valid (as @tvolkert also mentioned in his comments on PR #2871 ).

If the `pictureImageReader` should not be null this code will hide a mistake that is somewhere else in the code base. I will start looking into this.

panmari: Thanks Maurits for looking into this! I agree that this is only a workaround for a symptom, but not a proper fix. Do you have a hunch where the real problem could lie?

mvanbeusekom: Hi @panmari, we (me and a colleague) have been looking into this issue and can't quite reproduce it as of yet. We do have 2 theories at the moment:

1. Since version 0.5.0 developers are responsible for handling application lifecycle events that might affect the plugin (see also [this comment](https://github.com/flutter/flutter/issues/39109#issuecomment-524250114) mentioned by @kevin-lot). If these events are not handled correctly it might cause the issue where the permission system moves the app to the background and not reinitialize the camera plugin correctly (leaving it in an invalid state);
2. We could did get the same error when (using the example app) very rapidly switching between the front and back facing cameras. We noticed that in these cases the `onOpened` callback would be called when the `dispose` method has already been run (race condition).

I am currently discussing on how we should approach these problems.

panmari:
Abandoning in favor of #3127

PR #2967, opened 2020-08-26
[integration_test] Show stack trace of widget test errors on the platform side

jiahaog:
## Description

We keep things simple for now, because this might change once we have a proper test reporter.

## Breaking Change


packages/integration_test/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unpublished
```
dnfield, line 1: Let's make this 0.8.1 and update the pubspec.

PR #2972, opened 2020-08-27
[integration_test] Fix channel name for iOS, docs

jiahaog:
## Description

Fix channel name for iOS, docs

I haven't published `0.8.1` yet, will do so after this PR to get the fix for the channel name in.

### DIFF
--- a/packages/integration_test/README.md
+++ b/packages/integration_test/README.md
@@ -6,7 +6,7 @@ and native Android instrumentation testing.
 
 ## Usage
 
-Add a dependency on the `integration_test` package in the
+Add a dependency on the `integration_test` and `flutter_test` package in the
 `dev_dependencies` section of pubspec.yaml. For plugins, do this in the
 pubspec.yaml of the example app.
 
@@ -14,6 +14,7 @@ Invoke `IntegrationTestWidgetsFlutterBinding.ensureInitialized()` at the start
 of a test file, e.g.
 
 ```dart
+import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 
 void main() {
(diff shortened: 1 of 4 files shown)


packages/integration_test/README.md
```diff
@@ -47,28 +48,27 @@ import 'dart:async';
...
 
 Future<void> main() async => integrationDriver();
-
 ```
 
 To run a example app test with Flutter driver:
 
-```
+```sh
 cd example
```
jiahaog, line 56: I'm not really sure what this means... Is it referring to the example app in `packages/integration_test/example`? If so we should update the file names accordingly, and perhaps we should move the following docs there and direct users there from here.

dnfield: Yes, it's that directory.

packages/integration_test/ios/Classes/IntegrationTestPlugin.m
```diff
@@ -1,6 +1,6 @@
 #import "IntegrationTestPlugin.h"
 
-static NSString *const kIntegrationTestPluginChannel = @"plugins.flutter.io/integratoin_test";
```
dnfield, line 3: I could have sworn @bparrishMines fixed this previously.

packages/integration_test/README.md
jiahaog: Filed #64690, will follow up separately

packages/integration_test/ios/Classes/IntegrationTestPlugin.m
bparrishMines: Sorry, I was supposed to. I got distracted while waiting for our CI to go green.

PR #2986, opened 2020-09-01
[integration_test] Recommend tests to be in `integration_test/`, fix example

jiahaog:
## Description

In this way, there is a clear distinction between integration tests that run on a device (in `integration_test/`, and widget tests that run with the flutter tester in `test/`.

## Related Issues

flutter/flutter#64690


packages/integration_test/README.md
```diff
@@ -7,78 +7,91 @@ and native Android instrumentation testing.
...
-You can run tests on web in release or profile mode.
+### Web
+
+Make sure you have [enabled web support](https://flutter.dev/docs/get-started/web#set-up)
+then [download and run](https://flutter.dev/docs/cookbook/testing/integration/introduction#6b-web)
+the web driver in another process.
 
-First you need to make sure you have downloaded the driver for the browser.
+Use following command to execute the tests, passing either `--release` or
+`--profile` as debug mode is not supported.
```
ditman, line 79: Debug mode is now supported, and in fact, it's the only way of getting stack traces in web now.

packages/integration_test/README.md
```diff
@@ -7,78 +7,91 @@ and native Android instrumentation testing.
 ## Usage
 
 Add a dependency on the `integration_test` and `flutter_test` package in the
-`dev_dependencies` section of pubspec.yaml. For plugins, do this in the
-pubspec.yaml of the example app.
+`dev_dependencies` section of `pubspec.yaml`. For plugins, do this in the
+`pubspec.yaml` of the example app.
```
ditman, line 11: I might have a rare use case, but for plugins where "web" is the only platform, I normally (almost always) need dart:html to run tests, so all my tests are "integration", and live in the "test" directory. I guess this advice makes sense for more normal packages, and for consistency across all plugins!

packages/integration_test/README.md
ditman: (At least in `master` :P)

packages/integration_test/README.md
jiahaog: I see, would it then make sense to remove this sentence?

> For plugins, do this in the `pubspec.yaml` of the example app.

I think it's kind of implicit anyway – what we're trying to tell users is put tests in `integration_tests/` of their packages as needed – if they have sub-packages like `example/`, they can have a `example/integration_tests/` directory as well

packages/integration_test/README.md
jiahaog: Hmm, on master i get this error when I run the tests, am I doing something wrongly?

```
$ flutter drive   --driver=test_driver/integration_test_driver.dart   --target=integration_test/example_test.dart -d web-server
Starting application: integration_test/example_test.dart
Flutter Driver web does not support running in debug mode.

Use --profile mode for testing application performance.
Use --release mode for testing correctness (with assertions).
```

ditman: Hmmm, odd, I think your `master` may be a little outdated; here's the change that removes that limitation:

https://github.com/flutter/flutter/pull/62577

packages/integration_test/README.md
ditman: Nah, I think web plugins are "odd" because they normally need the browser to run (some times even unit tests), so we use the integration_test package to run normal tests. I think this is documenting the current "correct" behavior. If this changes in the future, we can update.

packages/integration_test/README.md
jiahaog: My apologies, there was a mistake in my setup. Updated, thanks!

packages/integration_test/README.md
jiahaog: Got it 🙂

packages/integration_test/README.md
```diff
@@ -7,78 +7,89 @@ and native Android instrumentation testing.
...
 
-Invoke `IntegrationTestWidgetsFlutterBinding.ensureInitialized()` at the start
-of a test file, e.g.
+Create a `integration_test/` directory for your package. In this directory,
```
nturgut, line 13: In the cirrus file we are still using test_driver/ directory even though it is not the recommended directory?

nturgut: Oh sorry, I just read the recommended package structure.

Is there a reason for these recommendations? One idea that comes to my mind is: integration_test_driver.dart won't be the same for different tests.

nturgut:
I think this approach blocks the possibilities that can be tested with e2e tests. Let me explain more in details:

For example in the [extended test example](https://github.com/flutter/plugins/blob/master/packages/integration_test/example/test_driver/example_integration_extended_test.dart#L9) we are creating a driver and give it to the `integrationDriver` method. We also give a handler for handling the screenshots.

This way, depending on the repository, developers can handle the screenshots differently. Even better, FlutterDriver can be created differently, for example a WebDriver can enable accessibility in advance. Or FlutterDriver can be provided with different [options](https://github.com/flutter/flutter/blob/d2fa384c319d715fa2f74a077444fd9cc7939732/packages/flutter_driver/lib/src/driver/driver.dart#L109) for the driver.

Also, the same application, can use different driver files since they can provide different functionalities.

Due to this reasons, we will probably won't switch to this usage in web integration tests on the engine side. I also recommend keeping the other option in the documentation for power users who wants to provide a wider set of tests.

packages/integration_test/README.md
jiahaog: It won't be the same but I would think for most use cases the bulk of the testing code should go into the test script that runs on the device, and the driver is mainly for reporting.

Currently, some packages place the test scripts in `test/foo_e2e.dart` or `test_driver/foo_e2e.dart`. This works today, because of the old e2e naming, but if they were to rename their tests to like `test/foo_integration_test.dart` and  `test_driver/foo_integration_test.dart`, this would conflict with globbing rules today that check for the `*_test.dart` suffix, hence the recommendation of a new `integration_test/` directory for such tests

nturgut: We did the renaming like this last week in the for the web engine: https://github.com/flutter/engine/pull/20954

jiahaog:
@nturgut Sorry, I might have missed something. Can you elaborate more about how this approach blocks certain testing use cases? Customizing the driver is still possible with a custom driver file – in this PR, I have moved the extended test example to follow this proposed structure. There is a change to the command used to kick off the tests, but there is no change in functionality.

nturgut:
> @nturgut Sorry, I might have missed something. Can you elaborate more about how this approach blocks certain testing use cases? Customizing the driver is still possible with a custom driver file – in this PR, I have moved the extended test example to follow this proposed structure. There is a change to the command used to kick off the tests, but there is no change in functionality.

Sorry, I might be missing something, the README suggestion is:

```
An accompanying driver script will be needed that can be shared across all
integration tests. Create a `integration_test_driver.dart` in the `test_driver/`
directory with the following contents:
```

This suggestion seems very limiting to me. For example we can add:
- a note saying multiple driver files can be used
- we can also add an example of how we can pass a FlutterDriver

In my opinion, if developer adds multiple driver files, then the advantage of the new structure is not very clear:
```
lib/
  ...
integration_test/
  foo_test.dart
  bar_test.dart
test/
  # Other unit tests go here.
test_driver/
  integration_test_driver.dart
  integration_test_driver_with_screenshot.dart
  integration_test_driver_with_cstm_setup.dart
```

jiahaog:
Thanks for clarifying – you're right that I wasn't clear enough in how we can use multiple driver files, I have updated the PR and linked it to the extended driver script.

My understanding is that one of the goals for `package:integration_test` is to push as much logic as possible into the test script, so that the test will be compatible with tools that do not provide access to the host, like Firebase Test Lab. I have not been exposed too much to web use cases, so let me know if I'm wrong, but I was going with the assumption that customizing the driver script should be only done rarely (such as to pull artifacts from the device), and there should be a significantly larger number of test scripts that run on the device to driver scripts which this recommendation will benefit. Taking a brief look at the [`e2etests/`](https://github.com/flutter/engine/tree/master/e2etests/web/regular_integration_tests) directory of the engine, I see that most of the driver scripts are delegating to `integrationDriver`, so sharing the same driver script might be helpful there.

nturgut (approved):
Thanks for updating the documentation.

jiahaog:
Thanks!

PR #2988, opened 2020-09-01
[shared_preferences] Add shared_preferences_windows

stuartmorgan:
## Description

Dart implementation of the shared_preferences plugin using ffi for Windows.

This is https://github.com/flutter/plugins/pull/2631, trimmed down to the initial stand-alone PR that adds shared_preferences_windows, since it must be landed and published before the shared_preferences registration and endorsement can land.

## Related Issues

https://github.com/flutter/flutter/issues/41719

## Checklist

- [x] I signed the [CLA].

## Breaking Change


stuartmorgan:
This is now working, and passing the experimental CI bots. However, looking again at the way it's doing path storage, and comparing with the Linux implementation, it makes a lot more sense to hold off on this until path_provider_windows is landed, and then use it to get the storage location instead of having the Win32 code duplicated here (with different logic for constructing the subpath, as compared to path_provider). Once path_provider_windows is in I'll rework this to depend on it, as we do with shared_preferences_linux.

stuartmorgan:
@franciscojma86 Can you approve the co-authorship here when you get a chance? I've done a fair amount of rewriting based on what we learned from some of the Linux plugins, especially aligning it with the basic structure of shared_preferences_linux due the fact that shared_preferences_windows is now mostly complete, but it's still very much based on your PR.

franciscojma86:
@googlebot I consent.

franciscojma86:
So exciting!

stuartmorgan:
This is largely done, it just needs some adjustment to pubspec and testing after path_provider_windows is finalized.

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -0,0 +1,103 @@
...
+
+  /// Cached file for storing preferences.
+  File _localDataFilePath;
+
+  /// Gets the file where the preferences are stored.
+  Future<File> _getLocalDataFile() async {
+    if (_localDataFilePath == null) {
+      final directory = await pathProvider.getApplicationSupportPath();
+      _localDataFilePath =
+          fs.file(path.join(directory, 'shared_preferences.json'));
```
csells, line 40: The Windows per-user roaming settings path (which is what I believe application support path maps to) will be something like:  "C:\Users\<user>\AppData\Roaming". Unless we differentiate between apps, multiple Flutter apps will step on each other's preferences. The convention to prevent that is to append company name and app name, e.g.  "C:\Users\<user>\AppData\Roaming\<company>\<app-name>\shared_preferences.json", e.g. https://github.com/csells/shared_prefs_proto/blob/master/SharedPreferences.h#L49.

On Windows, the VERSIONINFO resource is the place to get that data, which would require us to generate one in the Runner, e.g. company == Flutter and app-name == project name from 'flutter create' and the developer can update the resource as they choose.

If we don't add a VERSIONINFO or if the app doesn't have one for another reason, e.g. we give them one and they remove it for some reason, then the normal fallback is to hardcode a company name, e.g. Flutter, and use the basename of the currently running exe as the app-name, e.g. c:\foo\bar.exe gives app-name == bar.

packages/shared_preferences/shared_preferences_windows/test/shared_preferences_windows_test.dart
```diff
@@ -0,0 +1,107 @@
...
+  setUp(() {
+    fileSystem = MemoryFileSystem.test();
+    pathProvider = FakePathProviderWindows();
+  });
+
+  tearDown(() {});
+
+  Future<String> _getFilePath() async {
+    final directory = await pathProvider.getApplicationSupportPath();
+    return path.join(directory, 'shared_preferences.json');
```
csells, line 25: ditto to the same comment above

csells (changes requested):
I don't think we're calculating the path where the shared preferences are stored correctly.

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
stuartmorgan: The current path_provider_windows PR is appending the exe name (although not a company name; I missed that that was part of the convention on Windows). path_provider is the place where that should happen, because the expectation from people used to path_provider on mobile is that the folder returned by `getApplicationSupportPath` is already application-specific. I'll add a comment over there.

timsneath: I've implemented these methods in the win32 package, but I wonder if we're overthinking it. 

It's true that `VERSIONINFO` _could_ get the company name and app name, but it seems a little fragile (in particular, if the characters used aren't valid Windows file characters), and I don't see most apps using the convention above, even including those from Microsoft. On my machine:

```
[C:\Users\tim\AppData\Roaming] dir
 Volume in drive C is Local Disk
 Volume Serial Number is 4459-1FFE

 Directory of C:\Users\tim\AppData\Roaming

08/28/2020  08:50 AM    <DIR>          .
08/28/2020  08:50 AM    <DIR>          ..
08/28/2020  04:05 PM    <DIR>          .dart
07/30/2020  10:38 PM                78 .flutter
08/05/2020  04:03 PM                86 .flutter_settings
09/15/2020  10:17 AM               135 .flutter_tool_state
07/30/2020  06:37 PM    <DIR>          Adobe
09/16/2020  06:27 PM    <DIR>          Code
09/16/2020  10:03 AM    <DIR>          ColorPicker
07/31/2020  10:48 AM    <DIR>          Microsoft FxCop
08/28/2020  08:50 AM    <DIR>          Microsoft Visual Studio
07/31/2020  10:47 AM    <DIR>          NuGet
09/04/2020  02:38 PM    <DIR>          Visual Studio Setup
07/30/2020  09:21 PM    <DIR>          vstelemetry
07/30/2020  09:21 PM    <DIR>          vs_installershell
               3 File(s)            299 bytes
              12 Dir(s)  151,567,503,360 bytes free
```

I wonder if it's simpler to just use the app name from `pubspec.yaml` to create the folder?

csells: The Windows convention is ...\Roam\<company>\<app>, which is an attempt at uniqueness and not stepping on other app's settings (especially roaming, since that's the total settings from all Windows boxes for that user). Without a VERSIONINFO there's no way to pass the extra info along to specify the company name, losing one level of uniqueness. Of course, the standard trick with non-Windows-file-system characters is to replace them, e.g. underscore. Further, I don't know if we have any way to get the info out of the pubspec at runtime. Without the VERSIONINFO, the fallback is to use the exe base name, e.g. c:\foo\bar.exe == bar.

Honestly, I'd rather lean on the safe side of being more unique and following the Windows convention then get a reputation for stepping on other app's settings.

stuartmorgan: Given that path_provider_windows is now doing the full path, are you happy with the code here as-is? (I'll update pubspec to point to 0.0.2 to ensure that's picked up once that is published.)

csells: I am.

csells (approved):

PR #2993, opened 2020-09-02
Add LUCI try builders json config file

keyonghan:
## Description

This is to support `plugins` repo LUCI try builders, triggered by [`cocoon`](https://github.com/flutter/cocoon) backend.

To be consistent with other repos like `flutter/flutter`, `flutter/engine`, `flutter/cocoon`, etc. we put the config file under `/dev`. The json format will be validated from `plugins/plugins.py` recipe.

## Related Issues

https://github.com/flutter/flutter/issues/56143#issuecomment-685843581


godofredoc (approved):
LGTM but let's ask @stuartmorgan it is ok to add the files in this location.

stuartmorgan:
> LGTM but let's ask @stuartmorgan it is ok to add the files in this location.

I'm not the best authority for the flutter/plugins repo, but given that there's already a `.ci` folder at the top level is there any reason not to just use that?

(Also, `dev` seems like a potentially confusing name given that it's the name of a Flutter channel.)

stuartmorgan:
Oh, I just saw this part:

> To be consistent with other repos like flutter/flutter, flutter/engine, flutter/cocoon, etc. we put the config file under /dev.

flutter/flutter already had a dev/ folder containing CI stuff, and flutter/engine doesn't have a dev/ folder that I can see, so the consistency argument seems questionable. I think using the existing folder that this repo uses for CI would be better.

keyonghan:
> Oh, I just saw this part:
> 
> > To be consistent with other repos like flutter/flutter, flutter/engine, flutter/cocoon, etc. we put the config file under /dev.
> 
> flutter/flutter already had a dev/ folder containing CI stuff, and flutter/engine doesn't have a dev/ folder that I can see, so the consistency argument seems questionable. I think using the existing folder that this repo uses for CI would be better.

Yeah, you are right. `flutter`, `cocoon`, and `packages` are all with `dev`, whereas `engine` is a special case with `ci/dev`. Sure. We can do the same for `plugins` similar to `engine`. 

As for consistency, I meant to say all are with `dev/builders.json`. Sorry for the confusion.

PR #3009, opened 2020-09-08
[integration_test] Allow users to hook into test completion

jiahaog:
## Description

Allow users to hook into the result reporting lifecycle. This API is unstable pending flutter/flutter#50816, so temporarily marking it as `@deprecated`


jiahaog:
Turns out that `tearDownAll` has LIFO serial semantics, so we can use that as a workaround until the test reporting API is stable

PR #3015, opened 2020-09-10
Add url_launcher_windows

stuartmorgan:
## Description

Adds a federated Windows implementation of url_launcher (not yet endorsed).

Since this is the first C++ plugin, this also adds a .clang-format file to the repo root for formatting C++ plugins.

## Related Issues

Part of https://github.com/flutter/flutter/issues/41721 (will need follow-up to endorse it in url_launcher)

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/url_launcher/url_launcher_windows/README.md
+++ b/packages/url_launcher/url_launcher_windows/README.md
@@ -1,13 +1,31 @@
-# url_launcher_fde
+# url_launcher_windows
 
-Prototype Windows implementation of
-[url_launcher](https://pub.dev/packages/url_launcher)
+The Windows implementation of [`url_launcher`][1].
 
-See [the main flutter_plugins README](../README.md) for general information about what
-this plugin is and how to use it.
+## Backward compatible 1.0.0 version is coming
+The plugin has reached a stable API, we guarantee that version `1.0.0` will be backward compatible with `0.0.y+z`. If you use
+url_launcher_windows directly, rather than as an implementation detail
+of `url_launcher`, please use `url_launcher_windows: '>=0.0.y+x <2.0.0'`
+as your dependency constraint to allow a smoother ecosystem migration.
+For more details see: https://github.com/flutter/flutter/wiki/Package-migration-to-1.0.0
 
-## Supported Platforms
+## Usage
 
-- [x] Windows
+### Import the package
 
-macOS and Linux are already supported by `url_launcher`.
+This package has not yet been endorsed. Once it is you only need to add
+`url_launcher` as a dependency in your `pubspec.yaml`, but for now you
+need to include both `url_launcher` and `url_launcher_windows`.
+
+This is what the above means to your `pubspec.yaml`:
+
+```yaml
+...
+dependencies:
+  ...
+  url_launcher: ^5.5.3
+  url_launcher_windows: ^0.0.1
+  ...
+```
+
+[1]: ../url_launcher/url_launcher
--- a/packages/url_launcher/url_launcher_windows/README.md
+++ b/packages/url_launcher/url_launcher_windows/README.md
@@ -1,0 +1,13 @@
+# url_launcher_fde
+
+Prototype Windows implementation of
+[url_launcher](https://pub.dev/packages/url_launcher)
+
+See [the main flutter_plugins README](../README.md) for general information about what
+this plugin is and how to use it.
+
+## Supported Platforms
+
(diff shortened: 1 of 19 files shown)


packages/url_launcher/url_launcher_windows/README.md
```diff
@@ -0,0 +1,31 @@
+# url_launcher_windows
+
+The Windows implementation of [`url_launcher`][1].
+
+## Backward compatible 1.0.0 version is coming
+The plugin has reached a stable API, we guarantee that version `1.0.0` will be backward compatible with `0.0.y+z`. If you use
+url_launcher_windows directly, rather than as an implementation detail
+of `url_launcher`, please use `url_launcher_windows: '>=0.0.y+x <2.0.0'`
```
csells, line 8: can I ask why we're recommending this instead of putting the right version of the plugin as a dependency of url_launcher?

csells: answering my own question, I assume we're doing this as a temporary message until Windows hits alpha.

packages/url_launcher/url_launcher_windows/windows/url_launcher_plugin.cpp
```diff
@@ -0,0 +1,148 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+#include "include/url_launcher_windows/url_launcher_plugin.h"
+
+#include <Windows.h>
```
csells, line 6: the convention when I was a Windows dev was lower case: 

```c
#include <windows.h>
```

csells (approved):
lgtm

packages/url_launcher/url_launcher_windows/README.md
stuartmorgan: To endorse a federated plugin (i.e., make it a dependency of the main plugin), the plugin already has to be published. So our flow for a new federated implementation is:
1. Make a PR for the federated implementation (e.g., foo_windows).
2. Land foo_windows.
3. Publish foo_windows.
4. Make a PR (against foo) to endorse the new federated implementation.
5. Land foo.
6. Publish foo.

This README covers steps 3 through 5, where this plugin exists in the wild as a published plugin but isn't endorsed. In step 6, the foo_windows readme gets updated to say that it's included automatically by foo. (Then that gets published as a minor, README-only change to foo_windows).

It's cumbersome, but the alternative (putting everything in one PR) wouldn't pass CI because the build of the main plugin would fail due to the federated implementation not having been published.

packages/url_launcher/url_launcher_windows/windows/url_launcher_plugin.cpp
stuartmorgan: Done; I'd seen both and wasn't sure if there was a preferred version.

(In fact, our app template is internally inconsistent, now that I look at it :P I'll fix that up too.)

packages/url_launcher/url_launcher_windows/example/lib/main.dart
```diff
@@ -0,0 +1,93 @@
...
+
+  @override
+  _MyHomePageState createState() => _MyHomePageState();
+}
+
+class _MyHomePageState extends State<MyHomePage> {
+  Future<void> _launched;
+
+  Future<void> _launchInBrowser(String url) async {
+    if (await UrlLauncherPlatform.instance.canLaunch(url)) {
```
amirh, line 40: Not for this PR but still - 

This saddens me, I really want us to come up with a way to share relevant pieces of the example app and e2e tests (and not have an example app thats using the platform interface).

I'm planning to be focusing on these issues in the coming weeks.

amirh (approved):
LGTM

packages/url_launcher/url_launcher_windows/example/lib/main.dart
stuartmorgan: Per my recent comments is Discord, I think using the platform interface rather than the app-facing plugin is a better way to write the driver tests; maybe we can continue the discussion there.

The role of the example app in a federated implementation is questionable though; it's only going to be an example in the (rare?) cases where there is extra API in the platform implementation. Having it replicate things that are in the main example—which is what people should look at in almost every case—isn't really useful as an example. But it is useful to have a self-contained build when working on the implementation.

PR #3023, opened 2020-09-12
Fix formatting

jiahaog:
A new version of the formatter or something might have rolled in, causing CI to break


jiahaog:
Hmm, there's an issue where running `pub global activate flutter_plugin_tools && pub global run flutter_plugin_tools format` locally causes the block in `packages/connectivity/connectivity/ios/Classes/FLTConnectivityPlugin.m` to be formatted like:

```
default: {
  return @"unknown";
}
```

But the CI check expects the block to be formatted like:

```
default: { return @"unknown"; }
```

stuartmorgan (approved):
LGTM

PR #3055, opened 2020-09-19
[path_provider] Fix mistake in license headers

mehmetf:
MacOS source headers were mistakenly marked as Apache license rather than BSD.


cyanglaz (approved):
LGTM!

PR #3063, opened 2020-09-21
[integration_test] Test drivers should end with  `_test.dart`

jiahaog:
After working on tooling to migrate existing apps (draft #3046), I think it makes more sense for the driver files to keep the _test.dart suffix. This makes globbing by tooling easier.

## Breaking Change


jiahaog:
Please ignore the CI failures for now – I have opened flutter/plugin_tools#108 which will update the plugin tools to support this.

dnfield (approved):

dnfield:
I'm not sure why this is failing, looks like things passed?

jiahaog:
Thanks! It is failing because the CI script doesn't support this yet – please see flutter/plugin_tools#108!

jiahaog:
I moved just the failure test to the test_driver directory to avoid globbing for that, and added a note as well.

As mentioned on #hackers-ecosystem, the failure on master is a flake with Cocoapods, so landing this.

PR #3066, opened 2020-09-21
Revert "[webview_flutter] Add new entrypoint that uses hybrid composition on Android"

bparrishMines:
Reverts flutter/plugins#2883


blasten (approved):
LGTM

PR #3056, opened 2020-09-20
[path_provider] Move Windows FFI behind a conditional import

stuartmorgan:
## Description

Moves the real implementation of path_provider_windows behind a
conditional export, instead exporting a stub on platforms that don't
support dart:ffi. This avoids build breakage in web projects that have
transitive dependencies on path_provider (and thus path_provider_windows
due to manual endorsement).

This will no longer be necessary once
https://github.com/flutter/flutter/issues/52267 is fixed, since only
Windows builds will ever need to have code-level dependency on
path_provider_windows.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/66143

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #3077, opened 2020-09-24
[path_provider] Use the application ID in the application support path (#2845)

robert-ancell:
## Description

Change getApplicationSupportPath to use the Linux application ID instead of the executable name.

The was landed, then reverted in https://github.com/flutter/plugins/pull/2845.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
```diff
@@ -3,11 +3,21 @@
 // found in the LICENSE file.
 import 'dart:io';
 import 'dart:async';
+import 'dart:ffi';
```
stuartmorgan, line 6: As I recently learned the hard way, if you add FFI to a plugin that's manually registered (as all desktop Dart plugins currently have to be) you need to do [this](https://github.com/flutter/plugins/pull/3056) to avoid breaking web builds. This won't show up in CI currently, unfortunately.

stuartmorgan:
@robert-ancell Is this something you are still planning to do? It's a breaking change (e.g., for shared_preferences), so the earlier the better, although I think we'll need to make it opt-in at this point since there's been more time for people to start depending on the current behavior.

robert-ancell:
It's something we really *should* do, though I haven't been looking at it. I'll put it back on my TODO list.

robert-ancell:
I've updated this to use the legacy directory name if it exists, and use the new one if not. This should mean existing apps keep working as expected.

packages/path_provider/path_provider_linux/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.1
+
+* Re-add change to getApplicationSupportPath from using executable name to application ID (if provided).
```
stuartmorgan, line 3: I don't think we need to say "re-add" since we never published it.

packages/path_provider/path_provider_linux/lib/src/path_provider_linux_stub.dart
```diff
@@ -0,0 +1,20 @@
...
+/// Notably, because path_provider needs to manually register
+/// path_provider_linux, anything with a transitive dependency on
+/// path_provider will also depend on path_provider_linux, not just at the
+/// pubspec level but the code level.
+class PathProviderLinux extends PathProviderPlatform {
+  /// Errors on attempted instantiation of the stub. It exists only to satisfy
+  /// compile-time dependencies, and should never actually be created.
+  PathProviderLinux() {
+    assert(false);
+  }
```
stuartmorgan, line 19: The stub has to have everything the real version has, or you create analyzer errors, because the analyzer doesn't really understand conditional includes (and in fact it's not even clear how it *should* handle them)

I highly recommend that you make a utility file that has just the `getApplicationId` function, and only put that behind a conditional include so that there aren't two copies of things to maintain.

stuartmorgan (changes requested):

robert-ancell:
The test failure seems unrelated to this PR - a trivial PR (https://github.com/flutter/plugins/pull/3774) also shows the same failure.

packages/path_provider/path_provider_linux/lib/src/get_application_id_real.dart
```diff
@@ -0,0 +1,38 @@
...
+
+/// Gets the application ID for this app.
+String? getApplicationId() {
+  DynamicLibrary gio;
+  try {
+    gio = DynamicLibrary.open('libgio-2.0.so');
+  } on ArgumentError {
+    return null;
+  }
+  var _g_application_get_default = gio.lookupFunction<
```
stuartmorgan, line 24: Local variables don't need leading _s.

packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart
```diff
@@ -0,0 +1,69 @@
...
+  static void register() {
+    PathProviderPlatform.instance = PathProviderLinux();
+  }
+
+  @override
+  Future<String?> getTemporaryPath() {
+    return Future.value("/tmp");
+  }
+
+  // Gets the name of this executable.
```
stuartmorgan, line 27: Let's keep the overrides grouped together, and put the private methods after.

packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart
```diff
@@ -0,0 +1,69 @@
...
+    if (appId != null) return appId;
+
+    // Fall back to using the executable name.
+    return await _getExecutableName();
+  }
+
+  @override
+  Future<String?> getApplicationSupportPath() async {
+    // This plugin originally used the executable name as a directory.
+    // Use that if it exists for backwards compatibility.
```
stuartmorgan, line 45: I think to make this even safer, the logic should be:
- If the new path exists, use the new path
- If not, and the old path exists, use the old path
- Otherwise, use the new path

That avoids the possibility of some future name collision on a machine breaking an app that was already on the correct (new) path.

packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart
```diff
@@ -0,0 +1,69 @@
...
+    final legacyDirectory =
+        Directory(path.join(xdg.dataHome.path, await _getExecutableName()));
+    if (await legacyDirectory.exists()) {
+      return legacyDirectory.path;
+    }
+
+    final directory = Directory(path.join(xdg.dataHome.path, await _getId()));
+    // Creating the directory if it doesn't exist, because mobile implementations assume the directory exists
```
stuartmorgan, line 53: Nit: trailing .

packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart
```diff
@@ -0,0 +1,69 @@
...
+
+  @override
+  Future<String?> getApplicationSupportPath() async {
+    // This plugin originally used the executable name as a directory.
+    // Use that if it exists for backwards compatibility.
+    final legacyDirectory =
+        Directory(path.join(xdg.dataHome.path, await _getExecutableName()));
+    if (await legacyDirectory.exists()) {
+      return legacyDirectory.path;
+    }
```
stuartmorgan, line 50: The logic for this needs tests (currently at https://github.com/flutter/plugins/blob/master/packages/path_provider/path_provider_linux/example/integration_test/path_provider_test.dart for consistency with the other implementations.)

stuartmorgan (changes requested):

stuartmorgan (commented):
Are you still planning on updating this with tests for the new functionality?

stuartmorgan:
@robert-ancell Ping on this? I do still think this is something we should do, it just needs the final testing.

robert-ancell:
Updated now with tests. I've rebased because there have been enough upstream changes to warrant giving it a whole look over again.

packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart
```diff
@@ -0,0 +1,98 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
```
stuartmorgan, line 5: Do you actually need this?

packages/path_provider/path_provider_linux/test/path_provider_linux_test.dart
```diff
@@ -35,8 +36,19 @@ void main() {
   });
 
   test('getApplicationSupportPath', () async {
-    final PathProviderPlatform plugin = PathProviderPlatform.instance;
-    expect(await plugin.getApplicationSupportPath(), startsWith('/'));
+    final PathProviderPlatform plugin = PathProviderLinux.private(
+        executableName: 'test', applicationId: 'com.example.Test');
+    // Note this will fail if ${xdg.dataHome.path}/test exists on the local filesystem.
```
stuartmorgan, line 41: Maybe use something less likely to collide? E.g., `path_provider_linux_unit_test`

packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart
```diff
@@ -0,0 +1,98 @@
...
+  // Gets the unique ID for this application.
+  Future<String> _getId() async {
+    _applicationId ??= getApplicationId();
+
+    if (_applicationId != null) {
+      return _applicationId!;
+    }
+
+    // Fall back to using the executable name.
+    return await _getExecutableName();
```
stuartmorgan, line 96: You could collapse the block above and this to: `return _applicationId ?? _getExecutableName();` and not need to use `!`.

stuartmorgan (approved):
LGTM with nits.

robert-ancell:
@stuartmorgan is the "submit-queue" test failure related to this? Seems to be a video player issue.

stuartmorgan:
`submit-queue` is the current status of the tree, to prevent landing PRs into a broken tree; it's never related to the PR itself.

PR #3031, opened 2020-09-15
[template] [video_player] Add playback speed support on all platforms

creativecreatorormaybenot:
## Description

Playback speed on videos is a very important feature to us and seems to be highly requested as well.  
There have been earlier attempts to implement this feature into the `video_player` plugin, however, the PRs for those have remained stale.

Consequently, I decided to pick up this task and implement playback speed support into the latest version of the `video_player` plugin.

### What this PR accomplishes

This PR adds the playback speed feature on Android, iOS, and web.

I did not copy code from existing PRs (though I did use them as a reference for the Android and iOS implementations). Instead, I designed the API in the way that I figured was most sensible having the learnings from the other PRs in mind. (I actually later figured out just how similar the interface implementations were, even though they were created separately. I suppose it happened because we all followed the existing interface).

## Related Issues

* Resolves https://github.com/flutter/flutter/issues/29900.
* Resolves https://github.com/flutter/flutter/issues/16624.
* Closes https://github.com/flutter/plugins/pull/1400.
* Closes https://github.com/flutter/plugins/pull/860.

## Pigeon

I upgraded `pigeon` to `0.1.7`. You might say that this makes the PR harder to review because the generation changed, but I thought about it and I have to disagree:

* There are changes anyway using the older `pigeon` version.
* Generated files should be ignored in PR review anyway.

### Manual testing

I tested this both on Android and on iOS in a production app.  
Additionally, I added the feature to the example app and tested that on web and mobile.

### Unit (and integration) tests

Integration testing for this feature seems not only very difficult, but most importantly not sensible. I can test the platform calls, but the underlying native libraries and Flutter functionality should do the integration testing.

Having said that, I did include unit tests.

## Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
[...]

PR #3081, opened 2020-09-25
[video_player_web] Playback speed

creativecreatorormaybenot:
## Part

**Part 2/3** of https://github.com/flutter/plugins/pull/3031.  
This is the `video_player_web` part of that PR.

Checklist, full code, et al. can be found in the linked PR.

### DIFF
--- a/packages/video_player/video_player/CHANGELOG.md
+++ b/packages/video_player/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.12+5
+
+* Locked `video_player_platform_interface` to version `2.1.1` for test to succeed.
+
 ## 0.10.12+4
 
 * Keep handling deprecated Android v1 classes for backward compatibility.
--- a/packages/video_player/video_player/CHANGELOG.md
+++ b/packages/video_player/video_player/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.10.12+5
 
-* Locked `video_player_platform_interface` to version `2.1.1` for test to succeed.
+* Depend on `video_player_platform_interface` version that contains the new `TestHostVideoPlayerApi`
+  in order for tests to pass using the latest dependency.
 
 ## 0.10.12+4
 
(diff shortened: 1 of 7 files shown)


packages/video_player/video_player_web/pubspec.yaml
```diff
@@ -29,5 +29,5 @@ dev_dependencies:
   pedantic: ^1.8.0
 
 environment:
-  sdk: ">=2.1.0 <3.0.0"
+  sdk: ">=2.8.0 <3.0.0"
```
creativecreatorormaybenot, line 32: I am changing this here just to be consistent with the platform interface (the platform interface requires it because of `pigeon v0.1.7`, which annotates it on a file in preparation of null safety.

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.12+5
+
+* Locked `video_player_platform_interface` to version `2.1.1` for test to succeed.
```
creativecreatorormaybenot, line 3: I did not realize this before, but because the test host in the messages in the platform interface were adjusted and the `video_player` **tests** depend on that, CI will fail in the transition period.

So we need to lock the platform interface to the older platform interface version for just this transition period. When part 3/3 is merged, the dependency can be unlocked again.

---

I am sorry, I did not consider this additionally. I think this is the easiest solution. Alternatively, I could simply include the test adjustments in this PR, but I want to include those in part 3/3 instead because that is when the platform interface is logically bumped.

ditman: Can you please merge this on its own, so the tree becomes green again? That way we can keep merging stuff without ignoring the tooling :P

creativecreatorormaybenot: @ditman Sounds great - I will fix it differently in the extra PR however 😉

ditman: I should have caught this when reviewing, my bad!

ditman (approved):
This looks good to me!

PR #3086, opened 2020-09-27
[in_app_purchase] Fix crash iOS sandboxTesting default value

Vardiak:
## Description

Currently, in_app_purchase crashes when initiating a payment if `sandboxTesting` of `PurchaseParam` is `null`. But it is actually its default value (contrary to the definition which states that default value is false). In this pull request, the default value is fixed & I added an assert to check if the user didn't pass null explicitly which would cause a crash.

## Related Issues

https://github.com/flutter/flutter/issues/65767

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/in_app_purchase/lib/src/in_app_purchase/purchase_details.dart
+++ b/packages/in_app_purchase/lib/src/in_app_purchase/purchase_details.dart
@@ -79,7 +79,8 @@ class PurchaseParam {
   PurchaseParam(
       {@required this.productDetails,
       this.applicationUserName,
-      this.sandboxTesting});
+      this.sandboxTesting = false})
+      : assert(sandboxTesting != null);
 
   /// The product to create payment for.
   ///


Vardiak:
@googlebot I signed it!

cyanglaz:
@Vardiak Thanks for the fix! Could you add a test for this? We would also need to bump the minor version of the plugin since the public API has changed.

Vardiak:
> @Vardiak Thanks for the fix! Could you add a test for this?
> 
> In order to make it non-public API change fix, I would recommend to fix it on the OBJC side. We can always allow this value to be null and in the OBJC side, we just take the null value as `false`.

I'm not a OBJC developer, I may be able to help fix it there but I don't know what kind of test you are expecting me to write, since the fix is straightforward.

cyanglaz:
@Vardiak The test can be "if the `sandboxTesting` is nil, there is no crash"

Vardiak:
#3110 handles the issue at platform level, which is a better implementation

PR #3087, opened 2020-09-27
[package_info] Migrate package_info to federated design

mehmetf:
This is the only plugin left that is using an odd structure for supplying macos functionality. Let's migrate it.


mehmetf:
@ditman What is the proper set of steps to migrate a plugin to federated design? I assume I need to first write and publish the interface, then migrate the main plugin then finally migrate macos?

mehmetf:
Also I would appreciate it if you review the interface and the front end design. `PackageInfo` is a bit unique among plugins since its primary class is actually a data class rather than a plugin class. I left most of the functionality intact and opted to provide a Map of values from a `getAll` function for the interface so the data class can populate itself. I was not sure if there was another way of making this non-breaking.

mehmetf:
This plugin is being migrated off to package_info_plus and it seems like the work is already done. Closing this PR.

ditman:
Thanks for understanding @mehmetf!

PR #2882, opened 2020-07-16
[video_player] Add cookie support for network resources

jaysephjw:
## Description

Add cookie support for video_player network resources.

Avoids the problem of relying on `"AVURLAssetHTTPHeaderFieldsKey"` (undocumented ios API) seen in #953 and #897 by only supporting cookies.

**I Broke this change up in to 3 parts, #3136 #3137 and #3138. Review each step individually**.

#### Implementation

Dart:
Optionally include a List<Cookie> when calling VideoPlayerController.network
Pigeon these over to hosts via List<String>, which each cookie as a "Set-Cookie" value strings. 

iOS: 
Parse the strings into NSHTTPCookies and attach them using `AVURLAssetHTTPCookiesKey`

Android: 
Parse the strings into HttpCookies so we can then break name=value pairs back out, and build a "Cookie" header for EXOPlayer

Web: 
Set the cookies onto the document with the `document.cookie` setter, which supports full "Set-Cooke" value strings. _Note: Haven't tested web yet. Are these still there after the player disposes? Can we cleanup?_

## Related Issues

* Implements https://github.com/flutter/flutter/issues/61638
* Alleviates https://github.com/flutter/flutter/issues/16466 for users who just need a Cookie header

## Testing done

Tested on iOS and Android with an HLS server that requires an auth cookie. 

## Breaking Change

- [X] No, this is *not* a breaking change.

## Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I read and followed the [Flutter Style Guide].
- [X] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [X] I signed the [CLA].
- [X] I am willing to follow-up on review comments in a timely manner.
- [X] I have tested on Android
- [X] I have tested on iOS

TODO: 
- [ ] Test on web
- [X] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [X] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [X] All existing and new tests are passing.
- [X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [X] I updated CHANGELOG.md to add a description of the change.

PR #3136, opened 2020-10-11
[video_player_platform_interface] Add cookie support (1/3)

jaysephjw:
**Part 1 of #2882 - see main description there.**

* Add a cookies field for DataSourceType.network requests. 
* Re-run pigeon. Most diffs are formatting changes from that - I have left them in to avoid hand editing generated code. I used the same pigeon version.


jaysephjw:
Achieved in #3671!

PR #3140, opened 2020-10-12
[file_selector] initial implementation

tugorez:
## Description

Add the API for the file_selector plugin.

## Related Issues

https://github.com/flutter/flutter/issues/65403

## Checklist

- [x] I signed the [CLA].

## Breaking Change


tugorez:
@jasonpanelli @ditman PTAL

packages/file_selector/file_selector/README.md
```diff
@@ -0,0 +1,3 @@
+# file_selector
+
+A Flutter plugin that manages files and interactions with file dialogs.
```
ditman, line 3: This README needs a huge amount of love :)

packages/file_selector/file_selector/example/README.md
```diff
@@ -0,0 +1,16 @@
+# example
+
+A new Flutter project.
```
ditman, line 3: This also needs a little bit of love

packages/file_selector/file_selector/pubspec.yaml
```diff
@@ -0,0 +1,21 @@
+name: file_selector
+description: Flutter plugin for opening and saving files.
+homepage: https://github.com/flutter/plugins/tree/master/packages/file_selector/file_selector
+version: 1.0.0
```
ditman, line 4: Let's not make this 1.0.0 until it has gained some more users and stability. 1.0.0 is being used to mark plugins as "production ready" (I'm OK with... 0.7.0 :P)

ditman (changes requested):
I think we need to bump down the initial version of this plugin to adjust expectations from users :)

Other than that, let's get this shipped!

ditman:
Also cc @stuartmorgan: the plugin is coming!

packages/file_selector/file_selector/example/lib/main.dart
```diff
@@ -0,0 +1,298 @@
...
+              textColor: Colors.white,
+              child: Text('Press to try opening a file'),
+              onPressed: () => Navigator.pushNamed(context, '/open'),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+}
```
stuartmorgan, line 298: Unless I missed it, neither `openFiles` nor `getDirectoryPath` are exercised/demonstrated in this example.

packages/file_selector/file_selector/example/lib/main.dart
```diff
@@ -0,0 +1,298 @@
...
+        context: context,
+        builder: (context) {
+          return ImageDisplay(file: file);
+        });
+  }
+
+  void _onOpenTextFile() async {
+    final typeGroup = XTypeGroup(label: 'images', extensions: ['txt', 'json']);
+
+    final file = await openFile(acceptedTypeGroups: [typeGroup]);
```
stuartmorgan, line 132: The only examples of `openFile` use exactly one accepted type group; there should ideally also be an example with none, and an example with multiple.

stuartmorgan (commented):
Some very high level comments based on an initial skim.

In addition to the comments below: why isn't there a driver/integration test as part of this?

packages/file_selector/file_selector/example/README.md
tugorez: PTAL

packages/file_selector/file_selector/pubspec.yaml
tugorez: PTAL

packages/file_selector/file_selector/README.md
tugorez: PTAL

packages/file_selector/file_selector/example/lib/main.dart
tugorez: I agree that we require more examples/ better documenttation for those particular cases. In that sense: 

- I revamped the example app to make the process of adding new examples easier. 
- I added a new example "open multiple images" that exercises the usage of multiple XTypeGroups

Given that this is going to be published as v0.7.0, leaving room for changes/improvements: Can we merge this as it currently is (so we can start with platform implementations) and add more examples/improve the docs as follow ups?

packages/file_selector/file_selector/test/file_selector_test.dart
```diff
@@ -0,0 +1,91 @@
...
+    )).thenAnswer((_) => Future.value([file]));
+
+    final result = await openFiles(
+        acceptedTypeGroups: [typeGroup],
+        initialDirectory: 'dir',
+        confirmButtonText: 'load');
+
+    expect(result, isNotNull);
+  });
+}
```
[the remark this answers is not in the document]
tugorez: I might be missing something but my impression is that the platform implementations should test all those behaviours as at this particular moment the API is only receiving arguments and calling the platform interface with those.

packages/file_selector/file_selector/example/lib/main.dart
tugorez: Thanks for the feedback Stuart :) 
I added a new example for `openFiles`. Can we wait until the first non-web implementation to add an example for getDirectoryPath? Please take a look at my other comment.

tugorez:
> Some very high level comments based on an initial skim.
> 
> In addition to the comments below: why isn't there a driver/integration test as part of this?

Those tests should and will be added as part as each platform implementation.

packages/file_selector/file_selector/test/file_selector_test.dart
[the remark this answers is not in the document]
stuartmorgan: The platform implementations should test the API surface of the platform implementations. This plugin's test should be providing and end-to-end test, including the API that everyone will actually be using.

packages/file_selector/file_selector/example/lib/main.dart
stuartmorgan: We should be setting a good precedent in this plugin for what the examples platform implementations should cover. If it's easy to add examples, why not just do it now instead of hoping that someone will remember to come back and add it later?

packages/file_selector/file_selector/example/lib/main.dart
stuartmorgan: > Can we wait until the first non-web implementation

There are currently zero implementations of this plugin; why would web be the standard that defines what's covered in the main plugin?

I have several desktop implementations essentially complete locally, so it's not like we're talking about a long delay before adding other implementations.

packages/file_selector/file_selector/test/file_selector_test.dart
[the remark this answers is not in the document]
stuartmorgan: Sorry, I was thinking of integration tests when writing that last comment. For unit tests:

>  the API is only receiving arguments and calling the platform interface with those.

How do you know that the API is doing that correctly without tests?

stuartmorgan (changes requested):

stuartmorgan:
> > Some very high level comments based on an initial skim.
> > In addition to the comments below: why isn't there a driver/integration test as part of this?
> 
> Those tests should and will be added as part as each platform implementation.

Every plugin I've implemented for desktop here has integration tests at both layers. Why would we do less testing in this plugin?

stuartmorgan:
Oh, right. I wasn't thinking about the fact that this is UI based; sorry about that. Yes, that does probably mean integration tests can't live at this layer.

(We should still unit test the app-facing API though, using a mock platform interface implementaiton, per my other comments.)

ditman:
> We should still unit test the app-facing API though, using a mock platform interface implementaiton, per my other comments.

@stuartmorgan yes, those should be easy to add; it's just verify that the mock platform methods are being called (if done with Mockito and a Mock platform implementation) or by doing the method logging technique.

packages/file_selector/file_selector/test/file_selector_test.dart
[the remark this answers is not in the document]
tugorez: Hi @stuartmorgan, no problem at all :)!

> How do you know that the API is doing that correctly without tests?

Because we are mocking the platform interface and making it return something only when it is called with specific arguments. 

For instance, on this test, we indicate the mock to return `[file]` only if it is called with those arguments.
```dart
    when(mock.openFiles(
      acceptedTypeGroups: [typeGroup],
      initialDirectory: 'dir',
      confirmButtonText: 'load',
    )).thenAnswer((_) => Future.value([file]));
```
And then we call the API and make sure it returns `[file]`. 
```dart
    final result = await openFiles(
        acceptedTypeGroups: [typeGroup],
        initialDirectory: 'dir',
        confirmButtonText: 'load');

    expect(result, [file]);
```
Currently the only case where this test would pass is if the API returns the result of calling platform interface with the same arguments.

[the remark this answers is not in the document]
stuartmorgan: >> How do you know that the API is doing that correctly **without tests?**
>
> [...]
> For instance, **on this test**

That's a test, so isn't addressing my question.

This comment thread started because I was pointing out that there are many cases not covered by your unit tests (e.g., not providing the optional arguments). Your argument was that you don't need to test those cases because they are just calling through. I'm not asking how unit tests work, I'm asking how you can assert that the cases that you are *not* testing are, in fact, calling through correctly?

Let me put this a different way: which test will fail if someone adds code to `openFile` that attempts to do some preprocessing on, say, `initialDirectory`, without null-checking it?

packages/file_selector/file_selector/example/lib/main.dart
tugorez: Thanks for the feedback Stuart! I already added an example page for getDirectoryPath. With that I think our examples are now covering all the API methods. PTAL

packages/file_selector/file_selector/example/lib/main.dart
tugorez: I added an example for `getDirectoryPath`. PTAL

packages/file_selector/file_selector/test/file_selector_test.dart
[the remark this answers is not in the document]
tugorez: Hi Stuart, I expanded the tests coverage. PTAL

packages/file_selector/file_selector/example/lib/main.dart
tugorez: Friendly ping!

packages/file_selector/file_selector/example/lib/get_directory_page.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:file_selector/file_selector.dart';
+import 'package:flutter/material.dart';
+
+/// Screen that shows an example of openFile(s)
```
ditman, line 4: This screen shows an example of getDirectoryPath, it seems :P

ditman (approved):
Thanks for addressing this! I think it's ready to go, even though, there's a small change incoming regarding the `cross_file` package (we should be able to hide that fact from the platform_interface package, possibly!)

If Stuart is happy with the current implementation, let's get this merged!

ditman:
Note to all: `package: cross_file` is now a thing:

https://pub.dev/packages/cross_file

packages/file_selector/file_selector/example/lib/get_directory_page.dart
```diff
@@ -0,0 +1,66 @@
...
+class GetDirectoryPage extends StatelessWidget {
+  void _getDirectoryPath(BuildContext context) async {
+    final String initialDirectory = '~/Desktop';
+    final String confirmButtonText = 'Choose a cool directory';
```
stuartmorgan, line 8: Nit: This is a *lot* of text for a button. Maybe pick something more realistic to demonstrate, like "Choose"?

packages/file_selector/file_selector/example/lib/get_directory_page.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:file_selector/file_selector.dart';
+import 'package:flutter/material.dart';
+
+/// Screen that shows an example of openFile(s)
+class GetDirectoryPage extends StatelessWidget {
+  void _getDirectoryPath(BuildContext context) async {
+    final String initialDirectory = '~/Desktop';
```
stuartmorgan, line 7: Do the various OS file dialogs actually expand "~"? The robust way to do this would be to use `path_provider`; if we don't want to depend on that I would suggest we just don't set a starting directory in the example.

packages/file_selector/file_selector/example/lib/open_multiple_images_page.dart
```diff
@@ -0,0 +1,85 @@
...
+
+/// Screen that shows an example of openFile(s)
+class OpenMultipleImagesPage extends StatelessWidget {
+  void _openImageFile(BuildContext context) async {
+    final XTypeGroup jpgsTypeGroup = XTypeGroup(
+      label: 'images',
+      extensions: ['jpg', 'jpeg'],
+    );
+    final XTypeGroup pngTypeGroup = XTypeGroup(
+      extensions: ['png'],
```
stuartmorgan, line 14: Why no label?

packages/file_selector/file_selector/example/lib/open_multiple_images_page.dart
```diff
@@ -0,0 +1,85 @@
+import 'dart:io';
+import 'package:flutter/foundation.dart';
+import 'package:file_selector/file_selector.dart';
+import 'package:flutter/material.dart';
+
+/// Screen that shows an example of openFile(s)
+class OpenMultipleImagesPage extends StatelessWidget {
+  void _openImageFile(BuildContext context) async {
+    final XTypeGroup jpgsTypeGroup = XTypeGroup(
+      label: 'images',
```
stuartmorgan, line 10: It seems odd to call this "images" rather than "JPEGs" given that the other option is also an image type.

stuartmorgan (approved):
Thanks for the expanded test coverage and examples! LGTM with some minor nits.

packages/file_selector/file_selector/example/lib/get_directory_page.dart
tugorez: Great catch! Fixed here and (hopefully) everywhere.

tugorez:
> Note to all: `package: cross_file` is now a thing:
> 
> https://pub.dev/packages/cross_file

https://github.com/flutter/flutter/issues/70950 to track the integration. Thanks for the heads up :)!

packages/file_selector/file_selector/example/lib/get_directory_page.dart
tugorez: Done!

packages/file_selector/file_selector/example/lib/open_multiple_images_page.dart
tugorez: Done!

packages/file_selector/file_selector/example/lib/open_multiple_images_page.dart
tugorez: Fixed!

packages/file_selector/file_selector/example/lib/get_directory_page.dart
tugorez: That is a good point. I'm not sure if `~` gets actually expanded so I will play safe here and remove it. 
Thanks @stuartmorgan :)!

ditman: Note that `path_provider` doesn't work in web at all AFAIK (U_U)

ditman:
Let's go with this one! Thanks all for the hard work!

stuartmorgan:
While doing manual testing on Windows with the example I noticed that none of the example calls handle the user pressing cancel (in which case the returned path will be null). Right now canceling any dialog in the example will throw an exception.

ditman:
@stuartmorgan that's an interesting "edge" case for the web; in web you never get called back if the user cancels the selection (no events, null path... nothing!)

PR #3146, opened 2020-10-13
[path_provider] Log errors in the linux example

jiahaog:
When exceptions happen in the linux example, they are silently dropped and do not show up in the logs (flutter/flutter#66593).

This PR prints them out to the console so that we can understand what went wrong.


jiahaog:
@TimWhiting can you please take a look at this :)

TimWhiting:
Hi @jiahaog 
I'm not a member of the flutter team, but did help with the linux implementation and example for this plugin.
I would defer to them for what they want to do with this.

The reason why I implemented it this way was because if one call throws then the rest won't execute leaving the other strings as null. I'm pretty sure it is best practice to catch errors rather than letting them crash leaving null scattered through the app. And we should encourage plugin users to catch errors so they don't run into problems like I mentioned with not initializing something because an error was thrown before it got to that point. However, instead of just assigning an error string in the catch block we could catch each error and then rethrow the last error at the end of initialization if there was an error. But I'm not sure the best way to handle this.

jiahaog:
Thanks, that makes sense. I think to keep things simple we could also print the error to the console instead of just dropping it, I will send an update for this 🙂

jiahaog:
@stuartmorgan can you help to take a look at this?

stuartmorgan (commented):
Code changes seem fine, but please fix the title and/or description to make it clear this is about Linux specifically.

I'm also not sure what the repo policy on changing just examples is; this may need a version and changelog update?

jiahaog:
Updated the description. I don't think this will require a version update - if I understand correctly the example code is not shipped to users on pub. What do you think?

stuartmorgan:
> What do you think?

As I said, I don't know what the repo policy on that is. Someone more familiar with flutter/plugins will need to weigh in.

jiahaog:
Checked with @cyanglaz and we need to update the version and changelog, so did that.

stuartmorgan (approved):
LGTM

jiahaog:
Thanks!

PR #3154, opened 2020-10-14
Prepare url_launcher for the Link widget

mdebbar:
## Description
- **[THIS] Prepare url_launcher for the Link widget.**
- [PENDING] [web] Implement Link for web (https://github.com/flutter/plugins/pull/3155).
- [COMING] Implement Link for native platforms (link TBD).

Issues:

### DIFF
--- a/packages/url_launcher/url_launcher_platform_interface/lib/link.dart
+++ b/packages/url_launcher/url_launcher_platform_interface/lib/link.dart
@@ -59,9 +59,6 @@ class LinkTarget {
   static const blank = LinkTarget._(debugLabel: 'blank');
 }
 
-/// Used to override the delegate that builds the link.
-LinkDelegate linkDelegate;
-
 /// Encapsulates all the information necessary to build a Link widget.
 abstract class LinkInfo {
   /// Called at build time to construct the widget tree under the link.
@@ -80,9 +77,11 @@ abstract class LinkInfo {
 /// Pushes the [routeName] into Flutter's navigation system via a platform
 /// message.
 Future<ByteData> pushRouteNameToFramework(
-  String routeName, {
-  @required bool isUsingRouter,
-}) {
+  BuildContext context,
+  String routeName,
+) {
+  final bool isUsingRouter =
+      context.findAncestorWidgetOfExactType<Router>() != null;
   final MethodCall call = isUsingRouter
       ? MethodCall('pushRouteInformation', <dynamic, dynamic>{
           'location': routeName,
--- a/packages/url_launcher/url_launcher_platform_interface/lib/link.dart
+++ b/packages/url_launcher/url_launcher_platform_interface/lib/link.dart
@@ -77,21 +77,29 @@ abstract class LinkInfo {
 /// message.
 Future<ByteData> pushRouteNameToFramework(
   BuildContext context,
-  String routeName,
-) {
-  final bool isUsingRouter =
-      context.findAncestorWidgetOfExactType<Router>() != null;
-  final MethodCall call = isUsingRouter
-      ? MethodCall('pushRouteInformation', <dynamic, dynamic>{
+  String routeName, {
+  @visibleForTesting
+  bool debugForceRouter = false,
+}) {
+  final Completer<ByteData> completer = Completer<ByteData>();
+  if (debugForceRouter || Router.of(context, nullOk: true) != null) {
+    SystemNavigator.routeInformationUpdated(location: routeName);
+    window.onPlatformMessage(
+      'flutter/navigation',
+      _codec.encodeMethodCall(
+        MethodCall('pushRouteInformation', <dynamic, dynamic>{
           'location': routeName,
           'state': null,
-        })
-      : MethodCall('pushRoute', routeName);
-  final Completer<ByteData> completer = Completer<ByteData>();
-  window.onPlatformMessage(
-    'flutter/navigation',
-    _codec.encodeMethodCall(call),
-    completer.complete,
-  );
+        }),
+      ),
+      completer.complete,
+    );
+  } else {
+    window.onPlatformMessage(
+      'flutter/navigation',
+      _codec.encodeMethodCall(MethodCall('pushRoute', routeName)),
+      completer.complete,
+    );
+  }
   return completer.future;
 }
--- a/packages/url_launcher/url_launcher_platform_interface/lib/link.dart
+++ b/packages/url_launcher/url_launcher_platform_interface/lib/link.dart
@@ -5,7 +5,6 @@
 import 'dart:async';
 import 'dart:ui';
 
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
--- a/packages/url_launcher/url_launcher_platform_interface/lib/link.dart
+++ b/packages/url_launcher/url_launcher_platform_interface/lib/link.dart
@@ -1,0 +1,99 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:ui';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+/// Signature for a function provided by the [Link] widget that instructs it to
+/// follow the link.
+typedef FollowLink = Future<void> Function();
+
+/// Signature for a builder function passed to the [Link] widget to construct
+/// the widget tree under it.
+typedef LinkWidgetBuilder = Widget Function(
+  BuildContext context,
+  FollowLink followLink,
+);
+
+/// Signature for a delegate function to build the [Link] widget.
+typedef LinkDelegate = Widget Function(LinkInfo linkWidget);
+
+final MethodCodec _codec = const JSONMethodCodec();
+
+/// Defines where a Link URL should be open.
+///
+/// This is a class instead of an enum to allow future customizability e.g.
+/// opening a link in a specific iframe.
+class LinkTarget {
+  /// Const private constructor with a [debugLabel] to allow the creation of
+  /// multiple distinct const instances.
+  const LinkTarget._({this.debugLabel});
+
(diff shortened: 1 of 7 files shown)


packages/url_launcher/url_launcher_platform_interface/lib/link.dart
```diff
@@ -0,0 +1,99 @@
...
+          'location': routeName,
+          'state': null,
+        })
+      : MethodCall('pushRoute', routeName);
+  final Completer<ByteData> completer = Completer<ByteData>();
+  window.onPlatformMessage(
+    'flutter/navigation',
+    _codec.encodeMethodCall(call),
+    completer.complete,
+  );
```
ditman, line 97: Is there an imperative Router API that can be called, instead of platform messages?

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
```diff
@@ -0,0 +1,99 @@
...
+
+/// Signature for a delegate function to build the [Link] widget.
+typedef LinkDelegate = Widget Function(LinkInfo linkWidget);
+
+final MethodCodec _codec = const JSONMethodCodec();
+
+/// Defines where a Link URL should be open.
+///
+/// This is a class instead of an enum to allow future customizability e.g.
+/// opening a link in a specific iframe.
```
ditman, line 31: Good change, I had a comment about target being any String in the previous PR

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
```diff
@@ -0,0 +1,99 @@
...
+  /// On Android and iOS, this opens the link in the browser or the relevant
+  /// app.
+  static const blank = LinkTarget._(debugLabel: 'blank');
+}
+
+/// Used to override the delegate that builds the link.
+LinkDelegate linkDelegate;
+
+/// Encapsulates all the information necessary to build a Link widget.
+abstract class LinkInfo {
```
ditman, line 66: This class looks a lot like a PlatformInterface class, why is it not one?

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
```diff
@@ -0,0 +1,99 @@
...
+  /// On the web, this opens the link in a new tab or window (depending on the
+  /// browser and user configuration).
+  ///
+  /// On Android and iOS, this opens the link in the browser or the relevant
+  /// app.
+  static const blank = LinkTarget._(debugLabel: 'blank');
+}
+
+/// Used to override the delegate that builds the link.
+LinkDelegate linkDelegate;
```
ditman, line 63: This also looks like another platform interface method :P

ditman (commented):
This looks great! A few questions, but I don't see anything blocking!

Most of the code are very simple classes, but I think a test for the `pushRouteNameToFramework` is warranted (the isUsingRouter logic, for example)

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
mdebbar: That's a great idea actually.

@chunhtai if I get the Router instance from `Router.of(context)`, is there a way to push the route information name without sending a platform message?

Maybe I can do the same thing with `Navigator.of(context).pushNamed(...)` when there's no Router?

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
mdebbar: Can you elaborate?

This is a class that's shared between web and non-web implementations, that's why I put it in the `_platform_interface` package.

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
mdebbar: Same as my comment above, could you elaborate please? :)

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
chunhtai: There is no imperative RouterAPI for route update, but you should use SystemNavigator.routeInformationUpdated or SystemNavigator.routeUpdated

chunhtai: is the purpose of this function to just update the browser URL bar? calling this method alone will not make navigator/router update their content.

mdebbar: The purpose of this function is to tell the framework to push the given route name. This is invoked when a `Link` widget is pressed. Consequently, the browser URL bar will be updated when the framework sends the `routeUpdated`/`routeInformationUpdated` message to the engine.

mdebbar: Would it work if I do:
```dart
final Router router = Router.of(context);
if (router != null) {
  router.routeInformationProvider.value = {'location': routeName, 'state': null};
} else {
  Navigator.pushNamed(context, routeName);
}
```
?

chunhtai: It wouldn't work for either case, for router the routeInformationProvider is a value listenable, it does not have a setter.
For navigator, it wouldn't work if it has nested navigator, you can probably get the root navigator, but it is not systematic correct.

The best way to do this is somehow trigger didpushRoute/didPushRouteInformation on widgetbindingobserver.
This can be done either by exposing some WidgetBinding API, or send a message through the message channel in the web engine. Besides that you still need to send SystemNavigator.routeInformationUpdated or SystemNavigator.routeUpdated.

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
ditman: So in the federated plugins, a "PlatformInterface" is an abstract class describing the behavior that each platform must implement so the plugin works there. The Web implementation extends the PlatformInterface class and provides the web implementations, and there's normally a default MethodChannel implementation that is used by the native side.

The way you override the linkDelegate, and (especially) the builder inside LinkInfo, reminds me of the pattern above (but you reworked it from scratch), that's what I meant by "these look like PlatformInterface" classes, you're using the same pattern, but without leveraging the existing classes :P

mdebbar: You're right about link delegate. There's one difference though that makes it not fit exactly well in the platform interface pattern. The default link delegate implementation has to live in `url_launcher` (the main package) because it needs to call the real `canLaunch()` and `launch()`. I'll see what I can do to use the platform interface pattern as much as possible.

OTOH, `LinkInfo` isn't being overriden anywhere. It's a class that defines the contract between `Link` and `LinkDelegate`. It's common between `url_launcher` and `url_launcher_web` that's why I put it in the platform interface package.

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
```diff
@@ -0,0 +1,97 @@
...
+  String routeName,
+) {
+  final bool isUsingRouter =
+      context.findAncestorWidgetOfExactType<Router>() != null;
+  final MethodCall call = isUsingRouter
+      ? MethodCall('pushRouteInformation', <dynamic, dynamic>{
+          'location': routeName,
+          'state': null,
+        })
+      : MethodCall('pushRoute', routeName);
```
chunhtai, line 89: This synthesizes the engine sends the message to framework to push the new route. For navigator, this should be sufficient, but for router, it will think this push route information is coming from the engine, so it will not send routeInformationUpdate to the engine. you will still need to update the url/browser history entry by sending the routeInformationUpdated

mdebbar: I can confirm the behavior you are describing for Navigator. I haven't tried the Router yet. Is this a good app to test with (https://github.com/flutter/flutter/pull/63424/) ?

chunhtai: yes, that app should be a good example

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
ditman: About canLaunch/launch, if this ever makes it into the framework, it won't be able to use the url_launcher plugin, right? Then it'll need to be copied internally or exposed somehow. Anyway, I don't see anything blocking. (Note that there's an incoming null-safety migration to the package that might cause some merge conflicts with this PR.)

ditman (approved):
Other than my comments, I think this looks good!

packages/url_launcher/url_launcher_platform_interface/lib/url_launcher_platform_interface.dart
```diff
@@ -38,6 +39,9 @@ abstract class UrlLauncherPlatform extends PlatformInterface {
     _instance = instance;
   }
 
+  /// The delegate used by the Link widget to build itself.
+  LinkDelegate get linkDelegate;
```
ditman, line 43: This is defined as:

```dart
  @override
  LinkDelegate linkDelegate = (LinkInfo linkInfo) => WebLinkDelegate(linkInfo);
```

In the web implementation, how can you override a getter with the function above? Should the signature here be more explicit?

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
mdebbar: Ok I got both to work now. For Navigator, I'm just sending a "pushRoute" message to the framework (which goes through binding's `didPushRoute`). For Router, like you mentioned, I had to send a "pushRouteInformation" message to the framework (which goes through binding's `didPushRouteInformation`) AND call `SystemNavigator.routeInformationUpdated` to update the URL in the browser.

chunhtai (approved):
LGTM

Mravuri96:
This needs to be reverted. It's causing build failures.
https://github.com/flutter/flutter/issues/68545

ditman:
Closing the loop here: Mouad [published](https://pub.dev/packages/url_launcher_web/changelog#015) the fix for the build issue. Apologies for the breakage! /cc @Mravuri96

PR #3155, opened 2020-10-14
[web] Implement Link for web

mdebbar:
## Description
- [COMPLETE] Prepare url_launcher for the Link widget (https://github.com/flutter/plugins/pull/3154).
- **[THIS] [web] Implement Link for web.**
- [COMING] Implement Link for native platforms (link TBD).

Issues:
Fixes https://github.com/flutter/flutter/issues/68545

### DIFF
--- a/packages/url_launcher/url_launcher_web/lib/link.dart
+++ b/packages/url_launcher/url_launcher_web/lib/link.dart
@@ -1,0 +1,282 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:html' as html;
+import 'dart:js_util';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:flutter/services.dart';
+
+import 'package:url_launcher_platform_interface/link.dart';
+
+typedef _ClickListener = void Function(html.MouseEvent);
+
+/// The unique identifier for the view type to be used for link platform views.
+const String viewType = '__url_launcher::link';
+
+/// The name of the property used to set the viewId on the DOM element.
+const String _viewIdProperty = '__url_launcher::link::viewId';
+
+/// Signature for a function that takes a unique [id] and creates an HTML element.
+typedef PlatformViewFactory = html.Element Function(int viewId);
+
+/// Factory that returns the link DOM element for each unique view id.
+PlatformViewFactory get linkViewFactory => _LinkViewController._viewFactory;
+
+/// The delegate for building the [Link] widget on the web.
+///
+/// It uses a platform view to render an anchor element in the DOM.
+class WebLinkDelegate extends StatefulWidget {
+  /// Creates a delegate for the given [link].
+  const WebLinkDelegate(this.link);
+
+  /// Information about the link built by the app.
+  final LinkInfo link;
+
+  @override
+  _WebLinkDelegateState createState() => _WebLinkDelegateState();
+}
+
+class _WebLinkDelegateState extends State<WebLinkDelegate> {
+  _LinkViewController _controller;
+
+  @override
+  void didUpdateWidget(WebLinkDelegate oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (widget.link.uri != oldWidget.link.uri) {
+      _controller?._setUri(widget.link.uri);
+    }
+    if (widget.link.target != oldWidget.link.target) {
+      _controller?._setTarget(widget.link.target);
+    }
+  }
+
+  Future<void> _followLink() {
+    final Completer<void> completer = Completer<void>();
+    _LinkViewController._registerHitTest(
+      _controller,
+      onClick: (html.MouseEvent event) {
+        completer.complete(_onDomClick(event));
+      },
+    );
+    return completer.future;
+  }
+
+  Future<void> _onDomClick(html.MouseEvent event) {
+    if (!widget.link.uri.hasScheme) {
+      // A uri that doesn't have a scheme is an internal route name. In this
+      // case, we push it via Flutter's navigation system instead of letting the
+      // browser handle it.
+      event.preventDefault();
+      final String routeName = widget.link.uri.toString();
+      // TODO(mdebbar): how do we know if `isUsingRouter` should be true or false?
+      return pushRouteNameToFramework(routeName, isUsingRouter: false);
+    }
+
+    // External links will be handled by the browser, so we don't have to do
+    // anything.
+    return Future<void>.value(null);
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Stack(
+      children: <Widget>[
+        widget.link.builder(
+          context,
+          widget.link.isDisabled ? null : _followLink,
+        ),
+        Positioned.fill(
+          child: PlatformViewLink(
+            viewType: viewType,
+            onCreatePlatformView: _createController,
+            surfaceFactory:
+                (BuildContext context, PlatformViewController controller) {
+              return PlatformViewSurface(
+                controller: controller,
+                gestureRecognizers:
+                    Set<Factory<OneSequenceGestureRecognizer>>(),
+                hitTestBehavior: PlatformViewHitTestBehavior.transparent,
+              );
+            },
+          ),
+        ),
+      ],
+    );
+  }
+
+  _LinkViewController _createController(PlatformViewCreationParams params) {
+    _controller = _LinkViewController(params.id);
+    _controller._initialize().then((_) {
+      params.onPlatformViewCreated(params.id);
+    });
+    return _controller
+      .._setUri(widget.link.uri)
+      .._setTarget(widget.link.target);
+  }
+}
+
+/// Controls link views.
+class _LinkViewController extends PlatformViewController {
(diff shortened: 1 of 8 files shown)


packages/url_launcher/url_launcher_web/lib/link.dart
```diff
@@ -0,0 +1,282 @@
...
+
+  String _getHtmlTarget(LinkTarget target) {
+    switch (target) {
+      case LinkTarget.defaultTarget:
+      case LinkTarget.self:
+        return '_self';
+      case LinkTarget.blank:
+        return '_blank';
+      default:
+        throw Exception('Unknown LinkTarget value $target.');
```
ditman, line 228: I think we should make the link target more flexible, and just let a raw string through if specified.

Also, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target):

> Note: When using target, add `rel="noreferrer noopener"` to avoid exploitation of the window.opener API;

packages/url_launcher/url_launcher_web/lib/link.dart
```diff
@@ -0,0 +1,282 @@
...
+
+  @override
+  Future<void> dispose() async {
+    if (_isInitialized) {
+      assert(_instances[viewId] == this);
+      _instances.remove(viewId);
+      if (_instances.isEmpty) {
+        await _clickSubscribtion.cancel();
+      }
+      // Asynchronously dispose this view.
```
ditman, line 252: Is this asynchronous if we `await` for the invokeMethod result?

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
```diff
@@ -43,6 +47,9 @@ class UrlLauncherPlugin extends UrlLauncherPlatform {
   /// Registers this class as the default instance of [UrlLauncherPlatform].
   static void registerWith(Registrar registrar) {
     UrlLauncherPlatform.instance = UrlLauncherPlugin();
+    linkDelegate = (LinkInfo linkInfo) => WebLinkDelegate(linkInfo);
+    // ignore: undefined_prefixed_name
+    ui.platformViewRegistry.registerViewFactory(viewType, linkViewFactory);
```
ditman, line 52: _In TV Salesman Voice:_ You could hide all this logic, if the LinkDelegate was a PlatformInterface! :P And just leave it as: `LinkPlatform.instance = WebLinkPlatform();` :P

ditman (commented):
This looks good to me; I'd move the link.dart file under `src` so users can't import it; that way I wouldn't worry much about making things `_Private`.

My only real concern would be to auto-applying a 'rel' attribute, or letting users set it alongside the 'target', so window.opener does not become an issue.

Everything else is nitpicking, and since most of this file is private anyways, it seems it can be refactored later, if needed/wanted.

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
ditman: (I understand this is not going to stay as a plugin forever, so... whatever makes sense for the framework/engine is probably best :P)

packages/url_launcher/url_launcher_web/lib/link.dart
```diff
@@ -0,0 +1,282 @@
...
+typedef _ClickListener = void Function(html.MouseEvent);
+
+/// The unique identifier for the view type to be used for link platform views.
+const String viewType = '__url_launcher::link';
+
+/// The name of the property used to set the viewId on the DOM element.
+const String _viewIdProperty = '__url_launcher::link::viewId';
+
+/// Signature for a function that takes a unique [id] and creates an HTML element.
+typedef PlatformViewFactory = html.Element Function(int viewId);
```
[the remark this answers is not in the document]
mdebbar: This was a copy from the engine [here](https://github.com/flutter/engine/blob/b715d3f21d66e284364d290b0bd050c76e021d85/lib/web_ui/lib/ui.dart#L87). But I agree the word "platform" is overloaded in the context of plugins, so I'll rename it to `HtmlViewFactory`.

packages/url_launcher/url_launcher_web/lib/link.dart
```diff
@@ -0,0 +1,282 @@
...
+      event.preventDefault();
+      final String routeName = widget.link.uri.toString();
+      // TODO(mdebbar): how do we know if `isUsingRouter` should be true or false?
+      return pushRouteNameToFramework(routeName, isUsingRouter: false);
+    }
+
+    // External links will be handled by the browser, so we don't have to do
+    // anything.
+    return Future<void>.value(null);
+  }
```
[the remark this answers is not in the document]
mdebbar: It'll need to know about widgets soon (to get the right value for `isUsingRouter`).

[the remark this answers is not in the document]
ditman: Can you determine `isUsingRouter` when creating the _controller? :P

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
mdebbar: I would love to make it work like you described but I couldn't. The difference in my case is that the default implementation of "LinkPlatform" has to live in `url_launcher`, not in `url_launcher_platform_interface`. And AFAICT, there's no way to set `LinkPlatform.instance = DefaultLinkPlatform();` in `url_launcher`. I hope I'm wrong here, and hope there's a way to do it :)

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
```diff
@@ -47,11 +47,13 @@ class UrlLauncherPlugin extends UrlLauncherPlatform {
...
   static void registerWith(Registrar registrar) {
     UrlLauncherPlatform.instance = UrlLauncherPlugin();
-    linkDelegate = (LinkInfo linkInfo) => WebLinkDelegate(linkInfo);
     // ignore: undefined_prefixed_name
-    ui.platformViewRegistry.registerViewFactory(viewType, linkViewFactory);
+    ui.platformViewRegistry.registerViewFactory(linkViewType, linkViewFactory);
   }
 
+  @override
+  LinkDelegate linkDelegate = (LinkInfo linkInfo) => WebLinkDelegate(linkInfo);
```
ditman, line 55: This is a getter in the platform interface; does the compiler let you override a getter with this?

ditman (approved):
Ship it, let's give this a shot!

packages/url_launcher/url_launcher_web/lib/link.dart
mdebbar: I want to see a few feature requests before I make target flexible. I'll add the `rel` attribute though.

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
mdebbar: I changed it to a getter. But to answer your question, yes, the compiler lets me override a getter with a field. Remember the field is nothing but a getter and setter bundled together. So technically it is a getter.

packages/url_launcher/url_launcher_web/lib/link.dart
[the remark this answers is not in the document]
mdebbar: Done. What I did is pass the `context` to `LinkViewController` so it can use it later to look the Router up.

ditman (approved):
LGTM!

packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -0,0 +1,295 @@
...
+
+  static Map<int, LinkViewController> _instances = <int, LinkViewController>{};
+
+  static html.Element _viewFactory(int viewId) {
+    return _instances[viewId]?._element;
+  }
+
+  static int _hitTestedViewId;
+
+  static StreamSubscription _clickSubscribtion;
```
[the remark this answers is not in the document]
mdebbar: wow nice catch!

packages/url_launcher/url_launcher_web/test/test_driver/url_launcher_web_integration.dart
```diff
@@ -228,4 +232,88 @@ void main() {
       });
     });
   });
+
+  group('link', () {
+    testWidgets('creates anchor with correct attributes',
```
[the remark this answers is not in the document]
mdebbar: Yeah I want to also add more tests around hit testing but I couldn't make it work so it's gonna take some time. I wanted to land this ASAP to unbreak the url_launcher plugin.

cyanglaz (approved):
LGTM

mdebbar:
The red `submit-queue` is due to the breakage in `url_launcher_platform_interface` that this PR should fix.

Merging.

ditman:
Thanks for publishing this, @mdebbar!

neiljaywarner:
So what version does this work in, is it working now?

ditman:
@neiljaywarner This should work all the way up to flutter channel beta?

neiljaywarner:
@ditman will it get documentation on flutter.dev or "rsoon"?

ditman:
@neiljaywarner there is documentation for the widget, here:

https://pub.dev/documentation/url_launcher/latest/link/Link-class.html

(The widget comes in the url_launcher plugin)

stenlee:
Unfortunately, the **CMD+Left Mouse Click** still not working to open the link in the new tab as users are used in the "classic" web app.

PR #3164, opened 2020-10-16
[device_info][device_info_platform_interface] update to git deps

cyanglaz:
Updated the dependencies from local to git.


blasten (approved):
LGTM

PR #3168, opened 2020-10-19
fix pointer events by setting TestBindingEventSource

nturgut:
## Description

Web integration tests were failing for MousePointer events. This PR is fixing this by setting `TestBindingEventSource` to `TestBindingEventSource.test` for web integration_tests. 

## Related Issues

Fixes https://github.com/flutter/flutter/issues/68502

## Tests

Tests are added to test the pointer gestures.

## Checklist

- [] All existing and new tests are passing.
- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/integration_test/example/integration_test/_example_test_web.dart
+++ b/packages/integration_test/example/integration_test/_example_test_web.dart
@@ -6,7 +6,9 @@
 // tree, read text, and verify that the values of widget properties are correct.
 
 import 'dart:html' as html;
+import 'dart:js_util' as js_util;
 import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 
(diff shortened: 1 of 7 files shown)


nturgut:
@ferhatb This is the PR which will fix scroll for PR flutter/engine#21928

dnfield (changes requested):
Can we add a test that would fail without this change?

Also, I'm a bit surprised that we'd need this on we but not on desktop or mobile. Why is that?

nturgut:
I'm not sure about mobile or desktop actually that's why I only set web. Any opinion if the device

> Can we add a test that would fail without this change?
> 
> Also, I'm a bit surprised that we'd need this on we but not on desktop or mobile. Why is that?

This happens when we generate MouseEvents via dart:html. I will add a web tests, I don't know how  mobile/desktop generate PointerEvents for testing purposes therefore I'm not sure if they will have the same issue.

nturgut:
> I'm not sure about mobile or desktop actually that's why I only set web. Any opinion if the device
> 
> > Can we add a test that would fail without this change?
> > Also, I'm a bit surprised that we'd need this on we but not on desktop or mobile. Why is that?
> 
> This happens when we generate MouseEvents via dart:html. I will add a web tests, I don't know how mobile/desktop generate PointerEvents for testing purposes therefore I'm not sure if they will have the same issue.

Added a test for incrementing the counter by dispatching events. PHAL!

dnfield (approved):
LGTM.

We can worry about whether desktop platforms need this later, since it's not clear yet to me that they would.

packages/integration_test/example/integration_test/_example_test_web.dart
```diff
@@ -21,15 +23,74 @@ void main() {
...
+
+    dispatchEventToMouseLocation((offset.dx).toInt(), (offset.dy).toInt());
+    await tester.pumpAndSettle();
+
+    final Finder finder = find.byKey(const Key('counter'));
+    final Text counter = tester.widget(finder);
+
+    // Verify that counter is clicked.
+    expect(finder, findsOneWidget);
+    expect(counter.data, '1');
```
dnfield, line 54: Actually, can we assert earlier that this is '0'?

nturgut: Good idea! I'll make the change thanks :)

dnfield:
I'm a little confuse dabout the error in CI - why isn't it finding the super method when building?

dnfield (changes requested):
see comments

packages/integration_test/lib/integration_test.dart
```diff
@@ -70,6 +70,22 @@ class IntegrationTestWidgetsFlutterBinding extends LiveTestWidgetsFlutterBinding
...
+  void handlePointerEvent(
+    PointerEvent pointerEvent, {
+    TestBindingEventSource source = TestBindingEventSource.device,
+  }) {
+    if (kIsWeb) {
+      // Web driver does not create real device events. Routing web
+      // integration tests' pointer events with source set to `device`,
+      // results in exceptions.
+      super.handlePointerEvent(pointerEvent,
+          source: TestBindingEventSource.test);
```
dnfield, line 83: nit: trailing comma.

Why aren't we doing this in other test bindings? What's special about this one?

nturgut: I also added the other tests binding and update the comments. Fixing the assert statements didn't help since `WidgetTester.dispatchEvent` ([code](https://github.com/flutter/flutter/blob/master/packages/flutter_test/lib/src/widget_tester.dart#L786)) is not acting on the gesture bindings, using `TestBindingEventSource.device` does not generate the expected results for a gesture.

 `TestBindingEventSource.test` uses `RenderBinding.dispatchEvent`: https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/rendering/binding.dart#L271

nturgut:
@dnfield fixing the assert statements for PointerAdd/PointerRemove didn't help since the dispatchers used for the `TestBindingEventSource.device` also does not perform the operations we want during an end to end test.

Btw, looks like flutter driver is not working with integration_test package anymore, what is the quick solution we have? @ferhatb has a code that we need to get in (which uses scroll gestures)

nturgut:
Stable [channel tests](https://github.com/flutter/plugins/pull/3168/checks?check_run_id=1323465058) do not pass since the method we are using `handlePointerEvents` are not in the STABLE channel.  How to skip these tests (or what is the process this cases)?

stuartmorgan:
Since `integration_test` has moved, this is presumably obsolete; closing.

PR #3177, opened 2020-10-21
Implement Link for native platforms

mdebbar:
## Description
- [COMPLETE] Prepare url_launcher for the Link widget (https://github.com/flutter/plugins/pull/3154).
- [COMPLETE] [web] Implement Link for web (https://github.com/flutter/plugins/pull/3155).
- **[THIS] Implement Link for native platforms.**

Issues:

Fixes https://github.com/flutter/flutter/issues/59690
Fixes https://github.com/flutter/flutter/issues/44865
Fixes https://github.com/flutter/flutter/issues/54281

### DIFF
--- a/packages/url_launcher/url_launcher/lib/src/link.dart
+++ b/packages/url_launcher/url_launcher/lib/src/link.dart
@@ -9,9 +9,6 @@ import 'package:url_launcher/url_launcher.dart';
 import 'package:url_launcher_platform_interface/link.dart';
 import 'package:url_launcher_platform_interface/url_launcher_platform_interface.dart';
 
-export 'package:url_launcher_platform_interface/link.dart'
-    show LinkTarget, LinkWidgetBuilder, FollowLink;
-
 /// A widget that renders a real link on the web, and uses WebViews in native
 /// platforms to open links.
 ///
--- a/packages/url_launcher/url_launcher/lib/src/link.dart
+++ b/packages/url_launcher/url_launcher/lib/src/link.dart
@@ -1,0 +1,131 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:flutter/widgets.dart';
+import 'package:url_launcher/url_launcher.dart';
+import 'package:url_launcher_platform_interface/link.dart';
+import 'package:url_launcher_platform_interface/url_launcher_platform_interface.dart';
+
+export 'package:url_launcher_platform_interface/link.dart'
+    show LinkTarget, LinkWidgetBuilder, FollowLink;
+
+Widget defaultLinkDelegate(LinkInfo link) {
+  return DefaultLinkDelegate(link);
+}
+
+/// A widget that renders a real link on the web, and uses WebViews in native
+/// platforms to open links.
+///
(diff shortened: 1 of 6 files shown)


packages/url_launcher/url_launcher/lib/src/link.dart
```diff
@@ -0,0 +1,135 @@
...
+
+import 'dart:async';
+
+import 'package:flutter/widgets.dart';
+import 'package:url_launcher/url_launcher.dart';
+import 'package:url_launcher_platform_interface/link.dart';
+import 'package:url_launcher_platform_interface/url_launcher_platform_interface.dart';
+
+export 'package:url_launcher_platform_interface/link.dart'
+    show LinkTarget, LinkWidgetBuilder, FollowLink;
```
ditman, line 13: I don't think you need to export from this package; you only need to import these 3 classes to use them here, right?

packages/url_launcher/url_launcher/lib/url_launcher.dart
```diff
@@ -9,6 +9,10 @@ import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 import 'package:url_launcher_platform_interface/url_launcher_platform_interface.dart';
 
+export 'src/link.dart' show Link;
+export 'package:url_launcher_platform_interface/link.dart'
+    show FollowLink, LinkTarget, LinkWidgetBuilder;
```
ditman, line 14: (You do need this export though)

ditman: Another way of going about this, to minimize the exposure of people to the new Link class (it can cause collisions with people's code?) is to have a separate barrel file:

`url_launcher/lib/link.dart` with lines 12, 13, 14 from this pr.

That way people can continue using `url_launcher` pretty much as normal, but if they want the link widget they need to do:

import 'package:url_launcher/link.dart'`

WDYT?

ditman (approved):
Should we export `Link` from a separate barrel file?

packages/url_launcher/url_launcher/lib/src/link.dart
mdebbar: These 3 classes are needed by users. Here are examples of where they may need them:
```dart
Link(
  uri: Uri.parse('http://example.com/foo'),
  target: /**/LinkTarget/**/.blank,
  builder: (BuildContext context, /**/FollowLink/**/ followLink) {
    return Container();
  },
);
```
And in case they need to have a wrapper around the Link widget that takes the `builder` as an argument:
```dart
class GoogleLink extends StatelessWidget {
  GoogleLink({this.builder});

  final /**/LinkWidgetBuilder/**/ builder;

  final Uri uri = Uri.parse('https://google.com');

  @override
    return Link(uri: uri, builder: builder);
  }
}
```

mdebbar: Or maybe you mean that I can put this export in `lib/url_launcher.dart` directly instead of exporting them from here? I feel stupid now lol.

packages/url_launcher/url_launcher/lib/url_launcher.dart
mdebbar: I like it!

ditman (approved):
I like this better, thanks for moving the exports of the `Link` widget to a separate library!

PR #3179, opened 2020-10-21
[integration_test] Add a `run` method for proper reporting of test results

jiahaog:
## Description

This integrates with `directRunTests` from `package:test_core`, and allows users to use `test()` to wrap their tests, instead of only supporting `testWidgets()` and failing silently otherwise. In addition, exceptions thrown in other lifecycle blocks such as `setUp` will be caught and reported.

This is a non-breaking change, so the existing reporting mechanism will continue to work. A global variable is used to ensure that using `run` will disable the existing reporting mechanism.

### Example usage

```dart

void main() => run(_testMain);

void _testMain() {
  test('A test', () {
    expect(true, true);
  });
}
```

But we will want to make it easier for users to use this API. A CLI will be added as a follow up PR to wrap this method as part of flutter/flutter#66264, so users can write the following instead, and the CLI will take care of wrapping the users' script with `run` appropriately.

```dart

void main() {
  test('A test', () {
    expect(true, true);
  });
}
```

## Related Issues

Fixes flutter/flutter#50816.

## Breaking Change


packages/integration_test/example/pubspec.yaml
```diff
@@ -24,6 +24,12 @@ dev_dependencies:
   test: any
   pedantic: ^1.8.0
 
+dependency_overrides:
+  test_api:
+    path: /Users/jiahaog/dev/dart-test/pkgs/test_api
+  test_core:
+    path: /Users/jiahaog/dev/dart-test/pkgs/test_core
```
jiahaog, line 31: Please ignore this for now, this requires a release from package:test_core

packages/integration_test/pubspec.yaml
```diff
@@ -15,8 +15,16 @@ dependencies:
...
+  test_core:
+    path: /Users/jiahaog/dev/dart-test/pkgs/test_core
   vm_service: ^4.2.0
 
+dependency_overrides:
+  test_api:
+    path: /Users/jiahaog/dev/dart-test/pkgs/test_api
+  test_core:
+    path: /Users/jiahaog/dev/dart-test/pkgs/test_core
+
```
jiahaog, line 27: Please ignore this for now, this requires a release from package:test_core

jiahaog (commented):
Sending this draft for review first for preliminary feedback – this is still pending a release from package:test_core so that we can use `directRunTests`.

I also want to restructure the tests a bit, and do a bit more testing locally to ensure that nothing is broken

packages/integration_test/lib/common.dart
```diff
@@ -114,16 +114,20 @@ class Failure {
...
 
   /// Constructor requiring all fields during initialization.
-  Failure(this.methodName, this.details);
+  Failure(this.methodName, this.details, {this.error});
 
   /// Serializes the object to JSON.
   String toJson() {
     return json.encode(<String, String>{
       'methodName': methodName,
+      'error': error.toString(),
```
dnfield, line 130: I think we probably want this to be
```suggestion
      'error': error != null ? error.toString() : null,
```

To avoid having the string literally `'null`' in here.

packages/integration_test/lib/integration_test.dart
```diff
@@ -11,61 +11,117 @@ import 'package:flutter_test/flutter_test.dart';
...
+  return results;
+}
 
-const String _success = 'success';
+String _formatFailureForPlatform(Failure failure) =>
+    '${failure.error} ${failure.details}';
+
+Future<void> _reportResultsToNative(
+    IntegrationTestWidgetsFlutterBinding binding,
+    Map<String, Object> results) async {
```
dnfield, line 81:
```suggestion
    Map<String, Object> results,
  ) async {
```

packages/integration_test/lib/integration_test.dart
```diff
@@ -11,61 +11,117 @@ import 'package:flutter_test/flutter_test.dart';
...
+        'results': {
+          for (final result in binding.results.entries)
+            result.key: result.value is Failure
+                ? _formatFailureForPlatform(result.value)
+                : result.value
+        }
+      },
+    );
+  } on MissingPluginException {
+    print('Warning: integration_test test plugin was not detected.');
```
dnfield, line 102: I think this should be treated as a more serious failure, and probably make us `exit` with some non-zero exit code.

The risk is that if this isn't set up right in the application, it will print what looks like a warning in a sea of log output and exit with zero when tests may have failed.

packages/integration_test/lib/integration_test.dart
```diff
@@ -11,61 +11,117 @@ import 'package:flutter_test/flutter_test.dart';
...
-            })
-          },
-        );
-      } on MissingPluginException {
-        print('Warning: integration_test test plugin was not detected.');
-      }
-      if (!_allTestsPassed.isCompleted) _allTestsPassed.complete(true);
-    });
+    if (!_isUsingLegacyReporting) return;
+
```
dnfield, line 117: We should print a warning here that there's a new/better way to do this, probably with a link to instructions about how to do it. For now that link could be in the flutter/flutter/wiki area.

packages/integration_test/lib/integration_test.dart
```diff
@@ -11,61 +11,117 @@ import 'package:flutter_test/flutter_test.dart';
...
-              return MapEntry(name, result);
-            })
-          },
-        );
-      } on MissingPluginException {
-        print('Warning: integration_test test plugin was not detected.');
-      }
-      if (!_allTestsPassed.isCompleted) _allTestsPassed.complete(true);
-    });
+    if (!_isUsingLegacyReporting) return;
```
dnfield, line 116:
```suggestion
    if (!_isUsingLegacyReporting) {
      return;
    }
```

packages/integration_test/lib/src/reporter.dart
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:test_core/src/runner/engine.dart';
+import 'package:test_core/src/runner/reporter.dart';
```
dnfield, line 8: This will give warnings right? 

We should probably pin the dependency for this to make sure we don't inadventently get broken by a patch release that changes internals.

I'm also surprised we don't need analysis ignores here, like `// ignore: implementation_import` or something. We should enable that lint and add the ignore here.

packages/integration_test/lib/src/reporter.dart
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  void _onDone(bool _) async {
+    _cancel();
+    final Map<String, Object> results = {
+      for (final liveTest in _engine.liveTests)
+        liveTest.test.name: liveTest.state.result.name == success
+            ? success
+            : Failure(
+                liveTest.test.name, liveTest.errors.first.stackTrace.toString(),
```
dnfield, line 33: Is the first stack trace the right one? 

Is the first error the right one?

Is there some way we could add _all_ stack traces and errors to the reporting?

packages/integration_test/pubspec.yaml
```diff
@@ -15,8 +15,16 @@ dependencies:
   flutter_test:
     sdk: flutter
   path: ^1.6.4
+  test_core:
```
dnfield, line 18: @jakemac53 @natebosch - LGTY?

dnfield (commented):
This is pretty exciting - this will fix several classes of problems users can run into currently with the package!

packages/integration_test/pubspec.yaml
natebosch: This can be updated now

```suggestion
  test_core: 0.3.12-nullsafety.6
```

Unfortunately null safety leaves us in a state where we'll need to pin to a single version since we can't really pin to a feature version currently.

packages/integration_test/lib/common.dart
natebosch:
```suggestion
      'error': error?.toString(),
```

dnfield: @natebosch - that would have the unfortunate side effect of emitting `'null`' where we probably want `null`.

natebosch: It shouldn't. `'$error'` would do that, but `error?.toString()` should stay `null` and not `'null'`.

dnfield: Ahhh you are right.

packages/integration_test/lib/integration_test.dart
jiahaog: We don't want the application to exit when it is invoked using flutter driver, as the native plugin is not used. The platform side plugin is only used when running with FTL or through Xcode. The responsibility of failing when the infrastructure is not set up properly (i.e. the native plugin is not installed) should probably be up to the host side invoker though, and probably not here.

packages/integration_test/lib/integration_test.dart
jiahaog: Good idea, I added more text and a TODO instead, so that we can follow up after we have the executable to mention using the executable or even link to the readme

packages/integration_test/lib/src/reporter.dart
jiahaog: Yeah the lint isn't set up for this. This is a bit out of scope of this PR though – I have started #3187 to see if we can turn on the lint. There are some existing failures in the repo, will follow up tomorrow.

packages/integration_test/lib/src/reporter.dart
jiahaog: This was taken from the way other Dart tests are integrated with the Google3 infrastructure. I'm not sure if we should do anything else, @natebosch would you know more about this?

packages/integration_test/pubspec.yaml
jiahaog: Thanks, I updated it but there's a conflict with flutter_test that is pinned to the previous version. I have opened https://github.com/flutter/flutter/pull/68771 to update that separately.

jiahaog:
Still pending https://github.com/flutter/flutter/pull/68771 for CI to pass, but this is ready for review now.

jiahaog:
package:integration_test is being moved to the SDK repo, so this PR is superseded by https://github.com/flutter/plugins/pull/3179

PR #3182, opened 2020-10-21
[Camera] Return correct error message when supplied file points to a directory.

Sempakonka:
## Description

It is easy for a Camera plugin user to think that in the method startVideoRecording and takePicture they need to pass a directory file path. This PR points them in the right direction by supplying a better error message.

## Related Issues

[issue 56012](https://github.com/flutter/flutter/issues/56012)

## Checklist

- [x] I signed the [CLA].

## Breaking Change


Sempakonka:
@googlebot I consent.

mvanbeusekom:
@googlebot I consent.

packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -215,9 +215,10 @@ SurfaceTextureEntry getFlutterTexture() {
   public void takePicture(String filePath, @NonNull final Result result) {
     final File file = new File(filePath);
 
-    if (file.exists()) {
-      result.error(
-          "fileExists", "File at path '" + filePath + "' already exists. Cannot overwrite.", null);
+    FilePathValidator validator = new FilePathValidator(filePath);
+    validator.validate();
+    if (!validator.isValid()) {
+      result.error("filePathInvalid", validator.getErrorMessage(), null);
```
amirh, line 221: Changing the error code is a breaking change, I suggest we keep the error code to be `fileExists` to avoid a major version bump.

packages/camera/android/src/main/java/io/flutter/plugins/camera/FilePathValidator.java
```diff
@@ -0,0 +1,48 @@
...
+    this(filePath, new FileFactory());
+  }
+
+  FilePathValidator(String filePath, FilePathValidator.FileFactory factory) {
+    this.filePath = filePath;
+    this.fileFactory = factory;
+  }
+
+  public void validate() {
+    File file = fileFactory.makeFile(this.filePath);
```
bparrishMines, line 27: nit:
```suggestion
    final File file = fileFactory.makeFile(this.filePath);
```

packages/camera/android/src/main/java/io/flutter/plugins/camera/FilePathValidator.java
```diff
@@ -0,0 +1,48 @@
+package io.flutter.plugins.camera;
+
+import java.io.File;
+
+class FilePathValidator {
```
bparrishMines, line 5: Can you write a small javadoc explaining what the class does and that `validate()` must be called before calling `isValid()` or `getErrorMessage()`?

packages/camera/android/src/test/java/io/flutter/plugins/camera/FilePathValidatorTest.java
```diff
@@ -0,0 +1,97 @@
...
+    String filePath = "foo";
+
+    when(mockFile.exists()).thenReturn(true);
+    when(mockFactory.makeFile(filePath)).thenReturn(mockFile);
+
+    FilePathValidator validator = new FilePathValidator(filePath, mockFactory);
+    validator.validate();
+
+    assertEquals(
+        "File at path '" + filePath + "' already exists. Cannot overwrite.",
```
bparrishMines, line 47: nit:
Since "foo" is already known to be the file path, we typically just add it to the String in tests
```suggestion
        "File at path 'foo' already exists. Cannot overwrite.",
```

packages/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -215,9 +215,10 @@ SurfaceTextureEntry getFlutterTexture() {
   public void takePicture(String filePath, @NonNull final Result result) {
     final File file = new File(filePath);
 
-    if (file.exists()) {
-      result.error(
-          "fileExists", "File at path '" + filePath + "' already exists. Cannot overwrite.", null);
+    FilePathValidator validator = new FilePathValidator(filePath);
+    validator.validate();
+    if (!validator.isValid()) {
```
amirh, line 220: There's added cognitive load on the reader now, looking the code before it's pretty clear that we're checking whether the path exist. Reading the code now I see that we're validating and it's not clear exactly what is being validated just from looking at this file.

I would avoid the extra class, for testability purposes one alternative would be to add a package private method that receives a File object and in the unit test pass in some mock File.

Though I'm wondering whether we want any change in the Android behavior at all (see my top level comment)

packages/camera/android/src/test/java/io/flutter/plugins/camera/FilePathValidatorTest.java
```diff
@@ -0,0 +1,97 @@
...
+    when(mockFile.exists()).thenReturn(true);
+    when(mockFile.isDirectory()).thenReturn(true);
+    when(mockFactory.makeFile(filePath)).thenReturn(mockFile);
+
+    FilePathValidator validator = new FilePathValidator(filePath, mockFactory);
+    validator.validate();
+
+    assertEquals(
+        "filePathInvalid",
+        "File at path '"
```
bparrishMines, line 28: nit:
It's typically preferred to include "foo" in the String instead of concatenating the variable in tests.

bparrishMines (changes requested):

amirh (changes requested):
I left a few notes on the implementation, though more generally I'm wondering whether this confusion applies to Android only? if not would it make more sense to handle it on the Dart side?

Whatever we end up doing I would try to add minimal complexity to the codebase e.g if printing some extra information from the Dart side is simple enough maybe we should go with it, if we believe better documentation is enough here (I'm not sure?) maybe that's all we need to add?

mvanbeusekom:
Closing this PR, as it will become obsolete in version 0.6.0 of the camera plugin

PR #3188, opened 2020-10-22
[share] Fix bug on iPad where `origin` is null and enable XCUITests

cyanglaz:
Creates a XCUITests target in the share plugin.

Also fixes https://github.com/flutter/flutter/issues/66485


packages/share/example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -1,87 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
jmagman, line 1: Did you mean to remove this?

packages/share/example/ios/RunnerUITests/FLTShareExampleUITests.m
```diff
@@ -0,0 +1,43 @@
...
+
+- (void)testShareWithEmptyOrigin {
+  XCUIApplication* app = [[XCUIApplication alloc] init];
+  [app launch];
+
+  XCUIElement* shareWithEmptyOriginButton = [app.buttons
+      elementMatchingPredicate:[NSPredicate
+                                   predicateWithFormat:@"label == %@", @"Share With Empty Origin"]];
+  if (![shareWithEmptyOriginButton waitForExistenceWithTimeout:30]) {
+    NSLog(@"%@", app.debugDescription);
```
jmagman, line 25: Remove the NSLog

packages/share/example/ios/RunnerUITests/FLTShareExampleUITests.m
```diff
@@ -0,0 +1,43 @@
...
+}
+
+- (void)testShareWithEmptyOrigin {
+  XCUIApplication* app = [[XCUIApplication alloc] init];
+  [app launch];
+
+  XCUIElement* shareWithEmptyOriginButton = [app.buttons
+      elementMatchingPredicate:[NSPredicate
+                                   predicateWithFormat:@"label == %@", @"Share With Empty Origin"]];
+  if (![shareWithEmptyOriginButton waitForExistenceWithTimeout:30]) {
```
jmagman, line 24: 30 seconds seems really long.

packages/share/example/ios/RunnerUITests/FLTShareExampleUITests.m
```diff
@@ -0,0 +1,43 @@
...
+  XCTAssertNotNil(shareWithEmptyOriginButton);
+  [shareWithEmptyOriginButton tap];
+
+  // Find the share popup.
+  XCUIElement* activityListView = [app.otherElements
+      elementMatchingPredicate:[NSPredicate
+                                   predicateWithFormat:@"identifier == %@", @"ActivityListView"]];
+  if (![activityListView waitForExistenceWithTimeout:30]) {
+    NSLog(@"%@", app.debugDescription);
+    XCTFail(@"Failed due to not able to find activityListView with %@ seconds", @(30));
```
jmagman, line 38: literals as `@30`

packages/share/example/ios/RunnerUITests/FLTShareExampleUITests.m
```diff
@@ -0,0 +1,43 @@
...
+
+  XCTAssertNotNil(shareWithEmptyOriginButton);
+  [shareWithEmptyOriginButton tap];
+
+  // Find the share popup.
+  XCUIElement* activityListView = [app.otherElements
+      elementMatchingPredicate:[NSPredicate
+                                   predicateWithFormat:@"identifier == %@", @"ActivityListView"]];
+  if (![activityListView waitForExistenceWithTimeout:30]) {
+    NSLog(@"%@", app.debugDescription);
```
jmagman, line 37: Remove

packages/share/example/ios/RunnerUITests/FLTShareExampleUITests.m
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  XCTAssertNotNil(shareWithEmptyOriginButton);
+  [shareWithEmptyOriginButton tap];
+
+  // Find the share popup.
+  XCUIElement* activityListView = [app.otherElements
+      elementMatchingPredicate:[NSPredicate
+                                   predicateWithFormat:@"identifier == %@", @"ActivityListView"]];
+  if (![activityListView waitForExistenceWithTimeout:30]) {
```
jmagman, line 36: Ditto, 30 seconds seems too long.

packages/share/example/ios/share_exampleUITests/Info.plist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jmagman, line 1: Is this unused?

packages/share/example/ios/RunnerUITests/FLTShareExampleUITests.m
cyanglaz: We used the same waiting length on scenario app tests https://github.com/flutter/engine/blob/master/testing/scenario_app/ios/Scenarios/ScenariosUITests/GoldenPlatformViewTests.m#L11
I think the rational is that on ci, the animation might take some time to run and might cause false negative to happen.
This returns immediately after finding the item, so I think setting time-out to be longer should be fine?

packages/share/example/ios/RunnerUITests/FLTShareExampleUITests.m
cyanglaz: This is beneficial so that when a failure happened and we can see what's in the a11y tree and why the particular item was not found in the a11y tree.

packages/share/example/ios/share_exampleUITests/Info.plist
cyanglaz: Thanks for the good catch! Removed.

packages/share/example/ios/RunnerUITests/FLTShareExampleUITests.m
cyanglaz: Fixed. Moved all the "30" out to a static

packages/share/example/ios/RunnerUITests/FLTShareExampleUITests.m
jmagman: Can you print it when it fails?  Your test results will get clogged pretty fast with a bunch of debugging lines.

jmagman: Also NSLog is soft deprecated, should be using os_log.

cyanglaz: The log only happens if `![shareWithEmptyOriginButton waitForExistenceWithTimeout:kSecondsToWaitWhenFindingElements]` and we send a `XCTFail` signal afterwards. So it doesn't print unless the test fails.

Thanks for the info of os_log, updated!

cyanglaz: Or did you mean to combine the "app.debugDescription" log within XCTFail?

jmagman (approved):
LGTM!

PR #3193, opened 2020-10-23
[Camera] Made CameraController.isDisposed publicly accessible. Added unit Tests for the new implementation.

Sempakonka:
## Description

Currently when you call `CameraController.Dispose()` there is no way of knowing whether you disposed that CameraController.
With this PR, the user can check if that CameraController has been disposed, by getting `CameraController.isDisposed`.

Since the current `test/camera_test.dart` file contains test for the `CameraController` class located in the `lib/new/src/camera_controller.dart` I have moved it into the `test/new/camera_test.dart` sub folder. 

I have added a new `test/camera_test.dart` file which contains the unit tests covering the new `isDisposed` property which was added to the `CameraController` class located in the current `lib/camera.dart` implementation.

Additionally I have added unit tests to the `test/new/camera_test.dart` file to ensure the `isDisposed` property (already) implemented in the `CameraController` class located in the `lib/new/src/camera_controller.dart` file is working properly.

## Related Issues

Fixes [flutter/flutter#51630](https://github.com/flutter/flutter/issues/51630)

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/camera/lib/camera.dart
+++ b/packages/camera/lib/camera.dart
@@ -307,8 +307,12 @@ class CameraController extends ValueNotifier<CameraValue> {
   StreamSubscription<dynamic> _imageStreamSubscription;
   Completer<void> _creatingCompleter;
 
-  /// True after [CameraController.dispose] has completed successfully.
-  bool get isDisposed => _isDisposed;
+  /// Checks whether [CameraController.dispose] has completed successfully.
+  ///
+  /// This is a no-op when asserts are disabled.
+  void debugCheckIsDisposed() {
+    assert(_isDisposed);
+  }
 
   /// Initializes the camera on the device.
   ///
--- a/packages/camera/lib/camera.dart
+++ b/packages/camera/lib/camera.dart
@@ -307,6 +307,9 @@ class CameraController extends ValueNotifier<CameraValue> {
   StreamSubscription<dynamic> _imageStreamSubscription;
   Completer<void> _creatingCompleter;
 
+  /// True after [CameraController.dispose] has completed successfully.
+  bool get isDisposed => _isDisposed;
+
   /// Initializes the camera on the device.
   ///
   /// Throws a [CameraException] if the initialization fails.
--- a/packages/camera/test/camera_test.dart
+++ b/packages/camera/test/camera_test.dart
@@ -6,7 +6,8 @@ import 'package:flutter_test/flutter_test.dart';
 
 void main() {
   group('camera', () {
-    test('debugCheckIsDisposed should not throw assertion error when disposed', () {
+    test('debugCheckIsDisposed should not throw assertion error when disposed',
+        () {
       final MockCameraDescription description = MockCameraDescription();
       final CameraController controller = CameraController(
         description,
@@ -23,8 +24,7 @@ void main() {
       }
     });
 
-    test(
-        'debugCheckIsDisposed should throw assertion error when not disposed',
+    test('debugCheckIsDisposed should throw assertion error when not disposed',
         () {
       final MockCameraDescription description = MockCameraDescription();
       final CameraController controller = CameraController(
--- a/packages/camera/test/camera_test.dart
+++ b/packages/camera/test/camera_test.dart
@@ -6,7 +6,7 @@ import 'package:flutter_test/flutter_test.dart';
 
 void main() {
   group('camera', () {
-    test("isDisposed true when disposed", () {
+    test('debugCheckIsDisposed should not throw assertion error when disposed', () {
       final MockCameraDescription description = MockCameraDescription();
       final CameraController controller = CameraController(
         description,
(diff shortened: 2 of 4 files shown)


packages/camera/test/camera_test.dart
```diff
@@ -2,115 +2,32 @@
...
+    final MockCameraDescription description = MockCameraDescription();
+    final CameraController controller = CameraController(
+      description,
+      ResolutionPreset.low,
+    );
+
+    controller.dispose();
+    expect(controller.isDisposed, isTrue);
+  });
+  test("isDisposed false when not disposed", () {
```
bparrishMines, line 19: nit:
```suggestion
  });
  
  test("isDisposed false when not disposed", () {
```

packages/camera/test/new/camera_test.dart
```diff
@@ -0,0 +1,139 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
bparrishMines, line 1: The `lib/new` currently has no functionality and will most likely be removed in: https://github.com/flutter/plugins/pull/3229. So, you can just delete this file. And you don't need to add them back to `lib/camera_test.dart`.

mvanbeusekom:
@googlebot I consent.

packages/camera/lib/camera.dart
```diff
@@ -307,6 +307,9 @@ class CameraController extends ValueNotifier<CameraValue> {
   StreamSubscription<dynamic> _imageStreamSubscription;
   Completer<void> _creatingCompleter;
 
+  /// True after [CameraController.dispose] has completed successfully.
+  bool get isDisposed => _isDisposed;
```
dnfield, line 311: we typically expose things like this guarded by asserts, so they're only valid in debug mode.

This is to discourage people from adding checks like this to release mode code, where it should never really be necessary. 

IOW, making it easier to assert program correctness is great. Allowing someone to bog down code with 1000s of small checks that really should only be done during development is not as preferable.

dnfield (changes requested):
This should be guarded by asserts to discourage use in release mode.

packages/camera/lib/camera.dart
mvanbeusekom: thank you for providing feedback. I agree with the idea on only allowing this during during development. However I am not really sure how to "expose things like this guarded by asserts"?

Does this imply we don't expose a property but rather add assert statements to the methods? I feel a bit stupid here :P, but I just cannot wrap my head around it.

dnfield:
```suggestion
  /// Checks whether [CameraController.dispose] has completed successfully.
  ///
  /// This is a no-op when asserts are disabled.
  void debugDisposed() => assert(_isDisposed);
```

dnfield: Or maybe `debugCheckDisposed` or something.

mvanbeusekom: @bparrishMines what do you think? 

In general it would be bad to have any code that can potentially access the `CameraController` when it is already disposed. This means having the `debugCheckIsDisposed` around should be helpful for developers trying to debug their application, but should not be used in production. This will also prevent code constructs like this, which has high risk on state conflicts:

```dart
var cameraController = CameraController();

// do stuff with the camera controller
...

if (cameraController.isDisposed) {
  cameraController = CameraController();
}

// continue doing stuff with the camera controller
```

bparrishMines: Yea, that makes sense to avoid that type of pattern. I'm fine with going with a `debugCheckIsDisposed` instead.

mvanbeusekom: @dnfield thank you for clarifying the coding structure that was very helpful. I took your recommendations and updates the pull request. 

Note that I had to change:
```dart
void debugCheckIsDisposed() => assert(_isDisposed);
```
to:
```dart
void debugCheckIsDisposed() {
  assert(_isDisposed);
}
```

otherwise we would receive a compile error.

I would appreciate it if you could have another look.

dnfield (approved):
LGTM if LG to @bparrishMines

packages/camera/test/camera_test.dart
```diff
@@ -0,0 +1,49 @@
...
+        () {
+      final MockCameraDescription description = MockCameraDescription();
+      final CameraController controller = CameraController(
+        description,
+        ResolutionPreset.low,
+      );
+
+      controller.dispose();
+
+      try {
```
bparrishMines, line 19: I believe [returnsNormally](https://pub.dev/documentation/matcher/latest/matcher/returnsNormally-constant.html) is the standard way to make sure no Exception is thrown. 

```dart
expect(() => controller.debugCheckIsDisposed(), returnsNormally);
```

bparrishMines (commented):
Everything looks good! Just one small comment on the test.

bparrishMines (approved):
LGTM

I can merge for you once tests pass again.

mvanbeusekom:
Noticed that `flutter format` failed. I pushed a fix and will monitor the CI and merge when it is all green. Thank you for reviewing so quickly!

PR #3206, opened 2020-10-26
[google_maps_flutter] Clean up google_maps_flutter plugin

math1man:
## Description

Clean up google_maps_flutter plugin. This change should not have any functional differences.

1. A few minor formatting changes and additions of @Nullable annotations
2. Removed pass-through of activityHashCode. In the legacy plugin use case, the value is derived from `registrar.getActivity()`. In the new plugin use case, the value should never be used because it uses DefaultLifecycleCallbacks instead of ActivityLifecycleCallbacks.
3. Replaced custom lifecycle state ints with androidx.lifecycle.Lifecycle.State enum. Also simplify paused/stopped states, which don't need their own dedicated states. Paused == started and stopped == created.
4. Fixed a bug where the Lifecycle object was being leaked onDetachFromActivity by nulling out the field.
5. Moved GoogleMapListener to its own file. Declaring multiple top level classes in the same file is bad practice.

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapListener.java
```diff
@@ -0,0 +1,16 @@
+package io.flutter.plugins.googlemaps;
+
+import com.google.android.gms.maps.GoogleMap;
+
+interface GoogleMapListener
```
amirh, line 5: I'm fine with moving this here if that's how you prefer it, curious though - is there a reference to a style guide / best practices doc with the determination that multiple top level classes is bad?

amirh (changes requested):
Thanks! and thanks for the great PR description, it really helps walking through this PR.

Left a few nits, overall looks good.

One thing to check - in the PR description it says that activityHashCode is `-1` for the legacy plugin, my understanding is that it is `-1` for the v2 (new) embedding, am I correct?

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapBuilder.java
```diff
@@ -7,13 +7,14 @@
 import android.app.Application;
 import android.content.Context;
 import android.graphics.Rect;
+import androidx.annotation.Nullable;
```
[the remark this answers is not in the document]
math1man: Various classes in the plugin already depend on this annotation or the NonNull equivalent, so it appears not.

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapListener.java
math1man: I can't find any formal documentation, but it can cause compiler problems. See discussion in https://stackoverflow.com/q/2336692 for example.

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapListener.java
```diff
@@ -0,0 +1,16 @@
+package io.flutter.plugins.googlemaps;
```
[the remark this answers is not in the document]
math1man: Done, thanks

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -880,7 +856,7 @@ private int getActivityHashCode() {
     if (registrar != null && registrar.activity() != null) {
       return registrar.activity().hashCode();
     } else {
-      return activityHashCode;
+      return -1;
```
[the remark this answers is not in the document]
math1man: Done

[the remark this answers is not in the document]
math1man: Though it's worth noting that this could be considered a breaking change.

math1man:
> One thing to check - in the PR description it says that activityHashCode is `-1` for the legacy plugin, my understanding is that it is `-1` for the v2 (new) embedding, am I correct?

I updated the description to clarify that the value is derived from `registrar.getActivity()`. What I had been referring to was the legacy `new GoogleMapFactory` call site in `GoogleMapsPlugin#registerWith` passed in a value of `-1`.

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -552,18 +524,22 @@ private void setGoogleMapListener(@Nullable GoogleMapListener listener) {
   }
 
   // @Override
-  // The minimum supported version of Flutter doesn't have this method on the PlatformView interface, but the maximum
-  // does. This will override it when available even with the annotation commented out.
+  // The minimum supported version of Flutter doesn't have this method on the PlatformView
```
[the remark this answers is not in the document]
math1man: Done

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapBuilder.java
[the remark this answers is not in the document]
amirh: Probably wrong that we didn't have it so far, I think these get brought in by the Android embedding right now, though for consistency with other dependencies and for future compatibility (in the very unlikely event that the embedding is going to drop the dependency) I'd be explicit about the dependencies.

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -880,7 +852,11 @@ private int getActivityHashCode() {
     if (registrar != null && registrar.activity() != null) {
```
amirh, line 852: Mmmm staring at this snippet a little more I'm worried this may crash with the old embedding when the activity is in the background and some lifecycle event for the same application happens.
We used to cache the activityHashCode, and we stopped caching at https://github.com/flutter/plugins/pull/2488 cc @cyanglaz do you remember the context for this change? was it safe to stop caching the activity hashCode in that PR?

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.4
+
+* Minor cleanup of Android code.
```
amirh, line 3: nit: call out the specific things that were done (especially the leak fix).

amirh (commented):
Thanks! this looks good, I'm waiting to hear from @cyanglaz regarding the activityHashCode before landing to make sure we're not introducing a crash with the activityHashCode.

tarektaamali:
please look at this [issue](https://github.com/flutter/flutter/issues/40284) has opened more than a year ago  .you should fixed .it's happened with all version of the packages

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
math1man: Done

math1man:
> please look at this [issue](https://github.com/flutter/flutter/issues/40284) has opened more than a year ago .you should fixed .it's happened with all version of the packages

Thanks for the pointer. I don't think this change will address that, but I have some theories on how it might be fixed. I'll look into it in another pull request.

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -880,7 +852,11 @@ private int getActivityHashCode() {
...
       return registrar.activity().hashCode();
     } else {
-      return activityHashCode;
+      // This method should only be invoked in cases where registrar != null, and only at lifecycle
```
cyanglaz, line 855: This exception can actually throw when some lifecycle event happened when the activity is in the background as mentioned in https://github.com/flutter/plugins/pull/3206/files#r512392320
It is a regression caused by https://github.com/flutter/plugins/pull/2488. We should fix it by removing the `getActivityHashCode()` and use a cached hashCode value like how it was.
I created an issue to track this fix: https://github.com/flutter/flutter/issues/69128

@math1man Do you mind reverting it back to return -1 here and add a TODO to fix the regression?
Something like:
```
// TODO(cyanglaz): Remove the `getActivityHashCode()` and use a cached hashCode when creating the web view for V1 embedding.
// https://github.com/flutter/flutter/issues/69128
return activityHashCode;
```

math1man: Done, thanks for creating that.

amirh (approved):
LGTM

Thanks!
Will land on green.

PR #3210, opened 2020-10-27
[share] Fix bug on iPad where `origin` is null and enable XCUITests in the repo

cyanglaz:
xctests enabled xctests in repo, also added a sample test in the share plugin.


jmagman (approved):
LGTM based on https://github.com/flutter/plugins/pull/3188

PR #3213, opened 2020-10-27
[google_maps_flutter] Overhaul lifecycle management in GoogleMapsPlugin

math1man:
## Description

Overhaul lifecycle management in GoogleMapsPlugin.

GoogleMapController is now uniformly driven by implementing DefaultLifecycleObserver. That observer is registered to a lifecycle passed via the new LifecycleProvider interface, which has 3 implementations:

1. For v2 plugin registration, GoogleMapsPlugin implements the interface and holds the lifecycle in a field, which is itself controlled by ActivityAware methods.
2. For v1 plugin registration, if the activity implements LifecycleOwner, it's lifecycle is used directly.
3. For v1 plugin registration, if the activity does not implement LifecycleOwner, a proxy lifecycle is created via ActivityLifecycleCallbacks.

I'm suspicious that there may still be a bug where a GoogleMapController instance can be shared across multiple activities and therefore leak the activity and lifecycle.

## Related Issues

* Fixes flutter/flutter#69128
* Works towards addressing flutter/flutter#40284

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 1.1.0
+
+Overhaul lifecycle management in GoogleMapsPlugin.
+
+GoogleMapController is now uniformly driven by implementing `DefaultLifecycleObserver`. That observer is registered to a lifecycle from one of three sources:
+
+1. For v2 plugin registration, `GoogleMapsPlugin` obtains the lifecycle via `ActivityAware` methods.
+2. For v1 plugin registration, if the activity implements `LifecycleOwner`, it's lifecycle is used directly.
+3. For v1 plugin registration, if the activity does not implement `LifecycleOwner`, a proxy lifecycle is created and driven via `ActivityLifecycleCallbacks`.
+
 ## 1.0.4
 
 * Cleanup of Android code:
--- a/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.1.0
+## 1.0.5
 
 Overhaul lifecycle management in GoogleMapsPlugin.
 
--- a/packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
+++ b/packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
@@ -6,8 +6,6 @@ package io.flutter.plugins.googlemaps;
 
 import android.Manifest;
 import android.annotation.SuppressLint;
-import android.app.Activity;
-import android.app.Application;
 import android.content.Context;
 import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
@@ -19,7 +17,6 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.lifecycle.DefaultLifecycleObserver;
 import androidx.lifecycle.Lifecycle;
-import androidx.lifecycle.Lifecycle.State;
 import androidx.lifecycle.LifecycleOwner;
 import com.google.android.gms.maps.CameraUpdate;
 import com.google.android.gms.maps.GoogleMap;
@@ -39,7 +36,6 @@ import io.flutter.embedding.engine.plugins.activity.ActivityPluginBinding;
 import io.flutter.plugin.common.BinaryMessenger;
 import io.flutter.plugin.common.MethodCall;
 import io.flutter.plugin.common.MethodChannel;
-import io.flutter.plugin.common.PluginRegistry;
 import io.flutter.plugin.platform.PlatformView;
 import java.io.ByteArrayOutputStream;
 import java.util.ArrayList;
@@ -50,8 +46,7 @@ import java.util.Map;
 
 /** Controller of a single GoogleMaps MapView instance. */
 final class GoogleMapController
-    implements Application.ActivityLifecycleCallbacks,
-        DefaultLifecycleObserver,
+    implements DefaultLifecycleObserver,
         ActivityPluginBinding.OnSaveInstanceStateListener,
         GoogleMapOptionsSink,
         MethodChannel.MethodCallHandler,
@@ -75,12 +70,8 @@ final class GoogleMapController
   private boolean disposed = false;
   private final float density;
   private MethodChannel.Result mapReadyResult;
-  @Nullable private final Lifecycle lifecycle;
   private final Context context;
-  // Do not use directly, use getApplication() instead to get correct application object for both v1
-  // and v2 embedding.
-  @Nullable private final Application mApplication;
-  @Nullable private final PluginRegistry.Registrar registrar; // For v1 embedding only.
+  private final LifecycleProvider lifecycleProvider;
   private final MarkersController markersController;
   private final PolygonsController polygonsController;
   private final PolylinesController polylinesController;
@@ -94,9 +85,7 @@ final class GoogleMapController
       int id,
       Context context,
       BinaryMessenger binaryMessenger,
-      @Nullable Application application,
-      @Nullable Lifecycle lifecycle,
-      @Nullable PluginRegistry.Registrar registrar,
+      LifecycleProvider lifecycleProvider,
       GoogleMapOptions options) {
     this.id = id;
     this.context = context;
@@ -105,9 +94,7 @@ final class GoogleMapController
     this.density = context.getResources().getDisplayMetrics().density;
     methodChannel = new MethodChannel(binaryMessenger, "plugins.flutter.io/google_maps_" + id);
     methodChannel.setMethodCallHandler(this);
-    mApplication = application;
-    this.lifecycle = lifecycle;
-    this.registrar = registrar;
+    this.lifecycleProvider = lifecycleProvider;
     this.markersController = new MarkersController(methodChannel);
     this.polygonsController = new PolygonsController(methodChannel, density);
     this.polylinesController = new PolylinesController(methodChannel, density);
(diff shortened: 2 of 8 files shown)


packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 1.1.0
```
amirh, line 1: This doesn't introduce new features so should be a patch bump (1.0.5)

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
```diff
@@ -4,196 +4,172 @@
...
-  }
-
   // FlutterPlugin
 
   @Override
   public void onAttachedToEngine(FlutterPluginBinding binding) {
-    pluginBinding = binding;
+    binding
+        .getPlatformViewRegistry()
+        .registerViewFactory(VIEW_TYPE, new GoogleMapFactory(binding.getBinaryMessenger(), this));
```
amirh, line 73: What do you think about creating a separate `LifecycleProvider` object which is just the lifecycle provider and passing that here (instead of having GoogleMapsPlugin implement LifecycleProvider). My worry is that future developers adding members to GoogleMapsPlugins will miss the fact that a reference to it is being retained by GoogleMapFactory...

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -119,30 +106,8 @@ public View getView() {
...
-        // Nothing to do, the activity has been completely destroyed or not yet created.
-        break;
-    }
-    if (lifecycle != null) {
-      lifecycle.addObserver(this);
-    } else {
-      getApplication().registerActivityLifecycleCallbacks(this);
-    }
+  void init() {
+    lifecycleProvider.getLifecycle().addObserver(this);
```
amirh, line 110: Not sure so double checking - will we get the initial notification for the current state?

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -507,7 +472,10 @@ public void dispose() {
     methodChannel.setMethodCallHandler(null);
     setGoogleMapListener(null);
     destroyMapViewIfNecessary();
-    getApplication().unregisterActivityLifecycleCallbacks(this);
+    Lifecycle lifecycle = lifecycleProvider.getLifecycle();
+    if (lifecycle != null) {
+      lifecycle.removeObserver(this);
```
amirh, line 477: (not sure) do we need to protect against calling `removeObserver` twice? (here and from onDestroy)

packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: google_maps_flutter
 description: A Flutter plugin for integrating Google Maps in iOS and Android applications.
 homepage: https://github.com/flutter/plugins/tree/master/packages/google_maps_flutter/google_maps_flutter
-version: 1.0.4
+version: 1.1.0
```
amirh, line 4: don't forget to update the version here as well

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
```diff
@@ -4,196 +4,172 @@
...
-    state.set(STARTED);
+  public Lifecycle getLifecycle() {
+    return lifecycle;
   }
 
-  @Override
-  public void onStop(@NonNull LifecycleOwner owner) {
-    state.set(CREATED);
-  }
+  private static final class ProxyLifecycleProvider
```
amirh, line 107: nit: Can you add some comments here describing that this is used make a lifecycle our of the application lifecycle calls?

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
```diff
@@ -4,196 +4,172 @@
...
+                  registrar.messenger(),
+                  new LifecycleProvider() {
+                    @Override
+                    public Lifecycle getLifecycle() {
+                      return ((LifecycleOwner) activity).getLifecycle();
+                    }
+                  }));
+    } else {
+      ProxyLifecycleProvider proxyLifecycleProvider = new ProxyLifecycleProvider(activity);
+      activity.getApplication().registerActivityLifecycleCallbacks(proxyLifecycleProvider);
```
amirh, line 57: optional nit: I'd consider moving this into `ProxyLifecycleProvider` to keep all the hackery in one class.

amirh (changes requested):
This is great! thanks!

Left a few minor comments.

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
math1man: Done

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
math1man: Yes, from `addObserver`'s documentation:

> The given observer will be brought to the current state of the LifecycleOwner. For example, if the LifecycleOwner is in `State#STARTED` state, the given observer will receive `Event#ON_CREATE`, `Event#ON_START` events

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
math1man: Nope, operation is idempotent

packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
math1man: done

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
math1man: Done

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
math1man: Done

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
math1man: Makes sense, done

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
```diff
@@ -4,196 +4,182 @@
...
   @Override
   public void onAttachedToEngine(FlutterPluginBinding binding) {
-    pluginBinding = binding;
+    binding
+        .getPlatformViewRegistry()
+        .registerViewFactory(
+            VIEW_TYPE,
+            new GoogleMapFactory(
+                binding.getBinaryMessenger(),
+                new LifecycleProvider() {
```
amirh, line 76: Is this going to keep an implicit reference to the outer class?

math1man: It actually keeps an explicit reference to the outer class, because `lifecycle` is a field in the outer class. I don't think that's avoidable because only the plugin itself can implement ActivityAware.

amirh: I'm probably missing something obvious... what fails if:

GoogleMapsPlugin holds a reference to the provider (an implementation with a lifecycle field) and sets the lifecycle when attached/detached to the activity, and we pass that provider reference to GoogleMapFactory, won't we avoid a reference from GoogleMapFactory to GoogleMapsPlugin in this case?

math1man: I could do that, but it doesn't really matter and ends up being more boiler plate. My assumption is that the scope of GoogleMapsPlugin is greater than the scope of GoogleMapFactory, so it doesn't matter if GoogleMapFactory transitively captures the GooglemapsPlugin instance. If that assumption is violated, all sorts of worse things happen than this particular leak.

I guess my question is what problem are you trying to solve?

amirh (approved):
LGTM

amirh:
"Windows Plugins" was fixed and the re-run is green (the presubmit indicator is stale).

slightfoot:
@math1man Is there a reason you didn't use this? https://github.com/flutter/plugins/tree/master/packages/flutter_plugin_android_lifecycle

math1man:
> @math1man Is there a reason you didn't use this? https://github.com/flutter/plugins/tree/master/packages/flutter_plugin_android_lifecycle

I do exactly that [here](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java). That's not the problem AFAICT.

slightfoot:
> > @math1man Is there a reason you didn't use this? https://github.com/flutter/plugins/tree/master/packages/flutter_plugin_android_lifecycle
> 
> I do exactly that [here](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java).

My point was your suppose to import that package and use its adapter.

math1man:
Lines 90-93 of GoogleMapsPlugin are this:
```
  @Override
  public void onAttachedToActivity(ActivityPluginBinding binding) {
    lifecycle = FlutterLifecycleAdapter.getActivityLifecycle(binding);
  }
```
Could you elaborate on what you mean?

PR #3230, opened 2020-10-29
[google_maps_flutter_web][url_launcher_web][video_player_web] Remove custom analysis options

amirh:
Removes the custom analysis options files for google_maps_flutter_web, url_launcher_web, and video_player_web, I added local suppresses to the undefined_prefixed_name warnings.

This is done to enable https://github.com/flutter/flutter/issues/69309


ditman:
Local suppresses cannot be done to disable certain analysis checks, hence why they were added as `analysis_options.yaml` files.

See [this](https://github.com/flutter/plugins/pull/2658).

amirh:
Strangely enough analysis passes cleanly for me locally with the inline ignores... not sure what's the difference on CI

ditman:
@amirh maybe you have an old-ish version of the [`analyzer` package](https://pub.dev/packages/analyzer)?

amirh:
Seems like I had some locally cached state that prevented the analyzer from complaining.

PR #3232, opened 2020-10-29
[google_maps_flutter] Fix state-saving logic in google_maps_flutter

math1man:
## Description

GoogleMapController previously implemented `ActivityPluginBinding.OnSaveInstanceStateListener` but did not get registered as a listener. This meant that state-saving simply didn't work.

The new system powers state saving with a combination of [androidx.savedstate.SavedStateRegistry](https://developer.android.com/reference/androidx/savedstate/SavedStateRegistry) and `ActivityPluginBinding.OnSaveInstanceStateListener`. Unfortunately, due to bugs and shortcomings in the Flutter core embedding library, this required a lot of hackery to get right. I've thoroughly documented the hackery in code and described it at a high level in the changelog file.

## Related Issues

I believe this fixes flutter/flutter#40284

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -512,7 +516,38 @@ public void onCreate(@NonNull LifecycleOwner owner) {
     if (disposed) {
       return;
     }
-    mapView.onCreate(null);
+    if (owner instanceof SavedStateRegistryOwner) {
+      // Accessing SavedStateRegistry *must* be done in DefaultLifecycleObserver#onCreate for
+      // reasons described in the javadoc of GoogleMapsPlugin.StateSavingLifecycleProvider.
+      SavedStateRegistry savedStateRegistry =
+          ((SavedStateRegistryOwner) owner).getSavedStateRegistry();
+      String key = SAVED_STATE_KEY + id;
```
amirh, line 524: Mmmm these IDs are assigned in order of platform view creation.
To my understanding: if the app is coming back to a restored state the key may match a different view. e.g:

Let's say we have an app with 2 pages (pageA, pageB), and a map on each page (mapA, mapB). The app starts at pageA, and can then transition to pageB.

The ID for mapA would be 1 and form mapB will be 2.
Let's say the activity is cleared from memory when in mapB, when restoring the instance state - if the app is restoring back to pageB, now mapB will have id 1 and the wrong state will be restored.

I'm not exactly sure what will be the best approach here, I guess we'd need to use some sort of persistent widget identifier defined by the app developer?

cc @goderbauer (master of state restorations), is what I'm describing correct? assuming all the widgets in the app are restorable if we're coming back directly to pageB the widgets of pageA won't be created first correct? any advice here?

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -18,6 +18,9 @@
 import androidx.lifecycle.DefaultLifecycleObserver;
 import androidx.lifecycle.Lifecycle;
 import androidx.lifecycle.LifecycleOwner;
+import androidx.savedstate.SavedStateRegistry;
```
amirh, line 21: We need to add the savedstate dependency to the gradle file. IIUC it happens that we add the embedding dependencies as app-level dependencies which is why this currently build, but we actually don't want to allow apps to rely on the transitive dependencies (cc @blasten who I believe have some issue tracking this)

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
```diff
@@ -181,5 +194,88 @@ public void onActivityDestroyed(Activity activity) {
...
+
+    @NonNull
+    @Override
+    public SavedStateRegistry getSavedStateRegistry() {
+      return savedStateRegistryController.getSavedStateRegistry();
+    }
+  }
+
+  /**
+   * This logic is incredibly hacky, but is necessary to address multiple problems with the Flutter
```
amirh, line 206: nit: please start with a summary of what it does, the disclosure can come second 😄

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
```diff
@@ -181,5 +194,88 @@ public void onActivityDestroyed(Activity activity) {
...
+   * SavedStateRegistryOwner}. This would make this class completely unnecessary. However, {@link
+   * io.flutter.embedding.android.FlutterActivity} does not (at the time of this writing) do this.
+   *
+   * <p>Given that, ideally, every LifecycleOwner that drives this plugin would restore state at the
+   * correct time. This would allow us to consistently drive state restoration via {@link
+   * OnSaveInstanceStateListener} rather than possibly needing to delegate to the LifecycleOwner.
+   * Unfortunately, {@link io.flutter.embedding.android.FlutterFragment} (at the time of this
+   * writing) restores state in {@link androidx.fragment.app.Fragment#onActivityCreated(Bundle)}
+   * rather than in {@link androidx.fragment.app.Fragment#onCreate(Bundle)}. Note that this is a
+   * pretty serious bug in FlutterFragment.
```
amirh, line 219: @xster FYI

amirh (commented):
Sorry for the delayed review!

Thanks for digging deep pointing out embedding issues and coming up with workarounds!

I left a few notes, additionally we will need to test cover this before we can land. Seems like driving the activity lifecycle to trigger state restoration can be done with an Android instrumentation test the tricky part would be what to assert, I wish we could hack an Android instrumentation test that will use the method channel to check some internal map state. I wonder if the integration_test package can get us close, @dnfield do you have any testing advice here?
I guess an alternative would be to use a Robolectric style test which is also not completely trivial to pull off.

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
math1man: I've got a separate PR pending to fix this and a few other lifecycle timing bugs: https://github.com/flutter/engine/pull/22203

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
+## 1.0.6
+
+Fix state-saving logic in GoogleMapsPlugin.
+
+GoogleMapController previously implemented `ActivityPluginBinding.OnSaveInstanceStateListener`
```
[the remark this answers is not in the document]
math1man: Let me know if the update is sufficient.

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
math1man: Ah that's a great point, I didn't realize that's how the IDs work. That would certainly create a wrinkle in this design. I'll wait for a reply from @goderbauer for further guidance.

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
math1man: Done

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapsPlugin.java
math1man: Done

packages/google_maps_flutter/google_maps_flutter/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
goderbauer: There's no garantee that the widgets will be created in the same order again. In the state restoration framework for the flutter framework we require developers to assign stable `restorationId`s to widgets or properties that they need to restore. The ID needs to be stable across app launches.

math1man: Is there any way for the PlatformView to get access to that ID? If multiple maps are used, state saving needs a unique stable ID for each map.

goderbauer: The widget in question would have to take a restoration ID as an argument and send it over to this size here.

math1man: Got it. Since a situation with 2 maps should be rare, I can definitely add that as an optional argument and avoid breaking API changes.

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -57,6 +61,13 @@ class GoogleMap extends StatefulWidget {
   })  : assert(initialCameraPosition != null),
         super(key: key);
 
+  /// Restoration ID to save and restore state of the native widget.
+  ///
+  /// If multiple [GoogleMap] widgets are included in the same layout, you must
+  /// specify a unique value for each widget in order to ensure native state is
+  /// restored correctly. Otherwise, the argument is optional.
```
goderbauer, line 68: In the framework we usually only enable state restoration if an ID has been provided.

math1man: I could go that route, but the native Android component generally assumes native state is being saved correctly so I'd prefer to save state proactively.

PR #3235, opened 2020-10-30
[connectivity] Fix IllegalArgumentException

hamdikahloun:
## Description

* Android: Fix IllegalArgumentException.
* Android: Update Example project.

## Related Issues
https://github.com/flutter/flutter/issues/68265

## Checklist

- [x] I signed the [CLA].

## Breaking Change


packages/connectivity/connectivity/android/src/main/java/io/flutter/plugins/connectivity/ConnectivityBroadcastReceiver.java
```diff
@@ -30,6 +30,7 @@
   private Connectivity connectivity;
   private EventChannel.EventSink events;
   private Handler mainHandler = new Handler(Looper.getMainLooper());
+  private ConnectivityManager.NetworkCallback networkCallback;
```
dnfield, line 33: Why are we caching this?

And it looks like we're only setting it in `onCancel`, is that really right?

hamdikahloun: Ah sorry it's inversed .

dnfield (commented):
So connectivity expects us to give it the same exact object for the callback when registering/unregistering?

Given that, is there some way we can make this a little easier to think about? It seems like it would be easy to mess this up by failing to use the private var vs. calling the method.

I'm also concerned there's a bug now where we'd potentially call `context.unregisterReciever` when we don't want to.

packages/connectivity/connectivity/android/src/main/java/io/flutter/plugins/connectivity/ConnectivityBroadcastReceiver.java
```diff
@@ -49,8 +50,8 @@ public void onListen(Object arguments, EventChannel.EventSink events) {
 
   @Override
   public void onCancel(Object arguments) {
-    if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
-      connectivity.getConnectivityManager().unregisterNetworkCallback(getNetworkCallback());
+    if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && networkCallback != null) {
+      connectivity.getConnectivityManager().unregisterNetworkCallback(networkCallback);
     } else {
       context.unregisterReceiver(this);
```
dnfield, line 56: We can now fall into this branch where we didn't before. That seems like a bug.

dnfield (commented):
Also, we should have a test for this.

packages/connectivity/connectivity/android/src/main/java/io/flutter/plugins/connectivity/ConnectivityBroadcastReceiver.java
hamdikahloun: And I can't implement this interface because require API level 21 : [ConnectivityManager.NetworkCallback](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback)
> So connectivity expects us to give it the same exact object for the callback when registering/unregistering?

dnfield: Right - it seems like if that's the case, we should just be doing nothing rather than calling `context.unregisterReceiver`

packages/connectivity/connectivity/android/src/main/java/io/flutter/plugins/connectivity/ConnectivityBroadcastReceiver.java
```diff
@@ -50,7 +62,7 @@ public void onListen(Object arguments, EventChannel.EventSink events) {
   @Override
   public void onCancel(Object arguments) {
     if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
-      connectivity.getConnectivityManager().unregisterNetworkCallback(getNetworkCallback());
+      connectivity.getConnectivityManager().unregisterNetworkCallback(networkCallback);
```
dnfield, line 65: Is this safe to call with null?

dnfield (approved):
LGTM

dnfield:
CI isn't happy.

hamdikahloun:
My branch already updated with upstream/master . have you any suggestion please ?

hamdikahloun:
https://github.com/flutter/flutter/pull/69971

hamdikahloun:
CI is happy now

Inf0rmatix:
When will this be released? The Exception is a real issue and we need it gone asap ^^"

hamdikahloun:
Hi @Inf0rmatix 

connectivity 2.0.2 published.

```
2.0.2 :
Android: Fix IllegalArgumentException.
Android: Update Example project.
```

Thank you

PR #3239, opened 2020-10-31
Remove unused `test` dependencies add missing environment constraints

cyanglaz:
## Description

The `test` dependency seems to cause CI to hang, and we are not using them mostly in this repo so we can remove those and make the tree green. This PR only unblocks the tree, it didn't fix the root cause of the issue. See discussion in the issue for details. 

This PR also adds env constraints for some example apps who are missing env constraints. 

## Related Issues

https://github.com/flutter/flutter/issues/69439

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #3236, opened 2020-10-30
Remove upper constraints

mit-mit:
## Description

Align plugins with new templates, removing unused upper constraint

## Related Issues

https://github.com/flutter/flutter/pull/68437

PR #3240, opened 2020-10-31
Add Dart SDK constraint to example pubspecs that were missing them

zanderso:
If a pubspec.yaml for an example was missing a Dart SDK constraint, this PR copies the `environment` section from the `pubspec.yaml` file in the parent directory.

### DIFF
--- a/packages/android_alarm_manager/example/pubspec.yaml
+++ b/packages/android_alarm_manager/example/pubspec.yaml
@@ -21,3 +21,7 @@ dev_dependencies:
 
 flutter:
   uses-material-design: true
+
+environment:
+  sdk: ">=2.1.0 <3.0.0"
+  flutter: ">=1.12.13+hotfix.5 <2.0.0"
--- a/packages/android_alarm_manager/CHANGELOG.md
+++ b/packages/android_alarm_manager/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.5+17
+
+* Update Dart SDK constraint in example.
+
 ## 0.4.5+16
 
 * Remove unnecessary workaround from test.
--- a/packages/android_alarm_manager/pubspec.yaml
+++ b/packages/android_alarm_manager/pubspec.yaml
@@ -4,7 +4,7 @@ description: Flutter plugin for accessing the Android AlarmManager service, and
 # 0.4.y+z is compatible with 1.0.0, if you land a breaking change bump
 # the version to 2.0.0.
 # See more details: https://github.com/flutter/flutter/wiki/Package-migration-to-1.0.0
-version: 0.4.5+16
+version: 0.4.5+17
 homepage: https://github.com/flutter/plugins/tree/master/packages/android_alarm_manager
 
 dependencies:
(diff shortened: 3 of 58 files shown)


leafpetersen (approved):

amirh:
We should probably re-publish all these plugins (to make sure the pub.dev score won't get hit by analysis errors) so we need to bump the version numbers and add changelog entries.

Side note: it is worrying that so many packages had this problem and were affected by https://github.com/flutter/flutter/pull/69440#issuecomment-719887060, I'm wondering what's the status for all the existing packages in the wild (is it possible that we had a time range during which the plugin template didn't include these constraints?)

leafpetersen:
> Side note: it is worrying that so many packages had this problem and were affected by [flutter/flutter#69440 (comment)](https://github.com/flutter/flutter/pull/69440#issuecomment-719887060), I'm wondering what's the status for all the existing packages in the wild (is it possible that we had a time range during which the plugin template didn't include these constraints?)

I think @mit-mit would know about the template, and might have run the numbers for missing sdk (or at least flutter) constraints once?

zanderso:
> we need to bump the version numbers and add changelog entries.

Done

amirh (approved):
LGTM

packages/android_alarm_manager/example/pubspec.yaml
```diff
@@ -21,3 +21,7 @@ dev_dependencies:
 
 flutter:
   uses-material-design: true
+
+environment:
+  sdk: ">=2.1.0 <3.0.0"
+  flutter: ">=1.12.13+hotfix.5 <2.0.0"
```
mit-mit, line 27: For the Flutter SDK constraint, please add the lower constraint only. See https://github.com/flutter/plugins/pull/3236 and https://github.com/flutter/flutter/pull/68437

mit-mit (changes requested):

packages/android_alarm_manager/example/pubspec.yaml
zanderso: The goal of this PR is to get flutter/plugins CI green with tip-of-tree framework. Other cleanup work can happen in a subsequent PR.

mit-mit: Why would we add a constraint in this PR, and then remove it again later? I don't understand the value in incurring additional debt. The `flutter` constraint does nothing wrt. getting CI green; you just need the Dart constraint for that.

zanderso: This PR is consistent with the rest of the repo. Fixing every pubspec in the repo is out of scope for this PR. Choosing the right time to do other cleanups is up to @amirh to decide.

zanderso:
@amirh the Cirrus presubmits are timing out. Do you know what's up with that?

amirh:
> @amirh the Cirrus presubmits are timing out. Do you know what's up with that?

Seems to be https://github.com/flutter/flutter/issues/69439 😞

amirh:
This has been incorporated into https://github.com/flutter/plugins/pull/3239 which also works around the driver timeout.

zanderso:
Closing in favor of https://github.com/flutter/plugins/pull/3239

PR #3247, opened 2020-11-03
[integration_test] Remove test dependency and related tests.

cyanglaz:


amirh (commented):
Maybe comment out the test instead of removing with a TODO pointing to https://github.com/flutter/flutter/issues/69439 ?

Also - can you update the PR title?

cyanglaz:
@amirh done. I'm not sure if CI would complain an empty test file, I would imaging if the file exists, `flutter drive` would run against it. Hopefully, it can exit gracefully if it detects the file is empty.

amirh (approved):
LGTM

PR #3261, opened 2020-11-10
[file_selector] Allow empty type groups

stuartmorgan:
## Description

This was restricted under the assumption that there was no reason for an
empty group to exist, but that's not actually true; on platforms with
selectable groups in the native UI, it may be useful to provide
specific options as well as a fallback option (e.g., "Text files" and
"All files").

## Related Issues

None

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
+++ b/packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
@@ -5,18 +5,16 @@
 /// A set of allowed XTypes
 class XTypeGroup {
   /// Creates a new group with the given label and file extensions.
+  ///
+  /// A group with none of the type options provided indicates that any type is
+  /// allowed.
   XTypeGroup({
     this.label,
     this.extensions,
     this.mimeTypes,
     this.macUTIs,
     this.webWildCards,
-  }) : assert(
-            !((extensions == null || extensions.isEmpty) &&
-                (mimeTypes == null || mimeTypes.isEmpty) &&
-                (macUTIs == null || macUTIs.isEmpty) &&
-                (webWildCards == null || webWildCards.isEmpty)),
-            "At least one type must be provided for an XTypeGroup.");
+  });
 
   /// The 'name' or reference to this group of types
   final String label;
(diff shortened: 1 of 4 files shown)


packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
```diff
@@ -5,18 +5,16 @@
 /// A set of allowed XTypes
 class XTypeGroup {
   /// Creates a new group with the given label and file extensions.
+  ///
+  /// A group with none of the type options provided indicates that any type is
+  /// allowed.
```
ditman, line 10: It seems there's a valid UTI (`public.item`/`public.content`), MIME-type (`*/*`) and maybe extension (`*`?) to specify "any". Do we want to remove these asserts?

If this is for developer convenience, should we provide a pre-configured static XTypeGroup instance (`XTypeGroup.any`?) with the above preconfigured?

ditman (approved):
I'm OK with this change if my suggestion above is not a good way to go!

packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
ditman: (`any` would probably need to be a named constructor of XType so we can still override the `label` in a i18n-able way:

```
XTypeGroup.any(String label)
```

stuartmorgan: > It seems there's a valid UTI (`public.item`/`public.content`), MIME-type (`*/*`) and maybe extension (`*`?) to specify "any". Do we want to remove these asserts?

I don't think there's any reason to make it more complex to specify a type group that doesn't restrict anything. I was the one who suggested this assert IIRC, and it was purely because I didn't think there was a valid use case for having an empty type group, so I thought we should flag it as a mistake. But there is, so why not allow it in the simplest possible way (null, which we already have to allow for each of these properties anyway)?

> If this is for developer convenience, should we provide a pre-configured static XTypeGroup instance (`XTypeGroup.any`?) with the above preconfigured?

We could, but doubling the constructor surface to provide a secondary syntax for a rare case seems unnecessary.

ditman: 🚀 then!

PR #3268, opened 2020-11-15
[image_picker_for_web] update versions in readme

otopba:


ditman:
@otopba, is this PR motivated for cleanliness of the README, or is this solving some issue that happened because of having outdated versions?

According to Dart [documentation](https://dart.dev/tools/pub/dependencies#caret-syntax), `image_picker_for_web: ^0.1.0` should automatically pick up `0.1.0+2` as soon as it's available in pub.dev.

otopba:
> @otopba, is this PR motivated for cleanliness of the README, or is this solving some issue that happened because of having outdated versions?
> 
> According to Dart [documentation](https://dart.dev/tools/pub/dependencies#caret-syntax), `image_picker_for_web: ^0.1.0` should automatically pick up `0.1.0+2` as soon as it's available in pub.dev.

I was adding a library to my project and couldn't figure out why MIME Types didn't work. But it turns out that the readme is not the latest version of the library. And by the way, it is already available in the pub

image

ditman:
> the readme is not the latest version of the library. 

@otopba There's no need to specify the latest version of the library when using the caret notation (`^`). pub get knows to find the latest available version of the same major version.

_(It shouldn't be needed neither in the README of the package, nor in your app. If you want your app to ensure to grab the latest packages, you might need to delete your pubspec.lock, though, or check that your package cache is not stale.)_

otopba:
> > the readme is not the latest version of the library. 
> 
> @otopba There's no need to specify the latest version of the library when using the caret notation (`^`). pub get knows to find the latest available version of the same major version.
> 
> > I was adding a library to my project and couldn't figure out why MIME Types didn't work.
> 
> It shouldn't be needed neither in the README of the package, nor in your app. If you want your app to ensure to grab the latest packages, you might need to delete your pubspec.lock, and/or check that your pub package cache is not stale.

Thank you for such a detailed description!

ditman:
Thanks for understanding @otopba!

PR #3279, opened 2020-11-20
Use testWidgets instead of test to fix failures not surfacing on CI

jiahaog:
## Description

Only `testWidgets` is supported for testing for `package:integration_test`. Errors thrown in other declarers will not be caught and reported.

Also added a note to mention this in the README for `package:integration_test`

## Related Issues

flutter/flutter#60943

## Breaking Change

### DIFF
--- a/packages/integration_test/CHANGELOG.md
+++ b/packages/integration_test/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.9.2+3
+## 0.9.3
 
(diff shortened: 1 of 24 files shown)


stuartmorgan (approved):
LGTM

Thanks for investigating! And it's good to know that it was a couple of plugins (including the ones I happened to have tested this on), rather than affecting the entire repo's tests.

PR #3280, opened 2020-11-20
Fix broken link

johnpryan:
This is the correct link: https://github.com/flutter/plugins/blob/master/packages/integration_test/example/test_driver/extended_integration_test.dart


dnfield (approved):
LGTM

PR #3289, opened 2020-11-25
Add chromedriver instructions to example README

johnpryan:
Chromedriver needs to be running in the background to run `flutter drive` with the `web-server` or `chrome` device

## Breaking Change


stuartmorgan:
This copy of `integration_test` is now deprecated in favor of the copy in the Flutter SDK, so this should be updated there instead. Apologies for the delay in triaging this.

PR #3302, opened 2020-12-03
[camera] Add implementations for `camera_platform_interface` package.

BeMacized:
## Description

Updates the dart plugin and its native iOS and Android implementations to make use of the new platform interface that has been submitted in #3253

## Checklist

- [x] I signed the [CLA].

## Breaking Change


mvanbeusekom:
@googlebot I consent.

danielroek:
@googlebot I consent.

ditman:
(I've published camera_platform_interface, sorry for the delay!)

BeMacized:
> (I've published camera_platform_interface, sorry for the delay!)

Nice! The dependency should now be updated as well.

ditman:
Hey @BeMacized @mvanbeusekom, can you edit this PR so I'm allowed to push changes to it (and its branch?) I have a couple of fixes for the issues outlined by the `publishable` check that I'd like to push! I'll also add some extra comments to the code :)

mvanbeusekom:
> Hey @BeMacized @mvanbeusekom, can you edit this PR so I'm allowed to push changes to it (and its branch?) I have a couple of fixes for the issues outlined by the `publishable` check that I'd like to push! I'll also add some extra comments to the code :)

Hi @ditman, I gave you write access to our fork of the plugins repo (https://github.com/baseflow/flutter-plugins) which should give you the required access to push changes to the branch (which will show up in the PR). You should have received an invite by email. Thanks for helping out!

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -0,0 +1,492 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
```
ditman, line 6: If possible, please don't import `dart:io`, otherwise this won't be marked as a "web-compatible" plugin.

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -0,0 +1,492 @@
...
+  ///
+  /// Throws a [CameraException] if image streaming or video recording has
+  /// already started.
+  ///
+  /// The `startImageStream` method is only available on Android and iOS (other
+  /// platforms won't be supported in current setup).
+  ///
+  // TODO(bmparr): Add settings for resolution and fps.
+  Future<void> startImageStream(onLatestImageAvailable onAvailable) async {
+    assert(!Platform.isIOS && !Platform.isAndroid);
```
ditman, line 282: Please, use `defaultTargetPlatform` from Flutter instead of `Platform` from `dart:io` to do these checks (there's one here, and another one in the `stopImageStream` method)

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -0,0 +1,492 @@
...
+
+  /// Releases the resources of this camera.
+  @override
+  Future<void> dispose() async {
+    if (_isDisposed) {
+      return;
+    }
+    _isDisposed = true;
+    super.dispose();
+    if (_creatingCompleter != null) {
```
ditman, line 487: This is the only usage of the _creatingCompleter, which seems to be only checking if the initalize method has been called (but not finished running). I've modified this by a `FutureOr`, it's probable this can be simplified even further...

packages/camera/camera/test/camera_test.dart
```diff
@@ -32,9 +68,443 @@ void main() {
...
+      expect(
+          cameraController.startImageStream((image) => null),
+          throwsA(isA<CameraException>().having(
+            (error) => error.description,
+            'A camera has started streaming images.',
+            'startImageStream was called while a camera was streaming images.',
+          )));
+    });
+
+    test('startImageStream() calls CameraPlatform', () async {
```
ditman, line 308: Would it make sense to split the start/stop ImageStream tests to their own files, so we don't mix the PlatformInterface-based initialization and tests, with the MethodChannel ones?

ditman: (Also for later cleanup, it's easier, just deleting a file)

ditman (commented):
Hey, I've checked this (except for the Java bits, maybe @bparrishMines can take a look at those), and I have a few comments. I don't think there's anything blocking. Some things are not very idiomatic. My only major concerns:

* Reexporting the whole camera_platform_interface from camera. Please reexport only the things that are being removed from Camera. That way we can expose things on platform_interface that are useful to other platform authors, but that shouldn't be directly seen by the users of the plugin.
* Do not import `dart:io`, that way the plugin will be marked as "WEB" compatible in pub.dev later!

ditman:
@googlebot I consent.

mvanbeusekom:
@googlebot I consent.

ditman:
> Hi @ditman, I gave you write access to our fork of the plugins repo

I have added a couple of commits above, I don't think I've broken anything (tests seemed to continue passing), but please verify those! They have to do with the initialization sequence and how it used completers a little bit weirdly.

(Also, when creating PRs, I think it's enough to mark the "allow maintainers to do changes to this PR" checkbox to enable me to push stuff to the branch, without having to give me access to the whole repo (but thanks anyway, Maurits!))

packages/camera/camera/lib/src/camera_controller.dart
mvanbeusekom: Done!

ditman: Thanks!!

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -0,0 +1,492 @@
...
+            .onCameraInitialized(_cameraId)
+            .map((event) {
+              return Size(
+                event.previewWidth,
+                event.previewHeight,
+              );
+            })
+            .first
+            .then((previewSize) => _resolutionCompleter.complete(previewSize)),
+      );
```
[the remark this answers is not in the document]
mvanbeusekom: Yes that looks much better. Thanks

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -0,0 +1,492 @@
...
+
+      value = value.copyWith(
+        isInitialized: true,
+        previewSize: await _resolutionCompleter.future,
+      );
+    } on PlatformException catch (e) {
+      throw CameraException(e.code, e.message);
+    }
+    _creatingCompleter.complete();
+    return _creatingCompleter.future;
```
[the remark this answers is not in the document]
mvanbeusekom: Perfect. I agree this is the better solution

packages/camera/camera/lib/camera.dart
```diff
@@ -2,643 +2,8 @@
...
-      await _creatingCompleter.future;
-      await _channel.invokeMethod<void>(
-        'dispose',
-        <String, dynamic>{'textureId': _textureId},
-      );
-      await _eventSubscription?.cancel();
-    }
-  }
-}
+export 'package:camera_platform_interface/camera_platform_interface.dart';
```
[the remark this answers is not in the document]
mvanbeusekom: I seem to have some problems with this one. Locally I have updated this statement to:

```dart
export 'package:camera_platform_interface/camera_platform_interface.dart'
    show
        CameraDescription,
        CameraException,
        CameraLensDirection,
        ResolutionPreset,
        XFile;
```

But now the code in the example/lib/main.dart won't compile anymore because it can't locate any of the types mentioned by the `show` statement. I made sure I imported the `import package:camera/camera.dart` Dart file but still not working.

I can't figure out what I am doing wrong here, probably overlooking something (it is getting late here). Maybe you have some pointers? Will give it a fresh look tomorrow morning.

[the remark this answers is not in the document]
ditman: It should work, maybe you need a flutter clean in the example app, or to "restart" the analyzer (or both?).

If you want, push whatever you have to the branch and I can take a look on my end (it's going to break some tests, but oh well!)

packages/camera/camera/test/camera_test.dart
mvanbeusekom: I agree, thanks for pointing it out. I moved the tests into a new file called `camera_image_stream_test.dart`.

packages/camera/camera/lib/camera.dart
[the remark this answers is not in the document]
mvanbeusekom: it was indeed my environment that got a bit confused ;) I corrected everything and it works without issues now.

mvanbeusekom:
@ditman I have processed your feedback, would really appreciate it if you can have another look. 

The changes you made look good, I tested the app on several Android and iOS devices and still works nicely.

packages/camera/camera/android/src/test/java/io/flutter/plugins/camera/DartMessengerTest.java
```diff
@@ -54,55 +47,41 @@ public void setUp() {
...
-  public void send_handlesNullEventSinks() {
-    dartMessenger.send(DartMessenger.EventType.ERROR, "error description");
+  public void sendCameraErrorEvent_includesErrorDescriptions() {
+    dartMessenger.sendCameraErrorEvent("error description");
 
     List<ByteBuffer> sentMessages = fakeBinaryMessenger.getMessages();
-    assertEquals(0, sentMessages.size());
+    assertEquals(1, sentMessages.size());
+    MethodCall call = decodeSentMessage(sentMessages.get(0));
+    assertEquals(DartMessenger.EventType.ERROR.toString().toLowerCase(), call.method);
```
bparrishMines, line 56: nit: We typically prefer using literals in tests over constants. It makes it easier to read and self contained. (e.g. "error" instead of `DartMessenger.EventType.ERROR.toString().toLowerCase()`). Up to you though.

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -347,10 +382,7 @@ - (void)captureOutput:(AVCaptureOutput *)output
     }
   }
   if (!CMSampleBufferDataIsReady(sampleBuffer)) {
-    _eventSink(@{
-      @"event" : @"error",
-      @"errorDescription" : @"sample buffer is not ready. Skipping sample"
-    });
+    [_methodChannel invokeMethod:@"error" arguments:@"sample buffer is not ready. Skipping sample"];
```
bparrishMines, line 385: nit: I think `@"error"` would be better as a constant like it is in Java so it doesn't need to be copy and pasted everywhere.

bparrishMines (commented):
Platform code look good to me! I left a couple of small nits. I'll also let @ditman give the approval once he believes it's ready.

packages/camera/camera/ios/Classes/CameraPlugin.m
mvanbeusekom: Agreed, updated the code to use a constant called `errorMethod` and use it where we do a `invokeMethod:@"error"`.

mvanbeusekom:
@ditman, @bparrishMines, I have updated the code to include the feedback from Maurice. @ditman would be great if you can give it another look.

packages/camera/camera/lib/camera.dart
```diff
@@ -6,4 +6,10 @@ export 'src/camera_controller.dart';
...
 export 'src/camera_preview.dart';
 
-export 'package:camera_platform_interface/camera_platform_interface.dart';
+export 'package:camera_platform_interface/camera_platform_interface.dart'
+    show
+        CameraDescription,
+        CameraException,
+        CameraLensDirection,
+        ResolutionPreset,
+        XFile;
```
ditman, line 15: Compare how much cleaner this is now with everything that is exported by the platform_interface:
https://pub.dev/documentation/camera_platform_interface/latest/camera_platform_interface/camera_platform_interface-library.html! Thanks!!

ditman (approved):
I think this is good to go, however the analysis is reporting some formatting errors. Let's merge this after we get a full green CI run!

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -218,12 +219,13 @@ SurfaceTextureEntry getFlutterTexture() {
...
-    final File file = new File(filePath);
-
-    if (file.exists()) {
-      result.error(
-          "fileExists", "File at path '" + filePath + "' already exists. Cannot overwrite.", null);
+  public void takePicture(@NonNull final Result result) {
+    final File outputDir = applicationContext.getCacheDir();
+    final File file;
+    try {
+      file = File.createTempFile("CAP", ".jpg", outputDir);
```
hamdikahloun, line 226: Can we handle  [SecurityException](https://developer.android.com/reference/java/io/File#createTempFile(java.lang.String,%2520java.lang.String)) too ?
I see we have static file name for pictures. We can save one picture?

mvanbeusekom: This is not the case, the `createTempFile` method will create a unique file name. The first parameter ("CAP" in this case) is a prefix that will be used to generate the name (see [File.createTempFile](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#createTempFile(java.lang.String,%20java.lang.String,%20java.io.File) for details).

The actual file will look something like "CAP3602253894598046604.jpg"

mvanbeusekom: Good catch regarding the `SecurityException`, I will add support to handle this as well.

mvanbeusekom: done

hamdikahloun: Great

hamdikahloun (approved):
LGTM

AlexV525:
> * [x]  Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).

I just wondering why only the **MINOR** version has been increased?

mvanbeusekom:
@AlexV525  it actually is a major version number. When a package is not considered stable (version 1.0.0) Dart considers the second digit the major number, the third digit as minor and a fourth digit (added as +n) as the patch number: 

> Although semantic versioning doesn’t promise any compatibility between versions prior to 1.0.0, the Dart community convention is to treat those versions semantically as well. The interpretation of each number is just shifted down one slot: going from 0.1.2 to 0.2.0 indicates a breaking change, going to 0.1.3 indicates a new feature, and going to 0.1.2+1 indicates a change that doesn’t affect the public API. For simplicity’s sake, avoid using + after the version reaches 1.0.0.

The quote above is copied from the article [Package versioning | Dart](https://dart.dev/tools/pub/versioning), in particular the [Semantic versions](https://dart.dev/tools/pub/versioning#semantic-versions) chapter.

AlexV525:
https://github.com/flutter/flutter/wiki/Package-migration-to-1.0.0 This might be against to the migration proposal. The package using `>=0.x.y+z <2.0.0` directly suffered from breaking changes.

mvanbeusekom:
Not really sure how to answer this in a satisfying manor ;). The Flutter plugin teams follows the versioning semantics as discussed in the [Package versioning | Dart](https://dart.dev/tools/pub/versioning) article. As such we bumped the second digit to indicate a major version change.

AlexV525:
Thanks for the response. I think I'll consult others for the conflict later, since they're obviously against each other at present.

PR #3309, opened 2020-12-09
Pass a Uri to package:http APIs

natebosch:
Prepare for https://github.com/dart-lang/http/issues/375

### DIFF
--- a/packages/image_picker/image_picker_platform_interface/CHANGELOG.md
+++ b/packages/image_picker/image_picker_platform_interface/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 1.1.4-dev
+## 1.1.4
+
+* Pass `Uri`s to `package:http` methods, instead of strings, in preparation for a major version update in `http`.
 
 ## 1.1.3
 
--- a/packages/image_picker/image_picker_platform_interface/CHANGELOG.md
+++ b/packages/image_picker/image_picker_platform_interface/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 1.1.4-dev
+
 ## 1.1.3
 
 * Update documentation of `pickImage()` regarding HEIC images.
--- a/packages/image_picker/image_picker_platform_interface/lib/src/types/picked_file/html.dart
+++ b/packages/image_picker/image_picker_platform_interface/lib/src/types/picked_file/html.dart
@@ -24,7 +24,7 @@ class PickedFile extends PickedFileBase {
     if (_initBytes != null) {
       return Future.value(UnmodifiableUint8ListView(_initBytes));
     }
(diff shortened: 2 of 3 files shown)


ditman (commented):
Thanks for the fix @natebosch, before merging we'll have to bump the version in the pubspec + CHANGELOG so this becomes publishable!

natebosch:
I bumped to a `-dev` version so it is clear this has changed since it was last published. https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change

There is no need to publish just for this - it won't have any real impact for places where it is fetched from pub. Where this will make a difference is google3.

ditman:
I need to do a quick real world test with this, I remember Safari having trouble when parsing Blob URLs before doing the ajax fetch. I'll merge this once that's done.

Also, there's a similar package: `cross_file` that might need a similar fix to this (but it's probably not imported into g3), I'll take care of that one!

natebosch:
> I need to do a quick real world test with this, I remember Safari having trouble when parsing Blob URLs before doing the ajax fetch. I'll merge this once that's done.

There should be no impact. That `Uri.parse` call happens today already, it's just happening deep in the `http` functions instead of at the call site.

https://github.com/dart-lang/http/blob/d473635f0dbb72936664696162127c2209986356/lib/src/base_client.dart#L110

> Also, there's a similar package: `cross_file` that might need a similar fix to this (but it's probably not imported into g3), I'll take care of that one!

Thanks!

Note that this won't have an impact on any package until it changes it's dependency constraints in `pubspec.yaml`. This will come with a major version bump in `http`. The main reason I need to do google3 packages in advance is because pubspec constraints don't have any impact there.

natebosch:
It looks like tests are failing due to the `-dev` suffix on the version I mentioned in  https://github.com/flutter/plugins/pull/3309#issuecomment-741341246. This is a pattern that we use to indicate when the code in the repo does not match the code as published under that version, and until we go to do a publish we don't know whether the code will match.

Since this repo doesn't take that approach which is prefered - do you eagerly bump the version when the code changes even without user facing changes, or do you wait to bump the version until there is a user facing change to document in the changelog?

ditman:
> do you eagerly bump the version when the code changes even without user facing changes

@natebosch yes, let's bump the minor version and document in the CHANGELOG that this passes `Uri` objects instead of `Strings` to `package:http` methods. I can do this, now that you've touched the files themselves from the web UI, I think!

Thanks for the clarification about the major version and the links to the http source!

ditman (approved):
LGTM, thanks for the fix @natebosch!

ditman:
@googlebot I consent.

PR #3319, opened 2020-12-10
Update analysis options for nnbd

amirh:
With NNBD `assert(foo != null)` for a non nullable `foo` generates an analysis error.
However as long as we support mixed-mode we want these asserts it, so disabling the check.


blasten (approved):
LGTM

PR #3343, opened 2020-12-17
[webview_flutter] WebView.bouncesEnabled property to controls iOS scrolling

softkot:
## Description

New feature request to control native side  _webView.scrollView.bounces value the same way gestureNavigationEnabled does.

Fixes https://github.com/flutter/flutter/issues/57090


softkot:
No one to review so far ?

stuartmorgan:
Thanks for the submission! We apologize for the long delay in triaging this PR. We’re in the process of overhauling our PR triage system to respond much more quickly, as well as working through the backlog.

We’re currently working through a large backlog of PRs, and this will require non-trivial review, so it will take some time before we’re able to review it. As explained in [CONTRIBUTING.md](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#the-review-process), votes for the corresponding issue are the primary way we’re prioritizing non-trivial reviews, so we encourage you to file an issue and associate it with this PR.

I've also updated the PR description with the checklist we include in our template; please look it over and check off items as you address them; there are several things this PR will need (tests, version update, etc.) before it's ready for review.

packages/webview_flutter/webview_flutter/lib/src/webview.dart
```diff
@@ -83,6 +83,7 @@ class WebView extends StatefulWidget {
     this.javascriptChannels,
     this.navigationDelegate,
     this.gestureRecognizers,
+    this.bouncesEnabled = true,
```
stuartmorgan, line 86: Android has overscroll as well, and API to control it for web views, so this should be given a more generic name.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -432,6 +432,8 @@ private void applySettings(Map<String, Object> settings) {
           break;
         case "gestureNavigationEnabled":
           break;
+        case "bouncesEnabled":
+          break;
```
stuartmorgan, line 436: This should be implemented for Android as well (https://developer.android.com/reference/android/webkit/WebView#setOverScrollMode(int))

stuartmorgan (changes requested):
I've left a few high-level comments. As noted above, this will also need all of the elements in the PR checklist addressed (notably, adding tests).

packages/webview_flutter/webview_flutter/lib/src/webview.dart
softkot: It would be nice if you suggest the name instead of just simple reject it.

stuartmorgan: I didn't reject anything, I just left a review comment.

I would suggest something like `enableOverscroll`, since overscroll is the term Flutter uses for this.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
stuartmorgan: The implementation should be trivial, but per https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#platform-support since we know what the interface should look like it would be fine to land this as an iOS-only change if you aren't comfortable implementing it for Android.

stuartmorgan:
If you'd like to move this PR forward:

> As noted above, this will also need all of the elements in the PR checklist addressed (notably, adding tests).

You've instead deleted the checklist. Please don't do that; it's there for a reason. Please review the checklist, updating the PR as appropriate, and when that’s complete please feel free to mark the PR as ready for review.

Hixie:
@softkot I'm going to close this PR for now since it's missing tests and so is not really ready for review, but please don't hesitate to open a new PR if you have the change to add tests. If you would like help you can join our Discord, see our contributing documentation for details.

PR #3344, opened 2020-12-17
[camera] Fixed stale images in imageStream subscriptions

danielroek:
## Description

When streaming images the first image is the same as the last image of the previous stream of images. This is fixed in this PR

## Related Issues

Redo of #2655 
Fixes [#54837](https://github.com/flutter/flutter/issues/54837)

## Checklist

- [x] I signed the [CLA].

## Breaking Change

-  Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).


hamdikahloun (changes requested):
Thank you for your contribution.
Can you please fix the format exception by removing the extra line?

 ```

-
   private void closeCaptureSession() {
     if (cameraCaptureSession != null) {
       cameraCaptureSession.close();
```

File: packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java [hunk 1]
```diff
@@ -509,6 +509,11 @@ private void setImageStreamImageAvailableListener(final EventChannel.EventSink i
         null);
   }
 
+  public void stopImageStream() throws CameraAccessException {
+    imageStreamReader.setOnImageAvailableListener(null, null);
```
hamdikahloun, line 513: What happens if `imageStreamReader` is null ?

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -470,9 +470,7 @@ public void onListen(Object o, EventChannel.EventSink imageStreamSink) {
           }
 
           @Override
-          public void onCancel(Object o) {
-            imageStreamReader.setOnImageAvailableListener(null, null);
```
hamdikahloun, line 474: Why we should remove this ?

`imageStreamReader.setOnImageAvailableListener(null, null);`

danielroek: Because we call `setOnImageAvailableListener` in `stopImageStream` this would be redundant.

hamdikahloun: i dont think so , because `stopImageStream`  is in `MethodChannel` and you removed `onCancel` from `StreamHandler`

mvanbeusekom: The thing here is that the `CameraController.stopImageStream` executes both the `stopImageStream` method on the `MethodChannel` and will cancel the stream on the `EventChannel`. Making the logic in the `onCancel` method redundant:

```dart
  Future<void> stopImageStream() async {
    assert(defaultTargetPlatform == TargetPlatform.android ||
        defaultTargetPlatform == TargetPlatform.iOS);

   // Left out error handling to keep example compact.

    try {
      value = value.copyWith(isStreamingImages: false);
      await _channel.invokeMethod<void>('stopImageStream');
    } on PlatformException catch (e) {
      throw CameraException(e.code, e.message);
    }

    await _imageStreamSubscription.cancel();
    _imageStreamSubscription = null;
  }
```

mvanbeusekom (approved):
LGTM

hamdikahloun (approved):
LGTM

PR #3338, opened 2020-12-16
[camera] Add implementations for the torch flash mode.

BeMacized:
## Description

Adds implementations for torch mode for Android and iOS. It implements the interfaces defined in #3326.

~~This PR is currently still in draft state, as it still depends on:~~
- ~~#3326 Updated camera_platform_interface being published to pub.dev.~~
- ~~#3314 Flash mode implementations being merged.~~

## Related Issues

flutter/flutter#19845
flutter/flutter#37853

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #3314, opened 2020-12-10
[camera] Flash functionality for Android and iOS

BeMacized:
## Description

Adds flash support to the camera implementations for Android and iOS. It implements the interfaces defined #3313.

~~This PR is currently still in draft state, as it still depends on:~~
~~- #3302 General platform implementations being merged.~~
~~- #3313 Updated `camera_platform_interface` being published to pub.dev.~~

## Related Issues

- flutter/flutter#19845
- flutter/flutter#66252

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #3346, opened 2020-12-17
[camera] Add iOS and Android implementations for managing auto exposure.

BeMacized:
## Description

Adds iOS and Android implementations for managing auto exposure. It implements the interfaces defined in #3345.

It allows for:

- Locking the current exposure settings
- Checking if setting exposure focus points is supported
- Setting the exposure focus point
- Setting the exposure offset/bias/compensation
- Getting the possible range for exposure offset/bias/compensation
- Getting the step size for exposure offset/bias/compensation

~~This PR is currently still in draft state, as it still depends on:~~

- ~~#3345 Auto Exposure platform interface methods being merged and published.~~
- ~~#3314 Flash mode implementations being merged.~~
- ~~#3338 Torch mode implementations being merged.~~
- ~~#3367 Android flash/torch implementation fixes being merged.~~

## Related Issues

- flutter/flutter#41565
- flutter/flutter#24575

## Checklist

- [x] I signed the [CLA].

## Breaking Change


danielroek:
@googlebot I consent.

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -676,8 +681,151 @@ private void updateFlash(FlashMode mode) {
...
+    // Set the metering rectangle
+    aeMeteringRectangle = new MeteringRectangle(targetX, targetY, targetWidth, targetHeight, 1);
+    // Apply it
+    initPreviewCaptureBuilder();
+    this.cameraCaptureSession.setRepeatingRequest(
+        captureRequestBuilder.build(), pictureCaptureCallback, null);
+    result.success(null);
+  }
+
+  @SuppressLint("NewApi")
```
mvanbeusekom, line 736: This could potentially result in errors or crashes on phones using a lower API. The "NewApi" lint guards against usage of APIs that are introduced after the configured `minSdkVersion`. By suppressing these warnings we might introduce crashes on those devices.

It would be better if we can only call these "new" APIs conditionally testing for the installed SDK version and provide alternatives for those SDKs that don't support those APIs. If we do this, there shouldn't be a need to suppress the "NewApi" lint rule.

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -3,12 +3,14 @@
...
 
 import 'dart:async';
+import 'dart:math';
 
 import 'package:camera/camera.dart';
 import 'package:camera_platform_interface/camera_platform_interface.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:pedantic/pedantic.dart';
```
mvanbeusekom, line 13: do we need this dependency? Seems it shouldn't be here.

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -676,8 +681,151 @@ private void updateFlash(FlashMode mode) {
...
+    if (targetX > maxTargetX) targetX = maxTargetX;
+    if (targetY > maxTargetY) targetY = maxTargetY;
+    // Set the metering rectangle
+    aeMeteringRectangle = new MeteringRectangle(targetX, targetY, targetWidth, targetHeight, 1);
+    // Apply it
+    initPreviewCaptureBuilder();
+    this.cameraCaptureSession.setRepeatingRequest(
+        captureRequestBuilder.build(), pictureCaptureCallback, null);
+    result.success(null);
+  }
```
mvanbeusekom, line 734: This could use a unit test, especially all the math. It would also be nice to extract this logic into a separate helper class/ method which will do all required calculations and simply returns the `MeteringRectangle`. This will also simplify unit testing. 

Maybe have a look at the `CameraZoom` class which does something similar.

mvanbeusekom (changes requested):

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
BeMacized: This was added to suppress a warning that came up for a code path that did not exist. 
I split up the method now so it checks more directly and the warning disappeared.

packages/camera/camera/lib/src/camera_controller.dart
BeMacized: Yes, we need it for `unawaited` on line 212. It is [recommended](https://dart-lang.github.io/linter/lints/unawaited_futures.html) to do so by the dart linter.

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
BeMacized: Yep, agreed. That would probably be for the better. Implementing this now.

packages/camera/camera/lib/src/camera_controller.dart
mvanbeusekom: Then we need to change dependency from a `dev_dependency` to a normal `dependency` in the pubspec.yaml

BeMacized: Good catch! Moving it.

mvanbeusekom (approved):
LGTM

mvanbeusekom (approved):

bparrishMines (approved):
LGTM

PR #3379, opened 2020-12-30
Removed mobile platform plugins

tmarzeion:


tmarzeion:
Created that by mistake (wanted to make PR on my fork instead of official one), sorry!

PR #3398, opened 2021-01-07
[package_info] Migrate to null safety

bparrishMines:
Fixes https://github.com/flutter/flutter/issues/75239

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/package_info/example/integration_test/package_info_test.dart
+++ b/packages/package_info/example/integration_test/package_info_test.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// @dart = 2.9
+
 import 'dart:io';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
--- a/packages/package_info/example/integration_test/package_info_test.dart
+++ b/packages/package_info/example/integration_test/package_info_test.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// @dart = 2.9
-
 import 'dart:io';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
--- a/packages/package_info/example/test_driver/integration_test.dart
+++ b/packages/package_info/example/test_driver/integration_test.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// @dart = 2.9
-
 import 'dart:convert';
 import 'dart:io';
 
--- a/packages/package_info/example/test_driver/integration_test.dart
+++ b/packages/package_info/example/test_driver/integration_test.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// @dart = 2.9
+
 import 'dart:convert';
 import 'dart:io';
 
--- a/packages/package_info/lib/package_info.dart
+++ b/packages/package_info/lib/package_info.dart
@@ -35,17 +35,20 @@ class PackageInfo {
   /// Retrieves package information from the platform.
   /// The result is cached.
   static Future<PackageInfo> fromPlatform() async {
-    final PackageInfo? packageInfo = _fromPlatform;
+    PackageInfo? packageInfo = _fromPlatform;
     if (packageInfo != null) return packageInfo;
 
     final Map<String, dynamic> map =
         (await _kChannel.invokeMapMethod<String, dynamic>('getAll'))!;
-    return _fromPlatform = PackageInfo(
+
+    packageInfo = PackageInfo(
       appName: map["appName"],
       packageName: map["packageName"],
       version: map["version"],
       buildNumber: map["buildNumber"],
     );
+    _fromPlatform = packageInfo;
+    return packageInfo;
   }
 
   /// The app name. `CFBundleDisplayName` on iOS, `application/label` on Android.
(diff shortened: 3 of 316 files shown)


packages/package_info/example/integration_test/package_info_test.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// @dart = 2.9
+
 import 'dart:io';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
```
blasten, line 9: @dnfield I couldn't find an issue about migrating `vm_service` to null safe. Do you know if there's any work in the radar?

packages/package_info/example/test_driver/integration_test.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// @dart = 2.9
```
blasten, line 5: Which issue did you see that require opting out?

packages/package_info/lib/package_info.dart
```diff
@@ -24,30 +24,29 @@ class PackageInfo {
...
 
-  static PackageInfo _fromPlatform;
+  static PackageInfo? _fromPlatform;
 
   /// Retrieves package information from the platform.
   /// The result is cached.
   static Future<PackageInfo> fromPlatform() async {
     if (_fromPlatform != null) {
-      return _fromPlatform;
+      return _fromPlatform!;
```
blasten, line 39: shouldn't need `_fromPlatform!` since the analyzer is able to detect the enclosing `_fromPlatform != null` check.

packages/package_info/lib/package_info.dart
```diff
@@ -24,30 +24,29 @@ class PackageInfo {
...
     }
 
-    final Map<String, dynamic> map =
+    final Map<String, dynamic>? map =
         await _kChannel.invokeMapMethod<String, dynamic>('getAll');
-    _fromPlatform = PackageInfo(
-      appName: map["appName"],
+    return _fromPlatform = PackageInfo(
```
blasten, line 44: it's probably a bit more readable to have this as two separate statements

packages/package_info/lib/package_info.dart
```diff
@@ -24,30 +24,29 @@ class PackageInfo {
...
+      appName: map!["appName"],
```
blasten, line 45: why is `map!` required here, but not below?

packages/package_info/lib/package_info.dart
bparrishMines: So, I was wondering about that too. When I remove it, I get this error 

Screen Shot 2021-01-07 at 4 36 31 PM

I also made it synchronous just in case that was the problem. It seems like the static analysis is missing the null check. I'm probably missing something obvious, but have you seen this before?

packages/package_info/lib/package_info.dart
bparrishMines: I'm new to null safety, but it looks like you only have to null check the value once within the same method.

packages/package_info/example/test_driver/integration_test.dart
bparrishMines: There is an issue with flutter_driver being legacy: 

Screen Shot 2021-01-07 at 4 42 03 PM

I wasn't sure when we were planning on migrating test since `integration_test` isn't null safe yet either.

packages/package_info/example/integration_test/package_info_test.dart
dnfield: Looks like per https://github.com/dart-lang/sdk/issues/40046 it's been landed. We can look to migrate integration_test and driver.

blasten: Awesome!

packages/package_info/example/test_driver/integration_test.dart
blasten: ah - you right https://github.com/flutter/flutter/blob/master/packages/flutter_driver/pubspec.yaml#L13

packages/package_info/lib/package_info.dart
blasten: ah yeah. @mehmetf just pointed out that the variable needs to be defined in the local scope

packages/package_info/lib/package_info.dart
blasten: ah, what about:
```dart
final Map<String, dynamic> map = await _kChannel.invokeMapMethod<String, dynamic>('getAll')!;
```

packages/package_info/example/integration_test/package_info_test.dart
dnfield: I was wrong. That's for a different part of the SDK. Migration is still ongoing, Dart cl is https://dart-review.googlesource.com/c/sdk/+/156980

packages/package_info/lib/package_info.dart
bparrishMines: I tried this as well, but that only null checks the `Future` and not the value the `Future` returns. You would have to add parentheses around the `await`. Something like:

```dart
final Map<String, dynamic> map = (await _kChannel.invokeMapMethod<String, dynamic>('getAll'))!;
```

I wasn't really sure if this was cleaner or not.

blasten: correct. 

```dart
final Map<String, dynamic> map = (await _kChannel.invokeMapMethod<String, dynamic>('getAll'))!;
```
lgtm

blasten (approved):

elliotsayes:
Since this version has not been published to pub.dev as of writing https://pub.dev/packages/package_info/versions

For those in the null-safety beta wanting to access this patch now, you can import it in your `pubspec.yaml` as follows:
```
dependencies:
  package_info:
    git:
      url: git://github.com/bparrishMines/plugins.git
      ref: package_info_nnbd
      path: packages/package_info
```

PR #3399, opened 2021-01-08
[image_picker] fix version

cyanglaz:
Fix version mismatch introduced in https://github.com/flutter/plugins/commit/a7dd76950e5b240d2d878504a6f010ef24717e34


ditman (approved):
Ahh, I see the problem now. Apologies for not catching this in the previous PR.

PR #3403, opened 2021-01-11
[camera] Limit video length with maxVideoDuration on startVideoRecording

danielroek:

Adds implementation for defining a maximum video duration.

TODO

TODO

## Pre-launch Checklist

- [ ] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [ ] I signed the [CLA].


packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -170,16 +171,25 @@ private void initFps(CameraCharacteristics cameraCharacteristics) {
...
+                      .setMediaOrientation(getMediaOrientation())
+                      .setMaxVideoDuration(maxVideoDuration)
+                      .build();
+    } else {
+      mediaRecorder =
+              new MediaRecorderBuilder(recordingProfile, outputFilePath)
+                      .setEnableAudio(enableAudio)
+                      .setMediaOrientation(getMediaOrientation())
+                      .build();
+    }
```
danielroek, line 192: Simplify?

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -589,10 +599,26 @@ public void startVideoRecording(Result result) {
...
+     if(maxVideoDuration != null) {
+       mediaRecorder.setOnInfoListener((mr, what, extra) -> {
+         if (what == MEDIA_RECORDER_INFO_MAX_DURATION_REACHED) {
+           try {
+             dartMessenger.sendTimeLimitReachedEvent(videoRecordingFile.getAbsolutePath());
+             cameraCaptureSession.abortCaptures();
+             mediaRecorder.stop();
+             recordingVideo = false;
+             videoRecordingFile = null;
+             startPreview();
```
danielroek, line 615: Maybe combine this with the code in stopVideoRecording()?

packages/camera/camera/example/lib/main.dart
```diff
@@ -606,7 +606,14 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
...
     try {
-      await controller.startVideoRecording();
+      await controller.startVideoRecording(
+          maxVideoDuration: const Duration(milliseconds: 5000),
+      );
+      debugPrint('recording started');
+      controller.onCameraTimeLimitReachedEvent(onCameraTimeLimitReached: (file) {
+        debugPrint('onCameraTimeLimitReached ${file.path}');
+      });
+      debugPrint('listening');
```
danielroek, line 616: Remove here, add to README

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -482,6 +485,32 @@ class CameraController extends ValueNotifier<CameraValue> {
...
 
+  /// TODO: Documentation
+  void onCameraTimeLimitReachedEvent({onCameraTimeLimitReached}) {
+    if (!value.isInitialized || _isDisposed) {
+      throw CameraException(
+        'Uninitialized CameraController',
+        'cameraTimeLimitReachedEventStream was called on uninitialized CameraController',
+      );
+    }
+    if (!value.isRecordingVideo) {
```
danielroek, line 496: Is this correct behaviour? Should we only listen after starting the recording? Or should it be possible beforehand?

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -482,6 +485,32 @@ class CameraController extends ValueNotifier<CameraValue> {
...
+      throw CameraException(
+        'No video is recording',
+        'cameraTimeLimitReachedEventStream was called when no video is recording.',
+      );
+    }
+    debugPrint('ping');
+
+    CameraPlatform.instance.onCameraTimeLimitReached(_cameraId).listen((event) {
```
danielroek, line 504: Doesn't get closed after receiving an event, onCameraTimeLimitReached gets printed multiple times in 2nd+ recording.

packages/camera/camera/pubspec.yaml
```diff
@@ -8,7 +8,8 @@ homepage: https://github.com/flutter/plugins/tree/master/packages/camera/camera
 dependencies:
   flutter:
     sdk: flutter
-  camera_platform_interface: ^1.2.0
+  camera_platform_interface: #^1.5.0
+    path: ../camera_platform_interface
```
danielroek, line 12: Remove path

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -482,6 +485,30 @@ class CameraController extends ValueNotifier<CameraValue> {
...
+    }
+
+    CameraPlatform.instance
+        .onCameraTimeLimitReached(_cameraId)
+        .first
+        .then((event) {
+      value = value.copyWith(isRecordingVideo: false);
+      onCameraTimeLimitReached(event.path);
+    });
+  }
```
BeMacized, line 510: One issue I see with this, is that if the user sets a `maxVideoDuration`, but never calls `onCameraTimeLimitReachedEvent` to handle the callback, `value` will never be updated, causing `isRecordingVideo to still report `true`, while it has in fact already stopped. 

Ideally, the camera controller should always be listening for `CameraPlatform.instance.onCameraTimeLimitReached` to update the CameraValue, even when the user has not provided a callback, so that the state is always correct. 

Additionally, I think I would personally prefer seeing a method that returns a stream of these events (Like how the platform interface exposes events) rather than a method that allows for setting a callback, as that would imo offer a bit more flexibility. What do you think?

packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
```diff
@@ -235,3 +235,33 @@ class CameraErrorEvent extends CameraEvent {
   @override
   int get hashCode => super.hashCode ^ description.hashCode;
 }
+
+class CameraTimeLimitReachedEvent extends CameraEvent {
+  final XFile path;
+
```
BeMacized, line 241: Should the event maybe also include the time limit that was reached? I have a feeling there might be a use case for this.

packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
```diff
@@ -108,6 +108,11 @@ abstract class CameraPlatform extends PlatformInterface {
     throw UnimplementedError('unlockCaptureOrientation() is not implemented.');
   }
 
+  @override
+  Stream<CameraTimeLimitReachedEvent> onCameraTimeLimitReached(int cameraId) {
+    throw UnimplementedError('onCameraTimeLimitReached() is not implemented.');
+  }
```
BeMacized, line 114: This change to the platform interface should probably be split off to a separate PR, so that it can be published to pub.dev before the rest is merged to master.

BeMacized (commented):
Overall this already looks pretty good! 😄 
I've left a few comments below.

Additionally:
- Unit tests are still missing for your changes.
- iOS implementations are still missing.

packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
mvanbeusekom: I agree, this would be helpful information to have around.

packages/camera/camera/lib/src/camera_controller.dart
mvanbeusekom: I agree, this would be easier for users to manage if we can use this as a `Stream<XFile>` and simply listen for a file to arrive. 

We can even extend the `stopVideoRecording` method to add the final video file to the same stream. This way developers simply listen to the stream to get informed that a new video file is available (no matter if the end-user pressed the stop button or if the max time limit has been reached). Of course this would mean the name of the event and the stream should be updated a to reflect this change. Something like:

```dart
Stream<VideoRecordedEvent> onVideoRecorded(int cameraId); 
```

packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
```diff
@@ -235,3 +235,33 @@ class CameraErrorEvent extends CameraEvent {
   @override
   int get hashCode => super.hashCode ^ description.hashCode;
 }
+
+class CameraTimeLimitReachedEvent extends CameraEvent {
+  final XFile path;
```
mvanbeusekom, line 240: nit: Rename to `file` instead of `path`:
```suggestion
  final XFile file;
```

packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart
```diff
@@ -162,6 +162,11 @@ class MethodChannelCamera extends CameraPlatform {
         .whereType<DeviceOrientationChangedEvent>();
   }
 
+  @override
+  Stream<CameraTimeLimitReachedEvent> onCameraTimeLimitReached(int cameraId) {
```
mvanbeusekom, line 166: I think it would be better to rename this to `onVideoRecordingTimeLimitReached`, so it is clear this is directly related to the video recording feature. Of course if you accept my earlier comment about also emitting the recorded video on this stream when the user calls `stopVideoRecording` we should rename it to something more generic like `onVideoRecorded`.

packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
mvanbeusekom: I think it would be better to rename this to `onVideoRecordingTimeLimitReached`, so it is clear this is directly related to the video recording feature. Of course if you accept my earlier comment about also emitting the recorded video on this stream when the user calls stopVideoRecording we should rename it to something more generic like `onVideoRecorded`.

packages/camera/camera/README.md
```diff
@@ -66,6 +66,33 @@ As of version [0.5.0](https://github.com/flutter/plugins/blob/master/packages/ca
...
+        );
+        controller.onCameraTimeLimitReachedEvent(onCameraTimeLimitReached: (XFile file) {
+          //Handle the XFile
+        });
+      } on CameraException catch (e) {
+        _showCameraException(e);
+        return;
+      }
+    }
+```
```
mvanbeusekom, line 95: Really like that you didn't forget to updated the documentation in the README.md. But this should probably also be updated according the the discussion @BeMacized started on using a stream instead of registering a callback method.

mvanbeusekom (changes requested):

packages/camera/camera/lib/src/camera_controller.dart
danielroek: This might also solve possible issues with recording video's and stopping them for unusual reasons, like full storage

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -601,6 +619,18 @@ public void startVideoRecording(Result result) {
...
+  public void resetCaptureSession() throws CameraAccessException {
+    try {
+      cameraCaptureSession.abortCaptures();
+      mediaRecorder.stop();
+    } catch (CameraAccessException | IllegalStateException e) {
+      // Ignore exceptions and try to continue (changes are camera session already aborted capture)
+    }
+
+    mediaRecorder.reset();
+    startPreview();
```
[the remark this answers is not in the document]
danielroek: Removed the CameraException from the catch-clause. I'm going to leave the finally statement, as it does add the risk of making future problems harder to debug.

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -610,15 +640,8 @@ public void stopVideoRecording(@NonNull final Result result) {
...
-      } catch (CameraAccessException | IllegalStateException e) {
-        // Ignore exceptions and try to continue (changes are camera session already aborted capture)
-      }
+      resetCaptureSession();
 
-      mediaRecorder.reset();
-      startPreview();
       result.success(videoRecordingFile.getAbsolutePath());
       videoRecordingFile = null;
     } catch (CameraAccessException | IllegalStateException e) {
```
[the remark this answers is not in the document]
danielroek: That's right

kiaxseventh:
add a full example please

PR #3404, opened 2021-01-11
[video_player] Fix playback speed resetting on iOS

creativecreatorormaybenot:
This fixes the playback speed resetting on iOS on startup and when seeking.  
The approach is not the cleanest because there might be up to `500ms` (unrealistic because seeking will take some time) that are played back with reset playback speed after seeking.

However, I tried a lot of different approaches for fixing this and this seemed to be the best. The only way I figured we could do better would be by increasing the frequency of applying the playback speed. I am unaware of the underlying issue causing the issue in the first place, which prevents me from creating a better fix.

All in all, I think this is a good fix because it is 1. performant (setting every 500ms does not possibly impact performance) and 2. reliable for every kind of action the video player might take on iOS.  
I decided against increasing the frequency (e.g. every `100ms` or even more often) because I do not want to impact performance.

---

* Fixes https://github.com/flutter/flutter/issues/73643.
* Fixes https://github.com/flutter/flutter/issues/71264.

## Pre-launch Checklist

- [x] I signed the [CLA].


IchordeDionysos:
@cyanglaz any chance we can merge this now that the null safe version is finished? 🤔

IchordeDionysos:
@cyanglaz @iskakaushik any updates here?

creativecreatorormaybenot:
@cyanglaz @iskakaushik @stuartmorgan what do we need to do here to move this forward?

stuartmorgan:
Sorry for the delay; we're still working on triaging the PR backlog. From an initial look this will need tests, per policy, and it will need non-trivial review, so will need to go into [the backlog](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#the-review-process).

To elaborate on the second point:
- As described in the PR comment and the comment in the code this is essentially hacking around an issue we don't actually understand yet, which is not something we generally do until we've done more investigation, and
- It's not clear from the PR how you came to the conclusion that "setting every 500ms does not possibly impact performance" given that it's doing a method channel invocation, which happens on the main thread.

creativecreatorormaybenot:
> From an initial look this will need tests

@stuartmorgan if you have any idea how to test this, would love to know.

> It's not clear from the PR how you came to the conclusion that "setting every 500ms does not possibly impact performance" given that it's doing a method channel invocation, which happens on the main thread.

To be honest, I wrote some of the other code and this is the least impact but it is quite a critical bug (e.g. we cannot use the plugin without this PR).

It would just be nice to move this forward, given we have been on our own fork for 9 months for a 1 line change.

stuartmorgan (changes requested):
> > From an initial look this will need tests
> 
> @stuartmorgan if you have any idea how to test this, would love to know.

As currently written it would be somewhat difficult to test since the fix is at the wrong layer; I expect a native fix should be relatively straightforward to test with a native unit test using a mock AVPlayer.

> It would just be nice to move this forward, given we have been on our own fork for 9 months for a 1 line change.

The number of lines isn't the issue, it's the content. I've looked more at the linked issues, and this is definitely not a fix we would accept; the issue is:
- platform-specific
- specific to the seek action
- specific to certain video types

This workaround would affect every platform, constantly making calls across the Dart/native boundary even when no seeking is happening, with no targeting to cases that cause the bug.

From the comment there about it having reset by the time the completion callback fires it sounds like it should be straightforward to do a highly targeted native-side fix that would have no negative side effects.

creativecreatorormaybenot:
@stuartmorgan your concerns are great but also idealistic.

This is a major bug that has not been addressed for a whole year. Merging this would fix it.

Your concern is valid, however, negligible. The value is updated so often that the single call has 0 measurable effect.

creativecreatorormaybenot:
I am all in favor of a better fix. What I would do: merge this and create an issue to clean this up.

creativecreatorormaybenot:
And if your concern is quality of code, I am fully with you. If I owned this plugin, the first thing I would do is clean up all of the wanky handling.

The value is not stable at all, which causes this problem in the first place. You cannot trust the value one bit - there are so many ways to bring the value (Dart side) out of sync with the actual underlying video player state (native side).

stuartmorgan:
> @stuartmorgan your concerns are great but also idealistic.

I'm willing to listen to explanations for why a native-side fix that has none of those drawbacks would be hard to implement.

> Your concern is valid, however, negligible.

You asked me what was necessary to move the PR forward; asserting that my comments don't matter isn't going to move it forward.

> And if your concern is quality of code, I am fully with you. If I owned this plugin, the first thing I would do is clean up all of the wanky handling.

Adding hacks instead of addressing root causes directly works against that goal. If the first thing you would do is improve code quality, then I'm not sure why you are arguing against improving this PR as part of the review process.

> The value is not stable at all, which causes this problem in the first place. You cannot trust the value one bit - there are so many ways to bring the value (Dart side) out of sync with the actual underlying video player state (native side).

I'm happy to review PRs that address root causes of out-of-sync values, rather than layering hacks over them.

creativecreatorormaybenot:
@stuartmorgan 

> Adding hacks instead of addressing root causes directly works against that goal. If the first thing you would do is improve code quality, then I'm not sure why you are arguing against improving this PR as part of the review process.

I am not at all. What I am trying to say is that I am not familiar with the native side and therefore I cannot move **this** PR in the direction you pointed out.

Thus, I would merge this and create an issue to fix up the native side once there is someone with the knowledge to do so.

creativecreatorormaybenot:
(I would have fixed it on the native side if I knew how - I am sure I have the ability to learn how to do it quickly, but not quickly enough that I have the time to do it, unfortunately)

creativecreatorormaybenot:
> If the first thing you would do is improve code quality, then I'm not sure why you are arguing against improving this PR as part of the review process.

The issue is that I do not have the capacity to do it. Ideally, I would do it 😕

stuartmorgan:
Based on the comments above it doesn't sound like this PR is going to reach a point where it is landable, so I'm going to close it. If at some point in the future you are interested in revisiting a fix that addresses the comments above, please feel free to open a new PR.

creativecreatorormaybenot:
I guess we will have to fork the plugin then.

creativecreatorormaybenot:
https://github.com/flutter/plugins/pull/4331 should be the iOS-side fix 🎉 

But it is also lacking unit tests right now..

PR #3408, opened 2021-01-12
Change onMethodCall to protected

sagallea:
This will allow subclasses of MethodChannelWebViewPlatform to have
custom method call handling, calling up to the superclass if not needed.

Bug: b/169054926


sagallea:
Closed pull request since we no longer intend to depend on this plugin.

PR #3426, opened 2021-01-14
[google_maps_flutter] Fix iOS crash by observing map frame change only once

guykogus:
I cannot to tell you the stress this bug has caused me 😅

The `view` method gets called _a lot_. Every draw cycle, in fact. And if the frame should change, for example when the keyboard appears, every single one of those `observeValueForKeyPath:` observations will be called, which crashes an app with the following exception: `EXC_BAD_ACCESS KERN_PROTECTION_FAILURE`

Screen Shot 2021-01-15 at 00 17 58

## Pre-launch Checklist

- [x] I signed the [CLA].


MarceLifeBites:
Thx @guykogus , we are having de same bug here, and fix it with help a lot with our platform stability, so i wish they can release the solution as soon as possible

guykogus:
> I can confirm changes from this pull seems to resolves the problem. Beside, it should be released with version lower than 2.0.0 because plenty of projects depends on more libraries than only google maps and won't be migrated to flutter 2.0.0 for a while (dependency hell 😢 )

Unfortunately there's no branch in the original repo to merge with to keep it at version < 2.0 ☹️

pawlowskim:
> > I can confirm changes from this pull seems to resolves the problem. Beside, it should be released with version lower than 2.0.0 because plenty of projects depends on more libraries than only google maps and won't be migrated to flutter 2.0.0 for a while (dependency hell 😢 )
> 
> Unfortunately there's no branch in the original repo to merge with to keep it at version < 2.0 ☹️

I'm aware, but the fix won't be usable for old projects that have dozens of non flutter / dart team libraries dependencies. Until then, guess we will use your fork:
```
  google_maps_flutter:
      git:
          url: git://github.com/guykogus/flutter_plugins
          path: packages/google_maps_flutter/google_maps_flutter
          ref: 6156ea3fa2ecf4b0e7ff35ba9a5fb1414e181d6d
```

guykogus:
> > > I can confirm changes from this pull seems to resolves the problem. Beside, it should be released with version lower than 2.0.0 because plenty of projects depends on more libraries than only google maps and won't be migrated to flutter 2.0.0 for a while (dependency hell 😢 )
> > 
> > 
> > Unfortunately there's no branch in the original repo to merge with to keep it at version < 2.0 ☹️
> 
> I'm aware, but the fix won't be usable for old projects that have dozens of non flutter / dart team libraries dependencies. Until then, guess we will use your fork:
> 
> ```
>   google_maps_flutter:
>       git:
>           url: git://github.com/guykogus/flutter_plugins
>           path: packages/google_maps_flutter/google_maps_flutter
>           ref: 6156ea3fa2ecf4b0e7ff35ba9a5fb1414e181d6d
> ```

You can use `ref: via` if you want. I've got another fix there for Android, which can crash if the native Android map loads after the Flutter map is unmounted (i.e. the `StatefulWidget` is `disposed`). Unfortunately the PR was rejected cos they couldn't reproduce it.

sangTroo:
> Flutter 1.22.5, google_maps_flutter: 1.0.6. For us it happens when user stays on a map for ~30 minutes (app in foreground, screen is on) and we are tracking his position and update his position marker (custom one, so we are re-drawing circle on a map). To simulate it, I just use more or less this piece of code:

We have the same problem. Will try this branch to see if it fixes the issue.  The app that we're building is heavily use maps. And hasn't been stable since this bug keeps the app crashing

carlosmobile:
Hi! We have the same problem. We are using a fork with this changes and fixed the crash. Thanks!

cyanglaz (commented):
Thanks for providing the fix, the change looks good. 

It's not clear to me in what scenario the `view` is called multiple times for a single webview.
It looks like it's only called once per controller, see: https://github.com/flutter/engine/blob/master/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm#L161

Again, this is a good fix nonetheless, the old code is pretty confusing and didn't do what it says it wants to do.  

However, we should add some tests before landing this. I wonder If you know a way to reproduce this consistently.

via-guy:
> Thanks for providing the fix, the change looks good.
> 
> It's not clear to me in what scenario the `view` is called multiple times for a single webview.
> It looks like it's only called once per controller, see: https://github.com/flutter/engine/blob/master/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm#L161
> 
> Again, this is a good fix nonetheless, the old code is pretty confusing and didn't do what it says it wants to do.
> 
> However, we should add some tests before landing this. I wonder If you know a way to reproduce this consistently.

I couldn't come up with one when I made this fix. It happens in our project, and clearly it does for others too. Do any of you have a simple sample test that can be provided here, @pawlowskim @sangTroo @carlosmobile ?

sangTroo:
> > Thanks for providing the fix, the change looks good.
> > It's not clear to me in what scenario the `view` is called multiple times for a single webview.
> > It looks like it's only called once per controller, see: https://github.com/flutter/engine/blob/master/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm#L161
> > Again, this is a good fix nonetheless, the old code is pretty confusing and didn't do what it says it wants to do.
> > However, we should add some tests before landing this. I wonder If you know a way to reproduce this consistently.
> 
> I couldn't come up with one when I made this fix. It happens in our project, and clearly it does for others too. Do any of you have a simple sample test that can be provided here, @pawlowskim @sangTroo @carlosmobile ?

try this #72821

sangTroo:
Any update when this fix will be merged?

sangTroo:
> > Thanks for providing the fix, the change looks good.
> > It's not clear to me in what scenario the `view` is called multiple times for a single webview.
> > It looks like it's only called once per controller, see: https://github.com/flutter/engine/blob/master/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm#L161
> > Again, this is a good fix nonetheless, the old code is pretty confusing and didn't do what it says it wants to do.
> > However, we should add some tests before landing this. I wonder If you know a way to reproduce this consistently.
> 
> I couldn't come up with one when I made this fix. It happens in our project, and clearly it does for others too. Do any of you have a simple sample test that can be provided here, @pawlowskim @sangTroo @carlosmobile ?

Hi,
What kind of simple sample test do you want, sorry I don't get what you mean

stuartmorgan (changes requested):
It looks like this should be easily testable with a native unit test:
- Allow injecting a mock `GMSMapView`
- Call `view` several times
- Ensure that `addObserver:...` is not called on the mock repeatedly.

guykogus:
@stuartmorgan yes, I really want to because this change is essential, I just haven't had a chance to do it...

via-guy:
@stuartmorgan I've added some tests, but I'm not sure if I've really done it correctly. Basically I just tested to see if it doesn't lag a long time to loop over `view`. On the `master` branch the test fails, but it passes on mine.

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * Removes dependencies from `pubspec.yaml` that are only needed in `example/pubspec.yaml`
 * Updates Android compileSdkVersion to 31.
+* Fix iOS crash on `EXC_BAD_ACCESS KERN_PROTECTION_FAILURE` if the map frame changes long after creation.
```
stuartmorgan, line 5: Fixes

(Per the repo CHANGELOG style guide.)

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
```
stuartmorgan, line 2: This change needs a version bump.

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/FlutterMock.h
```diff
@@ -0,0 +1,19 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import Foundation;
+@import Flutter;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MockRegistrar : NSObject <FlutterPluginRegistrar>
```
stuartmorgan, line 10: Why are you mocking these manually instead of with OCMock?

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/GoogleMapsTests.m
```diff
@@ -15,4 +17,26 @@ - (void)testPlugin {
...
+                                     viewIdentifier:0
+                                          arguments:@{}
+                                          registrar:object];
+
+  for (NSInteger i = 0; i < 10000; ++i) {
+    [controller view];
+  }
+  [[controller view] setValue:[NSValue valueWithCGRect:CGRectMake(0, 0, 0, 0)] forKey:@"frame"];
+
+  [viewExpectation fulfill];
```
stuartmorgan, line 39: I don't understand what this test is supposed to be doing. It's not asserting what this PR is actually intended to fix.

I [described what a test should look like above](https://github.com/flutter/plugins/pull/3426#pullrequestreview-771597382); I'm not sure why you've done something completely different.

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/GoogleMapsTests.m
via-guy: That's a good point. Got distracted with seeing if I could actually test for the crash itself to happen. I'll update it.

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/FlutterMock.h
via-guy: I've never used OCMock. Thanks for the tip!

via-guy:
@stuartmorgan I've updated according to your suggestions, but I'm not sure about the placement of GoogleMapController-Test.h. I added it into the example project itself, because otherwise it would be exported with the pod. Is that correct, or is there some way in the podspec to specify target test files?

stuartmorgan:
> @stuartmorgan I've updated according to your suggestions, but I'm not sure about the placement of GoogleMapController-Test.h. I added it into the example project itself, because otherwise it would be exported with the pod. Is that correct, or is there some way in the podspec to specify target test files?

It should live next to the source file; you can use a module map to control how it's exposed. See https://github.com/flutter/plugins/pull/4430 for an example of setting that up.

packages/google_maps_flutter/google_maps_flutter/example/ios/Podfile
```diff
@@ -31,6 +31,8 @@ target 'Runner' do
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
   target 'RunnerTests' do
     inherit! :search_paths
+
+    pod 'OCMock'
```
stuartmorgan, line 35: This should have a version. See [other uses](https://github.com/flutter/plugins/search?l=Ruby&q=pod+ocmock) for examples.

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/GoogleMapController-Test.h
```diff
@@ -0,0 +1,19 @@
...
+// found in the LICENSE file.
+
+#import <Flutter/Flutter.h>
+#import <GoogleMaps/GoogleMaps.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FLTGoogleMapController (Test)
+
+- (instancetype)initWithMapView:(GMSMapView *)mapView
```
stuartmorgan, line 12: All methods need declaration comments, per the ObjC style guide.

packages/google_maps_flutter/google_maps_flutter/ios/Classes/GoogleMapController.m
```diff
@@ -119,24 +125,20 @@ - (instancetype)initWithFrame:(CGRect)frame
...
-  [_mapView addObserver:self forKeyPath:@"frame" options:0 context:nil];
   return _mapView;
 }
 
 - (void)observeValueForKeyPath:(NSString *)keyPath
                       ofObject:(id)object
                         change:(NSDictionary *)change
                        context:(void *)context {
-  if (_cameraDidInitialSetup) {
-    // We only observe the frame for initial setup.
```
stuartmorgan, line 136: Rather than remove this comment, it would be good to move it down to just above the remaining call to `removeObserver:forKeyPath:`.

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/MockGMSMapView.h
```diff
@@ -0,0 +1,11 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import GoogleMaps;
+
+@interface MockGMSMapView : GMSMapView
```
stuartmorgan, line 7: Needs a declaration comment.

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/MockGMSMapView.h
```diff
@@ -0,0 +1,11 @@
...
+
+@property(nonatomic, assign, readonly) NSInteger frameObserverCount;
```
stuartmorgan, line 9: Needs a declaration comment.

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/MockGMSMapView.h
stuartmorgan, line 7: You should probably call this `PartiallyMockedMapView` or something along those lines; the assumption for a mock object is not that it is the real object but with some overridden behavior, but that it has no behavior other than what it is given in test setup.

stuartmorgan (changes requested):

stuartmorgan:
@cyanglaz Do you want to do the second review here since you've already looked at it previously?

via-guy:
@stuartmorgan ok, I think I've finally handled all your comments. Not sure how great my documentation comments are, but hopefully they're good enough.
Thanks so much for your help!

packages/google_maps_flutter/google_maps_flutter/ios/Classes/GoogleMapController_Test.h
```diff
@@ -0,0 +1,27 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FLTGoogleMapController (Test)
+
+/**
+ * Initialize a map controller with a concrete map view.
+ *
+ * @param mapView A map view that will be displayed by the controller
+ * @param viewId A unique identifier for the controller (unused).
```
cyanglaz, line 16: What does the (unused) mean? We did store it in an ivar: `_viewId = viewId;`
Does it mean the ivar itself is unused in the current implementation? I think this makes the reader confused. We should just remove the `(unused)` part. If the ivar `_viewId` is currently not accessed in the implementation, we can just remove the ivar itself. (can be in a different PR)

cyanglaz (approved):
LGTM module nits

packages/google_maps_flutter/google_maps_flutter/ios/Classes/GoogleMapController_Test.h
via-guy: Yeah, the ivar isn't actually being used in any way. I removed the comment.

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Removes dependencies from `pubspec.yaml` that are only needed in `example/pubspec.yaml`
 * Updates Android compileSdkVersion to 31.
 * Internal code cleanup for stricter analysis options.
+* Fixes iOS crash on `EXC_BAD_ACCESS KERN_PROTECTION_FAILURE` if the map frame changes long after creation.
```
stuartmorgan, line 6: You can't add changes to a version that has already been published. It needs a new version.

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/PartiallyMockedMapView.h
```diff
@@ -0,0 +1,13 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import GoogleMaps;
+
+// Defines a map view used for testing key-value observing.
```
stuartmorgan, line 7: Please use Doxygen style: https://google.github.io/styleguide/objcguide.html#declaration-comments.

packages/google_maps_flutter/google_maps_flutter/example/ios/RunnerTests/PartiallyMockedMapView.h
```diff
@@ -0,0 +1,13 @@
...
+@interface PartiallyMockedMapView : GMSMapView
+
+/// Counts the number of times that `frame` KVO has been added
```
stuartmorgan, line 10: Same. You're also missing the period; comments should be properly punctuated, per the style guide.

stuartmorgan: Also, this is a property, so should describe a noun rather than a verb; methods do things, properties are things. I.e., "The number of times that [...]".

packages/google_maps_flutter/google_maps_flutter/ios/Classes/GoogleMapController_Test.h
```diff
@@ -0,0 +1,27 @@
...
+
+#import <Flutter/Flutter.h>
+#import <GoogleMaps/GoogleMaps.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FLTGoogleMapController (Test)
+
+/**
+ * Initialize a map controller with a concrete map view.
```
stuartmorgan, line 13: Nit: Initializes

(Per style guide.)

packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h
```diff
@@ -0,0 +1,9 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Foundation/Foundation.h>
+#import <google_maps_flutter/FLTGoogleMapsPlugin.h>
```
stuartmorgan, line 6: From CI:

```
Testing failed:
	Umbrella header for module 'google_maps_flutter' does not include header 'JsonConversions.h'
	Umbrella header for module 'google_maps_flutter' does not include header 'FLTGoogleMapTileOverlayController.h'
```

(JsonConversions.h shouldn't actually be public, but you don't need to fix that in this PR; it's a pre-existing mistake. @jmagman Maybe we should do an audit of the headers our plugins export, and make a repo convention for paths to put headers that aren't intended to be public?)

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h
via-guy: I added `FLTGoogleMapTileOverlayController.h`.

via-guy:
@stuartmorgan done

packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h
jmagman: Still getting 
```
 - WARN  | [iOS] xcodebuild:  Headers/Private/google_maps_flutter/google_maps_flutter-umbrella.h:10:1: warning: umbrella header for module 'google_maps_flutter' does not include header 'JsonConversions.h' [-Wincomplete-umbrella]
 - ```

Comment (stuartmorgan) [reply packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h:6]: To clarify my "JsonConversions.h shouldn't actually be public, but you don't need to fix that in this PR" comment: you don't need to fix the fact that it's public and shouldn't be in this PR. You do need to fix the fact that as a (currently) public header, it needs to be in the umbrella.

Comment (via-guy) [reply packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h:6]: Ah wait, do I want these 2 headers in `FLTGoogleMapsPlugin.h` or in the umbrella? Right now they're in the plugin header...

Comment (via-guy) [reply packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h:6]: I moved them to the umbrella

**Comment by jmagman** [comment]
There's a format error, please run:
```
patch -p1 <<DONE
diff --git a/packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h b/packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h
index 5af927f61..50880a2b9 100644
--- a/packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h
+++ b/packages/google_maps_flutter/google_maps_flutter/ios/Classes/google_maps_flutter-umbrella.h
@@ -3,8 +3,8 @@
 // found in the LICENSE file.
 
 #import <Foundation/Foundation.h>
-#import <google_maps_flutter/FLTGoogleMapsPlugin.h>
 #import <google_maps_flutter/FLTGoogleMapTileOverlayController.h>
+#import <google_maps_flutter/FLTGoogleMapsPlugin.h>
 #import <google_maps_flutter/JsonConversions.h>
 
 FOUNDATION_EXPORT double google_maps_flutterVersionNumber;

DONE
```

via-guy:
@jmagman done

stuartmorgan (approved):
LGTM!

PR #2895, opened 2020-07-24
[webview_flutter] add option to set the background to transparent

crizant:
Regarding this issue:
https://github.com/flutter/flutter/issues/29300

PR #3431, opened 2021-01-15
[webview_flutter] Add an option to set the background color of the webview

e-adrien:
This PR add an option to set a transparent background for the webview.

There are already 2 outdated PR for the same feature :
 * https://github.com/flutter/plugins/pull/1336
 * https://github.com/flutter/plugins/pull/2895

I took the previous PR made by @crizant and rebase it.

It fixes this issue :
 * https://github.com/flutter/flutter/issues/29300


fundup:
So close to have this feature  ... waiting for month as others !!! Hope this merge will be accepted and in the next Webview version !

crizant:
There are conflicts, seems @e-adrien you need to rebase again. 🤪

davedbase:
Very excited to have this merged in and available. The overlay/opacity hacks are quite annoying. Also thanks @crizant and @e-adrien for your effort on this.

brenoasm:
Any updates?

packages/webview_flutter/example/lib/main.dart
```diff
@@ -45,6 +46,7 @@ class _WebViewExampleState extends State<WebViewExample> {
   @override
   Widget build(BuildContext context) {
     return Scaffold(
+      backgroundColor: Colors.green,
```
gaaclarke, line 49: Why is this hardcoded as green?

e-adrien: It was to show the effect of the transparent background option in the example.
I can remove this change in the example and just keep the "opaque" option to the default value (true).

gaaclarke: Oh sorry, i missed this was in the example code.  Nevermind.

stuartmorgan:
Thanks for the submission! As mentioned in the other PR, this will need tests in order to move forward:

> This will need a test of the code in webview_flutter.dart (that it's sent to the interface); ideally there would native testing as well, but we need infrastructure plumbed up for that for webview_flutter still, and I don't think we need to gate this change on that.

Since that was written, the necessary test harness on the iOS side has been added, so native tests for iOS are now also possible.

e-adrien:
Hi, I added some tests, but I can't find a way to take a screenshot of the webview for Android.
I always have the container without it's child webview in the image.
I think I have to wait for https://github.com/flutter/flutter/issues/83856 to do a proper test for that.

packages/webview_flutter/example/ios/RunnerUITests/FLTWebViewUITests.m
```diff
@@ -18,6 +41,57 @@ - (void)setUp {
...
+  XCTAttachment* colorCenterAtt =
+      [XCTAttachment attachmentWithString:centercicolor.stringRepresentation];
+  [colorCenterAtt setLifetime:XCTAttachmentLifetimeKeepAlways];
+  [colorCenterAtt setName:@"Center color"];
+  [self addAttachment:colorCenterAtt];
+
+  CIColor* flutterGreenColor = [CIColor colorWithRed:76.0f / 255.0f
+                                               green:175.0f / 255.0f
+                                                blue:80.0f / 255.0f
+                                               alpha:1.0f];
```
gaaclarke, line 89: Add comment with a link to the source file in Flutter that is using this color.

e-adrien:
Hi @gaaclarke , thank you for your review. 

I made the changes you asked to the iOS native test.

For the integration test there is no getter for the background color in the [Android Webview](https://developer.android.com/reference/android/webkit/WebView).
I was thinking about adding a method to get the pixel color at a relative position as the canvas has the color defined by the setBackgroundColor call.
```java
private void getPixelColorAt(MethodCall methodCall, Result result) {
  // Parameters
  Map<String, Object> request = methodCall.arguments();
  double x = (double) request.get("x");
  double y = (double) request.get("y");

  // Get an image of the WebView
  Bitmap webviewBitmap = Bitmap.createBitmap(webView.getWidth(), webView.getHeight(), Bitmap.Config.ARGB_8888);
  Canvas bitmapCanvas = new Canvas(webviewBitmap);
  webView.draw(bitmapCanvas);

  // Find the pixel at the given position
  int pixelColor = webviewBitmap.getPixel((int) Math.floor(x * webviewBitmap.getWidth()), (int) Math.floor(y * webviewBitmap.getHeight()));

  // Send the result
  result.success(String.format("#%08X", pixelColor));
}
```

gaaclarke (commented):
There still isn't an assert for the android tests.

e-adrien:
Hi.
I'll make the change for iOS.
For the test on Android I wanted to have your opinion on the way to do that (https://github.com/flutter/plugins/pull/3431#issuecomment-874900915).
Thank you for your time.

gaaclarke:
@e-adrien That sounds good to me if it works.  I'm not as familiar with the Android API so I'm not sure if rendering to the bitmap like that will work for you.  I think what you are proposing is ideal, otherwise a platform channel that reads that value is fine for me.

e-adrien:
I renamed the opaque parameter to transparent (false by default).
I also added the test for Android.
I'm not familiar with Android too, so maybe there is a better way to do that.

gaaclarke (commented):
This is great, thanks.  I do have one niggle though.  You added the functionality for getting the color of the pixel to the interface for the webview plugin instead of making a separate platform channel that lives just in the test harness.

It's kind of in a weird state now where it is in there just for testing, which isn't bad, but it isn't implemented for all platforms.  Technically if you implemented it for ios and android you could have the tests just written in Flutter.

I think we should either:
1) Move the 'getPixelColorAt' to the test harness and in it's own platform channel
2) Implement the 'getPixelColorAt' for all platforms and unify the tests to use it

If we leave it as is, we have an obligation to maintain and implement the other platforms.  Do you see what I'm saying?

stuartmorgan:
Special test accessors should be their own separate platform interface; they should not presume that implementations use method channels.

stuartmorgan:
Stepping back a bit though: do we actually need an accessor, or can this be tested via native integration tests rather than a Dart integration test? That would be much simpler than a new test-only mechanism for querying native state from the Dart side.

gaaclarke:
> Stepping back a bit though: do we actually need an accessor, or can this be tested via native integration tests rather than a Dart integration test? That would be much simpler than a new test-only mechanism for querying native state from the Dart side.

That would mean they could just do the assert in the Android code, right?  So, they wouldn't need to add a platform channel and would have to rewrite the test for a different test harness?  They are so close with this approach, imo writing the platform channel is quicker to getting where we want.  I guess not having a platform channel is simpler, but I think having the test written more like how typical Flutter development happens is probably more accessible too.

packages/webview_flutter/lib/webview_flutter.dart
```diff
@@ -230,6 +230,7 @@ class WebView extends StatefulWidget {
     this.userAgent,
     this.initialMediaPlaybackPolicy =
         AutoMediaPlaybackPolicy.require_user_action_for_all_media_types,
+    this.transparent = false,
```
stuartmorgan, line 233: Sorry to throw a wrench into the design aspect at this point, but I don't think we should add this; it will interact badly with implementing https://github.com/flutter/flutter/issues/37697 unless we make a breaking change, which is painful for everyone.

Is there a reason this can't just be a `backgroundColor` paramater that includes an alpha channel in the color specification?

stuartmorgan:
> That would mean they could just do the assert in the Android code, right? So, they wouldn't need to add a platform channel and would have to rewrite the test for a different test harness?

Yes. Which, looking at the whole PR (sorry, previously I was just responding to the comment I saw go by since I was subscribed to the PR) is already what this PR is doing on iOS.

> They are so close with this approach, imo writing the platform channel is quicker to getting where we want.

It's not just a platform channel, it's a new, separate platform interface with its own registration, default implementation, etc. It's doable, but it's more design work.

> I guess not having a platform channel is simpler, but I think having the test written more like how typical Flutter development happens is probably more accessible too.

Plugins that involve non-trivial native code often require native tests; that's a reality of plugin development. I don't really see how writing a custom native test channel for a single test is more typical of Flutter development than just writing a native test.

I see three ways this could be tested:
1. A chain of unit tests:
  - In the app-facing portion, testing that the param is passed to the platform interface correctly
  - In the platform implementations, testing that the parameter is passed to the next layer correctly (in this case, to the method channel implementation of the interface)
  - In the native code, that including the param in the method channel results in the right property being set on the web view (I don't think we actually need to check the pixel value, since handling the background color correctly is up to the OS code, not our code, although it's nice to have).
2. Native UI tests (like what is being done for iOS already)
3. Flutter integration tests with a native helper (as you suggested)

I think any of the options is fine here, and there are pros and cons to each. If 3 is the easiest, that's fine, but it needs to be done in a way that is consistent with the federated design rather than hard-coded to be based on method channels. It's also not clear to me why we would do 2 for iOS and 3 for Android, that seems unnecessarily complicated.

e-adrien:
I wasn't aware of that request. I can update the PR to take a color with an alpha channel instead of the transparent boolean.
For the Android Native UI test I think I need to wait for https://github.com/flutter/plugins/pull/4147 (tell me if I'm wrong).

e-adrien:
I replace the `transparent` boolean parameter by a `backgroundColor` parameter.

e-adrien:
Hi @stuartmorgan 
I replaced the integration test for Android by an espresso test.
Let me know if you want me to change anything.

packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
```diff
@@ -10,10 +10,17 @@ import 'dart:typed_data';
...
+import 'package:integration_test/integration_test.dart';
 import 'package:webview_flutter/platform_interface.dart';
 import 'package:webview_flutter/webview_flutter.dart';
-import 'package:integration_test/integration_test.dart';
+import 'package:webview_flutter_example/main.dart' as app;
+
+void appMain() {
+  enableFlutterDriverExtension();
+  app.main();
+}
```
stuartmorgan, line 23: Why is this needed? We have integration test drivers in each application already.

packages/webview_flutter/webview_flutter/example/ios/RunnerUITests/FLTWebViewUITests.m
```diff
@@ -5,6 +5,30 @@
 @import XCTest;
 @import os.log;
 
+static UIColor* getPixelColorInImage(UIImage* image, int x, int y) {
+  CFDataRef pixelData = CGDataProviderCopyData(CGImageGetDataProvider(image.CGImage));
+  const UInt8* data = CFDataGetBytePtr(pixelData);
+
+  int imageWidth = floor(image.size.width * image.scale);
```
stuartmorgan, line 12: This is not safe; you want the image's stride, which is `CGImageGetBytesPerRow`.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -95,6 +95,20 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+    NSNumber* backgroundColorNSNumber = args[@"backgroundColor"];
+    if ([backgroundColorNSNumber isKindOfClass:[NSNumber class]]) {
+      int backgroundColorInt = [backgroundColorNSNumber intValue];
+      UIColor* backgroundColor = [UIColor colorWithRed:(backgroundColorInt >> 16 & 0xff) / 255.f
+                                                 green:(backgroundColorInt >> 8 & 0xff) / 255.f
+                                                  blue:(backgroundColorInt & 0xff) / 255.f
+                                                 alpha:(backgroundColorInt >> 24 & 0xff) / 255.f];
+      _webView.opaque = NO;
+      _webView.backgroundColor = backgroundColor;
+      _webView.scrollView.backgroundColor = backgroundColor;
```
stuartmorgan, line 109: Does setting both of these actually work with semi-opaque values? I would expect it to layer and give the wrong value.

It would be good to have a test covering that.

packages/webview_flutter/webview_flutter/lib/platform_interface.dart
```diff
@@ -488,6 +489,9 @@ class CreationParams {
   /// Which restrictions apply on automatic media playback.
   final AutoMediaPlaybackPolicy autoMediaPlaybackPolicy;
 
+  /// The background color of the webview.
```
stuartmorgan, line 492: Please explicitly document that `null` is the OS-default background.

packages/webview_flutter/webview_flutter/lib/webview_flutter.dart
```diff
@@ -414,6 +415,11 @@ class WebView extends StatefulWidget {
   /// The default policy is [AutoMediaPlaybackPolicy.require_user_action_for_all_media_types].
   final AutoMediaPlaybackPolicy initialMediaPlaybackPolicy;
 
+  /// The background color of the webview.
+  ///
+  /// By default `backgroundColor` is null.
```
stuartmorgan, line 420: Same here; document what that means.

stuartmorgan (changes requested):
Some small comments, but the bigger issue is that the iOS test is failing in CI. Does it pass for you locally?

(The Android test won't run in presubmit CI unfortunately since it's a Firebase Test Lab test which requires commit access, but I can validate that later.)

stuartmorgan:
(FYI, we're in the process of restructuring this plugin to be fully federated (so, moving the platform interface and the platform implementations to separate packages). I can help with rebasing this once that happens, since it'll only affect this PR because of the long review delay.)

e-adrien:
Hi @stuartmorgan 

I did the following :
- rebasing this PR on the master
- moving iOS native test into webview_flutter_wkwebview & use CGImageGetBytesPerRow
- moving Android native test into webview_flutter_android
- adding missing comments
- adding an iOS test for semi-opaque colors

For the Android test I had to add ```enableFlutterDriverExtension();``` in the example of the webview_flutter_android otherwise the test was failing with the error :
```
java.lang.RuntimeException: java.util.concurrent.ExecutionException: androidx.test.espresso.flutter.internal.protocol.impl.FlutterProtocolException: Flutter testing APIs not registered with Dart isolate isolates/4418401856323539.
	at androidx.test.espresso.flutter.EspressoFlutter$WidgetInteraction$ExceptionPropagator.<init>(EspressoFlutter.java:179)
```

Native tests on both platforms works locally.
```
[✓] Flutter (Channel stable, 2.5.1, on macOS 11.6 20G165 darwin-arm, locale en-GB)
    • Flutter version 2.5.1 at /Users/adrien/Library/Flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision ffb2ecea52 (13 days ago), 2021-09-17 15:26:33 -0400
    • Engine revision b3af521a05
    • Dart version 2.14.2

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at /Users/adrien/Library/Android/sdk
    • Platform android-31, build-tools 30.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 13.0, Build version 13A233
    • CocoaPods version 1.11.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2020.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)

[✓] VS Code (version 1.60.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.26.0

[✓] Connected device (1 available)
    • Chrome (web) • chrome • web-javascript • Google Chrome 94.0.4606.61

• No issues found!
```

packages/webview_flutter/webview_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.1.2
```
stuartmorgan, line 1: These should all be minor version bumps, not bugfix version bumps, as it's adding a new feature.

stuartmorgan:
> For the Android test I had to add `enableFlutterDriverExtension();` in the example of the webview_flutter_android otherwise the test was failing with the error :
> 
> ```
> java.lang.RuntimeException: java.util.concurrent.ExecutionException: androidx.test.espresso.flutter.internal.protocol.impl.FlutterProtocolException: Flutter testing APIs not registered with Dart isolate isolates/4418401856323539.
> 	at androidx.test.espresso.flutter.EspressoFlutter$WidgetInteraction$ExceptionPropagator.<init>(EspressoFlutter.java:179)
> ```

This can be done via a driver, rather than directly in `main`; see https://github.com/flutter/plugins/search?l=Dart&q=enableFlutterDriverExtension&type=

But... @dnfield shouldn't [integration_test_driver.dart](https://github.com/flutter/flutter/blob/master/packages/integration_test/lib/integration_test_driver.dart), which we switched to for most of the plugins (including this one) recently, be doing this automatically?

packages/webview_flutter/webview_flutter_wkwebview/example/ios/RunnerUITests/FLTWebViewUITests.m
```diff
@@ -18,6 +42,97 @@ - (void)setUp {
...
+  XCUIScreenshot* screenshot = [[XCUIScreen mainScreen] screenshot];
+
+  UIImage* screenshotImage = screenshot.image;
+  UIColor* centerLeftColor = getPixelColorInImage(
+      screenshotImage, 0, (screenshotImage.scale * screenshotImage.size.height) / 2);
+  UIColor* centerColor = getPixelColorInImage(
+      screenshotImage, (screenshotImage.scale * screenshotImage.size.width) / 2,
+      (screenshotImage.scale * screenshotImage.size.height) / 2);
+  CIColor* centerLeftCIColor = [CIColor colorWithCGColor:centerLeftColor.CGColor];
+  CIColor* centerCIColor = [CIColor colorWithCGColor:centerColor.CGColor];
```
stuartmorgan, line 77: Why are you converting from CGColors to CIColors to compare rather than using `CGColorEqualToColor`?

dnfield:
> But... @dnfield shouldn't [integration_test_driver.dart](https://github.com/flutter/flutter/blob/master/packages/integration_test/lib/integration_test_driver.dart), which we switched to for most of the plugins (including this one) recently, be doing this automatically?

That file runs on the host side. The line to enable the extension must be run device side.

stuartmorgan:
> > But... @dnfield shouldn't [integration_test_driver.dart](https://github.com/flutter/flutter/blob/master/packages/integration_test/lib/integration_test_driver.dart), which we switched to for most of the plugins (including this one) recently, be doing this automatically?
> 
> That file runs on the host side. The line to enable the extension must be run device side.

Oh, I was misremembering the boilerplate that that file replaced. It looks like the real problem is that we haven't [migrated all of the tests off of `flutter_driver`](https://flutter.dev/docs/testing/integration-tests#migrating-from-flutter_driver).

stuartmorgan:
Except this plugin is using integration_test already. I think I confused myself as to what the problem was.

> For the Android test I had to add `enableFlutterDriverExtension();` in the example of the webview_flutter_android otherwise the test was failing with the error

@e-adrien Which test specifically had this issue? The new native Java test? If so, that test shouldn't actually care about the Dart integration tests (it just runs them as an accidental side effect, IIRC).

stuartmorgan (changes requested):

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m
```diff
@@ -95,6 +95,20 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+    NSNumber* backgroundColorNSNumber = args[@"backgroundColor"];
+    if ([backgroundColorNSNumber isKindOfClass:[NSNumber class]]) {
+      int backgroundColorInt = [backgroundColorNSNumber intValue];
+      UIColor* backgroundColor = [UIColor colorWithRed:(backgroundColorInt >> 16 & 0xff) / 255.f
+                                                 green:(backgroundColorInt >> 8 & 0xff) / 255.f
+                                                  blue:(backgroundColorInt & 0xff) / 255.f
+                                                 alpha:(backgroundColorInt >> 24 & 0xff) / 255.f];
+      _webView.opaque = NO;
+      _webView.backgroundColor = backgroundColor;
+      _webView.scrollView.backgroundColor = backgroundColor;
```
stuartmorgan, line 109: I'm still confused as to how this works for semi-transparent colors. Could you add a comment here explaining why setting both to the same color is correct?

stuartmorgan:
> the bigger issue is that the iOS test is failing in CI. Does it pass for you locally?

This doesn't actually seem to have been addressed; the new tests are failing in CI. See the two failing `ios-platform_tests` tasks for details.

e-adrien:
> Except this plugin is using integration_test already. I think I confused myself as to what the problem was.
> 
> > For the Android test I had to add `enableFlutterDriverExtension();` in the example of the webview_flutter_android otherwise the test was failing with the error
> 
> @e-adrien Which test specifically had this issue? The new native Java test? If so, that test shouldn't actually care about the Dart integration tests (it just runs them as an accidental side effect, IIRC).

Yes it's the native one. I run it with this command :
dart run ./script/tool/bin/flutter_plugin_tools.dart native-test --android --no-unit --packages webview_flutter

e-adrien:
I made the change.
The test runs fine on my laptop but I can't find its execution results in Cirrus CI logs. Did I miss something ?

stuartmorgan:
> I made the change. The test runs fine on my laptop but I can't find its execution results in Cirrus CI logs. Did I miss something ?

We can't currently run the Android integration tests for PRs from non-contributors in presubmit, unfortunately.

It sounds like this is ready for another review round then? We're currently doing a significant restructuring of the webview plugin implementation (moving the platform channel boundary much closer to the OS API), so I'll hold off for the moment, but once that's complete we can move forward with this. Since the tests here are integration tests, I don't expect there to actually be much impact on the PR, just some adjustment to the `_android` and `_wkwebview` implementation details.

e-adrien:
> We can't currently run the Android integration tests for PRs from non-contributors in presubmit, unfortunately.

Okay, no problem. It was just to be sure that I didn't do something wrong.

> It sounds like this is ready for another review round then? We're currently doing a significant restructuring of the webview plugin implementation (moving the platform channel boundary much closer to the OS API), so I'll hold off for the moment, but once that's complete we can move forward with this. Since the tests here are integration tests, I don't expect there to actually be much impact on the PR, just some adjustment to the `_android` and `_wkwebview` implementation details.

No problem. I'll check this PR once a week to see if it needs rebasing until we can move forward.

stuartmorgan:
I pushed a merge that should update the PR for the new Dart-based Android implementation (feel free to replace with a rebase instead if that's your preferred workflow).

stuartmorgan (commented):
I think this is pretty much good to go, other than needing a test of the change to the method channel in the platform interface package, and of the background color parameter in the app-facing package. Both should be simple Dart tests that the values are passed through to the next layer as expected.

If you split out the platform interface change now (with that test added), and we can start getting this incrementally landed.

stuartmorgan:
FYI the analyzer issue you are hitting is unrelated to your change; I just landed a fix for it.

stuartmorgan:
The individual components have all landed (the Android implementation will auto-publish once the post-submit tests finish), so this can now be reduced down to just the `webview_flutter` package changes, and a simple test added [here](https://github.com/flutter/plugins/blob/master/packages/webview_flutter/webview_flutter/test/webview_flutter_test.dart), and it should be good to go!

e-adrien:
> The individual components have all landed (the Android implementation will auto-publish once the post-submit tests finish), so this can now be reduced down to just the `webview_flutter` package changes, and a simple test added [here](https://github.com/flutter/plugins/blob/master/packages/webview_flutter/webview_flutter/test/webview_flutter_test.dart), and it should be good to go!

Thanks. I will open a PR when the android package will be published.

stuartmorgan:
There are infrastructure issues that caused https://github.com/flutter/plugins/pull/4584 not to land correctly, and are currently preventing the creation of **new** PRs that land correctly; while we get that resolved I've updated this PR to match https://github.com/flutter/plugins/pull/4584 so that we get it landed.

stuartmorgan (approved):

PR #3439, opened 2021-01-17
[image_picker_for_web] Added image resize functionality

jesusrp98:
Added image resize functionality to the web implementation of `image_picker` to make it par with the mobile implementation of the plugin.

This implementation comes from @postacik, detailed [here](https://github.com/flutter/plugins/pull/2802#issuecomment-760216207) on #2802.
Huge thanks to @ditman as well for all the help!

This **SHOULD NOT** be a breaking change.

*Also this is my first PR here so bare with me :)*

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
+++ b/packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
@@ -37,9 +37,63 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     double maxHeight,
     int imageQuality,
     CameraDevice preferredCameraDevice = CameraDevice.rear,
-  }) {
+  }) async {
     String capture = computeCaptureAttribute(source, preferredCameraDevice);
-    return pickFile(accept: _kAcceptImageMimeType, capture: capture);
+    return pickFile(
+      accept: _kAcceptImageMimeType,
+      capture: await _resizeImage(
+        capture,
+        maxWidth.toInt(),
+        maxHeight.toInt(),
+        imageQuality,
+      ),
+    );
+  }
+
+  static Future<String> _resizeImage(
+    String src,
+    int maxWidth,
+    int maxHeight,
+    int imageQuality,
+  ) {
+    final completer = Completer<String>();
+    final img = html.ImageElement();
+    img.onError.listen((event) {
+      completer.complete("");
+    });
+    img.onLoad.listen((event) {
+      final canvas = html.CanvasElement();
+      final ctx = canvas.context2D;
+      double ratio = 1;
+
+      if (img.width > img.height) {
+        if (img.width > maxWidth) ratio = maxWidth / img.width;
+      } else if (img.height > maxWidth) {
+        ratio = maxWidth / img.height;
+      }
+      if (img.width > img.height) {
+        if (img.width < maxHeight) ratio = maxHeight / img.width;
+      } else if (img.height < maxHeight) {
+        ratio = maxHeight / img.height;
+      }
+
+      canvas.height = (img.height * ratio).floor();
+      canvas.width = (img.width * ratio).floor();
+      // Draw the image to canvas and resize
+      ctx.drawImageScaled(img, 0, 0, canvas.width, canvas.height);
+      final base64 = canvas.toDataUrl("image/png", imageQuality / 100);
+      completer.complete(base64);
+    });
+    img.src = src;
+    // make sure the load event fires for cached images too
+    if (img.complete) {
+      // Flush cache
+      img.src =
+          'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///ywAAAAAAQABAAACAUwAOw==';
+      // Try again
+      img.src = src;
+    }
+    return completer.future;
   }
 
   @override


jesusrp98:
@googlebot I signed it!

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -37,9 +37,63 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
-    return pickFile(accept: _kAcceptImageMimeType, capture: capture);
+    return pickFile(
+      accept: _kAcceptImageMimeType,
+      capture: await _resizeImage(
+        capture,
+        maxWidth.toInt(),
+        maxHeight.toInt(),
+        imageQuality,
+      ),
+    );
```
ditman, line 50: I don't think this'll work. A couple of comments:

* Only apply the resizing if the maxWidth and maxHeight (and possibly imageQuality) are passed in (otherwise do we end up with 0x0 images maybe?)
* This is replacing the value of the "capture" attribute, which is not what we want. Capture does [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#htmlattrdefcapture).

I think the correct flow for this would be to let `pickFile` do its thing, then you'd have a `src` to pass to your _resizeImage method (the `path` of the returned `PickedFile`).

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -37,9 +37,63 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
+  }
+
+  static Future<String> _resizeImage(
```
ditman, line 53: I think this should return a `Future<PickedFile>`, with a PickedFile that contains the resized image. You can create a PickedFile from its bytes.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -37,9 +37,63 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
+    String src,
+    int maxWidth,
+    int maxHeight,
+    int imageQuality,
+  ) {
+    final completer = Completer<String>();
+    final img = html.ImageElement();
```
ditman, line 60: Don't you need to add this `img` to the DOM for the events to fire?

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -37,9 +37,63 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
+      final ctx = canvas.context2D;
+      double ratio = 1;
+
+      if (img.width > img.height) {
+        if (img.width > maxWidth) ratio = maxWidth / img.width;
+      } else if (img.height > maxWidth) {
+        ratio = maxWidth / img.height;
+      }
+      if (img.width > img.height) {
+        if (img.width < maxHeight) ratio = maxHeight / img.width;
```
ditman, line 75: This is odd, comparing a width to a height. I think there's some typos in this algo to determine the aspect ratio / scale factor (?)

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -37,9 +37,63 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
+      ctx.drawImageScaled(img, 0, 0, canvas.width, canvas.height);
+      final base64 = canvas.toDataUrl("image/png", imageQuality / 100);
+      completer.complete(base64);
+    });
+    img.src = src;
+    // make sure the load event fires for cached images too
+    if (img.complete) {
+      // Flush cache
+      img.src =
+          'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///ywAAAAAAQABAAACAUwAOw==';
```
ditman, line 92: Will this trigger the onLoad event above? If so, you might want to check for this special image to do nothing with it; otherwise you'll attempt to complete the `completer` several times, and it'll throw an exception (you can only complete a completer once)

ditman (commented):
This needs some tests ([here](https://github.com/flutter/plugins/tree/master/packages/image_picker/image_picker_for_web/test)), also, I made a couple of comments in the general approach.

Have you verified this works with the `example` app of the plugin in web?

jesusrp98:
Sorry @ditman. The first commit was just copying and pasting the resize function to the code base. I'll continue with the work now and implement all your feedback. I'm really thankful for that because I haven't really tinkered with flutter web plugins before :)

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
jesusrp98: OK I understand the logic here and what is doing what.
Thank you very much for the explanation and how to procede here :)
This change has now been commited.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
jesusrp98: I'm haven't had many experience of developing flutter web plugins and I don't know how to proceed here.
Could you elaborate on this @ditman please?

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
jesusrp98: I'll take a look a the proper algorithm here, maybe I can refine this.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
jesusrp98: Yes it's a bit weird here and I don't really understand what's going on.
Maybe @postacik could help us here (?).

ditman:
Haha thanks for the clarification @jesusrp98! I didn't realize this was a Draft PR :facepalm: :P

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
ditman: Hm, some times for some DOM elements to fire their load/error events they must be injected into the markup of the page (anywhere, but they need to be injected).

I'm not sure this is the case, but I'd test this code across all main browsers (Edge/Safari/Firefox/Chrome) to ensure that the on*** events of the ImageElement are firing as expected.

Same approach for the comment below where we're resetting the image to something blank, and then the desired source again :)

postacik: @ditman , you are right, events are generally fired if the element is added to DOM tree but while converting my JavaScript code to Dart, I realized that it worked without adding it to DOM for Chrome. I haven't tested for other mentioned browsers.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
postacik: @ditman, you are right, if it does trigger the onload event several times, it would cause exceptions. The JavaScript code I used was not using a Promise (or Completer) but simply copying the result to the canvas so the hack to avoid cases where onload did not fire did not have any side effects but here it would. The base64 image is a 1x1 gif. Maybe the completer may be avoided if the loaded image size is 1x1.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
postacik: @ditman , the scaling algorithm was just a copy paste from an old JavaScript project. I really don't recall how I constructed it or tested it thoroughly. 

The best algorithm for the resize function would be using the algorithm used in the mobile counter part. Maybe the algorithm may be copied from the following Java file so that it will be consistent with the mobile version.

https://github.com/flutter/plugins/blob/6783cd5ebd54e034a77e2404f5565b48bd19e121/packages/image_picker/image_picker/android/src/main/java/io/flutter/plugins/imagepicker/ImageResizer.java#L56

jesusrp98: I agree with you @postacik. I'll update the current algorith to resemble something similar to the Android implementation.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
jesusrp98: Added a simple `img.width > 1 && img.height > 1` check before performing the resize logic.
Give me your feedback here!

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
jesusrp98: OK I understand what you're saying here, but I don't know how to exactly add this element to the DOM tree :/

I'd need some help here @postacik @ditman .

PR #3412, opened 2021-01-13
[google_maps_flutter_web] Support for Holes in Polygons

ABausG:
## Description

@sanekyy added support for Holes in Polygons in `google_maps_flutter_platform_interface` in #3135 and is adding the support for iOS and Android to `google_maps_flutter` with #1721 

## Pre-launch Checklist

- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

PR #3440, opened 2021-01-18
[google_maps_flutter_web] Hole Direction in Polygons

ABausG:
As observed by @ditman #3412 showed that the Web Version needs the holes in a polygon to be opposite direction to the parent polygon.

This PullRequests reverts the holes in the example to show the expected result as well as printing a warning if a hole is the same direction as the the polygon based on [this](https://stackoverflow.com/a/1165943)

Closes https://github.com/flutter/flutter/issues/74096

## Pre-launch Checklist

- [x] I signed the [CLA].


ABausG:
@ditman I fixed the issue you found.
I tried to add a test for this however I did not find a way to check for a specific console output. COuld you maybe lead me in the direction I should look at?

Also I observed that while the hole is filled it still registers as a valid hole and [containsLocation](https://pub.dev/documentation/google_maps/latest/google_maps_geometry/Poly/containsLocation.html) returns false.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -482,9 +482,15 @@ gmaps.PolygonOptions _polygonOptionsFromPolygon(
...
   });
+  final polygonDirection = _isPolygonClockwise(path);
   List<List<gmaps.LatLng>> paths = [path];
   polygon.holes?.forEach((hole) {
-    paths.add(hole.map((point) => _latLngToGmLatLng(point)).toList());
+    final holePath = hole.map((point) => _latLngToGmLatLng(point)).toList();
+    if (_isPolygonClockwise(holePath) == polygonDirection) {
+      print(
+          'Holepath is the same direction as polygon. In order to be displayed properly you should reverse the hole Path. More Info: https://developers.google.com/maps/documentation/javascript/examples/polygon-hole');
+    }
```
ditman, line 492: You might be able to replace this by an `assert(_isPolygonClockwise(holePath) == _isPolygonClockwise(path), 'The path of the hole should be reversed. See: blah');`. IIRC this is catchable in tests and will nag users (in development mode) for them to use the correct polygon winding.

Another thing we can do is just `.revert` the holePath before using it, and log the error. Then in the test you can just verify that the hole has been automatically reversed. I kind of like this second idea better, this brings the web to feature parity with the mobile version!

ABausG: I changed it so the holepath gets reversed if the direction matches and added a test for it.
Also reverted the example to the original state

ditman (approved):
This is great. Thank you very much for the quick fix, and the tests!

ditman:
@googlebot I consent.

ditman:
(I just added a small change to log the error only in debug mode, also bumped the version, which I should have done in the version where we added the Holes feature :))

ditman:
This has been uploaded to pub.dev:

https://pub.dev/packages/google_maps_flutter_web/versions/0.1.1

PR #3444, opened 2021-01-20
[camera] Ensure that channel.invokeMethod runs on the main thread

jsroest:
channel.invokeMethod may only be called from the main thread. This is in the camera plugin not always the case, which causes a crash of the application. This change ensures that all calls to channel.invokeMethod are done from the main thread.

Issue fixed by this PR: [camera] [crash] Methods marked with @UiThread must be executed on the main thread #72340
https://github.com/flutter/flutter/issues/72340

### DIFF
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.7.0+1
 
-* Fix issue: [camera] [crash] Methods marked with @UiThread must be executed on the main thread #72340
+* Ensure communication from JAVA to Dart is done on the main UI thread.
 
 ## 0.7.0
 
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.0+1
+
+* Fix issue: [camera] [crash] Methods marked with @UiThread must be executed on the main thread #72340
+
 ## 0.7.0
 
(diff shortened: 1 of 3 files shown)


mvanbeusekom (commented):
@jsroest thank you for this pull request. The fix looks good, however could you also update the version in the `pubspec.yaml` file and add an entry to the `CHANGELOG.md`?

I also noticed the build is failing on a formatting error, could you ensure the code is formatted properly? You can use the `flutter_plugin_tools` to auto format the Dart, Java and Objective C code for you, instructions can be found here: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#contributing-code

jsroest:
@mvanbeusekom Thank you for your comments. I've fixed it. Can you review the PR again?

packages/camera/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.7.0+1
+
+* Fix issue: [camera] [crash] Methods marked with @UiThread must be executed on the main thread #72340
```
mvanbeusekom, line 3: nit: It would be nice if we can update this to form a correct sentence. Maybe something like this:
```suggestion
* Ensure communication from JAVA to Dart is done on the main UI thread.
```

mvanbeusekom (commented):
Thanks for going through the feedback and making the suggested changes. The code looks good, I have just one more little request regarding the entry in the [CHANGELOG]. Would be great if you could have a look at it.

packages/camera/camera/CHANGELOG.md
jsroest: Done as you requested

mvanbeusekom (approved):
LGTM

PR #3451, opened 2021-01-22
[file_selector_platform_interface]: Verify that extensions don't have leading dots.

tugorez:
Make sure extensions doesn't have a leading dot.

Fixes flutter/flutter#73959.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
+++ b/packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
@@ -41,4 +41,11 @@ class XTypeGroup {
       'webWildCards': webWildCards,
     };
   }
+
+  bool operator ==(o) =>
+      label == o.label &&
+      listEquals(extensions, o.extensions) &&
+      listEquals(mimeTypes, o.mimeTypes) &&
+      listEquals(macUTIs, o.macUTIs) &&
+      listEquals(webWildCards, o.webWildCards);
 }
--- a/packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
+++ b/packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
@@ -1,7 +1,6 @@
 // Copyright 2020 The Flutter Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
-import 'package:flutter/foundation.dart';
 
 /// A set of allowed XTypes
 class XTypeGroup {
@@ -11,27 +10,11 @@ class XTypeGroup {
   /// allowed.
   XTypeGroup({
     this.label,
-    this.extensions,
+    List<String>? extensions,
     this.mimeTypes,
     this.macUTIs,
     this.webWildCards,
-  }) {
-    _verifyExtensions();
-  }
-
-  void _verifyExtensions() {
-    if (extensions == null) return;
-    final exts = extensions!;
-    for (var i = 0; i < exts.length; i++) {
-      if (!exts[i].startsWith('.')) continue;
-      if (kDebugMode) {
-        print('extensions[${i}] with value "${exts[i]}" is invalid.'
-            ' The leading dots are being removed from the extensions'
-            ' Please fix it.');
-      }
-      exts[i] = exts[i].substring(1);
-    }
-  }
+  }) : this.extensions = _removeLeadingDots(extensions);
 
   /// The 'name' or reference to this group of types
   final String? label;
@@ -58,4 +41,7 @@ class XTypeGroup {
       'webWildCards': webWildCards,
     };
   }
+
+  static List<String>? _removeLeadingDots(List<String>? exts) =>
+      exts?.map((ext) => ext.startsWith('.') ? ext.substring(1) : ext).toList();
 }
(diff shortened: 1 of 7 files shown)


tugorez:
Fixes https://github.com/flutter/flutter/issues/73959.

tugorez:
cc @ditman

packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
```diff
@@ -14,7 +15,22 @@ class XTypeGroup {
     this.mimeTypes,
     this.macUTIs,
     this.webWildCards,
-  });
+  }) {
+    _verifyExtensions();
+  }
+
+  void _verifyExtensions() {
+    if (extensions == null) return;
```
stuartmorgan, line 23: Could you use a local variable for extensions, mutate that, then re-assign at the end? That way you won't need `!` everywhere.

packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
```diff
@@ -14,7 +15,22 @@ class XTypeGroup {
...
+    for (var i = 0; i < extensions!.length; i++) {
+      if (!extensions![i].startsWith('.')) continue;
+      if (kDebugMode) {
+        print('extensions[${i}] with value "${extensions![i]}" is invalid.'
+            ' We are mutating the extensions list to remove the leading dot.'
```
stuartmorgan, line 28: "We" is very ambiguous here. How about "The leading dots are being removed from the extensions".

stuartmorgan (changes requested):
Sorry for the slow review. Minor nits, then it should be good to go.

packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
tugorez: Great catch. Done. PTAL

packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
tugorez: Agree. Done. PTAL

packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
```diff
@@ -14,7 +15,23 @@ class XTypeGroup {
...
+
+  void _verifyExtensions() {
+    if (extensions == null) return;
+    final exts = extensions!;
+    for (var i = 0; i < exts.length; i++) {
+      if (!exts[i].startsWith('.')) continue;
+      if (kDebugMode) {
+        print('extensions[${i}] with value "${exts[i]}" is invalid.'
+            ' The leading dots are being removed from the extensions'
+            ' Please fix it.');
```
stuartmorgan, line 30: Sorry, one more nit. Looking at this again, I don't think there's really any value in telling the developer, but also fixing it; if we're going to do runtime fixup, then it's not really invalid. We should either assert (and clearly document the requirement), or just silently fix it (which I think is the better option from a developer-friendliness standpoint, since there's no ambiguity in the intent here).

So let's just remove this whole print.

stuartmorgan (changes requested):

packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
tugorez: > or just silently fix it (which I think is the better option from a developer-friendliness standpoint, since there's no ambiguity in the intent here).

The only thing that concerns me is that we are mutating the extensions list. This mutation could lead to unexpected errors + hard debugging time if we don't inform them.

```dart
final extensions = await api.getExtensions();
final files = openFiles(...);
// By this time, extensions will be muted
doSomethingElseWithExtensions(extensions);
```

stuartmorgan: Good point; we should copy the list (and document the behavior):
```

  /// Creates a new group with the given label and file extensions.
  ///
  /// A group with none of the type options provided indicates that any type is
  /// allowed.
  /// 
  /// [extensions] are normalized to a form without a leading '.'.
  XTypeGroup({
    this.label,
    List<String>? extensions,
    this.mimeTypes,
    this.macUTIs,
    this.webWildCards,
  }) : this.extensions =
            extensions == null ? null : List<String>.from(extensions);
```

packages/file_selector/file_selector_platform_interface/test/x_type_group_test.dart
```diff
@@ -41,5 +41,13 @@ void main() {
       expect(jsonMap['macUTIs'], null);
       expect(jsonMap['webWildCards'], null);
     });
+
+    test('Validates extensions have not leading dots', () {
```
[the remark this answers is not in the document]
tugorez: Done. PTAL

packages/file_selector/file_selector_platform_interface/lib/src/types/x_type_group/x_type_group.dart
tugorez: Done. PTAL

stuartmorgan (approved):
LGTM!

ditman:
Published!

PR #3462, opened 2021-01-25
Support breaking http change

grouma:
The next version of `package:http` expects `URI`s. See https://github.com/dart-lang/http/pull/507

### DIFF
--- a/packages/cross_file/CHANGELOG.md
+++ b/packages/cross_file/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.2.1
+
+* Prepare for breaking `package:http` change. 
+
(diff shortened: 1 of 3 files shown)


natebosch (approved):

ditman:
This is coming:

https://github.com/flutter/plugins/pull/3452

It doesn't use `package:http`.

I'll merge the nnbd version, it should be API compatible with whatever is imported now :)

(If I don't have time to merge the nnbd version, I'll land this one as an emergency fix, and rebase mine on top of this)

ditman (approved):
LGTM, thanks for sending the fix!

PR #3466, opened 2021-01-26
Migrate shared_preferences_platform_interfaces to null safety

bparrishMines:
Migrate to null safety.

## Pre-launch Checklist

- [] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [x] I signed the [CLA].

### DIFF
--- a/packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
+++ b/packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
@@ -40,7 +40,10 @@ class MethodChannelSharedPreferencesStore
 
   @override
   Future<Map<String, Object>> getAll() async {
-    return await _kChannel.invokeMapMethod<String, Object>('getAll')
-        as Map<String, Object>;
+    final Map<String, Object>? preferences =
+        await _kChannel.invokeMapMethod<String, Object>('getAll');
+
+    if (preferences == null) return <String, Object>{};
+    return preferences;
   }
 }
--- a/packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
+++ b/packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
@@ -19,23 +19,23 @@ class MethodChannelSharedPreferencesStore
     extends SharedPreferencesStorePlatform {
   @override
   Future<bool> remove(String key) async {
-    return await _kChannel.invokeMethod<bool>(
+    return (await _kChannel.invokeMethod<bool>(
       'remove',
       <String, dynamic>{'key': key},
-    ) as bool;
+    ))!;
   }
 
   @override
   Future<bool> setValue(String valueType, String key, Object value) async {
-    return await _kChannel.invokeMethod<bool>(
+    return (await _kChannel.invokeMethod<bool>(
       'set$valueType',
       <String, dynamic>{'key': key, 'value': value},
-    ) as bool;
+    ))!;
   }
 
   @override
   Future<bool> clear() async {
-    return await _kChannel.invokeMethod<bool>('clear') as bool;
+    return (await _kChannel.invokeMethod<bool>('clear'))!;
   }
 
   @override
--- a/packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
+++ b/packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
@@ -18,39 +18,29 @@ const MethodChannel _kChannel =
 class MethodChannelSharedPreferencesStore
     extends SharedPreferencesStorePlatform {
   @override
-  Future<bool> remove(String key) {
-    return _invokeBoolMethod('remove', <String, dynamic>{
-      'key': key,
-    });
+  Future<bool> remove(String key) async {
+    return await _kChannel.invokeMethod<bool>(
+      'remove',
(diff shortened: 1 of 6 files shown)


packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
```diff
@@ -18,39 +18,29 @@ const MethodChannel _kChannel =
...
+  Future<bool> clear() async {
+    return await _kChannel.invokeMethod<bool>('clear') as bool;
   }
 
   @override
-  Future<Map<String, Object>> getAll() {
-    return _kChannel.invokeMapMethod<String, Object>('getAll');
+  Future<Map<String, Object>> getAll() async {
+    return await _kChannel.invokeMapMethod<String, Object>('getAll')
+        as Map<String, Object>;
```
stuartmorgan, line 44: Why do we need the cast? I thought the purpose of `invokeMapMethod` was do to that cast internally.

packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
```diff
@@ -18,39 +18,29 @@ const MethodChannel _kChannel =
...
-  Future<bool> clear() {
-    return _kChannel.invokeMethod<bool>('clear');
+  Future<bool> clear() async {
+    return await _kChannel.invokeMethod<bool>('clear') as bool;
   }
 
   @override
-  Future<Map<String, Object>> getAll() {
-    return _kChannel.invokeMapMethod<String, Object>('getAll');
+  Future<Map<String, Object>> getAll() async {
```
stuartmorgan, line 42: Are we certain that no platform implementation ever returns null here? It might be a good idea to bullet-proof this with a null check on the results, returning an empty map if the channel's result was null.

packages/shared_preferences/shared_preferences_platform_interface/test/method_channel_shared_preferences_test.dart
```diff
@@ -44,9 +44,9 @@ void main() {
           return await testData.clear();
         }
         final RegExp setterRegExp = RegExp(r'set(.*)');
-        final Match match = setterRegExp.matchAsPrefix(methodCall.method);
-        if (match.groupCount == 1) {
-          final String valueType = match.group(1);
+        final Match? match = setterRegExp.matchAsPrefix(methodCall.method);
+        if (match != null && match.groupCount == 1) {
```
stuartmorgan, line 48: Nit: couldn't this be `if (match?.groupCount == 1)` (since null != 1)?

stuartmorgan (approved):
LGTM with some optional nits.

packages/shared_preferences/shared_preferences_platform_interface/test/method_channel_shared_preferences_test.dart
bparrishMines: Ah, good suggestion. done

packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
bparrishMines: All the invoke methods return the nullable version of the type parameters. As in `invokeMapMethod<String, Object>` returns `Map<String, Object>?`.

packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
bparrishMines: done

packages/shared_preferences/shared_preferences_platform_interface/lib/method_channel_shared_preferences.dart
stuartmorgan: Wouldn't that just need a `!` though, rather than a cast?

bparrishMines: Yes that would work but you would have to add parentheses around the await keyword. Basically the two options are

```dart
final Map<String, Object> map = (await invokeMapMethod<String, Object>())!;
```

or 

```dart
final Map<String, Object> map = await invokeMapMethod<String, Object>() as Map<String, Object>;
```

I don't really have a preference on which is better. I think the top one is shorter, but the second is easier to read/cleaner.

stuartmorgan: I'd vote strongly for the top one, for the same reason that people prefer C++-style casts to C-style casts: it's explicit about what it's doing. `!` does exactly one thing (remove the nullability from the type), while `as` could do other kinds of casting. The fact that I didn't know what the cast was for in the second form seems like pretty good evidence that it has readability issues.

Plus, if there were a mistake with the types here (say, a mismatch in the contents of the `<>` in invokeMapMethod vs the cast), the `as` would be a runtime failure, while the `!` would be a compile-time failure.

packages/shared_preferences/shared_preferences_platform_interface/test/method_channel_shared_preferences_test.dart
MrBirb: But we are already checking if match != null in the first condition. I believe it won’t evaluate the second statement if match is null

PR #3473, opened 2021-01-28
[local_auth] Fix incorrect switch fallthrough

jiahaog:
This breaks internal lints when imported, because the switch case continues to the next case. I'm not familiar with the code, but I assume that since the error is handled in `completionHandler.onError` this `case` is complete and we should `break`,


jiahaog:
@cyanglaz can you please take a look at this?

cyanglaz (approved):
LGTM, please update version and CHANGELOG

jiahaog:
Done, will land when the checks are green. Thanks!

PR #3478, opened 2021-01-28
[package_info] Register IntegrationTestPlugin in the example app.

ditman:
This PR registers the IntegrationTestPlugin in the example app, so test results are correctly reported back to FTL.

The end-to-end firebase tests for package_info haven't passed in a while (nor have been reported as broken before), but after this change, they start passing. See:

https://firebase.corp.google.com/project/flutter-infra/testlab/histories/bh.fd85a82ead1f4de

Screen Shot 2021-01-28 at 4 21 07 PM

---

Fixes https://github.com/flutter/flutter/issues/74944

PR #3479, opened 2021-01-28
[ci] skip package_info tests on firebase test lab

cyanglaz:
Skipping the package_info from firebase test lab command. 

See also: https://github.com/flutter/flutter/issues/74944


cyanglaz:
Fixed in https://github.com/flutter/plugins/pull/3478

PR #3480, opened 2021-01-29
[integration_test] Fix tests from changes to `flutter test` machine output

jiahaog:
There was a change in `flutter master` on how machine readable output is reported (https://github.com/flutter/flutter/pull/74176). It required a change to integration_test in the SDK, but that wasn't patched to the plugins repo. This PR does that.

### DIFF
--- a/packages/integration_test/test/binding_fail_test.dart
+++ b/packages/integration_test/test/binding_fail_test.dart
@@ -61,12 +61,17 @@ Future<Map<String, dynamic>> _runTest(String scriptPath) async {
   final String testResults = (await process.stdout
           .transform(utf8.decoder)
           .expand((String text) => text.split('\n'))
-          .map((String line) {
+          .map<dynamic>((String line) {
             try {
               return jsonDecode(line);
             } on FormatException {
               // Only interested in test events which are JSON.
             }
+          })
+          .expand<Map<String, dynamic>>((dynamic json) {
+            return json is List<dynamic>
+                ? json.cast()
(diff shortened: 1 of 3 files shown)


packages/integration_test/test/binding_fail_test.dart
```diff
@@ -61,13 +61,18 @@ Future<Map<String, dynamic>> _runTest(String scriptPath) async {
   final String testResults = (await process.stdout
           .transform(utf8.decoder)
           .expand((String text) => text.split('\n'))
-          .map((String line) {
+          .map<dynamic>((String line) {
```
ditman, line 64: Not sure you need this `<dynamic>` here? But if no linter complains, let's gooo!

ditman (approved):
Can you please bump the changelog and version of the integration_test package?

ditman:
(This will have to be merged in red to fix the tree... FOR REAL THIS TIME!)

packages/integration_test/test/binding_fail_test.dart
jiahaog: Yeah I think since this is copied from the SDK repo, where type annotations are necessary

ditman:
All is green, merging on red, to hopefully make the submit-queue green!

ditman:
This indeed fixed the tree, and was [published to pub](https://pub.dev/packages/integration_test/changelog#1022) last night. Thanks again @jiahaog !!

PR #3485, opened 2021-01-30
[path_provider_windows] Resolve FFI stabilization changes

timsneath:
Fixes flutter/flutter#75050. When ffi 0.2.0-nullsafety.2 lands, the `calloc` method will be provided by ffi itself, which might trigger another set of changes, but this change will ensure that path_provider_windows works on current null safety builds. 

## Pre-launch Checklist

- [X] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [X] I updated CHANGELOG.md to add a description of the change.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I signed the [CLA].
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

PR #3488, opened 2021-02-01
Revert "[path_provider_windows] Resolve FFI stabilization changes"

xu-baolin:
Reverts flutter/plugins#3485 to fix the tree.
```dart
Running command: "pub global run tuneup check" in /b/s/w/ir/tmp/t/pluginsFHFRRF/packages/path_provider/path_provider_windows
Checking project path_provider_windows...

  error - The method 'setGUID' isn't defined for the type 'Pointer' at lib/src/path_provider_windows_real.dart:119:57 - (undefined_method)

1 issue found; analyzed 11 source files in 4.0s.
```

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket.appspot.com/8856572090238051232/+/steps/run_test.dart_for_flutter_plugins_shard_and_subshard_analyze/0/stdout

CC @timsneath @stuartmorgan

### DIFF
--- a/packages/path_provider/path_provider_windows/CHANGELOG.md
+++ b/packages/path_provider/path_provider_windows/CHANGELOG.md
@@ -1,7 +1,3 @@
-## 0.1.0-nullsafety.1
-
-* Bump win32 dependency to latest version.
-
 ## 0.1.0-nullsafety
 
 * Migrate to null safety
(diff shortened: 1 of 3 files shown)


stuartmorgan:
I'm confused; the revert *introduces* the error you say you are fixing.

And from your build log:
> `! path_provider_windows 0.1.0-nullsafety from path ..`
That's the version from before the patch you are trying to revert.

stuartmorgan:
If you are seeing this error in current builds, the most likely explanation is a pub cache on the machine in question that has the older version, in which case the fix is to update the minimum `path_provider_windows` constraint on `path_provider`, and then update the minimum version of `path_provider`.

PR #3491, opened 2021-02-01
Remove Dart stubs from macOS plugins

stuartmorgan:
When these federated plugins were created, plugins had to have at least
one Dart file to avoid issues with the analyzer, so were created with a
stub file since all the code is native. The analyzer no longer has this
limitation, so the stub is no longer necesssary.


cyanglaz (approved):
LGTM assuming CI passes.

PR #3497, opened 2021-02-02
[camera_platform_interface] Migrate to null safety

mvanbeusekom:
This PR migrates the `camera_platform_interface` package to null safety.

flutter/flutter#75228
flutter/flutter#75247

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_platform_interface/CHANGELOG.md
+++ b/packages/camera/camera_platform_interface/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.0-nullsafety.0
+
+- Migrate to null safety.
+
 ## 1.6.0
 
 - Added VideoRecordedEvent to support ending a video recording in the native implementation. 
--- a/packages/camera/camera_platform_interface/CHANGELOG.md
+++ b/packages/camera/camera_platform_interface/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.0.0-nullsafety.0
+## 2.0.0-nullsafety
 
 - Migrate to null safety.
 
--- a/packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
+++ b/packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
@@ -71,9 +71,9 @@ class CameraInitializedEvent extends CameraEvent {
     int cameraId,
     this.previewWidth,
     this.previewHeight, [
-    this.exposureMode,
+    this.exposureMode = ExposureMode.auto,
     this.exposurePointSupported = false,
-    this.focusMode,
+    this.focusMode = FocusMode.auto,
     this.focusPointSupported = false,
   ]) : super(cameraId);
 
--- a/packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
+++ b/packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
@@ -242,7 +242,7 @@ class VideoRecordedEvent extends CameraEvent {
   final XFile file;
 
   /// Maximum duration of the recorded video.
-  final Duration maxVideoDuration;
+  final Duration? maxVideoDuration;
 
   /// Build a VideoRecordedEvent triggered from the camera with the `cameraId`.
   ///
--- a/packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
+++ b/packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
@@ -70,12 +70,12 @@ class CameraInitializedEvent extends CameraEvent {
   CameraInitializedEvent(
     int cameraId,
     this.previewWidth,
-    this.previewHeight, [
-    this.exposureMode = ExposureMode.auto,
-    this.exposurePointSupported = false,
-    this.focusMode = FocusMode.auto,
-    this.focusPointSupported = false,
-  ]) : super(cameraId);
+    this.previewHeight,
+    this.exposureMode,
+    this.exposurePointSupported,
+    this.focusMode,
+    this.focusPointSupported,
+  ) : super(cameraId);
 
   /// Converts the supplied [Map] to an instance of the [CameraInitializedEvent]
   /// class.
--- a/packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart
+++ b/packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart
@@ -11,6 +11,7 @@ import 'package:camera_platform_interface/src/types/focus_mode.dart';
 import 'package:camera_platform_interface/src/types/image_format_group.dart';
 import 'package:camera_platform_interface/src/utils/utils.dart';
 import 'package:cross_file/cross_file.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 import 'package:meta/meta.dart';
@@ -58,8 +59,13 @@ class MethodChannelCamera extends CameraPlatform {
   @override
   Future<List<CameraDescription>> availableCameras() async {
     try {
-      final List<Map<dynamic, dynamic>> cameras = await _channel
+      final cameras = await _channel
           .invokeListMethod<Map<dynamic, dynamic>>('availableCameras');
+
+      if (cameras == null) {
+        return <CameraDescription>[];
+      }
+
       return cameras.map((Map<dynamic, dynamic> camera) {
         return CameraDescription(
           name: camera['name'],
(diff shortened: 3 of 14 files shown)


packages/camera/camera_platform_interface/test/method_channel/method_channel_camera_test.dart
```diff
@@ -352,8 +375,8 @@ void main() {
     });
 
     group('Function Tests', () {
-      MethodChannelCamera camera;
-      int cameraId;
+      late MethodChannelCamera camera;
+      int? cameraId;
```
bparrishMines, line 379: I think you should make `cameraId` non-nullable since `camera.createCamera` sets it in `setUp`. This way you don't have to null check `cameraId` in every test.

mvanbeusekom: Good catch, I have updated this to `late int cameraId;` and removed all the unnecessary null checks.

packages/camera/camera_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.0-nullsafety.0
```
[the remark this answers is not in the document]
mvanbeusekom: Good to know, I noticed some different strategies and picked the wrong one ;). Updated to remove the .0.

packages/camera/camera_platform_interface/lib/src/events/camera_event.dart
```diff
@@ -71,9 +71,9 @@ class CameraInitializedEvent extends CameraEvent {
     int cameraId,
     this.previewWidth,
     this.previewHeight, [
-    this.exposureMode,
+    this.exposureMode = ExposureMode.auto,
```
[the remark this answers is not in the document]
mvanbeusekom: You are right indeed, this constructor is only used internally (when information from the native side arrives) so we indeed should enforce all parameters to be set. Updated as per your recommendation.

packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart
```diff
@@ -131,7 +137,7 @@ class MethodChannelCamera extends CameraPlatform {
     );
 
     if (_channels.containsKey(cameraId)) {
-      _channels[cameraId].setMethodCallHandler(null);
+      _channels[cameraId]!.setMethodCallHandler(null);
```
bparrishMines, line 140: nit: My preference is that `!` should only be used if you can guarantee that something is non-null by looking within the same scope, otherwise it should be handled. But, I also just started using null safety, so my intuition may be wrong on that.

Also, the `MethodCall` is still invoked earlier in the method, but a `null` exception can still be thrown after. This could probably be checked and handled before the `MethodCall`.

mvanbeusekom: I had to think a bit about this, but I agree. If possible it makes sense to only use the `!` within the same scope and where access is clear. Thanks for pointing this out. I have updated the code.

bparrishMines (approved):
LGTM pending tests

Most of your test failures should be fixed in the master branch, so it should just need a pull.

script/nnbd_plugins.sh
```diff
@@ -29,7 +30,7 @@ readonly NNBD_PLUGINS_LIST=(
 
 readonly NON_NNBD_PLUGINS_LIST=(
   # "android_alarm_manager"
-  "camera"
+  # "camera"
```
bparrishMines, line 33: @mvanbeusekom I think the issue might be that `camera` and `google_maps_flutter` now conflict because of `stream_transform`. I think uncommenting `google_maps_flutter` in this list might fix the build errors.

What are your thoughts @ditman?

packages/camera/camera_platform_interface/pubspec.yaml
```diff
@@ -3,23 +3,23 @@ description: A common platform interface for the camera plugin.
...
-  async: ^2.4.2
-  mockito: ^4.1.1
-  pedantic: ^1.8.0
+  async: ^2.5.0-nullsafety.3
+  mockito: ^5.0.0-nullsafety.5
+  pedantic: ^1.10.0-nullsafety.3
 
 environment:
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: '>=2.12.0-133.7.beta <3.0.0'
```
bparrishMines, line 24: This should be 
```suggestion
  sdk: '>=2.12.0-0 <3.0.0'
```

mvanbeusekom: I don't mind changing this, but this is how the Dart migration tool updated the version number. That was why I didn't change it afterwards. Just checking what would be better?

script/nnbd_plugins.sh
bparrishMines: So it looks like camera also has a problem with `google_sign_in` on master:

```
ERR : Because every version of google_sign_in_platform_interface from path depends on quiver ^3.0.0-nullsafety.2 and every version of camera from path depends on quiver ^2.1.5, google_sign_in_platform_interface from path is incompatible with camera from path.

    | So, because all_plugins depends on both camera from path and google_sign_in_platform_interface from path, version solving failed.

FINE: Exception type: SolveFailure
```

I was able to get it to build on master locally when keeping `camera` uncommented and `google_maps_flutter" commented. I think `camera` is going to have to stay in this list until the `camera` package is migrated.

script/nnbd_plugins.sh
```diff
@@ -7,6 +7,7 @@
 readonly NNBD_PLUGINS_LIST=(
   "android_intent"
   "battery"
+  "camera_platform_interface"
```
bparrishMines, line 10: This should be

```suggestion
  "camera"
```

I wasn't able to get it to run locally on stable, but this might help on the CI.

packages/camera/camera_platform_interface/pubspec.yaml
bparrishMines: Yes, we've been setting the rest of our plugins to `2.12.0-0`, so it's mostly to keep them consistent. I don't believe it should cause any dependency problems.

script/nnbd_plugins.sh
bparrishMines: So I was finally able to get the test running locally. The error with `google_maps_flutter` for the `build_all_plugins` test went away once this was changed.

bparrishMines: I believe the issue is that `camera` needs to be excluded from `stable` and `master` until both the `camera` and `camera_platform_interface` package are migrated. `camera` is in a weird spot where it conflicts with migrated and non-migrated plugins.

Let me know if this works

mvanbeusekom: Wow thanks @bparrishMines, this completely boggled me ;). 

I have added a new commit where "camera" is now listed in both the "NNBD_PLUGINS_LIST" and the "NON_NNBD_PLUGINS_LIST". I will keep an eye on the tree with fingers crossed ;).

mvanbeusekom: Nice stuff that did it, thanks so much @bparrishMines. I will merge this now and will create the NNBD PR for the camera package.

packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart
```diff
@@ -58,8 +59,13 @@ class MethodChannelCamera extends CameraPlatform {
   @override
   Future<List<CameraDescription>> availableCameras() async {
     try {
-      final List<Map<dynamic, dynamic>> cameras = await _channel
+      final cameras = await _channel
```
ditman, line 62: I think we want left-hand types (but I'm not 100% sure), @bparrishMines?

ditman (commented):
This is looking great! The only critical change, I think, is the change to `nnbd_plugins.sh`; I think it's preventing tests from running on this package and PR.

script/nnbd_plugins.sh
```diff
@@ -7,6 +7,7 @@
 readonly NNBD_PLUGINS_LIST=(
   "android_intent"
   "battery"
+  "camera"
```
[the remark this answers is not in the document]
bparrishMines: I suggested this to fix the failing `build_all_plugins` test. `camera` has a conflict with `google_sign_in` while `camera_platform_interface` has a conflict with `google_maps_flutter`. I suggested that we keep them excluded from both until the camera plugin is migrated. 

The alternative to this would be to remove the camera plugin from the `build_all_plugins` test and run the camera tests on master. Would that be alright?

packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart
bparrishMines: It differs between plugins. If the general style within the plugin has left-hand types, then we should probably keep the pattern the same. We could also decide on a preference for the repo as a whole and add it to the global `analysis_options.yaml`. https://github.com/flutter/plugins/blob/master/analysis_options.yaml

packages/camera/camera_platform_interface/test/method_channel/method_channel_camera_test.dart
```diff
@@ -48,7 +51,7 @@ void main() {
             arguments: {
               'cameraName': 'Test',
               'resolutionPreset': 'high',
-              'enableAudio': null
+              'enableAudio': true
```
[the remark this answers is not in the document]
mvanbeusekom: I think you are right. Looking closer at the native Android and iOS implementations I see no specific treatment of catching a `null` value and converting it too `true`. Same goes for the method channel implementation and the camera_controller.

packages/camera/camera_platform_interface/test/method_channel/method_channel_camera_test.dart
```diff
@@ -70,7 +73,11 @@ void main() {
         // Act
         expect(
           () => camera.createCamera(
-            CameraDescription(name: 'Test'),
+            CameraDescription(
+              name: 'Test',
+              lensDirection: CameraLensDirection.back,
+              sensorOrientation: 0,
```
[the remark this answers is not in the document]
mvanbeusekom: My reasoning was that the `CameraDescription` instance used should always coming from the native side (as a result of the `availableCameras` method). Being such these parameters should always have a value and should never be defaulted to something just because they are not available (this might lead to bugs).

packages/camera/camera_platform_interface/pubspec.yaml
```diff
@@ -3,23 +3,23 @@ description: A common platform interface for the camera plugin.
...
+  stream_transform: ^2.0.0-nullsafety.0
 
 dev_dependencies:
   flutter_test:
     sdk: flutter
-  async: ^2.4.2
-  mockito: ^4.1.1
-  pedantic: ^1.8.0
+  async: ^2.5.0-nullsafety.3
+  mockito: ^5.0.0-nullsafety.5
```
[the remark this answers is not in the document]
mvanbeusekom: That indeed makes sense, I will refactor this.

packages/camera/camera_platform_interface/test/method_channel/method_channel_camera_test.dart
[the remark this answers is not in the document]
ditman: Ahhh good point. If users are not going to be creating their own CameraDescription, I'm fine with them having a more involved constructor. Thanks for the explanation!

packages/camera/camera_platform_interface/pubspec.yaml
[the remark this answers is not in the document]
ditman: Thanks for cleaning this up!

script/nnbd_plugins.sh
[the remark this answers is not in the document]
ditman: Ahhh, I see! As this is right now, `build_all_plugins` is already not building the `camera` plugin, right? As long as some 'camera' tests run in some shard in `master`, it's probably enough!

PR #3500, opened 2021-02-02
[google_maps_flutter_platform_interface ]fix version

cyanglaz:
The version should be 1.2.0. see https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/CHANGELOG.md


stuartmorgan (approved):
LGTM (once the script test part is reverted, obviously)

bparrishMines:
Could we change the `plugin_tools` function in `common.sh` script to 
```
function plugin_tools() {
  pub_command global activate flutter_plugin_tools > /dev/null
  pub_command global run flutter_plugin_tools "$@"
}
```

So that we can still use the function here

PR #3511, opened 2021-02-04
Add a note about Plus plugins to CONTRIBUTING.md

stuartmorgan:
Adds a prominent note to CONTRIBUTING.md about the new policy regarding PRs for plugins for which there is a Flutter Community Plus Plugins equivalent.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,10 +19,10 @@ for any of the following plugins, we encourage you to submit it
 - `battery`
 - `connectivity`
 - `device_info`
-- `wifi_info_flutter` (corresponds to `network_info_plus`)
 - `package_info`
 - `sensors`
 - `share`
+- `wifi_info_flutter` (corresponds to `network_info_plus`)
 
 ## Things you will need
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ _See also: [Flutter's code of conduct](https://github.com/flutter/flutter/blob/m
 
 As of January 2021, we are no longer accepting non-critical PRs for plugins
 for which there is a corresponding [Flutter Community Plus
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -5,6 +5,25 @@
 
 _See also: [Flutter's code of conduct](https://github.com/flutter/flutter/blob/master/CODE_OF_CONDUCT.md)_
 
+## Important note
+
+As of January 2021, we are no longer accepting non-critical PRs for plugins
```
cyanglaz, line 10: Do we have an example or explanation of what a "non-critical" PR is? Or is it just PRs for all the plugins in the list below?

cyanglaz (approved):
LGTM

CONTRIBUTING.md
stuartmorgan: Good point, reworded below to give examples of "critical".

(I expect we'll get some amount of pushback regardless, with people arguing that, e.g., a specific feature is critical to them, but this should help better set expectations.)

bparrishMines (approved):
LGTM

CONTRIBUTING.md
```diff
@@ -5,6 +5,25 @@
...
+for which there is a corresponding [Flutter Community Plus
+Plugin](https://plus.fluttercommunity.dev/), as we hope that in time we're able
```
csells, line 12: "as we hope in time to be able"

stuartmorgan: Sure, done. Also made the same change in the snippet we're using to close PRs, since this was copied directly from there for consistency.

I'll also be sure to pass along your feedback to the original author of that snippet ;)

csells: Excellent. I'm sure they'll appreciate the feedback. : )

PR #3526, opened 2021-02-08
[shared_preferences] Migrate main package to null-safety

gaetschwartz:
Migrates `shared_preferences` to nnbd.

Fixes https://github.com/flutter/flutter/issues/74876.

**Note:** also recreated the android and macos `example` applications because the macOS app was apparently a copy of the connectivity app and had `connectivity_example` as its name, the Android app was using the embedding v1. 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/shared_preferences/shared_preferences/pubspec.yaml
```diff
@@ -26,16 +26,16 @@ dependencies:
...
   # TODO(franciscojma): Revisit this (either update this part in the  design or the pub tool).
   # https://github.com/flutter/flutter/issues/46264
-  shared_preferences_linux: ^0.0.2
-  shared_preferences_macos: ^0.0.1
-  shared_preferences_web: ^0.1.2
-  shared_preferences_windows: ^0.0.1
+  shared_preferences_linux: ^0.0.4-nullsafety
+  shared_preferences_macos: ^0.0.2-nullsafety
+  shared_preferences_web: ^0.2.0-nullsafety
+  shared_preferences_windows: ^0.0.3-nullsafety
```
stuartmorgan, line 39: You didn't change the SDK min version, so there's no actual migration here.

stuartmorgan (changes requested):
This doesn't actually do a null-safety migration; there's no SDK version change (which is the controlling factor), and none of the nullability issues is shared_preferences.dart are resolved.

rockingdice:
Looking forward to it!

gaetschwartz:
Correct, forgot to. Done now.

packages/shared_preferences/shared_preferences/pubspec.yaml
```diff
@@ -5,7 +5,7 @@ homepage: https://github.com/flutter/plugins/tree/master/packages/shared_prefere
 # 0.5.y+z is compatible with 1.0.0, if you land a breaking change bump
 # the version to 2.0.0.
 # See more details: https://github.com/flutter/flutter/wiki/Package-migration-to-1.0.0
-version: 0.5.13+2
+version: 0.6.0-nullsafety
```
stuartmorgan, line 8: I almost forgot (and may have missed this for other plugins): we actually need to make this 2.0.0-nullsafety. The README currently says that we guarantee 1.0 will be compatible with the current version, and encourages people setting their constraints to <2, but this is a breaking change.

(This should also update the README to remove that section.)

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -54,19 +53,19 @@ class SharedPreferences {
     if (_completer == null) {
       _completer = Completer<SharedPreferences>();
```
stuartmorgan, line 54: Let's use a non-nullable local variable here, which is assigned to `_completer` at the end of the `try`, in order to remove most of the uses of `!`. `!` is a tool of last resort.

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -77,79 +76,80 @@ class SharedPreferences {
   ///
   /// It is NOT guaranteed that this cache and the device prefs will remain
   /// in sync since the setter method might fail for any reason.
-  final Map<String, Object> _preferenceCache;
+  final Map<String, Object?> _preferenceCache;
```
stuartmorgan, line 79: Why are the values nullable? That's not what the platform interface that this is caching vends.

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -77,79 +76,80 @@ class SharedPreferences {
...
       _preferenceCache[key] = list;
     }
     // Make a copy of the list so that later mutations won't propagate
-    return list?.toList();
+    return list?.toList() as List<String>?;
   }
 
   /// Saves a boolean [value] to persistent storage in the background.
   ///
   /// If [value] is null, this is equivalent to calling [remove()] on the [key].
```
stuartmorgan, line 120: All of the setters should take non-nullable values, and these comments removed. Nullability is a breaking change, which means we don't have to keep backwards compatibility; we should design the interface to be hard to use incorrectly in a null-safe world. In that context, we should require people to be explicit about removing values rather than accidentally passing nulls through and getting a different behavior.

(That's also consistent with `Map`, which I think is a feature.)

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -77,79 +76,80 @@ class SharedPreferences {
...
-  Future<bool> setStringList(String key, List<String> value) =>
+  Future<bool> setStringList(String key, List<String>? value) =>
       _setValue('StringList', key, value);
 
   /// Removes an entry from persistent storage.
   Future<bool> remove(String key) => _setValue(null, key, null);
 
-  Future<bool> _setValue(String valueType, String key, Object value) {
+  Future<bool> _setValue(String? valueType, String key, Object? value) {
```
stuartmorgan, line 152: With the changes mentioned above, this can take all non-nullable values, and the remove logic can move into `remove`.

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -161,7 +161,7 @@ class SharedPreferences {
       } else {
         _preferenceCache[key] = value;
       }
-      return _store.setValue(valueType, prefixedKey, value);
+      return _store.setValue(valueType!, prefixedKey, value);
     }
   }
```
stuartmorgan, line 167: Let's remove `commit` while making the breaking change.

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -181,17 +181,17 @@ class SharedPreferences {
   /// Use this method to observe modifications that were made in native code
   /// (without using the plugin) while the app is running.
   Future<void> reload() async {
-    final Map<String, Object> preferences =
+    final Map<String, Object?> preferences =
```
stuartmorgan, line 184: Same question for all of these about why the values are nullable.

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -213,7 +213,8 @@ class SharedPreferences {
       return MapEntry<String, dynamic>(newKey, value);
     });
     SharedPreferencesStorePlatform.instance =
-        InMemorySharedPreferencesStore.withData(newValues);
+        InMemorySharedPreferencesStore.withData(
+            newValues.cast<String, Object>());
```
stuartmorgan, line 217: This cast isn't safe, since you are allowing `dynamic` in the API (meaning it won't do any type checking). You should change the API to `<String, Object>` instead of making a potentially invalid conversion at the end.

packages/shared_preferences/shared_preferences/test/shared_preferences_test.dart
```diff
@@ -29,14 +27,14 @@ void main() {
       'flutter.List': <String>['baz', 'quox'],
     };
 
-    FakeSharedPreferencesStore store;
-    SharedPreferences preferences;
+    FakeSharedPreferencesStore? store;
```
stuartmorgan, line 30: Why not `late`?

stuartmorgan (changes requested):

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
gaetschwartz: Not sure what you are talking about. The `commit` method is deprecated isn't it ?

stuartmorgan: Right, `deprecated` means it shouldn't be used and will be removed in the future. Since it's been deprecated for a long time, and we're doing a breaking change for the plugin anway, it's a good time to actually remove it.

packages/shared_preferences/shared_preferences/test/shared_preferences_test.dart
```diff
@@ -29,11 +27,11 @@ void main() {
...
     };
 
-    FakeSharedPreferencesStore store;
-    SharedPreferences preferences;
+    late FakeSharedPreferencesStore store;
+    late SharedPreferences preferences;
 
     setUp(() async {
-      store = FakeSharedPreferencesStore(kTestValues);
+      store = FakeSharedPreferencesStore(kTestValues.cast<String, Object>());
```
stuartmorgan, line 34: Can't you change the declarations above rather than casting?

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -83,86 +83,76 @@ class SharedPreferences {
...
   /// Android doesn't support storing doubles, so it will be stored as a float.
-  ///
-  /// If [value] is null, this is equivalent to calling [remove()] on the [key].
   Future<bool> setDouble(String key, double value) =>
       _setValue('Double', key, value);
 
   /// Saves a string [value] to persistent storage in the background.
-  ///
-  /// If [value] is null, this is equivalent to calling [remove()] on the [key].
+
```
stuartmorgan, line 132: Please remove.

stuartmorgan (changes requested):
One last small question, but otherwise looking good!

gaetschwartz:
I was also thinking about something regarding the different setters. Imagine someone sets some arbitrary data (let's say `String` here and thus could be null). The user would have to always check if null and then if it is null use `remove` if so and use `setString` if not. What about adding nullable setters as well ? Something like `setNullableString` that would itself check if null, if so use `remove` and else use `setString` ? A lot of users might probably be confused as the change to non-nullable fields in the setters are breaking and not really document how to use them. I'm pretty sure a certain amount of people used `setString(null)` to delete stuff.

What do you think ? 

I think at least updating the changelog with instructions on how to migrate something like `setString(String? s)` to 
```dart
if(s == null) prefs.remove('key');
else prefs.setString('key', s);
```

would be preferred. Thoughts ? @stuartmorgan

stuartmorgan:
> Imagine someone sets some arbitrary data (let's say `String` here and thus could be null).

`String` can't be null though. Do you mean `String?`? If so, the question is, what did they intend that to mean? Because `shared_preferences` *cannot* store null data for a key. If the answer is that they intended setting null to mean removing, why? That's not "set[ting] arbitrary data", it's either setting data or removing data, and they should consider why they are structuring their APIs that way.

> The user would have to always check if null and then if it is null use `remove` if so and use `setString` if not.

If they want an API that even in a NNBD world treats setting null as meaning removal, then yes. I would argue that that's poor nullsafe API design on their part, however.

> What about adding nullable setters as well ? Something like `setNullableString` that would itself check if null, if so use `remove` and else use `setString` ?

That's a badly named API though; nothing is being set in the `null` case, and in fact the opposite is happening. We certainly can't stop people from writing such wrappers if they really want them in their own code, I don't see a reason to encourage that code structure by providing them.

> A lot of users might probably be confused as the change to non-nullable fields in the setters are breaking and not really document how to use them. I'm pretty sure a certain amount of people used `setString(null)` to delete stuff.

Agreed we should call out that aspect of the breaking change in the changelog. Something like "Setters no longer accept null to mean removing values. If you were previously using `set*(key, null)` for removing, instead call `remove(key)`.

> ```dart
> void setName(String? s) async {
>   if(s == null) prefs.remove('key');
>   else prefs.setString('key', s); // Eventually prefs.setNullableString('key', s);
> }
> ```

I do not want to provide that example, because I think in most cases it's an anti-pattern for NNBD.

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -83,86 +83,75 @@ class SharedPreferences {
...
   /// Removes an entry from persistent storage.
-  Future<bool> remove(String key) => _setValue(null, key, null);
+  Future<bool> remove(String key) {
+    final String prefixedKey = '$_prefix$key';
+    _preferenceCache.remove(key);
+    return _store.remove(prefixedKey);
+  }
 
   Future<bool> _setValue(String valueType, String key, Object value) {
     final String prefixedKey = '$_prefix$key';
```
stuartmorgan, line 147: The discussion about null setters made me realize that we should, as an added protection against mistakes when people are running in unsound mode, add a `ArgumentError.checkNotNull` on `value` here to make sure we never pass nulls through to the platform implementations (where the behavior would be platform-dependent).

gaetschwartz:
Alright, I agree, I updated the changelog and added the ArgumentError.checkNotNull(). Can you think of another example or is it okay the way I put it ?

stuartmorgan (approved):
LGTM (including README changes; thanks again for pointing out we should highlight this change)! Thanks for all the iteration on this!

I'll get it landed and published once all the CI runs are green.

gaetschwartz:
CI is failing for an obscure reason @stuartmorgan any idea why ? `test CHANNEL:stable` succeeded but it doesn't update https://cirrus-ci.com/task/4857579824218112.

stuartmorgan:
Yes, I'm not sure why it didn't update the status here. Since it did indeed pass though, we're good to land.

PR #3530, opened 2021-02-09
[camera_platform_interface] Migrate to null safety

mvanbeusekom:
This PR is a follow up on PR #3497 and contains fixes for the following feedback:

- Improve documentation of `availableCameras` method ([original comment](https://github.com/flutter/plugins/pull/3497#discussion_r572389747));
- Removed dependency on Mockito ([original comment](https://github.com/flutter/plugins/pull/3497#discussion_r572394822));
- Use a default value of `false` for the `enableAudio` parameter ([original comment](https://github.com/flutter/plugins/pull/3497#discussion_r572396714));
- Add left hand type definition to declared variable ([original comment](https://github.com/flutter/plugins/pull/3497#discussion_r572424013)).

Note: the version information in the pubspec.yaml and CHANGELOG.md is not updated on purpose since this should be published with the same release as the original PR.

## Pre-launch Checklist

- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [-] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [-] I updated CHANGELOG.md to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[...]

PR #3533, opened 2021-02-10
[camera] NNBD migration of the camera plugin

mvanbeusekom:
Migrates the camera plugin to NNBD. 

Note: the PR is marked "Draft" because it depends on the NNBD migration of the camera_platform_interface package which is yet to be published (see PR #3530).

flutter/flutter#75228

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/lib/src/camera_controller.dart
+++ b/packages/camera/camera/lib/src/camera_controller.dart
@@ -88,7 +88,7 @@ class CameraValue {
 
   /// The size of the preview in pixels.
   ///
-  /// Is `null` until  [isInitialized] is `true`.
+  /// Is `null` until [isInitialized] is `true`.
   final Size? previewSize;
 
   /// Convenience getter for `previewSize.width / previewSize.height`.
--- a/packages/camera/camera/lib/src/camera_controller.dart
+++ b/packages/camera/camera/lib/src/camera_controller.dart
@@ -32,19 +32,19 @@ Future<List<CameraDescription>> availableCameras() async {
 class CameraValue {
   /// Creates a new camera controller state.
   const CameraValue({
-    this.isInitialized,
+    required this.isInitialized,
     this.errorDescription,
     this.previewSize,
-    this.isRecordingVideo,
-    this.isTakingPicture,
-    this.isStreamingImages,
-    bool isRecordingPaused,
-    this.flashMode,
+    required this.isRecordingVideo,
+    required this.isTakingPicture,
+    required this.isStreamingImages,
+    required bool isRecordingPaused,
+    required this.flashMode,
     this.exposureMode,
     this.focusMode,
-    this.exposurePointSupported,
-    this.focusPointSupported,
-    this.deviceOrientation,
+    required this.exposurePointSupported,
+    required this.focusPointSupported,
+    required this.deviceOrientation,
     this.lockedCaptureOrientation,
     this.recordingOrientation,
   }) : _isRecordingPaused = isRecordingPaused;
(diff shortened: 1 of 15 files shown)


packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -84,17 +84,17 @@ class CameraValue {
   ///
   /// This is null while the controller is not in an error state.
   /// When [hasError] is true this contains the error description.
-  final String errorDescription;
+  final String? errorDescription;
 
   /// The size of the preview in pixels.
   ///
   /// Is `null` until  [isInitialized] is `true`.
```
AlexV525, line 91: Maybe we can update typos all together?

```suggestion
  /// Is `null` until [isInitialized] is `true`.
```

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -105,10 +105,10 @@ class CameraValue {
   final FlashMode flashMode;
```
AlexV525, line 105: I'm curious, if below variables can be null when the value is not initialized (just guessing), why this can be non-null? Also can we improve the nullability in other way to make less nullable variable in the `CameraValue`?

mvanbeusekom:
@googlebot I fixed it.

AlexV525:
Sorry for the late notice, but CLA bot seems cannot recognized suggestions' commits' CLAs directly, so you need to remove the *Co-authored-by* message when you apply suggestions.

mvanbeusekom:
> Sorry for the late notice, but CLA bot seems cannot recognized suggestions' commits' CLAs directly, so you need to remove the _Co-authored-by_ message when you apply suggestions.

Check, didn't realise that, but thank you for letting me know. Do you know if there is anything I can do now to fix it?

AlexV525:
> Do you know if there is anything I can do now to fix it?

Let's try rebase and commit again?

packages/camera/camera/lib/src/camera_controller.dart
mvanbeusekom: Looking at the native code we could indeed initialise the `focusMode` and `exposureMode` fields to `auto` (as this is the default values used on iOS and Android if these are not set). I will update these values. 

The `lockedCaptureOrientation` parameter uses `null` to indicate that no orientation lock is set, meaning it can be set (and default is set) to `null` which should not make it required. The same goes for `recordingOrientation` which is only set during video recording.

Lastly we have the `previewSize` and `errorDescription` parameters which are allowed to be `null` since they are only set once the a camera has been successfully initialised.

AlexV525: > Lastly we have the `previewSize` and `errorDescription` parameters which are allowed to be `null` since they are only set once the a camera has been successfully initialised.

How about `late final NonNullType` ?

mvanbeusekom: Not sure that would work, my understanding is that `late` should only be used in cases where we can ensure variables are initialised before they are used. In this case both fields can be accessed (or used) before they are initialised (for example when you have instantiated the `CameraController` but haven't called `initialize` (and awaited) yet. Accessing these fields before they are initialised would result in a `LateInitializationError `which might introduce bugs. What do you think?

AlexV525: > `late` should only be used in cases where we can ensure variables are initialised before they are used

Yes, that situation is quite similar to the `CameraValue`, while nothing is valid for read or use before it gets initialized.

Here's my common usage with the camera:
* Initialize a controller, and wait til it's initialized.
* After `isInitialized`, call another layout or value binding widgets for use.

In this case, nothing will be reached before initialized, so no exception will be thrown.

mvanbeusekom: But the problem here is that `CameraValue` is an immutable class. This would mean that once the class is constructed, it shouldn't change. This means we need to be able to set the `previewSize` field to some default value (which currently is `null`) when it is first constructed. Then when the `CameraController` is initialised a new instance of the `CameraValue` class is created and a new value is assigned to the `previewSize` field.

If we want to convert the `previewSize` field to a `late` field we need to make it private (otherwise people can set it from outside, which is something we don't want being an immutable class), and convert the `final Size previewSize;` field into a getter: `Size get previewSize => _previewSize;` 

Now we would need multiple constructors where one doesn't set the `previewSize` field (this would be the `uninitialized()` named constructor) and the other "normal" constructor does initialise it. This would mean some code duplication. And another side effect would be that we now need to get rid of the `const` keyword for the constructors (which isn't really a big problem in my opinion but still).

AlexV525: Agreed. We can ship this and see if any good ideas will be pop up from the community.

AlexV525 (approved):
LGTM

PR #3542, opened 2021-02-12
Added comment about path depenendencies

jonasfj:
Adds a comment `# ^x.y.z in your application` where examples have path dependencies.

This was motivated by https://github.com/dart-lang/pub/issues/1850#issuecomment-581117550 where a user is running into problems by copying the dependency from an `example/pubspec.yaml`.

Maybe we shouldn't make it all one-line... and maybe we could phrase the comment better.
Or maybe we shouldn't even do this -- feel free to close this if you think this is too obvious :)

### DIFF
--- a/packages/android_alarm_manager/example/pubspec.yaml
+++ b/packages/android_alarm_manager/example/pubspec.yaml
@@ -5,6 +5,11 @@ dependencies:
   flutter:
     sdk: flutter
   android_alarm_manager:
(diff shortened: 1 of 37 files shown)


stuartmorgan (approved):
LGTM!

stuartmorgan:
Removing need-publishing; these only affect the examples, which people aren't going to be using from published versions.

PR #3550, opened 2021-02-15
[file_selector_web] Migrated to null-safety

SirusCodes:
## Issue
https://github.com/flutter/flutter/issues/75235

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
```diff
@@ -1,89 +0,0 @@
...
-import 'package:file_selector_web/src/dom_helper.dart';
-
-void main() {
-  group('FileSelectorWeb', () {
-    IntegrationTestWidgetsFlutterBinding.ensureInitialized();
-    MockDomHelper mockDomHelper;
-    FileSelectorWeb plugin;
-
-    setUp(() {
-      mockDomHelper = MockDomHelper();
```
stuartmorgan, line 22: This file should definitely not be removed; a manual mock should be straightforward to write here.

stuartmorgan (changes requested):

packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
SirusCodes: Yes I made the mock after your comment on the issue, but I'm unable to run and test it on my local machine.

packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
```diff
@@ -81,7 +80,17 @@ void main() {
...
+class MockDomHelper extends Mock implements DomHelper {
+  @override
+  Future<List<XFile>> getFiles({
+    String accept = '',
+    bool multiple = false,
+    FileUploadInputElement? input,
+  }) {
+    return super
+        .noSuchMethod(Invocation.method(#getFiles, [accept, multiple, input]));
+  }
```
stuartmorgan, line 92: I'm confused by this override; it's not doing any of the nullability/non-nullability transformations described as necessary in mockito's NULL_SAFETY_README. Are you sure you were compiling and running the tests in strong mode?

(To clarify, when I referred to a manual mock I was actually thinking one that didn't use mockito at all, but instead just a simple class that implements `DomHelper` and has a method to set a return value for `getFiles` (and any argument checking to do). We could potentially do this too, but it doesn't look like this actually handles nullability correctly.)

packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
stuartmorgan: Can you elaborate on this? What's the issue with running the test?

SirusCodes: I opened an [issue](https://github.com/flutter/flutter/issues/76213) for this

packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
SirusCodes: Ohh I thought I need to [manually mock](https://github.com/dart-lang/mockito/blob/master/NULL_SAFETY_README.md#manual-mock-implementaion) them.

Can you please share an already implemented such type of mock for reference?

stuartmorgan: I don't think a different example would be useful given that it's class-specific. What I'm talking about is a class that implements the interface of the thing being mocked/faked, and also has whenever else is needed to populate it with fake data. For this, the simplest form would be something like (this is freeform so probably doesn't compile):

```
class FakeDomHelper implements DomHelper {
  List<XFile> filesToReturn = <XFile>[];

  @override
  Future<List<XFile>> getFiles({
    String accept = '',
    bool multiple = false,
    FileUploadInputElement? input,
  }) {
    return filesToReturn;
  }

  void setFiles(List<XFile> files) {
    filesToReturn = files;
  }
}
```

To keep the same level of correctness enforcement that the tests currently have you'd probably want to add `expectedAcceptStringParam` and `expectedMultipleParam` to that, and have `getFiles` return an empty array when there's no match. Or make the member variable a map of return values keyed by a combination of `accept` and `multiple`.

SirusCodes: > To keep the same level of correctness enforcement that the tests currently have you'd probably want to add `expectedAcceptStringParam` and `expectedMultipleParam` to that, and have `getFiles` return an empty array when there's no match. Or make the member variable a map of return values keyed by a combination of `accept` and `multiple`.

I couldn't get this but I have implemented it in a different way. Can you have a look at it?

packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
```diff
@@ -15,16 +13,18 @@ import 'package:file_selector_web/src/dom_helper.dart';
...
-      plugin = FileSelectorWeb(domHelper: mockDomHelper);
-    });
 
     group('openFile', () {
-      final mockFile = createXFile('1001', 'identity.png');
+      late MockDomHelper mockDomHelper;
+      late FileSelectorWeb plugin;
+      late XFile mockFile;
+
+      setUp(() {
```
stuartmorgan, line 22: Setup that's specific to a single test, as you have now made both of these, should be in the test itself, not `setUp`.

packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
stuartmorgan: That's not implementing correctness enforcement a different way, it's ignoring it. This test does not validate that the calls are correct the way the previous test did.

Please use one of the approaches I described, or something equivalent, so that this isn't regressing test coverage. Alternately, @ditman, perhaps you could take over this portion of the PR?

packages/file_selector/file_selector_web/pubspec.yaml
```diff
@@ -11,22 +11,21 @@ flutter:
...
 dependencies:
-  file_selector_platform_interface: ^1.0.2
-  platform_detect: ^1.4.0
+  file_selector_platform_interface: ^2.0.0-nullsafety.0
   flutter:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  meta: ^1.1.7
+  meta: ^1.3.0-nullsafety.6
```
stuartmorgan, line 19: This can just be 1.3.0

stuartmorgan (changes requested):

ditman:
@googlebot I consent.

ditman:
I've tweaked a little bit the directory structure of the package to better match how we mix unit + integration_tests in web now (very similar to the url_launcher_web package). I'll review the testing approach next.

packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
ditman: I've resolved this issue in the branch, the way to run the integration tests now is documented in the PR :)

packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
ditman: I've added expectation checks to the getFiles method. If the expected accept/multiple params don't match, the test will fail. Since we don't reuse the same mock twice, this is probably good enough.

packages/file_selector/file_selector_web/integration_test/file_selector_web_test.dart
ditman: Removed

packages/file_selector/file_selector_web/example/integration_test/dom_helper_test.dart
```diff
@@ -2,29 +2,27 @@
...
 import 'package:file_selector_platform_interface/file_selector_platform_interface.dart';
 
 void main() {
-  group('FileSelectorWeb', () {
+  group('dom_helper', () {
     IntegrationTestWidgetsFlutterBinding.ensureInitialized();
-    DomHelper domHelper;
-    FileUploadInputElement input;
+    late DomHelper domHelper;
+    FileUploadInputElement? input;
```
stuartmorgan, line 15: Can't this be `late`?

packages/file_selector/file_selector_web/example/pubspec.yaml
```diff
@@ -0,0 +1,21 @@
...
+    path: ../
+  flutter_driver:
+    sdk: flutter
+  flutter_test:
+    sdk: flutter
+  integration_test:
+    sdk: flutter
+
+environment:
+  sdk: ">=2.12.0-0 <3.0.0"
```
stuartmorgan, line 20: We're using 2.12.0-259.9.beta for all the stable ranges.

packages/file_selector/file_selector_web/example/pubspec.yaml
```diff
@@ -0,0 +1,21 @@
...
+  flutter: ">=1.26.0-0" # For integration_test from sdk
```
stuartmorgan, line 21: I thought null safety support was post-1.26?

packages/file_selector/file_selector_web/lib/src/dom_helper.dart
```diff
@@ -14,50 +14,51 @@ class DomHelper {
...
-      final ErrorEvent error = event;
+    inputElement.onError.first.then((event) {
+      final ErrorEvent error = event as ErrorEvent;
       final platformException = PlatformException(
         code: error.type,
         message: error.message,
       );
-      input.remove();
-      _completer.completeError(platformException);
+      input!.remove();
```
stuartmorgan, line 49: `inputElement`? This isn't safe otherwise.

packages/file_selector/file_selector_web/lib/src/dom_helper.dart
```diff
@@ -14,50 +14,51 @@ class DomHelper {
...
-    return _completer.future;
+    return completer.future;
   }
 
   XFile _convertFileToXFile(File file) => XFile(
         Url.createObjectUrl(file),
         name: file.name,
         length: file.size,
-        lastModified: DateTime.fromMillisecondsSinceEpoch(file.lastModified),
+        lastModified: DateTime.fromMillisecondsSinceEpoch(file.lastModified!),
```
stuartmorgan, line 62: Why is this safe?

packages/file_selector/file_selector_web/pubspec.yaml
```diff
@@ -11,22 +11,18 @@ flutter:
...
-  mockito: ^4.1.1
-  pedantic: ^1.8.0
-  integration_test:
-    path: ../../integration_test
+  pedantic: ^1.10.0
 
 environment:
-  sdk: ">=2.2.0 <3.0.0"
-  flutter: ">=1.10.0"
+  sdk: '>=2.12.0-259.9.beta <3.0.0'
```
stuartmorgan, line 27: Let's stick to " for repo consistency.

stuartmorgan (changes requested):

packages/file_selector/file_selector_web/example/integration_test/dom_helper_test.dart
ditman:
```
All tests passed.
```

Yup! Done

packages/file_selector/file_selector_web/example/pubspec.yaml
ditman: Yeah, I changed this to `1.27.0-0`. I think this slipped through the cracks in url_launcher_web, though. I might revisit that one after everything is migrated.

packages/file_selector/file_selector_web/example/pubspec.yaml
ditman: Arrrrghghgh

packages/file_selector/file_selector_web/lib/src/dom_helper.dart
ditman: Ah yes, good catch! inputElement everywhere in this file, after the first initialization!

packages/file_selector/file_selector_web/lib/src/dom_helper.dart
ditman: I think this is `?` in the definition because of older browsers that might not expose this feature [see docs](https://developer.mozilla.org/en-US/docs/Web/API/File/lastModified).

I've defaulted this to `DateTime.now().millisecondsSinceEpoch`, which is the default for files that don't have this field according to MDN.

stuartmorgan (approved):
LGTM

ditman:
Thanks @SirusCodes!! This saved me a TON of time, it'll get published as soon as CI succeeds in `master`!

ditman:
This has been published: https://pub.dev/packages/file_selector_web/versions/0.8.0

SirusCodes:
> Thanks @SirusCodes!! This saved me a TON of time, it'll get published as soon as CI succeeds in `master`!

Thank you @ditman!! I had a lot of issues while writing those tests, now I can go and learn a LOT from this PR and now I know what I need to focus on... getting better at.

SirusCodes:
> This has been published: https://pub.dev/packages/file_selector_web/versions/0.8.0

Will start migrating `file_selector`

stuartmorgan:
@SirusCodes  I have `file_selector` essentially finished; it's just waiting on a small fix to `cross_file`.

SirusCodes:
Ok 👍 thanks @stuartmorgan

PR #3552, opened 2021-02-15
Fix the build-all exclusion list

stuartmorgan:
build_all_plugins_app.sh contains an exclusion list, which currently contains almost all of the non-app-facing plugins. However, the script those exclusions are passed to expects federated plugin exclusions to be of the form `plugin_name/plugin_name_subplugin_name`, not just `plugin_name_subplugin_name`, so in practice almost nothing on that list has actually been doing anything.

This fixes the script to allow either mode of exclusion (since clearly people expect using just the name to work), and scrubs everything from the list that clearly wasn't actually needed.

### DIFF
--- a/script/nnbd_plugins.sh
+++ b/script/nnbd_plugins.sh
@@ -38,13 +38,9 @@ readonly NNBD_PLUGINS_LIST=(
 # building the all plugins app. This list should be kept empty.
 
 readonly NON_NNBD_PLUGINS_LIST=(
-  "camera"
-  "google_maps_flutter" # half migrated
-  # "image_picker"
-  # "in_app_purchase"
-  # "quick_actions"
-  # "sensors"
-  # "wifi_info_flutter"
+  "camera" # partially migrated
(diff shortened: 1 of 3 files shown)


script/nnbd_plugins.sh
```diff
@@ -38,13 +38,9 @@ readonly NNBD_PLUGINS_LIST=(
...
 
 readonly NON_NNBD_PLUGINS_LIST=(
-  "camera"
-  "google_maps_flutter" # half migrated
-  # "image_picker"
-  # "in_app_purchase"
-  # "quick_actions"
-  # "sensors"
-  # "wifi_info_flutter"
+  "camera" # partially migrated
```
mvanbeusekom, line 41: This morning I merged a null safety version of the `camera` plugin. I believe this line can now be removed (it is removed in `master` so maybe a rebase should be sufficient).

mvanbeusekom (changes requested):

mvanbeusekom (approved):
LGTM

PR #3556, opened 2021-02-16
Publish check

bparrishMines:


bparrishMines:
@stuartmorgan There was no `test/` folder in the tool directory: https://github.com/flutter/plugins/tree/master/script/tool.
So this PR doesn't include any unit tests for the command

script/tool/lib/src/publish_check_command.dart
```diff
@@ -0,0 +1,95 @@
...
+    if (failedPackages.isNotEmpty) {
+      final String error =
+          'FAIL: The following ${failedPackages.length} package(s) failed the '
+          'publishing check:';
+      final String joinedFailedPackages = failedPackages.join('\n');
+
+      final Colorize colorizedError = Colorize('$error\n$joinedFailedPackages')
+        ..red();
+      print(colorizedError);
+      throw ToolExit(64);
```
stuartmorgan, line 45: Why 64 (vs just 1)?

script/tool/lib/src/publish_check_command.dart
```diff
@@ -0,0 +1,95 @@
...
+        Colorize('All packages passed publish check!')..green();
+    print(passedMessage);
+  }
+
+  Pubspec tryParsePubspec(Directory package) {
+    final File pubspecFile = package.childFile('pubspec.yaml');
+
+    Pubspec pubspec;
+    try {
+      pubspec = Pubspec.parse(pubspecFile.readAsStringSync());
```
stuartmorgan, line 58: Couldn't this just be `return Pubspec.parse(pubspecFile.readAsStringSync());` to avoid the uninitialized variable?

stuartmorgan (approved):
LGTM with nits!

Please save the test code you wrote so that we can add it later once we have unit tests set up for the tool in the new location.

script/tool/lib/src/publish_check_command.dart
bparrishMines: It's been awhile, but I believe it was out of habit from reading something like https://www.freebsd.org/cgi/man.cgi?query=sysexits. I changed it to 1.

PR #3560, opened 2021-02-17
Publish check ignores prerelease sdk

bparrishMines:


script/tool/lib/src/publish_check_command.dart
```diff
@@ -63,6 +64,33 @@ class PublishCheckCommand extends PluginCommand {
...
+    });
+
+    process.stderr.listen((List<int> event) {
+      io.stderr.add(event);
+      output += String.fromCharCodes(event);
+    });
+
+    final int exitCode = await process.exitCode;
+
+    final bool hasPreReleaseSdk = output.contains('Package has 1 warning.') &&
```
stuartmorgan, line 87: Aren't errors and warnings reported separately? What does the output for a package with, say, 1 warning and 1 error look like? We need to make sure this won't pass a packages in that case.

stuartmorgan: From https://github.com/dart-lang/pub/blob/master/lib/src/command/lish.dart#L176 it looks like maybe we need to ensure that it doesn't contain "can't be published".

stuartmorgan: Oh wait, it early-returns after errors, before getting to warnings. So apparently we can't get the warning output if it's actually failed.

stuartmorgan, line 87: We need to wait on the output to ensure we're getting it all, to avoid potentially flaky results from the `contains` check. The `exitCode` docs say:

> There is no guarantee that stdout and stderr have finished reporting the buffered output of the process when the returned future completes. To be sure that all output is captured, wait for the done event on the streams.

stuartmorgan (changes requested):
Mostly looks great, just one issue with output wait.

stuartmorgan (approved):
LGTM!

PR #3567, opened 2021-02-19
Disable in_app_purchase for master

stuartmorgan:
It depends on the old version shared_preferences, which ultimately means
an indirect dependency on old win32, which no longer works on master.

This will be resolved when in_app_purchase is migrated to NNBD.


blasten (approved):

stuartmorgan:
Ah, this won't help because it only disable it from build-all; it will still fail on building the individual example.

I think since I just approved the IAP migration PR we can probably just wait for that to land rather than expand the workaround.

PR #3573, opened 2021-02-19
[image_picker_platform_interface] Bump NNBD version to stable

cyanglaz:
Bump the version to stable

### DIFF
--- a/packages/image_picker/image_picker_platform_interface/pubspec.yaml
+++ b/packages/image_picker/image_picker_platform_interface/pubspec.yaml
@@ -10,7 +10,7 @@ dependencies:
     sdk: flutter
   meta: ^1.3.0
   http: ^0.13.0
-  plugin_platform_interface: ^2.0.0
+  plugin_platform_interface: ">=1.0.0 <3.0.0"
 
 dev_dependencies:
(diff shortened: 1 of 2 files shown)


packages/image_picker/image_picker_platform_interface/pubspec.yaml
```diff
@@ -3,20 +3,20 @@ description: A common platform interface for the image_picker plugin.
...
 
 dependencies:
   flutter:
     sdk: flutter
-  meta: ^1.3.0-nullsafety.6
-  http: ^0.13.0-nullsafety.0
-  plugin_platform_interface: ^1.1.0-nullsafety.2
+  meta: ^1.3.0
+  http: ^0.13.0
+  plugin_platform_interface: ^2.0.0
```
stuartmorgan, line 13: Please use ">=1.0.0 <3.0.0" so that we don't create unnecessary (in this case) segmentation between prerelease and stable NNBD plugins.

cyanglaz: done

stuartmorgan (approved):
LGTM

PR #3579, opened 2021-02-19
[image_picker] NNBD stable

cyanglaz:
release image_picker nnbd version as stable

PR #3580, opened 2021-02-19
[image_picker] use nnbd version of deps to resolve ci failure

cyanglaz:
Landing this to fix CI.
This change will get published with https://github.com/flutter/plugins/pull/3579

### DIFF
--- a/packages/image_picker/image_picker/pubspec.yaml
+++ b/packages/image_picker/image_picker/pubspec.yaml
@@ -27,7 +27,7 @@ dev_dependencies:
     path: ../../integration_test
   mockito: ^5.0.0-nullsafety.7
   pedantic: ^1.10.0
-  plugin_platform_interface: ^2.0.0
+  plugin_platform_interface: ">=1.0.0 <3.0.0"
 
 environment:
   sdk: ">=2.12.0-0 <3.0.0"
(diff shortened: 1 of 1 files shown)


packages/image_picker/image_picker/pubspec.yaml
```diff
@@ -26,8 +26,8 @@ dev_dependencies:
   integration_test:
     path: ../../integration_test
   mockito: ^5.0.0-nullsafety.7
-  pedantic: ^1.8.0
-  plugin_platform_interface: ^1.0.3
+  pedantic: ^1.10.0
+  plugin_platform_interface: ^2.0.0
```
stuartmorgan, line 30: This needs to be a range.

cyanglaz: oh, forget about it. Done.

stuartmorgan (approved):
LGTM

jmagman:
This also closed the framework tree.

ditman:
If we're fairly confident this will work, we probably need to merge before pre-submit CI finishes.

At this time of the day, we're heavily throttled both on Cirrus and Firebase some times.

ditman:
Merging in red, to unblock the framework tree.

PR #3583, opened 2021-02-20
[url_launcher] Fix SDK copypasta in platform interface

stuartmorgan:
The lower end of the SDK range was wrong due to a bad copy/paste.


cyanglaz (approved):
LGTM

PR #3591, opened 2021-02-22
[camera] Fix iOS rotation issue

mvanbeusekom:
This PR fixes an issue where on iOS the default orientation was set to landscape right while it should have been set to portrait up. To compensate for this problem the `CameraPreview` widget used an additional offset when calculating the rotation on iOS. 

This PR solves the problem at the root and also removed the extra offset needed on iOS to adjust the orientation.

- Fixes flutter/flutter#76210

## Pre-launch Checklist

- [x] I signed the [CLA].


bparrishMines (commented):
Everything looks good, but this still lacks tests. Potentially a widget test?

mvanbeusekom:
@bparrishMines on iOS this indeed looks good but on Android I am running into the following problem: https://discord.com/channels/608014603317936148/608020293944082452/815873332574158890 (if you have a solution or advise I would love to hear it)

I will mark the PR as draft until this gets resolved. And I will also make sure to add tests.

packages/camera/camera/lib/src/camera_preview.dart
```diff
@@ -41,26 +40,25 @@ class CameraPreview extends StatelessWidget {
...
       DeviceOrientation.portraitUp: 0,
       DeviceOrientation.landscapeLeft: 1,
       DeviceOrientation.portraitDown: 2,
       DeviceOrientation.landscapeRight: 3,
     };
-    return turns[_getApplicableOrientation()]! + platformOffset;
+    return turns[_getApplicableOrientation()]!;
+  }
+
+  DeviceOrientation _getApplicableOrientation() {
```
cyanglaz, line 58: Just to confirm, there is no diff in this method right? The method was just moved down?

mvanbeusekom: Correct, it made more sense to me to keep it closer to where it was used. I see now that it causes a but of noice

cyanglaz: keeping it close to where it was used makes sense to me!

mvanbeusekom:
> Everything looks good, but this still lacks tests. Potentially a widget test?

@bparrishMines I have added some widget tests to make sure the rotated box turns correctly according to the different settings provided by the `CameraController`. Would appreciate it is you could have another look.

bparrishMines:
@mvanbeusekom As stated on Discord, Camera2 on Android automatically changes the orientation of the output while iOS requires you to set the value. 

On iOS, have you tried listening to the the [UIDeviceOrientationDidChangeNotification](https://developer.apple.com/documentation/uikit/uideviceorientationdidchangenotification?language=objc) and automatically updating the output camera preview orientation? I don't think there's anything that can be done on Android about Orientation since we use a `SurfaceTextureEntry` and not a `TextureView`.

mvanbeusekom:
> On iOS, have you tried listening to the the [UIDeviceOrientationDidChangeNotification](https://developer.apple.com/documentation/uikit/uideviceorientationdidchangenotification?language=objc) and automatically updating the output camera preview orientation?

Yes that is exactly what I did in this pull request, which we tested extensively and solves all orientation issues on iOS without the need of a `RotatedBox` on the Dart side. This is now ready and added additional automation tests to ensure this is working properly. 

I was just wondering if we could do something similar on Android so we could remove the `RotatedBox` on the Dart side altogether. I had an offline discussion with @cyanglaz and we agreed to land this PR first fixing the issues on iOS (which are currently the biggest problem) and file a new feature request for the Flutter Engine to see if its possible to expose the `TextureView` in the future so we could also solve this issue in a more elegant way for Android (without the need for additional coding on the Dart side).

So in short, this PR should be ready and would be great if you could review it.

bparrishMines (approved):
LGTM

snoopdoggy322:
In my case, updating to a new version did not fix this problem
Device: iPhone SE(old)

https://user-images.githubusercontent.com/38835822/122252946-67839f80-ced4-11eb-89c3-787586fbfcd7.MP4

PR #3588, opened 2021-02-20
[connectivity_macos] move NNBD to stable

cyanglaz:
Move the plugin to stable version

PR #3599, opened 2021-02-22
[connectivity_macos] fix version

cyanglaz:
Fix the wrong version introduced for conectvitiy_macos in https://github.com/flutter/plugins/pull/3588


stuartmorgan:
What was the issue here? 0.2.0 is a breaking change from 0.1.0

cyanglaz:
The last PR moved the version all the way to 2.0.0, it should be 0.2.0

stuartmorgan (approved):
Oops, I can't read apparently. I thought it was changing the other way.

PR #3606, opened 2021-02-23
Move plugin tool tests over

blasten:


stuartmorgan (approved):
LGTM++! Thanks for following up with this.

@bparrishMines Once this lands, could you add the tests you had written for `publish --dry-run`?

PR #3612, opened 2021-02-23
Webview not showing HTMLs from documents directory on ios 12.5.1

Vinod-Sutar:
WebView not showing HTMLs from documents directory on iOS 12.5.1 this has to be handled on native level allowing the access to url

Assets HTMLS files are loaded but not from the documents directory.

This does not work on iOS 12.5.1 and most of the iPad Air are upgradable upto iOS 12.5.1

This is very critical issue for our apps.


Vinod-Sutar:
@googlebot I fixed it.

PR #3613, opened 2021-02-23
[webview_flutter] webview not showing html on ios 12.5.1 

Vinod-Sutar:
WebView not showing HTMLs from Documents Directory on iOS 12.5.1  this has to be handled on native level allowing the access to url

All assets HTMLs are showing but not from Documents Directory

This issue is only in 12.5.1 and the iPad Air is only upgradable to iOS 12.5.1

This is a critical issue as most of our users are using iPad and also has reported the content blank issue

### DIFF
--- a/packages/webview_flutter/ios/Classes/FlutterWebView.m
+++ b/packages/webview_flutter/ios/Classes/FlutterWebView.m
@@ -426,7 +426,18 @@
   }
   NSMutableURLRequest* request = [NSMutableURLRequest requestWithURL:nsUrl];
   [request setAllHTTPHeaderFields:headers];
-  [_webView loadRequest:request];
+  if([url hasPrefix:@"file"]) {
+    NSString * allowingReadAccessToURL = [request.URL.path stringByDeletingLastPathComponent];
+    if (@available(iOS 9.0, *)) {
+      [_webView loadFileURL:request.URL allowingReadAccessToURL:[NSURL fileURLWithPath:allowingReadAccessToURL]];
+    }
+    else {
+      [_webView loadRequest:request];
+    }
+  }
+  else {
+    [_webView loadRequest:request];
+  }
   return true;
 }
 


stuartmorgan:
Can you elaborate on this issue being version-specific? There have been separate APIs for local and web requests for quite some time; it seems odd that it would suddenly be enforced only in 12.5.1. Is there a reference for this behavior change (e.g., from Apple)? It would help to file an issue with more details/context and link to the issue from this PR. (That's also how we prioritize reviews, as described in CONTRIBUTING.md)

Also, is this something that can be caught with automated tests?

Vinod-Sutar:
@stuartmorgan 

yes. we are aware about the loading the HTMLs from local and from web.

As the webview_flutter uses the WKWebView in iOS, and if the webview is loading the file-based URL (file://var/..../index.html) then it is needd to invoke the correct native iOS method of WKWebView.
https://developer.apple.com/documentation/webkit/wkwebview/1414973-loadfileurl?language=objc

The current implementation surely works on the latest version of iOS 14.0, 14.1, 14.2, 14.3 and 14.4 but I am also sure that it is not working on my iPad which has iOS 12.5.1 and also one of the user is also having the issue on the iPhone which has 12.5.1

On 12.5.1 the following callback is invoked but not on iOS 14.x and this is solved by checking the **file** prefix before loading the url in webview.

```
onWebResourceError: (error) {
    print("onWebResourceError: ${error.description}"); 

    // This prints: flutter: onWebResourceError: Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted" 
    // UserInfo={_WKRecoveryAttempterErrorKey=<WKReloadFrameErrorRecoveryAttempter: 0x281459a20>, 
    //_NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <4F78FC0D-9528-4229-A5F3-D69D92EEE0AE>.<1>, 
    //_kCFStreamErrorDomainKey=1, _kCFStreamErrorCodeKey=1}
}
```

stuartmorgan:
> As this invokes the onWebResourceError, this can be caught in automated tests!

Great; per the PR checklist that was in the template when you opened the PR, this will need automated tests as part of the change in order to move forward. Please let me know when that's been added and I can review.

Vinod-Sutar:
> Please let me know when that's been added and I can review.

I am not sure what further steps are needed from my side. Please guide.

stuartmorgan:
See https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md for discussion of various kinds of plugin tests; this would need integration tests and/or XCTests that cover this change (i.e., that fail without your change, but pass with it).

stuartmorgan:

PR #3618, opened 2021-02-23
[quick_action] null safety stable release

cyanglaz:
Release the null safety version as stable and move the example app to null safety.

### DIFF
--- a/packages/quick_actions/pubspec.yaml
+++ b/packages/quick_actions/pubspec.yaml
@@ -2,7 +2,7 @@ name: quick_actions
 description: Flutter plugin for creating shortcuts on home screen, also known as
   Quick Actions on iOS and App Shortcuts on Android.
 homepage: https://github.com/flutter/plugins/tree/master/packages/quick_actions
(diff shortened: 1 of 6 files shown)


packages/quick_actions/pubspec.yaml
```diff
@@ -16,17 +16,16 @@ flutter:
 dependencies:
   flutter:
     sdk: flutter
-  meta: ^1.3.0-nullsafety
+  meta: ^1.3.0
 
 dev_dependencies:
-  test: ^1.16.0-nullsafety
-  mockito: ^5.0.0-nullsafety.0
```
stuartmorgan, line 23: That's the easiest case I've seen of removing Mockito dependencies during migration ;)

stuartmorgan (approved):
LGTM

packages/quick_actions/pubspec.yaml
cyanglaz: haha it's just not used as all.

PR #3627, opened 2021-02-25
[wifi_info_flutter] null safety stable

cyanglaz:
Release null safety as stable

### DIFF
--- a/packages/wifi_info_flutter/wifi_info_flutter/example/lib/main.dart
+++ b/packages/wifi_info_flutter/wifi_info_flutter/example/lib/main.dart
@@ -42,7 +42,7 @@ class MyApp extends StatelessWidget {
 }
 
 class MyHomePage extends StatefulWidget {
-  MyHomePage({Key key, this.title}) : super(key: key);
+  MyHomePage({Key? key, required this.title}) : super(key: key);
 
   final String title;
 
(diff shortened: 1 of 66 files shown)


packages/wifi_info_flutter/wifi_info_flutter/example/lib/main.dart
```diff
@@ -103,7 +103,7 @@ class _MyHomePageState extends State<MyHomePage> {
   Future<void> _updateConnectionStatus(ConnectivityResult result) async {
     switch (result) {
       case ConnectivityResult.wifi:
-        String wifiName, wifiBSSID, wifiIP;
+        String? wifiName, wifiBSSID, wifiIP;
```
stuartmorgan, line 106: Are these actually `?`? They look `late`.

stuartmorgan (approved):
LGTM with one question

packages/wifi_info_flutter/wifi_info_flutter/example/lib/main.dart
cyanglaz: Yeah, the getter methods returning these infos returns nullables:
For example in `wifiBSSID = await _wifiInfo.getWifiBSSID()`, _wifiInfo.getWifiBSSID() returns `Future<String?>`

PR #3631, opened 2021-02-25
[file_selector] Migrate to null safety

stuartmorgan:
Migrates the app-facing package to null safety. Includes replacing Mockito with a custom fake/mock.

Fixes an issue where the example didn't handle dialogs being canceled, which was highlighted by the NNBD migration. (Previously, they would cause null assertions at runtime, which wasn't noticed during development. NNBD for the win!)

Fixes https://github.com/flutter/flutter/issues/75235

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/file_selector/file_selector/lib/file_selector.dart
```diff
@@ -10,10 +10,10 @@ export 'package:file_selector_platform_interface/file_selector_platform_interfac
     show XFile, XTypeGroup;
 
 /// Open file dialog for loading files and return a file path
-Future<XFile> openFile({
-  List<XTypeGroup> acceptedTypeGroups,
-  String initialDirectory,
-  String confirmButtonText,
+Future<XFile?> openFile({
+  List<XTypeGroup>? acceptedTypeGroups,
```
cyanglaz, line 14: optional nits: I'd prefer empty list and empty strings as default none values.
But since the platform interface has already landed, it's probably not worth to go back and introduce a breaking change.

cyanglaz (approved):
LGTM

packages/file_selector/file_selector/lib/file_selector.dart
stuartmorgan: We can make this an empty set without changing the platform interface (and that makes it much easier to change the platform interface later); I'll do that now. I should have caught it during initial review but wasn't as familiar with NNBD migrations yet.

For the strings, I don't think empty strings are valid for this API. An empty initial directory is somewhat ambiguous; is that intended to be the device root, for instance? And there's no fundamental reason someone couldn't decide to make the confirm button blank with an empty string (it would be bad UI design, but so are many other valid strings).

Empty collections are I think almost alway going to be the right choice, but empty strings need real thought about whether there's really no difference between empty and null. Here, that's definitely not the case for most of them.

stuartmorgan: (I checked the web implementation, and it already appears to treat empty sets and null the same. This will break two of the unendorsed desktop implementations, but I can fix those trivially.)

PR #3633, opened 2021-02-25
Make executor an instance property

blasten:
This should be an instance property instead.


stuartmorgan (approved):
LGTM, but please add a changelog/version so we can publish (since we just published the previous iteration)

PR #3634, opened 2021-02-25
Fix Cirrus script for firebase testlab tests

stuartmorgan:
Updates the way the script checks for the availability of encrypted variables to account for changes in Cirrus behavior.

Fixes https://github.com/flutter/flutter/issues/76805


cyanglaz (approved):
LGTM

PR #3641, opened 2021-02-26
[camera] Stable release for null safety.

mvanbeusekom:
Stable NNBD release for the camera package. 

With this PR I also migrated the example App to sound null safe. Note that I didn't update the dev-dependency on `mockito: ^5.0.0-nullsafety.7` since the stable version causes a conflict with the current flutter_driver dependency. I have marked the dependency in the pubspec.yaml with a `# TODO(mvanbeusekom)` comment to remind me to remove it when the conflict is resolved.

Conflict details:
```text
Because mockito >=5.0.0 depends on analyzer ^1.0.0 which depends on crypto ^3.0.0, mockito >=5.0.0 requires crypto ^3.0.0.
And because every version of flutter_driver from sdk depends on crypto 2.1.5, mockito >=5.0.0 is incompatible with flutter_driver from sdk.
So, because camera depends on both flutter_driver any from sdk and mockito ^5.0.0, version solving failed.
```

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/example/lib/main.dart
+++ b/packages/camera/camera/example/lib/main.dart
@@ -11,6 +11,10 @@ import 'package:camera/camera.dart';
 import 'package:flutter/material.dart';
 import 'package:video_player/video_player.dart';
 
+import '../../lib/camera.dart';
+import '../../lib/camera.dart';
+import '../../lib/camera.dart';
+
 class CameraExampleHome extends StatefulWidget {
   @override
   _CameraExampleHomeState createState() {
@@ -32,8 +36,13 @@ IconData getCameraLensIcon(CameraLensDirection direction) {
   }
 }
 
-void logError(String code, String? message) =>
+void logError(String code, String? message) {
+  if (message != null) {
     print('Error: $code\nError Message: $message');
+  } else {
+    print('Error: $code');
+  }
+}
 
 class _CameraExampleHomeState extends State<CameraExampleHome>
     with WidgetsBindingObserver, TickerProviderStateMixin {
@@ -46,14 +55,14 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
   double _minAvailableExposureOffset = 0.0;
   double _maxAvailableExposureOffset = 0.0;
   double _currentExposureOffset = 0.0;
-  AnimationController? _flashModeControlRowAnimationController;
-  Animation<double>? _flashModeControlRowAnimation;
-  AnimationController? _exposureModeControlRowAnimationController;
-  Animation<double>? _exposureModeControlRowAnimation;
-  AnimationController? _focusModeControlRowAnimationController;
-  Animation<double>? _focusModeControlRowAnimation;
-  double? _minAvailableZoom;
-  double? _maxAvailableZoom;
+  late AnimationController _flashModeControlRowAnimationController;
+  late Animation<double> _flashModeControlRowAnimation;
+  late AnimationController _exposureModeControlRowAnimationController;
+  late Animation<double> _exposureModeControlRowAnimation;
+  late AnimationController _focusModeControlRowAnimationController;
+  late Animation<double> _focusModeControlRowAnimation;
+  double _minAvailableZoom = 1.0;
+  double _maxAvailableZoom = 1.0;
   double _currentScale = 1.0;
   double _baseScale = 1.0;
 
@@ -102,14 +111,16 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) {
     // App state changed before we got the chance to initialize.
-    if (controller == null || !controller!.value.isInitialized) {
+    if ((controller?.value.isInitialized ?? false)) {
       return;
     }
 
+    final CameraController cameraController = controller!;
+
     if (state == AppLifecycleState.inactive) {
-      controller?.dispose();
+      cameraController.dispose();
     } else if (state == AppLifecycleState.resumed) {
-      onNewCameraSelected(controller!.description);
+      onNewCameraSelected(cameraController.description);
     }
   }
 
@@ -163,7 +174,7 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
 
   /// Display the preview from the camera (or a message if the preview is not available).
   Widget _cameraPreviewWidget() {
-    if (controller == null || !controller!.value.isInitialized) {
+    if (!(controller?.value.isInitialized ?? false)) {
       return const Text(
         'Tap a camera',
         style: TextStyle(
@@ -203,7 +214,7 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
     }
 
     _currentScale = (_baseScale * details.scale)
-        .clamp(_minAvailableZoom ?? 1, _maxAvailableZoom ?? 1);
+        .clamp(_minAvailableZoom, _maxAvailableZoom);
 
     await controller!.setZoomLevel(_currentScale);
   }
@@ -290,8 +301,8 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
   }
 
   Widget _flashModeControlRowWidget() {
-    return _wrapInSizeTransition(
-      animation: _flashModeControlRowAnimation!,
+    return SizeTransition(
+      sizeFactor: _flashModeControlRowAnimation,
       child: ClipRect(
         child: Row(
           mainAxisAlignment: MainAxisAlignment.spaceEvenly,
@@ -351,8 +362,8 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
           : Colors.blue,
     );
 
-    return _wrapInSizeTransition(
-      animation: _exposureModeControlRowAnimation!,
+    return SizeTransition(
+      sizeFactor: _exposureModeControlRowAnimation,
       child: ClipRect(
         child: Container(
           color: Colors.grey.shade50,
@@ -429,8 +440,8 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
           : Colors.blue,
     );
 
-    return _wrapInSizeTransition(
-      animation: _focusModeControlRowAnimation,
+    return SizeTransition(
+      sizeFactor: _focusModeControlRowAnimation,
       child: ClipRect(
         child: Container(
           color: Colors.grey.shade50,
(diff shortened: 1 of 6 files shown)


packages/camera/camera/example/lib/main.dart
```diff
@@ -27,32 +27,33 @@ IconData getCameraLensIcon(CameraLensDirection direction) {
...
       return Icons.camera;
+    default:
+      throw ArgumentError('Unknown lens direction');
   }
-  throw ArgumentError('Unknown lens direction');
 }
 
-void logError(String code, String message) =>
+void logError(String code, String? message) =>
     print('Error: $code\nError Message: $message');
```
stuartmorgan, line 36: If we have cases where there's no `message`, the format of this should probably conditionalize showing the second line. (Alternately, can we always pass a message?)

packages/camera/camera/example/lib/main.dart
```diff
@@ -27,32 +27,33 @@ IconData getCameraLensIcon(CameraLensDirection direction) {
...
-  AnimationController _focusModeControlRowAnimationController;
-  Animation<double> _focusModeControlRowAnimation;
-  double _minAvailableZoom;
-  double _maxAvailableZoom;
+  AnimationController? _flashModeControlRowAnimationController;
+  Animation<double>? _flashModeControlRowAnimation;
+  AnimationController? _exposureModeControlRowAnimationController;
+  Animation<double>? _exposureModeControlRowAnimation;
+  AnimationController? _focusModeControlRowAnimationController;
+  Animation<double>? _focusModeControlRowAnimation;
```
stuartmorgan, line 54: It looks like all of these could be `late` since they are initialized in `initState`.

packages/camera/camera/example/lib/main.dart
```diff
@@ -62,53 +63,53 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
...
-    _exposureModeControlRowAnimationController.dispose();
+    WidgetsBinding.instance?.removeObserver(this);
+    _flashModeControlRowAnimationController?.dispose();
+    _exposureModeControlRowAnimationController?.dispose();
     super.dispose();
   }
 
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) {
     // App state changed before we got the chance to initialize.
```
stuartmorgan, line 104: Let's set a local variable for `controller` as the first line and use it everywhere after this, so we don't need the `!` and `?` everywhere.

packages/camera/camera/example/lib/main.dart
```diff
@@ -161,7 +163,7 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
 
   /// Display the preview from the camera (or a message if the preview is not available).
   Widget _cameraPreviewWidget() {
-    if (controller == null || !controller.value.isInitialized) {
+    if (controller == null || !controller!.value.isInitialized) {
```
stuartmorgan, line 166: Same here.

packages/camera/camera/example/lib/main.dart
```diff
@@ -196,14 +198,14 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
...
   Future<void> _handleScaleUpdate(ScaleUpdateDetails details) async {
     // When there are not exactly two fingers on screen don't scale
-    if (_pointers != 2) {
+    if (controller == null || _pointers != 2) {
       return;
     }
 
     _currentScale = (_baseScale * details.scale)
-        .clamp(_minAvailableZoom, _maxAvailableZoom);
+        .clamp(_minAvailableZoom ?? 1, _maxAvailableZoom ?? 1);
```
stuartmorgan, line 206: If we always need values, then lets make them non-nullable and initialized to 1 (and if the thing they are set to later is nullable, add a `?? 1` at that point).

packages/camera/camera/example/lib/main.dart
```diff
@@ -466,6 +470,20 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
...
+    }
+
+    return SizeTransition(
+      sizeFactor: animation,
+      child: child,
+    );
+  }
+
   /// Display the control bar with buttons to take pictures and record videos.
   Widget _captureControlRowWidget() {
```
stuartmorgan, line 488: Use a local `controller` copy here as well.

packages/camera/camera/example/lib/main.dart
```diff
@@ -547,21 +574,25 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
 
   void showInSnackBar(String message) {
     // ignore: deprecated_member_use
-    _scaffoldKey.currentState.showSnackBar(SnackBar(content: Text(message)));
+    _scaffoldKey.currentState?.showSnackBar(SnackBar(content: Text(message)));
   }
 
   void onViewFinderTap(TapDownDetails details, BoxConstraints constraints) {
+    if (controller == null) {
```
stuartmorgan, line 581: Local controller.

stuartmorgan (changes requested):

packages/camera/camera/pubspec.yaml
```diff
@@ -2,26 +2,31 @@ name: camera
...
     sdk: flutter
-  mockito: ^5.0.0-nullsafety.5
-  plugin_platform_interface: ^1.1.0-nullsafety.2
+
+  # TODO(mvanbeusekom): Update to stable 5.0.0 release when dependency conflict
+  # with flutter_driver has been resolved:
+  # Because mockito >=5.0.0 depends on analyzer ^1.0.0 which depends on crypto ^3.0.0 
+  # every version of flutter_driver from sdk depends on crypto 2.1.5.
+  mockito: ^5.0.0-nullsafety.7
+  plugin_platform_interface: ^2.0.0
```
stuartmorgan, line 29: I think since this is the last plugin to be published as stable, this will work. Let's see how our bots handle it.

(Ideally we can now sweep the repo to make everything require 2.0.0 instead of a range that includes a non-nullable version.)

mvanbeusekom:
@stuartmorgan thanks for the feedback. I have gone through all your comments and made the necessary adjustments.

Could you have another look please?

packages/camera/camera/example/lib/main.dart
```diff
@@ -100,15 +107,16 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) {
     // App state changed before we got the chance to initialize.
-    if (controller == null || !controller.value.isInitialized) {
+    if ((controller?.value.isInitialized ?? false)) {
       return;
     }
+
+    final CameraController cameraController = controller!;
```
stuartmorgan, line 114: If you put this before the early-return, and use `cameraController` in the condition above, you can keep the logic as-is without any additional `!` or `?` or `??`.

packages/camera/camera/example/lib/main.dart
```diff
@@ -100,15 +107,16 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) {
     // App state changed before we got the chance to initialize.
-    if (controller == null || !controller.value.isInitialized) {
+    if ((controller?.value.isInitialized ?? false)) {
```
stuartmorgan, line 110: You've changed the meaning of this condition; it will now return early if `controller` is non-null and initialized, which is exactly the opposite of the intent.

packages/camera/camera/example/lib/main.dart
```diff
@@ -218,15 +227,15 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
...
                             child: Center(
                               child: AspectRatio(
                                   aspectRatio:
-                                      videoController.value.size != null
-                                          ? videoController.value.aspectRatio
+                                      videoController!.value.size != null
+                                          ? videoController!.value.aspectRatio
                                           : 1.0,
-                                  child: VideoPlayer(videoController)),
+                                  child: VideoPlayer(videoController!)),
```
stuartmorgan, line 238: Consider a local video controller variable to eliminate the need for these `!`s

packages/camera/camera/example/lib/main.dart
```diff
@@ -468,49 +479,52 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
 
   /// Display the control bar with buttons to take pictures and record videos.
   Widget _captureControlRowWidget() {
+    VoidCallback? takePictureButton;
+    VoidCallback? recordButtonPressed;
+    VoidCallback? pauseOrResumeButtonPressed;
+    VoidCallback? stopButtonPressed;
+
+    if (controller != null) {
+      final CameraController cameraController = controller!;
```
stuartmorgan, line 488: Here as well, you can do this first (as `CameraController?`), use it in the condition, and then everything else will automatically know that it's non-null.

packages/camera/camera/example/pubspec.yaml
```diff
@@ -9,23 +9,24 @@ dependencies:
...
 
 dev_dependencies:
   flutter_test:
     sdk: flutter
   flutter_driver:
     sdk: flutter
-  pedantic: ^1.8.0
+  integration_test:
+    path: ../../../integration_test
+  test: any
```
stuartmorgan, line 24: Was this to resolve a conflict? We don't usually use `any` unless we have to; I would expect `1.16.0` (or 1.16.3 if you need `Fake`).

stuartmorgan (changes requested):
A few more nits, but otherwise looks good.

packages/camera/camera/example/lib/main.dart
mvanbeusekom: The "(as `CameraController?`)" was the magic hint I needed. Now I understand that the analyser can figure out that a null check is done when dealing with local variables. 

Thanks for this feedback!

packages/camera/camera/example/pubspec.yaml
mvanbeusekom: This was actually a mistake on my end. I added this dependency because I was receiving a CI failure with the dependencies related to the `integration_test` in combination with running in NNBD. I started comparing with the `video_player` plugin and noticed that this dependency was included this way and decided to give that a try. Reading back in the Discord channel I realised that we shouldn't run integration_tests in NNBD mode (and add the `// @dart = 2.9` instruction to the top of the file). Unfortunately I forgot to remove this dependency which I did this time ;).

stuartmorgan (approved):
LGTM!

PR #3645, opened 2021-02-26
[connectivity] migrate unit tests to null safety

cyanglaz:
Migrate unit tests to null safety. This is mistakenly left out in the 3.0.0 release.


stuartmorgan (approved):
LGTM, good catch!

I was going to do another sweep for `@dart` and old-mininum `sdk`s on Friday, but forgot.

PR #3650, opened 2021-03-01
Update plugin_platform_interface min version

stuartmorgan:
To avoid intra-repo plugin conflicts during the NNBD stable migration,
`plugin_platform_interface` allowed either 1.x or 2.0. However, 1.0.x
isn't null-safe so this can cause apps that don't have all their
packages fully updated can fail to run in strong mode (due to having an
old local `plugin_platform_interface`).

Now that everything has been updated, we can bump all the minimums so
that people updating their plugins will get new versions of the
dependency.

### DIFF
--- a/packages/url_launcher/url_launcher/example/pubspec.yaml
+++ b/packages/url_launcher/url_launcher/example/pubspec.yaml
@@ -19,7 +19,7 @@ dev_dependencies:
     sdk: flutter
   pedantic: ^1.10.0
   mockito: ^5.0.0-nullsafety.7
-  plugin_platform_interface: ">=1.0.0 <3.0.0"
+  plugin_platform_interface: ^2.0.0
 
 flutter:
   uses-material-design: true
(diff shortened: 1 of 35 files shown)


packages/url_launcher/url_launcher/example/pubspec.yaml
```diff
@@ -19,7 +19,7 @@ dev_dependencies:
     sdk: flutter
   pedantic: ^1.10.0
   mockito: ^5.0.0-nullsafety.7
-  plugin_platform_interface: ">=1.0.0 <3.0.0"
+  plugin_platform_interface: ^2.0.0
```
ditman, line 22: This is the odd-one out, for those looking to why a change that should have an even number of files changed has touched 35 files :P

ditman (approved):
LGTM!

mvanbeusekom:
LGTM!

mvanbeusekom (approved):

PR #3562, opened 2021-02-18
[camera] Solves delay when using the zoom feature on iOS.

mvanbeusekom:
Initial version of the zoom feature used the `rampToVideoZoomFactor:withRate` method on iOS to update the zoom scale. This method introduces smooth transition at the specified rate, which is ideal when jumping in large zoom increments (e.g. directly zooming from factor 1 to factor 5). 

However in most use cases you would zoom in very small increments (using pinch to zoom or a slider). Using the `rampToVideoZoomFactor:withRate` method would cause an annoying delay in zoom functionality. Especially since the `rate` parameter is hardcoded since it is not cross platform compatible. Therefore this PR updates the iOS implementation to use the `videoZoomFactor` property instead which jumps directly to the desired zoom level. 

Added benefit is that this now behaves the same way on Android and iOS. It should also be relative easy to implement a smooth transition in cases which require jumping in large zoom increments by breaking the jump down in small increments and send them one by one over a certain period of time.

- Fixes flutter/flutter#74350

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#style
[...]

PR #3411, opened 2021-01-13
[camera] Fixes crash when taking a picture on iOS devices without flash

mvanbeusekom:
The current version of the camera plugin crashes when taking a picture on iOS devices that don't have support for flash. This is because the camera is automatically initialised with the `FlashModeAuto` value. 

This PR adds an additional check which initialises the `_flashMode` to `FlashModeAuto` if the capture device has flash capabilities and otherwise initialises the `_flashMode` too `FlashModeOff`.

- Resolves flutter/flutter#73726

## Pre-launch Checklist

- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

PR #3651, opened 2021-03-01
[camera] Fix all capture timeout, autofocus, flash, exposure, and device crash issues

acoutts:
This PR is a full rework of the android plugin. There were a number of issues before, including crashes on some legacy devices and some features just didn't work. I found that the camera API wasn't quite implemented correctly, and it wasn't transitioning between the states / modes as it should have. This led to instabilities on some devices.

I've tested this PR on the following devices and every feature in the plugin now works as expected:
* Pixel 4
* LG k50s
* Samsung Galaxy S7
* Sony XZ (F8331)

Summary of major changes / improvements:
* Added comments throughout the plugin detailing what each part is doing.
* Fixed capture timeout on devices, related to auto exposure / focus.
* Added correct handling for starting/stopping auto exposure, to not crash on cameras with unsupported modes.
* Added correct handling to stop/start autofocus when switching between focus lock and autofocus. Tapping the lock button now triggers a one-time autofocus of the current point/scene, and then locks it until later unlocked.
* Added support for `NV21` image streams in Android for more efficient use with Firebase ML Vision- no need to concatenate the YUV420 planes anymore, just feed the one plane from NV21 directly to ML Vision.
* Fixed flickering issue on Sony front facing cameras when capturing photos and changing AE mode.
* Moved capture callbacks / image processing to run on a background handler/thread.
* Instead of managing the camera state in `PictureCaptureRequest`, we now have a new `CameraState` as well as the `PictureCaptureRequestState`. The `PictureCaptureRequestState` handles timing out captures while the `CameraState` tracks the state of the camera as it's processing the current request. This made it easier to keep track of how the camera is moving between states during capture.

## Testing this PR
In order to use this PR before the update to `camera` and `camera_platform_interface` are published, you need to include the following in your `pubspec.yaml`:
```yaml
  camera:
    git:
      url: https://github.com/bottlepay/plugins.git
      path: packages/camera/camera
      ref: android-rework
```

## Future improvements
There is still a delay when capturing images on some devices. I believe it's because we are blocking the UI thread and freezing flutter, so the image preview freezes. I think what we can do to improve this is when calling take picture from android, we need to return a completer future so we can immediately call the result function to unblock flutter, and then once the image is processed we can complete it to return the image path back to dart.
The same could be applied to starting/stopping video capture too.

Related issues:
* https://github.com/flutter/flutter/issues/76782 (potentailly)
* https://github.com/flutter/flutter/issues/76498
* https://github.com/flutter/flutter/issues/76311
* https://github.com/flutter/flutter/issues/76305
* https://github.com/flutter/flutter/issues/75641
* https://github.com/flutter/flutter/issues/75506
* https://github.com/flutter/flutter/issues/75133
* https://github.com/flutter/flutter/issues/75208
* https://github.com/flutter/flutter/issues/73561
* https://github.com/flutter/flutter/issues/73504

## Pre-launch Checklist

- [x] I signed the [CLA].


acoutts:
cc @mvanbeusekom

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/types/FlashMode.java
```diff
@@ -4,6 +4,16 @@
...
 
+import android.hardware.camera2.CameraCaptureSession;
+import android.hardware.camera2.CameraCharacteristics;
+import android.hardware.camera2.CaptureFailure;
+import android.hardware.camera2.CaptureRequest;
+import android.hardware.camera2.TotalCaptureResult;
+
+import androidx.annotation.NonNull;
+
+import io.flutter.plugin.common.MethodChannel;
```
nt4f04uNd, line 15: these imports look like unused

nt4f04uNd, line 15: these imports look like unused

acoutts: Fixed

mvanbeusekom:
@acoutts first of all thank you for this contribution, at first glance this looks very promising and would like to work with you to land this PR. 

To start I would like to ask you to split the PR into two separate pull requests? One PR updating the `camera_platform_interface` containing support for the new NV21 format and a second PR for the `camera` plugin? Unfortunately we need to first land the the `camera_platform_interface` before we can continue with the `camera` package as the tree will never turn green otherwise. It will also make it a little bit easier for me to review the code ;).

Also I noticed that the formatting is giving some problems, this should be relatively easy to fix using the `flutter_plugin_tools` as described [here](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#contributing-code).

acoutts:
Great! Just submitted this PR for `camera_platform_interface`:
https://github.com/flutter/plugins/pull/3656

I also used the guide you linked to fix the java formatting.

acoutts:
I'm really curious if you have any ideas on how we can fix the shutter delay. The pixel 4 is intriguing because it's a very new / fast device. The stock camera app has no lag between tapping the shutter and image being available to view.

But from studying closely, there is a 1-2s delay between when submitting the capture request and when the capture request is ready. Writing to jpeg file only takes about 20ms so it's not that either.
```java
      captureSession.capture(stillBuilder.build(), captureCallback, mBackgroundHandler);
```

Immediately after this, but before `onCaptureStarted` within `captureCallback` is called, there's about a 2s delay.

acoutts:
I improved the delay a bit by only requesting 1 image in the ImageReaders, which in theory should reduce the memory allocated for the capture (and image stream in real time). We only need 1 because we only ever do a single capture.

mvanbeusekom:
@acoutts will you also update this PR to remove the NV21 format logic (since you found it is not supported)?

acoutts:
Alright that's cleaned up now.

stuartmorgan:

Where are the new tests for issues fixed by this PR? I just see some remaking of constants in existing tests; there should also be tests for your fixes (i.e., tests that fail without your PR but pass with it) to ensure that they don't regress in the future.

acoutts:
> > * [x]   I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
> 
> Where are the new tests for issues fixed by this PR? I just see some remaking of constants in existing tests; there should also be tests for your fixes (i.e., tests that fail without your PR but pass with it) to ensure that they don't regress in the future.

I'm not sure how to approach this-  Can you give some specific examples of what else needs to be tested? Do we need to test that the camera 2 api works as it's specified in its documentation? I'd be happy to add some coverage, just not sure what we need.

For what it's worth, the iOS lib has **not a single test** and that has never been raised as an issue for any of the prior PRs:
https://github.com/flutter/plugins/pull/3562
https://github.com/flutter/plugins/pull/3411
https://github.com/flutter/plugins/pull/3390
etc

Screen Shot 2021-03-02 at 9 05 46 AM

stuartmorgan:
> Can you give some specific examples of what else needs to be tested?

As much as reasonably possible. You've listed a number of explicit fixes, and referenced ten issues you believe to be related to this change. As one example, you've mentioned fixing a crash for certain unsupported modes; can you write a test that triggers that codepath and ensures that it doesn't crash?

> Do we need to test that the camera 2 api works as it's specified in its documentation?

You shouldn't be testing the Android code, you should be testing the code you wrote. E.g., you have at least one state machine in this PR that's plugin code, not Android code. It should be relatively easy to write tests for a state machine, and doing so ensure that it correctly handles all the transitions you expect and that it doesn't regress with future changes.

There may be aspects of the fixes that you can't test due to specific hardware interactions. If that's the case, then those parts may need to be untested (although consider whether mocking out the Android APIs and simulating certain kinds of failures can exercise the code you have to handle them); it's not ideal, but even when we can't do perfect testing we should be doing what testing we can.

> For what it's worth, the iOS lib has **not a single test** and that has never been raised as an issue for any of the prior PRs

Which is problematic, and something I hope @BeMacized and @mvanbeusekom will watch for more carefully going forward (and if there's no native test harness set up for the iOS camera code, that should definitely be a priority). It's [documented Flutter policy](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#write-test-find-bug), and it's in the PR checklist for that reason.

While I understand the point you are trying to make, there is a goal for flutter/plugins to improving testing, and I've never yet seen a project's testing situation improve when the standard for new PRs is "no worse than the worst we've previously done".

lukepighetti:
Improve testing or force people to contribute to https://github.com/fluttercommunity/plus_plugins ?

stuartmorgan:
@lukepighetti There's no Plus Plugins version of camera, and the Plus Plugins roadmap lists the following as the very first focus area of the reviews/updates they are doing of the plugins they are working on:

> Testing: covering the public API (as much as possible) with the dedicated unit and integration tests is essential. We can be confident in adding new features, squashing bugs, and opening up the repository to more external contributors.

So I'm not sure what point you think you are making here.

lukepighetti:
Ask yourself this question: are your triage requests reasonable and in good faith? If the answer is yes, please, by all means, carry on. I've been watching this repo from the community side of things long enough to believe that your requests are unreasonable, and in my opinion, in bad faith.

lukepighetti:
I'm concerned that PRs are not getting the triage/review effort they deserve because there is an effort to push flutter/plugins to community plus. I'm concerned that this PR is being stonewalled, perhaps unintentionally.

mvanbeusekom:
I agree with @stuartmorgan that testing will eventually improve the quality of the plugin. I agree it is hard and I also noticed that some PRs are being abandoned because of this requirement. But I am willing to help out where possible and see these changes covered and eventually landed.

I will also be focusing on getting a test bed ready for the iOS side (as we have been doing on the Dart side and started on Java).

acoutts:
Can anyone help answer how you to about mocking the android `CameraCharacteristics` behavior?

Screen Shot 2021-03-02 at 12 34 46 PM

stuartmorgan:
@lukepighetti I hope you deleted your other comment because you understand that it was not at all consistent with Flutter's code of conduct. If not, please take a moment to review it; it applies to all comments posted in Flutter repositories, even those that are deleted after triggering email notifications.

> I'm concerned that PRs are not getting the triage/review effort they deserve because there is an effort to push flutter/plugins to community plus.

Steering developers toward contributing to Plus Plugins instead of here looks like [this](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#important-note) and [this](https://github.com/flutter/plugins/pull/3384). Hopefully that clarifies things.

> I'm concerned that this PR is being stonewalled, perhaps unintentionally.

It is not generally my experience that a PR that's making hundreds of lines of changes and fixing 10+ issues in a codebase that has a testing framework in place cannot reasonably include *any* meaningful new tests, so I don't consider it stonewalling to ask about testing. If this particular case is an outlier, I'm certainly willing to listen to arguments for why from the PR author.

It is in fact common in Flutter repositories to ask for PRs to add tests, so I'm not sure why you assume that a routine process constitutes stonewalling in the first place. Feel free to peruse flutter/flutter for many, many examples of reviewers asking for tests (and for those PRs getting tests, and then being landed).

> I am concerned that if reviews keep looking like this less people will contribute.

It is certainly the case that some people abandon PRs rather than bring them up to the standards that the Flutter project lays out for PRs in its contribution guidelines. (It's also the case that some people abandon PRs rather than responding to literally any review feedback, but I suspect you would not argue that we should simply accept every PR as-is because of that.) I certainly hope that won't be the case here.

> I am concerned that this is net loss for flutter/plugins and the community.

In some cases, contributions may be lost. It is also the case that if we have no testing of anything, plugins will be brittle, regressions will be extremely common, and people will have no confidence that they can update to newer versions of plugins; this would also be a significant loss. The Flutter project has made specific choices about the tradeoffs, and laid them out in its contribution guidance. We're actively working to bring flutter/plugins into better alignment with the current Flutter standards.

If you want to continue this discussion, perhaps Discord would be a better option; it doesn't seem particularly related to this PR, and we shouldn't derail the efforts to get this PR reviewed with meta-discussion.

mvanbeusekom:
> Can anyone help answer how you to about mocking the android `CameraCharacteristics` behavior?
> 
> 
> 
> Screen Shot 2021-03-02 at 12 34 46 PM
> 
> 

The only approach I can think of is using the Roboelectric test runner. This will let you access the `CameraCharacteristic` constants.

You can have a look at an example in the [CameraZoomTest.java](https://github.com/flutter/plugins/blob/master/packages/camera/camera/android/src/test/java/io/flutter/plugins/camera/CameraZoomTest.java).

acoutts:
Aha! That did it. Ok let me write some tests and we can review again.

acoutts:
Unfortunately this is just beyond my abilities to sort out for the tests. I'm not an android developer, but I was inconvenienced enough by the published package being broken that I spent my weekend learning the camera2 api and rewriting this until it worked on all of my devices.

The state machine in here is going to be tough to test because the behavior of the state machine running is handled by android. If you can test that you can set it to state B, you still don't actually trigger that change and whether or not it does move depends on different hardware implementations, so the effectiveness of testing will be limited here.

Some other quirks:
* when sending `CONTROL_AF_TRIGGER_CANCEL`, some devices reset their focus to infinity, and some keep it at their last position. Unit testing won't provide much value here.
* One device would set its noise reduction to the max setting by default, causing all of the images and preview to look blurry, so this had to be accounted for.
* Some devices never achieve focus lock, but instead send a `CONTROL_AF_STATE_PASSIVE_SCAN` signal which we have to interpret as acceptable. It again comes down to how the OEMs implement the camera API and what signals they send.
* Some devices return null for some camera characteristics. The only way to be sure is to check the docs on every one, and test it on physical devices by hand.
* Passing invalid parameters to the capture session like an invalid surface, resolution, ae mode, focus mode, etc- will simply crash at runtime. Testing will be hard for this too.

stuartmorgan:
@mvanbeusekom Perhaps you could partner on building out tests here?

I haven't had a chance to look closely at the details here, but some high-level thoughts on the cases you've described here:

> The state machine in here is going to be tough to test because the behavior of the state machine running is handled by android. If you can test that you can set it to state B, you still don't actually trigger that change and whether or not it does move depends on different hardware implementations, so the effectiveness of testing will be limited here.

This sounds like a case where ideally you'd write tests that create various synthetic event streams—corresponding to actual things seen from different hardware—and ensuring that the state machine is in the expected state. That ensures that future changes that would regress only certain hardware would be caught in unit tests.

> * when sending `CONTROL_AF_TRIGGER_CANCEL`, some devices reset their focus to infinity, and some keep it at their last position. Unit testing won't provide much value here.
> * One device would set its noise reduction to the max setting by default, causing all of the images and preview to look blurry, so this had to be accounted for.

The goal here should be to have tests that the calls that mitigate this are all actually made, with comments on the test explaining why. This ensures that, e.g., someone doesn't "optimize" that code later by removing calls that seem unnecessary in their own testing because they are on different hardware.

> * Some devices never achieve focus lock, but instead send a `CONTROL_AF_STATE_PASSIVE_SCAN` signal which we have to interpret as acceptable. It again comes down to how the OEMs implement the camera API and what signals they send.

Similar to the above, the idea here would be to make a test that synthetically sends that to the plugin code, and ensure that the plugin code does in fact interpret it as acceptable.

> * Some devices return null for some camera characteristics. The only way to be sure is to check the docs on every one, and test it on physical devices by hand.

And here, to make sure that for known cases where this happens, there are tests that ensure that if the plugin code receives null, the right things happen.

> * Passing invalid parameters to the capture session like an invalid surface, resolution, ae mode, focus mode, etc- will simply crash at runtime. Testing will be hard for this too.

If you've identified cases where invalid parameters were being sent, then the goal would be tests that trigger those codepaths and use mocks to ensure they only sending valid parameters.


In all cases, the goal is to make unit tests which isolate the code you control and feed it inputs that match various real-world hardware scenarios so that we know that changes aren't going to regress your fixes on hardware that people writing later PRs haven't tested against.

mvanbeusekom:
I would be happy to help. I can start tomorrow morning (I am Central European time).

@acoutts could you give my write access to this PR if you'd like my assistance?

acoutts:
> I would be happy to help. I can start tomorrow morning (I am Central European time).
> 
> @acoutts could you give my write access to this PR if you'd like my assistance?

Excellent! Thanks a lot. Let me sort that for you now.

acoutts:
@mvanbeusekom should be in your inbox now.

mvanbeusekom:
Yes got it, will get on it tomorrow. Thanks

Petrol-:
Hello, thank you for this amazing PR.

I tried your work using the example app in the camera package and with the following devices:

- OnePlus Nord 5G
- Samsung Galaxy S7 Edge (europe)
- Huawei P20 Lite

It is a very big improvement ! I see a lot less camera timeouts.

The FlashMode does not seem to affect the capture anymore (in the previous version, I had a lot of Catpure timeout when setting the FlashMode to auto / always).

When taking a photo while shaking the phone (so it can't focus), I had a few timeout on my Galaxy S7 and Huawei P20 Lite. (3 - 4 exceptions in a 30min tests sessions). 


In the example app, we can set the Exposure Offset, which always throws the following exception:

`Null check operator used on a null value`
```
#0      MethodChannelCamera.setExposureOffset (package:camera_platform_interface/src/method_channel/method_channel_camera.dart:338:25)
<asynchronous suspension>
#1      _CameraExampleHomeState.setExposureOffset (package:new_camera/main.dart:728:16)
<asynchronous suspension>
```

I've never try to set the exposure offset using the current camera (0.7.x) so I don't know if it's a new problem.


I have a colleague who owns a Redmi phone (don't remember the exact model). But this phone is totally unable to take any photo with camera 0.7.x. (with or without flash / autofocus)
I'll send him a sample app so he can test this PR.

Thank again !

acoutts:
@Petrol- great feedback! Thank you. I’m really curious how the xiaomi redmi works on it. We had a lot of customers with that phone unable to use our app because the camera didn’t work at all on the 0.7/0.8 versions of the plugin.

For your timeouts, did you run in profile / release mode as well? There is a 5s timeout right now as a sane limit.

I noticed there’s a capture delay on the pixel 4 too. My final conclusion is it’s the camera2 api. Using print statements I found the delay is between when we submit the capture request to the camera manager, and when it actually starts the capture sequence. I guess it’s a black box there when it’s allocating the buffers and warming up. We will need to get creative for this. 

Even capturing as YUV420 didn’t affect the delay, so it’s not in the jpeg processing. The delay isn’t as bad on other phones but they all seem to have one. 

What really gets me is there must be some way we can make it faster. I assume the stock camera app on the pixel 4 uses the same underlying api so they’ve done something to make it fast.

acoutts:
Even open camera has about a 1s delay on the shutter for the pixel 4. 
https://opencamera.org.uk/

nt4f04uNd:
i'm confused, i wanted to try this out, but i'm getting
```
Package doesn't exist (the pubspec for camera 0.8.1 from git has version 0.8.0-nullsafety.3).
pub get failed (server unavailable) -- attempting retry 1 in 1 second...
```

acoutts:
@nt4f04uNd try this:
```yaml
  camera:
    git:
      url: https://github.com/bottlepay/plugins.git
      path: packages/camera/camera
      ref: 60f7eeb5099e63a0e8c598f6676d11a7c1879ed4
```

nt4f04uNd:
i tested and there were some troubles on my Redmi Note 5, i will tell you results of my tests later

prior to this i had to fix the camera example app itself, because it was poor written
i PRed it to your fork https://github.com/bottlepay/plugins/pull/1

acoutts:
Thanks @nt4f04uNd I'll give that a look.

Just tested on a Mi A2 Lite, and it seems like it's running as legacy support for camera2. I'm getting the following warnings printing:
```
W/LegacyRequestMapper(19663): convertRequestMetadata - control.awbRegions setting is not supported, ignoring value
W/LegacyRequestMapper(19663): convertRequestToMetadata - Ignoring unsupported noiseReduction.mode = 0
W/LegacyRequestMapper(19663): convertRequestMetadata - control.awbRegions setting is not supported, ignoring value
W/LegacyRequestMapper(19663): convertRequestToMetadata - Ignoring unsupported noiseReduction.mode = 0
```

So we need to check that awb / noise reduction are supported before trying to set those on the capture request.

When I record a video, the preview freezes on this device too, I suspect it's something with legacy devices. The following exceptions printed after capturing a video:
```
E/ExoPlayerImplInternal(19663): Playback error
E/ExoPlayerImplInternal(19663):   com.google.android.exoplayer2.ExoPlaybackException: Source error
E/ExoPlayerImplInternal(19663):       at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:554)
E/ExoPlayerImplInternal(19663):       at android.os.Handler.dispatchMessage(Handler.java:103)
E/ExoPlayerImplInternal(19663):       at android.os.Looper.loop(Looper.java:214)
E/ExoPlayerImplInternal(19663):       at android.os.HandlerThread.run(HandlerThread.java:67)
E/ExoPlayerImplInternal(19663):   Caused by: com.google.android.exoplayer2.source.UnrecognizedInputFormatException: None of the available extractors (FragmentedMp4Extractor, Mp4Extractor, FlvExtractor, FlacExtractor, WavExtractor, AmrExtractor, PsExtractor, OggExtractor, TsExtractor, MatroskaExtractor, AdtsExtractor, Ac3Extractor, Ac4Extractor, Mp3Extractor) could read the stream.
E/ExoPlayerImplInternal(19663):       at com.google.android.exoplayer2.source.BundledExtractorsAdapter.init(BundledExtractorsAdapter.java:92)
E/ExoPlayerImplInternal(19663):       at com.google.android.exoplayer2.source.ProgressiveMediaPeriod$ExtractingLoadable.load(ProgressiveMediaPeriod.java:1024)
E/ExoPlayerImplInternal(19663):       at com.google.android.exoplayer2.upstream.Loader$LoadTask.run(Loader.java:415)
E/ExoPlayerImplInternal(19663):       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
E/ExoPlayerImplInternal(19663):       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
E/ExoPlayerImplInternal(19663):       at java.lang.Thread.run(Thread.java:919)
E/flutter (19663): [ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception: PlatformException(VideoError, Video player had error com.google.android.exoplayer2.ExoPlaybackException: Source error, null, null)
E/flutter (19663): 
I/CameraDeviceState(19663): Legacy camera service transitioning to state CAPTURING
W/System  (19663): A resource failed to call release. 
I/chatty  (19663): uid=10157(io.flutter.plugins.cameraexample) FinalizerDaemon identical 3 lines
W/System  (19663): A resource failed to call release. 
```

No video appears to have saved.

nt4f04uNd:
i bet you didn't see this coming ;D

here's what i found broken on my redmi note 5

### syntax 
`N.` `[L?]` `[NV?]` `[cameraType: back | front | all]` ***topic***: description

* for each issue, see for video with `N` name in the https://drive.google.com/drive/folders/1urdiT5YF4rnz4tTLOCRJxhHOqxo6Kfjm?usp=sharing, except for marked as [NV]
* if issue has `[L]`, then search for logs with `N` name
* i tested each issue with both cameras so: 
     * `back` only applies to back camera
     * `front` only applies to front camera
     * `both` - you got me

### issues

1. [back] ***`FocusMode`***: `auto` and locked are messed up
2. [back] ***`FocusMode`***: actual `auto` (i.e. now `locked`) is constantly trying to find a focus point without success
3. [back] ***`FocusMode`***: actual `locked` (i.e. now `auto`) cannot refocus on tap
4. [back] ***`ExposureMode`***: `locked` cannot refocus on tap
5. [all] ***`FlashMode`***: `torch` works but when you `takePicture`, it will turn off take the shot, save the not lightened black image, and then turn on again
6. [back] ***`FlashMode`***: camera starts flashing if you shot was made with `FlashMode.auto` or `FlashMode.always` (or if you take a picture that will prevent the camera shot in future (see issue [9]) and then try to change to these modes). if it goes to this state switching back to any other mode does nothing, it continues flashing like beacon. but the shot will end up be lightened 🙃
7. [back] ***`FlashMode`***: default flash mode after camera creation is not applied (on the video first time i just `takePicture`, then re-init the camera -> tap the same default `FlashMode.auto` -> `takePicture`)
8. ***generic***: sometimes camera fails to initialize on fast taps
9. `[L]` `[NV]` [back] ***`takePicture`***: `takePicture` only works first time for a single controller, fails on subsequent calls BUT if you
   - `takePicture` (successfully)
   - do recording (start, stop)
   - then you sometimes will be able to do takePicture again 🤔
10. [all] ***orientation***: auto orientation doesn't work - shots and records are always in portrait mode
11. `[NV]` [front] `FocusMode` and `ExposureMode` not refocusing on taps. actually, Redmi Note 5 has a fixed focus position for front camera, and when you tap somewhere on screen in default camera app, it only adjusts exposure, but because `ExposureMode` isn't working, i suppose on devices that support that
it won't be working. with `ExposureMode.auto` is sometimes works but really really rare (like when the exopure change is too big it simply performs automatic ajustment on tap)
12. [front] ***`FlashMode`***:
   * `auto` -> `always` if auto was turned on - works as expected
   * `auto` -> `always` if auto was turned off - stays turned off
   * `torch` -> `always` - turns off
13. `[L]` ***generic***: crash happens if you fast spam taps onto photo and recording buttons

### logs

<details>
<summary>9</summary>

```
Restarted application in 1 420ms.
W/e.flutter_issue( 7880): type=1400 audit(0.0:40384): avc: denied { read } for name="u:object_r:vendor_camera_prop:s0" dev="tmpfs" ino=14616 scontext=u:r:untrusted_app:s0:c182,c258,c512,c768 tcontext=u:object_r:vendor_camera_prop:s0 tclass=file permissive=0
E/libc    ( 7880): Access denied finding property "vendor.camera.aux.packagelist"
I/RequestQueue( 7880): Repeating capture request cancelled.
E/BufferQueueProducer( 7880): [SurfaceTexture-1-7880-9] cancelBuffer: BufferQueue has been abandoned
I/chatty  ( 7880): uid=10694(com.example.flutter_issue) Binder:7880_4 identical 1 line
E/BufferQueueProducer( 7880): [SurfaceTexture-1-7880-9] cancelBuffer: BufferQueue has been abandoned
E/BufferQueueProducer( 7880): [SurfaceTexture-1-7880-9] connect: BufferQueue has been abandoned
E/BufferQueueProducer( 7880): [SurfaceTexture-1-7880-9] setMaxDequeuedBufferCount: BufferQueue has been abandoned
W/Binder:7880_4( 7880): type=1400 audit(0.0:40405): avc: denied { read } for name="u:object_r:vendor_camera_prop:s0" dev="tmpfs" ino=14616 scontext=u:r:untrusted_app:s0:c182,c258,c512,c768 tcontext=u:object_r:vendor_camera_prop:s0 tclass=file permissive=0
E/libc    ( 7880): Access denied finding property "vendor.camera.aux.packagelist"
W/Binder:7880_4( 7880): type=1400 audit(0.0:40406): avc: denied { read } for name="u:object_r:vendor_camera_prop:s0" dev="tmpfs" ino=14616 scontext=u:r:untrusted_app:s0:c182,c258,c512,c768 tcontext=u:object_r:vendor_camera_prop:s0 tclass=file permissive=0
E/libc    ( 7880): Access denied finding property "vendor.camera.aux.packagelist"
I/Camera  ( 7880): [FPS Range Available] is:[15, 15]
I/Camera  ( 7880): [FPS Range Available] is:[20, 20]
I/Camera  ( 7880): [FPS Range Available] is:[24, 24]
I/Camera  ( 7880): [FPS Range Available] is:[7, 30]
I/Camera  ( 7880): [FPS Range Available] is:[30, 30]
I/Camera  ( 7880): [FPS Range] is:[7, 30]
W/e.flutter_issue( 7880): type=1400 audit(0.0:40407): avc: denied { read } for name="u:object_r:vendor_camera_prop:s0" dev="tmpfs" ino=14616 scontext=u:r:untrusted_app:s0:c182,c258,c512,c768 tcontext=u:object_r:vendor_camera_prop:s0 tclass=file permissive=0
E/libc    ( 7880): Access denied finding property "persist.vendor.camera.privapp.list"
I/CameraManager( 7880): Using legacy camera HAL.
W/Thread-17( 7880): type=1400 audit(0.0:40408): avc: denied { read } for name="u:object_r:vendor_default_prop:s0" dev="tmpfs" ino=14624 scontext=u:r:untrusted_app:s0:c182,c258,c512,c768 tcontext=u:object_r:vendor_default_prop:s0 tclass=file permissive=0
E/libc    ( 7880): Access denied finding property "vendor.camera.hal1.packagelist"
W/Binder:7880_4( 7880): type=1400 audit(0.0:40409): avc: denied { read } for name="u:object_r:vendor_camera_prop:s0" dev="tmpfs" ino=14616 scontext=u:r:untrusted_app:s0:c182,c258,c512,c768 tcontext=u:object_r:vendor_camera_prop:s0 tclass=file permissive=0
3
E/libc    ( 7880): Access denied finding property "vendor.camera.aux.packagelist"
I/CameraDeviceState( 7880): Legacy camera service transitioning to state CONFIGURING
I/RequestThread-0( 7880): Configure outputs: 2 surfaces configured.
D/Camera  ( 7880): app passed NULL surface
D/Camera  ( 7880): Lct package name is com.example.flutter_issue
E/libc    ( 7880): Access denied finding property "vendor.camera.aux.packagelist"
I/RequestThread-0( 7880): configureOutputs - set take picture size to 720x480
W/System  ( 7880): A resource failed to call release.
I/CameraDeviceState( 7880): Legacy camera service transitioning to state IDLE
I/RequestQueue( 7880): Repeating capture request set.
W/LegacyRequestMapper( 7880): convertRequestMetadata - control.awbRegions setting is not supported, ignoring value
I/CameraDeviceState( 7880): Legacy camera service transitioning to state CAPTURING
--------->|||||||||fisrt takePicture (ok)||||||||| I/RequestQueue( 7880): Repeating capture request cancelled.
I/RequestQueue( 7880): Repeating capture request set.
W/LegacyRequestMapper( 7880): convertRequestMetadata - control.awbRegions setting is not supported, ignoring value
D/LegacyFocusStateMapper( 7880): onAutoFocusMoving - ignoring move callbacks from old af run8
I/RequestQueue( 7880): Repeating capture request cancelled.
W/LegacyRequestMapper( 7880): convertRequestMetadata - control.awbRegions setting is not supported, ignoring value
2
W/LegacyRequestMapper( 7880): Only received metering rectangles with weight 0.
I/RequestThread-0( 7880): Received jpeg.
I/RequestThread-0( 7880): Producing jpeg buffer...
I/CameraDeviceState( 7880): Legacy camera service transitioning to state IDLE
I/RequestQueue( 7880): Repeating capture request set.
W/LegacyRequestMapper( 7880): convertRequestMetadata - control.awbRegions setting is not supported, ignoring value
I/CameraDeviceState( 7880): Legacy camera service transitioning to state CAPTURING
--------->|||||||||another takePicture (timed out)|||||||||        I/RequestQueue( 7880): Repeating capture request cancelled.
I/RequestQueue( 7880): Repeating capture request set.
W/LegacyRequestMapper( 7880): convertRequestMetadata - control.awbRegions setting is not supported, ignoring value
I/flutter ( 7880): Error: captureTimeout
I/flutter ( 7880): Error Message: Picture capture request timed out
```
</details>


<details>
<summary>13</summary>

```
I/CameraDeviceState(26004): Legacy camera service transitioning to state CAPTURING
I/RequestQueue(26004): Repeating capture request cancelled.
D/AndroidRuntime(26004): Shutting down VM
E/AndroidRuntime(26004): FATAL EXCEPTION: main
E/AndroidRuntime(26004): Process: com.example.flutter_issue, PID: 26004
E/AndroidRuntime(26004): java.lang.IllegalArgumentException: CaptureRequest contains unconfigured Input/Output Surface!
E/AndroidRuntime(26004): 	at android.hardware.camera2.CaptureRequest.convertSurfaceToStreamId(CaptureRequest.java:674)
E/AndroidRuntime(26004): 	at android.hardware.camera2.impl.CameraDeviceImpl.submitCaptureRequest(CameraDeviceImpl.java:1056)
E/AndroidRuntime(26004): 	at android.hardware.camera2.impl.CameraDeviceImpl.capture(CameraDeviceImpl.java:926)
E/AndroidRuntime(26004): 	at android.hardware.camera2.impl.CameraCaptureSessionImpl.capture(CameraCaptureSessionImpl.java:173)
E/AndroidRuntime(26004): 	at io.flutter.plugins.camera.Camera.runPictureCapture(Camera.java:675)
E/AndroidRuntime(26004): 	at io.flutter.plugins.camera.Camera.access$1500(Camera.java:75)
E/AndroidRuntime(26004): 	at io.flutter.plugins.camera.Camera$3.processCapture(Camera.java:541)
E/AndroidRuntime(26004): 	at io.flutter.plugins.camera.Camera$3.onCaptureCompleted(Camera.java:462)
E/AndroidRuntime(26004): 	at android.hardware.camera2.impl.CameraCaptureSessionImpl$1.lambda$onCaptureCompleted$3(CameraCaptureSessionImpl.java:641)
E/AndroidRuntime(26004): 	at android.hardware.camera2.impl.-$$Lambda$CameraCaptureSessionImpl$1$OA1Yz_YgzMO8qcV8esRjyt7ykp4.run(Unknown Source:8)
E/AndroidRuntime(26004): 	at android.os.Handler.handleCallback(Handler.java:873)
E/AndroidRuntime(26004): 	at android.os.Handler.dispatchMessage(Handler.java:99)
E/AndroidRuntime(26004): 	at android.os.Looper.loop(Looper.java:201)
E/AndroidRuntime(26004): 	at android.app.ActivityThread.main(ActivityThread.java:6810)
E/AndroidRuntime(26004): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime(26004): 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:547)
E/AndroidRuntime(26004): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:873)
I/CameraDeviceState(26004): Legacy camera service transitioning to state IDLE
3
V/AudioTrack(26004): getMinFrameCount=3844: afFrameCount=1920, afSampleRate=48000, afLatency=80
I/Process (26004): Sending signal. PID: 26004 SIG: 9
```
</details>



<details>
<summary>flutter doctor -v</summary>

```
[✓] Flutter (Channel texture_buildable, 1.27.0-2.0.pre.100, on Microsoft Windows [Version 10.0.19041.804], locale ru-RU)    • Flutter version 1.27.0-2.0.pre.100 at c:\dev\src\flutter
    • Framework revision 76080cd46c (9 days ago), 2021-02-21 18:22:32 +0300
    • Engine revision 1d537824d6
    • Dart version 2.13.0 (build 2.13.0-30.0.dev)

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
    • Android SDK at C:\Users\danya\AppData\Local\Android\sdk
    • Platform android-30, build-tools 30.0.2
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)
    • All Android licenses accepted.

[✓] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[!] Visual Studio - develop for Windows (Visual Studio Community 2019 16.7.7)
    • Visual Studio at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community
    • Visual Studio Community 2019 version 16.7.30621.155
    • Windows 10 SDK version 10.0.18362.0
    ✗ Visual Studio is missing necessary components. Please re-run the Visual Studio installer for the "Desktop
      development with C++" workload, and include these components:
        MSVC v142 - VS 2019 C++ x64/x86 build tools
         - If there are multiple build tool versions available, install the latest
        C++ CMake tools for Windows
        Windows 10 SDK

[✓] Android Studio (version 4.0)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin version 50.0.1
    • Dart plugin version 193.7547
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)

[✓] VS Code (version 1.53.2)
    • VS Code at C:\Users\danya\AppData\Local\Programs\Microsoft VS Code
    • Flutter extension version 3.20.0

[✓] Connected device (4 available)
    • Redmi Note 5 (mobile) • 40c9f14 • android-arm64  • Android 9 (API 28)
    • Windows (desktop)     • windows • windows-x64    • Microsoft Windows [Version 10.0.19041.804]
    • Chrome (web)          • chrome  • web-javascript • Google Chrome 88.0.4324.190
    • Edge (web)            • edge    • web-javascript • Microsoft Edge 88.0.705.74

! Doctor found issues in 1 category.
```
</details>

nt4f04uNd:
with 0.8.0 it's less buggy but still almost unusable for some of the use cases
if you need precise comparison i can make it tomorrow

mvanbeusekom:
@googlebot I consent.

nt4f04uNd:
i also noticed that to change these properties of `CameraController` you have to recreate the controller:

* `enableAudio`
* `imageFormatGroup`

within this PR, can we allow to change them without recreating the controller?

mvanbeusekom:
> i also noticed that to change these properties of `CameraController` you have to recreate it:
> 
> * `enableAudio`
> * `imageFormatGroup`
> 
> within this PR, can we allow to change them without recreating the controller?

I don't think we should do that here if it can be done at all. Updating these values will have several complications:

1. This would mean a change on both iOS and Android, while the current change only affects Android;
2. When you update the `enableAudio` parameter too `true` (default is `false`) we need to check for permissions on the microphone. Currently this only happens when the camera is initialised;
3. Updating the `enableAudio` parameter on Android would require to release and rebuild the `MediaRecorder` class, doing this while recording a movie will cause problems (in the least we need to nicely stop recording, update the value and start a new recording session which is not really user friendly). So reinitialising the camera here is a lot easier and we are sure all members are nicely released and can be reinitialised.

Petrol-:
I can confirm that this PR fixes all issues with Redmi 9 phones.
In version 0.7.x, this phone was unable to take any photo (with / without flash )

nt4f04uNd:
in the current state, it still has a whole bunch of problems on my redmi note 5

for example i just encountered

```
Unhandled Exception: CameraException(error, Attempt to invoke virtual method 'int android.hardware.camera2.CameraCaptureSession.capture(android.hardware.camera2.CaptureRequest, android.hardware.camera2.CameraCaptureSession$CaptureCallback, android.os.Handler)' on a null object reference)
E/flutter ( 9595): #0      CameraController.setFocusMode (package:camera/s
```

and then i tapped the volume button and the app just hang on forever

i can make detailed report again later

nt4f04uNd:
is this fully addressed? https://github.com/flutter/plugins/pull/3651#issuecomment-789407331

mvanbeusekom:
> is this fully addressed? [#3651 (comment)](https://github.com/flutter/plugins/pull/3651#issuecomment-789407331)

not yet. I think what would be smart now is to first finish writing the unit tests and from that point on start looking into your feedback (which is very clear and much appreciated ;)).

pedromassangocode:
@acoutts we may want to add the *fixes text before the issues you are sure will be fixed with this PR so that those issues are closed on this is merged!

acoutts:
@pedromassangocode I thought about that too but i'm not 100% sure it will solve their issue, as a lot of these problems were dependent on certain android manufacturers so it would be good if each person can verify the PR fixes it.

acoutts:
This is really coming along. @mvanbeusekom is doing amazing work with the testing and we've done some good refactoring to make it easier to add features in the future.

acoutts:
@mvanbeusekom just tested the latest commit on a Sony Xperia 1 and it works like a dream now. In a previous commit it still had some errors, but after the refactoring of features this past weekend, it's working great.

mvanbeusekom:
> @mvanbeusekom just tested the latest commit on a Sony Xperia 1 and it works like a dream now. In a previous commit it still had some errors, but after the refactoring of features this past weekend, it's working great.

That is awesome! Tomorrow I will finish testing all features and move on to the `CameraCaptureCallback` to test all state changes. Then the `Camera` class itself and we should have most of the things covered. This will be a big quality boost I think.

b.t.w. @account would you have some time to have a look at the timeout implementation we discussed earlier?

acoutts:
Sure thing I can get started on that tomorrow.

acoutts:
@mvanbeusekom in https://github.com/bottlepay/plugins/commit/2ddc5b81a834ad0c5a4bbe604a87e40dadbd415b I've fixed the timeout and cleaned up the picture capture request. I think it has value now, as it scopes the timeouts to the current capture.

Now when you initiate a capture,
* During focusing, it will move forward to metering either when the focus conditions are met OR if the focus timeout has lapsed.
* During metering, it will move forward to capturing if the metering conditions are met OR if the metering timeout has lapsed.

This way no matter if you get focus/metering, there will always be an image captured, as one would expect from a camera. The timeouts are probably too generous right now (3000ms each), but figured it would be a good place to start.

acoutts:
I found the Pixel 5 has something different about how AE metering works. The first picture taken is fine: AE is immediately converged during focusing, so it skips right to capturing the picture.

For every subsequent picture after that - during focusing, the AE state is `CONTROL_AE_STATE_SEARCHING` for a second, then it goes to `CONTROL_AE_STATE_PRECAPTURE`. It then hangs here indefinitely, even though the camera is now in `STATE_WAITING_PRECAPTURE_DONE`. So for this case, we just take the picture after the metering timeout.

Obviously in a perfect case we never need to use the timeout on metering (or focus), but I suppose this is a good use for it- in the unexpected case a device like the pixel 5 has something different about it, at least we can still capture a picture (even if the metering isn't perfect).

I've now tested the plugin on the following devices:
* Pixel 5
* Pixel 4
* LG k50s
* Samsung Galaxy S7
* Sony XZ (F8331)
* Sony Xperia 1
* Xiaomi A2

Annoyingly, the pixel 5 is behaving different from the pixel 4, so this really shows how hard it is to build an android camera that satisfies how all of these different devices are implemented- even some of the same manufacturer and family are completely different.

mvanbeusekom:
@stuartmorgan, @acoutts, I have finished writing unit tests and made sure the PR is up to date to pass all checks. I think this is ready for review.

nt4f04uNd:
@mvanbeusekom can i now give feedback on how it works?

mvanbeusekom:
> @mvanbeusekom can i now give feedback on how it works?

Yes, I think so. @acoutts made all the functional changes, I simply helped out making sure all changes are covered by tests. But feedback would be valuable.

stuartmorgan:
I'll take a look, but @bparrishMines it would be great if you could review as well since you're familiar with `camera`.

nt4f04uNd:
@acoutts this is an updated feedback from https://github.com/flutter/plugins/pull/3651#issuecomment-789407331 - those issues that were not fixed are included here (some of them were reformulated and better demonstrated)

this is tested with an example app from https://github.com/bottlepay/plugins/pull/1

[logs](https://drive.google.com/drive/folders/1dmX-HTyydnudcQdwObhHxWKGxqnBH7Ap?usp=sharing)
[videos](https://drive.google.com/drive/folders/1UzEv3PMbamuu0den8oGBR7_xtZRNT3QC?usp=sharing)

1. back camera record video broken
2. back camera setting focus mode to auto turns the torch on (only on the first time)
3. fast taps crash - this is in realese mode. in debug app will just hang on forever
4. camera huge freeze with flash turned on
5. front camera flash `always` does nothing
6. front camera flash `always` breaks flash mode after shot
7. front camera record video breaks back camera
8. rotation not working if orientation is locked - on the video i compared this to the native camera, achieve something similar is not possible (i am not sure if this should be fixed in terms of this PR)
9. back camera torch turns off after shot
10. focus mode locked not reacting to taps
11. exposure mode locked not reacting to taps

stuartmorgan:
Before diving into the 6000 line review here, it would be good to evaluate whether individual aspects of this can be separated out into individual PRs. Smaller PRs have a lot of benefits:
- They are far easier to review, since the reviewer doesn't need to hold the entire enormous change in their head at the same time.
- It's easier to evaluate whether the change is likely to have side effects beyond what was intended by the change, since the change should be relatively targeted.
- They allow pinpointing the source of any regressions, instead of having to figure out which piece of an enormous change caused them.
- Relatedly, they allow reverting specific aspects that are found to be problematic later, rather than having to potentially revert the whole change (in the case of a critical issue that we can't immediately pinpoint the cause of among the 6k lines).

Ideally we have small PRs by having PRs written that way in the first place, but I've had many cases over the years where I've done very large changes locally then carved them up after the fact into separable pieces (and when I've done that, it has almost always been worth the extra effort). This PR would serve as a demonstration of what all the individual PRs are headed in order to provide context, but each PR could be reviewed and landed separately, and this PR would shrink over time until the remaining changes were more manageable.

To be clear, the goal here would not be to make every tiny change a separate PR, as that would be nearly impossible to do after the fact. But as an example, you have a bunch of new classes that gather/track information about the camera (`CameraFeature` and its various subclasses), which could presumably be added in one or more separate PRs which are adopted within the context of the existing code as incremental improvements.

@acoutts Please let me know if you have thoughts about parts that would be relatively easy to separate out, since you are best positioned to know; if not I can take a closer look and try to suggest some cut points.

acoutts:
Great points @stuartmorgan. This is a tough one. I would like to see it as smaller PRs, but given we have basically binned a majority of the existing plugin and rebuilt it from scratch, it would be hard to break it up. Given there were no tests before, and a lot of user reports of broken functionality, it's even harder to compare if we have a reversion since this is really the first proper implementation of the plugin.

Just thinking from the top of my head about what exactly we changed:
* There was one small change to the camera example app UI which allows more camera options to fit. This could be broken out, but it's a small change and doesn't help break down the java code part.
* The camera features are a refactor from previously having a mix-match of private class variables for the various settings. I don't think we can break this one out without having to write all new tests to make it fit halfway into the old plugin and the new one.
* We removed the state machine from PictureCaptureRequest and kept just the main state machine in Camera.java for the actual state of the camera. This maybe can be split out.

Come to think of it, the camera features was a big bulk of the refactoring to clean things up (and make it testable). Really tough call on this one.

I don't want to sound defensive out of laziness or because I don't want to do it, but i'm really just thinking if it's feasible to do.

stuartmorgan:
> Given there were no tests before, and a lot of user reports of broken functionality, it's even harder to compare if we have a regression since this is really the first proper implementation of the plugin.

Well, we'd know after the fact because we'd get issue reports saying that a specific device/case worked with the old version, and fails with the new version. Then we'd need to figure out what happened.

> * The camera features are a refactor from previously having a mix-match of private class variables for the various settings. I don't think we can break this one out without having to write all new tests to make it fit halfway into the old plugin and the new one.
> [...]
> Come to think of it, the camera features was a big bulk of the refactoring to clean things up (and make it testable). Really tough call on this one.

Yes, that's why I was focusing there; it seems like they are leaf classes (thus more likely to be separable), and also a significant portion of the PR. If they can be carved off, it would be easier to focus on the core logic rewrite here.

What if the camera features PR(s) didn't add tests beyond the unit tests of the feature class themselves (which can presumably move as-is to a new PR)? Would it be straightforward to swap them in for the existing variables other than the testing aspect? While obviously the ideal would be for the incremental PRs to have substantial incremental tests, if we can get most of the benefit of incremental PRs without much incremental effort by deferring all the new behavioral testing to this PR, that seems like a good tradeoff to me. (Unlike most cases of deferred testing this wouldn't be "we'll come back and write tests later", which rarely happens, but "we already have the tests, and they'll be landed soon".)

carman247:
If I "watch" this PR will I be notified when it is reviewed/merged? 😄

acoutts:
I've thought on this some more and I just don't have the bandwidth to break this down into separate PRs right now. If it's a hard requirement for this to be merged then I will try to get to it in a couple of weeks but I cant promise anything until then. In the meantime I invite others to please test and provide feedback on your various devices.

I think no matter what the PR is going to be quite large to review here given the scope of the rewrite.

mvanbeusekom:
@stuartmorgan, @acoutts, would it be an idea to one by one add the `...Feature` classes and their tests to the current implementation without actually using them yet? 

This will allow us to add each feature class including their unit tests without having to worry about the current implementation and will keep the PRs nice and small. When all the features classes and utility classes have been moved over we can swap out the actual implementation which will then be a lot smaller (still is a big chunk of code but would be more manageable). That would take out 40 files of the current 66 files.

We could even merge this to a separate "staging" branch so it doesn't get in the way as "temporary" dead code on the master branch.

stuartmorgan:
If incremental adoption of the feature classes isn't feasible, then yes, just adding the classes separately would still help keep the reviews more manageable.

vimaxwell:
@acoutts. this PR is really good. I fixed most the issues our users encountered. 
 But I still got 2 crashes in firebase crashlytics. I hope you can figure out.

>
Device
Brand: HUAWEI
Model: nova 2i
Orientation: Portrait
Version: 8.0.0

```
Fatal Exception: java.lang.RuntimeException: Methods marked with @UiThread must be executed on the main thread. Current thread: CameraBackground
       at io.flutter.embedding.engine.FlutterJNI.ensureRunningOnMainThread(FlutterJNI.java:36)
       at io.flutter.embedding.engine.FlutterJNI.invokePlatformMessageResponseCallback(FlutterJNI.java)
       at io.flutter.embedding.engine.dart.DartMessenger$Reply.reply(DartMessenger.java:27)
       at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler$1.a(MethodChannel.java:14)
       at io.flutter.plugins.camera.PictureCaptureRequest.a(PictureCaptureRequest.java:2)
       at io.flutter.plugins.camera.Camera.takePictureAfterPrecapture(Camera.java:114)
       at io.flutter.plugins.camera.Camera.access$000(Camera.java)
       at io.flutter.plugins.camera.CameraCaptureCallback.a(CameraCaptureCallback.java:18)
       at io.flutter.plugins.camera.CameraCaptureCallback.a(CameraCaptureCallback.java:196)
       at io.flutter.plugins.camera.CameraCaptureCallback.onCaptureCompleted(CameraCaptureCallback.java)
       at java.lang.reflect.Method.invoke(Method.java)
       at android.hardware.camera2.dispatch.InvokeDispatcher.dispatch(InvokeDispatcher.java:39)
       at android.hardware.camera2.dispatch.HandlerDispatcher$1.run(HandlerDispatcher.java:65)
       at android.os.Handler.handleCallback(Handler.java:808)
       at android.os.Handler.dispatchMessage(Handler.java:101)
       at android.os.Looper.loop(Looper.java:166)
       at android.os.HandlerThread.run(HandlerThread.java:65)
```

----
Device
Brand: HMD Global
Model: Nokia 1
Orientation: Portrait
Operating System
Version: 8.1.0
```
Fatal Exception: java.lang.IllegalArgumentException: Surface was abandoned
       at android.hardware.camera2.utils.SurfaceUtils.getSurfaceSize(SurfaceUtils.java:70)
       at android.hardware.camera2.params.OutputConfiguration.<init>(OutputConfiguration.java:216)
       at android.hardware.camera2.params.OutputConfiguration.<init>(OutputConfiguration.java:101)
       at android.hardware.camera2.impl.CameraDeviceImpl.createCaptureSession(CameraDeviceImpl.java:499)
       at io.flutter.plugins.camera.Camera.a(Camera.java:4)
       at io.flutter.plugins.camera.Camera.a(Camera.java:163)
       at io.flutter.plugins.camera.Camera.a(Camera.java:1)
       at io.flutter.plugins.camera.Camera.startPreview(Camera.java:30)
       at io.flutter.plugins.camera.Camera$3.onOpened(Camera.java:14)
       at android.hardware.camera2.impl.CameraDeviceImpl$1.run(CameraDeviceImpl.java:139)
       at android.os.Handler.handleCallback(Handler.java:790)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:164)
       at android.app.ActivityThread.main(ActivityThread.java:6523)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:857)
```

arrrrny:
WOW!! Thank you so much!!! All stupid camera errors are gone!! It was impossible to take picture with my Xiaomi 10 Lite. everything works so smooth with your patch.  Here is my settings for fast capturing with your patch.

`      await cameraController.initialize();
      await cameraController.lockCaptureOrientation();
      await cameraController.setFocusPoint(
        Offset(0.5, 0.5),
      );
      await cameraController.setFocusMode(FocusMode.locked);`

acoutts:
@googlebot I consent.

mvanbeusekom:
@googlebot I consent.

Bringoff:
Kudos for this monumental work. I'm only wondering why don't you use Jetpack Camera library and go for raw Camera2 API.

Bringoff:
By the way, using Camera2 this way might be the reason that I still cannot use camera on some older devices. I've switched to the latest commit from this branch

```
  # TODO: Switch back to upstream after merged
  camera:
    git:
      url: https://github.com/bottlepay/plugins.git
      path: packages/camera/camera
      ref: 38b0849eb1c4f5b59a6e9b4fc87b78304610454f
```

And when I try to open a screen where we use camera on a device like Moto E 2 Gen (Android 5.1) I get following errors in logs:
```
I/CameraManager( 6903): Using legacy camera HAL.
W/Camera  ( 6903): An error occurred while connecting to camera: 0
I/Camera  ( 6903): open | onError
I/Camera  ( 6903): close
```
And of course there is no camera preview and it's impossible to take a picture. Our QA reports that he observes the same behavior on Samsung SM-T285 Android 5.1.1.
This is frustrating as we rely on camera heavily and work on this PR is too slow. I feel like we'll have to fork camera plugin and do all the fixes ourselves.

Bringoff:
Also, probably because of this error the Future inside camera_controller `initialize()` method never finishes.
```
await CameraPlatform.instance.initializeCamera(
        _cameraId,
        imageFormatGroup: imageFormatGroup ?? ImageFormatGroup.unknown,
      );
```

Bringoff:
I've connected this branch to our project and played with it a little. Looks promising, can't wait to see this in main repo. Unfortunately, for now I'll have to make a fork of this fork :) There are several problems described in [this comment](https://github.com/flutter/plugins/pull/3651#issuecomment-805069753). We don't use front camera and don't record video, so most of them are not critical for us. Still, we observe major freezing when flash is enabled. Also, on older devices (Android 5.1 for our case) where legacy camera HAL used, our app freezes forever when we call `await cameraController.initialize()` and camera device is busy. I guess, we should get an exception, but this doesn't happen and I see `WaitingInMainSignalCatcherLoop` warning in logs.

So, I guess, because of all the separating-in-multiple-PRs work for now I'll have to fork this branch and do some workarounds myself if you don't mind.

IanDarwin:
This solved my problems on the Teracube 2e phone (Android 10).
No regressions on various Android (10,11) devices.
No regressions on iPhone (iOS 14.4.2)
Unfortunately it does break on Android 5, as I think was noted earlier. The preview is solid black and the 
"Take Picture" button does nothing. Android 5 is NOT a show stopper for me; might be for others.
[Uploading Z220-camera-android5.log…](Console log attached in case it helps.)

Bringoff:
@IanDarwin had the same problem. Turned out that was because we used another library to check if flash is available, that was likely implemented via Camera 1 API internally. Everything worked well on newer devices, but Android 5 devices often doesn't support Camera 2 API used in this library and switch to legacy mode, so we got an error about camera being locked by lower API. Works fine after removing that library usages.

IanDarwin:
> ... had the same problem. Turned out that was because we used another library...

Thanks @Bringoff but not using any other camera-related APIs.

psrcek:
If someone else is looking to use this right now before it's merged, use this in your pubspec:

```
  camera:
    git:
      url: https://github.com/bottlepay/plugins.git
      ref: android-rework
      path: packages/camera/camera
```

Just be warned, this is really BLEEDING EDGE and uses a branch that is being worked on and may break at any time as far as I'm aware! (unless @acoutts or someoe that worked on this confirms it is fine to use)

carman247:
> If someone else is looking to use this right now before it's merged, use this in your pubspec:
> 
> ```
>   camera:
>     git:
>       url: https://github.com/bottlepay/plugins.git
>       ref: android-rework
>       path: packages/camera/camera
> ```
> 
> Just be warned, this is really BLEEDING EDGE and uses a branch that is being worked on and may break at any time as far as I'm aware! (unless @acoutts or someone that worked on this confirms it is fine to use)

I've been using this for a while but intermittently getting crashes when taking photos on Android .. Haven't noticed if the same on iOS.

I'll try and get a proper crash report added but not sure if it's relevant to this issue.

wisnuwiry:
Conflict @acoutts

stuartmorgan:
> Conflict @acoutts

The conflict is with the first of the incremental patches that are being extracted; there's not a lot of point in resolving them until all five have landed.

carman247:
Hi,

is this still the most current fix to use?

```
  camera:
    git:
      url: https://github.com/bottlepay/plugins.git
      path: packages/camera/camera
      ref: android-rework
```

carman247:
Any news on this?

I'm desperate to release an update for an app but cannot use the current compatible versions or this PR because I'm still getting crashes that just say `wrote stack traces to tombstoned` and I'm unable to get any more info.

stuartmorgan:
> Any news on this?

Individual pieces continue to be extracted, reviewed, and landed.

> I'm desperate to release an update for an app but cannot use the current compatible versions or this PR

If you have a crash even when using this PR then this isn't going to be relevant to your use case; you should file an issue with details.

BeMacized:
Closing this, as the final changes from this PR have finally been merged with #4059. Many thanks to those who worked on this!

PR #3661, opened 2021-03-02
Remove unused plugin (integration_test_macos)

cyanglaz:
This plugin is not used nor published. Removing


stuartmorgan (approved):
LGTM.

(I remember now: when I looked at this last time I found that it was implemented, but never endorsed, and nothing was pulling in the unendorsed version.)

PR #3662, opened 2021-03-02
[google_maps_flutter_web] update min flutter sdk version to 1.20.0

cyanglaz:
Since the plugin doesn't have an iOS folder, it is required that the min sdk version is 1.20.0
Also, it looks like the last 0.1.2 version was a mistake (made in https://github.com/cyanglaz/plugins/commit/6fa2ead23e8d58de047f38c0f90ce5e218ef8ea5) and has been published, so i only updated the changelog.


ditman (approved):
LGTM, thanks for fixing! I'll publish 0.1.2 as soon as this lands :)

PR #3653, opened 2021-03-02
Adopt Xcode 12 for podspec lints

jmagman:
Adopt Xcode 12 for all Cirrus tasks.
Add `--configuration=Debug` option ([available as of CocoaPods 1.10](https://github.com/CocoaPods/CocoaPods/pull/9760)) to `pod lib lint` CI script so the build only targets the supported x86_64 simulator, not the unsupported (by Flutter) arm64 simulator.

See also https://github.com/flutter/plugins/pull/3304.

PR #3145, opened 2020-10-12
[in_app_purchase] Add example test target to Podfile, add OCMock dependency

jmagman:
## Description

Test target wasn't building in example app because it wasn't inheriting search paths from the Runner target (which allows the linker to find the plugin frameworks, plus Flutter framework).  Also add OCMock dependency to match test_spec podspec.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/67175

## Checklist
- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change
- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

PR #3663, opened 2021-03-02
[image_picker] Run CocoaPods iOS tests in RunnerUITests target

jmagman:
`pod lib lint` has been used to run iOS unit tests and the static analyzer.  Unfortunately due to [Flutter not supporting ARM iOS simulators](https://github.com/flutter/flutter/issues/64502) and the fact that we [can't tell the `pod` command to skip analyzing on that architecture](https://github.com/flutter/plugins/pull/3653), this has prevented the plugins CI from migrating to Xcode 12 and is blocking Flutter from [increasing the Xcode minimum version to 12](https://github.com/flutter/flutter/pull/77025).

Now that we have a full-fledged [xctest_command](https://github.com/flutter/plugins/blob/master/script/tool/lib/src/xctest_command.dart) we can migrate some of the value we were getting from `pod lib lint`.

Migrate `image_picker` package to run all the unit tests as part of the `RunnerUITests` target.  Remove the old `image_picker_exampleTests` in favor of `RunnerUITests`.

Also swap `VALID_ARCHS` to the Xcode 12 `EXCLUDED_ARCHS` generated by the Flutter template as of flutter/flutter#64504

### DIFF
--- a/packages/image_picker/image_picker/example/ios/Podfile
+++ b/packages/image_picker/image_picker/example/ios/Podfile
@@ -1,0 +1,42 @@
+# Uncomment this line to define a global platform for your project
+# platform :ios, '9.0'
+
+# CocoaPods analytics sends network stats synchronously affecting flutter build latency.
+ENV['COCOAPODS_DISABLE_STATS'] = 'true'
+
+project 'Runner', {
+  'Debug' => :debug,
+  'Profile' => :release,
+  'Release' => :release,
+}
+
+def flutter_root
+  generated_xcode_build_settings_path = File.expand_path(File.join('..', 'Flutter', 'Generated.xcconfig'), __FILE__)
+  unless File.exist?(generated_xcode_build_settings_path)
+    raise "#{generated_xcode_build_settings_path} must exist. If you're running pod install manually, make sure flutter pub get is executed first"
+  end
+
+  File.foreach(generated_xcode_build_settings_path) do |line|
+    matches = line.match(/FLUTTER_ROOT\=(.*)/)
+    return matches[1].strip if matches
+  end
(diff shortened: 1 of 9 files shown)


packages/image_picker/image_picker/example/ios/Podfile
```diff
@@ -0,0 +1,42 @@
...
+require File.expand_path(File.join('packages', 'flutter_tools', 'bin', 'podhelper'), flutter_root)
+
+flutter_ios_podfile_setup
+
+target 'Runner' do
+  flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
+
+  target 'RunnerUITests' do
+    inherit! :search_paths
+  end
```
jmagman, line 35: Checked in the `Podfile` to inherit `search_paths` so Flutter linker flags are hooked up. See https://github.com/flutter/plugins/pull/3145 for similar fix.

cyanglaz (commented):
I remember these are unit tests, should we create a separate target for unit tests?
I didn't know you can combine unit tests and ui tests in a same target.

jmagman:
> I remember these are unit tests, should we create a separate target for unit tests?
> I didn't know you can combine unit tests and ui tests in a same target.

You can run whatever you want in the UI tests as long as the symbols are available.  The `Runner` app symbols aren't available, but the UI tests are linking directly on the plugin framework, and it can test those.

It worked:
https://cirrus-ci.com/task/5631298589949952?command=build#L1930

cyanglaz (approved):
LGTM, maybe we should rename the target to RunnerTest in the future which would be more accurate.

jmagman:
> LGTM, maybe we should rename the target to RunnerTest in the future which would be more accurate.

Sure, we can use whatever test target name we want once https://github.com/flutter/plugins/pull/3667 merges.

PR #3664, opened 2021-03-02
[webview_flutter] Run CocoaPods iOS tests in RunnerUITests target

jmagman:
Migrate `webview_flutter` package to run all the unit tests as part of the `RunnerUITests` target.  Remove the old `webview_flutter_exampleTests ` in favor of `RunnerUITests`.

Also swap `VALID_ARCHS` to the Xcode 12 `EXCLUDED_ARCHS` generated by the Flutter template as of flutter/flutter#64504

See https://github.com/flutter/plugins/pull/3667 for more details.

PR #3667, opened 2021-03-03
Run static analyzer during xctest

jmagman:
`pod lib lint` has been used to run iOS unit tests and the static analyzer.  Unfortunately due to [Flutter not supporting ARM iOS simulators](https://github.com/flutter/flutter/issues/64502) and the fact that we [can't tell the `pod` command to skip analyzing on that architecture](https://github.com/CocoaPods/CocoaPods/issues/10476), this has prevented the plugins CI from migrating to Xcode 12 and is blocking Flutter from [increasing the Xcode minimum version to 12](https://github.com/flutter/flutter/pull/77025).

Now that we have a full-fledged [xctest_command](https://github.com/flutter/plugins/blob/master/script/tool/lib/src/xctest_command.dart) we can migrate the static analyzer from `pod lib lint` to the custom test runner.

1. Run static analyzer as part of `xctest`. Automatically skip `test` and just `analyze` if the unit tests aren't set up.  Remove `--target` from xctest, always just use the test target associated with the `Runner` app.
2. Stop analyzing in the `lint_podspecs_command`.

Unblocks https://github.com/flutter/plugins/pull/3653 and therefore https://github.com/flutter/flutter/pull/77025.

- [x] Currently skipping `webview_flutter` until https://github.com/flutter/plugins/pull/3664 merges, which will hook up the existing pod unit tests to the `Runner` target.
- [x] Same for `image_picker` until https://github.com/flutter/plugins/pull/3663 merges.

Fixes https://github.com/flutter/flutter/issues/68419

### DIFF
--- a/script/tool/lib/src/xctest_command.dart
+++ b/script/tool/lib/src/xctest_command.dart
@@ -12,17 +12,15 @@ import 'package:path/path.dart' as p;
 import 'common.dart';
 
 const String _kiOSDestination = 'ios-destination';
-const String _kTarget = 'target';
 const String _kSkip = 'skip';
 const String _kXcodeBuildCommand = 'xcodebuild';
 const String _kXCRunCommand = 'xcrun';
 const String _kFoundNoSimulatorsMessage =
     'Cannot find any available simulators, tests failed';
 
-/// The command to run iOS' XCTests in plugins, this should work for both XCUnitTest and XCUITest targets.
-/// The tests target have to be added to the xcode project of the example app. Usually at "example/ios/Runner.xcodeproj".
-/// The command takes a "-target" argument which has to match the target of the test target.
-/// For information on how to add test target in an xcode project, see https://developer.apple.com/library/archive/documentation/ToolsLanguages/Conceptual/Xcode_Overview/UnitTesting.html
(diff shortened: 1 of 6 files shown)


script/tool/lib/src/xctest_command.dart
```diff
@@ -164,6 +107,31 @@ class XCTestCommand extends PluginCommand {
...
+      _kXcodeBuildCommand,
+      if (runTests)
+        'test',
+      'analyze',
+      '-workspace',
+      'ios/Runner.xcworkspace',
+      '-configuration',
+      'Debug',
+      '-scheme',
+      'Runner',
```
cyanglaz, line 121: I didn't know this, making the scheme to be the app target runs all the test targets?

cyanglaz (approved):
LGTM assuming all test targets run

script/tool/lib/src/xctest_command.dart
jmagman: No, it runs whatever tests are hooked up in the specified scheme. So in this case, `xcodebuild test -scheme Runner` will run the tests set up in the `Runner` scheme.
Screen Shot 2021-03-02 at 5 52 00 PM

PR #3669, opened 2021-03-03
[Video_Player] Remove the deprecated API reference.

zhenqiu1101:
exoPlayer.setAudioStreamType has been deprecated. Remove it from the video_player plugin.

Doc: https://developer.android.com/reference/android/media/MediaPlayer#setAudioStreamType(int)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player/CHANGELOG.md
+++ b/packages/video_player/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.1
+
+* Remove the deprecated API "exoPlayer.setAudioAttributes".
+
 ## 2.0.0
 
 * Migrate to null safety.
--- a/packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
+++ b/packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
@@ -222,12 +222,8 @@ final class VideoPlayer {
 
   @SuppressWarnings("deprecation")
   private static void setAudioAttributes(SimpleExoPlayer exoPlayer, boolean isMixMode) {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-      exoPlayer.setAudioAttributes(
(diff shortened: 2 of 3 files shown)


stuartmorgan:
It's deprecated in API 21, but the plugin supports building back to 16, so this will regress running on old versions. Can you explain the problem you are trying to solve? Are you using alternate build files with a newer target?

zhenqiu1101:
This API has been removed entirely from the internal exoplayer library. So the internal plugin is diverged from the upstream one. But I'm not sure if it's safe to remove this reference in the upstream as you mentioned. I've talked with Chris. Chris suggest me to create a PR to get some feedback. We can handle this difference for now if this causes regression.

cyanglaz:
Is it possible to catch `NoSuchMethodException`?

stuartmorgan:
Oh I see, I missed that this is ExoPlayer rather than Media player. So I think the question becomes, does the newer ExoPlayer API work down-level?

zhenqiu1101:
I think so. Actually, the exoPlayer.setAudioStreamType API has been removed from the source. But I'm not familiar with the upstream dependencies. I may look at the wrong repo. https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java

zhenqiu1101:
Kindly ping.

stuartmorgan:
> I think so.

Can you link to documentation/source/discussion that supports this?

> Actually, the exoPlayer.setAudioStreamType API has been removed from the source.

Is that because the new method works downlevel so isn't needed, or has the current version dropped support for older versions of Android?

We need definitive confirmation that this isn't going to break the plugin for Android versions we still support before proceeding.

zhenqiu1101:
See here: https://github.com/google/ExoPlayer/blob/r2.9.6/library/core/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java#L499

setAudioStreamType internally use setAudioAttributes(AudioAttributes) which is also deprecated later and should use setAudioAttributes(AudioAttributes, boolean) instead.

stuartmorgan (approved):
Thanks, makes sense then! LGTM

PR #897, opened 2018-11-13
[video_player]Added HTTP headers setting to video_player plugin

Yarikk26:
Not the best solution for https://github.com/flutter/flutter/issues/16466 especially for iOS.
It seems that there is the only one way to set headers for iOS player, but it isn't public (https://stackoverflow.com/a/23713028).

PR #3671, opened 2021-03-03
[video_player] add http headers

PiN73:
This enables to pass HTTP headers to `VideoPlayerController.network`.

Actually this is rebasing @Yarikk26's implementation https://github.com/flutter/plugins/pull/897 on top of video_player 2.0.0.

Fixes: https://github.com/flutter/flutter/issues/16466

## Pre-launch Checklist

- [x] I signed the [CLA].


PiN73:
As described in [CONTRIBUTING.md](https://github.com/flutter/plugins/blob/cbee8561c07495d542ccca6dc8572bd0b4c92cd9/packages/video_player/video_player/CONTRIBUTING.md), video_player in the PR is temporary using overriden video_player_platform_interface_dependency because it isn't published yet

stuartmorgan:
Thanks for picking up this PR! I'll try to take a look in the next week or so.

BartusZak:
Looking forward to see that.

packages/video_player/video_player/example/pubspec.yaml
```diff
@@ -14,6 +14,15 @@ dependencies:
...
     path: ../
 
+# For development / testing purpose only.
+# When using video_player in a real application
+# you won't directly depend on video_player_platform_interface at all.
+# The example app is bundled with the plugin so we use a path dependency on
+# the parent directory to use the current plugin's version.
+dependency_overrides:
+  video_player_platform_interface:
+    path: ../../video_player_platform_interface
```
stuartmorgan, line 24: This will need to be removed, and the PR split into two parts, before landing.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
```diff
@@ -162,8 +164,15 @@ - (void)createVideoOutputAndDisplayLink:(FLTFrameUpdater*)frameUpdater {
   _displayLink.paused = YES;
 }
 
-- (instancetype)initWithURL:(NSURL*)url frameUpdater:(FLTFrameUpdater*)frameUpdater {
-  AVPlayerItem* item = [AVPlayerItem playerItemWithURL:url];
+- (instancetype)initWithURL:(NSURL*)url
+               frameUpdater:(FLTFrameUpdater*)frameUpdater
+                httpHeaders:(NSDictionary<NSString*, NSString*>*)headers {
+  NSDictionary<NSString*, id>* options = nil;
+  if (headers != NULL) {
```
stuartmorgan, line 171: `nil`, not `NULL`

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -194,8 +195,13 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
   /// **Android only**: The [formatHint] option allows the caller to override
   /// the video format detection code.
+  /// [httpHeaders] option allows to specify HTTP headers
+  /// for the request to the [dataSource].
   VideoPlayerController.network(this.dataSource,
-      {this.formatHint, this.closedCaptionFile, this.videoPlayerOptions})
+      {this.formatHint,
+      this.closedCaptionFile,
+      this.videoPlayerOptions,
+      this.httpHeaders})
```
stuartmorgan, line 204: This will format better if you add a comma after `httpHeaders`

packages/video_player/video_player_platform_interface/lib/messages.dart
```diff
@@ -43,7 +45,8 @@ class CreateMessage {
       ..asset = pigeonMap['asset'] as String?
       ..uri = pigeonMap['uri'] as String?
       ..packageName = pigeonMap['packageName'] as String?
-      ..formatHint = pigeonMap['formatHint'] as String?;
+      ..formatHint = pigeonMap['formatHint'] as String?
+      ..httpHeaders = pigeonMap['httpHeaders'] as Map<Object?, Object?>?;
```
stuartmorgan, line 49: I haven't used pigeon myself yet; does this need to be Object rather than String?

stuartmorgan (changes requested):
Overall this looks really good. Two things it'll need:
- The platform interface change moved to its own PR
- Dart tests in the video_player package that ensure that headers passed into the API reach the platform interface (via a mock/fake platform interface implementation)

packages/video_player/video_player_platform_interface/lib/messages.dart
PiN73: As far as I understand, pigeon doesn't support generics at the moment. httpHeaders field is actually [declared](https://github.com/flutter/plugins/blob/0761edf6402e8b3e1321effb003cf21b92262490/packages/video_player/video_player/pigeons/messages.dart#L34) as `Map<String, String>` in pigeons/messages.dart. But lib/messages.dart is generated so if we follow "do not manually edit generated files" rule, we should keep it as is

packages/video_player/video_player/example/pubspec.yaml
PiN73: Doesn't it make sense to keep the override to make anyone cloning the source to work with local copy?

Just like `video_player:` dependency in the example is declared as `path: ../` to point to local copy of `video_player` package, this would do the same for `video_player_platform_interface` package

stuartmorgan: > Doesn't it make sense to keep the override to make anyone cloning the source to work with local copy?

There are several issues with that, most notably that it causes the integration tests to run against a configuration that is different from what actual clients of this package would run, making it easy to have a PR pass CI and be landed and published, but break everyone because it doesn't work with the published interface package.

> Just like `video_player:` dependency in the example is declared as `path: ../` to point to local copy of `video_player` package

The example is *part* of that package. It's an internal reference.

> this would do the same for `video_player_platform_interface` package

Overriding dependencies in a multi-package graph is definitely not the same thing.

PiN73: OK, thanks, I understand the problem now. May I keep it commented out or should completely delete?

stuartmorgan: We don't check in commented out code, generally speaking. It's very easy for someone who wants this behavior for temporary testing to re-add it.

PiN73: OK, I'll remove the override. But the https://github.com/flutter/plugins/pull/3702 need to be landed first (if there aren't any other issues in the current PR) because the example won't built otherwise

PiN73:
@stuartmorgan I fixed all problems you mentioned except removing dependency_overrides in example app. If I remove it now, the example won't compile. Should https://github.com/flutter/plugins/pull/3702 be merged and I remove the override then, right?

PiN73:
Any additional edits needed from me?

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -212,12 +219,16 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
         formatHint = null,
+        httpHeaders = null,
         super(VideoPlayerValue(duration: Duration.zero));
 
   /// The URI to the video file. This will be in different formats depending on
   /// the [DataSourceType] of the original video.
   final String dataSource;
 
+  /// HTTP headers used for the request to the [dataSource].
+  final Map<String, String>? httpHeaders;
```
stuartmorgan, line 230: We generally prefer to make collections non-nullable and defaulting to empty, so that we don't have both null and empty as codepaths that mean the same thing (except in cases where null and empty actually mean different things, but that's not true here). It avoids bugs where, e.g., one platform implementation handles them the same, but one fails with one or the other (or treats them differently on accident).

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -30,6 +30,9 @@ class FakeController extends ValueNotifier<VideoPlayerValue>
   @override
   String get dataSource => '';
 
+  @override
+  Map<String, String>? get httpHeaders => null;
```
stuartmorgan, line 34: Same.

stuartmorgan (commented):
Sorry for the review delay; doing both now.

packages/video_player/video_player/lib/video_player.dart
PiN73: As pigeon doesn't currently support non-nullable fields, generated fields will be still nullable. If we make headers non-nullable in VideoPlayerController class, should we check for nulls in platform code then?

PiN73: Well I made it non-nullable and empty by default. Also I kept check for null and added check for empty in platform code. So the default behavior is untouched (don't call function setting headers in native code if no headers were passed). And also, if null will git into native side somehow, it will be also ignored

packages/video_player/video_player/example/pubspec.yaml
PiN73: Deleted it now

PiN73:
Now, as video_player_platform_interface 4.1.0 is published, I removed dependency overrides. Any additional changes needed, @stuartmorgan?

stuartmorgan (approved):
Looks good, thanks!

BartusZak:
Thank you very much.

I'm going to test it right away :)

PR #3684, opened 2021-03-05
[shared_preferences] Fix concurrent modification of the shared preferences on Android

konstantintuev:
When value is set before the last value has been committed to the shared preferences, the commit of the last value is rejected with the following error: 
`java.util.concurrent.RejectedExecutionException: Task io.flutter.plugins.sharedpreferences.MethodCallHandlerImpl rejected from java.util.concurrent.ThreadPoolExecutor`.

This happens because a `SynchronousQueue` is being used, which can handle only one request at a time and isn't really queue unless a request is removed or done.
By replacing it with a `LinkedBlockingQueue` the commits are queue to execute one after another. This preserves the async nature of the library, prevents data loss and fixes the `RejectedExecutionException`.

This fixes this issue: [[shared_preferences] 2.0.3 throws a lot of Java Tracebacks on Android](https://github.com/flutter/flutter/issues/77155)

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/packages/shared_preferences/shared_preferences/CHANGELOG.md
+++ b/packages/shared_preferences/shared_preferences/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.4
+
+* Fix a regression with simultaneous writes on Android.
+
 ## 2.0.3
 
 * Android: don't create additional Handler when method channel is called.
--- a/packages/shared_preferences/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/MethodCallHandlerImpl.java
+++ b/packages/shared_preferences/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/MethodCallHandlerImpl.java
@@ -23,7 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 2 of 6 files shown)


monsieurtanuki:
My 2 cents: probably this has been discussed before, but using `apply()` instead of `commit()` seems to be the best practice on Android (and you wouldn't have to bother about the threads). Unless you need the return code of the operation (very unlikely I guess). It looks like the iOS version always returns true anyway.

stuartmorgan:
Thanks for the contribution! This certainly looks good; it looks like the review missed that it wasn't using an actual queue when making it an instance variable rather than a one-off.

However, we should get a test in place for this in the PR to ensure that it doesn't regress again. There aren't currently native unit tests in place for it, but based on the discussion in the issue it seems like an integration tests that does two sets in row than waits for both to succeed should be able to catch this.

If you can add it that works be great, otherwise I can take a look tomorrow. This will also need a version and CHANGELOG update.

stuartmorgan:
> Unless you need the return code of the operation (very unlikely I guess).

It's definitely not unlikely; it's used 100% of the time (to send back in the response).

> It looks like the iOS version always returns true anyway.

There is no mechanism on iOS for NSUserDefaults to fail and that to be communicated to the caller, so on iOS one has to trust that it always succeeds. That's not the case on Android. The existence of one behavior on one platform doesn't mean that we should treat a different API on a different platform as if it had that behavior.

stuartmorgan:
Test added; I verified that it triggered the bug for me locally without the fix, so it would catch the issue at least sometimes; I don't see a way to test it deterministically unfortunately.

mvanbeusekom (approved):
LGTM!

stuartmorgan:
> Besides, I know that you return the appropriate value on Android, I just doubt that many developers check the return code from `setxxx` calls in SharedPreferences.

There is a fundamental difference between an individual developer choosing not to check the return code, and the plugin making it impossible for any developer to check the return code.

monsieurtanuki:
I have nothing to add to your excerpts about Android. I guess that the notion of synchronous calls on the main thread on Android (probably one of the first things you learn to avoid on Android java) (and where it's suggested to consider using `apply()`) is not that relevant on flutter. And I must say I'm feeling much better developing on flutter than on Android java (though obviously I sometimes still think like an Android java dev).

You consider that the developer should be given the possibility to check _synchronously_ the return code, it's safer, fair enough.
The fact that the behavior is different on other platforms (ios and web always returning `true`) is problematic.

I have absolutely no idea about how I could gracefully add the return code checks to all my `set` calls, and what I should do in case the return code is `false` (I would probably be in big trouble if even `SharedPreferences` does not work).
I admit that I would enjoy being able to check the return code for some important commits (important data, large data).

That's all I have to say about that.

PR #3691, opened 2021-03-08
[path_provider_linux] When getApplicationDocumentsPath() is used, call getApplicationSupportPath()

UnicornsOnLSD:
Currently, `getApplicationDocumentsDirectory()` returns the documents directory on Linux (usually $HOME/Documents). This directory is used for the users documents and shouldn't really be used by programs for storing data. On mobile, `getApplicationDocumentsDirectory()` returns a directory that only the app can access, which will make the developer assume that they can store their app files in the directory returned by `getApplicationDocumentsDirectory()`. For example, [hive_flutter](https://pub.dev/packages/hive_flutter) uses `getApplicationDocumentsDirectory()` to store database files. On Linux, this causes Hive to dump database files in the user's documents directory instead of the app's own folder. In a best case scenario, this annoys the end user. In a worst case scenario, this could cause apps to have conflicts over file names. For example, if two apps saved the user's credentials in a file called `user.json`, those apps would keep overwriting each other, causing those apps to keep breaking.

The solution I've submitted here is kind of jank. `getApplicationSupportPath()` does what I'd expect `getApplicationDocumentsPath()` to do, so I've just made `getApplicationDocumentsPath()` call `getApplicationSupportPath()`. There may be better and more robust solutions to this, but the solution I've submitted is better than what we have now IMO.

A screenshot of my Documents folder, which has the Hive databases for one of my apps

Fixes https://github.com/flutter/flutter/issues/77628

Didn't have to change anything, but this is a breaking change as some apps will now get a new documents directory. This may cause apps to lose their configuration/user data. Since Linux support is still very early, I doubt this will actually affect many people. Still, it'd probably need a big version bump or at least a mention in the changelog.

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
+++ b/packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
@@ -7,7 +7,6 @@ import 'dart:async';
 import 'package:xdg_directories/xdg_directories.dart' as xdg;
 import 'package:path/path.dart' as path;
 import 'package:path_provider_platform_interface/path_provider_platform_interface.dart';
-import 'package:package_info/package_info.dart';
 
 /// The linux implementation of [PathProviderPlatform]
 ///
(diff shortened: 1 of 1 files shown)


UnicornsOnLSD:
@googlebot I signed it!

UnicornsOnLSD:
Closing this since Android just seems to be the outlier and getSupportDirectory should be used by apps for internal files.

PR #2583, opened 2020-03-06
[webview_flutter]  Fix the issue that webview cannot load assets html and local html

asjqkkkk:
Fix flutter/flutter#27086 

Now, flutter webview can load html file in two ways:

#### Assetfiles
```
    WebView(
      onWebViewCreated: (WebViewController controller) {
        final url = 'assets/index.html';
        controller.loadAssetHtmlFile(url);
      },
      javascriptMode: JavascriptMode.unrestricted,
    );
```
#### Localfiles
```
    WebView(
      onWebViewCreated: (WebViewController controller) {
        controller.loadLocalHtmlFile(filePath)
      },
      javascriptMode: JavascriptMode.unrestricted,
    );

  Future<void> downloadFile(String url) async {
    Dio dio = Dio();
    String path = (await getApplicationDocumentsDirectory()).path;
    String fileName=url.substring(url.lastIndexOf("/")+1);
    await Dio().download(url, '$path/$fileName',onReceiveProgress: (rec,total){
      progress = ((rec/total) * 100).floor();
      if(progress >= 100){
        filePath= '$path/$fileName';
      }
      setState(() {});
    });
  }
```

PR #3693, opened 2021-03-09
[webview_flutter] Fix issue where local and asset HTML files cannot be loaded

Wian-TMC:
This PR enables a way for local and asset HTML files to be loaded in by webview_flutter. This was initially completed by @asjqkkkk issue number #2583 but it requires a fresh PR as the branch has been closed.

- Fixed loading local or asset HTML files


stuartmorgan:
Thanks for the submission! We’re currently working through a large backlog of PRs, and this will require non-trivial review (since we need to evaluate whether there are API changes we need to make at the same time), so it will take some time before we’re able to review it. As explained in [CONTRIBUTING.md](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#the-review-process), votes for the corresponding issue are the primary way we’re prioritizing non-trivial reviews, so we encourage you to file an issue and associate it with this PR.

It looks like this is trying to address the same underlying need as https://github.com/flutter/plugins/pull/3613, but with a significantly different approach; it might be useful to discuss and settle on a single PR.

From an initial look, having an explicit method for local content seems like a good approach. I would suggest not trying to solve asset path resolution in this context though; there are many cases where a developer might need to get paths from the assets folder, so embedding that logic in each plugin, vs. having a single place (e.g., path_provider), would be a maintenance issue. There are also non-trivial issues with getting that logic correct for various packaging sceranios such as add-to-app.

Wian-TMC:
@stuartmorgan I'm looking forward to the progress of this PR! Thank you for your insight

Wian-TMC:
Fix flutter/flutter#27086

This PR addresses this issue mentioned above

FelixYew (approved):

stuartmorgan:
@FelixYew Please read https://github.com/flutter/flutter/wiki/Tree-hygiene#how-to-review-code: "contributors without commit access are welcome to comment on PRs, but they should NOT be approving them in the GitHub UI"

stuartmorgan:
Closing as obsolete, since https://github.com/flutter/plugins/pull/4446, https://github.com/flutter/plugins/pull/4486, and to-be-written-soon follow-up PRs add local file handling based on discussion about how we want the suite of load APIs to interact. Thanks again for the submission though!

PR #3697, opened 2021-03-10
Streamline CI setup, and reenable macOS credits

stuartmorgan:
This makes a few improvements to CI throughput:
- Moves the flutter channel+upgrade step into each step, rather then doing it for both channels in the setup step. This avoids pointlessly doing twice the work in every setup (downloading two channels, only one of which will be used)
- Re-enables credits for macOS tasks. These were disabled in a recent chance since I was unifying configurations, but it turns out that without credits there is *no* parallelization of macOS tasks, which makes the end-to-end time very long.
- Dials back the Linux VM requirements based on looking at the new CPU and memory usage graphs for runs:
  - No task was ever reaching 4 CPUs or 12 GB of memory, so use those values for heavy workload; this will double concurrency of the heavy-workload tasks and should have no impact on run length.
  - Linux build+drive was extremely low usage, so move it to the light workload group.

This will increase credit usage relative to the last week or so, but not relative to most of the recent repository history.

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
@jmagman FYI, since this is relevant to compute credit discussions.

script/set_channel.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
jmagman, line 1: If you don't hoist the logic up to the top per my previous suggestion, instead of creating a new bash file you could also instead add a [repeated node](https://yaml.org/spec/1.2/spec.html#id2760395) template to .cirrus.yml so it's in the same file.  Same comment though about needing to add `CHANNEL: "master"` in the few places missing it.
```yaml
set_channel_template: &SET_CHANNEL_TEMPLATE
  script:
    - flutter channel $CHANNEL
    - flutter upgrade
...
  matrix:
    ### Platform-agnostic tasks ###
    - name: plugin_tools_tests
      << : *SET_CHANNEL_TEMPLATE
...
    - name: build_all_plugins_apk
      env:
        matrix:
          CHANNEL: "master"
          CHANNEL: "stable"
      << : *SET_CHANNEL_TEMPLATE
```
etc.

.cirrus.yml
```diff
@@ -9,49 +9,46 @@ task:
...
-    - flutter channel master
-    - flutter upgrade
-    - flutter config --enable-linux-desktop
-    - git fetch origin master
+  setup_script:
+    - git fetch origin master # To set FETCH_HEAD for "git merge-base" to work
   matrix:
     ### Platform-agnostic tasks ###
     - name: plugin_tools_tests
+      upgrade_script: ./script/set_channel.sh master
```
jmagman, line 17: You could pull this up to the `task` level instead of making each named task redeclare it:
```
  setup_script:
    - git fetch origin master # To set FETCH_HEAD for "git merge-base" to work
   upgrade_script: ./script/set_channel.sh $CHANNEL
```
and then set `CHANNEL` in the few places missing it.
```
matrix:
    ### Platform-agnostic tasks ###
    - name: plugin_tools_tests 
    - env:
        CHANNEL: "master"
```
etc

jmagman: Or you can do it the way I tried in 2019 and set CHANNEL to `master` at the top level, and let individual tests override it
https://github.com/jmagman/plugins/blob/41147b069a15a6199b27a4becfcd6d4cdf5b502c/.cirrus.yml#L13-L15
```yaml
  environment:
    CHANNEL: "master" # Default to flutter master, tests can override.
```

stuartmorgan: For some reason I assumed I couldn't use env variables "above" the level in the tree they are set, but in retrospect I'm not sure why I would have thought that. Will reworking to avoid the script.

.cirrus.yml
```diff
@@ -186,35 +180,21 @@ task:
...
-        matrix:
-          CHANNEL: "master"
-          CHANNEL: "stable"
-      build_script:
-        - flutter channel $CHANNEL
-        - ./script/incremental_build.sh build-examples --linux
-        - xvfb-run ./script/incremental_build.sh drive-examples --linux
 
 task:
   # Xcode 12 task
```
jmagman, line 177: Would you mind removing this comment while you're here?  I should have removed it with #3653.

script/set_channel.sh
stuartmorgan: I ended up doing both, for even less duplication :)

stuartmorgan:
@jmagman Thanks for the Cirrus-fu pointers! This is much cleaner, and seem to be working as intended.

.cirrus.yml
```diff
@@ -139,17 +134,9 @@ task:
...
-    INTEGRATION_TEST_PATH: "./packages/integration_test"
-  upgrade_script:
-    - flutter channel stable
-    - flutter upgrade
-    - flutter channel master
-    - flutter upgrade
-    - flutter config --enable-linux-desktop
-    - git fetch origin master
+    cpu: 4
+    memory: 12G
```
jmagman, line 138: Might have to play with this if you start seeing OOM failures.

jmagman (approved):
LGTM if removing the extra channel change works.

.cirrus.yml
stuartmorgan: Yep, I figured this was safe since I wasn't seeing it going above 10 in general, but we can certainly adjust it back up if necessary.

.cirrus.yml
stuartmorgan: Done.

PR #3702, opened 2021-03-11
[video_player_platform_interface] add http headers

PiN73:
This PR is a subset of https://github.com/flutter/plugins/pull/3671 and includes changes only in `video_player_platform_interface` package, as required in [CONTRIBUTING.md](https://github.com/flutter/plugins/blob/d7724ad6f24a6deeec4a88a68d6903252b7961ff/packages/video_player/video_player/CONTRIBUTING.md#publishing-plugin-updates-that-span-multiple-plugin-packages)


packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -169,6 +170,10 @@ class DataSource {
   /// detection with whatever is set here.
   final VideoFormat? formatHint;
 
+  /// HTTP headers used for the request to the [uri].
+  /// Only for [DataSourceType.network] videos.
+  Map<String, String>? httpHeaders;
```
stuartmorgan, line 175: Non-nullable with empty default; see comment in other review.

packages/video_player/video_player_platform_interface/test/method_channel_video_player_test.dart
```diff
@@ -141,8 +141,41 @@ void main() {
...
+        sourceType: DataSourceType.network,
+        uri: 'someUri',
+        httpHeaders: {},
+      ));
+      expect(log.log.last, 'create');
+      expect(log.createMessage?.asset, null);
+      expect(log.createMessage?.uri, 'someUri');
+      expect(log.createMessage?.packageName, null);
+      expect(log.createMessage?.formatHint, null);
+      expect(log.createMessage?.httpHeaders, {});
```
stuartmorgan, line 163: Note that you won't need two different test cases once it's not nullable (which is a feature, since it indicates that there aren't two paths for the same thing).

stuartmorgan (changes requested):
Just one small change, otherwise looks good.

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
PiN73: Fixed it

stuartmorgan (approved):
LGTM!

aliyazdi75:
This seems to have no `httpHeaders` in https://github.com/flutter/plugins/blob/708bf9c09a6c1059b8e3dbba27965e7cd5989b28/packages/video_player/video_player/pigeons/messages.dart#L33

PiN73:
@aliyazdi75 yes, because this PR is just platform_interface part of #3671 which isn't merged yet

aliyazdi75:
@PiN73 oh, ok. They must be separated, for a PR?

PiN73:
@aliyazdi75 yes, as described in [here](https://github.com/flutter/plugins/blob/708bf9c09a6c1059b8e3dbba27965e7cd5989b28/packages/video_player/video_player/CONTRIBUTING.md#publishing-plugin-updates-that-span-multiple-plugin-packages).

In short, the reason is that video_player depends on video_player_platform_interface, new version of which must be published to pub.dev first

PR #3703, opened 2021-03-11
Prep for alignment with Flutter analysis options

stuartmorgan:
Renames the old analysis_options.yaml to analysis_options_legacy.yaml,
replacing it with a slightly modified copy of flutter/flutter's analysis
options.

Each plugins has a temporary local analysis_options.yaml that points to
the legacy version. This allows for inceremental conversion on a
per-plugin basis, which should make the problem more tractable.

Since this hasn't yet been enabled for any packages, it's likely that as
it is we'll find a few local modification we need to make to the root
analysis_options (e.g., things that conflict with 'dart format').

## Pre-launch Checklist

- [x] I signed the [CLA].


analysis_options.yaml
```diff
@@ -1,13 +1,240 @@
-include: package:pedantic/analysis_options.1.8.0.yaml
+# This is a copy (as of March 2021) of flutter/flutter's analysis_options file,
```
ditman, line 1: Note that if you took this from the root of the flutter repo, it gets "extended" in its subdirectories:

* [flutter/packages](https://github.com/flutter/flutter/blob/master/packages/analysis_options.yaml)
* [flutter/packages/flutter](https://github.com/flutter/flutter/blob/master/packages/flutter/analysis_options.yaml)

(I'm not sure which exact flavor is being used when we get tested, but I think this file at least needs the `no_runtimeType_toString` linter rule enabled.

ditman (approved):
Let's go! I'd mark the Plus plugins with a comment in their .yaml so we leave those for last.

analysis_options.yaml
stuartmorgan: Good call, added the two extra rules in a marked section..

ditman:
This change is breaking the rolls into flutter/flutter:

```
Activated flutter_plugin_tools 0.0.45+3.
▌09:47:03▐ ELAPSED TIME: 5.143s for ../../../k/flutter/bin/cache/dart-sdk/bin/pub global activate flutter_plugin_tools in ../../tmp/t/pluginsDYZTNQ
▌09:47:03▐ RUNNING: cd ../../tmp/t/pluginsDYZTNQ; ../../../k/flutter/bin/cache/dart-sdk/bin/pub global run flutter_plugin_tools analyze
Verifying analysis settings...
Found an extra analysis_options.yaml in /b/s/w/ir/tmp/t/pluginsDYZTNQ/packages/android_alarm_manager/analysis_options.yaml.
If this was deliberate, pass the package to the analyze command with the --custom-analysis flag and try again.
```

We should let the analysis command that all our plugins have a custom `analysis_options.yaml` for now?

Investigating.

ditman:
The change needs to happen around here:

https://github.com/flutter/flutter/blob/2e06ed078726e31c5afca36ae9c6736060c7b7e8/dev/bots/test.dart#L851-L860

PR #3719, opened 2021-03-16
[video_player] dont hardcode video fps to 30 on ios

skela:
Videos on iOS were playing at 30 fps regardless of what the fps was according to the media file. This PR fixes that. It will now try to read the fps from the media file, if it fails it falls back on 30 fps.

https://github.com/flutter/flutter/issues/78210


skela:
@googlebot I signed it!

stuartmorgan:
Thanks for the contribution!

You've removed the checklist from the PR template; please refer back to that for some pieces this is missing (notably, version and CHANGELOG updates). Also you'll need to format the code with clang-format to pass CI.

Is this change correct for interlaced video? It would be helpful if you could explain how you've tested this and in what conditions.

stuartmorgan:
I'm going to mark this as a Draft while waiting on the above updates/responses; please feel free to remove the Draft marker once you've updated.

PR #3720, opened 2021-03-16
Add missing licenses, and add a check

stuartmorgan:
Adds a new CI check that all code files have a copyright+license block (and that it's one we are expecting to see).

Fixes the ~350 files (!) that did not have them. This includes all of the files in the .../example/ directories, following the example of flutter/flutter. (This does mean some manual intervention will be needed when generating new example directories in the future, but it's one-time per example.)

Also standardized some variants that used different line breaks than most of the rest of the repo (likely added since I standardized them all a while ago, but didn't add a check for at the time to enforce going forward), to simplify the checks.

Fixes https://github.com/flutter/flutter/issues/77114

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
@Hixie I added license blocks to pigeon-generated files since they didn't have a set naming scheme. I assume that's harmless to do (similar to putting it on files created by `flutter create`), but if it's better to filter them out by the header comment saying they are generated by pigeon I can do that instead.

script/tool/lib/src/license_check_command.dart
```diff
@@ -0,0 +1,162 @@
...
+final RegExp _copyrightRegex = RegExp(r'^(?://|#) Copyright', multiLine: true);
+// All Flutter-authored code.
+final RegExp _bsdLicenseRegex = RegExp(
+    r'^(?://|#) Use of this source code is governed by a BSD-style license',
+    multiLine: true);
+// Other code. When adding license regexes here, include the copyright info to
+// ensure that any new additions are flagged for added scrutiny in review.
+// -----
+// Third-party code used in url_launcher_web.
+final RegExp _workivaLicenseRegex = RegExp(
```
Hixie, line 58: We should enforce that all code that doesn't have our copyright notice is in a third_party directory.

script/tool/lib/src/license_check_command.dart
```diff
@@ -0,0 +1,162 @@
...
+//
+// These are intentionally very simple, since almost all source in this
+// repository should be using the same license text, comment style, etc., so
+// they shouldn't need to be very flexible. Complexity can be added as-needed
+// on a case-by-case basis.
+final RegExp _copyrightRegex = RegExp(r'^(?://|#) Copyright', multiLine: true);
+// All Flutter-authored code.
+final RegExp _bsdLicenseRegex = RegExp(
+    r'^(?://|#) Use of this source code is governed by a BSD-style license',
+    multiLine: true);
```
Hixie, line 53: Can we use a static string rather than a pattern, to match our license block?

stuartmorgan: We could, but it would mean duplicating it for bash vs. other files. What's the issue with the pattern?

(I can put the full text in the pattern, if that's the goal.)

Hixie: Seems like it would be simpler, but I see what you're saying.
FWIW this is what the flutter/flutter repo does: https://github.com/flutter/flutter/blob/master/dev/bots/analyze.dart#L226-L271
It does seem shorter than what we have here and handles bash etc.

Putting the full text in the pattern would be good regardless.

packages/url_launcher/url_launcher_linux/lib/url_launcher_linux.dart
```diff
@@ -1,3 +0,0 @@
-// The url_launcher_platform_interface defaults to MethodChannelUrlLauncher
```
ditman, line 1: Not needed anymore?

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
```diff
@@ -1,3 +1,7 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
 // Autogenerated from Pigeon (v0.1.19), do not edit directly.
```
ditman, line 5: Should Pigeon allow for a "prefix" for the generated file(s)? This change will probably get wiped the next time the files are auto-generated.

script/tool/lib/src/license_check_command.dart
```diff
@@ -0,0 +1,162 @@
...
+  '.c',
+  '.cc',
+  '.cpp',
+  '.dart',
+  '.h',
+  '.java',
+  '.m',
+  '.mm',
+  '.swift',
+  '.sh',
```
ditman, line 22: Needs `.html`? I'm sure our example/web/index.html files don't have a license :/ (I can add them later)

script/tool/lib/src/license_check_command.dart
stuartmorgan: Done.

script/tool/lib/src/license_check_command.dart
stuartmorgan: Made it stricter as a starting point; it checks the full text of both lines, without any whitespace variation allowed.

.cirrus.yml
```diff
@@ -50,6 +50,11 @@ task:
         - ./script/check_publish.sh
     - name: format
       format_script: ./script/incremental_build.sh format --fail-on-change
+      license_script:
+        - cd script/tool
+        - pub get
+        - cd ../..
+        - dart script/tool/lib/src/main.dart license-check
```
ditman, line 57: Is this an incremental command, or does it check the whole repo on every run?

ditman (approved):
I took a quick look at all the licenses, and they seem to be atop of their corresponding files. Made a few comments below. I think the only missing licenses in the repo now are the example/web/index.html files, but it seems those can be corrected later very easily.

script/tool/lib/src/license_check_command.dart
stuartmorgan: Since we have a bunch of variants (four different author phrasings, different years, a handful with commas after the year) I think the static string version would be too messy right now.

This PR did inspire me to continue my planned series of PRs to standardize this whole repo though (derailed by not having sign-off from repo owners to add AUTHORS to all plugins), so I'll pick that back up after this lands, and tighten this down as I go until it's at the point where we can use a string.

.cirrus.yml
stuartmorgan: The latter. Originally I wrote it with the intention of having two modes, complete with a flag for doing only changed files and a marker to myself to implement it once everything else was done. Then I discovered that the version that checks everything takes <5 seconds on my machine, so it seemed like extra work and complexity for nothing 🤷

ditman:
(Also how come it's not running in this PR? Shouldn't this run pre-submit?)

script/tool/lib/src/license_check_command.dart
stuartmorgan: Ah, good call. Will add right now and fix the files (what's another 13 files in this PR)

packages/url_launcher/url_launcher_linux/lib/url_launcher_linux.dart
stuartmorgan: Correct, publishing no longer warns on having no Dart code. I figured it was easier to just remove given that.

ditman: Absolutely, this is great!

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
stuartmorgan: If we want to ignore them, I think we could just do it by matching this line rather than naming convention. @Hixie thoughts?

.cirrus.yml
ditman: > I discovered that the version that checks everything takes <5 seconds on my machine

D A M N

stuartmorgan:
> (Also how come it's not running in this PR? Shouldn't this run pre-submit?)

It is, it's just part of the existing "format" task. I didn't want all the overhead of another task for something that's ~instant to run, so combined it. It seemed closely linked enough to format checks conceptually to add there.

script/tool/lib/src/license_check_command.dart
stuartmorgan: I updated to use a string (using similar code to the flutter/flutter code linked above) for the license part, but not yet the copyright part. I'll switch copyright over once I've standardized.

mvanbeusekom (approved):
LGTM! 

(only thing I noticed but don't think is a problem, is that the year mentioned in the "Copyright 20xx" differs between files)

stuartmorgan:
> (only thing I noticed but don't think is a problem, is that the year mentioned in the "Copyright 20xx" differs between files)

The author and year in this repo are not yet standardized. I had to abandon an initial effort to fix that a while ago, but plan to pick it up again after this.

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/Messages.java
stuartmorgan: Per Discord discussion, we'll keep this (re-adding on regeneration) and revisit if that becomes problematic.

PR #3724, opened 2021-03-17
Null safety broken for method isMarkerInfoWindowShown #78426

stefalda:
The method isMarkerInfoWindowShown in file method_channel_google_maps_flutter.dart isn't working with null safety because the channel(mapId).invokeMethod has an optional return type (<T?>), so forcing it to bool doesn't work.

In the original code a cast has been added to silence the compiler warnings but there is a runtime error when the method is invoked because the signature invokeMethod should be invokeMethod<bool?>.

This is a fix for [issue 78426](https://github.com/flutter/flutter/issues/78426) where there's the original code.


packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
```diff
@@ -414,10 +414,11 @@ class MethodChannelGoogleMapsFlutter extends GoogleMapsFlutterPlatform {
   Future<bool> isMarkerInfoWindowShown(
     MarkerId markerId, {
     required int mapId,
-  }) {
+  }) async {
     assert(markerId != null);
-    return channel(mapId).invokeMethod<bool>('markers#isInfoWindowShown',
-        <String, String>{'markerId': markerId.value}) as Future<bool>;
+    bool? res  = await channel(mapId).invokeMethod<bool?>('markers#isInfoWindowShown',
+        <String, String>{'markerId': markerId.value}) ;
```
stuartmorgan, line 420: Why the intermediate value, vs `return (await ...)!`?

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
```diff
@@ -414,10 +414,11 @@ class MethodChannelGoogleMapsFlutter extends GoogleMapsFlutterPlatform {
   Future<bool> isMarkerInfoWindowShown(
     MarkerId markerId, {
     required int mapId,
-  }) {
+  }) async {
     assert(markerId != null);
-    return channel(mapId).invokeMethod<bool>('markers#isInfoWindowShown',
-        <String, String>{'markerId': markerId.value}) as Future<bool>;
+    bool? res  = await channel(mapId).invokeMethod<bool?>('markers#isInfoWindowShown',
```
stuartmorgan, line 419: Why are you changing this to `<bool?>`? The signature is `Future<T?> invokeMethod <T>`, so I would not expect this to need an optional type.

stuartmorgan (changes requested):
Interesting, this was generated by the migration tool; I wonder why it generated that. Thanks for the contribution!

Please add a test that exercises this codepath (i.e., that fails without your fix and passes with it) per Flutter PR policy and the PR template checklist. This will also need a version and CHANGELOG update per the checklist (in the future, please don't delete the checklist when opening a PR; it's there to help smooth the process).

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
stefalda: You're right, I didn't think of wrapping with () and forcing the result at the end with !

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
stefalda: You're right again, I confused T with the Future<T?> return type.

This should be the final code:
```
  return (await channel(mapId).invokeMethod<bool>('markers#isInfoWindowShown',
        <String, String>{'markerId': markerId.value}))! ;
```

stefalda:
@googlebot I fixed it

stuartmorgan:
Thanks for the version update! Are you planning on adding the test coverage as well, or do you need feedback on something first?

ditman:
Missing CLA for `M@tilde07`, not sure what commit triggered that.

stefalda:
> Missing CLA for `M@tilde07`, not sure what commit triggered that.

Sincerely I don't understand either, I was surprised it asked me to sign the CLA again...

stefalda:
> Thanks for the version update! Are you planning on adding the test coverage as well, or do you need feedback on something first?

I've tried but I've not been able to find the right test where to put the logic, because most of the tests use a fakeGoogleMap instead of the real one, so I'm unable to reproduce the problem that I've discovered while building a real app...

stuartmorgan:
> > Missing CLA for `M@tilde07`, not sure what commit triggered that.
> 
> Sincerely I don't understand either, I was surprised it asked me to sign the CLA again...

The "Updated version and changelog" commit appears to be using a different email address; notice that it doesn't have your profile image. You probably need to ammend that commit locally to use this account's email address, and then force push.

packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: google_maps_flutter
 description: A Flutter plugin for integrating Google Maps in iOS and Android applications.
 homepage: https://github.com/flutter/plugins/tree/master/packages/google_maps_flutter/google_maps_flutter
-version: 2.0.1
+version: 2.0.2
```
stuartmorgan, line 4: Sorry, I hadn't looked at this closely enough; you've updated the version and changelog of the app-facing package, rather than the platform interface package where you made the change. You need to update the version of the package that your changes are in.

stuartmorgan:
> > Thanks for the version update! Are you planning on adding the test coverage as well, or do you need feedback on something first?
> 
> I've tried but I've not been able to find the right test where to put the logic, because most of the tests use a fakeGoogleMap instead of the real one, so I'm unable to reproduce the problem that I've discovered while building a real app...

The test should go in [this group](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/test/platform_interface/google_maps_flutter_platform_test.dart#L40). Currently there are no realy tests there it looks like (just a placeholder that should be removed), but you can look at, e.g., [this](https://github.com/flutter/plugins/blob/67bdfcb5b6f49286dcc17a58107ddd1ab2b20472/packages/device_info/device_info_platform_interface/test/method_channel_device_info_test.dart#L13) to see what a test that sets mock method channel data looks like.

hlavki:
@stefalda can be this problem related also to [issue #78856](https://github.com/flutter/flutter/issues/78856)? thanks

stefalda:
> @stefalda can be this problem related also to [issue #78856](https://github.com/flutter/flutter/issues/78856)? thanks

Yes, I guess that it could be the same problem just in another method call

stuartmorgan:
Since this is a more general problem, I'll follow up shortly with a PR that tests all of the APIs that could have this issue, and fixes all of them, which will replace this one.

stuartmorgan:
Closing in favor of https://github.com/flutter/plugins/pull/3754. Thanks for the clear issue report, debugging, and PR, which definitely accelerated the fix here!

PR #3725, opened 2021-03-17
Add an AUTHORS file to each plugin

stuartmorgan:
This ensures that the AUTHORS list is distributed with the plugins whose copyright blocks refer to the project authors, for ease of reference. This is currently a copy of the top-level authors list, but over time they are expected to diverge as people contribute to specific plugins.

This does not bump plugin versions or CHANGLELOGs, since it has no effect on the plugin itself; it will be picked up over time as we ship new versions of plugins.

Part of https://github.com/flutter/flutter/issues/78448

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -188,8 +188,9 @@ to merge the pull request and `pub submit` any affected packages.
 You must complete the
 [Contributor License Agreement](https://cla.developers.google.com/clas).
 You can do this online, and it only takes a minute.
-If you've never submitted code before, you must add your (or your
-organization's) name and contact info to the [AUTHORS](AUTHORS) file.
+If you've never submitted code for that plugin before, you may also add your (or
+your organization's) name and contact info to the AUTHORS file for the plugin.
+You may also add it to the AUTHORS file for [the repository](AUTHORS).
 
 ### The review process
 
--- a/packages/android_alarm_manager/AUTHORS
+++ b/packages/android_alarm_manager/AUTHORS
@@ -1,0 +1,65 @@
+# Below is a list of people and organizations that have contributed
+# to the Flutter project. Names should be added to the list like so:
+#
+#   Name/Organization <email address>
+
+Google Inc.
+German Saprykin <saprykin.h@gmail.com>
+Benjamin Sauer <sauer.benjamin@gmail.com>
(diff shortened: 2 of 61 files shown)


CONTRIBUTING.md
```diff
@@ -188,8 +188,9 @@ to merge the pull request and `pub submit` any affected packages.
 You must complete the
 [Contributor License Agreement](https://cla.developers.google.com/clas).
 You can do this online, and it only takes a minute.
-If you've never submitted code before, you must add your (or your
-organization's) name and contact info to the [AUTHORS](AUTHORS) file.
+If you've never submitted code for that plugin before, you may also add your (or
+your organization's) name and contact info to the AUTHORS file for the plugin.
+You may also add it to the AUTHORS file for [the repository](AUTHORS).
```
ditman, line 193: Should we compute the initial version of the per-plugin AUTHORS from the git log?

stuartmorgan: I considered it, but it would be significantly more time-intensive, and I couldn't see any real benefit. The authors file isn't a legal source of truth (the CLA and the git history are the definitive information, as I understand it), just a way to recognize contributions, and I don't think anyone will be unhappy about being slightly over-represented.

It's an odd situation in the first place since normally there is just one such file in a repo; parsing contributions by subfolder isn't the norm. Looking at it another way: if someone currently wanted to find out who "The Flutter Authors" referred to in a specific plugin, they would find exactly this same list, it would just be harder to do so, so nothing is fundamentally changing.

cyanglaz (approved):
LGTM!

CONTRIBUTING.md
ditman: Do we want a per-package AUTHORS, then? flutter/flutter has a root AUTHORS file, but then each individual package doesn't have an individual one.

stuartmorgan: Hixie wanted these because the published packages (i.e., what is downloaded from pub.dev) have a license file that reference "the authors", but no authors list, which is kind of weird. (This is weird in a different way, but here the weirdness is restricted to the repo instead of what we distribute.)

I assume nothing in flutter/flutter is published as a package.

PR #3726, opened 2021-03-18
[google_maps_flutter_web] Migrate to null-safety.

ditman:
This PR migrates the `google_maps_flutter_web` package to null safety. To do so, certain things needed to change:

* The underlying `google_maps` plugin has been updated to `^5.0.0`. Thanks @a14n!
* Tests had to be partially re-written due to the differences in null-safety `mockito`, and how `google_maps` is now put together (mockito can't mock extension methods).

The following changes were applied to the mocking strategy of this package:

* Use Mockito 5 to generate mocks of the classes defined in this package (mainly, geometry controllers)
* Use partially initialized JS Objects from the js-interop layer (functioning "Fakes"). These are supported by the underlying Maps JS SDK, so we leverage that!

### Issues

Fixes https://github.com/flutter/flutter/issues/74261
Fixes https://github.com/flutter/flutter/issues/75247

### Testing

* All existing tests pass. Modifications as minimal as possible.
* Deployed manually here: https://dit-maps-tests.web.app

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/google_maps_plugin_test.dart
```diff
@@ -74,16 +78,7 @@ void main() {
...
-            null,
-            onPlatformViewCreated,
-            initialCameraPosition: initialCameraPosition,
-          ),
-          throwsAssertionError,
-          reason:
-              '_webOnlyMapCreationId is mandatory to prevent unnecessary reloads in web.',
-        );
+        // The argument type 'Null' can't be assigned to the parameter type 'int'.
       });
```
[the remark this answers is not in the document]
ditman: In fact, I'm going to try and create a couple of small @dart=2.9 test with these tests, see what happens :P

packages/google_maps_flutter/google_maps_flutter_web/lib/src/circle.dart
```diff
@@ -26,21 +26,23 @@ class CircleController {
...
-  gmaps.Circle get circle => _circle;
+  gmaps.Circle? get circle => _circle;
 
   /// Returns `true` if this Controller will use its own `onTap` handler to consume events.
   bool get consumeTapEvents => _consumeTapEvents;
 
   /// Updates the options of the wrapped [gmaps.Circle] object.
   void update(gmaps.CircleOptions options) {
-    _circle.options = options;
+    _circle!.options = options;
```
stuartmorgan, line 36: Either this should have a null check, or the docs should say that it's a programming error to call it after `remove()`.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -202,48 +128,53 @@ bool _isJsonMapStyle(Map value) {
...
+        })
+        .cast<gmaps.MapTypeStyle>()
+        .toList();
+    // .toList calls are required so the JS API understands the underlying data structure.
   }
   return styles;
 }
 
-gmaps.LatLng _latLngToGmLatLng(LatLng latLng) {
+gmaps.LatLng? _latLngToGmLatLng(LatLng latLng) {
```
stuartmorgan, line 152: It looks like this can only be null if the param is null, which would already be a programming error. It seems like we shouldn't bake this into the API just for mixed mode support.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -202,48 +128,53 @@ bool _isJsonMapStyle(Map value) {
...
   if (latLng == null) return null;
   return gmaps.LatLng(latLng.latitude, latLng.longitude);
 }
 
-LatLng _gmLatLngToLatLng(gmaps.LatLng latLng) {
+LatLng _gmLatLngToLatLng(gmaps.LatLng? latLng) {
```
stuartmorgan, line 157: Related to the above: why the different nullability semantics in this reverse method?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -265,14 +196,14 @@ gmaps.InfoWindowOptions _infoWindowOptionsFromMarker(Marker marker) {
   if (marker.infoWindow.title?.isNotEmpty ?? false) {
     final HtmlElement title = HeadingElement.h3()
       ..className = 'infowindow-title'
-      ..innerText = marker.infoWindow.title;
+      ..innerText = marker.infoWindow.title!;
```
stuartmorgan, line 199: Totally optional nit: could be nice to extract locals for `marker.infoWindow.title` and `...snippet`, at which point you wouldn't need the `!`

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -351,14 +282,14 @@ gmaps.PolygonOptions _polygonOptionsFromPolygon(
     gmaps.GMap googleMap, Polygon polygon) {
   List<gmaps.LatLng> path = [];
   polygon.points.forEach((point) {
-    path.add(_latLngToGmLatLng(point));
+    path.add(_latLngToGmLatLng(point)!);
```
stuartmorgan, line 285: It's not clear why this and the below are safe (argument for changing the API to return non-null)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -391,8 +322,8 @@ bool _isPolygonClockwise(List<gmaps.LatLng> path) {
   var direction = 0.0;
   for (var i = 0; i < path.length; i++) {
     direction = direction +
-        ((path[(i + 1) % path.length].lat - path[i].lat) *
-            (path[(i + 1) % path.length].lng + path[i].lng));
+        ((path[(i + 1) % path.length].lat! - path[i].lat!) *
+            (path[(i + 1) % path.length].lng! + path[i].lng!));
```
stuartmorgan, line 326: Why are these safe?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -401,7 +332,7 @@ gmaps.PolylineOptions _polylineOptionsFromPolyline(
     gmaps.GMap googleMap, Polyline polyline) {
   List<gmaps.LatLng> paths = [];
   polyline.points.forEach((point) {
-    paths.add(_latLngToGmLatLng(point));
+    paths.add(_latLngToGmLatLng(point)!);
```
stuartmorgan, line 335: Same as earlier comment.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -460,16 +391,16 @@ void _applyCameraUpdate(gmaps.GMap map, CameraUpdate update) {
           // print('Error computing new focus LatLng. JS Error: ' + e.toString());
         }
       }
-      map.zoom = map.zoom + newZoomDelta;
+      map.zoom = map.zoom! + newZoomDelta;
```
stuartmorgan, line 394: Why are all these `zoom!`'s safe?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -481,17 +412,17 @@ void _applyCameraUpdate(gmaps.GMap map, CameraUpdate update) {
 
 // original JS by: Byron Singh (https://stackoverflow.com/a/30541162)
 gmaps.LatLng _pixelToLatLng(gmaps.GMap map, int x, int y) {
-  final ne = map.bounds.northEast;
-  final sw = map.bounds.southWest;
-  final projection = map.projection;
+  final ne = map.bounds!.northEast;
+  final sw = map.bounds!.southWest;
+  final projection = map.projection!;
```
stuartmorgan, line 417: Why is all of this safe?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
```diff
@@ -41,14 +41,14 @@ class GoogleMapController {
...
+  gmaps.TrafficLayer? _trafficLayer;
 
   /// A getter for the current traffic layer. Only for tests.
   @visibleForTesting
-  gmaps.TrafficLayer get trafficLayer => _trafficLayer;
+  gmaps.TrafficLayer? get trafficLayer => _trafficLayer;
 
   // The underlying GMap instance. This is the interface with the JS SDK.
-  gmaps.GMap _googleMap;
+  gmaps.GMap? _googleMap;
```
stuartmorgan, line 51: `late`?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
```diff
@@ -255,30 +255,30 @@ class GoogleMapController {
...
-    return _gmLatLngBoundsTolatLngBounds(await _googleMap.bounds);
+    return _gmLatLngBoundsTolatLngBounds(await _googleMap!.bounds);
   }
 
   /// Returns the [ScreenCoordinate] for a given viewport [LatLng].
   Future<ScreenCoordinate> getScreenCoordinate(LatLng latLng) async {
     final point =
-        _googleMap.projection.fromLatLngToPoint(_latLngToGmLatLng(latLng));
-    return ScreenCoordinate(x: point.x, y: point.y);
+        _googleMap!.projection!.fromLatLngToPoint!(_latLngToGmLatLng(latLng))!;
```
stuartmorgan, line 264: That's a lot of `!`...

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker.dart
```diff
@@ -6,33 +6,33 @@ part of google_maps_flutter_web;
...
         onTap.call();
       });
     }
     if (onDragEnd != null) {
-      _marker.onDragend.listen((event) {
-        _marker.position = event.latLng;
-        onDragEnd.call(event.latLng);
+      _marker!.onDragend.listen((event) {
+        _marker!.position = event.latLng;
+        onDragEnd.call(event.latLng!);
```
stuartmorgan, line 35: Why is this safe?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker.dart
```diff
@@ -6,33 +6,33 @@ part of google_maps_flutter_web;
...
+      _marker!.onClick.listen((event) {
         onTap.call();
       });
     }
     if (onDragEnd != null) {
-      _marker.onDragend.listen((event) {
-        _marker.position = event.latLng;
-        onDragEnd.call(event.latLng);
+      _marker!.onDragend.listen((event) {
+        _marker!.position = event.latLng;
```
stuartmorgan, line 34: Is it guaranteed that this can't fire after `dispose`?

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/google_maps_plugin_test.dart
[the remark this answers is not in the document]
ditman: In the end, since this new package is a major version, I've decided to drop support for this feature, and remove the obsolete tests. I've added a "breaking changes" section to the README to document these changes in behavior. Thanks for bringing them up!

packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
```diff
@@ -2,8 +2,8 @@ name: google_maps_flutter_web_integration_tests
 publish_to: none
 
 environment:
-  sdk: ">=2.2.2 <3.0.0" # Bump this to 2.12 when migrating to nnbd.
-  flutter: ">=1.27.0-0" # For integration_test from sdk
+  sdk: ">=2.12.0 <3.0.0" # Bump this to 2.12 when migrating to nnbd.
```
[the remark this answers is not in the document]
ditman: Definitely

packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/google_maps_plugin_test.dart
```diff
@@ -119,7 +115,7 @@ void main() {
 
       testWidgets('returns cached instance if it already exists',
           (WidgetTester tester) async {
-        final expected = HtmlElementView(viewType: 'only-for-testing');
+        final Widget expected = HtmlElementView(viewType: 'only-for-testing');
```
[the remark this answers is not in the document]
ditman: No need, I think this is a leftover from when I was experimenting with a mockable return value for `controller.widget` (had to move that to `Widget?` for everything to work nicely)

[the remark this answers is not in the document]
ditman: Removed.

packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
```diff
@@ -2,8 +2,8 @@ name: google_maps_flutter_web_integration_tests
...
+  flutter: ">=2.1.0-0" # For integration_test from sdk
```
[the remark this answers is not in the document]
ditman: Done!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -202,48 +128,53 @@ bool _isJsonMapStyle(Map value) {
 }
 
 // Converts an incoming JSON-encoded Style info, into the correct gmaps array.
-List<gmaps.MapTypeStyle> _mapStyles(String mapStyleJson) {
+List<gmaps.MapTypeStyle?> _mapStyles(String? mapStyleJson) {
```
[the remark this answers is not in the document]
ditman: Yep, totally right.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -202,48 +128,53 @@ bool _isJsonMapStyle(Map value) {
...
+gmaps.LatLng? _latLngToGmLatLng(LatLng latLng) {
   if (latLng == null) return null;
   return gmaps.LatLng(latLng.latitude, latLng.longitude);
 }
 
-LatLng _gmLatLngToLatLng(gmaps.LatLng latLng) {
+LatLng _gmLatLngToLatLng(gmaps.LatLng? latLng) {
   if (latLng == null) return _nullLatLng;
-  return LatLng(latLng.lat, latLng.lng);
+  return LatLng(latLng.lat!.toDouble(), latLng.lng!.toDouble());
```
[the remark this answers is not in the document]
ditman: Have you ever seen cases of a non-null latLng object that had either lat/lng null? I'd rather that case crash at runtime so the user is aware, than return (0, 0) by default :S

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -202,48 +128,53 @@ bool _isJsonMapStyle(Map value) {
...
+LatLngBounds _gmLatLngBoundsTolatLngBounds(gmaps.LatLngBounds? latLngBounds) {
   if (latLngBounds == null) {
     return _nullLatLngBounds;
   }
 
   return LatLngBounds(
-    southwest: _gmLatLngToLatLng(latLngBounds.southWest),
-    northeast: _gmLatLngToLatLng(latLngBounds.northEast),
+    southwest: _gmLatLngToLatLng(latLngBounds.southWest!),
+    northeast: _gmLatLngToLatLng(latLngBounds.northEast!),
```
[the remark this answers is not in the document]
ditman: (Same idea as above)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
```diff
@@ -213,78 +213,78 @@ class GoogleMapsPlugin extends GoogleMapsFlutterPlatform {
...
     PlatformViewCreatedCallback onPlatformViewCreated, {
-    @required CameraPosition initialCameraPosition,
+    required CameraPosition initialCameraPosition,
     Set<Marker> markers = const <Marker>{},
     Set<Polygon> polygons = const <Polygon>{},
     Set<Polyline> polylines = const <Polyline>{},
     Set<Circle> circles = const <Circle>{},
     Set<TileOverlay> tileOverlays = const <TileOverlay>{},
-    Set<Factory<OneSequenceGestureRecognizer>> gestureRecognizers =
+    Set<Factory<OneSequenceGestureRecognizer>>? gestureRecognizers =
```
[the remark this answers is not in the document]
ditman: This being nullable is being required by the [platform interface](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_flutter_platform.dart#L351).

packages/google_maps_flutter/google_maps_flutter_web/lib/src/circle.dart
```diff
@@ -6,15 +6,15 @@ part of google_maps_flutter_web;
 
 /// The `CircleController` class wraps a [gmaps.Circle] and its `onTap` behavior.
 class CircleController {
-  gmaps.Circle _circle;
+  gmaps.Circle? _circle;
```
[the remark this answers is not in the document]
ditman: I'm just trying to guarantee that I'm not holding on to JS instances (`gmaps.*`) after the programmer has decided they're done with them. That's why I want to set the reference to `null` on `remove`. I'll add some more null checks / asserts to the methods that can't be called after `remove`.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/circle.dart
ditman: I've added an assertion, and extra docs to specify what methods can't be called after `remove`. I'll add some tests too.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: Ohh, good catch on this file. Let me give it another pass. Thanks for noticing!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: Everything coming from the web (gmaps.*) can be null. These methods are not very popular, so I'm going to just provide default values for where these are used.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: Added some docs with regards to how this is used. It is safe now to use `!` now, because of the LatLng transformation function not returning maybe nulls anymore, and what this is actually looking at.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: I just can't imagine how an user can get to this point of the code having null `zoom` in their Maps :)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
ditman: Haha yeah, the underlying API is all nullable `?`, and the previous code did a lot of calls without null-checking in the middle.

There seems to be some wiring that can be done within the map to know when `projection` is ready, but that hasn't been required so far. The rest is still nullable.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker.dart
ditman: `_marker` is `required` and non-null in the constructor of the `MarkerController`.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker.dart
ditman: dispose immediately removes the map from the geometry. These won't fire after dispose, because the user won't be able to interact with the actual object.

I suspect this can be faked in unit-testing, however.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker.dart
ditman: event is a gmaps.MapMouseEvent, and it seems latLng [may be null](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/google.maps/index.d.ts#L3197)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
ditman: No, I want to clear `gmaps.GMap` on `dispose`, same as with the other geometry objects.

ditman: I've added a note to dispose that you can't call "many of the methods" of the controller after `dispose`. However this code is used through the gmaps plugin, and not directly by end users.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: Yep, good call!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: I don't understand why the next line of code didn't fail :/

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: Replaced by `(map.zoom ?? 0)` and made 0 the default value for zoom (according to the JS api, when zoom is not initialized)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker.dart
ditman: I've added some extra sanity checks around this onDragEnd call so `!` are warranted.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: I added a bunch of asserts to make it safe, or throwy-with-an-explanation.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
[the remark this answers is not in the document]
a14n: As `new google.maps.LatLng().lat()` returns `NaN` and `new google.maps.LatLng(null, null).lat()` returns `0` I guess I can make lat and lng non-nullable on google_maps

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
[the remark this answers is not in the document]
a14n: On JS side `new google.maps.LatLngBounds().getNorthEast()` and  `new google.maps.LatLngBounds(null,null).getNorthEast()` returns something non-null. So I should make them non-nullable.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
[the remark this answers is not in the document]
a14n: Yes but shouldn't the platform interface be change to non-nullable as well? Is there a meaning to call `buildView(....., gestureRecognizers: null)` instead of `buildView(.....)` ?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
a14n: >I don't understand why the next line of code didn't fail :/

Because of 
https://github.com/flutter/plugins/blob/15d4f9c4e06014225807c5fbcf681bd6ea1f765d/analysis_options.yaml#L49-L50 ?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
a14n: On JS side you can do really strange things :-/

```
> map.setZoom(null)
undefined
> map.getZoom()
null
```

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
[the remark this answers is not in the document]
stuartmorgan: IIRC I allowed it to be nullable because it's interacting with a framework API that has the same signature, and I opted for consistency with the framework. Maybe the framework API should be non-nullable too, but I didn't want to shave that yak while we were in a time crunch to migrate all the plugins.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: (x_x) totally

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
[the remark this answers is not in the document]
ditman: In our case, plugin users never create (or see) the underlying gmaps.LatLng objects, I think we only expose our own LatLng from the platform_interface of the plugin (which has [non-nullable latitude/longitude](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/location.dart#L18)).

@a14n do you have to bump the major version of the package every time you make something non-nullable?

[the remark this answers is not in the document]
a14n: > @a14n do you have to bump the major version of the package every time you make something non-nullable?

Making a parameter non-null seems more breaking (user can not pass `null` anymore) than making a return type non-null (user may see new diagnostic about unnecessary null checks).

But one could also argument that it was a bug in the current implementation that doesn't required a major version bump ;)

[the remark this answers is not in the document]
a14n: I have been told that a 5.1.0 has been released.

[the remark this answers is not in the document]
ditman: I'll update to 5.1.0 right away :P

ditman:
I need to rebase this with the latest master, to accomodate the changes made in [ci] and web plugins.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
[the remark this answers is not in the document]
ditman: Done!

ditman:
(This should now attempt to run the modified tests in the `build-web+drive-examples CHANNEL:master` step!)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -481,17 +416,27 @@ void _applyCameraUpdate(gmaps.GMap map, CameraUpdate update) {
...
+  final ne = bounds!.northEast;
+  final sw = bounds.southWest;
 
-  final topRight = projection.fromLatLngToPoint(ne);
-  final bottomLeft = projection.fromLatLngToPoint(sw);
+  final topRight = projection!.fromLatLngToPoint!(ne)!;
+  final bottomLeft = projection.fromLatLngToPoint!(sw)!;
 
-  final scale = 1 << map.zoom; // 2 ^ zoom
+  final scale = 1 << (zoom! as int); // 2 ^ zoom
```
stuartmorgan, line 436: What is the type of `zoom`? I'm curious why this isn't `.toInt()`

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
```diff
@@ -144,8 +144,8 @@ class GoogleMapController {
     // Create the map...
     _googleMap = _createMap(_div, options);
 
-    _attachMapEvents(_googleMap);
-    _attachGeometryControllers(_googleMap);
+    _attachMapEvents(_googleMap!);
+    _attachGeometryControllers(_googleMap!);
```
stuartmorgan, line 148: Optional: You could make a local, set `_googleMap` to the local, then use the local everywhere instead of `!`.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
```diff
@@ -255,30 +255,31 @@ class GoogleMapController {
...
+        await _googleMap!.bounds ?? _nullGmapsLatLngBounds);
   }
 
   /// Returns the [ScreenCoordinate] for a given viewport [LatLng].
   Future<ScreenCoordinate> getScreenCoordinate(LatLng latLng) async {
     final point =
-        _googleMap.projection.fromLatLngToPoint(_latLngToGmLatLng(latLng));
-    return ScreenCoordinate(x: point.x, y: point.y);
+        _googleMap!.projection!.fromLatLngToPoint!(_latLngToGmLatLng(latLng))!;
+    return ScreenCoordinate(x: point.x! as int, y: point.y! as int);
```
stuartmorgan, line 266: `toInt()`?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
```diff
@@ -314,6 +314,6 @@ class GoogleMapsPlugin extends GoogleMapsFlutterPlatform {
 
     onPlatformViewCreated.call(creationId);
 
-    return mapController.widget;
+    return mapController.widget!;
```
stuartmorgan, line 317: Why is this safe? Could use a comment.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker.dart
```diff
@@ -6,33 +6,35 @@ part of google_maps_flutter_web;
...
-    LatLngCallback onDragEnd,
-    ui.VoidCallback onTap,
+    LatLngCallback? onDragEnd,
+    ui.VoidCallback? onTap,
   })  : _marker = marker,
         _infoWindow = infoWindow,
         _consumeTapEvents = consumeTapEvents {
     if (onTap != null) {
-      _marker.onClick.listen((event) {
+      _marker!.onClick.listen((event) {
```
stuartmorgan, line 28: Maybe this method should use `marker` instead everywhere.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -481,17 +416,27 @@ void _applyCameraUpdate(gmaps.GMap map, CameraUpdate update) {
...
+  assert(zoom != null,
+      'Current map zoom level required to compute LatLng of screen x/y');
+
+  final ne = bounds!.northEast;
+  final sw = bounds.southWest;
 
-  final topRight = projection.fromLatLngToPoint(ne);
-  final bottomLeft = projection.fromLatLngToPoint(sw);
+  final topRight = projection!.fromLatLngToPoint!(ne)!;
+  final bottomLeft = projection.fromLatLngToPoint!(sw)!;
```
stuartmorgan, line 434: I think a comment here about why this is safe would be useful.

stuartmorgan (changes requested):
Mostly small comment. I'm still concerned at a high level by the amount of `!` usage in ways that are not locally obviously correct (e.g., not cases like using a field right after checking that it's not null).

A lot of that is driven by the `dispose` system. I think it would be worth trying to separate the objects that have that pattern into a wrapper and an impl, where the impl has essentially all of the logic and all the fields are non-null, and the wrapper forwards to that object, and has `dispose` throw the whole impl away. It might be relatively straightforward to change?

packages/google_maps_flutter/google_maps_flutter_web/example/README.md
```diff
@@ -19,3 +19,17 @@ Make sure you have updated to the latest Flutter master.
...
+
+Mocks are [auto-generated by `package:mockito`](https://github.com/dart-lang/mockito/blob/master/NULL_SAFETY_README.md#code-generation).
+
+Mocks can be manually re-generated with the following command: `flutter pub run build_runner build`, or by running the `regen_mocks.sh` script.
+
+Mocks may change with changes to the API of the classes they're mocking, but also by how they're used within the tests.
+
+Remember to regenerate the mocks often, and feel free to commit any changes in them.
+
+(Mocks will be auto-generated by the `run_test.sh` script as well.)
```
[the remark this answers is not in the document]
ditman: Rephrased a little bit!

packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/google_maps_controller_test.dart
```diff
@@ -57,7 +44,7 @@ void main() {
       Set<Polygon> polygons = const <Polygon>{},
       Set<Polyline> polylines = const <Polyline>{},
       Set<Circle> circles = const <Circle>{},
-      Map<String, dynamic> options,
+      Map<String, dynamic>? options,
```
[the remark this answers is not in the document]
ditman: Done

packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/marker_test.dart
```diff
@@ -2,100 +2,144 @@
...
-
-class _MockInfoWindow extends Mock implements gmaps.InfoWindow {}
 
 /// Test Markers
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
-  bool called = false;
+  late Completer<bool> _called;
+  late Future<bool> called;
```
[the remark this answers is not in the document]
ditman: Yep, renamed and documented.

packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/shape_test.dart
```diff
@@ -2,114 +2,179 @@
...
-  Stream<gmaps.PolyMouseEvent> get onClick => onClickController.stream;
-}
 
 /// Test Shapes (Circle, Polygon, Polyline)
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
-  bool called = false;
+  late Completer<bool> _called;
+  late Future<bool> called;
```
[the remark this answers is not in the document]
ditman: Renamed and documented.

packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/shape_test.dart
```diff
@@ -2,114 +2,179 @@
...
+      expect(circle.draggable, isTrue);
+    });
+
+    group('remove', () {
+      late CircleController controller;
+
+      setUp(() {
+        controller = CircleController(circle: circle);
+
+        controller.remove();
```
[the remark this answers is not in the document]
ditman: Yeah, this felt very weird when writing it too :/ Moved this line to the test.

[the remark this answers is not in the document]
ditman: Also corrected in the marker_test.dart

packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
```diff
@@ -1,9 +1,10 @@
 name: google_maps_flutter_web_integration_tests
 publish_to: none
 
+# Tests require flutter beta or greater to run.
 environment:
-  sdk: ">=2.2.2 <3.0.0" # Bump this to 2.12 when migrating to nnbd.
-  flutter: ">=1.27.0-0" # For integration_test from sdk
+  sdk: ">=2.12.0 <3.0.0"
+  flutter: ">=1.27.0" # For integration_test with null safety from sdk (should be >=2.1.0)
```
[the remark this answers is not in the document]
ditman: I've added better documentation and a >=2.0.0 constraint.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -202,37 +127,36 @@ bool _isJsonMapStyle(Map value) {
...
+              ..elementType = value['elementType']
+              ..featureType = value['featureType']
+              ..stylers =
+                  (value['stylers'] as List).map((e) => jsify(e)).toList();
+          }
+          return value;
+        })
+        .cast<gmaps.MapTypeStyle>()
+        .toList();
+    // .toList calls are required so the JS API understands the underlying data structure.
```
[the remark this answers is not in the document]
ditman: If you remove the .toList() call right above this, everything will [compile as normal](https://api.dart.dev/stable/2.12.2/dart-convert/jsonDecode.html), but map styling will crash at runtime because JS is not able to iterate the ["view" returned by `cast`](https://api.flutter.dev/flutter/dart-core/List/cast.html).

Precisely because the stylers seemed more "obvious" that were being cast .toList(), I left the comment here. The temptation to remove this last .toList seemed far greater (maybe it's just me?)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -241,38 +165,42 @@ LatLngBounds _gmLatLngBoundsTolatLngBounds(gmaps.LatLngBounds latLngBounds) {
...
 
-gmaps.InfoWindowOptions _infoWindowOptionsFromMarker(Marker marker) {
-  if ((marker.infoWindow?.title?.isEmpty ?? true) &&
-      (marker.infoWindow?.snippet?.isEmpty ?? true)) {
+gmaps.InfoWindowOptions? _infoWindowOptionsFromMarker(Marker marker) {
+  final markerTitle = marker.infoWindow.title;
+  final markerSnippet = marker.infoWindow.snippet;
+
+  // If both the title and the snippet are null or empty, bail out...
+  if ((markerTitle?.isEmpty ?? true) && (markerSnippet?.isEmpty ?? true)) {
```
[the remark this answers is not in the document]
ditman: Ugh, there's days that I should be as far as possible from source code!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -460,16 +395,16 @@ void _applyCameraUpdate(gmaps.GMap map, CameraUpdate update) {
...
       }
-      map.zoom = map.zoom + newZoomDelta;
+      map.zoom = (map.zoom ?? 0) + newZoomDelta;
       if (focusLatLng != null) {
         map.panTo(focusLatLng);
       }
       break;
     case 'zoomIn':
-      map.zoom++;
+      map.zoom = (map.zoom ?? 0) + 1;
```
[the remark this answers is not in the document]
ditman: xD in a correctly initialized map, zoom is always defined, bbbbbut I can create a partially initialized map that has zoom = undefined, so this needs to stay `?`.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: This boi is `num?`. I did a quick test, and `.toInt()` is a much more robust cast:

```dart
void main() {
  num ber = 7.5;
  print(ber.toInt());
  print(ber as int);
}
```

Outputs:

```
7
Uncaught Error: TypeError: 7.5: type 'JSDouble' is not a subtype of type 'int'
```

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
ditman: Yep, sweeping all these `as int` as `toInt`!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
ditman: Indeed, done.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
ditman: I've added an assert here. `widget` should never be null, unless `dispose` has been called on a map.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker.dart
ditman: Yes, in fact the other geometry controllers use the incoming object, instead of the local version.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: (Asserts to the rescue)

ditman:
> A lot of that is driven by the dispose system. I think it would be worth trying to separate the objects that have that pattern into a wrapper and an impl, where the impl has essentially all of the logic and all the fields are non-null, and the wrapper forwards to that object, and has dispose throw the whole impl away. It might be relatively straightforward to change?

@stuartmorgan for now, I ended up guarding all the `!` with `asserts`. The plugin seems a little bit more throwy now than before, but all tests continued normally without real changes (in fact, I added tests to trigger some of those asserts). I'd rather not add another level of indirection to the controllers; what you're describing is essentially what the code is already doing (with the internal gmaps instances).

This could benefit from a refactor more in the style that you did with the geometry stuff in the core package (all "geometry" shares a set of base classes that have some default behavior, like dispose), but I'd rather introduce that change in another PR. This one is complex enough as is :/

stuartmorgan (approved):
Okay, LGTM then. I do think followup refactoring to reduce the use of `!` would be helpful, but I'm fine with proceeding as-is.

ditman:
Rebasing with the latest master to resolve example/pubspec.yaml conflict.

PR #3727, opened 2021-03-18
[video_player] Pause video when it completes

KyleFin:
Instead of just updating the `VideoPlayerValue` `isPlaying` and `position` fields when a video completes, have the platform pause and seek to the last frame of the video.

Fixes https://github.com/flutter/flutter/issues/77674

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/video_player/video_player/lib/video_player.dart
```diff
@@ -375,6 +374,9 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   /// has been sent to the platform, not when playback itself is totally
   /// finished.
   Future<void> play() async {
+    if (value.position == value.duration) {
+      await seekTo(const Duration());
+    }
```
KyleFin, line 379: If video is at the end, play() restarts the video from the beginning. This isn't necessary to fix the original issue, but it makes the behavior on iOS and Android consistent with web.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -314,8 +314,7 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
           _applyPlayPause();
           break;
         case VideoEventType.completed:
-          value = value.copyWith(isPlaying: false, position: value.duration);
-          _timer?.cancel();
+          pause().then((void pauseResult) => seekTo(value.duration));
```
KyleFin, line 317: `pause()` sets isPlaying false and cancels the timer, but also notifies the platform to pause the video. Previously, the video would pause in the UI, but after the video completed:
- Invoking `seekTo(duration)` would cause the video to resume playing from duration without updating value.position or value.isPlaying (the original bug).
- Invoking `play()`  would cause the timestamp counter to continue from wherever the video was still counting beyond when the video ended.

`then(seekTo(value.duration))` updates `position` to `value.duration` and tells the platform to update to the last frame in the video. Without this, we may receive a `completed` event slightly before the video ends, resulting in `position` not being equal to the `duration` (e.g. UI will show something like "00:03.97 / 00:04.00"). I originally just used `value.copyWith(position: value.duration)` to update the UI, but then we're technically displaying video a few frames different from the timestamp displayed in the UI.

KyleFin, line 317: Is `void pauseResult` the right way to "[avoid anonymous parameter names](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-anonymous-parameter-names)"?

stuartmorgan:
Thanks for the contribution! Per the checklist and Flutter's PR policies, this will need tests covering the change(s) you are making here as part of the PR. Once those are in place, we can move forward with a review.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -429,7 +431,37 @@ void main() {
         await tester.pumpAndSettle();
 
         expect(controller.value.isPlaying, isFalse);
-        expect(controller.value.position, controller.value.duration);
+        expect(controller.value.position, nonzeroDuration);
```
KyleFin, line 434: Previously, `duration` was zero so this assertion would pass even if `completed` didn't make any change to `position`.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -429,7 +431,37 @@ void main() {
...
+        expect(
+            fakeVideoPlayerPlatform
+                .calls[fakeVideoPlayerPlatform.calls.length - 2],
+            'pause');
+        expect(fakeVideoPlayerPlatform.calls.last, 'seekTo');
```
KyleFin, line 439: I would prefer to check that the underlying platform actually stopped playing (instead of testing implementation details), but it wasn't clear to me how to do that using the existing test fakes.

If it's worth the effort, I'll dig more into how I might access the underlying platform in unit tests or set up a [golden file](https://github.com/flutter/flutter/wiki/Writing-a-golden-file-test-for-package%3Aflutter) test or [device lab](https://github.com/flutter/flutter/wiki/Running-and-writing-tests#running-device-lab-tests-locally) tests for `video_player`. (Those feel rather heavy-handed for this change, but the setup seems like it would be generally useful for `video_player`.)

KyleFin (commented):
Is this change worth mentioning in CHANGELOG.md and updating version in pubspec.yaml?

packages/video_player/video_player/test/video_player_test.dart
KyleFin: Integration tests in packages/video_player/video_player/example/integration_test/video_player_test.dart looks like it may be a good place to test this. I'll look into it more now.

KyleFin: I'll plan to wait until a reviewer has bandwidth to provide guidance.

Here are some things I found in the rabbit hole:
- Running tests in packages/video_player/video_player/example/integration_test/video_player_test.dart with `flutter test` fails (maybe there's a different way to run them?).
- @cyanglaz's TODO to un-skip the test in `example/test_driver/video_player_test.dart` may be resolved (https://github.com/flutter/flutter/issues/43012 was marked fixed in Oct). `flutter drive test_driver/video_player_test.dart` passed once on my machine, but failed with timeouts about 4 times. [The issue](https://github.com/flutter/flutter/issues/43012#issuecomment-548077843) had notes about running these tests on Firebase device lab, so I'm not sure if or where these tests are currently intended to be run.

stuartmorgan: Just a quick note on this question (I haven't looked at the PR in general):

> * Running tests in packages/video_player/video_player/example/integration_test/video_player_test.dart with `flutter test` fails (maybe there's a different way to run them?).

Things in `integration_test` folders are intended to be run via `flutter drive`, using the test driver in `test_driver`; see https://flutter.dev/docs/testing/integration-tests for discussion of the directory structure.

packages/video_player/video_player/example/integration_test/video_player_test.dart
```diff
@@ -129,6 +129,51 @@ void main() {
       },
     );
 
+    testWidgets(
+      'stay paused when seeking after video completed',
```
KyleFin, line 133: The original issue (flutter/flutter#77674)

packages/video_player/video_player/example/integration_test/video_player_test.dart
```diff
@@ -129,6 +129,51 @@ void main() {
...
+        await _controller.seekTo(tenMillisBeforeEnd);
+        await tester.pumpAndSettle(_playDuration);
+
+        expect(_controller.value.isPlaying, false);
+        expect(_controller.value.position, tenMillisBeforeEnd);
+      },
+    );
+
+    testWidgets(
+      'do not exceed duration on play after video completed',
```
KyleFin, line 156: The additional symptom I observed in https://github.com/flutter/flutter/issues/77674#issuecomment-800263277

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -277,6 +277,23 @@ void main() {
       expect(fakeVideoPlayerPlatform.calls.last, 'setPlaybackSpeed');
     });
 
+    test('play restarts from beginning if video is at end', () async {
```
KyleFin, line 280: The new behavior I added to make Android and iOS consistent with web.

packages/video_player/video_player/test/video_player_test.dart
KyleFin: Thanks Stuart! I added 2 integration tests. This PR is ready for review.

KyleFin:
Re-opening this PR. (I accidentally closed it while I was working on renaming git branches.)

packages/video_player/video_player/lib/video_player.dart
stuartmorgan: Could you put a mini version of this explanation in the code here as a comment? The fact that it needed this explanation is good evidence that the code is going to be non-obvious to someone reading it.

stuartmorgan: Looks reasonable.

stuartmorgan (changes requested):
Sorry this was in my queue for so long; I lost track of the notification :(

Overall it looks great, just a couple small nits.
- One inline
- The version in pubspec.yaml and CHANGELOG.md need to be updated to the change will be published

I'm not sure what's going on with the CI failures there; I suspect merging in master will resolve it.

stuartmorgan (approved):
LGTM, sorry again for losing the review!

nogaten:
If I play video when it is not initialized, I will get an error.
I want you to check `value.isInitialized` before seek in play.

stuartmorgan:
> If I play video when it is not initialized, I will get an error.
> I want you to check `value.isInitialized` before seek in play.

Please use the issue tracker, not comments on submitted PRs, to report issues.

PR #3751, opened 2021-03-25
Revert "[in_app_purchase] Fix error message when trying to consume purchase on iOS"

mvanbeusekom:
Reverts flutter/plugins#3750

Reverting this PR since I updated with the wrong version number (should have been `0.5.1+1` instead of `0.5.2`). Will resubmit the PR with the correct version number.


BeMacized (approved):
LGTM, awaiting new PR

PR #3756, opened 2021-03-25
Drop redundant question marks used with dynamic.

aam:
Question marks used with `dynamic` are redundant and are flagged by dart analysis tools as of https://dart-review.googlesource.com/c/sdk/+/190041.

Plugin code need to be analysis-clean.

Fixes https://github.com/flutter/flutter/issues/79089


ditman (approved):
Not bumping the version because at runtime this code makes no difference. Landing to fix the rollers.

PR #3757, opened 2021-03-25
Drop redundant ? for dynamic

GaryQian:
Continuation of https://github.com/flutter/plugins/pull/3756, which missed one question mark.


ditman (approved):

ditman:
Thanks for noticing @GaryQian!

PR #3758, opened 2021-03-26
[android_platform_images] add support for reuse android platform images

vimerzhao:
* add support for reuse android platform images


stuartmorgan:
Adding a new plugin is a major decision, and there are significant questions about how to structure this if we do want to do it (e.g., should we instead rename the existing iOS plugin and make it more generic, since other platforms may need the same functionality in the future? can the APIs be aligned?). Please start with a proposal issue, and potentially a design doc, rather than a PR, so that the proposal can be discussed.

vimerzhao:
> Adding a new plugin is a major decision, and there are significant questions about how to structure this if we do want to do it (e.g., should we instead rename the existing iOS plugin and make it more generic, since other platforms may need the same functionality in the future? can the APIs be aligned?). Please start with a proposal issue, and potentially a design doc, rather than a PR, so that the proposal can be discussed.

got it

PR #3762, opened 2021-03-27
Capitalise Platform class in README

ColonisationCaptain:
When using `dart:io`, the `Platform` class is capitalised in the latest stable Flutter release. This PR fixes this minor issue in the README, where this class was previously uncapitalised. I've briefly looked at the other docs available and I don't think this mistake exists anywhere else.


stuartmorgan (approved):
LGTM, thanks for the contribution!

PR #3763, opened 2021-03-27
[local_auth] local_auth plugin wrongfully reports available methods on android

bharat-biradar:
This PR is for issue - [#45497](https://github.com/flutter/flutter/issues/45497).

### Note for reviewers:
- This PR fixes the issue for android version <= 27.
- For android version > 27 it is more of a work around than actual fix.
- For version > 27 we can determine if a [Biometric_Strong](https://developer.android.com/reference/android/hardware/biometrics/BiometricManager.Authenticators#BIOMETRIC_STRONG) or Biometric.Weak is available for authentication [refer this](https://developer.android.com/reference/android/hardware/biometrics/BiometricManager#canAuthenticate(int))
- According to [stackoverflow answer](https://stackoverflow.com/a/53973970/13171509) the current behaviour is not going to be changed google (I could not find the official issue where it was mentioned)
- For api>27 available _getAvailableBiometrics() could return either Biometric_strong or Biometric_weak as available methods to fix this issue
- If a reviewer gives a go head for above approach I'll make further changes to this PR

## Pre-launch Checklist

- [ ] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I signed the [CLA].


stuartmorgan:

Note that this will need tests before moving forward per Flutter policy. Unfortunately this plugin predates strict enforcement of that, so there aren't currently any Java unit tests, meaning that the file would need to be created and added to the build. But beyond that it seems like it would be fairly straightforward to make a new unit tests for this by extracting this logic to a helper method that takes the biometric manager, and passing in a mock.

bharat-biradar:
@stuartmorgan I would like to implement the unit tests. But can please expand a bit on what you meant `by extracting this logic to a helper method that takes the biometric manager, and passing in a mock` before I start implementing tests so that I have clear picture of what to do. Thanks

stuartmorgan:
Currently the method is getting the `BiometricManager` from the context inline, which means there's no way to unit test the different paths through the conditionals. We need the logic to be able to take a passed instance instead (or do DI for that object at the plugin level), so that it can be given a mock for tests.

bharat-biradar:
Thanks for explaining, I'll implement the test and open a separate PR for it.

bharat-biradar:
Can you please point me to other plugins already having unit tests with this implementation. This is my first time coming accross DI and also writing unit tests I had to spend last 3 days learning about it. Seeing a practical implementation would be great help. If this seems out of my scope please mention I'll wait until the tests are implemented by others and then move forward with this PR. Thank you.

bharat-biradar:
I have written the tests but before opening a PR I wanted to check my implementation as this is my first time coming across DI and also writing java unit test. Can you please point me to any existing plugins that already have similar tests so that can compare them, it will make it easier for reviewers to go through PR. Thanks

stuartmorgan:
Having now looked at the updated Android APIs more, it seems like we should probably, rather than try to reverse-engineer a guess about the available biometrics, add `weak` and `strong` as types to the Dart API. The downside is that anyone checking for specific options (vs it just being non-empty) will need to handled different platforms differently. However, it will a) be more future-proof, and b) doesn't have the potential to give incorrect information as this implementation does (e.g., a device with both touch and face hardware, but only touch enrolled, would incorrectly report that face is an enrolled option with this change; it's an edge case, but it could happen).

bharat-biradar:
Yes due to this problem I was suggesting implementing `weak` and `strong` for API level 28 in dart API. At least the result be will correct even if can't specify exactly which one of the biometrics is enrolled. So should I take this as a yes to make the changes to `weak` and `strong`?

bharat-biradar:
Regarding the tests, IOS tests seem a lot different than my implementation. Maybe I have implemented them wrongly. Is there a deadline before which the tests should be added?

stuartmorgan:
> Yes due to this problem I was suggesting implementing `weak` and `strong` for API level 28 in dart API. At least the result be will correct even if can't specify exactly which one of the biometrics is enrolled. So should I take this as a yes to make the changes to `weak` and `strong`?

Yes, let's proceed with that approach. (I'll discuss with the team to see if we should consider this a breaking change.)

stuartmorgan:
> Regarding the tests, at a glance, IOS tests seem a lot different than my implementation. Maybe I have implemented them wrongly.

There are many ways to solve the problem; just because it doesn't look like that one doesn't mean it's incorrect. If you post what you've implemented I can comment.

> Is there a deadline before which the tests should be added?

I'm not sure what you're asking exactly. The PR needs a test before it can land, if that's what you mean.

bharat-biradar:
Sorry for replying this late. By post did you mean opening a PR or a gist?

hectorvillag:
any update on this issue? I have this problem and need a solution, but I think this PR is not close to merging.
I understand that is needed to do unit tests in java and better implementation to correct the behavior, right?

mxa0079:
> Having now looked at the updated Android APIs more, it seems like we should probably, rather than try to reverse-engineer a guess about the available biometrics, add `weak` and `strong` as types to the Dart API. The downside is that anyone checking for specific options (vs it just being non-empty) will need to handled different platforms differently. However, it will a) be more future-proof, and b) doesn't have the potential to give incorrect information as this implementation does (e.g., a device with both touch and face hardware, but only touch enrolled, would incorrectly report that face is an enrolled option with this change; it's an edge case, but it could happen).

We are constantly running into the exact  edge case described by @stuartmorgan - our app is used by 1M+ MAU and at this volume it should not be considered an edge case -- it happens constantly.

stuartmorgan:
Per the discussion above, the PR as written isn't the approach we want to move forward with for fixing this; we should instead be changing the plugin API slightly to better accommodate the fundamental change in the way Android handles auth. Given that I'm going to close this, but if you're interested in working on that approach please feel free to open a new PR!

PR #3768, opened 2021-03-30
[google_maps_flutter] Fix NNBD migration mistake in example

stuartmorgan:
The TileOverlay portion of the example app incorrectly constructed a
Set<TileOverlay?> and then attempted to cast it to a Set<TileOverlay>,
rather than making a set of non-nullable types as required by the API.

Fixes https://github.com/flutter/flutter/issues/79242

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
+++ b/packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
@@ -1,7 +1,7 @@
 name: google_maps_flutter
 description: A Flutter plugin for integrating Google Maps in iOS and Android applications.
 homepage: https://github.com/flutter/plugins/tree/master/packages/google_maps_flutter/google_maps_flutter
-version: 2.0.2
+version: 2.0.2-dev
 
 dependencies:
   flutter:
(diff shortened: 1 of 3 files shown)


stuartmorgan:
No test for this change since it looks like this integration test is one of the ones that uses the APIs directly rather than driving the example app.

packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: google_maps_flutter
 description: A Flutter plugin for integrating Google Maps in iOS and Android applications.
 homepage: https://github.com/flutter/plugins/tree/master/packages/google_maps_flutter/google_maps_flutter
-version: 2.0.1
+version: 2.0.2
```
cyanglaz, line 4: Probably not necessary? Updating the version has no harm either.

cyanglaz (approved):
LGTM

packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
stuartmorgan: I thought more of the example was visible on pub.dev than it is.

I think it's worth mentioning in CHANGELOG, so I'll leave it. (Something I think we should consider is a way to add CHANGELOG lines for cases where it's not worth actually publishing yet. E.g., allow for "<NEXT>" as a version as long as the version in pubspec hasn't changed.)

ditman: Dart people do `-dev` releases for these cases, until they need to publish for real, and then they collapse everything into a next version with multiple entries.

ditman (approved):

PR #3776, opened 2021-03-31
[tool] add `all` and `dry-run` flags to publish-plugin command

cyanglaz:
Add a new `all` flag  to the `publish plugin` command does the following:
1. Look for all the pubspec changes in the current commit and determine what plugin(s) need to be released.  If the new version is lower than the existing latest tagged version, skip. 
1. Release the plugin as how `publish-plugin` works: publishing to pub, git tag release etc. 

Add a new `dry-run` flag to dry-run the publish and tag release process. This is for testing locally and ci without actually publishing or git tag anything.

This command is going to be part of auto-publish workflow.

part of https://github.com/flutter/flutter/issues/79830

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/publish_plugin_command.dart
+++ b/script/tool/lib/src/publish_plugin_command.dart
@@ -69,7 +69,7 @@ class PublishPluginCommand extends PluginCommand {
       defaultsTo: 'upstream',
     );
     argParser.addFlag(
-      _allFlag,
+      _allChangedFlag,
       help:
           'Release all plugins that contains pubspec changes at the current commit compares to the base-sha.\n'
           'The $_packageOption option is ignored if this is on.',
@@ -91,7 +91,7 @@ class PublishPluginCommand extends PluginCommand {
   static const String _pushTagsOption = 'push-tags';
   static const String _pubFlagsOption = 'pub-publish-flags';
   static const String _remoteOption = 'remote';
-  static const String _allFlag = 'all';
+  static const String _allChangedFlag = 'all-changed';
   static const String _dryRunFlag = 'dry-run';
 
   // Version tags should follow <package-name>-v<semantic-version>. For example,
@@ -108,13 +108,12 @@ class PublishPluginCommand extends PluginCommand {
   final Print _print;
   final io.Stdin _stdin;
   StreamSubscription<String> _stdinSubscription;
-  bool _startedListenToStdStream = false;
 
   @override
   Future<void> run() async {
     final String package = argResults[_packageOption] as String;
-    final bool all = argResults[_allFlag] as bool;
-    if (package == null && !all) {
+    final bool publishAllChanged = argResults[_allChangedFlag] as bool;
+    if (package == null && !publishAllChanged) {
       _print(
           'Must specify a package to publish. See `plugin_tools help publish-plugin`.');
       throw ToolExit(1);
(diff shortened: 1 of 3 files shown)


cyanglaz:
As we discussed in the doc, we are probably going to use the alternative if LUCI push to git works.
If that's the case, the command should include both publish and git tag. I will convert this to draft until implement the publish part.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -84,14 +106,14 @@ class PublishPluginCommand extends PluginCommand {
...
-  final Stdin _stdin;
-  // The directory of the actual package that we are publishing.
+  final io.Stdin _stdin;
   StreamSubscription<String> _stdinSubscription;
 
   @override
   Future<void> run() async {
     final String package = argResults[_packageOption] as String;
-    if (package == null) {
+    final bool all = argResults[_allFlag] as bool;
```
stuartmorgan, line 115: Nit: how about `publishAll` as the variable name so it's more descriptive.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -64,13 +68,31 @@ class PublishPluginCommand extends PluginCommand {
       // Flutter convention is to use "upstream" for the single source of truth, and "origin" for personal forks.
       defaultsTo: 'upstream',
     );
+    argParser.addFlag(
+      _allFlag,
```
stuartmorgan, line 72: Maybe we should call this flag `all-changed`? `--all` sounds more like it should do something like the Melos publish-all command.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -111,7 +135,74 @@ class PublishPluginCommand extends PluginCommand {
     }
     _print('Local repo is ready!');
 
-    final Directory packageDir = _getPackageDir(package);
+    if (all) {
+      await _publishAllPackages(
```
stuartmorgan, line 139: Similarly, `_publishAllChangedPackages`

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -111,7 +135,74 @@ class PublishPluginCommand extends PluginCommand {
...
+      {String remote,
+      String remoteUrl,
+      bool shouldPushTag,
+      GitDir baseGitDir}) async {
+    final List<String> packagesReleased = <String>[];
+    final GitVersionFinder gitVersionFinder = await retrieveVersionFinder();
+    final List<String> changedPubspecs =
+        await gitVersionFinder.getChangedPubSpecs();
+    if (changedPubspecs.isEmpty) {
+      _print('No version updates in this commit, exiting...');
```
stuartmorgan, line 164: "..." indicates more to follow, which seems odd for something that says it's exiting.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -111,7 +135,74 @@ class PublishPluginCommand extends PluginCommand {
...
+      return;
+    }
+    _print('Getting existing tags...');
+    final io.ProcessResult existingTagsResult =
+        await baseGitDir.runCommand(<String>['tag', '--sort=-committerdate']);
+    final List<String> existingTags = (existingTagsResult.stdout as String)
+        .split('\n')
+          ..removeWhere((element) => element == '');
```
stuartmorgan, line 172: `isEmpty`

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -111,7 +135,74 @@ class PublishPluginCommand extends PluginCommand {
...
+    for (final String pubspecPath in changedPubspecs) {
+      final File pubspecFile =
+          fileSystem.directory(baseGitDir.path).childFile(pubspecPath);
+      if (!pubspecFile.existsSync()) {
+        printErrorAndExit(
```
stuartmorgan, line 177: Why is this a fatal error? Isn't this expected if we delete (or move) a plugin?

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -111,7 +135,74 @@ class PublishPluginCommand extends PluginCommand {
...
+          remote: remote,
+          remoteUrl: remoteUrl,
+          shouldPushTag: shouldPushTag);
+      packagesReleased.add(pubspecFile.parent.basename);
+      _print('\n');
+    }
+    _print('Packages released: ${packagesReleased.join(', ')}');
+  }
+
+  Future<void> _publishAndReleasePackage(
```
stuartmorgan, line 201: `_publishAndTagPackage`? I would consider publishing to be releasing.

Wouldn't hurt to add a comment to this method saying what it does.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -120,7 +211,50 @@ class PublishPluginCommand extends PluginCommand {
           remoteUrl: remoteUrl,
           shouldPushTag: shouldPushTag);
     }
-    await _finishSuccesfully();
+    _print('Release ${packageDir.basename} successful.');
```
stuartmorgan, line 214: 'Released [...] successfully.'

stuartmorgan, line 214: There aren't any bools for the sub-steps; does that mean they are fatal errors when something goes wrong?

If we try to release 3 packages, an error with one of them presumably shouldn't kill the whole thing, but instead move on to the next, and then at the end list all the successes and failures.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -120,7 +211,50 @@ class PublishPluginCommand extends PluginCommand {
...
+  }
+
+  // Returns `true` if needs to release the version, `false` if needs to skip
+  Future<bool> _checkNeedsRelease({
+    @required String pubspecPath,
+    @required File pubspecFile,
```
stuartmorgan, line 220: Why are the path and the file separate arguments, rather than getting one from the other internally?

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -120,7 +211,50 @@ class PublishPluginCommand extends PluginCommand {
...
+        await gitVersionFinder.getPackageVersion(pubspecPath, gitRef: 'HEAD');
+    if (headVersion == null) {
+      printErrorAndExit(
+          errorMessage: 'No version found. A package that '
+              'intentionally has no version should be marked '
+              '"publish_to: none".');
+    }
+
+    if (pubspec.name == null) {
+      printErrorAndExit(errorMessage: 'Fatal: Package name is null.');
```
stuartmorgan, line 239: Same concerns here; if this is fatal, then all packages will fail, not just the one that's wrong.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -120,7 +211,50 @@ class PublishPluginCommand extends PluginCommand {
...
+    }
+    // Get latest tagged version and compare with the current version.
+    // TODO(cyanglaz): Check latest version of the package on pub instead of git
+    // https://github.com/flutter/flutter/issues/81047
+    final String latestTag = existingTags.isNotEmpty
+        ? existingTags
+            .firstWhere((String tag) => tag.split('-v').first == pubspec.name)
+        : '';
```
stuartmorgan, line 247: You can use `orElse` here instead of checking `isNotEmpty`.

In fact, I think without that change the current logic is wrong, since just because the set of tags isn't empty doesn't mean that this particular package has been released before, so it seems like this would throw as soon as we had a new plugin. (Which suggests there's a test case missing.)

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -135,6 +269,14 @@ class PublishPluginCommand extends PluginCommand {
...
       @required bool shouldPushTag}) async {
     final String tag = _getTag(packageDir);
+    if (argResults[_dryRunFlag] as bool) {
+      _print('DRY RUN: Tagging release $tag...');
+      if (!shouldPushTag) {
+        return;
+      }
+      _print('DRY RUN: Pushing tag to $remote...');
+      return;
+    }
```
stuartmorgan, line 279: Duplicating all the logic and log messages for all dry run commands will be a maintenance mess. It also means a dry run is only testing the special dry-run logic, not the actual logic that will run, which is a landmine waiting to happen.

I'd much rather we have a big "===== DRY RUN =====" (or similar) banner at the start of the output, and have the dry-run short-circuit individual commands, but otherwise run through the normal logic.

stuartmorgan (changes requested):

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: done

cyanglaz (commented):
Updated per review comments. PTAL @stuartmorgan

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -110,45 +134,219 @@ class PublishPluginCommand extends PluginCommand {
...
+          shouldPushTag: shouldPushTag);
+    }
+    await _finish(successful);
+  }
+
+  Future<bool> _publishAllChangedPackages(
+      {String remote,
+      String remoteUrl,
+      bool shouldPushTag,
+      GitDir baseGitDir}) async {
```
stuartmorgan, line 162: Optional nit: add trailing commas here and below to improve formatting

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -110,45 +134,219 @@ class PublishPluginCommand extends PluginCommand {
...
+      _print('Packages released: ${packagesReleased.join(', ')}');
+    }
+    if (packagesFailed.isNotEmpty) {
+      _print(
+          'Packages failed to release: ${packagesFailed.join(', ')}, see above for details.');
+    }
+    if (packagesFailed.isNotEmpty) {
+      return false;
+    }
+    return true;
```
stuartmorgan, line 218: These four lines can be simplified to: `return packagesFailed.isEmpty;`

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -110,45 +134,219 @@ class PublishPluginCommand extends PluginCommand {
...
+          fileSystem.directory(baseGitDir.path).childFile(pubspecPath);
+      try {
+        final bool needsRelease = await _checkNeedsRelease(
+            pubspecFile: pubspecFile,
+            gitVersionFinder: gitVersionFinder,
+            existingTags: existingTags);
+        if (!needsRelease) {
+          continue;
+        }
+      } on ToolExit catch (e) {
```
stuartmorgan, line 191: Catching a `ToolExit` seems potentially very confusing. How about making _checkNeedsRelease return a release/norelease/failure enum?

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -162,56 +360,67 @@ class PublishPluginCommand extends PluginCommand {
...
+          'git',
+          <String>[
+            'status',
+            '--porcelain',
+            '--ignored',
+            packageDir.absolute.path
+          ],
+          workingDir: packageDir);
+    } on ToolExit catch (e) {
+      print(e);
```
stuartmorgan, line 376: Why `print` and `_print`?

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -162,56 +360,67 @@ class PublishPluginCommand extends PluginCommand {
...
-          'status',
-          '--porcelain',
-          '--ignored',
-          packageDir.absolute.path
-        ],
-        workingDir: packageDir);
+  Future<bool> _checkGitStatus(Directory packageDir) async {
+    io.ProcessResult statusResult;
+    try {
+      statusResult = await processRunner.runAndExitOnError(
```
stuartmorgan, line 366: The naming here suggests there's a version that doesn't exit on error, so we wouldn't need to catch ToolExits?

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -237,9 +449,19 @@ class PublishPluginCommand extends PluginCommand {
...
-    await processRunner.runAndExitOnError('git', <String>['push', remote, tag],
-        workingDir: packagesDir);
+    try {
+      if (!(argResults[_dryRunFlag] as bool)) {
+        await processRunner.runAndExitOnError(
+            'git', <String>['push', remote, tag],
+            workingDir: packagesDir);
+      }
+    } on ToolExit catch (e) {
+      print(e);
```
stuartmorgan, line 461: Same question as above.

stuartmorgan (changes requested):

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: done

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: Yup, sounds good!

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: done

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: oops, this was a testing code, removed

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: `runAndExitOnError` does something extra (printing error message in a specific format etc), if we don't use it, then we have to re-write that code. I think we should rename this method to `run(bool logError = true, bool exitOnError = true...)`. 

Actually, I just saw that the `run` method that we have already has a `exitOnError` flag which doesn't do anything, which suggests that we used to have it like this but somehow refactored into 2 methods. https://github.com/flutter/plugins/blob/master/script/tool/lib/src/common.dart#L507

I'm happy to do the refactoring in a separate PR! Then rebase this on top after that lands.

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: same answer above :)

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: https://github.com/flutter/plugins/pull/3827 @stuartmorgan here you go

cyanglaz (commented):
@stuartmorgan Fixed all review comments as well as removing all the unnecessary try-catch. PTAL

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -110,50 +134,231 @@ class PublishPluginCommand extends PluginCommand {
...
     }
-    await _finishSuccesfully();
+    _print('Released [${packageDir.basename}] successfully.');
+    return true;
   }
 
-  Future<void> _publishPlugin({@required Directory packageDir}) async {
-    await _checkGitStatus(packageDir);
-    await _publish(packageDir);
+  // Returns `true` if needs to release the version, `false` if needs to skip
```
stuartmorgan, line 250: Fix comment for new return type.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -110,50 +134,231 @@ class PublishPluginCommand extends PluginCommand {
...
+        packagesFailed.add(pubspecFile.parent.basename);
+      }
+      _print('\n');
+    }
+    if (packagesReleased.isNotEmpty) {
+      _print('Packages released: ${packagesReleased.join(', ')}');
+    }
+    if (packagesFailed.isNotEmpty) {
+      _print(
+          'Packages failed to release: ${packagesFailed.join(', ')}, see above for details.');
```
stuartmorgan, line 218: Nit: "Failed to release the following packages:"

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -110,50 +134,231 @@ class PublishPluginCommand extends PluginCommand {
...
+
+    final Pubspec pubspec = Pubspec.parse(pubspecFile.readAsStringSync());
+    if (pubspec.publishTo == 'none') {
+      return _CheckNeedsReleaseResult.noRelease;
+    }
+
+    if (pubspec.version == null) {
+      _print('No version found. A package that '
+          'intentionally has no version should be marked '
+          '"publish_to: none".');
```
stuartmorgan, line 272: Nit: Fix the excessive line breaking here?

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -235,23 +446,44 @@ class PublishPluginCommand extends PluginCommand {
...
 }
+
+enum _CheckNeedsReleaseResult {
+  // The package needs to be released.
+  release,
+
+  // The package does not need to be released.
+  noRelease,
+
+  // There's an error when trying to access if the package needs to be released.
```
stuartmorgan, line 487: Nit: s/access if/determine whether/

script/tool/test/util.dart
```diff
@@ -85,14 +87,14 @@ Directory createFakePlugin(
     final Directory exampleDir = pluginDirectory.childDirectory('example')
       ..createSync();
     createFakePubspec(exampleDir,
-        name: '${name}_example', isFlutter: isFlutter);
+        name: '${name}_example', isFlutter: isFlutter, includeVersion: false, publish_to: 'none');
```
stuartmorgan, line 90: While you're touching the utils, could you fix the naming of this parameter? It should be `publishTo`.

stuartmorgan (approved):
LGTM with nits.

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: done

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: done

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: done

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: done

script/tool/test/util.dart
cyanglaz: done

PR #3778, opened 2021-04-01
[local_auth] Fix callback thread handling

stuartmorgan:
Ensure that all auth replies, which are sent on an internal framework queue per documentation, are dispatched back to the main thread for handling, as all resulting operations (method channel callbacks, display of UI) are things that must be done on the main thread

In order to test this, sets up local_auth with XCTest-based tests, and adds the ability to inject a mock `LAContext`. (This does not do full unit test backfill, to limit the scope of the PR.)

Fixes https://github.com/flutter/flutter/issues/47465

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Before landing this I need to make sure CI is actually running the new unit tests, since I'm not sure if there were extra steps to wire that up that I missed, or if it's automatic.

packages/local_auth/ios/Classes/FLTLocalAuthPlugin.m
```diff
@@ -40,6 +42,13 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
 
 #pragma mark Private Methods
 
+- (LAContext *)authContext {
+  if (_authContextOverride != nil) {
+    return _authContextOverride;
+  }
+  return [[LAContext alloc] init];
```
cyanglaz, line 49: Each `self.authContext` is going to call the constructor (when no  _authContextOverride is provided)

If we always wants a new instance (which i think that's the case), we should rename it do something doesn't sound too much like a getter`. Maybe:
`- (LAContext *)getNewAuthContext`?

packages/local_auth/ios/Tests/FLTLocalAuthPluginTests.m
```diff
@@ -0,0 +1,182 @@
...
+                                                              arguments:@{
+                                                                @"biometricOnly" : @(YES),
+                                                                @"localizedReason" : reason,
+                                                              }];
+
+  XCTestExpectation* expectation = [self expectationWithDescription:@"Result is called"];
+  [plugin handleMethodCall:call
+                    result:^(id _Nullable result) {
+                      XCTAssertTrue([NSThread isMainThread]);
+                      [expectation fulfill];
```
cyanglaz, line 58: Shouldn't this fulfill be the last call of the block? After `XCTAssertTrue([result boolValue]);`? Unless I'm missing something? Ditto below if make sense.

packages/local_auth/ios/Tests/FLTLocalAuthPluginTests.m
```diff
@@ -0,0 +1,182 @@
...
+                      XCTAssertTrue([result isKindOfClass:[NSNumber class]]);
+                      XCTAssertTrue([result boolValue]);
+                    }];
+  [self waitForExpectationsWithTimeout:1.0 handler:nil];
```
cyanglaz, line 62: We should increase the timeout as running tests on CI might be slow and we don't want to many false negatives (flake) because of the environment is slow. How about 30s?  and set `continueAfterFailure` to NO in `setup`

cyanglaz (changes requested):
It seems there are some issues with the test code. I might be missing something tho.

packages/local_auth/ios/Classes/FLTLocalAuthPlugin.m
stuartmorgan: Renamed to `createAuthContext` since `create` is a standard idiom. Also future-proofed the override by making it better fit the fact that this is really, as you point out, a factory rather than a getter; it now uses an array of overrides, so a test could run a flow with multiple API calls.

packages/local_auth/ios/Tests/FLTLocalAuthPluginTests.m
stuartmorgan: Good call; done. It's been a long time since I've done async XCTests!

packages/local_auth/ios/Tests/FLTLocalAuthPluginTests.m
stuartmorgan: Done.

cyanglaz (approved):
LGTM

PR #3779, opened 2021-04-01
[tool] refactor publish plugin command

cyanglaz:
Refactoring publish plugin command to get ready for a new feature to publish and tag release based on git diff.

See https://github.com/flutter/plugins/pull/3776#issuecomment-811574215

behavioral change: The command now validate the git remote first before checking package directory. Because when we change this command later to run multiple publishes, the git remote validation only needs to happen once.

### DIFF
--- a/script/tool/lib/src/publish_plugin_command.dart
+++ b/script/tool/lib/src/publish_plugin_command.dart
@@ -120,7 +120,7 @@ class PublishPluginCommand extends PluginCommand {
     _print('Local repo is ready!');
 
     final Directory packageDir = _checkPackageDir(package);
-    await _publishPlugin(packageDir: packagesDir);
+    await _publishPlugin(packageDir: packageDir);
     await _tagRelease(packageDir: packageDir, remote: remote, remoteUrl: remoteUrl, shouldPushTag: shouldPushTag);
     await _finishSuccesfully();
   }
--- a/script/tool/lib/src/publish_plugin_command.dart
+++ b/script/tool/lib/src/publish_plugin_command.dart
@@ -105,6 +105,7 @@ class PublishPluginCommand extends PluginCommand {
       throw ToolExit(1);
     }
 
+    _print('Checking local repo...');
     if (!await GitDir.isGitDir(packagesDir.path)) {
       _print('$packagesDir is not a valid Git repository.');
       throw ToolExit(1);
@@ -116,6 +117,7 @@ class PublishPluginCommand extends PluginCommand {
     if (shouldPushTag) {
       remoteUrl = await _verifyRemote(remote);
     }
+    _print('Local repo is ready!');
 
     final Directory packageDir = _checkPackageDir(package);
     await _publishPlugin(packageDir: packagesDir);
@@ -124,9 +126,7 @@ class PublishPluginCommand extends PluginCommand {
   }
 
   Future<void> _publishPlugin({@required Directory packageDir}) async {
-    _print('Checking local repo...');
     await _checkGitStatus(packageDir);
-    _print('Local repo is ready!');
     await _publish(packageDir);
     _print('Package published!');
   }
--- a/script/tool/lib/src/publish_plugin_command.dart
+++ b/script/tool/lib/src/publish_plugin_command.dart
@@ -63,6 +63,12 @@ class PublishPluginCommand extends PluginCommand {
       // Flutter convention is to use "upstream" for the single source of truth, and "origin" for personal forks.
       defaultsTo: 'upstream',
     );
+    argParser.addOption(
+      _allOption,
+      help:
+          'Based the diff between $_kBaseShaOption and HEAD, publish all the plugins that has pubspec change in this diff.\n'
+          'The --package option is ignored if this flag is on. (This is currently under construction)' ,
+    );
   }
 
   static const String _packageOption = 'package';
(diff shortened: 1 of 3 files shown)


script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -63,13 +63,21 @@ class PublishPluginCommand extends PluginCommand {
       // Flutter convention is to use "upstream" for the single source of truth, and "origin" for personal forks.
       defaultsTo: 'upstream',
     );
+    argParser.addOption(
+      _allOption,
+      help:
+          'Based the diff between $_kBaseShaOption and HEAD, publish all the plugins that has pubspec change in this diff.\n'
```
stuartmorgan, line 69: Nit: "have pubspec changes"

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -63,13 +63,21 @@ class PublishPluginCommand extends PluginCommand {
...
+          'The --package option is ignored if this flag is on. (This is currently under construction)' ,
```
stuartmorgan, line 70: What does "under construction" mean here? Is the plan for this behavior to change?

script/tool/lib/src/publish_plugin_command.dart
stuartmorgan, line 69: Do we need to handle the case where there have been multiple releases of a single plugin in the range, or are we going to avoid that issue in automation by never doing a range spanning multiple changes?

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -85,15 +93,24 @@ class PublishPluginCommand extends PluginCommand {
...
   final Stdin _stdin;
   // The directory of the actual package that we are publishing.
-  Directory _packageDir;
   StreamSubscription<String> _stdinSubscription;
 
   @override
   Future<Null> run() async {
     checkSharding();
+    final String package = argResults[_packageOption];
+    if (package == null) {
```
stuartmorgan, line 102: The help said --package is ignored if --all is provided, but this is bailing early regardless.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -102,60 +119,55 @@ class PublishPluginCommand extends PluginCommand {
...
+  }
+
+  Future<void> _publishPlugin({@required Directory packageDir}) async {
+    await _checkGitStatus(packageDir);
+    await _publish(packageDir);
     _print('Package published!');
+  }
+
+  Future<void> _tagRelease({@required Directory packageDir, @required String remote, @required String remoteUrl, @required bool shouldPushTag}) async {
     if (!argResults[_tagReleaseOption]) {
```
stuartmorgan, line 135: I think it would make more sense to wrap the call to _tagRelease in this check, rather than making the function silently no-op based on an argument.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -102,60 +119,55 @@ class PublishPluginCommand extends PluginCommand {
...
-      return await _finishSuccesfully();
+      return;
     }
-
     _print('Tagging release...');
-    final String tag = _getTag();
-    await processRunner.runAndExitOnError('git', <String>['tag', tag],
-        workingDir: _packageDir);
+    final String tag = _getTag(packageDir);
```
stuartmorgan, line 139: Let's flip this and the line above, so the print can include the tag for ease of debugging issues.

script/tool/lib/src/publish_plugin_command.dart
```diff
@@ -102,60 +119,55 @@ class PublishPluginCommand extends PluginCommand {
...
     await _stdinSubscription.cancel();
     _print('Done!');
   }
 
-  Directory _checkPackageDir() {
-    final String package = argResults[_packageOption];
-    if (package == null) {
-      _print(
-          'Must specify a package to publish. See `plugin_tools help publish-plugin`.');
+  Directory _checkPackageDir(String package) {
```
stuartmorgan, line 154: Can we rename this to something like _getPackageDir (and document that it throws if the dir doesn't exist)? I would expect a `check` function to either return a bool, or return nothing.

script/tool/lib/src/publish_plugin_command.dart
stuartmorgan: Returning after finishing the review... why is this being added here? Nothing in the PR uses it; it's unimplemented and untested.

stuartmorgan (changes requested):
Structure looks good, just small suggestions

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: The issue won't exist when we start the automate process, because every commit that contains a pubspec change is going to run a publish command. 

There might be leftovers before the automation so we might need to handle that, handle it manually before automation doesn't seem to be too bad.

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: It meant it is not supported yet. I had everything wired up in a different branch but i thought to introduce this pr as an intermediate step is easier to review. 

 I should just remove the flag int this PR as well and add this flag when actually adding this feature.

script/tool/lib/src/publish_plugin_command.dart
cyanglaz: Sorry to cause confusion. This PR doesn't handle --all yet. And this PR still tries to keep the original functionality of the command. 

Because I wanted to move `_checkPackageDir` after the `_verifyRemote` check, so I moved this check out of `_checkPackageDir` to keep the validation step consistent. 

The test is covered by `requires a package flag`

cyanglaz (commented):
@stuartmorgan Updated per comments. PTAL

stuartmorgan (approved):
LGTM

PR #3780, opened 2021-04-02
[local_auth] Fix iOS crash when no localizedReason

enricobenedos:
If you fill `localizedReason` with an empty string the app crash without any exception or log on iOS simulator. I've added an assertion and a related test in order to notify developer that the field cannot be null but also cannot be an empty string.

This PR should fix [#77794](https://github.com/flutter/flutter/issues/77794). 

## Pre-launch Checklist

- [x] I signed the [CLA].


enricobenedos:
@googlebot I fixed it.

packages/local_auth/lib/local_auth.dart
```diff
@@ -101,6 +101,8 @@ class LocalAuthentication {
     bool biometricOnly = false,
   }) async {
     assert(localizedReason != null);
```
cyanglaz, line 103: Since this plugin has migrated to null safety, this line is actually not necessary anymore. Feel free to remove it while you are here.

packages/local_auth/lib/local_auth.dart
```diff
@@ -101,6 +101,8 @@ class LocalAuthentication {
...
+    assert(localizedReason.isNotEmpty);
```
cyanglaz, line 104: What's the behavior on android if this is empty?

enricobenedos: On Android 11 (simulator) there are no problems if the `localizedReason` is empty. Is it a good idea to add the assert only on iOS case using `Platform.isIOS`?

cyanglaz: > Is it a good idea to add the assert only on iOS case using Platform.isIOS? 

Yes. If android can handle null, especially if null and empty string means different things on android, we should allow null on android.  Do you know if there are difference between null and empty string on android?

cyanglaz: Looking at this again. I think you are right, we really shouldn't allow empty strings or null on android either. So the change looks good.

cyanglaz: @enricobenedos Let's make sure to mention the `localizedReason` must not be an empty String in the documentation.

enricobenedos: Hello @cyanglaz, sorry for the delay but I'm a little bit busy. I'm happy that you check that is it a good idea to also not give the possibility to input an empty string on Android. I will update the docs soon.

packages/local_auth/lib/local_auth.dart
enricobenedos: 👍🏼

packages/local_auth/lib/local_auth.dart
cyanglaz: Awesome! Thanks, once you update the doc we will land this!

enricobenedos: I've updated the `authenticate` method documentation.

packages/local_auth/lib/local_auth.dart
enricobenedos: Done

cyanglaz (commented):
Almost ready! Just some nits :)

enricobenedos:
Thank you @cyanglaz, I've applied your suggestions 😃

cyanglaz (approved):
Awesome! LGTM

PR #3792, opened 2021-04-07
Fix and update version checks

stuartmorgan:
Currently our version update checks aren't actually working; the script doesn't work correctly if no explicit --base-sha is passed, but that's always how CI is calling it.

Fixes https://github.com/flutter/flutter/issues/79823 (and version checks in general)

This makes a number of changes:
- Fixes it to work without --base-sha
- Adds tests that it works in that mode
  - And tightens existing tests to require ToolExit, not just any error, to reduce false-positive test success
- Adds verbose logging of the checks being done, to make it easier to debug this kind of issue in the future
- Tightens the exception handling for missing previous versions to just the line that's expected to fail in that case
- Only allows missing versions when "publish_to: none" is set
  - Adds that everywhere it's missing
  - Standardize the format in the repo to "none" (instead of also having "'none'").
- Allows the use of NEXT in CHANGELOG as a way of gathering changes that are worth noting, but not
  doing a publish cycle for. (Replaces the plan of using -dev versions, since that's actually harder to implement,
  and more confusing.)
  - Ensures that we don't forget to clean up NEXT entries when bumping versions

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/path_provider/path_provider_macos/example/pubspec.yaml
+++ b/packages/path_provider/path_provider_macos/example/pubspec.yaml
@@ -2,7 +2,7 @@ name: path_provider_example
 description: Demonstrates how to use the path_provider plugin.
 publish_to: none
 
-dependencie:
+dependencies:
   flutter:
     sdk: flutter
   path_provider_macos:
--- a/packages/path_provider/path_provider_macos/example/pubspec.yaml
+++ b/packages/path_provider/path_provider_macos/example/pubspec.yaml
@@ -1,7 +1,8 @@
 name: path_provider_example
 description: Demonstrates how to use the path_provider plugin.
+publish_to: none
 
-dependencies:
+dependencie:
   flutter:
     sdk: flutter
   path_provider_macos:
--- a/script/tool/lib/src/version_check_command.dart
+++ b/script/tool/lib/src/version_check_command.dart
@@ -13,8 +13,6 @@ import 'package:pubspec_parse/pubspec_parse.dart';
 
 import 'common.dart';
 
-const String _kBaseSha = 'base-sha';
-
 /// Categories of version change types.
 enum NextVersionType {
   /// A breaking change.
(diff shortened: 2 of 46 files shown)


stuartmorgan:
I imagine that now that this works, we'll run into things we need to relax or tweak over time, but getting it actually working is a good first step to getting it working the way we want :)

script/tool/lib/src/version_check_command.dart
```diff
@@ -175,6 +184,14 @@ class VersionCheckCommand extends PluginCommand {
     }
     // Remove all leading mark down syntax from the version line.
     final String versionString = firstLineWithText.split(' ').last;
+
+    // Skip validation for the special NEXT version that's used to accumulate
+    // changes that don't warrant publishing on their own.
+    if (versionString == 'NEXT') {
```
cyanglaz, line 190: We also need to make sure that the version in pubspec equals to the next version after `NEXT` in CHANGELOG. 

This is to prevent someone updated the version in pubspec but forgot to change the "NEXT"

cyanglaz: If we only cares about presubmit, which looks like so in the current code (base-sha not configurable), we can easily do
```
if (versionString == 'NEXT' && pubspecVersion not changed) 
```

script/tool/lib/src/version_check_command.dart
```diff
@@ -96,48 +94,59 @@ class VersionCheckCommand extends PluginCommand {
     final List<String> changedPubspecs =
         await gitVersionFinder.getChangedPubSpecs();
 
-    final String baseSha = argResults[_kBaseSha] as String;
```
cyanglaz, line 99: I still think we should allow explicit base-sha so this can be used in post-submit where we can set base-sha to be HEAD~

cyanglaz (changes requested):

script/tool/lib/src/version_check_command.dart
```diff
@@ -96,48 +94,59 @@ class VersionCheckCommand extends PluginCommand {
...
     for (final String pubspecPath in changedPubspecs) {
+      print('Checking versions for $pubspecPath...');
+      final File pubspecFile = fileSystem.file(pubspecPath);
+      if (!pubspecFile.existsSync()) {
+        print('  Deleted; skipping.');
+        continue;
+      }
+      final Pubspec pubspec = Pubspec.parse(pubspecFile.readAsStringSync());
+      if (pubspec.publishTo == 'none') {
+        print('  Found "publish_to: none"; skipping.');
```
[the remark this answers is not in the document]
stuartmorgan: Yes, this is instead of doing the `----------` between each section like the CHANGELOG check; it's more compact, and I think it's more readable.

script/tool/lib/src/version_check_command.dart
stuartmorgan: We do; all the support for that flag is in the common.dart code. See the addition of _getBaseSha when null is passed, which uses the existing (working) logic to either use the flag, or the merge base when not passed.

Most of the tests use the flag, and still pass.

stuartmorgan: (Also, we should definitely fix the post-submit invocation since currently I believe it does nothing. I'll do that in a follow-up. And longer term we should maybe reconsider how this check works, since it ignores the usual packages/all system.)

script/tool/lib/src/version_check_command.dart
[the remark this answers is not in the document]
stuartmorgan: Updated to use an `indentation` variable to make it clearer and ensure consistency.

script/tool/lib/src/version_check_command.dart
```diff
@@ -96,48 +94,59 @@ class VersionCheckCommand extends PluginCommand {
...
-                NextVersionType.BREAKING_MAJOR) {
-          final String error = '$pubspecPath breaking change detected.\n'
-              'Breaking changes to platform interfaces are strongly discouraged.\n';
-          printErrorAndExit(errorMessage: error);
-        }
+        masterVersion = await gitVersionFinder.getPackageVersion(pubspecPath);
       } on io.ProcessException {
-        print('Unable to find pubspec in master for $pubspecPath.'
-            ' Safe to ignore if the project is new.');
+        print('  Unable to find pubspec in master. '
```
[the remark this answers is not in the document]
stuartmorgan: Yes, that makes way more sense; the type of exception here is entirely dependent on the implementation details of that method, which is odd. Fixed.

script/tool/lib/src/version_check_command.dart
stuartmorgan: Good call, added validation for that and added a test.

script/tool/lib/src/version_check_command.dart
cyanglaz: ah i actually developed that code not even long time ago lol my memory is getting really bad these days due to lack of sleep. :)

script/tool/lib/src/version_check_command.dart
```diff
@@ -169,12 +178,29 @@ class VersionCheckCommand extends PluginCommand {
...
+    // Skip validation for the special NEXT version that's used to accumulate
+    // changes that don't warrant publishing on their own.
+    bool hasNextSection = versionString == 'NEXT';
+    if (hasNextSection) {
+      print('Found NEXT; validating next version in the CHANGELOG.');
+      // Ensure that the version in pubspec hasn't changed without updating
+      // CHANGELOG. That means the next version entry in the CHANGELOG pass the
+      // normal validation.
+      while (iterator.moveNext()) {
+        if (iterator.current.trim().startsWith('## ')) {
```
cyanglaz, line 197: some of our CHANGELOG.md files do no follow the standard format, see: https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher_web/CHANGELOG.md

This would fail. :(

cyanglaz (changes requested):

script/tool/lib/src/version_check_command.dart
stuartmorgan: I'm fine with a model where trying to use NEXT won't work until the format has been fixed to follow our standard format.

We can also do a one-time sweep if that becomes annoying.

cyanglaz: Sounds good!

cyanglaz (approved):
LGTM

script/tool/lib/src/version_check_command.dart
```diff
@@ -96,48 +94,59 @@ class VersionCheckCommand extends PluginCommand {
...
-          continue;
-        }
-        final Pubspec pubspec = Pubspec.parse(pubspecFile.readAsStringSync());
-        if (pubspec.publishTo == 'none') {
-          continue;
-        }
+      print('Checking versions for $pubspecPath...');
+      final File pubspecFile = fileSystem.file(pubspecPath);
+      if (!pubspecFile.existsSync()) {
+        print('${indentation}Deleted; skipping.');
```
ditman, line 102:
```suggestion
        print('${indentation}pubspec.yaml not found; skipping.');
```

script/tool/lib/src/version_check_command.dart
```diff
@@ -96,48 +94,59 @@ class VersionCheckCommand extends PluginCommand {
...
+        print('$indentation$headVersion -> $masterVersion');
+      }
+
+      final bool isPlatformInterface =
+          pubspec.name.endsWith('_platform_interface');
+      if (isPlatformInterface &&
+          allowedNextVersions[headVersion] == NextVersionType.BREAKING_MAJOR) {
+        final String error = '$pubspecPath breaking change detected.\n'
+            'Breaking changes to platform interfaces are strongly discouraged.\n';
+        printErrorAndExit(errorMessage: error);
```
ditman, line 149: These happened all the time with the null safety migration, and will probably happen again. Is there any flag to make a major version change to a platform interface a non-error?

packages/path_provider/path_provider_macos/example/pubspec.yaml
```diff
@@ -1,7 +1,8 @@
 name: path_provider_example
 description: Demonstrates how to use the path_provider plugin.
+publish_to: none
 
-dependencies:
+dependencie:
```
ditman, line 5: (Please revert this typo)

```suggestion
dependencies:
```

ditman (changes requested):
Looks good to me, but there's a small typo that slipped through?

script/tool/lib/src/version_check_command.dart
stuartmorgan: The assumption here is that if there's no pubspec.yaml it's because the whole plugin has been deleted, so I didn't think it was useful to call out the specific file that's not there.

script/tool/lib/src/version_check_command.dart
stuartmorgan: I didn't see one. I didn't want to keep expanding the scope of this PR even further so I left this for us to revisit later.

I think we'll need a way to add a special comment to the version line to override this (I was thinking about the same for new plugins that don't start at version 0.1.0 or 1.0.0 to reduce the chance of plugins accidentally falling into the "allow anything; it's a new plugin!" case as was happening previously).

packages/path_provider/path_provider_macos/example/pubspec.yaml
stuartmorgan: Oh, that's why CI is failing. Thanks :)

script/tool/lib/src/version_check_command.dart
ditman: Ahhh, I understand now!

ditman (approved):
`format` CI check is taking issue, a pass of `dartfmt -w .` should fix it!

PR #2616, opened 2020-03-23
[local_auth] support localizedFallbackTitle in IOSAuthMessages

kaparray:
## Description

 Add support localizedFallbackTitle in IOSAuthMessages.

## Related Issues

https://github.com/flutter/flutter/issues/53205

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] All existing and new tests are passing.
- [X] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [X] I read and followed the [Flutter Style Guide].
- [X] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [X] I updated CHANGELOG.md to add a description of the change.
- [X] I signed the [CLA].
- [X] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [X] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #3806, opened 2021-04-12
[local_auth] support localizedFallbackTitle in IOSAuthMessages

furaiev:
## Description
Add support localizedFallbackTitle in IOSAuthMessages.

## Related Issues
Fixes flutter/flutter#34638

## Pre-launch Checklist

- [x] I signed the [CLA].


furaiev:
@googlebot I signed it!

furaiev:
@stuartmorgan could you please review?

cyanglaz:
This is pretty similar but does have some improvement of https://github.com/flutter/plugins/pull/2616.
CC @stuartmorgan since he already started reviewing this other one.

packages/local_auth/ios/Classes/FLTLocalAuthPlugin.m
```diff
@@ -122,7 +122,10 @@ - (void)authenticateWithBiometrics:(NSDictionary *)arguments
   NSError *authError = nil;
   self.lastCallArgs = nil;
   self.lastResult = nil;
-  context.localizedFallbackTitle = @"";
+  context.localizedFallbackTitle = nil;
+  if (arguments[@"localizedFallbackTitle"] != (NSString *)[NSNull null]) {
```
stuartmorgan, line 126: Why are you casting an `NSNull` to an `NSString`? They are completely different types.

Also, you should ideally handle both `nil` and `NSNull` here for safety.

packages/local_auth/ios/Classes/FLTLocalAuthPlugin.m
```diff
@@ -122,7 +122,10 @@ - (void)authenticateWithBiometrics:(NSDictionary *)arguments
   NSError *authError = nil;
   self.lastCallArgs = nil;
   self.lastResult = nil;
-  context.localizedFallbackTitle = @"";
+  context.localizedFallbackTitle = nil;
```
stuartmorgan, line 125: Nil is the default value, so this line is unnecessary.

packages/local_auth/lib/auth_strings.dart
```diff
@@ -68,20 +68,23 @@ class IOSAuthMessages {
...
   final String? lockOut;
   final String? goToSettingsButton;
   final String? goToSettingsDescription;
   final String? cancelButton;
+  final String? localizedFallbackTitle;
 
-  Map<String, String> get args {
-    return <String, String>{
+  Map<String, String?> get args {
+    return <String, String?>{
```
stuartmorgan, line 81: You can conditionally include `'localizedFallbackTitle'` rather than making the values nullable. (Which will require the change to handle `nil` mentioned above.)

stuartmorgan (changes requested):
Apologies, for the long delay in getting back to this review. Specific comments inline below, but this will also need a test, per the contribution guide. In this case, since it involves native UI, I think the easiest approach would be to add a native unit test that uses a mock `LAContext` and ensures that the property is set (or not set) correctly based on a constructor method call object. See https://github.com/flutter/plugins/blob/master/packages/local_auth/example/ios/RunnerTests/FLTLocalAuthPluginTests.m

furaiev:
@stuartmorgan thank you for your review.
I've updated the PR according to your comments but faced an issue with tests.
`localizedFallbackTitle` isn't passed via the constructor, it is just a `LAContext` property and I didn't found a way to check if it was assigned or not.
Could you please advise other ways to test it?

stuartmorgan:
> `localizedFallbackTitle` isn't passed via the constructor, it is just a `LAContext` property

Sorry, "constructor" was a typo in my comment. That should have said a **constructed** method call argument.

> and I didn't found a way to check if it was assigned or not.

See `OCMVerify`.

furaiev:
Thank you @stuartmorgan ,
I've tried `OCMVerify` and it is triggered on `get` but not on `set`.

`OCMVerify([mockAuthContext localizedFallbackTitle])` returns `OCMockObject(LAContext): Method localizedFallbackTitle was not invoked.`

stuartmorgan:
> I've tried `OCMVerify` and it is triggered on `get` but not on `set`.

That's because you are providing the method name of the getter. The setter for a property called `localizedFallbackTitle` is `setLocalizedFallbackTitle:`

furaiev:
thank you @stuartmorgan !
PR is updated

packages/local_auth/test/local_auth_test.dart
```diff
@@ -73,6 +73,30 @@ void main() {
...
+          log,
+          <Matcher>[
+            isMethodCall('authenticate',
+                arguments: <String, dynamic>{
+                  'localizedReason': 'Needs secure',
+                  'useErrorDialogs': true,
+                  'stickyAuth': false,
+                  'sensitiveTransaction': true,
+                  'biometricOnly': true,
+                }..addAll(iosAuthMessages.args)),
```
stuartmorgan, line 95: I see that this is standard practice in the existing tests in this file, but it's actually a very bad idea; this test wouldn't have failed with your incorrect key bug, because it's using part of the object under test to generate the test expectations, which undermines the test. Please replace this with an explicit list of expected arguments. (You don't have to fix the existing instances since it's not related to your PR, but if you wanted to that would be great.)

packages/local_auth/test/local_auth_test.dart
```diff
@@ -149,6 +173,29 @@ void main() {
...
+                arguments: <String, dynamic>{
+                  'localizedReason': 'Needs secure',
+                  'useErrorDialogs': true,
+                  'stickyAuth': false,
+                  'sensitiveTransaction': true,
+                  'biometricOnly': false,
+                }..addAll(iosAuthMessages.args)),
+          ],
+        );
+      });
```
stuartmorgan, line 197: You don't need to have this second test; there's nothing different about the code paths at this layer, so this isn't testing anything different than the first one.

stuartmorgan (changes requested):

stuartmorgan (approved):
LGTM! Sorry for the delay on this final review round.

@gaaclarke for secondary review per Flutter policy.

packages/local_auth/example/ios/RunnerTests/FLTLocalAuthPluginTests.m
```diff
@@ -56,13 +56,15 @@ - (void)testSuccessfullAuthWithBiometrics {
                                                               arguments:@{
                                                                 @"biometricOnly" : @(YES),
                                                                 @"localizedReason" : reason,
+                                                                @"localizedFallbackTitle" : @"",
```
gaaclarke, line 59: Shouldn't we have a test where `localizedFallbackTitle` isn't defined?  And if it isn't we'd want `OCMVerify([mockAuthContext setLocalizedFallbackTitle:@""]);` right?  That isn't how it would turn on the way the code is written.  It'd send in `nil`.

stuartmorgan: `nil` is correct:

> Use the localizedFallbackTitle property to set a title for the fallback button. When the property is nil—as it is by default—the user sees an appropriate default title, like “Use Passcode”.

(From Apple's docs)

stuartmorgan: But you're right about the test cases; we should be not sending anything in the normal tests, because that's what the Dart code does when there's no fallback. I missed that.

stuartmorgan (changes requested):

packages/local_auth/example/ios/RunnerTests/FLTLocalAuthPluginTests.m
```diff
@@ -186,4 +186,47 @@ - (void)testFailedAuthWithoutBiometrics {
...
+                    result:^(id _Nullable result) {
+                      XCTAssertTrue([NSThread isMainThread]);
+                      XCTAssertTrue([result isKindOfClass:[NSNumber class]]);
+                      OCMVerify([mockAuthContext setLocalizedFallbackTitle:localizedFallbackTitle]);
+                      XCTAssertFalse([result boolValue]);
+                      [expectation fulfill];
+                    }];
+  [self waitForExpectationsWithTimeout:kTimeout handler:nil];
+}
+
```
stuartmorgan, line 231: There should be a test that when no fallback is passed, `setLocalizedFallbackTitle:` is not called at all, since part of what this PR is fixing is the call with `@""` that disabled the fallback.

furaiev: @stuartmorgan thank you for the fast feedback.
The previous 4 tests are called with arguments:
```
arguments:@{
  @"biometricOnly" : @(NO), // or YES
  @"localizedReason" : reason,
}];
```
Should I add `OCMVerify([mockAuthContext setLocalizedFallbackTitle:nil]);` to each of them?

stuartmorgan: I would add a specific new test whose purpose is to make the specific assertion; unit tests should generally be as discuss as possible.

(You're right that it should asset that it's called with nil rather than not called; I misremembered the final code.)

packages/local_auth/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.1.11
+
+* Add support `localizedFallbackTitle` in authenticateWithBiometrics on iOS.
```
stuartmorgan, line 3: Nit: Adds

(Per recently add CHANGELOG style guide.)

stuartmorgan (approved):
LGTM with nit. @gaaclarke this is ready for re-review.

stuartmorgan:
@gaaclarke Ping for review.

stuartmorgan:
@furaiev If you can resolve the conflicts quickly we can land this ASAP, otherwise it'll need to wait until after https://github.com/flutter/plugins/pull/4699 and https://github.com/flutter/plugins/pull/4701 land and be reconciled against those changes.

(I don't expect it would be difficult to resolve either way, so no rush.)

furaiev:
@stuartmorgan conflicts are resolved. Thanks!

stuartmorgan:
Could you re-run the autoformatter? We're enforcing `*` alignment in ObjC now.

furaiev:
@stuartmorgan done

stuartmorgan:
Thanks!

furaiev:
@stuartmorgan it seems like the bot found some unresolved requests but I see only resolved conversations

stuartmorgan:
No clue 🤷🏻 

Landing manually.

PR #2887, opened 2020-07-21
[quick_actions] Fixed plugin on iOS. Entire lifecycle was not taken care of

olivierbrand:
## Description

The PR includes an entire iOS lifecycle, allowing the actions to trigger when the app is in the background and/or not running.
Also modified the Dart example to provide a more accurate view of what is going on by using a dialog rather than changing the title bar (the title bar was not changing when triggered without actions, making it more confusing for the developer)

## Related Issues

N/A

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [X ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ X] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [ X] All existing and new tests are passing.
- [ X] I updated/added relevant documentation (doc comments with `///`).
- [ X] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [ X] I read and followed the [Flutter Style Guide].
- [X ] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [ X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [X ] I updated CHANGELOG.md to add a description of the change.
- [ X] I signed the [CLA].
- [ X] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [X ] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #3811, opened 2021-04-16
[quick_actions] handle cold start on iOS correctly

mvanbeusekom:
According to Apple's documentation (see the "Discussion" section [here](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623032-application?language=objc)) the `application:performActionForShortcutItem:` method is not executed when the `willFinishLaunchingWithOptions:` or the `didFinishLaunchingWithOptions:` return `NO`. In these cases the `willFinishLaunchingWithOptions:` or `didFinishLaunchingWithOptions:` method are responsible for handling quick actions.

Basic Flutter applications will always return `YES` for the `willFinishLaunchingWithOptions:` and `didFinishLaunchingWithOptions:` methods, however in more advanced cases developers might override these methods and provide custom implementations which might result in one of these method to return `NO`. In these situations the current implementation of the quick_actions plugin will not work. 

This PR will solve the problem and correctly handle quick actions in the `didFinishLaunchingWithOptions:` delegate of the plugin. It also contains an additional XCUITest to proof this method is working correctly. To be able to simulate the situation the `didFinishLaunchingWithOptions:` method in the `example/ios/Runner/AppDelegate.m` is updated to always return `NO`. 

Running the XCUITest on the current codebase will fail if you update the `didFinishLaunchingWithOptions:` method in the `example/ios/Runner/AppDelegate.m` to return `NO`. Running the same test on the code modified in this PR the test will succeed.

- Fixes flutter/flutter#13634
- Closes #2887

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/quick_actions/quick_actions/ios/Classes/FLTQuickActionsPlugin.m
+++ b/packages/quick_actions/quick_actions/ios/Classes/FLTQuickActionsPlugin.m
@@ -8,11 +8,10 @@ static NSString *const CHANNEL_NAME = @"plugins.flutter.io/quick_actions";
 
 @interface FLTQuickActionsPlugin ()
 @property(nonatomic, retain) FlutterMethodChannel *channel;
+@property(nonatomic, retain) NSString *shortcutType;
 @end
 
-@implementation FLTQuickActionsPlugin {
-  NSString *shortcutType;
-}
+@implementation FLTQuickActionsPlugin
 
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
   FlutterMethodChannel *channel =
@@ -52,7 +51,7 @@ static NSString *const CHANNEL_NAME = @"plugins.flutter.io/quick_actions";
     performActionForShortcutItem:(UIApplicationShortcutItem *)shortcutItem
                completionHandler:(void (^)(BOOL succeeded))completionHandler
     API_AVAILABLE(ios(9.0)) {
-  [self _handleShortcut:shortcutItem.type];
+  [self handleShortcut:shortcutItem.type];
   return YES;
 }
 
@@ -62,7 +61,7 @@ static NSString *const CHANNEL_NAME = @"plugins.flutter.io/quick_actions";
     UIApplicationShortcutItem *shortcutItem =
         launchOptions[UIApplicationLaunchOptionsShortcutItemKey];
     if (shortcutItem) {
-      shortcutType = shortcutItem.type;
+      self.shortcutType = shortcutItem.type;
       return NO;
     }
   }
@@ -70,15 +69,15 @@ static NSString *const CHANNEL_NAME = @"plugins.flutter.io/quick_actions";
 }
 
 - (void)applicationDidBecomeActive:(UIApplication *)application {
-  if (shortcutType) {
-    [self _handleShortcut:shortcutType];
-    shortcutType = nil;
+  if (self.shortcutType) {
+    [self handleShortcut:self.shortcutType];
+    self.shortcutType = nil;
   }
 }
 
 #pragma mark Private functions
 
-- (void)_handleShortcut:(NSString *)shortcut {
+- (void)handleShortcut:(NSString *)shortcut {
   [self.channel invokeMethod:@"launch" arguments:shortcut];
 }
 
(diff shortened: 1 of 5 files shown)


packages/quick_actions/quick_actions/ios/Classes/FLTQuickActionsPlugin.m
```diff
@@ -10,7 +10,9 @@ @interface FLTQuickActionsPlugin ()
 @property(nonatomic, retain) FlutterMethodChannel *channel;
 @end
 
-@implementation FLTQuickActionsPlugin
+@implementation FLTQuickActionsPlugin {
+  NSString *shortcutType;
```
cyanglaz, line 14: nits: Let's make this a property instead of ivar to match the code style of this file.

packages/quick_actions/quick_actions/ios/Classes/FLTQuickActionsPlugin.m
```diff
@@ -50,12 +52,36 @@ - (BOOL)application:(UIApplication *)application
...
+- (void)applicationDidBecomeActive:(UIApplication *)application {
+  if (shortcutType) {
+    [self _handleShortcut:shortcutType];
+    shortcutType = nil;
+  }
+}
+
 #pragma mark Private functions
 
+- (void)_handleShortcut:(NSString *)shortcut {
```
cyanglaz, line 81: nits: OBJC method should not start with `handleShortcut` is fine.

packages/quick_actions/quick_actions/ios/Classes/FLTQuickActionsPlugin.m
```diff
@@ -50,12 +52,36 @@ - (BOOL)application:(UIApplication *)application
...
     API_AVAILABLE(ios(9.0)) {
-  [self.channel invokeMethod:@"launch" arguments:shortcutItem.type];
+  [self _handleShortcut:shortcutItem.type];
   return YES;
 }
 
+- (BOOL)application:(UIApplication *)application
+    didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+  if (@available(iOS 9.0, *)) {
+    UIApplicationShortcutItem *shortcutItem =
```
cyanglaz, line 62: Why don't we  directly call `handleShortcut` here?

mvanbeusekom: We have to wait here for the application to become fully active before the method channel are fully initialised. When we directly call the `handleShortcut` in the `didFinishLaunchingWithOptions` method the channels on the Dart side are not initialised yet and the `invokeMethod:@"launch"` isn't processed.

packages/quick_actions/quick_actions/ios/Classes/FLTQuickActionsPlugin.m
mvanbeusekom: done

packages/quick_actions/quick_actions/ios/Classes/FLTQuickActionsPlugin.m
mvanbeusekom: done

mvanbeusekom:
@cyanglaz thank you very much for your feedback. I corrected the "nits" and answered the question regarding why not to call the `handleShortcut` directly from the`didFinishLaunchingWithOptions` method. Would appreciate it if you could have another look.

packages/quick_actions/quick_actions/ios/Classes/FLTQuickActionsPlugin.m
cyanglaz: Ah makes sense. Let's add a comment here for future reference.

Why do we return `No` below?

mvanbeusekom: According to Apple's documentation you should return `NO` if you already handled the shortcut action which will make sure the `application:performActionFor:` method and not handle the same action again.

The explanation can be found in the last alinea of the "Discussion" section of this article: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623032-application

cyanglaz: OK. Let's also add comment here too. Then I think we are good to go!

mvanbeusekom: Thanks @cyanglaz, I have added comments as explanation.

cyanglaz (approved):
LGTM

PR #3817, opened 2021-04-21
[url_launcher] Add a workaround for Uri encoding

stuartmorgan:
`Uri`'s constructor doesn't handle query parameters correctly for
non-http(s) schemes, so the `mailto` example in the README is
misleading. This updates the README to show using a simple method
to work around that bug, and a warning about the need to use it.

Fixes https://github.com/flutter/flutter/issues/75552
Fixes https://github.com/flutter/flutter/issues/73717

## Pre-launch Checklist

- [x] I signed the [CLA].


patreu22:
Hey @stuartmorgan , I'm not sure if you are still working on that fix here, but I'm running into an issue that looks like your PR could possibly fix my issue with the encoding of linebreaks: https://stackoverflow.com/questions/67859136/flutter-url-launcher-line-break-doesnt-work-in-ios-mail-program

Maybe you find some time to finalize your PR, I think it could be very helpful!

Best regards
Patrick

stuartmorgan:
A decision on what to do here is pending a decision in https://github.com/dart-lang/sdk/issues/43838

I may just do a README update in the short term.

patreu22:
I see, thx for coming back to me so quickly! Will use the suggested workaround from there and just replace the `+` with a `%20`. Still not fixing my line break issue but I'll look out for other fixes to that.

stuartmorgan:
Based on discussion with Hixie we're not going to put the workaround in the plugin itself, where we'll need to maintain it (or break people) even after the Dart bug is fixed, so I've updated this to just show the workaround in the README. This will avoid the issue of people following the README and getting the wrong behavior.

gaaclarke (approved):

ditman (approved):
There's a simpler version where people can call `Uri.toString().replace("+", "%20")`, but this one is more "standard" (especially in JS you'd end up some code like this if you were concatenating your own URLs).

Anyway, ship it! 🚀

stuartmorgan:
> There's a simpler version where people can call `Uri.toString().replace("+", "%20")`

Only in the same sense that there's a simpler version where you write `'mailto:$emailAddress?subject=$subject`. I.e., it's easier to write, but is incorrect and may not actually work.

> but this one is more "standard"

If by "standard" you mean "correct", then yes 🙂  The URI spec allows for `+` to exist unencoded in certain places, and also specifically says that "URIs that differ in the replacement of a reserved character with its corresponding percent-encoded octet are not equivalent.  Percent-encoding a reserved character, or decoding a percent-encoded octet that corresponds to a reserved character, will change how the URI is interpreted by most applications."

Which means that your `replace` has the potential to corrupt a URI. People should not be writing that.

stuartmorgan:
Landing this simple PR on red to force post-submit CI to run, since we believe the last round of post-submit failures was due to an infra migration that was happening that impacted Cirrus jobs.

PR #3819, opened 2021-04-21
Add a todo WRT correctly setting the X-Goog-AuthUser header

kevmoo:
See https://github.com/flutter/flutter/issues/80905


ditman (approved):
LGTM, thanks for noticing this!

PR #3820, opened 2021-04-22
[video_player] Android: Rotate videos recorded in landscapeRight

KyleFin:
This change rotates videos recorded in landscapeRight orientation to fix the bug where they were upside-down.

https://github.com/flutter/flutter/issues/60327

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
+++ b/packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
@@ -279,6 +279,9 @@ final class VideoPlayer {
           width = exoPlayer.getVideoFormat().height;
           height = exoPlayer.getVideoFormat().width;
         }
+        if (rotationDegrees == 180) {
+          event.put("rotation", Math.PI);
+        }
         event.put("width", width);
         event.put("height", height);
       }
--- a/packages/video_player/video_player/example/integration_test/video_player_test.dart
+++ b/packages/video_player/video_player/example/integration_test/video_player_test.dart
@@ -129,6 +129,51 @@ void main() {
       },
     );
 
+    testWidgets(
+      'stay paused when seeking after video completed',
+      (WidgetTester tester) async {
+        await _controller.initialize();
+        // Mute to allow playing without DOM interaction on Web.
+        // See https://developers.google.com/web/updates/2017/09/autoplay-policy-changes
+        await _controller.setVolume(0);
+        Duration tenMillisBeforeEnd =
+            _controller.value.duration - const Duration(milliseconds: 10);
+        await _controller.seekTo(tenMillisBeforeEnd);
+        await _controller.play();
+        await tester.pumpAndSettle(_playDuration);
+        expect(_controller.value.isPlaying, false);
+        expect(_controller.value.position, _controller.value.duration);
+
+        await _controller.seekTo(tenMillisBeforeEnd);
+        await tester.pumpAndSettle(_playDuration);
+
+        expect(_controller.value.isPlaying, false);
+        expect(_controller.value.position, tenMillisBeforeEnd);
+      },
+    );
+
+    testWidgets(
+      'do not exceed duration on play after video completed',
+      (WidgetTester tester) async {
+        await _controller.initialize();
+        // Mute to allow playing without DOM interaction on Web.
+        // See https://developers.google.com/web/updates/2017/09/autoplay-policy-changes
+        await _controller.setVolume(0);
+        await _controller.seekTo(
+            _controller.value.duration - const Duration(milliseconds: 10));
+        await _controller.play();
+        await tester.pumpAndSettle(_playDuration);
+        expect(_controller.value.isPlaying, false);
+        expect(_controller.value.position, _controller.value.duration);
+
+        await _controller.play();
+        await tester.pumpAndSettle(_playDuration);
+
+        expect(_controller.value.position,
+            lessThanOrEqualTo(_controller.value.duration));
+      },
+    );
+
     testWidgets('test video player view with local asset',
         (WidgetTester tester) async {
       Future<bool> started() async {
--- a/packages/video_player/video_player/example/integration_test/video_player_test.dart
+++ b/packages/video_player/video_player/example/integration_test/video_player_test.dart
@@ -129,51 +129,6 @@ void main() {
       },
     );
 
-    testWidgets(
-      'stay paused when seeking after video completed',
-      (WidgetTester tester) async {
-        await _controller.initialize();
-        // Mute to allow playing without DOM interaction on Web.
-        // See https://developers.google.com/web/updates/2017/09/autoplay-policy-changes
-        await _controller.setVolume(0);
-        Duration tenMillisBeforeEnd =
-            _controller.value.duration - const Duration(milliseconds: 10);
-        await _controller.seekTo(tenMillisBeforeEnd);
-        await _controller.play();
-        await tester.pumpAndSettle(_playDuration);
-        expect(_controller.value.isPlaying, false);
-        expect(_controller.value.position, _controller.value.duration);
-
-        await _controller.seekTo(tenMillisBeforeEnd);
-        await tester.pumpAndSettle(_playDuration);
-
-        expect(_controller.value.isPlaying, false);
-        expect(_controller.value.position, tenMillisBeforeEnd);
-      },
-    );
-
-    testWidgets(
-      'do not exceed duration on play after video completed',
-      (WidgetTester tester) async {
-        await _controller.initialize();
-        // Mute to allow playing without DOM interaction on Web.
-        // See https://developers.google.com/web/updates/2017/09/autoplay-policy-changes
-        await _controller.setVolume(0);
-        await _controller.seekTo(
-            _controller.value.duration - const Duration(milliseconds: 10));
-        await _controller.play();
-        await tester.pumpAndSettle(_playDuration);
-        expect(_controller.value.isPlaying, false);
-        expect(_controller.value.position, _controller.value.duration);
-
-        await _controller.play();
-        await tester.pumpAndSettle(_playDuration);
-
-        expect(_controller.value.position,
-            lessThanOrEqualTo(_controller.value.duration));
-      },
-    );
-
     testWidgets('test video player view with local asset',
         (WidgetTester tester) async {
       Future<bool> started() async {
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -314,8 +314,7 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
           _applyPlayPause();
           break;
         case VideoEventType.completed:
-          value = value.copyWith(isPlaying: false, position: value.duration);
-          _timer?.cancel();
+          pause().then((void pauseResult) => seekTo(value.duration));
           break;
         case VideoEventType.bufferingUpdate:
           value = value.copyWith(buffered: event.buffered);
@@ -375,6 +374,9 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   /// has been sent to the platform, not when playback itself is totally
   /// finished.
   Future<void> play() async {
+    if (value.position == value.duration) {
+      await seekTo(const Duration());
+    }
     value = value.copyWith(isPlaying: true);
     await _applyPlayPause();
   }
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -370,6 +370,8 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
 
   /// Starts playing the video.
   ///
+  /// If the video is at the end, this method starts playing from the beginning.
+  ///
   /// This method returns a future that completes as soon as the "play" command
   /// has been sent to the platform, not when playback itself is totally
   /// finished.
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -370,8 +370,6 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
 
   /// Starts playing the video.
   ///
-  /// If the video is at the end, this method starts playing from the beginning.
-  ///
   /// This method returns a future that completes as soon as the "play" command
   /// has been sent to the platform, not when playback itself is totally
   /// finished.
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -314,7 +314,8 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
           _applyPlayPause();
           break;
         case VideoEventType.completed:
-          pause().then((void pauseResult) => seekTo(value.duration));
+          value = value.copyWith(isPlaying: false, position: value.duration);
+          _timer?.cancel();
           break;
         case VideoEventType.bufferingUpdate:
           value = value.copyWith(buffered: event.buffered);
@@ -374,9 +375,6 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   /// has been sent to the platform, not when playback itself is totally
   /// finished.
   Future<void> play() async {
-    if (value.position == value.duration) {
-      await seekTo(const Duration());
-    }
     value = value.copyWith(isPlaying: true);
     await _applyPlayPause();
   }
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -39,6 +39,7 @@ class VideoPlayerValue {
     this.isBuffering = false,
     this.volume = 1.0,
     this.playbackSpeed = 1.0,
+    this.rotation = 0.0,
     this.errorDescription,
   });
 
@@ -93,6 +94,9 @@ class VideoPlayerValue {
   /// The [size] of the currently loaded video.
   final Size size;
 
+  /// The [rotation] of the video.
+  final double rotation;
+
   /// Indicates whether or not the video has been loaded and is ready to play.
   final bool isInitialized;
 
@@ -118,7 +122,7 @@ class VideoPlayerValue {
   }
 
   /// Returns a new instance that has the same values as this current instance,
-  /// except for any overrides passed in as arguments to [copyWidth].
+  /// except for any overrides passed in as arguments to [copyWith].
   VideoPlayerValue copyWith({
     Duration? duration,
     Size? size,
@@ -131,6 +135,7 @@ class VideoPlayerValue {
     bool? isBuffering,
     double? volume,
     double? playbackSpeed,
+    double? rotation,
     String? errorDescription,
   }) {
     return VideoPlayerValue(
@@ -145,6 +150,7 @@ class VideoPlayerValue {
       isBuffering: isBuffering ?? this.isBuffering,
       volume: volume ?? this.volume,
       playbackSpeed: playbackSpeed ?? this.playbackSpeed,
+      rotation: rotation ?? this.rotation,
       errorDescription: errorDescription ?? this.errorDescription,
     );
   }
@@ -306,6 +312,7 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
           value = value.copyWith(
             duration: event.duration,
             size: event.size,
+            rotation: event.rotation,
             isInitialized: event.duration != null,
           );
           initializingCompleter.complete(null);
@@ -639,7 +646,10 @@ class _VideoPlayerState extends State<VideoPlayer> {
   Widget build(BuildContext context) {
     return _textureId == VideoPlayerController.kUninitializedTextureId
         ? Container()
-        : _videoPlayerPlatform.buildView(_textureId);
+        : Transform.rotate(
+            angle: widget.controller.value.rotation,
+            child: _videoPlayerPlatform.buildView(_textureId),
+          );
   }
 }
 
--- a/packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
+++ b/packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
@@ -213,12 +213,13 @@ class VideoEvent {
   ///
   /// The [eventType] argument is required.
   ///
-  /// Depending on the [eventType], the [duration], [size] and [buffered]
-  /// arguments can be null.
+  /// Depending on the [eventType], the [duration], [size], [rotation], and
+  /// [buffered] arguments can be null.
   VideoEvent({
     required this.eventType,
     this.duration,
     this.size,
+    this.rotation,
     this.buffered,
   });
 
@@ -234,6 +235,11 @@ class VideoEvent {
   ///
   /// Only used if [eventType] is [VideoEventType.initialized].
   final Size? size;
+
+  /// Rotation of the video.
+  ///
+  /// Only used if [eventType] is [VideoEventType.initialized].
+  final double? rotation;
 
   /// Buffered parts of the video.
   ///
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,13 +1,245 @@
-include: package:pedantic/analysis_options.1.8.0.yaml
+# This is a copy (as of March 2021) of flutter/flutter's analysis_options file,
+# with minimal changes for this repository. The goal is to move toward using a
+# shared set of analysis options as much as possible, and eventually a shared
+# file.
+#
+# Plugins that have not yet switched from the previous set of options have a
+# local analysis_options.yaml that points to analysis_options_legacy.yaml
+# instead.
+
+# Specify analysis options.
+#
+# Until there are meta linter rules, each desired lint must be explicitly enabled.
+# See: https://github.com/dart-lang/linter/issues/288
+#
+# For a list of lints, see: http://dart-lang.github.io/linter/lints/
+# See the configuration guide for more
+# https://github.com/dart-lang/sdk/tree/master/pkg/analyzer#configuring-the-analyzer
+#
+# There are other similar analysis options files in the flutter repos,
+# which should be kept in sync with this file:
+#
+#   - analysis_options.yaml (this file)
+#   - packages/flutter/lib/analysis_options_user.yaml
+#   - https://github.com/flutter/plugins/blob/master/analysis_options.yaml
+#   - https://github.com/flutter/engine/blob/master/analysis_options.yaml
+#
+# This file contains the analysis options used by Flutter tools, such as IntelliJ,
+# Android Studio, and the `flutter analyze` command.
+
 analyzer:
+  strong-mode:
+    implicit-casts: false
+    implicit-dynamic: false
+  errors:
+    # treat missing required parameters as a warning (not a hint)
+    missing_required_param: warning
+    # treat missing returns as a warning (not a hint)
+    missing_return: warning
+    # allow having TODOs in the code
+    todo: ignore
+    # allow self-reference to deprecated members (we do this because otherwise we have
+    # to annotate every member in every test, assert, etc, when we deprecate something)
+    deprecated_member_use_from_same_package: ignore
+    # Ignore analyzer hints for updating pubspecs when using Future or
+    # Stream and not importing dart:async
+    # Please see https://github.com/flutter/flutter/pull/24528 for details.
+    sdk_version_async_exported_from_core: ignore
+    ### Local flutter/plugins changes ###
+    # Allow null checks for as long as mixed mode is officially supported.
+    unnecessary_null_comparison: false
+    always_require_non_null_named_parameters: false # not needed with nnbd
   exclude:
     # Ignore generated files
     - '**/*.g.dart'
     - 'lib/src/generated/*.dart'
     - '**/*.mocks.dart' # Mockito @GenerateMocks
-  errors:
-    always_require_non_null_named_parameters: false # not needed with nnbd
-    unnecessary_null_comparison: false # Turned as long as nnbd mix-mode is supported.
+
 linter:
   rules:
-    - public_member_api_docs
+    # these rules are documented on and in the same order as
+    # the Dart Lint rules page to make maintenance easier
+    # https://github.com/dart-lang/linter/blob/master/example/all.yaml
+    - always_declare_return_types
+    - always_put_control_body_on_new_line
+    # - always_put_required_named_parameters_first # we prefer having parameters in the same order as fields https://github.com/flutter/flutter/issues/10219
+    - always_require_non_null_named_parameters
+    - always_specify_types
+    # - always_use_package_imports # we do this commonly
+    - annotate_overrides
+    # - avoid_annotating_with_dynamic # conflicts with always_specify_types
+    # - avoid_as # required for implicit-casts: true
+    - avoid_bool_literals_in_conditional_expressions
+    # - avoid_catches_without_on_clauses # we do this commonly
+    # - avoid_catching_errors # we do this commonly
+    - avoid_classes_with_only_static_members
+    # - avoid_double_and_int_checks # only useful when targeting JS runtime
+    - avoid_empty_else
+    - avoid_equals_and_hash_code_on_mutable_classes
+    # - avoid_escaping_inner_quotes # not yet tested
+    - avoid_field_initializers_in_const_classes
+    - avoid_function_literals_in_foreach_calls
+    # - avoid_implementing_value_types # not yet tested
+    - avoid_init_to_null
+    # - avoid_js_rounded_ints # only useful when targeting JS runtime
+    - avoid_null_checks_in_equality_operators
+    # - avoid_positional_boolean_parameters # not yet tested
+    # - avoid_print # not yet tested
+    # - avoid_private_typedef_functions # we prefer having typedef (discussion in https://github.com/flutter/flutter/pull/16356)
+    # - avoid_redundant_argument_values # not yet tested
+    - avoid_relative_lib_imports
+    - avoid_renaming_method_parameters
+    - avoid_return_types_on_setters
+    # - avoid_returning_null # there are plenty of valid reasons to return null
+    # - avoid_returning_null_for_future # not yet tested
+    - avoid_returning_null_for_void
+    # - avoid_returning_this # there are plenty of valid reasons to return this
+    # - avoid_setters_without_getters # not yet tested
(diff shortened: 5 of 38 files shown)


packages/video_player/video_player/lib/video_player.dart
```diff
@@ -639,7 +646,10 @@ class _VideoPlayerState extends State<VideoPlayer> {
   Widget build(BuildContext context) {
     return _textureId == VideoPlayerController.kUninitializedTextureId
         ? Container()
-        : _videoPlayerPlatform.buildView(_textureId);
+        : Transform.rotate(
+            angle: widget.controller.value.rotation,
+            child: _videoPlayerPlatform.buildView(_textureId),
+          );
```
KyleFin, line 652: This is the heart of this PR (the rest is plumbing so Android plugin can set rotation if rotationDegrees == 180). Transforming the widget doesn't seem terrible, but I'd prefer to fix it at the Java level.

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -279,6 +279,9 @@ private void sendInitialized() {
           width = exoPlayer.getVideoFormat().height;
           height = exoPlayer.getVideoFormat().width;
         }
+        if (rotationDegrees == 180) {
+          event.put("rotation", Math.PI);
+        }
```
KyleFin, line 284: This seems ok IMHO, but I'd prefer to rotate the video correctly within the Java code instead of rotating the flutter VideoPlayer widget.

One option I started looking into is using a [TextureView](https://source.android.com/devices/graphics/arch-tv) with [exoPlayer.setVideoTextureView](https://google.github.io/ExoPlayer/doc/reference/com/google/android/exoplayer2/SimpleExoPlayer.html#setVideoTextureView-android.view.TextureView-) instead of a [Surface](https://developer.android.com/reference/android/view/Surface.html) on line 170 in this file. (The top answers on these old StackOverflow questions [[1](https://stackoverflow.com/questions/48841506/rotating-exoplayer), [2](https://stackoverflow.com/questions/44564356/how-to-rotate-exoplayer)] about rotating exoplayer and [the docs](https://source.android.com/devices/graphics/arch-tv#surface_or_texture) comparing SurfaceView and TextureView suggest TextureView is better, or required, for rotation. The docs suggest TextureView can reduce performance because of duplicate compositing and shouldn't be used in newer APIs.)

It seems switching from Surface to TextureView would be a much larger refactoring of how the plugin works. Combined with the performance concerns of TextureView, maybe the Flutter rotation isn't such a bad solution.

I do want to look more into why these videos aren't upside down in the ExoPlayer sample app (according to cyanglaz's [comment](https://github.com/flutter/flutter/issues/60327#issuecomment-713019545)). Does it use a TextureView?

packages/video_player/video_player/example/integration_test/video_player_test.dart
```diff
@@ -162,5 +162,24 @@ void main() {
...
+    testWidgets('test Android video is rotated for landscapeRight',
+        (WidgetTester tester) async {
+      _controller =
+          VideoPlayerController.asset('assets/keyboard-landscape.mp4');
+
+      VideoPlayer player = VideoPlayer(_controller);
+      await tester.pumpWidget(player);
+
+      expect(find.byType(Transform), findsOneWidget);
+    }, skip: kIsWeb); // TODO(KyleFin): Skip for non-Android tests?
```
KyleFin, line 175: @stuartmorgan do you know if there's a way I can run this test only for Android? Do CI tests run on Android? When I run the test on my machine with `flutter drive` ([docs](https://github.com/flutter/flutter/wiki/Plugin-Tests#integration-tests)), it runs in Chrome. (https://github.com/flutter/flutter/issues/32927 doesn't look too promising for finding device platform in integration tests.)

packages/video_player/video_player/example/integration_test/video_player_test.dart
```diff
@@ -162,5 +162,24 @@ void main() {
...
+
+    testWidgets('test Android video is rotated for landscapeRight',
+        (WidgetTester tester) async {
+      _controller =
+          VideoPlayerController.asset('assets/keyboard-landscape.mp4');
+
+      VideoPlayer player = VideoPlayer(_controller);
+      await tester.pumpWidget(player);
+
+      expect(find.byType(Transform), findsOneWidget);
```
KyleFin, line 174: @stuartmorgan is this a reasonable way to inspect the widget and determine if Transform.rotate is present?

packages/video_player/video_player/example/integration_test/video_player_test.dart
KyleFin: I just realized too that when you said "test on the native side that the right metadata is sent, and on the Dart side that the metadata is handled with a transform" maybe I could just use a Dart unit test for this instead of an integration test with an actual video file.

KyleFin: I created a sample test in packages/video_player/video_player/test/video_player_test.dart that verifies the rotation value is used in Transform (doesn't verify the event from Java code will actually set rotation field though).

stuartmorgan: > @stuartmorgan do you know if there's a way I can run this test only for Android?

`flutter drive` takes the same `-d` option that `flutter run` does (I've added that to the wiki, thanks!)

> I just realized too that when you said "test on the native side that the right metadata is sent, and on the Dart side that the metadata is handled with a transform"

I was thinking a Dart unit test + a Java unit test. https://github.com/flutter/plugins/pull/4108 will help with the latter.

packages/video_player/video_player/example/integration_test/video_player_test.dart
stuartmorgan: This isn't an area I have familiarity with; but it seems plausible 🤷🏻

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -50,27 +51,51 @@
...
       Map<String, String> httpHeaders,
-      VideoPlayerOptions options) {
+      VideoPlayerOptions options,
+      SimpleExoPlayer exoPlayer) {
     this.eventChannel = eventChannel;
     this.textureEntry = textureEntry;
     this.options = options;
 
-    exoPlayer = new SimpleExoPlayer.Builder(context).build();
+    if (exoPlayer != null) {  // Skip remaining setup in tests if using a mock exoPlayer.
```
stuartmorgan, line 94: I think the logic would be much clearer here (and less error-prone; someone could easily add non-ExoPlayer-specific code after this block on accident) to have the other constructor do ExoPlayer setup and pass in the ExoPlayer, instead of passing null and then handling null specially here.

(Then you can also drop some parameters from this constructor, which will make it less confusing, since currently it sometimes ignores a bunch of parameters.)

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
stuartmorgan: I think it's reasonable to move forward with this, but let's add a comment here since it's non-obvious outside the context of this PR why this is only being sent for one specific rotation. E.g.:

```
// Rotating the video with ExoPlayer does not seem to be possible with a Surface,
// so inform the Flutter code that the widget needs to be rotated to prevent
// upside-down playback.
```

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -219,12 +219,13 @@ class VideoEvent {
   ///
   /// The [eventType] argument is required.
   ///
-  /// Depending on the [eventType], the [duration], [size] and [buffered]
-  /// arguments can be null.
+  /// Depending on the [eventType], the [duration], [size], [rotation], and
+  /// [buffered] arguments can be null.
```
stuartmorgan, line 223: These changes will need to be split out per https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
stuartmorgan: Although actually: what is the behavior here for the 90/270 cases? Does the video play sideways, or is that handled differently by ExoPlayer than 180?

I'm wondering if we should always send the rotation, so that more information is available to plugin clients on the Dart side.

stuartmorgan (changes requested):

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
KyleFin: I think I'll just rename `rotation` to be `rotationCorrection` and add a comment as you suggested.

I confirmed videos recorded in 0, 90, and 270 orientations are displayed correctly without this PR (only 180 needs to be fixed).

I like the idea of passing along more information to clients, but it seems a bit messy to add a `rotation` value that's currently only populated on Android and only has an effect in the case of 180 on Android.
(If we do pass it along, the `Transform.rotate` widget in video_player.dart would potentially have:
`angle: widget.controller.value.rotation == 180 ? pi : 0.0,`)

Another argument is that clients shouldn't have to care about the recording orientation if they know the default display orientation is correct and they can compare height/width if they want to treat portrait/landscape differently.

WDYT?

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
KyleFin: I'll leave the original constructor alone and just create a VisibleForTesting constructor for setting required (private final) fields to mocks.

(I like your idea of setting up the ExoPlayer then passing it in, but after setup there's really no remaining logic for a secondary constructor to handle.)

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
KyleFin: Ack. I'll fetch latest changes and split the PR when the other content is approved. Thanks!

stuartmorgan:
@KyleFin Are you planning on addressing the latest comments?

packages/video_player/video_player/example/integration_test/video_player_test.dart
KyleFin: I ended up not going this route.

packages/video_player/video_player/example/integration_test/video_player_test.dart
KyleFin: Done. Added Dart and Java unit tests. (Reverted this integration test)

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
KyleFin: (Done in https://github.com/flutter/plugins/pull/3820/commits/5f95287a54b82ca90cde6153285ab9e9b10e2493)

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
KyleFin: (Done in https://github.com/flutter/plugins/pull/3820/commits/5f95287a54b82ca90cde6153285ab9e9b10e2493)

KyleFin:
Sorry for the miscommunication:
I addressed your Sep 16 comments by implementing the changes I suggested in my Sep 23 comments as https://github.com/flutter/plugins/pull/3820/commits/5f95287a54b82ca90cde6153285ab9e9b10e2493

If you agree with those changes, this is ready to submit (aside from splitting out the changes for federated plugins https://github.com/flutter/plugins/pull/3820#discussion_r710225750)

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -96,6 +97,18 @@
     setupVideoPlayer(eventChannel, textureEntry);
```
stuartmorgan, line 97: Why isn't this needed in the other constructor? It seems like it should be.

This is why having two unrelated constructor paths tends to be problematic, and why I'd rather see this constructor call the other constructor.

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -282,6 +296,13 @@ private void sendInitialized() {
         }
         event.put("width", width);
         event.put("height", height);
+
+        // Rotating the video with ExoPlayer does not seem to be possible with a Surface,
+        // so inform the Flutter code that the widget needs to be rotated to prevent
+        // upside-down playback.
```
stuartmorgan, line 302: Could you add a comment here saying why this only adds a correction for 180?

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -661,7 +668,10 @@ class _VideoPlayerState extends State<VideoPlayer> {
   Widget build(BuildContext context) {
     return _textureId == VideoPlayerController.kUninitializedTextureId
         ? Container()
-        : _videoPlayerPlatform.buildView(_textureId);
+        : Transform.rotate(
```
stuartmorgan, line 671: We should probably make a helper that only wraps with a `Transform` if there's a non-zero rotation.

stuartmorgan (changes requested):

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
KyleFin: This new constructor is only needed for creating a mock VideoPlayer in tests. We don't need any additional set up, but these private final fields are required to be set.  (See tests in VideoPlayerTest.java)

Added a docstring about this constructor being used for mocks.

packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
KyleFin: Done.

packages/video_player/video_player/lib/video_player.dart
KyleFin: Agreed. Done.

KyleFin:
I believe remaining check failures should be resolved by [splitting the platform interface changes](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins) into a separate PR

packages/video_player/video_player/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java
```diff
@@ -4,12 +4,72 @@
...
+        new VideoPlayer(
+            mock(EventChannel.class),
+            mock(TextureRegistry.SurfaceTextureEntry.class),
+            mock(VideoPlayerOptions.class),
+            mockExoPlayer);
+    QueuingEventSink mockEventSink = mock(QueuingEventSink.class);
+    player.eventSink = mockEventSink;
+    player.isInitialized = true;
+
+    player.sendInitialized();
```
stuartmorgan, line 43: Why do you need to use this internal method? You're mocking the exoplayer that has a listener set on it, which means you should be able to capture the listener and invoke the correct state change on it to drive this flow rather than exposing several pieces of internals. That would make the test much less fragile.

stuartmorgan (changes requested):
Sorry, one more comment on the tests.

This is close enough that you could go ahead and make the platform interface PR though.

KyleFin:
https://github.com/flutter/plugins/pull/4634 is the platform interface PR (ready for review).

KyleFin:
@stuartmorgan
tl;dr - I reverted the Java unit tests (Commit https://github.com/flutter/plugins/pull/3820/commits/04555284b05a610960559d99241a2a054d8bc9c4), and I think this is ready to merge.

Scrolling through comments on the [original issue](https://github.com/flutter/flutter/issues/60327), it seems there are a number of people (and their [customers](https://github.com/flutter/flutter/issues/60327#issuecomment-1022986875)) who would really appreciate having this fix merged.

______

More details:

I was _really_ hoping to test the Java code, but at this point that seems infeasible (as far as I can tell, the only Java test for the rest of this plugin is the [placeholder initializer test](https://github.com/flutter/plugins/blob/main/packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java) that doesn't actually do anything. So I suppose not having tests for this PR won't be any less tested than the rest of the plugin 😞 ).

In those reverted tests I was hoping to have a simple unit test to check that the rotation gets added to the event channel when `sendInitialized()` is called, but as [you said](https://github.com/flutter/plugins/pull/3820#discussion_r771986067) that requires exposing a lot of class internals.

I think making this Java class more testable would require refactoring well beyond the scope of this PR. The [existing constructor](https://github.com/flutter/plugins/blob/7d1923f4528e416529f0c5d1f02a706c9adad45b/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java#L62) does not lend itself to injecting mocks for tests because it requires many realistic dependencies (see all its parameters) that interact in complex ways. I tried creating a new constructor that would allow me to inject a mock ExoPlayer, but even then, [`setupVideoPlayer`](https://github.com/flutter/plugins/blob/7d1923f4528e416529f0c5d1f02a706c9adad45b/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java#L156) requires very realistic dependencies too so I don't think your suggestion to [capture the listener](https://github.com/flutter/plugins/pull/3820#discussion_r771986067) is feasible without significant refactoring.

Before attempting unit tests, my [original thought](https://github.com/flutter/plugins/pull/3820#discussion_r659058667) was to write an integration test (so I wouldn't have to expose private sendInitialized method where logic is implemented) to test the end-to-end flow of loading a video file and confirming rotation is applied. That doesn't seem feasible to have an integration test that runs only on Android in CI (and again, I don't see that there are any existing tests).

It feels like defeat to not add tests, but at this point I don't know what else to do. I know you have a lot on your plate too. After waiting weeks to months between reviews it's frustrating for me to find a few hours in my "spare time" to wrestle with merge conflicts and environment set up and trying to remember what I was trying to do 10 months ago.

**I'd really appreciate if you could merge this PR or give me a very specific outline of what else I need to finish in a final push to get this over the finish line.**

Thanks for your help on this PR and in making Flutter more awesome in general! :)

packages/video_player/video_player/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java
KyleFin: I reverted the tests. (See https://github.com/flutter/plugins/pull/3820#issuecomment-1086419754)

KyleFin:
@stuartmorgan, 
I created https://github.com/flutter/plugins/pull/5158 for the video_player_android package changes. Once that is merged, I believe this PR can be rebased and the iOS CI errors should be resolved:
`Because every version of video_player from path depends on video_player_android ^2.3.3 which doesn't match any versions, video_player from path is forbidden.
So, because video_player_example depends on video_player from path, version solving failed.`

I recorded sample videos and confirmed that those recorded in landscapeLeft or landscapeRight both send the correct rotationCorrection value and display upright in my simulator using the changes in this PR:

Screen Shot 2022-04-02 at 12 46 26 AM
Screen Shot 2022-04-02 at 12 50 38 AM

stuartmorgan:
> as far as I can tell, the only Java test for the rest of this plugin is the [placeholder initializer test](https://github.com/flutter/plugins/blob/main/packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java) that doesn't actually do anything.

That's correct; many of our plugins had no native unit test harnesses set up, which made adding a native unit test extremely difficult; that no-op is a result of a sweep we did over all plugins to ensure that they had the harnesses in place to allow adding a native unit test without having to also learn how to configure the build to actually run it.

> So I suppose not having tests for this PR won't be any less tested than the rest of the plugin 😞 

This is addressed in https://github.com/flutter/flutter/wiki/Plugin-Tests#faq-do-i-need-to-add-tests-even-if-the-part-of-the-code-im-changing-isnt-already-tested

> I tried creating a new constructor that would allow me to inject a mock ExoPlayer, but even then, [`setupVideoPlayer`](https://github.com/flutter/plugins/blob/7d1923f4528e416529f0c5d1f02a706c9adad45b/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java#L156) requires very realistic dependencies too so I don't think your suggestion to [capture the listener](https://github.com/flutter/plugins/pull/3820#discussion_r771986067) is feasible without significant refactoring.

Can you elaborate on what you mean by "requires very realistic dependencies"? What was hard to handle with mocks there? Depending on the scope of the issue, there may be small adjustments we could make to make it work better. (E.g., when I added unit tests to a new area of `camera` I introduced a trivial wrapper around one problematic class to make it much easier to mock; something like that might be useful here.)

> Before attempting unit tests, my [original thought](https://github.com/flutter/plugins/pull/3820#discussion_r659058667) was to write an integration test (so I wouldn't have to expose private sendInitialized method where logic is implemented) to test the end-to-end flow of loading a video file and confirming rotation is applied. That doesn't seem feasible to have an integration test that runs only on Android in CI (and again, I don't see that there are any existing tests).

Integration tests that run only on Android are definitely possible, and we do have existing tests. See https://github.com/flutter/plugins/blob/main/packages/video_player/video_player_android/example/integration_test/video_player_test.dart

> **I'd really appreciate if you could merge this PR or give me a very specific outline of what else I need to finish in a final push to get this over the finish line.**

It needs a test that would fail in CI if your changes to `VideoPlayer.java` were reverted. I'd prefer if that were a native unit test since they are much faster and more reliable (and I'm happy to advise on how we could do that if you can provide a bit more info about what was hard to do with injected mocks, especially since making it easier to use mocks will be useful in the future as well), but if you can write a reasonable integration test from the Dart side instead that's fine as well.

stuartmorgan:
@camsim99 Any chance you'd have time to look at adding Java unit tests to this? See https://github.com/flutter/plugins/pull/3820#issuecomment-1086419754 and https://github.com/flutter/plugins/pull/3820#issuecomment-1087969806 for context. This PR is very close, so it would be great if we could do whatever testability changes in the plugin structure are necessary to get this testable.

Chetan-Goyal:
I am facing issues with recorded video in latest version of `camera` plugin and this PR while testing `landscapeRight`.
https://user-images.githubusercontent.com/55137330/164749318-87eecb79-1698-4838-8651-47930c007bb5.mp4

camsim99:
> @camsim99 Any chance you'd have time to look at adding Java unit tests to this? See [#3820 (comment)](https://github.com/flutter/plugins/pull/3820#issuecomment-1086419754) and [#3820 (comment)](https://github.com/flutter/plugins/pull/3820#issuecomment-1087969806) for context. This PR is very close, so it would be great if we could do whatever testability changes in the plugin structure are necessary to get this testable.

@stuartmorgan @KyleFin I added a couple of unit tests to test the changes and did a bit of refactoring to make the tests simpler. Please take a look when you get the chance!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -231,81 +248,74 @@ private static void setAudioAttributes(ExoPlayer exoPlayer, boolean isMixMode) {
...
         event.put("width", width);
         event.put("height", height);
-
-        // Rotating the video with ExoPlayer does not seem to be possible with a Surface,
-        // so inform the Flutter code that the widget needs to be rotated to prevent
-        // upside-down playback for videos with rotationDegrees of 180 (other orientations work
-        // correctly without correction).
-        if (rotationDegrees == 180) {
-          event.put("rotationCorrection", rotationDegrees);
-        }
```
KyleFin: This is the new code that is being added to fix https://github.com/flutter/flutter/issues/60327.

The purpose of adding tests in this PR is to verify that `rotationCorrection` is correctly set to 180 when `rotationDegrees == 180` and unspecified for other cases.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -90,10 +92,24 @@ final class VideoPlayer {
...
     MediaSource mediaSource = buildMediaSource(uri, dataSourceFactory, formatHint, context);
+
     exoPlayer.setMediaSource(mediaSource);
     exoPlayer.prepare();
 
-    setupVideoPlayer(eventChannel, textureEntry);
+    setupVideoPlayer(exoPlayer, new QueuingEventSink());
+  }
+
+  VideoPlayer(
```
KyleFin, line 102: I suppose this constructor could be used in non-test code, but it might be helpful to mark it as VisibleForTesting and/or add a comment explaining this constructor's purpose (why would I choose one constructor over the other?)

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java
```diff
@@ -4,12 +4,107 @@
...
+    assertEquals(event.get("height"), 200);
+
+    testFormat = new Format.Builder().setWidth(100).setHeight(200).setRotationDegrees(180).build();
+
+    videoPlayer.sendInitialized();
+
+    assertEquals(event.get("event"), "initialized");
+    assertEquals(event.get("duration"), 10L);
+    assertEquals(event.get("width"), 100);
+    assertEquals(event.get("height"), 200);
```
KyleFin, line 108: Here we should also assert `rotationCorrection == 180`. In other cases, we could assert that `rotationCorrection` is not set.

KyleFin (commented):
> > @camsim99 Any chance you'd have time to look at adding Java unit tests to this? See [#3820 (comment)](https://github.com/flutter/plugins/pull/3820#issuecomment-1086419754) and [#3820 (comment)](https://github.com/flutter/plugins/pull/3820#issuecomment-1087969806) for context. This PR is very close, so it would be great if we could do whatever testability changes in the plugin structure are necessary to get this testable.
> 
> @stuartmorgan @KyleFin I added a couple of unit tests to test the changes and did a bit of refactoring to make the tests simpler. Please take a look when you get the chance!

Thanks so much @camsim99 !! I added a few comments.

If I correctly understand the package dependencies and Cirrus CI results, your Java test code will need to be added in https://github.com/flutter/plugins/pull/5158 and merged. Then the remainder of the changes in this PR will be ready to merge after rebasing. Thanks again!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -45,17 +46,17 @@ final class VideoPlayer {
   private static final String FORMAT_HLS = "hls";
   private static final String FORMAT_OTHER = "other";
 
-  private ExoPlayer exoPlayer;
+  private ExoPlayer mExoPlayer;
```
stuartmorgan, line 49: This change violates Google style:

> In Google Style, special prefixes or suffixes are not used. For example, these names are not Google Style: `name_`, `mName`, `s_name` and `kName`.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -45,17 +46,17 @@ final class VideoPlayer {
...
 
   private Surface surface;
 
   private final TextureRegistry.SurfaceTextureEntry textureEntry;
 
-  private QueuingEventSink eventSink = new QueuingEventSink();
+  private QueuingEventSink mEventSink;
```
stuartmorgan, line 55: Same.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -153,26 +168,28 @@ private MediaSource buildMediaSource(
     }
   }
 
-  private void setupVideoPlayer(
-      EventChannel eventChannel, TextureRegistry.SurfaceTextureEntry textureEntry) {
+  private void setupVideoPlayer(ExoPlayer exoPlayer, QueuingEventSink eventSink) {
```
stuartmorgan, line 171: While you're touching this, could you fix the name to `setUpVideoPlayer`? (This mistake is widespread in the plugin code, unfortunately; I'm trying to fix it incrementally whenever we find it.)

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java
```diff
@@ -4,12 +4,107 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.MockitoAnnotations;
+import org.robolectric.RobolectricTestRunner;
 
+@RunWith(RobolectricTestRunner.class)
 public class VideoPlayerTest {
-  // This is only a placeholder test and doesn't actually initialize the plugin.
+  private VideoPlayer videoPlayer;
```
stuartmorgan, line 27: I very strongly prefer (across all languages) that we not make the object under test part of the test fixture. It makes it harder to understand what the setup for the test is, and also doesn't scale as soon as there's a case where a test needs to set the object up slightly differently. It's better to just have the boilerplate of the object creating in each test to make it more locally understandable.

(Setting up the mocks here is fine.)

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
camsim99: @KyleFin @stuartmorgan Ah, I apologize for misunderstanding the changes! I'll address the comments, add tests for this, and then re-requests your reviews when finished.

camsim99: @KyleFin I can't request your review, but please take a look when you get the chance!

KyleFin: This code to set `rotationCorrection` still needs to be re-added (along with the comment explaining it), but otherwise LGTM! Thanks again!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -264,7 +283,7 @@ long getPosition() {
   }
 
   @SuppressWarnings("SuspiciousNameCombination")
-  private void sendInitialized() {
+  void sendInitialized() {
```
KyleFin, line 286: VisibleForTesting?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
KyleFin: LGTM!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -214,6 +230,9 @@ public void onPlayerError(final PlaybackException error) {
             }
           }
         });
+
+    this.exoPlayer = exoPlayer;
+    this.eventSink = eventSink;
```
stuartmorgan, line 235: Should we do these at the start of the method, in case `setUpVideoPlayer` ever calls a method that uses the members?

stuartmorgan (commented):
Looks good (with one question)! Let's go ahead and move the changes over to the Android PR.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
camsim99: That makes sense to me, but the reason I didn't do that is because I would have to specify the `this` for each usage of these in this method. If that's not a problem, then I'll change it!

stuartmorgan: Why would we need to change the other uses? Wouldn't that only be true if we were assigning new values to the arguments (which is generally not a good idea anyway from a readability perspective)?

camsim99: Oh, I see. That was a misunderstanding on my part.

stuartmorgan:
@camsim99 Can you go ahead and copy the tests over to https://github.com/flutter/plugins/pull/5158 so we can get that reviewed and landed?

stuartmorgan:
This should be ready to rebase and land, I believe?

KyleFin:
Done! @stuartmorgan this should be ready to merge once the submit-queue turns green. Thanks for all the reviews and huge thanks to @camsim99 for adding the Java tests!

I made one small change to use a _VideoPlayerWithRotation widget instead of helper method ([rationale](https://youtu.be/IOyq-eTRhvo?t=92)).

I confirmed the [original video sample](https://github.com/flutter/flutter/issues/60327#issuecomment-714729881) correctly passes the rotationCorrection value and rotates the video.
Screen Shot 2022-06-02 at 4 53 32 PM

stuartmorgan:
@camsim99 Do you want to do the second review here since you have a lot of context on the overall change?

stuartmorgan:
@KyleFin Thanks for the infinite patience getting this one completed, and thanks again @camsim99 for getting the testing set up!

PR #3821, opened 2021-04-22
[in_app_purchase] platform interface improvement

cyanglaz:
Adds some improvements to the platform interface. 

1. Make `InAppPurchasePlatformAddition` a subclass of `PlatformInterface` to take advantage of the existing token checking logic. (I also consider `InAppPurchasePlatformAddition` a `PlatformInterface` because it works in a similar way.
2. Make the `instance` variable `late` as we should never access it before setter.
3. Add tests for `InAppPurchasePlatformAddition`

part of flutter/flutter#78525

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
```diff
@@ -2,18 +2,34 @@
...
 /// The interface that platform implementations must implement when they want to
 /// provide platform specific in_app_purchase features.
-abstract class InAppPurchasePlatformAddition {
+///
+/// Platform implementations should extend this class rather than implement it as `in_app_purchase`
+/// does not consider newly added methods to be breaking changes. Extending this class
+/// (using `extends`) ensures that the subclass will get the default implementation, while
+/// platform implementations that `implements` this interface will be broken by newly added
+/// [InAppPurchasePlatformAddition] methods.
+abstract class InAppPurchasePlatformAddition extends PlatformInterface {
```
cyanglaz, line 16: @stuartmorgan 

I'm making the `InAppPurchasePlatformAddition` also a `PlatformInterface`.
We didn't talk about this offline but I think this makes sense. Please share your thoughts.

mvanbeusekom (approved):
LGTM, only thing I am thinking about is adding another test to verify that accessing `instance` throws a `LateInitializationError` when accessing it before it is initialized.

cyanglaz:
> LateInitializationError

Sounds good, weird tho the code wasn't able to find `LateInitializationError`

mvanbeusekom:
> > LateInitializationError
> 
> 
> 
> Sounds good, weird tho the code wasn't able to find `LateInitializationError`

I tested it on dartpad.dev since I wasn't sure what would happen ;)

cyanglaz:
@mvanbeusekom Seems like the class `LateInitializationError ` isn't accessible. I think it's ok to not test a default NNBD behavior.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform.dart
```diff
@@ -24,15 +24,16 @@ abstract class InAppPurchasePlatform extends PlatformInterface {
   /// The instance of [InAppPurchasePlatform] to use.
   ///
   /// Defaults to `null`.
```
stuartmorgan, line 26: This is no longer correct. Instead, you should document that it must be set before using.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform.dart
```diff
@@ -24,15 +24,16 @@ abstract class InAppPurchasePlatform extends PlatformInterface {
...
-  static InAppPurchasePlatform? get instance => _instance;
+  static InAppPurchasePlatform get instance => _instance;
 
-  static InAppPurchasePlatform? _instance;
+  // Should only be accessed after setter is called.
+  static late InAppPurchasePlatform _instance;
```
stuartmorgan, line 30: This should be either above or below the getter and setter, rather than between them.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
```diff
@@ -2,18 +2,34 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// ignore: avoid_classes_with_only_static_members
+import 'package:in_app_purchase_platform_interface/in_app_purchase_platform_interface.dart';
+import 'package:plugin_platform_interface/plugin_platform_interface.dart';
+
 /// The interface that platform implementations must implement when they want to
 /// provide platform specific in_app_purchase features.
```
stuartmorgan, line 9: Nit: platform-specific

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
stuartmorgan: What's the advantage of doing this? It is not at all clear to me that if we added new methods to `PlatformInterface` they would be applicable to these addition classes.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
```diff
@@ -36,5 +52,13 @@ abstract class InAppPurchasePlatformAddition {
   ///   }
   /// }
   /// ```
-  static InAppPurchasePlatformAddition? instance;
+  static InAppPurchasePlatformAddition get instance => _instance;
```
stuartmorgan, line 55: Why have you made this non-nullable? Why should a platform with no additions need to set this up?

stuartmorgan (changes requested):

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform.dart
cyanglaz: done

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform.dart
cyanglaz: done

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
cyanglaz: done. Also fixed some other appearance in the plugin.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
cyanglaz: This is mainly to take advantage of using the token verification of the `PlatformInterface`.
I couldn't think of a scenario where we need to add some methods in the "PlatformAddition" that will not be applicable to all addition classes. 

Also, Now thinking about the usage of other "PlatformInterface"s, I'm not confident with `InAppPurchasePlatformAddition` being a "PlatformInterface". Maybe for this plugin i'll just duplicate the token verification code instead of extending `PlatformInterface`.

WDYT?

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
cyanglaz: Ah makes sense! Made this nullable.

cyanglaz (commented):
@stuartmorgan Updated per your comments. PTAL :)

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
mvanbeusekom: Isn't the `verifyToken` meant to enforce users `extend` the class instead of `implement`ing it?

Is this a concern for the `InAppPurchasePlatformAddition` class?

I don't think we will need to be afraid of the `InAppPurchasePlatformAddition` class to have any methods. Meaning I don't expect any methods to be added in the future that would break implementing classes.

cyanglaz: We do need some sort of enforcement to make sure the addition is not a type of `InAppPurchasePlatform`.

Make sure an addition class to extend `InAppPurchasePlatformAddition` is one of the solution.

mvanbeusekom: Check, understood

cyanglaz: Extending happens to be the first easy solution I can think of; If there's an easier solution, we absolutely should adopt it!

stuartmorgan: > We do need some sort of enforcement to make sure the addition is not a type of `InAppPurchasePlatform`.

Why do we need to enforce that? The concern is purely around pubspec dependency cycles with endorsement IIRC, in which case it's the pubspec cycle, not the classes, that are the issue. This seems like a very indirect way of trying to enforce a lack of cycles (and is both not sufficient to prevent a cycle, and not necessary in the case of an in-package implementation if this is something we adopt as a general solution).

cyanglaz: If the addition is a type of `InAppPurchasePlatform`, the plugin user would easily make a mistake to access the `InAppPurchasePlatform` object (e.g. InAppPurchasePlatformAndroid) and call the unified API there.

stuartmorgan: Sorry, I'm not following. Can you give a snippet of what that mistake would look like in code?

cyanglaz: Let's say we allow an `InAppPurchasePlatform` to also be a type of `InAppPurchasePlatformAddition`, the implementer could potentially create a class like this:

```dart
class InAppPurchasePlatformGooglePlay extends InAppPurchasePlatform implements InAppPurchasePlatformAddition {
   
    @override
    Future<void> buyConsumable(...){...}

    Future<void> someAdditionMethod(...){...}
}
```

Then the user could potentially use this class to do something like:

```dart
InAppPurchasePlatformGooglePlay googlePlayAddition = InAppPurchaseConnection.instance.getPlatformAddition<InAppPurchasePlatformGooglePlay>();
googlePlayAddition.buyConsumable(...);
```

In this case, the user is accessing the platform interface API directly, which we want to avoid.

stuartmorgan: I don't think we need to go out of our way to forbid that construction. It's a bad idea, but only in the same way that, say, making a bunch of implementation details of a `InAppPurchaseGoogleplayAdditions` public instead of private would be. I.e., it's the implementor of the addition class designing their own extension API badly.

What I wanted to avoid was the design where we *forced* extensions to use that, so that every extension would be required to have that poor design. If we make it so that it's easy to do the right thing, and model the right thing in our plugins, that should be enough.

cyanglaz (commented):
@stuartmorgan Updated with review. comments about not making `InAppPurchasePlatformAddition` a platform interface.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
```diff
@@ -2,18 +2,33 @@
...
 
+import 'package:in_app_purchase_platform_interface/in_app_purchase_platform_interface.dart';
+
 // ignore: avoid_classes_with_only_static_members
 /// The interface that platform implementations must implement when they want to
-/// provide platform specific in_app_purchase features.
+/// provide platform-specific in_app_purchase features.
+///
+/// Platforms that wants to introduce platform-specific public APIs should create
+/// a class that either extend or implements [InAppPurchasePlatformAddition]. Then replace
```
stuartmorgan, line 12: s/replace ... with/set ... to/ (by default there's no instance, so there's nothing to replace)

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
```diff
@@ -2,18 +2,33 @@
...
+/// the [InAppPurchasePlatformAddition.instance] with an instance of that class.
+///
+/// All the APIs added by [InAppPurchasePlatformAddition] implementers will be accessed from
```
stuartmorgan, line 15: s/implementers/implementations/

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
```diff
@@ -2,18 +2,33 @@
...
+/// [InAppPurchasePlatformAdditionProvider.getPlatformAddition] by the client APPs.
+/// To avoid client APPs to directly call the [InAppPurchasePlatform] APIs, an [InAppPurchasePlatformAddition] implementation must not
```
stuartmorgan, line 17: "To avoid clients directly calling [InAppPurchasePlatform] APIs, an [InAppPurchasePlatformAddition] implementation should not be"

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
```diff
@@ -36,5 +51,13 @@ abstract class InAppPurchasePlatformAddition {
   ///   }
   /// }
   /// ```
-  static InAppPurchasePlatformAddition? instance;
+  static InAppPurchasePlatformAddition? get instance => _instance;
+
+  /// Set the instance to a desired [InAppPurchasePlatformAddition] implementation.
```
stuartmorgan, line 56: Sets

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
```diff
@@ -36,5 +51,13 @@ abstract class InAppPurchasePlatformAddition {
...
+  ///
+  /// The `instance` must not be a type of [InAppPurchasePlatform].
```
stuartmorgan, line 58: s/must/should/

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/types/product_details_response.dart
```diff
@@ -18,7 +18,7 @@ class ProductDetailsResponse {
 
   /// The list of identifiers that are in the `identifiers` of [InAppPurchasePlatform.queryProductDetails] but failed to be fetched.
   ///
-  /// There's multiple platform specific reasons that product information could fail to be fetched,
+  /// There's multiple platform-specific reasons that product information could fail to be fetched,
```
stuartmorgan, line 21: s/There's/There are/

packages/in_app_purchase/in_app_purchase_platform_interface/test/in_app_purchase_platform_test.dart
```diff
@@ -143,3 +188,36 @@ class ExtendsInAppPurchasePlatform extends InAppPurchasePlatform {}
...
+class ImplementsInAppPurchasePlatformAddition
+    implements InAppPurchasePlatformAddition {
+  @override
+  dynamic noSuchMethod(Invocation invocation) => super.noSuchMethod(invocation);
+}
+
+class MockInAppPurchasePlatformAddition extends Mock
+    with
+        // ignore: prefer_mixin
+        MockPlatformInterfaceMixin
```
stuartmorgan, line 201: This shouldn't still be needed, right?

stuartmorgan (approved):
LGTM with nits.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
cyanglaz: done

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
cyanglaz: done

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
cyanglaz: done

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
cyanglaz: done

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/in_app_purchase_platform_addition.dart
cyanglaz: done

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/types/product_details_response.dart
cyanglaz: done

packages/in_app_purchase/in_app_purchase_platform_interface/test/in_app_purchase_platform_test.dart
cyanglaz: done

mvanbeusekom (approved):
LGTM

PR #3822, opened 2021-04-22
Move all null safety packages' min dart sdk to 2.12.0

cyanglaz:
This is mainly to remove the warning when `pub publish`. We can skip publishing this change for now.

### DIFF
--- a/script/tool/lib/src/publish_check_command.dart
+++ b/script/tool/lib/src/publish_check_command.dart
@@ -22,7 +22,7 @@ class PublishCheckCommand extends PluginCommand {
     argParser.addFlag(
       _allowPrereleaseFlag,
       help: 'Allows the pre-release SDK warning to pass.\n'
-          'When enabled, a pub warning, which asks to publish the package as a pre-release version when'
+          'When enabled, a pub warning, which asks to publish the package as a pre-release version when '
           'the SDK constraint is a pre-release version, is ignored.',
       defaultsTo: false,
     );
--- a/script/tool/lib/src/publish_check_command.dart
+++ b/script/tool/lib/src/publish_check_command.dart
@@ -74,37 +74,8 @@ class PublishCheckCommand extends PluginCommand {
       workingDirectory: package,
     );
 
-    final StringBuffer outputBuffer = StringBuffer();
 
-    final Completer<void> stdOutCompleter = Completer<void>();
-    process.stdout.listen(
-      (List<int> event) {
-        io.stdout.add(event);
-        outputBuffer.write(String.fromCharCodes(event));
-      },
-      onDone: () => stdOutCompleter.complete(),
-    );
-
-    final Completer<void> stdInCompleter = Completer<void>();
-    process.stderr.listen(
-      (List<int> event) {
-        io.stderr.add(event);
-        outputBuffer.write(String.fromCharCodes(event));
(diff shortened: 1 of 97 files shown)


stuartmorgan (commented):
Looks good, but can you also update the publish check script to no longer allow this warning to pass?

ditman:
Ssssso many packages! :P

cyanglaz:
> Looks good, but can you also update the publish check script to no longer allow this warning to pass?

Good point! done!

stuartmorgan:
Needs test updates for the tool behavioral change :)

cyanglaz:
I think maybe we can add a flag to keep the feature that ignores pre-release check, for possible feature release requirement.

script/tool/lib/src/publish_check_command.dart
```diff
@@ -18,7 +18,17 @@ class PublishCheckCommand extends PluginCommand {
     Directory packagesDir,
     FileSystem fileSystem, {
     ProcessRunner processRunner = const ProcessRunner(),
-  }) : super(packagesDir, fileSystem, processRunner: processRunner);
+  }) : super(packagesDir, fileSystem, processRunner: processRunner) {
+    argParser.addFlag(
+      _allowPrereleaseFlag,
+      help: 'Allows the pre-release SDK warning to pass.'
+          'When enabled, a pub warning, which asks to publish the package as a pre-release version when'
```
stuartmorgan, line 25: You're missing spaces at the ends of these string literals, so they won't concatenate correctly.

script/tool/test/publish_check_command_test.dart
```diff
@@ -90,7 +90,7 @@ void main() {
           throwsA(isA<ToolExit>()));
     });
 
-    test('pass on prerelease', () async {
+    test('pass on prerelease is --allow-pre-release flag is on', () async {
```
stuartmorgan, line 93: Nit: s/is/if/

script/tool/test/publish_check_command_test.dart
```diff
@@ -105,7 +105,8 @@ void main() {
 
       processRunner.processesToReturn.add(process);
 
-      expect(runner.run(<String>['publish-check']), completes);
+      expect(runner.run(<String>['publish-check', '--allow-pre-release']),
+          completes);
     });
```
stuartmorgan, line 110: We also need a new test that the same setup fails without `--allow-pre-release`

stuartmorgan (changes requested):
Keeping it behind a flag SGTM.

script/tool/test/publish_check_command_test.dart
cyanglaz: done

script/tool/test/publish_check_command_test.dart
cyanglaz: done

cyanglaz (commented):
@stuartmorgan Fixed

script/tool/lib/src/publish_check_command.dart
stuartmorgan: The second part still has the same problem; this will print "[...] a pre-release verison whenthe SDK [...]"

stuartmorgan (approved):
LGTM with nit

PR #3824, opened 2021-04-23
Update integration_test README 

johnpryan:
Changes "deprecated" to "moved" to avoid confusion, since this is still the recommended way to write unit tests, but needs to be imported differently. Also adds a link to the flutter.dev docs + the new location.

FYI @sfshaza2

### DIFF
--- a/packages/integration_test/README.md
+++ b/packages/integration_test/README.md
@@ -1,15 +1,19 @@
-# integration_test (deprecated)
+# integration_test (moved)
 
-## DEPRECATED
+## MOVED
 
-This package has been moved to the Flutter SDK. Starting with Flutter 2.0,
-it should be included as:
+This package has been [moved to the Flutter
+SDK](https://github.com/flutter/flutter/tree/master/packages/integration_test).
+Starting with Flutter 2.0, it should be included as:
 
 ```
 dev_dependencies:
   integration_test:
     sdk: flutter
 ```
+
+For the latest documentation, see [Integration
+testing](https://flutter.dev/docs/testing/integration-tests).
 
 ## Old instructions
 


packages/integration_test/README.md
```diff
@@ -1,16 +1,20 @@
-# integration_test (deprecated)
+# integration_test (moved)
 
-## DEPRECATED
+## MOVED
 
-This package has been moved to the Flutter SDK. Starting with Flutter 2.0,
-it should be included as:
+This package has been [moved to the Flutter
```
stuartmorgan, line 5: I'm not convinced this is clearer; "moved" isn't generally a meaningful term for clients of a package (since where the source lives is generally irrelevant). The version on pub, where users are likely to be reading this, is deprecated.

johnpryan: The version on Pub is deprecated, but the package isn't. We don't want to give the impression that integration testing should be done some other way. @sfshaza2 do you have any input here?

stuartmorgan (approved):
Looks like we never got resolution on this. I'm fine to go ahead and land it, but you'll need to merge in the latest master so that the tests are green (re-running the flake will fail since this is so old.)

packages/integration_test/README.md
```diff
@@ -1,16 +1,20 @@
...
+SDK](https://github.com/flutter/flutter/tree/master/packages/integration_test).
+Starting with Flutter 2.0, it should be included as:
```
sfshaza2, line 7:
```suggestion
As of Flutter 2.0, include it in your pubspec's
dev dependencies section, as follows:
```

packages/integration_test/README.md
sfshaza2, line 5:
```suggestion
This package has [moved to the Flutter
```

packages/integration_test/README.md
```diff
@@ -1,16 +1,20 @@
-# integration_test (deprecated)
+# integration_test (moved)
 
-## DEPRECATED
+## MOVED
 
-This package has been moved to the Flutter SDK. Starting with Flutter 2.0,
-it should be included as:
+This package has been [moved to the Flutter
+SDK](https://github.com/flutter/flutter/tree/master/packages/integration_test).
```
sfshaza2, line 6:
```suggestion
SDK](https://github.com/flutter/flutter/tree/master/packages/integration_test),
and the pub.dev version is deprecated.
```

sfshaza2 (commented):
I've made a few suggested tweaks. I never saw this back in April!

packages/integration_test/README.md
sfshaza2: I made some suggested changes to the text.

PR #3829, opened 2021-04-27
[google_sign_in] Fix issue obtaining serverAuthCode on Android and add forceCodeForRefreshToken parameter // Update

dreenot:
This PR is a follow to #3356. It merges into the master keeping the master's preferences. It has been a long time since it was requested and the code has significantly changed. A few changes were needed to make it work with the current master. Therefore, I'm just updating #3356. The Pre-launch Checklist belongs to this PR, but the following info is a reference.

tl;dr `serverAuthCode` is returning the right value, not null. `forceCodeForRefreshToken` is now an option, it wasn't available before

To test it, sign in and get the `serverAuthCode`
`final user = await googleSignIn.signIn();`
`final googleAuth = await user.authentication;`
`final serverAuthCode = googleAuth.serverAuthCode`

To test `forceCodeForRefreshToken` update `values/bools.xml`
```
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <bool name="force_code_for_refresh_token">false</bool>
</resources>
```
And refresh the sign in

## Pre-launch Checklist

- [x] I signed the [CLA].

## Description
This is a follow up to #2116 to fix the following issues:

- `serverAuthCode` is always null on Android because it is not passed to the `GoogleSignInAuthentication` object as that is created by the `getTokens` call rather than from the data obtained during the initial `signIn` call (which does contain a `serverAuthCode`, it's just not accessible anywhere that I can tell)

- A refresh token is only given to the server on the first sign in without the ability to set the `forceCodeForRefreshToken` parameter

## Related Issues

- flutter/flutter#17813

- flutter/flutter#14245

## Breaking Change

- [x] No, this is not a breaking change.

## Festivus Note
As the season of Festivus is upon us, I do have a grievance to air from the last time I submitted a PR related to this plugin. I was told on #879 that my pull request to add the serverAuthCode (which worked on android, as a side note) would not be accepted without E2E tests using a test harness that wasn't even ready yet. You can imagine my surprise when #2116 was merged a few months later without them (no disrespect to the author of that PR, they did what was asked of them).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -65,3 +65,5 @@ Anton Borries <mail@antonborri.es>
 Alex Li <google@alexv525.com>
 Rahul Raj <64.rahulraj@gmail.com>
 Daniel Roek <daniel.roek@gmail.com>
+Twin Sun, LLC <google-contrib@twinsunsolutions.com>
+Andrew Cardinot <andrewcardinot@gmail.com>
--- a/packages/google_sign_in/google_sign_in/CHANGELOG.md
+++ b/packages/google_sign_in/google_sign_in/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.0.3
+
+* Add `force_code_for_refresh_token` option for Android, returns the right serverAuthCode
+
 ## 5.0.2
 
 * Fix flutter/flutter#48602 iOS flow shows account selection, if user is signed in to Google on the device.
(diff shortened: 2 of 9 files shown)


stuartmorgan:
Given that the original PR has been updated, it doesn't seem like there's a need for this second copy; closing. If you disagree, please leave a comment explaining why and we can revisit.

PR #3832, opened 2021-04-28
[in_app_purchase] Federated iOS implementation

mvanbeusekom:
This PR adds the iOS implementation of the new "in_app_purchase_platform_interface" (see PR #3781).

**Note:** this PR is marked as "Draft" since the following points are open:
- [x] The iOS specific functions `presentCodeRedemptionSheet` and `refreshPurchaseVerificationData` methods are currently not implemented as part of the `InAppPurchaseIosPlatform` class. These methods are not enforced by the `InAppPurchasePlatform` platform interface as they are iOS specific. These methods are currently accessible through the `SKPaymentQueueWrapper` and `SKReceiptManager` APIs. @cyanglaz I am wondering is if we need to also make them available through an `InAppPurchaseIosPlatformAddition` implementation?
 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/in_app_purchase/in_app_purchase_ios/lib/src/in_app_purchase_ios_platform.dart
```diff
@@ -0,0 +1,223 @@
...
+    _skPaymentQueueWrapper.setTransactionObserver(observer);
+    return _instance!;
+  }
+
+  @override
+  Future<bool> isAvailable() => SKPaymentQueueWrapper.canMakePayments();
+
+  @override
+  Future<bool> buyNonConsumable({required PurchaseParam purchaseParam}) async {
+    if (!(purchaseParam is AppStorePurchaseParam)) {
```
cyanglaz, line 76: Note here: this check is not necessary, the only thing missing from `PurchaseParam` is `simulatesAskToBuyInSandbox`, which is used only when testing. So we can actually make simulatesAskToBuyInSandbox defaults to `false`. so:
```dart
    await _skPaymentQueueWrapper.addPayment(SKPaymentWrapper(
        productIdentifier: purchaseParam.productDetails.id,
        quantity: 1,
        applicationUsername: purchaseParam.applicationUserName,
        simulatesAskToBuyInSandbox: (purchaseParam is AppStorePurchaseParam) ? purchaseParam.simulatesAskToBuyInSandbox: false
        requestData: null));
```
We can document it in the `AppStorePurchaseParam` how to test `simulatesAskToBuyInSandbox` in details, something like:

/// If you want to test `simulatesAskToBuyInSandbox`, you should create a `AppStorePurchaseParam` and use it in `buyNonConsumable` or `buyConsumable`.

packages/in_app_purchase/in_app_purchase_ios/lib/src/in_app_purchase_ios_platform.dart
```diff
@@ -0,0 +1,223 @@
...
+      {required PurchaseParam purchaseParam, bool autoConsume = true}) {
+    assert(autoConsume == true, 'On iOS, we should always auto consume');
+    return buyNonConsumable(purchaseParam: purchaseParam);
+  }
+
+  @override
+  Future<void> completePurchase(PurchaseDetails purchase) {
+    if (!(purchase is AppStorePurchaseDetails)) {
+      throw ArgumentError(
+        'On iOS, the `purchase` should always be of type `AppStorePurchaseDetails`.',
```
cyanglaz, line 103: Will this ever be a user error? `PurchaseDetails` is never generated by users right, it is given to them. So it should always be a type of `AppStorePurchaseDetails` unless there's a bug in our code, so assertion maybe the better approach here.

packages/in_app_purchase/in_app_purchase_ios/pubspec.yaml
```diff
@@ -0,0 +1,34 @@
...
+  
+dev_dependencies:
+  build_runner: ^1.11.1
+  json_serializable: ^4.1.1
+  flutter_test:
+    sdk: flutter
+
+environment:
+  sdk: ">=2.12.0 <3.0.0"
+  flutter: ">=1.20.0"
```
cyanglaz, line 34: new line EOF

packages/in_app_purchase/in_app_purchase_ios/pubspec.yaml
```diff
@@ -0,0 +1,34 @@
+name: in_app_purchase_ios
+description: An implementation for the iOS platform of the Flutter `in_app_purchase` plugin. This uses the iOS StoreKit Framework.
+repository: https://github.com/flutter/plugins/tree/master/packages/in_app_purchase/in_app_purchase_ios
+version: 1.0.0
```
cyanglaz, line 4: let's release this as 0.1.0 as the first version.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/types/purchase_details.dart
```diff
@@ -37,7 +38,7 @@ class PurchaseDetails {
   final String? transactionDate;
 
   /// The status that this [PurchaseDetails] is currently on.
-  PurchaseStatus? status;
+  PurchaseStatus status;
```
cyanglaz, line 41: this is added in a separate PR.

cyanglaz (changes requested):

packages/in_app_purchase/in_app_purchase_ios/ios/in_app_purchase.podspec
```diff
@@ -0,0 +1,27 @@
...
+  s.version          = '0.0.1'
+  s.summary          = 'Flutter In App Purchase'
+  s.description      = <<-DESC
+A Flutter plugin for in-app purchases. Exposes APIs for making in-app purchases through the App Store.
+Downloaded by pub (not CocoaPods).
+                       DESC
+  s.homepage         = 'https://github.com/flutter/plugins'
+  s.license          = { :type => 'BSD', :file => '../LICENSE' }
+  s.author           = { 'Flutter Dev Team' => 'flutter-dev@googlegroups.com' }
+  s.source           = { :http => 'https://github.com/flutter/plugins/tree/master/packages/in_app_purchase' }
```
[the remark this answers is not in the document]
mvanbeusekom: check, I guess the actual url should be:  https://github.com/flutter/plugins/tree/master/packages/in_app_purchase/in_app_purchase_ios/

packages/in_app_purchase/in_app_purchase_ios/ios/in_app_purchase.podspec
```diff
@@ -0,0 +1,27 @@
...
+  s.documentation_url = 'https://pub.dev/packages/in_app_purchase'
```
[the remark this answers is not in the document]
mvanbeusekom: check

packages/in_app_purchase/in_app_purchase_ios/lib/src/in_app_purchase_ios_platform.dart
```diff
@@ -0,0 +1,223 @@
...
+import 'package:flutter/foundation.dart';
+import 'package:flutter/services.dart';
+import 'package:in_app_purchase_ios/src/in_app_purchase_ios_platform_addition.dart';
+import 'package:in_app_purchase_platform_interface/in_app_purchase_platform_interface.dart';
+
+import '../in_app_purchase_ios.dart';
+import '../store_kit_wrappers.dart';
+
+/// [IAPError.code] code for failed purchases.
+final String kPurchaseErrorCode = 'purchase_error';
```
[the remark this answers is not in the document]
mvanbeusekom: Sorry old habit,  in Java and C# (and probably other languages as well) having a public const is considered a code smell. Reason is that compile time the const is replaced with the literal value in consuming libraries. This could lead to bugs when you update the const value and replace the library without re-compiling all dependent code. 

I guess this is not a problem in Dart since code is always compiled entirely including dependencies. I have updated it to `const`.

packages/in_app_purchase/in_app_purchase_ios/lib/src/in_app_purchase_ios_platform.dart
```diff
@@ -0,0 +1,223 @@
...
+
+/// [IAPError.code] code used when a query for previouys transaction has failed.
+final String kRestoredPurchaseErrorCode = 'restore_transactions_failed';
+
+/// [IAPError.code] code used when a consuming a purchased item fails.
+final String kConsumptionFailedErrorCode = 'consume_purchase_failed';
```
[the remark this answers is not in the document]
mvanbeusekom: Correct this field is not used anywhere in the iOS package so I removed it.

packages/in_app_purchase/in_app_purchase_ios/lib/src/in_app_purchase_ios_platform.dart
```diff
@@ -0,0 +1,223 @@
...
+import 'package:in_app_purchase_platform_interface/in_app_purchase_platform_interface.dart';
+
+import '../in_app_purchase_ios.dart';
+import '../store_kit_wrappers.dart';
+
+/// [IAPError.code] code for failed purchases.
+final String kPurchaseErrorCode = 'purchase_error';
+
+/// [IAPError.code] code used when a query for previouys transaction has failed.
+final String kRestoredPurchaseErrorCode = 'restore_transactions_failed';
```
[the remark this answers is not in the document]
mvanbeusekom: Correct this field is not used anywhere in the iOS package so I removed it.

packages/in_app_purchase/in_app_purchase_ios/lib/src/in_app_purchase_ios_platform.dart
mvanbeusekom: I really like this one, thanks for pointing it out. I have updated the code and the documentation of the `AppStorePurchaseParam` class.

packages/in_app_purchase/in_app_purchase_ios/lib/src/in_app_purchase_ios_platform.dart
mvanbeusekom: No the user should never get this error indeed. I (ab)used the check so I didn't have to do a double cast. When using an `assert` I have to do another cast when calling the `finishTransaction` method:

```dart
    assert(
      purchase is AppStorePurchaseDetails, // first cast (or check)
      'On iOS, the `purchase` should always be of type `AppStorePurchaseDetails`.',
    );

    return _skPaymentQueueWrapper.finishTransaction(
      (purchase as AppStorePurchaseDetails).skPaymentTransaction, // second cast
    );
```

But since the assert is removed from release versions it doesn't really matter ;). So I think you are right and I updated the code.

packages/in_app_purchase/in_app_purchase_ios/pubspec.yaml
mvanbeusekom: done

packages/in_app_purchase/in_app_purchase_ios/pubspec.yaml
mvanbeusekom: Good point. I make the change and also updated the CHANGELOG.md to reflect the same version number.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/types/purchase_details.dart
mvanbeusekom: Correct, I still needed to merge with master. This should be gone now.

packages/in_app_purchase/in_app_purchase_ios/ios/in_app_purchase.podspec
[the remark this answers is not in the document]
mvanbeusekom: note that this is causing linter warnings in the tree since the url does not exists yet:

```
url: The URL (https://pub.dev/packages/in_app_purchase_ios) is not reachable.
```

not sure how serious we should treat this?

cyanglaz (approved):
LGTM!

cyanglaz:
I think the lint error will block the tree. We can revert the doc url and add a tODO to update it to the correct one when the plugin is published.

mvanbeusekom:
> I think the lint error will block the tree. We can revert the doc url and add a tODO to update it to the correct one when the plugin is published.

Done, thanks for the review @cyanglaz

PR #3840, opened 2021-04-30
[tool] version-check publish-check commands can check against pub

cyanglaz:
Add a `PubVersionFinder` class to easily fetch the version from pub.

Add an `against-pub` flag to `check-version` command, which allows it to check the version against pub server

Make the 'publish-check' command to check against pub to determine if the specific versions of packages need to be published.
Add a `log-status` flag, which allows the `publish-check` command to log the final status of the result. This helps other ci tools to easily grab the results and use it to determine what to do next. See option 3 in  https://github.com/flutter/flutter/issues/81444

This PR also fixes some tests.

partially https://github.com/flutter/flutter/issues/81444

## Pre-launch Checklist

- [x] I signed the [CLA].


script/tool/lib/src/publish_check_command.dart
```diff
@@ -26,9 +29,22 @@ class PublishCheckCommand extends PluginCommand {
           'the SDK constraint is a pre-release version, is ignored.',
       defaultsTo: false,
     );
+    argParser.addFlag(_logStatusFlag,
+        help:
+            'Logs the check-publish final status to a defined string as the last line of the command output.\n'
```
stuartmorgan, line 34: Mixing output for humans and output for machines is generally not a good idea; I would rather we make a `--machine` flag that switches entirely to an output meant for parsing.

script/tool/lib/src/publish_check_command.dart
```diff
@@ -121,24 +159,54 @@ class PublishCheckCommand extends PluginCommand {
...
     }
   }
+
+  // Check if `packageName` already has `version` published on pub.
+  Future<bool> _checkIfAlreadyPublished(
+      {String packageName, Version version}) async {
+    final PubVersionFinder pubVersionFinder = PubVersionFinder(
+        package: packageName, httpClient: httpClient ?? http.Client());
+    final PubVersionFinderResponse pubVersionFinderResponse =
+        await pubVersionFinder.getPackageVersion();
```
stuartmorgan, line 199: Don't we want a try/catch here to deal with issues reaching pub?

stuartmorgan (changes requested):

script/tool/lib/src/publish_check_command.dart
cyanglaz: We don't need try-catch, but we should check the status of pubVersionFinderResponse. Done.

cyanglaz (commented):
@stuartmorgan Updated per review comments. PTAL

script/tool/lib/src/publish_check_command.dart
```diff
@@ -121,24 +184,69 @@ class PublishCheckCommand extends PluginCommand {
...
+    bool published;
+    switch (pubVersionFinderResponse.result) {
+      case PubVersionFinderResult.success:
+        published = pubVersionFinderResponse.versions.contains(version);
+        break;
+      case PubVersionFinderResult.fail:
+        printErrorAndExit(errorMessage: '''
```
stuartmorgan, line 231: Won't this break the `--machine` case, because the final message will never print?

stuartmorgan (changes requested):

script/tool/lib/src/common.dart
```diff
@@ -549,6 +550,97 @@ class ProcessRunner {
...
+}
+
+/// Represents a response for [PubVersionFinder].
+class PubVersionFinderResponse {
+  /// Constructor.
+  const PubVersionFinderResponse({this.versions, this.result, this.httpResponse});
+
+  /// The versions found in [PubVersionFinder].
+  ///
+  /// This is sorted by largest to smallest, so the first element in the list is the largest version.
```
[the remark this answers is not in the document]
cyanglaz: done

script/tool/lib/src/common.dart
```diff
@@ -549,6 +550,97 @@ class ProcessRunner {
...
+  final List<Version> versions;
+
+  /// The result of the version finder.
+  final PubVersionFinderResult result;
+
+  /// The response object of the http request.
+  final http.Response httpResponse;
+}
+
+/// An enum represents the result of [PubVersionFinder].
```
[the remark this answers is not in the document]
cyanglaz: done

script/tool/lib/src/common.dart
```diff
@@ -549,6 +550,97 @@ class ProcessRunner {
...
+  success,
+
+  /// The version finder failed to find a valid version.
+  ///
+  /// This might due to http connection errors or user errors.
+  fail,
+
+  /// The version finder failed to locate the package.
+  ///
+  /// This is usually OK when the package is new.
```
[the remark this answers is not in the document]
cyanglaz: reworded, this was not what I meant.

script/tool/lib/src/publish_check_command.dart
```diff
@@ -56,12 +106,19 @@ class PublishCheckCommand extends PluginCommand {
...
+    } else {
+      final Colorize passedMessage =
+          Colorize('All packages passed publish check!')..green();
+      print(passedMessage);
     }
 
-    final Colorize passedMessage =
-        Colorize('All packages passed publish check!')..green();
-    print(passedMessage);
+    // This has to be last output of this command because it is promised in the help section.
```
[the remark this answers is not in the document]
cyanglaz: done

script/tool/lib/src/publish_check_command.dart
cyanglaz: done.

cyanglaz (commented):
Updated per your comments, PTAL @stuartmorgan

script/tool/lib/src/publish_check_command.dart
stuartmorgan: I was thinking this should return something, rather than bailing completely. Is the reason for this that we assume that all the other checks will fail the same way because it's likely a network issue?

If we do want to keep this construction, this line (and the line above) need comments explaining why they are doing what they are doing, since it's non-obvious.

script/tool/lib/src/publish_check_command.dart
```diff
@@ -121,24 +188,67 @@ class PublishCheckCommand extends PluginCommand {
...
+      case PubVersionFinderResult.success:
+        published = pubVersionFinderResponse.versions.contains(version);
+        break;
+      case PubVersionFinderResult.fail:
+        print(_machineMessageError);
+        printErrorAndExit(errorMessage: '''
+Error fetching version on pub for $packageName.
+HTTP Status ${pubVersionFinderResponse.httpResponse.statusCode}
+HTTP response: ${pubVersionFinderResponse.httpResponse.body}
+''');
```
stuartmorgan, line 238: I'm realizing that failures here are going to be almost impossible to debug, because we won't have any information in the log.

Perhaps we should consider doing something like a JSON dictionary of output, with status and detailed log messages as separate keys? That would mean we'd need to consume the result with something like a Dart script that can parse the JSON, rather than just a one-liner in the GitHub Action. Thoughts?

stuartmorgan (changes requested):

script/tool/lib/src/publish_check_command.dart
cyanglaz: ah ok, I updated to print the message and handled the error later

script/tool/lib/src/publish_check_command.dart
cyanglaz: Good Idea, updated the code to do exactly that

script/tool/lib/src/publish_check_command.dart
```diff
@@ -56,12 +119,21 @@ class PublishCheckCommand extends PluginCommand {
       final Colorize colorizedError = Colorize('$error\n$joinedFailedPackages')
         ..red();
       print(colorizedError);
-      throw ToolExit(1);
+    } else {
+      final Colorize passedMessage =
+          Colorize('All packages passed publish check!')..green();
```
stuartmorgan, line 124: Looking at the new tests, it's very strange to have terminal color commands in the `--machine` output. I think we should replace all the Colorize calls with a utility method like:
```
_printImportantStatusMessage(String message, {@required bool isError}) {
  if (argResults[_machineFlag] as bool) {
    print('${isError ? 'ERROR' : 'SUCCESS'}: $message');
  } else {
    final Colorize colorizedMessage = Colorize(message);
    if (isError) {
      colorizedMessage.red();
    } else {
      colorizedMessage.green();
    }
    print(colorizedMessage);
  }
}

stuartmorgan (changes requested):

script/tool/lib/src/publish_check_command.dart
cyanglaz: I updated so that both machine mesasge and human message include the `error` and `success` prefix.

cyanglaz (commented):
@stuartmorgan Updated the status message bit. PTAL

stuartmorgan (approved):
LGTM!

PR #3841, opened 2021-05-03
[in_app_purchase] Federated Android implementation

mvanbeusekom:
This PR adds the Android implementation of the new "in_app_purchase_platform_interface" (see PR #3781).

Note: when running Flutter commands for the `in_app_purchase_android` package I keep getting the warning below. I think this is caused by the fact that the `in_app_purchase_android` is missing a `MainActivity.java` (since is contains pure plugin code and no `example` which will be part of the `in_app_purchase` package), but I am not sure. If I made a configuration mistake I would be happy to fix it.

Warning:
```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Warning
──────────────────────────────────────────────────────────────────────────────
Your Flutter application is created using an older version of the Android
embedding. It's being deprecated in favor of Android embedding v2. Follow the
steps at

https://flutter.dev/go/android-project-migration

to migrate your project.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

## Pre-launch Checklist

- [x] I signed the [CLA].


cyanglaz:
>since is contains pure plugin code and no example which will be part of the in_app_purchase package

I think we should add examples for this plugin as well, also have a readme to document well how to use only android implementations if people want to (since we advertised it in our readme already)

We should also do the same for the iOS package.

However, these are not priorities since most of the people will use the unified API. But still, I think this can be on our future roadmap. We should create an issue and add TODOs to track it.

I created the issue: https://github.com/flutter/flutter/issues/81695, do you mind adding TODO in this PR?

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
```diff
@@ -0,0 +1,327 @@
...
+
+  @override
+  Future<bool> isAvailable() async {
+    await _readyFuture;
+    return billingClient.isReady();
+  }
+
+  @override
+  Future<bool> buyNonConsumable({required PurchaseParam purchaseParam}) async {
+    if (!(purchaseParam is GooglePlayPurchaseParam)) {
```
cyanglaz, line 92: Similar to iOS, I think we don't have to force it here. Instead, we can document that if the client app wants to use the change subscription feature, they should use a `GooglePlayPurchaseParam` instead. Otherwise, `changeSubscriptionParam` will default to `null`

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
```diff
@@ -0,0 +1,327 @@
...
+        .toSet();
+
+    String errorMessage =
+        errorCodeSet.isNotEmpty ? errorCodeSet.join(', ') : '';
+
+    List<PurchaseDetails> pastPurchases =
+        responses.expand((PurchasesResultWrapper response) {
+      return response.purchasesList;
+    }).map((PurchaseWrapper purchaseWrapper) {
+      return GooglePlayPurchaseDetails.fromPurchase(purchaseWrapper);
```
cyanglaz, line 183: Do we need to set the status to be "restored"?

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
```diff
@@ -0,0 +1,327 @@
...
+
+  Future<void> _connect() =>
+      billingClient.startConnection(onBillingServiceDisconnected: () {});
+
+  /// Query the product detail list.
+  ///
+  /// This method only returns [ProductDetailsResponse].
+  /// To get detailed Google Play sku list, use [BillingClient.querySkuDetails]
+  /// to get the [SkuDetailsResponseWrapper].
+  Future<ProductDetailsResponse> queryProductDetails(
```
cyanglaz, line 221: Where is this method used? This is an addition API right? Which should be in the addition class?

packages/in_app_purchase/in_app_purchase_android/lib/src/types/in_app_purchase_exception.dart
```diff
@@ -0,0 +1,23 @@
+/// Thrown to indicate that an action failed while interacting with the
+/// in_app_purchase plugin.
+class InAppPurchaseException implements Exception {
```
cyanglaz, line 3: Is this android only? Does it make sense to be in the platform_interface?

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
```diff
@@ -0,0 +1,327 @@
...
+    return await billingClient
+        .acknowledgePurchase(purchase.verificationData.serverVerificationData);
+  }
+
+  /// Mark that the user has consumed a product.
+  ///
+  /// You are responsible for consuming all consumable purchases once they are
+  /// delivered. The user won't be able to buy the same product again until the
+  /// purchase of the product is consumed.
+  Future<BillingResultWrapper> consumePurchase(PurchaseDetails purchase) {
```
cyanglaz, line 149: This should also be in the addition class

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
```diff
@@ -0,0 +1,327 @@
...
+  static bool _enablePendingPurchase = false;
+
+  /// Enable the [InAppPurchaseConnection] to handle pending purchases.
+  ///
+  /// This method is required to be called when initialize the application.
+  /// It is to acknowledge your application has been updated to support pending purchases.
+  /// See [Support pending transactions](https://developer.android.com/google/play/billing/billing_library_overview#pending)
+  /// for more details.
+  /// Failure to call this method before access [instance] will throw an exception.
+  static void enablePendingPurchases() {
```
cyanglaz, line 66: This should also be in the addition class.

cyanglaz (changes requested):

packages/in_app_purchase/in_app_purchase_android/analysis_options.yaml
```diff
@@ -0,0 +1 @@
+include: ../../../analysis_options_legacy.yaml
```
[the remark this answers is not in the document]
mvanbeusekom: Done.

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
mvanbeusekom: This method is declared in the `InAppPurchasePlatform` plugin interface, so it should have been decorated with the `@override` attribute. I corrected this mistake.

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
mvanbeusekom: Done.

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
mvanbeusekom: Done.

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
mvanbeusekom: Yes is should. Updated thanks.

packages/in_app_purchase/in_app_purchase_android/lib/src/types/in_app_purchase_exception.dart
mvanbeusekom: At the moment it is only used in the Android package. On iOS we don't need this specific exception. However I can imagine the exception to be useful on other platforms in the future. 

So I wasn't sure to add it now to the platform_interface or later if we really need it on multiple platform (by that time we might have forgotten that there is a similar class in the Android implementation).

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform.dart
mvanbeusekom: Agreed, I updated the code to reflect this behaviour.

cyanglaz:
> > > since is contains pure plugin code and no example which will be part of the in_app_purchase package
> > 
> > 
> > I think we should add examples for this plugin as well, also have a readme to document well how to use only android implementations if people want to (since we advertised it in our readme already)
> > We should also do the same for the iOS package.
> > However, these are not priorities since most of the people will use the unified API. But still, I think this can be on our future roadmap. We should create an issue and add TODOs to track it.
> > I created the issue: [flutter/flutter#81695](https://github.com/flutter/flutter/issues/81695), do you mind adding TODO in this PR?
> 
> No problem at all, but I am not sure where I would add these TODOs (as in which files)? Since there is no example, where would it make the most sense to add a TODO so we encounter it in the future?

In the README.md, we can add todo saying we need to add example apps.

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform_addition.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:in_app_purchase_platform_interface/in_app_purchase_platform_interface.dart';
+
+import '../billing_client_wrappers.dart';
+
+/// Contains InApp Purchase features that are only available on Android.
```
cyanglaz, line 9: nits: on PlayStore

packages/in_app_purchase/in_app_purchase_android/lib/src/types/in_app_purchase_exception.dart
cyanglaz: Yes, I think it belongs to the platform_interface

cyanglaz (changes requested):

packages/in_app_purchase/in_app_purchase_android/lib/src/types/in_app_purchase_exception.dart
mvanbeusekom: I created a separate PR (#3852) which adds the `InAppPurchaseException` to the platform_interface. Once that is merged I will make sure it gets removed from this PR.

mvanbeusekom:
> In the README.md, we can add todo saying we need to add example apps.

Done.

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform_addition.dart
mvanbeusekom: Fixed.

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform_addition.dart
```diff
@@ -0,0 +1,52 @@
...
+  InAppPurchaseAndroidPlatformAddition(this._billingClient) {
+    if (InAppPurchaseAndroidPlatformAddition.enablePendingPurchase) {
+      _billingClient.enablePendingPurchases();
+    }
+  }
+
+  /// Whether pending purchase is enabled.
+  ///
+  /// See also [enablePendingPurchases] for more on pending purchases.
+  static bool get enablePendingPurchase => _enablePendingPurchase;
```
cyanglaz, line 23: Since users will be getting the `InAppPurchaseAndroidPlatformAddition` from the provider, which is going to be a single instance, we can make `_enablePendingPurchase` an instance variable. And `enablePendingPurchases` below an instance method.

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform_addition.dart
```diff
@@ -0,0 +1,52 @@
...
+
+import '../billing_client_wrappers.dart';
+
+/// Contains InApp Purchase features that are only available on PlayStore.
+class InAppPurchaseAndroidPlatformAddition
+    extends InAppPurchasePlatformAddition {
+  /// Creates a [InAppPurchaseAndroidPlatformAddition] which uses the supplied
+  /// `BillingClient` to provide Android specific features.
+  InAppPurchaseAndroidPlatformAddition(this._billingClient) {
+    if (InAppPurchaseAndroidPlatformAddition.enablePendingPurchase) {
```
cyanglaz, line 15: Making this an assert

packages/in_app_purchase/in_app_purchase_android/lib/src/in_app_purchase_android_platform_addition.dart
mvanbeusekom: As per offline discussion we decided to leave it static as it is important the `enablePendingPurchases()` method is called before initialising the `InAppPurchaseAndroidPlatform` class.

cyanglaz (approved):
LGTM

PR #2924, opened 2020-08-11
[google_maps_flutter] Pin GoogleMaps cocoapod dep to last good known version

ditman:
## Description

GoogleMaps 3.10 seems to have increased the minimum requirement for XCode beyond what our test infra can handle (undocumented?).

This change pins the GoogleMaps podspec to <3.10 until the situation clears, to unblock the build trees of both flutter/plugins and flutter/flutter.

**Why are our apps resolving to 3.10, when our Podspec states that ios 8.0 needs to be supported?** 

Apps were resolving to 3.10 because `flutter create` now creates apps that have a minimum iOS version of 9.0 (instead of 8.0).

**Why did 3.10 break us?**

3.10 breaks us because they're using some APIs only available in XCode 11.4 and our infra is in 11.3.1. (However, as @jmagman pointed out, they should check for the availability of these APIs or a runtime error will happen in devices running iOS < 13.4.)

## Related Issues

* https://github.com/flutter/flutter/issues/63447
* b/163359804
* b/163474612

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #3850, opened 2021-05-05
[google_maps_flutter] Unpin iOS GoogleMaps pod dependency version

jmagman:
The plugins infrastructure has migrated past Xcode 11.5, which is related to the original reason this was pinned.

This is already tested in presubmit and a failure will look like the original reported https://github.com/flutter/flutter/issues/63447.

Fixes https://github.com/flutter/flutter/issues/63530.  
See also https://github.com/flutter/plugins/pull/2924

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.4
+
+* Unpin iOS GoogleMaps pod dependency version.
+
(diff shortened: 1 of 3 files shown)


cyanglaz (approved):
LGTM! Thanks

cyanglaz:
@jmagman 
I'll do a query for all the PRs with "last-mile" when "Windows-plugins" are fixed, then I'll land all these PRs.

ditman:
Thanks for the fix @jmagman!!

PR #3859, opened 2021-05-06
[google_sign_in] fix registration links

kevmoo:
Fixes https://github.com/flutter/flutter/issues/81847


stuartmorgan (changes requested):
Most users are reading the README from pub.dev, so this needs a version and CHANGELOG update so we can publish it.

Also, we don't usually strictly enforce an 80-char limit for .md files; is there a reason you've reformatted the whole file?

kevmoo:
> Most users are reading the README from pub.dev, so this needs a version and CHANGELOG update so we can publish it.
> 
> Also, we don't usually strictly enforce an 80-char limit for .md files; is there a reason you've reformatted the whole file?

Updated changelog. I was hoping to bump the example, too – before this is published.

RE formatting: my editor does that for me. I find it easier to read. I also cleaned out some bits at the bottom.

Happy to revert unrelated changes if you prefer.

stuartmorgan:
> * This commit has empty status or empty checks. Please check the Google CLA status is present and Flutter Dashboard application has multiple checks.

Weird. I have no idea what this means.

PR #3882, opened 2021-05-12
Update tool README for packages use

stuartmorgan:
Updates to better reflect that this tooling is still used in flutter/packages, and how that differs from the flutter/plugins usage.

Part of an overall overhaul of flutter/plugins and flutter/packages docs.


cyanglaz (approved):
LGTM

PR #3885, opened 2021-05-13
Revert "Temporarily run Windows tests via GitHub Actions"

stuartmorgan:
Reverts flutter/plugins#3875
Windows bots were restored faster than I expected so we no longer need this


cyanglaz (approved):
LGTM

PR #3888, opened 2021-05-13
[google_sign_in] Remove unneeded null checks affecting package score

kevmoo:
Also fix some lint warnings

### DIFF
--- a/packages/google_sign_in/google_sign_in/CHANGELOG.md
+++ b/packages/google_sign_in/google_sign_in/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Remove unneeded null checks affecting package score.
+
 ## 5.0.3
 
 * Fixed links in `README.md`.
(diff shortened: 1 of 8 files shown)


stuartmorgan:
The asserts here are not unnecessary; see https://github.com/flutter/plugins/blob/master/analysis_options.yaml#L49-L50. there is no guarantee that clients have migrated to null safety.

As for the lints, CI should be enforcing that there aren't analyzer issues. Were you seeing these locally without changing analysis options? (If not, is much rather that analyzer improvements be done by removing--or moving to sub-packages to reduce the change scope-- the temporary https://github.com/flutter/plugins/blob/master/packages/google_sign_in/analysis_options.yaml and fixing all the resulting issues.)

kevmoo:
RE asserts: understood, although I'm guessing in normal  usage the user will face failures either way. I wish we didn't have to pay points in the package score!

image

The other lints: I should have been more clear. I experimented with adding other lints because I was confused why my IDE was NOT showing the null issues. So these are just general cleanup items that happen to map to a set of lints I often use.

erdemsitki4222 (approved):

stuartmorgan:
I'm going to go ahead and close this, since we're doing this deliberately. When we decide it's time to remove these checks (which should probably be a Flutter-wide decision; that setting is mirrored from flutter/flutter) the right way to proceed here will be to change the analysis options and mass-fix the issues it flags, rather than just changing one package's behavior.

PR #3898, opened 2021-05-17
[image_picker] Image picker fix camera device

ydag:
This PR fixes an issue where `preferredCameraDevice` is not working as expected for `getVideo` method. I also refactor device-only unit tests to make sure we can test camera devices on the CI as well.

Related issues :
- Fixes [flutter/flutter#74674](https://github.com/flutter/flutter/issues/74674)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -457,7 +461,7 @@ - (void)imagePickerController:(UIImagePickerController *)picker
       image = [FLTImagePickerImageUtil scaledImage:image
                                           maxWidth:maxWidth
                                          maxHeight:maxHeight
-                               isMetadataAvailable:originalAsset != nil];
+                               isMetadataAvailable:YES];
```
stuartmorgan, line 464: What is this change for? It doesn't seem obviously related.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -158,9 +158,13 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
     }
 
     switch (imageSource) {
-      case SOURCE_CAMERA:
+      case SOURCE_CAMERA: {
+        NSInteger cameraDevice = [[_arguments objectForKey:@"cameraDevice"] intValue];
+        _device = (cameraDevice == 1) ? UIImagePickerControllerCameraDeviceFront
+                                      : UIImagePickerControllerCameraDeviceRear;
```
stuartmorgan, line 164: This mapping from a magic number, as well as the hard-coded key, is now duplicated in two places. Please extract a small helper method that takes an arguments dictionary and returns the device, so that both locations can share the same code.

stuartmorgan (commented):
We should have tests covering this. Long-term that will probably require some refactoring for testability, but short term: was [this test](https://github.com/flutter/plugins/blob/master/packages/image_picker/image_picker/ios/Tests/ImagePickerPluginTests.m#L98) failing and we just didn't know because it's device-only?

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
ydag: Yes, this is not related to this PR but since it is just a small fix I wanted to add it here. This change is related to [#3873](https://github.com/flutter/plugins/pull/3873). `isMetadataAvailable` parameter is used to handle rotation problem for limited access to the photo library and it should be `YES` for all other occasions.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
ydag: done.

ydag:
> We should have tests covering this. Long-term that will probably require some refactoring for testability, but short term: was [this test](https://github.com/flutter/plugins/blob/master/packages/image_picker/image_picker/ios/Tests/ImagePickerPluginTests.m#L98) failing and we just didn't know because it's device-only?

I think the problem is not the test. Since `UIImagePickerControllerCameraDevice _device` is a global variable we need to re-assign before we call the `showCamera` method from both `pickImage` and `pickVideo` in `handleMethodCall`.

But I double-check and I think we even don't need a `UIImagePickerControllerCameraDevice _device` variable at all if we call the `getCameraDevice` helper method inside the `showCamera` method.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -183,9 +186,9 @@ - (void)showCamera {
   }
   // Camera is not available on simulators
   if ([UIImagePickerController isSourceTypeAvailable:UIImagePickerControllerSourceTypeCamera] &&
-      [UIImagePickerController isCameraDeviceAvailable:_device]) {
+      [UIImagePickerController isCameraDeviceAvailable:[self getCameraDevice:_arguments]]) {
     _imagePickerController.sourceType = UIImagePickerControllerSourceTypeCamera;
-    _imagePickerController.cameraDevice = _device;
+    _imagePickerController.cameraDevice = [self getCameraDevice:_arguments];
```
stuartmorgan, line 191: Please use a local variable instead of calling this twice.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -158,9 +160,10 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
     }
 
     switch (imageSource) {
-      case SOURCE_CAMERA:
+      case SOURCE_CAMERA: {
         [self checkCameraAuthorization];
         break;
+      }
```
stuartmorgan, line 166: Why this change?

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
stuartmorgan: If it's not related, then it would be much better to create a new PR for it. Having behavior-changing fixes hidden in unrelated PRs makes things much more difficult when doing things like trying to find the reason for certain changes with blame, or looking for causes of regressions.

It will also need tests, which would require adding even more (unrelated) code here.

stuartmorgan (changes requested):
> > We should have tests covering this. Long-term that will probably require some refactoring for testability, but short term: was [this test](https://github.com/flutter/plugins/blob/master/packages/image_picker/image_picker/ios/Tests/ImagePickerPluginTests.m#L98) failing and we just didn't know because it's device-only?
> 
> I think the problem is not [this test](https://github.com/flutter/plugins/blob/master/packages/image_picker/image_picker/ios/Tests/ImagePickerPluginTests.m#L98). Since `UIImagePickerControllerCameraDevice _device` is a global variable we need to re-assign before we call the `showCamera` method from both `pickImage` and `pickVideo` in `handleMethodCall`.

Sorry, I'm not seeing how this answers my question. To clarify, there are two parts:
- Where are the test changes that this PR needs to verify the fix, per Flutter policy?
- What is the status of the test referenced above, which sounds like it should already have been testing this scenario, but had not alerted us to the failure previously given that the tree is green?

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
ydag: I am not sure why I changed this. I reverted it.

ydag: I am not sure why I changed this. I reverted it.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
ydag: done.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -67,6 +66,12 @@ - (UIViewController *)viewControllerWithWindow:(UIWindow *)window {
   return topController;
 }
 
+- (UIImagePickerControllerCameraDevice)getCameraDevice:(NSDictionary *)arguments {
```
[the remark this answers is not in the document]
ydag: Done.

stuartmorgan:
That means we don't have any meaningful test coverage here, since a regression would (as evidenced by this PR) never show up in our CI.

It looks like this should be testable via native unit tests. It would require class mocking of `AVCaptureDevice` and `UIImagePickerController` unfortunately, since there are class methods called on them, but class mocks are still better than no test coverage at all. Alternately, you could wrap those calls in a helper that can be instantiated and mocked less invasively via DI.

packages/image_picker/image_picker/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.8.0+2
+* Fixes an issue where `preferredCameraDevice` option is not working for `getVideo` method.
```
cyanglaz, line 2:
```suggestion
## 0.8.0+2

* Fixes an issue where `preferredCameraDevice` option is not working for `getVideo` method.
```

packages/image_picker/image_picker/example/ios/Podfile
```diff
@@ -31,7 +31,10 @@ target 'Runner' do
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
 
   target 'RunnerUITests' do
+    platform :ios, '9.0'
     inherit! :search_paths
+    # Pods for testing
+    pod 'OCMock', '~> 3.8.1'
```
cyanglaz, line 37: The target of the unit tests have been changed to `RunnerTests` recently, so this needs to be updated.

packages/image_picker/image_picker/CHANGELOG.md
ydag: Done.

packages/image_picker/image_picker/example/ios/Podfile
ydag: Done.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,151 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
+  // Run test
   FLTImagePickerPlugin *plugin = [FLTImagePickerPlugin new];
   FlutterMethodCall *call =
       [FlutterMethodCall methodCallWithMethodName:@"pickImage"
                                         arguments:@{@"source" : @(0), @"cameraDevice" : @(0)}];
   [plugin handleMethodCall:call
                     result:^(id _Nullable r){
                     }];
+
+  // Expect cameraDevice is equal to UIImagePickerControllerCameraDeviceRear
```
cyanglaz, line 70: This comment seems not necessary, the code below is self-explanatory.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,151 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
+  // Run test
   FLTImagePickerPlugin *plugin = [FLTImagePickerPlugin new];
   FlutterMethodCall *call =
       [FlutterMethodCall methodCallWithMethodName:@"pickImage"
                                         arguments:@{@"source" : @(0), @"cameraDevice" : @(1)}];
   [plugin handleMethodCall:call
                     result:^(id _Nullable r){
                     }];
+
+  // Expect cameraDevice is equal to UIImagePickerControllerCameraDeviceFront
```
cyanglaz, line 98: This comment seems not necessary, the code below is self-explanatory.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,151 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
+  // Run test
   FLTImagePickerPlugin *plugin = [FLTImagePickerPlugin new];
   FlutterMethodCall *call =
       [FlutterMethodCall methodCallWithMethodName:@"pickVideo"
                                         arguments:@{@"source" : @(0), @"cameraDevice" : @(0)}];
   [plugin handleMethodCall:call
                     result:^(id _Nullable r){
                     }];
+
+  // Expect cameraDevice is equal to UIImagePickerControllerCameraDeviceRear
```
cyanglaz, line 126: This comment seems not necessary, the code below is self-explanatory.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,151 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
                                         arguments:@{@"source" : @(0), @"cameraDevice" : @(1)}];
   [plugin handleMethodCall:call
                     result:^(id _Nullable r){
                     }];
-  plugin.result = ^(id result) {
 
-  };
-  [plugin imagePickerControllerDidCancel:[plugin getImagePickerController]];
-  [plugin imagePickerControllerDidCancel:[plugin getImagePickerController]];
+  // Expect cameraDevice is equal to UIImagePickerControllerCameraDeviceFront
```
cyanglaz, line 154: This comment seems not necessary, the code below is self-explanatory.

cyanglaz (approved):
LGTM % nits

mvanbeusekom:
> That means we don't have any meaningful test coverage here, since a regression would (as evidenced by this PR) never show up in our CI.
> 
> It looks like this should be testable via native unit tests. It would require class mocking of `AVCaptureDevice` and `UIImagePickerController` unfortunately, since there are class methods called on them, but class mocks are still better than no test coverage at all. Alternately, you could wrap those calls in a helper that can be instantiated and mocked less invasively via DI.

@stuartmorgan could you have another look at this PR? @ydag updated the tests so the device specific features are mocked using OCMock and unit-tests will run correctly on the simulator. Chris already approved but the PR is blocked on your requested changes.

packages/image_picker/image_picker/example/ios/Podfile
```diff
@@ -31,7 +31,10 @@ target 'Runner' do
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
 
   target 'RunnerTests' do
+    platform :ios, '9.0'
```
stuartmorgan, line 34: Why are you setting the tests to iOS 9? We still support iOS 8 for now. (The messages about deployment target when building with recent versions of Xcode should be harmless.)

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,147 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
 
-#pragma mark - Test camera devices, no op on simulators
+- (void)setUp {
+  _mockUIImagePicker = OCMClassMock([UIImagePickerController class]);
+  _mockAVCaptureDevice = OCMClassMock([AVCaptureDevice class]);
+}
+
+- (void)tearDown {
+  // Put teardown code here. This method is called after the invocation of each test method in the
+  // class.
```
stuartmorgan, line 44: If you're not using tearDown you can remove it.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,147 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
+}
+
 - (void)testPluginPickImageDeviceBack {
-  if (![UIImagePickerController isSourceTypeAvailable:UIImagePickerControllerSourceTypeCamera]) {
-    return;
-  }
+  // UIImagePickerControllerSourceTypeCamera is supported
+  OCMStub(ClassMethod(
+              [_mockUIImagePicker isSourceTypeAvailable:UIImagePickerControllerSourceTypeCamera]))
+      .andReturn(true);
```
stuartmorgan, line 51: The true value for a BOOL is `YES`, not `true`.

Applies throughout the PR.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,147 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
+
+  // UIImagePickerControllerCameraDeviceRear is supported
+  OCMStub(ClassMethod(
+              [_mockUIImagePicker isCameraDeviceAvailable:UIImagePickerControllerCameraDeviceRear]))
+      .andReturn(true);
+
+  // AVAuthorizationStatusAuthorized is supported
+  OCMStub([_mockAVCaptureDevice authorizationStatusForMediaType:AVMediaTypeVideo]).andReturn(3);
```
stuartmorgan, line 59: Why isn't this returning an `AVAuthorizationStatus` constant?

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,147 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
-  };
-  [plugin imagePickerControllerDidCancel:[plugin getImagePickerController]];
-  [plugin imagePickerControllerDidCancel:[plugin getImagePickerController]];
+  XCTAssertEqual([plugin getImagePickerController].cameraDevice,
+                 UIImagePickerControllerCameraDeviceFront);
 }
 
-- (void)testPluginPickVideoDeviceFront {
-  if (![UIImagePickerController isSourceTypeAvailable:UIImagePickerControllerSourceTypeCamera]) {
+#pragma mark - Test camera devices, no op on simulators
```
stuartmorgan, line 155: Blank line after `pragma mark` please.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,147 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
+- (void)testPluginPickImageDeviceCancelClickMultipleTimes {
+  if ([UIImagePickerController isSourceTypeAvailable:UIImagePickerControllerSourceTypeCamera]) {
```
stuartmorgan, line 157: Why are you aborting the test if the camera *is* available?

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,147 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
   [plugin handleMethodCall:call
                     result:^(id _Nullable r){
                     }];
-  XCTAssertEqual([plugin getImagePickerController].cameraDevice,
-                 UIImagePickerControllerCameraDeviceFront);
+  plugin.result = ^(id result) {
+
+  };
+  [plugin imagePickerControllerDidCancel:[plugin getImagePickerController]];
+  [plugin imagePickerControllerDidCancel:[plugin getImagePickerController]];
```
stuartmorgan, line 171: Why is this called twice?

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -27,87 +28,147 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
...
 }
```
stuartmorgan, line 172: Why isn't this test asserting anything?

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -70,6 +69,21 @@ - (UIViewController *)viewControllerWithWindow:(UIWindow *)window {
...
 }
 
+/**
+ * Applies UIImagePickerControllerCameraDevice on the UIImagePickerController.
+ *
+ * If the cameraDevice value that is fetched from @c arguments is 1 then returns
+ * UIImagePickerControllerCameraDeviceFront. If the cameraDevice value that is fetched
+ * from @c arguments is 0 then returns UIImagePickerControllerCameraDeviceRear.
+ *
+ * @param @arguments that should be used to get cameraDevice value.
```
stuartmorgan, line 79: `arguments` is the name of the parameter, not a Doxygen command, so should not have a `@`.

stuartmorgan (changes requested):

packages/image_picker/image_picker/example/ios/Podfile
ydag: I set it because `OCMock` requires iOS 9.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
ydag: Hi Stuart,
this change made by [this commit](https://github.com/flutter/plugins/commit/25a585dfeef6d2dc224fd0dddee327372de63782#) and it already exists in master as well and got merged in my PR. Do you think we should remove it? Or ask what was this test about and try to fix it?

stuartmorgan: @cyanglaz Was this a bug, or is this intentionally only supposed to run on simulator? That seems bizarre.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
stuartmorgan: This hasn't been changed.

ydag: Oops, sorry. I also changed other comments with the same mistake. :)

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -70,6 +69,21 @@ - (UIViewController *)viewControllerWithWindow:(UIWindow *)window {
   return topController;
 }
 
+/**
+ * Applies UIImagePickerControllerCameraDevice on the UIImagePickerController.
```
stuartmorgan, line 73: It doesn't though; it returns the device that should be used. How about "Returns the UIImagePickerControllerCameraDevice to use given [arguments]."

stuartmorgan (approved):
LGTM with one comment nit (no need to get another review once fixed)

PR #3936, opened 2021-05-19
Added Windows to the description

mikolertesx:
Since the plugin is already supposed to support Windows, I think it should mention it as well.


mikolertesx:
@googlebot I signed it!

stuartmorgan (approved):
Thanks for the contribution, LGTM! (I made some minor edits to the list format as long as it's being updated.)

PR #3943, opened 2021-05-20
Feature/greedy gestures

Rexios80:
Use greedy gesture handling so the map does not require 2 fingers to pan on mobile web browser. I really don't see why that would ever be what anyone wants. This probably isn't a good implementation of this, but the ability to do this definitely needs to be implemented somehow.

This fixes MY ISSUE and probably a lot of other people's issues.

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
+++ b/packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
@@ -529,6 +529,7 @@ class _GoogleMapOptions {
       'indoorEnabled': indoorViewEnabled,
       'trafficEnabled': trafficEnabled,
       'buildingsEnabled': buildingsEnabled,
+      'gestureHandling': 'greedy',
     };
   }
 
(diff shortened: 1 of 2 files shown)


Rexios80:
@googlebot I signed it!

stuartmorgan:
Could you file an issue describing the issue please, and link it from here?

Note that the change you've made here is failing a test, so clearly some discussion will be needed; having a clear issue description would help with that.

Rexios80:
Maybe I'm dumb but it looks like issues are disabled for this repo? Where should I submit it?

stuartmorgan:
> Maybe I'm dumb but it looks like issues are disabled for this repo? Where should I submit it?

See https://github.com/flutter/plugins#issues

Note also that you have deleted the PR template rather than filling it out and following the steps there; the template (and the process it lays out) exists for a reason. It would be helpful if you could update the PR description to re-add the template and follow the steps laid out there (naming the PR correctly, providing an issue link, writing tests, etc.)

For now, as this is not suitable for review as-is, I'm going to mark this as a draft. Once you've updated, feel free to remove the draft marker to indicate that it's ready for review.

PR #3949, opened 2021-05-20
[integration_test] Remove the package

stuartmorgan:
This package has been discontinued since the release of Flutter 2.0, as
it's now part of the SDK. Remove the source as it's no longer used
anywhere, and there will be no further need to publish the plugin.

As with e2e, the README is left for now in case of external links to the
package source dir.


dnfield (approved):
LGTM!

dnfield:
I think you'll need to update the pubspecs of the plugins that test using this so that they pull from `sdk: flutter` now instead of the local path, right?

blasten (approved):

dnfield:
Ah, looks like that's already done. Cool.

stuartmorgan:
> Ah, looks like that's already done. Cool.

Yes, I did that a while back; this is really just cleaning up dead code.

PR #3954, opened 2021-05-21
[url_launcher] Fix Link test

stuartmorgan:
One of the Link tests relied on internal implementation details of
url_launcher_platform_interface that have changed, so started failing
when the new version of that plugin was published.

This replaces that test with a test that the relevant
url_launcher_platform_interface function is called, without asserting
anything about how it works internally.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/url_launcher/url_launcher/lib/src/link.dart
+++ b/packages/url_launcher/url_launcher/lib/src/link.dart
@@ -4,10 +4,17 @@
 
 import 'dart:async';
 
+import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
+import 'package:meta/meta.dart';
 import 'package:url_launcher/url_launcher.dart';
 import 'package:url_launcher_platform_interface/link.dart';
 import 'package:url_launcher_platform_interface/url_launcher_platform_interface.dart';
+
+/// The function used to push routes to the Flutter framework.
+@visibleForTesting
+Future<ByteData> Function(Object?, String) pushRouteToFrameworkFunction =
+    pushRouteNameToFramework;
 
 /// A widget that renders a real link on the web, and uses WebViews in native
 /// platforms to open links.
@@ -98,7 +105,7 @@ class DefaultLinkDelegate extends StatelessWidget {
       // case, we push it via Flutter's navigation system instead of letting the
       // browser handle it.
       final String routeName = link.uri.toString();
-      await pushRouteNameToFramework(context, routeName);
+      await pushRouteToFrameworkFunction(context, routeName);
       return;
     }
 
--- a/packages/url_launcher/url_launcher/pubspec.yaml
+++ b/packages/url_launcher/url_launcher/pubspec.yaml
@@ -24,6 +24,7 @@ flutter:
 dependencies:
   flutter:
     sdk: flutter
+  meta: ^1.3.0
   url_launcher_platform_interface: ^2.0.0
   # The design on https://flutter.dev/go/federated-plugins was to leave
   # this constraint as "any". We cannot do it right now as it fails pub publish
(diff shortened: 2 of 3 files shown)


stuartmorgan:

creativecreatorormaybenot:
@Hixie @stuartmorgan With a combination of these changes you broke the plugin - this is a breaking change.

Specifically, one of the two changes would have needed to be a major version bump (`2.0.3` #3893 in the platform interface changed the type and this PR `6.0.5` also changed the type, which means that `url_launcher: 6.0.5` and `url_launcher_platform_interface: 2.0.2` are incompatible, same for `url_launcher: 6.0.4` and `url_launcher_platform_interface: 2.0.3`). Just helped someone who spent 1 hour not knowing how to debug this problem.

I think you should publish a new version of `url_launcher` that explicitly declares the dependency.

stuartmorgan:
> Specifically, one of the two changes would have needed to be a major version bump

Making this a major version change wouldn't really fix anything; the issue here is that this change uncovered that the other change is breaking in a very subtle way even though it was intended not to be.

I'll need to think about the best combination of releases to minimize the chances of resolving to a bad combination.

creativecreatorormaybenot:
@stuartmorgan Right, #4000 is **a** fix in this case (making `2.0.3` mandatory).

> Making this a major version change wouldn't really fix anything

I do not think this is true. If the platform interface version would have been `3.0.0`, then you would have **been required to** explicitly bump the dependency.

stuartmorgan:
> > Making this a major version change wouldn't really fix anything
> 
> I do not think this is true. If the platform interface version would have been `3.0.0`, then you would have **been required to** explicitly bump the dependency.

This isn't the platform interface PR, it's the app-facing package PR. The part I was replying to said "one of the two changes would have needed to be a major version bump", which isn't right; a breaking change needs to be changed in the version of the package containing the breaking change, not clients of that package.

creativecreatorormaybenot:
@stuartmorgan I see, makes sense. Thinking about it that way, I commented on the wrong PR (did not put thought into which one to choose since I thought they were connected).

But you are right: #3893 would have needed a major version bump because it changed public-facing API.

stuartmorgan:
Having had time to dig into this more, and look more closely at the type mismatch, I think most of the discussion above is incorrect.

> @Hixie @stuartmorgan With a combination of these changes you broke the plugin - this is a breaking change.

I don't think there were any breaking changes, just a broken change (this one).

> Specifically, one of the two changes would have needed to be a major version bump (`2.0.3` #3893 in the platform interface changed the type and this PR `6.0.5` also changed the type

... and as a consequence, I don't think this is true either.

> which means that `url_launcher: 6.0.5` and `url_launcher_platform_interface: 2.0.2` are incompatible

This is definitely true. Specifically: `A value of type 'Future<ByteData> Function(BuildContext, String, {bool debugForceRouter})' can't be assigned to a variable of type 'Future<ByteData> Function(Object?, String)'`, which makes sense.

To be compatible with `url_launcher_platform_interface` 2.0.0-2.0.2, this would have needed to use `BuildContext`, not `Object?`, as the type of the first argument. I.e., of the two variants of the interface, it needs to use the one whose arguments are more specifically typed in order for both of the assignments to be valid.

> same for `url_launcher: 6.0.4` and `url_launcher_platform_interface: 2.0.3`).

I don't think this is true, based both on looking at the code (`pushRouteNameToFramework(context, routeName)` should be a perfectly valid call with the newer signature), and on actually trying it. What error were you seeing with this combination?

creativecreatorormaybenot:
@stuartmorgan Ok that makes sense, thanks 🙏 I did test the second combination properly, your conclusion is correct.

PR #3955, opened 2021-05-21
[connectivity] Removed unnecessary parentheses in readme example

Freelenzer:
Just a little change in the example so you can copy and paste without removing the parentheses.


stuartmorgan:
Thanks for the contribution, but please see https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#important-note. This is not a critical fix, so closing.

PR #3982, opened 2021-05-27
[script/tool] speed up the pub get portion of the analyze command

devoncarew:
This PR speeds up the `analyze` command for the plugins tool. We were running pub get twice for all 42 example directories (flutter pub get automatically runs pub get for an example/ child directory when running pub get for the parent). The analyze command was spending 60% of it's time running flutter pub get; after this change the overall command should be about 25% faster.

- work towards https://github.com/flutter/flutter/issues/83261 (and https://github.com/dart-lang/sdk/issues/45873)

## Pre-launch Checklist

- [x] I signed the [CLA].
- [-] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [-] I updated CHANGELOG.md to add a description of the change.

### DIFF
--- a/script/tool/lib/src/analyze_command.dart
+++ b/script/tool/lib/src/analyze_command.dart
@@ -65,8 +65,8 @@ class AnalyzeCommand extends PluginCommand {
     }
 
     final List<Directory> packageDirectories = await getPackages().toList();
-    final List<String> packagePaths =
-        packageDirectories.map((Directory dir) => dir.path).toList();
+    final Set<String> packagePaths =
+        packageDirectories.map((Directory dir) => dir.path).toSet();
     packageDirectories.removeWhere((Directory directory) {
       // We remove the 'example' subdirectories - 'flutter pub get' automatically
       // runs 'pub get' there as part of handling the parent directory.
--- a/script/tool/lib/src/analyze_command.dart
+++ b/script/tool/lib/src/analyze_command.dart
@@ -41,6 +41,7 @@ class AnalyzeCommand extends PluginCommand {
   @override
   Future<void> run() async {
     print('Verifying analysis settings...');
+
     final List<FileSystemEntity> files = packagesDir.listSync(recursive: true);
     for (final FileSystemEntity file in files) {
       if (file.basename != 'analysis_options.yaml' &&
@@ -64,6 +65,14 @@ class AnalyzeCommand extends PluginCommand {
     }
 
     final List<Directory> packageDirectories = await getPackages().toList();
+    final List<String> packagePaths =
+        packageDirectories.map((Directory dir) => dir.path).toList();
+    packageDirectories.removeWhere((Directory directory) {
+      // We remove the 'example' subdirectories - 'flutter pub get' automatically
+      // runs 'pub get' there as part of handling the parent directory.
+      return directory.basename == 'example' &&
+          packagePaths.contains(directory.parent.path);
+    });
     for (final Directory package in packageDirectories) {
       await processRunner.runAndStream('flutter', <String>['packages', 'get'],
           workingDir: package, exitOnError: true);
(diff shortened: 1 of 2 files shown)


script/tool/lib/src/analyze_command.dart
```diff
@@ -64,6 +65,14 @@ class AnalyzeCommand extends PluginCommand {
     }
 
     final List<Directory> packageDirectories = await getPackages().toList();
+    final List<String> packagePaths =
+        packageDirectories.map((Directory dir) => dir.path).toList();
```
stuartmorgan, line 69: `toSet()`? You're only using this for inclusion testing.

script/tool/lib/src/analyze_command.dart
```diff
@@ -64,6 +65,14 @@ class AnalyzeCommand extends PluginCommand {
...
+    packageDirectories.removeWhere((Directory directory) {
+      // We remove the 'example' subdirectories - 'flutter pub get' automatically
+      // runs 'pub get' there as part of handling the parent directory.
+      return directory.basename == 'example' &&
+          packagePaths.contains(directory.parent.path);
+    });
```
stuartmorgan, line 75: I was hoping we could just use `getPlugins` instead of doing the filtering, but it looks like flutter/packages has some other nested constructions.

script/tool/test/analyze_command_test.dart
```diff
@@ -53,6 +53,24 @@ void main() {
...
+    expect(
+        processRunner.recordedCalls,
+        orderedEquals(<ProcessCall>[
+          ProcessCall(
+              'flutter', const <String>['packages', 'get'], plugin1Dir.path),
+          ProcessCall('dart', const <String>['analyze', '--fatal-infos'],
+              plugin1Dir.path),
+        ]));
+  });
+
```
stuartmorgan, line 73: Could you also add a test of a package with a sub-package that's not `example`, to make sure nobody makes the "simplification" that I was about to suggest before checking flutter/packages?

stuartmorgan (changes requested):
Nice, just minor comments.

script/tool/lib/src/analyze_command.dart
devoncarew: will change

script/tool/lib/src/analyze_command.dart
devoncarew: Yeah, I would prefer something more semantic here as well. The logic for how these are computed - getPlugins() and getPackages() - seems complicated, and I wasn't comfortable modifying it.

It looks like `plugins` actually returns packages? And `packages` returns packages plus there example directories?

I would have assumed that `getPlugins()` would return the directory containing the n dart packages that make up a plugin - like `url_launcher/`, `url_launcher_platform_interface/`, `url_launcher_macos/`, ...

I'll want a call like that - the top level plugin directory - for a follow up PR, where I'm going to try and speed up the analysis portion of the analyze call.

script/tool/test/analyze_command_test.dart
devoncarew: We have one like that already - the `analyzes all packages` test - but I can add one where the 2nd package is named 'example' but not contained by the 1st package.

devoncarew (commented):
Some comments - code changes to follow separately.

script/tool/lib/src/analyze_command.dart
stuartmorgan: > It looks like plugins actually returns packages? And packages returns packages plus there example directories?

I suspect what happened is that this was written for flutter/plugins, then expanded for use in flutter/packages without renaming. Fixing the fact that "plugins" is used to refer to packages in general in the tool (including the `--plugins` argument) is on my list.

stuartmorgan (approved):
LGTM

PR #3994, opened 2021-06-01
[script/tool] speed up the repo analysis command

devoncarew:
This PR speeds up the repo's `analyze` command. It's currently a proposal - no tests as yet, and may not be something we want to land.

This PR:
- in-lines the content from pedantic 1.8.0 into the top level analysis options file
- groups the plugin packages into packages related to each plugin (so, grouping all the 6 `url_launcher` related packages together)
- runs `dart analyze` at that plugin granularity instead of individually for each package

From the times below:
- the analysis times are roughly 90% faster w/ this PR, however
- even before this PR (and definitely after), the CI time is dominated by the `pub get` times

Some thoughts:
- the code to get the list of package and plugins (common.dart, getPackages() and getPlugins()) is pretty complicated, w/ some ways to influence then via the command line, and logic to shard things up
- we should have a higher level concept of a 'Plugin' here (a set of related packages), implemented in common.dart, that people can call instead of needing to do some post-processing
- perhaps the sharding can be based on the plugins, and people can just act on all the plugins they get back from the common.dart call

```
--- before PR ---

[pub time    ] 130.87s
[analyze time] 106.947s

[pub time    ] 132.62s
[analyze time] 99.753s

--- after PR ---

[pub time    ] 145.714s
[analyze time] 55.715s

[pub time    ] 133.109s
[analyze time] 51.134s

[pub time    ] 132.943s
[analyze time] 51.523s
```

### DIFF
--- a/analysis_options_legacy.yaml
+++ b/analysis_options_legacy.yaml
@@ -1,4 +1,6 @@
-include: package:pedantic/analysis_options.1.8.0.yaml
+# The pedantic 1.8.0 rules are in-lined below; see the 'linter' section.
+# include: package:pedantic/analysis_options.1.8.0.yaml
+
 analyzer:
   exclude:
     # Ignore generated files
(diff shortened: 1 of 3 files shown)


devoncarew:
cc @stuartmorgan - I don't think this is something we want to land (partly due to the complexity in the `_calculatePluginGroups()` method to infer plugins), but might be useful in thinking about re-doing some of the APIs from common.dart.

stuartmorgan:
> * in-lines the content from pedantic 1.8.0 into the top level analysis options file

Does this make an appreciable difference? Only the legacy file uses pedantic, so we could trivially do this; there's no long-term maintenance tradeoff to consider.

> * groups the plugin packages into packages related to each plugin (so, grouping all the 6 `url_launcher` related packages together)
> [...]
> * the analysis times are roughly 90% faster w/ this PR, however

The counterpoint here is that it will make incremental analysis slower; that's what's done on most PRs, which are the ones developers are most sensitive to the time of. So I don't think we'd want to do it in general. And then the question is whether the benefit to non-incremental is worth the extra complexity of maintaining both.

> * the code to get the list of package and plugins (common.dart, getPackages() and getPlugins()) is pretty complicated, w/ some ways to influence then via the command line, and logic to shard things up

Yes, simplifying this is on my to-do list.

> * we should have a higher level concept of a 'Plugin' here (a set of related packages), implemented in common.dart, that people can call instead of needing to do some post-processing

I think this may fall out of simplifying/standardizing the logic. I'm just not sure how many uses cases there actually are for it.

> * perhaps the sharding can be based on the plugins, and people can just act on all the plugins they get back from the common.dart call

Sharding is used for the slowest tests, which are platform-specific tests, so I'm not sure this would really change anything.

PR #3998, opened 2021-06-02
[path_provider] Add iOS unit tests

jmagman:
Add `RunnerTests` XCTest iOS target to example app that creates the plugin and asserts it's not nil, proving linking is set up correctly.

path_provider part of https://github.com/flutter/flutter/issues/83357

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan (changes requested):
There's no native UI in `path_provider`, so there's no actual value I can imagine from having native UI tests for it. The goal is to have the scaffolding in place for every test we might actually want to add, but adding scaffolding that we won't ever use will just mean maintenance overhead and longer bot cycles.

jmagman:
> There's no native UI in `path_provider`, so there's no actual value I can imagine from having native UI tests for it. The goal is to have the scaffolding in place for every test we might actually want to add, but adding scaffolding that we won't ever use will just mean maintenance overhead and longer bot cycles.

Removed the target.

stuartmorgan (approved):
LGTM!

PR #4003, opened 2021-06-02
[quick_actions] Optimised UI test to prevent flaky CI failures

mvanbeusekom:
This PR will hopefully fixes the flaky `testQuickActionWithFreshStart` iOS UI test (see flutter/flutter#83483).

Since I wasn't able to reproduce the issue on my local environment it is difficult to determine if this really resolves the problem. However the test was launching and terminating the app under test and I think this was interfering to the point where the app is terminated right before the test to check for the "action-two" screen shows up.

I have restructured the test script to ensure the app under test is terminated after each test so the environment should start in a fresh state for each of the tests.

Added benefit is that is makes the UI Test logic easier to read and understand and less prone to mistakes when additional tests are added.

- flutter/flutter#83483

## Pre-launch Checklist

- [x] I signed the [CLA].
  - The version number is not updated since this PR only restructured an existing UI Test so it would run more stable. In my opinion this change didn't warrant a new release to be published.
  - The version number is not updated since this PR only restructured an existing UI Test so it would run more stable.

### DIFF
--- a/packages/quick_actions/quick_actions/example/ios/RunnerUITests/RunnerUITests.m
+++ b/packages/quick_actions/quick_actions/example/ios/RunnerUITests/RunnerUITests.m
@@ -11,11 +11,20 @@ static const int kElementWaitingTime = 30;
 
 @end
 
-@implementation RunnerUITests
+@implementation RunnerUITests {
+  XCUIApplication *_exampleApp;
+}
 
 - (void)setUp {
   [super setUp];
   self.continueAfterFailure = NO;
+  _exampleApp = [[XCUIApplication alloc] initWithBundleIdentifier:@"io.flutter.plugins.quickActionsExample"];
+}
+
+- (void)tearDown {
+  [super tearDown];
+  [_exampleApp terminate];
+  _exampleApp = nil;
 }
 
 - (void)testQuickActionWithFreshStart {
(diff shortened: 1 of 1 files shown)


packages/quick_actions/quick_actions/example/ios/RunnerUITests/RunnerUITests.m
```diff
@@ -11,18 +11,24 @@ @interface RunnerUITests : XCTestCase
...
-@implementation RunnerUITests
+@implementation RunnerUITests {
+  XCUIApplication *_exampleApp;
+}
 
 - (void)setUp {
   [super setUp];
   self.continueAfterFailure = NO;
+  _exampleApp =
+      [[XCUIApplication alloc] initWithBundleIdentifier:@"io.flutter.plugins.quickActionsExample"];
```
stuartmorgan, line 22: Why do we need to specify the app by bundle ID?

stuartmorgan (commented):
Looks good other than one question. It's not really clear to me why this would fix it since it was waiting for the springboard, so the termination should have been complete, but it's worth a try.

packages/quick_actions/quick_actions/example/ios/RunnerUITests/RunnerUITests.m
mvanbeusekom: Not necessarily, however since we also reference the "Springboard" App in the tests I found it easier (more explicit) to reference the example App through the bundle identifier as wel.

Otherwise it would read:

```
_exampleApp = [[XCUIApplication alloc] init];
```

And we need to assume this is the example app.

mvanbeusekom:
> Looks good other than one question. It's not really clear to me why this would fix it since it was waiting for the springboard, so the termination should have been complete, but it's worth a try.

The old test starten with the following code block:

```
XCUIApplication app = [[XCUIApplication alloc] init];
[app launch];
[app terminate];
```

I suspect that the terminate command takes a while to complete and terminates the process to late. So when the test is already progressing the quick action button is tapped the app is launched but then the terminate from earlier kicks in basically cancelling the launch causing the test to timeout because the `action_two` screen is never displayed.

mvanbeusekom:
Looking at the failing step it seems this actually made it worse :(. As now both UI tests seem to fail. I will look into it.

ditman:
@mvanbeusekom since this is intended to fix a flake, can we re-trigger the failing check several times while in PR, to see if this actually fixes the flakiness?

ditman:
This is the flaky test:

**build-ipas+drive-examples CHANNEL:stable PLUGIN_SHARDING:--shardIndex 2 --shardCount 4**

mvanbeusekom:
> @mvanbeusekom since this is intended to fix a flake, can we re-trigger the failing check several times while in PR, to see if this actually fixes the flakiness?
> 
> 
> 
> (I can re-run the test several times today, I just need to know what's the task that usually failed)

@ditman that would be much appreciated. The questionable step in question would be `build-ipas+drive-examples CHANNEL:stable PLUGIN_SHARDING:--shardIndex 2 --shardCount 4`.

I have just triggered it to run for a second time.

ditman:
Thanks @mvanbeusekom, I'll leave it in a window and retrigger it a bunch of times today. Thanks for the info!

mvanbeusekom:
> Thanks for the info @mvanbeusekom, I'll leave it in a window and retrigger it a bunch of times today. [Link](https://cirrus-ci.com/task/4932445645570048).

@ditman and @stuartmorgan I have just run the flaky job 10 times in a row and all green. So it looks good so far. Maybe @ditman you can run it a view times on some other times during the day (maybe it is some caching issue and all builds succeed now because I run them quickly after each other)?

ditman:
> @ditman you can run it a view times on some other times during the day (maybe it is some caching issue and all builds succeed now because I run them quickly after each other)?

@mvanbeusekom yes, I definitely can run this a few more times later in the PM. The main difference between this and `master` is that in the PR, this is running pretty much by itself, and when it runs in `master` this runs with other tasks.

In any case, promising! :D

stuartmorgan (approved):
LGTM; let's give it a try and see what happens over the next few days.

ditman:
I've retried this in `master` a few times, and it seems solid: https://cirrus-ci.com/task/5610373752553472

PR #4018, opened 2021-06-05
[flutter_plugin_tools] Remove global state from tests

stuartmorgan:
Eliminates the global test filesystem and global test packages directory, in favor of local versions. This guarantees that each test runs with a clean filesystem state, rather than relying on cleanup. It also simplifies understanding the tests, since everything is done via params and return values instead of needing to know about the magic global variables and which methods mutate them.

## Pre-launch Checklist

- [x ] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides]. (Note that unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`. See [plugin_tool format])
- [x] I signed the [CLA].

### DIFF
--- a/script/tool/test/common_test.dart
+++ b/script/tool/test/common_test.dart
@@ -34,8 +34,7 @@ void main() {
 
   setUp(() {
     fileSystem = MemoryFileSystem();
-    packagesDir =
-        initializeFakePackages(parentDir: fileSystem.currentDirectory);
+    packagesDir = createPackagesDirectory(fileSystem: fileSystem);
     thirdPartyPackagesDir = packagesDir.parent
         .childDirectory('third_party')
         .childDirectory('packages');
@@ -330,7 +329,7 @@ packages/plugin3/plugin3.dart
 
     setUp(() {
       fileSystem = MemoryFileSystem();
-      initializeFakePackages(parentDir: fileSystem.currentDirectory);
+      createPackagesDirectory(fileSystem: fileSystem);
       gitDirCommands = <List<String>?>[];
       gitDiffResponse = '';
       gitDir = MockGitDir();
--- a/script/tool/test/common_test.dart
+++ b/script/tool/test/common_test.dart
@@ -34,7 +34,8 @@ void main() {
 
   setUp(() {
     fileSystem = MemoryFileSystem();
-    packagesDir = fileSystem.currentDirectory.childDirectory('packages');
+    packagesDir =
+        initializeFakePackages(parentDir: fileSystem.currentDirectory);
     thirdPartyPackagesDir = packagesDir.parent
         .childDirectory('third_party')
         .childDirectory('packages');
@@ -52,7 +53,6 @@ void main() {
       }
       return Future<ProcessResult>.value(mockProcessResult);
     });
-    initializeFakePackages(parentDir: packagesDir.parent);
     processRunner = RecordingProcessRunner();
     plugins = <String>[];
     final SamplePluginCommand samplePluginCommand = SamplePluginCommand(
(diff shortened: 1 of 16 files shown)


script/tool/test/firebase_test_lab_test.dart
```diff
@@ -134,7 +135,7 @@ void main() {
               '/packages/plugin/example'),
           ProcessCall(
               '/packages/plugin/example/android/gradlew',
-              'app:assembleDebug -Pverbose=true -Ptarget=/packages/plugin/example/test_driver/plugin_e2e.dart'
+              'app:assembleDebug -Pverbose=true -Ptarget=/packages/plugin/example/test/plugin_e2e.dart'
```
stuartmorgan, line 138: These seemingly-random changes are because we were in fact not cleaning up the in-memory filesystem between tests in this suite, and so the order these files were evaluated (which I'm assuming was creation date) changed.

script/tool/test/common_test.dart
```diff
@@ -298,10 +268,10 @@ packages/plugin1/plugin1/plugin1.dart
 packages/plugin1/plugin1_platform_interface/plugin1_platform_interface.dart
 packages/plugin1/plugin1_web/plugin1_web.dart
 ''';
-      final Directory plugin1 = createFakePlugin('plugin1',
-          parentDirectoryName: 'plugin1', packagesDirectory: packagesDir);
-      createFakePlugin('plugin2', packagesDirectory: packagesDir);
-      createFakePlugin('plugin3', packagesDirectory: packagesDir);
+      final Directory plugin1 = createFakePlugin('plugin1', packagesDir,
```
gaaclarke, line 271: Nit: It's a bit of a step backward removing the label to the argument.  We could  have made it required instead of making it an argument without a label.

script/tool/test/test_command_test.dart
```diff
@@ -49,16 +49,14 @@ void main() {
               'flutter', const <String>['test', '--color'], plugin2Dir.path),
         ]),
       );
-
-      cleanupPackages();
```
gaaclarke, line 53: Why don't we need cleanupPackages anymore?  The function isn't showing up in the diff, are we still using it somewhere else, should we remove it?

gaaclarke (approved):

script/tool/test/test_command_test.dart
stuartmorgan: It's removed; see script/tool/test/util.dart diff.

script/tool/test/common_test.dart
stuartmorgan: That's what I did initially, and essentially every call was "packagesDirectory: packagesDir", which just seemed noisy, so I made it positional.

cyanglaz (commented):
LGTM % nits

script/tool/test/util.dart
```diff
@@ -11,32 +11,29 @@ import 'package:file/file.dart';
...
+/// Creates a packages directory in the given location.
 ///
-/// If [parentDir] is set the mock packages dir will be creates as a child of
-/// it. If not [mockFileSystem] will be used instead.
-void initializeFakePackages({Directory? parentDir}) {
-  mockPackagesDir =
-      (parentDir ?? mockFileSystem.currentDirectory).childDirectory('packages');
-  mockPackagesDir.createSync();
+/// If [parentDir] is set the packages directory will be created there,
+/// otherwise [fileSystem] must be provided and it will be created an arbitrary
```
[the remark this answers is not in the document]
stuartmorgan: I intentionally said that it would be created in "an arbitrary location" because tests shouldn't be written to assume anything about where in the in-memory filesystem will create the directory. They should save the return value if they need to reference it, not construct a (theoretically) matching path manually as some tests used to d.

PR #4020, opened 2021-06-07
[in_app_purchase] Added userIds to google play purchase

renefloor:
This PR adds the obfuscatedAccountId and obfuscatedProfileId to the PurchaseWrapper of a Google Play Purchase.
It contains documentation about how these id's can be set on the purchase.

Fixes #83651
## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
+++ b/packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
@@ -5,7 +5,6 @@
 package io.flutter.plugins.inapppurchase;
 
 import androidx.annotation.Nullable;
-
 import com.android.billingclient.api.AccountIdentifiers;
 import com.android.billingclient.api.BillingResult;
 import com.android.billingclient.api.Purchase;
@@ -66,7 +65,7 @@ import java.util.List;
     info.put("isAcknowledged", purchase.isAcknowledged());
     info.put("purchaseState", purchase.getPurchaseState());
     AccountIdentifiers accountIdentifiers = purchase.getAccountIdentifiers();
-    if(accountIdentifiers != null) {
+    if (accountIdentifiers != null) {
       info.put("obfuscatedAccountId", accountIdentifiers.getObfuscatedAccountId());
       info.put("obfuscatedProfileId", accountIdentifiers.getObfuscatedProfileId());
     }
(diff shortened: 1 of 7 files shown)


packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
```diff
@@ -63,6 +64,11 @@
     info.put("developerPayload", purchase.getDeveloperPayload());
     info.put("isAcknowledged", purchase.isAcknowledged());
     info.put("purchaseState", purchase.getPurchaseState());
+    AccountIdentifiers accountIdentifiers = purchase.getAccountIdentifiers();
+    if (accountIdentifiers != null) {
```
mvanbeusekom, line 68: There is already a test checking that the `obfuscatedAccountId` and `obfuscatedProfileId` are present in the `HashMap` when the the `accountIdentifiers` object is not `null`. However I think it is also good to test that the keys are not present when the `accountIdentifiers` object is `null`.

This guards against somebody removing the `if` statement without thinking about it.

mvanbeusekom (commented):
Looks very good, I think the code would benefit from one more test.

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
renefloor: @mvanbeusekom this is difficult to test, because although getAccountIdentifiers is nullable, I don't think it will ever be null.

This is the decompiled source code of `getAccountIdentifiers()`
```
    @Nullable
    public AccountIdentifiers getAccountIdentifiers() {
        String var1 = this.zzc.optString("obfuscatedAccountId");
        String var3 = this.zzc.optString("obfuscatedProfileId");
        if (var1 == null && var3 == null) {
            return null;
        } else {
            AccountIdentifiers var2 = new AccountIdentifiers(var1, var3);
            return var2;
        }
    }
```

The account and profile id are retrieved from the JSONObject using [optString(key)](https://developer.android.com/reference/org/json/JSONObject?authuser=1#optString(java.lang.String)).

> Returns the value mapped by name if it exists, coercing it if necessary, or the empty string if no such mapping exists.

Not setting the `obfuscatedAccountId` and `obfuscatedProfileId` in the json will lead to getting a `new AccountIdentifiers("", "")`. Setting them to `null` leads to the same result as that is ignored by the JSONObject decoder. It is marked as `@Nullable` so we can't remove `if (accountIdentifiers != null)`, but it is actually never `null` in the current implementation.

mvanbeusekom: What if you pass in a mock instance of the `Purchase` class? Something like:

```java
class FakePurchase extends Purchase {
  @Override
  public AccountIdentifiers getAccountIdentifiers() {
    return null;
  }
}
```

mvanbeusekom (approved):
LGTM

PR #4030, opened 2021-06-08
[path_provider] Error handle for WidgetsFlutterBinding added

evolvingkid:
In the latest null safety path provider package for flutter, the getApplicationDocumentsDirectory() called on void main  without   WidgetsFlutterBinding.ensureInitialized() cause error as [ERROR:flutter/lib/ui/ui_dart_state.cc(199)] Unhandled Exception: Null check operator used on a null value.
The error actually occurs because of WidgetsFlutterBinding.ensureInitialized() not called .
So I added an error handler to check it. Error handler named as MissingFlutterWidgetInitialized Exception.

### DIFF
--- a/packages/path_provider/path_provider/CHANGELOG.md
+++ b/packages/path_provider/path_provider/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## NEXT
 
 * Add iOS unit test target.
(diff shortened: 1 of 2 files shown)


evolvingkid:
forgot to sign google cla

PR #4047, opened 2021-06-11
[flutter_plugin_tool] Add support for building UWP plugins

stuartmorgan:
This allows building UWP plugin examples with `build-examples --winuwp`. As with previous pre-stable-template desktop support, this avoids the issue of unstable app templates by running `flutter create` on the fly before trying to build, so a template that will bitrot doesn't need to be checked in.

This required some supporting tool changes:
- Adds the ability to check for the new platform variants in a pubspec
- Adds the ability to write test pubspecs that include variants, for testing

The PR also includes a number of cleanups that stemmed from the changes above:
- To avoid adding even more cases where `createFakePlugin` required tests to then overwrite the pubspec immediately due to the limited API of that method, it reworks the way platform support is given to that method.
- Since it was already making that major change to the API, cleans up a number of other oddities:
  - It had two incompatible ways of creating example folders
  - It had two flags for controlling versions instead of one
  - It was being used in a handful of places to make packages, so had an `isFlutter` parameter that could be set to false, which makes no sense for a plugin
  - It defaulted many standard files to off, rather than on, meaning that by default tests ran against something that looked very little like a plugin; this is reversed so that tests that specifically need unusual changes must explicitly change that.
These changes cause some churn across many tests, mostly code removal/reduction.

Part of https://github.com/flutter/flutter/issues/82817

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/CHANGELOG.md
+++ b/script/tool/CHANGELOG.md
@@ -5,6 +5,7 @@
   compatibility.
 - `xctest` now supports running macOS tests in addition to iOS
   - **Breaking change**: it now requires an `--ios` and/or `--macos` flag.
+- `build-examples` now supports UWP plugins via a `--winuwp` flag.
 
 ## 0.2.0
 
--- a/script/tool/lib/src/build_examples_command.dart
+++ b/script/tool/lib/src/build_examples_command.dart
@@ -135,8 +135,8 @@ class BuildExamplesCommand extends PluginCommand {
         }
 
         if (getBoolArg(kPlatformWindows)) {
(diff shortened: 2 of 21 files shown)


stuartmorgan:
Sorry this turned into a yak-shave; once I was changing that test utility method it was very hard to stop...

stuartmorgan:
Hold off on reviewing; I re-implemented most of the yak as a prequel PR.

stuartmorgan:
Now with 100% less yak.

PR #4062, opened 2021-06-17
[integration_test] Update README installation instructions

jmagman:
`integration_test` has moved into the Flutter SDK at `flutter/packages/integration_test`.  However, the "old instructions" `iOS Device Testing` instructions are wrong, which is confusing users who find the README through a Google search.

Update the README to match the instructions in https://github.com/flutter/flutter/tree/master/packages/integration_test#ios-device-testing.  This is not for `pub` publication so won't update https://pub.dev/packages/integration_test instructions.

https://github.com/flutter/flutter/issues/72876

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/integration_test/README.md
+++ b/packages/integration_test/README.md
@@ -57,7 +57,7 @@ Future<void> main() => integrationDriver();
 You can also use different driver scripts to customize the behavior of the app
 under test. For example, `FlutterDriver` can also be parameterized with
 different [options](https://api.flutter.dev/flutter/flutter_driver/FlutterDriver/connect.html).
-See the [extended driver](https://github.com/flutter/plugins/blob/master/packages/integration_test/example/test_driver/extended_integration_test.dart) for an example.
+See the [extended driver](https://github.com/flutter/flutter/blob/master/packages/integration_test/example/test_driver/extended_integration_test.dart) for an example.
 
 ### Package Structure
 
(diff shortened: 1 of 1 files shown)


stuartmorgan (approved):
LGTM in that it's certainly better, but:
- If we're concerned about people finding outdated instructions, don't we *want* to publish this to pub, so there are fewer copies that are wrong?
- We're unlikely to remember to keep updating this; if these are actively confusing, how about we just remove everything but the deprecation notice (and publish that) so there's only one copy going forward?

jmagman:
> If we're concerned about people finding outdated instructions, don't we want to publish this to pub, so there are fewer copies that are wrong?

I would need to resurrect the old code to publish to pub, do we want that?

> We're unlikely to remember to keep updating this; if these are actively confusing, how about we just remove everything but the deprecation notice (and publish that) so there's only one copy going forward?

That makes more sense.  I can update this README to point to a link to https://github.com/flutter/flutter/tree/master/packages/integration_test.

stuartmorgan:
> I would need to resurrect the old code to publish to pub, do we want that?

Oops, I forgot I deleted it.

Let's start with just changing this copy then, and if we see evidence people are being confused by the pub.dev copy we can update the README on a branch that's split off from just before I removed the code, and publish from that.

PR #4064, opened 2021-06-17
[flutter_plugin_tool] Refactor createFakePlugin

stuartmorgan:
createFakePlugin is used extensively in tests, but has become more and more complex over time. This substantially refactors it and related test utilities:
- Extracts a createFakePackage for the cases where we want non-plugin packages (allowing the removal of isFlutter from createFakePlugin).
- Makes most things created automatically instead of being opt-in, so that tests are operating on things that look more like actual packages/plugins unless they specifically override that behavior:
  - Version is added by default.
  - An example/ directory is added by default.
  - A CHANGELOG is added (it can be deleted manually if we ever need to test for that).
- Combines the two different ways of defining the example format.
- Replaces the list of platform boolean with a map of platform name to details. This allows more control over how things are set in the pubspec (currently, federated vs inline), rather than requiring rewriting the pubspec after creation.
- Uses the platform constants in more places instead of raw strings, and renames them to reflect the more general use.
- Eliminates almost all calls to createFakePubspec, as it is almost never needed any more.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/test/drive_examples_command_test.dart
+++ b/script/tool/test/drive_examples_command_test.dart
@@ -6,6 +6,7 @@ import 'package:args/command_runner.dart';
 import 'package:file/file.dart';
 import 'package:file/memory.dart';
 import 'package:flutter_plugin_tools/src/common/core.dart';
+import 'package:flutter_plugin_tools/src/common/plugin_utils.dart';
 import 'package:flutter_plugin_tools/src/drive_examples_command.dart';
 import 'package:path/path.dart' as p;
 import 'package:platform/platform.dart';
@@ -40,8 +41,10 @@ void main() {
             <String>['example', 'test_driver', 'plugin_test.dart'],
             <String>['example', 'test', 'plugin.dart'],
           ],
-          isIosPlugin: true,
-          isAndroidPlugin: true);
+          platformSupport: <String, PlatformSupport>{
+            kPlatformAndroid: PlatformSupport.inline,
+            kPlatformIos: PlatformSupport.inline,
+          });
 
       final Directory pluginExampleDirectory =
           pluginDirectory.childDirectory('example');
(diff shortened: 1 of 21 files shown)


stuartmorgan:
A lot of these changes are pretty mechanical. `utils.dart` is where most of the interesting changes are.

script/tool/test/drive_examples_command_test.dart
```diff
@@ -165,21 +154,20 @@ void main() {
...
-            <String>['example', 'test_driver', 'integration_test.dart'],
-            <String>['example', 'integration_test', 'bar_test.dart'],
-            <String>['example', 'integration_test', 'foo_test.dart'],
-            <String>['example', 'integration_test', 'ignore_me.dart'],
-          ],
-          isAndroidPlugin: true,
-          isIosPlugin: true);
+      final Directory pluginDirectory =
+          createFakePlugin('plugin', packagesDir, extraFiles: <List<String>>[
+        <String>['example', 'test_driver', 'integration_test.dart'],
```
gaaclarke, line 159: nit: it would be a bit more concise if this was just a list of strings that are paths, then we replaced the '/' with platform specific characters.  No need to address it here, just calling it out.
```dart
<String>[
        'example/test_driver/integration_test.dart',
        'example/integration_test/bar_test.dart',
        ...
```

script/tool/test/util.dart
```diff
@@ -110,45 +120,38 @@ void createFakeCHANGELOG(Directory parent, String texts) {
...
   String? version,
 }) {
+  isPlugin |= platformSupport.isNotEmpty;
   parent.childFile('pubspec.yaml').createSync();
   String yaml = '''
 name: $name
+''';
+  if (isFlutter) {
+    if (isPlugin) {
+      yaml += '''
```
gaaclarke, line 144: nit: Instead of manipulating strings, we could create the datastructure then use a yaml library to generate the yaml.  That would make this a bit more maintainable.

gaaclarke (approved):
lgtm, it was a lot to look at.  I just had a few nits.

script/tool/test/util.dart
stuartmorgan: Yes, I've thought about that as well. Initial investigation turned up mostly libraries that can read YAML, unfortunately, rather than write it, but I'd like to look into it more at some point.

script/tool/test/drive_examples_command_test.dart
stuartmorgan: I think that may actually already work by accident, but I'll look at formalizing it. I agree the current system is needlessly messy.

PR #3388, opened 2021-01-05
[image_picker_platform_interface] use XFile instead of PickedFile

balvinderz:
## Description

In this PR,  i have migrated the image_picker_platform_interface to use XFile instead of PickedFile. In further PRs i will migrate the image_picker and image_picker_web to use XFile also. This will allow users to get the length of the file(https://github.com/flutter/flutter/issues/58764) and name of the file(https://github.com/flutter/plugins/pull/2802#issuecomment-642360958) .

## Related Issues

flutter/flutter#58764 flutter/flutter#70886

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [X] All existing and new tests are passing.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [X] I read and followed the [Flutter Style Guide].
- [X] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [X] I updated CHANGELOG.md to add a description of the change.
- [X] I signed the [CLA].
- [X] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [X] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [ ] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #4072, opened 2021-06-21
[image_picker_platform_interface] Migrate image_picker to package:cross_file

BeMacized:
Removed the PickedFile implementations and migrated the platform interface to make use of the `cross_file` library.
Changes have been based on #3388.

Web implementation can be found in: #4083.

Android/iOS implementation can be found in #4073.

Relevant issue:
- flutter/flutter#70886

## Pre-launch Checklist

- [x] I signed the [CLA].


ditman:
@BeMacized I've been going back and forth with this one for a couple of days, but we have some internal customers that would benefit greatly from this being a non-breaking change.

Instead of changing the signature of the methods to return an XFile, could we add a new set of methods that return the new type, while keeping the old methods available (and mark them as deprecated?)

That way our customers are going to be able to see the incoming change before forcing them to update their apps.

At a later date, after a few months, or when the next inescapable major version comes, we can delete the deprecated methods and the PickedFile class :/

What do you think?

BeMacized:
@ditman If you believe this is something important enough that we should take that route, I'd say we do it. 

The only thing I'm a bit vague on atm is what the naming should be for the new alternatives to `pickImage`, `pickVideo` and `LostData`. Just putting a "XFile" suffix on them feels a bit odd, especially considering they will be the only ones left after their deprecated counterparts have been removed in a later version.

Do you happen to have any suggestions?

ditman:
@BeMacized I think we're at a point where we can reuse the names from v0.6, that have already been removed from everywhere :)

I'd say we replace `pick` by `get`, maybe? `getImage`, `getVideo`, `getMultipleImages`... It's not great because we can trigger the camera with this, but at least it doesn't need to use XFile anywhere in the name.

As for the LostData, we might reuse the names from the earlier version? `LostData` class -> `LostDataResponse`, and `retrieveLostData` -> `getLostData`?

There's some notes about the old method names here: https://pub.dev/documentation/image_picker/latest/#deprecation-warnings-in-pickimage-pickvideo-and-lostdataresponse

BeMacized:
@ditman Ah, that indeed sounds like a good idea! I'll see if I can update the PRs for this tomorrow.

ditman:
Thanks for the patience and extra work @BeMacized!

BeMacized:
@ditman Would it be possible for you to take another look? :)

ayushin:
Really looking forward to this merged

ditman:
Apologies, I was working in fixing a framework [P2](https://github.com/flutter/flutter/issues/84832), I'll get on this today.

ditman:
I just realized that making things `@Deprecated` here will break other packages when this gets published, see for example the image_picker with this patch:

```
dit@dit:~/github/plugins/packages/image_picker/image_picker$ flutter analyze
Warning: You are using these overridden dependencies:
! image_picker_platform_interface 2.2.0 from path ../image_picker_platform_interface
Running "flutter pub get" in image_picker...                       807ms
Analyzing image_picker...                                               

   info • 'LostData' is deprecated and shouldn't be used. Switch to using LostDataResponse instead • test/image_picker_test.dart:313:17 •
          deprecated_member_use
   info • 'LostData' is deprecated and shouldn't be used. Switch to using LostDataResponse instead • test/image_picker_test.dart:326:17 •
          deprecated_member_use
```

For other users the breakage might be different. It's a little bit late today, but tomorrow I'll update this PR removing the `@Deprecated` annotations. We'll add the deprecation warnings to the `image_picker` package directly, when people attempt to use the old methods that return PickedFile, instead of the new ones that return XFile. That way people will get the deprecation warnings only once everything is in place for them to switch to the new code.

**My bad, I'll fix this PR ASAP!**

ditman:
@googlebot I consent.

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
```diff
@@ -180,4 +180,92 @@ class MethodChannelImagePicker extends ImagePickerPlatform {
...
+  @override
+  Future<LostDataResponse> getLostData() async {
+    final Map<String, dynamic>? result =
+        await _channel.invokeMapMethod<String, dynamic>('retrieve');
+
+    if (result == null) {
+      return LostDataResponse.empty();
+    }
+
+    assert(result.containsKey('path') ^ result.containsKey('errorCode'));
```
ditman, line 245: I need to ask internally if this bitwise-XOR works with booleans... or *how* it works (I'm guessing some implicit boolean -> int casting)... Anyway, this is non-documented behavior from what I could find in the language tour (and iffy)  :/

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
```diff
@@ -180,4 +180,92 @@ class MethodChannelImagePicker extends ImagePickerPlatform {
...
+      maxHeight: maxHeight,
+      imageQuality: imageQuality,
+    );
+    if (paths == null) return null;
+
+    final List<XFile> files = [];
+    for (final path in paths) {
+      files.add(XFile(path));
+    }
+    return files;
```
ditman, line 219: This can be rewritten as:

```dart
return paths.map((String path) => XFile(path)).toList();
```

(especially since we're not doing any validation of `path` within the `for` loop?)

ditman (approved):
This looks good to me, I'll take an eye to the tests and see what's going on there (the branch needed a refresh from `master` to pick some changes from earlier today).

@BeMacized, can you verify that you can still work on the other packages with my changes applied? Once you verify that you can still keep working, we can merge this change.

ditman:
This needs another refresh from master... refreshing now...

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
BeMacized: I'm not actually sure of this myself, the contents of all of these methods are basically just 1:1 copies of their existing counterparts. I updated both instances of this to hopefully be a bit more conventional.

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
BeMacized: Yep, agreed. Came from the existing function as well, fixed it in both instances.

BeMacized:
@ditman I can confirm the changes have no effect on the implementation PRs, they have been updated accordingly.

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
ditman: Ah, yes, this is valid since Dart 2.1, according to the docs:

https://api.flutter.dev/flutter/dart-core/bool/operator_bitwise_exclusive_or.html #TIL

ditman:
Finally published: https://pub.dev/packages/image_picker_platform_interface/changelog#220

cedvdb:
cross_file cannot be used in `Image.file(File(uri), fit: BoxFit.cover);` while universal_io File can.  What is the reason for using the former ?

PR #4077, opened 2021-06-22
[in_app_purchase] billingClient launchPriceChangeConfirmationFlow

renefloor:
This PR adds a launchPriceChangeConfirmationFlow function in the BillingClientWrapper and the platform additions.

In the android example of the in_app_purchase I replaced the checkmark for subscriptions with an arrow up, this now checks for price changes. I could not get a real price change showing in the test environment.

Fixes flutter/flutter#85038

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/MethodCallHandlerImpl.java
+++ b/packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/MethodCallHandlerImpl.java
@@ -24,7 +24,6 @@ import com.android.billingclient.api.BillingFlowParams.ProrationMode;
 import com.android.billingclient.api.BillingResult;
 import com.android.billingclient.api.ConsumeParams;
 import com.android.billingclient.api.ConsumeResponseListener;
-import com.android.billingclient.api.PriceChangeConfirmationListener;
 import com.android.billingclient.api.PriceChangeFlowParams;
 import com.android.billingclient.api.PurchaseHistoryRecord;
 import com.android.billingclient.api.PurchaseHistoryResponseListener;
@@ -149,11 +148,9 @@ class MethodCallHandlerImpl
         break;
       case InAppPurchasePlugin.MethodNames.IS_FEATURE_SUPPORTED:
         isFeatureSupported((String) call.argument("feature"), result);
-		break;
+        break;
       case InAppPurchasePlugin.MethodNames.LAUNCH_PRICE_CHANGE_CONFIRMATION_FLOW:
-        launchPriceChangeConfirmationFlow(
-                (String) call.argument("sku"),
-                result);
+        launchPriceChangeConfirmationFlow((String) call.argument("sku"), result);
         break;
       default:
         result.notImplemented();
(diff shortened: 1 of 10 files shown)


packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/MethodCallHandlerImpl.java
```diff
@@ -374,6 +378,41 @@ private void updateCachedSkus(@Nullable List<SkuDetails> skuDetailsList) {
...
+          "ACTIVITY_UNAVAILABLE",
+          "launchPriceChangeConfirmationFlow is not available. "
+              + "This method must be run with the app in foreground.",
+          null);
+      return;
+    }
+    if (billingClientError(result)) {
+      return;
+    }
+    assert billingClient != null;
```
mvanbeusekom, line 393: I know it `assert` is used on more places in the code base (and in other plugins as well) but I recently learned that in most situations `assert` doesn't work on Android development (see also https://stackoverflow.com/a/6176529/5167831 and https://stackoverflow.com/a/8164195/5167831). With this in mind the recommended solution would be to throw an `AssertionError`:

```suggestion
    if (billingClient == null) {
      throw new AssertionError("The `billingClient` should not be null.");
    }
```

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/MethodCallHandlerImpl.java
```diff
@@ -374,6 +378,41 @@ private void updateCachedSkus(@Nullable List<SkuDetails> skuDetailsList) {
...
 
+  private void launchPriceChangeConfirmationFlow(String sku, MethodChannel.Result result) {
+    if (activity == null) {
+      result.error(
+          "ACTIVITY_UNAVAILABLE",
+          "launchPriceChangeConfirmationFlow is not available. "
+              + "This method must be run with the app in foreground.",
+          null);
+      return;
+    }
```
mvanbeusekom, line 389: This code block should be tested. The same goes for the `if (billingClientError(result))`, the assert and `if (skuDetails == null)` checks.

mvanbeusekom (changes requested):
Looks good, just have two remarks considering the Java implementation.

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/MethodCallHandlerImpl.java
renefloor: The `if (billingClientError(result))` is already doing `if (billingClient == null)`, so this is only so the analyzer doesn't warn it can be null, because it can't be. Changing this doesn't really add anything, but maybe we could write a test for `billingClientError` that it returns true if billingClient == null.

This is the billingClientError method:
```
  private boolean billingClientError(MethodChannel.Result result) {
    if (billingClient != null) {
      return false;
    }

    result.error("UNAVAILABLE", "BillingClient is unset. Try reconnecting.", null);
    return true;
  }
```

mvanbeusekom: Then let's add a comment to the `assert` line that it is only there to silence the analyser. Maybe something like:
```suggestion
    // Note that assert doesn't work on Android (see https://stackoverflow.com/a/6176529/5167831 and https://stackoverflow.com/a/8164195/5167831)
    // and that this assert is only added to silence the analyser. The actual null check 
    // is handled by the `billingClientError()` call. 
    assert billingClient != null;
```

mvanbeusekom (approved):
LGTM

PR #4082, opened 2021-06-23
Android maps

bparrishMines:

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter/README.md
```diff
@@ -46,6 +46,19 @@ This means that app will only be available for users that run Android SDK 20 or
                android:value="YOUR KEY HERE"/>
 ```
 
+#### Hybrid Composition
+
+To use Hybrid Composition to render the `GoogleMap` widget on Android. Set the
```
stuartmorgan, line 51: Nit: s/. Set the/, set/

stuartmorgan, line 51: Nit: maybe link "Hybrid Composition" to an explanation of what that is for people who don't know?

packages/google_maps_flutter/google_maps_flutter/example/lib/main.dart
```diff
@@ -66,5 +68,11 @@ class MapsDemo extends StatelessWidget {
 }
 
 void main() {
+  WidgetsFlutterBinding.ensureInitialized();
```
stuartmorgan, line 71: Why is this necessary? Setting that flag doesn't send a message, does it?

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -250,7 +258,23 @@ class _GoogleMapState extends State<GoogleMap> {
 
   @override
   Widget build(BuildContext context) {
-    return GoogleMapsFlutterPlatform.instance.buildView(
+    final GoogleMapsFlutterPlatform platform =
+        GoogleMapsFlutterPlatform.instance;
+    if (platform is MethodChannelGoogleMapsFlutter) {
```
stuartmorgan, line 263: Why can't the new method be called unconditionally? The delegation to the old method was supposed to be handled by a default implementation in the platform interface.

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
bparrishMines: The default implementation (`MethodChannelGoogleMapsFlutter`) does call the old method if the platform is not android or hybrid composition is turned off. However, web switches to another platform instance: https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart#L13. 

So in this case, we would need to call the other method, right?

packages/google_maps_flutter/google_maps_flutter/example/lib/main.dart
bparrishMines: You're right. Force of habit. removed

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
stuartmorgan: Sorry, that was poorly phrased; by "default implementation" I mean the code that's part of the base class that all implementations extend from. Having a fallback implementation there is what makes this overall change harmless even for federated implementations we don't even know about.

bparrishMines: Ah ok. So [buildViewWithTextDirection](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart#L464) was only added to `MethodChannelGoogleMapsFlutter` and not to [GoogleMapsFlutterPlatform](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_flutter_platform.dart#L26). This is same for the flag `useAndroidViewSurface`. This was done because this was an Android only requirement. 

Were you thinking that `buildViewWithTextDirection` should have been added to `GoogleMapsFlutterPlatform` as well? I think that makes more sense and could be done without a breaking change.

bparrishMines: I made the change in https://github.com/flutter/plugins/pull/4121. How would we make this temporary? Would we just make a breaking change for the platform_interface to make the transition to the new method?

stuartmorgan: Maybe I misunderstood about it being temporary. What's the long-term plan for these methods? Are we keeping both forever?

bparrishMines: For this method to only be temporary, the platform_interface would have to be broken and bumped to version 3.0. This is because any implementation would need to add the additional parameter to their method. However, I believe the policy is to avoid breaking changes to the platform_interface.

An alternative is to have two methods, and to call `buildViewWithTextDirection` by default in `google_maps_flutter`. But this would still break any implementation that doesn't implement `buildViewWithTextDirection` as the method would throw an `UnimplementedError()`. 

This PR essentially makes `buildViewWithTextDirection` unique to `MethodChannelPlatform`. The `GoogleMap` widget then calls it iff it is `MethodChannelPlatform`. I believe this avoids any breaking changes and any of the [implementations](https://pub.dev/packages?q=dependency%3Agoogle_maps_flutter_platform_interface) would still work with the plugin.

stuartmorgan: > An alternative is to have two methods, and to call `buildViewWithTextDirection` by default in `google_maps_flutter`. But this would still break any implementation that doesn't implement `buildViewWithTextDirection` as the method would throw an `UnimplementedError()`.

That's why I had recommended having the method not throw UnimplementedError, but instead calling the old method.

But my question wasn't about implemented, but the ideal goal: do we want this new parameter forever, or is the need for this parameter temporary? I had thought it was the latter, but again, I may have misunderstood.

bparrishMines: Ah, you wanted that done at the interface level and not the implementation level. That makes since. 

This parameter is required for Android hybrid composition unless we don't want the user to be able to change the text direction. I haven't seen a way around it yet that doesn't involve framework changes. My intention is that it's a permanent parameter.

stuartmorgan: Okay, sorry for the confusion on my part. In that case yes, I think we should update the platform interface to have a default implementation that calls the old method with a default value. Then implementations (those that want a different behavior than that default) can override both, but nobody will be broken and the details of who implements what won't bleed into this package.

bparrishMines: That works for me! I updated https://github.com/flutter/plugins/pull/4121 to reflect this.

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -250,9 +258,12 @@ class _GoogleMapState extends State<GoogleMap> {
 
   @override
   Widget build(BuildContext context) {
-    return GoogleMapsFlutterPlatform.instance.buildView(
+    return GoogleMapsFlutterPlatform.instance.buildViewWithTextDirection(
       _mapId,
       onPlatformViewCreated,
+      textDirection: widget.textDirection ??
+          Directionality.maybeOf(context) ??
```
bparrishMines, line 265: This may return `null` if there is no `Directionality` ancestor. This was the case with the tests, so I also added a default `TextDirection.ltr`.

bparrishMines:
@stuartmorgan I believe this is ready for another review.

packages/google_maps_flutter/google_maps_flutter/README.md
```diff
@@ -46,6 +46,19 @@ This means that app will only be available for users that run Android SDK 20 or
                android:value="YOUR KEY HERE"/>
 ```
 
+#### Hybrid Composition
+
+To use [Hybrid Composition](https://flutter.dev/docs/development/platform-integration/platform-views) to render the `GoogleMap` widget on Android. Set
```
stuartmorgan, line 51: Nit: s/. Set/, set/

packages/google_maps_flutter/google_maps_flutter/README.md
```diff
@@ -46,6 +46,19 @@ This means that app will only be available for users that run Android SDK 20 or
...
+`MethodChannelGoogleMapsFlutter.useAndroidViewSurface` to true.
```
stuartmorgan, line 52: The use of method channels (and especially the fact that we are sharing the same impl between Android and iOS currently) is an implementation detail, so I would really like to not expose it. How about a trivial `AndroidGoogleMapsFlutter` subclass that can be exported and used here.

stuartmorgan (changes requested):
Sorry for the delay, just a couple of small things.

stuartmorgan:
Was this ready for me to review again?

bparrishMines:
> Was this ready for me to review again?

@stuartmorgan I forgot that this hadn't landed yet. But, yes it's ready for another review.

PR #4087, opened 2021-06-23
[flutter_plugin_tools] Migrate build-examples to new base command

stuartmorgan:
Switches build-examples to the new base command that handles the boilerplate of looping over target packages.

While modifying the command, also does some minor cleanup:
- Extracts a helper to reduce duplicated details of calling `flutter build`
- Switches the flag for iOS to `--ios` rather than `--ipa` since `ios` is what is actually passed to the build command
- iOS no longer defaults to on, so that it behaves like all the other platform flags
- Passing no platform flags is now an error rather than a silent pass, to ensure that we never accidentally have CI doing a no-op run without noticing.
- Rewords the logging slightly for the versions where the label for what is being built is a platform, not an artifact (which is now everything but Android).

Part of flutter/flutter#83413

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/build_examples_command.dart
+++ b/script/tool/lib/src/build_examples_command.dart
@@ -3,7 +3,6 @@
 // found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:io' as io;
 
 import 'package:file/file.dart';
 import 'package:path/path.dart' as p;
--- a/script/tool/lib/src/build_examples_command.dart
+++ b/script/tool/lib/src/build_examples_command.dart
@@ -14,11 +14,8 @@ import 'common/package_looping_command.dart';
 import 'common/plugin_utils.dart';
 import 'common/process_runner.dart';
 
-/// Key for IPA.
-const String kIpa = 'ipa';
-
 /// Key for APK.
-const String kApk = 'apk';
+const String _platformFlagApk = 'apk';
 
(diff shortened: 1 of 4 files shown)


script/tool/lib/src/build_examples_command.dart
```diff
@@ -49,164 +47,125 @@ class BuildExamplesCommand extends PluginCommand {
...
-                ],
-                workingDir: example);
-            if (exitCode != 0) {
-              failingPackages.add('$packageName (macos)');
-            }
-          } else {
-            print('macOS is not supported by this plugin');
+      if (getBoolArg(kPlatformMacos)) {
+        print('\nBUILDING $packageName for macOS');
+        if (isMacOsPlugin(package)) {
```
gaaclarke, line 88: Should there be an else branch for isMacOsPlugin?  Printing out a warning?

gaaclarke (approved):

script/tool/lib/src/build_examples_command.dart
stuartmorgan: There is an `else`; it's the `printSkip`. It's not a warning because it's normal for many plugins not to support a given platform, particularly with federation.

stuartmorgan:
The tree is green, but `submit-queue` isn't updating correctly, so landing manually.

PR #4099, opened 2021-06-24
[flutter_plugin_tools] Overhaul drive-examples

stuartmorgan:
Significantly restructures drive-examples:
- Migrates it to the new package-looping base command
- Enforces that only one platform is passed, since in practice multiple platforms never actually worked. (The logic is structured so that it will be easy to enable multi-platform if `flutter drive` gains multi-platform support.)
- Fixes the issue where `--ios` and `--android` were semi-broken, by doing explicit device targeting for them rather than relying on the default device being the right kind
- Extracts much of the logic to helpers so it's easier to understand the flow
- Removes support for a legacy integration test file structure that is no longer used
- Adds more test coverage; previously no failure cases were actually tested.

Fixes https://github.com/flutter/flutter/issues/85147
Part of https://github.com/flutter/flutter/issues/83413

## Pre-launch Checklist

- [x] I signed the [CLA].


script/tool/lib/src/drive_examples_command.dart
```diff
@@ -43,213 +48,252 @@ class DriveExamplesCommand extends PluginCommand {
...
   @override
   final String description = 'Runs driver tests for plugin example apps.\n\n'
-      'For each *_test.dart in test_driver/ it drives an application with a '
-      'corresponding name in the test/ or test_driver/ directories.\n\n'
-      'For example, test_driver/app_test.dart would match test/app.dart.\n\n'
-      'This command requires "flutter" to be in your path.\n\n'
-      'If a file with a corresponding name cannot be found, this driver file'
-      'will be used to drive the tests that match '
-      'integration_test/*_test.dart.';
+      'For each *_test.dart in test_driver/ it drives an application with the '
```
gaaclarke, line 51: s/the/

script/tool/lib/src/drive_examples_command.dart
```diff
@@ -43,213 +48,252 @@ class DriveExamplesCommand extends PluginCommand {
...
+    }
+
+    String? androidDevice;
+    if (getBoolArg(kPlatformAndroid)) {
+      final List<String> devices = await _getDevicesForPlatform('android');
+      if (devices.isEmpty) {
+        printError('No Android devices available');
+        throw ToolExit(_exitNoAvailableDevice);
+      }
+      androidDevice = devices.first;
```
gaaclarke, line 89: Should we warn if there are more than one device?

script/tool/lib/src/drive_examples_command.dart
```diff
@@ -43,213 +48,252 @@ class DriveExamplesCommand extends PluginCommand {
...
     }
-    print('\n\n');
+    if (!testsRan) {
+      printError('No driver tests were run ($examplesFound example(s) found).');
+      errors.add('No tests ran (use --exclude if this is intentional).');
+    }
+    return errors;
+  }
+
+  Future<List<String>> _getDevicesForPlatform(String platform) async {
```
gaaclarke, line 201: nit: Splitting this into parsing the results of `devices` and calling `devices` would make it more testable.  (Although it looks like you can use MockProcess to test it fine.)

script/tool/lib/src/drive_examples_command.dart
```diff
@@ -43,213 +48,252 @@ class DriveExamplesCommand extends PluginCommand {
...
+        }
+      }
     }
-    if (isWindows) {
-      return isWindowsPlugin(plugin);
+    return tests;
+  }
+
+  /// Runs each file from [targets] using `flutter drive`, returning a list of
+  /// any failing test targets.
```
gaaclarke, line 267: nit: if you are going to have a docstring, it helps to mention all the arguments.  It's kind of hard to guess what they all mean.

script/tool/test/drive_examples_command_test.dart
```diff
@@ -35,52 +39,55 @@ void main() {
...
-      ]);
+    test('fails for iOS if no iOS devices are present', () async {
+      setMockFlutterDevicesOutput(hasIosDevice: false);
+      expect(
+        () => runCapturingPrint(runner, <String>['drive-examples']),
+        throwsA(isA<ToolExit>()),
+      );
+    });
 
+    test('fails if no platforms are provided', () async {
```
gaaclarke, line 76: Shouldn't this test have a device, but fail since there is no platform provided?

gaaclarke (approved):
The diff was a bit hard to parse, but it looks like all the old tests are still intact, just refactored.

script/tool/lib/src/drive_examples_command.dart
stuartmorgan: Done.

script/tool/test/drive_examples_command_test.dart
stuartmorgan: This test was copypasta'd (it was supposed to be the Android version of the test above but I never actually changed it); fixed. I also did a sweep and fixed all tests that asserted `ToolExit`s to actually check the output as well so that they aren't failing for the wrong reasons (which was true of several more tests).

script/tool/lib/src/drive_examples_command.dart
stuartmorgan: Done

script/tool/lib/src/drive_examples_command.dart
stuartmorgan: I thought about adding more control over devices via flags, but I figured that absent anyone actually wanting that functionality I'd just do the simplest thing for now, and we can expand later (e.g., adding device choice flags to the command itself) if needed.

script/tool/lib/src/drive_examples_command.dart
stuartmorgan: For now relying on the indirect testing is fine; if we end up needing this in more places and pull this out as a utility method we can split it apart and do direct unit testing.

stuartmorgan:
Landing on "red" since `submit-queue` isn't updating correctly.

PR #4111, opened 2021-06-28
[flutter_plugin_tools] Restructure version-check

stuartmorgan:
Combines the two different aspects of version-checking into a single looping structure based on plugins, using the new base command, rather than one operating on plugins as controlled by the usual flags and the other operating on a git list of changed files.

Also simplifies the determination of the new version by simply checking the file, rather than querying git for the HEAD state of the file. Tests setup is simplified since we no longer need to set up nearly as much fake `git` output.

Minor changes to base commands:
- Move indentation up to PackageLoopingCommand so that it is consistent across commands
- Add a new post-loop command for any cleanup, which is needed by version-check
- Change the way the GitDir instance is managed by the base PluginCommand, so that it's always cached even when not overridden, to reduce duplicate work and code.

Part of https://github.com/flutter/flutter/issues/83413

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/common/plugin_command.dart
+++ b/script/tool/lib/src/common/plugin_command.dart
@@ -76,8 +76,8 @@ abstract class PluginCommand extends Command<void> {
   /// This can be overridden for testing.
   final ProcessRunner processRunner;
 
-  /// The git directory to use. If unset, [getGitDir] uses the packages
-  /// directory's enclosing repository.
+  /// The git directory to use. If unset, [gitDir] populates it from the
+  /// packages directory's enclosing repository.
   ///
   /// This can be mocked for testing.
   GitDir? _gitDir;
@@ -102,7 +102,7 @@ abstract class PluginCommand extends Command<void> {
   }
 
   /// Returns the [GitDir] containing [packagesDir].
-  Future<GitDir> getGitDir() async {
+  Future<GitDir> get gitDir async {
     GitDir? gitDir = _gitDir;
     if (gitDir != null) {
       return gitDir;
@@ -309,7 +309,7 @@ abstract class PluginCommand extends Command<void> {
     final String baseSha = getStringArg(_kBaseSha);
 
     final GitVersionFinder gitVersionFinder =
-        GitVersionFinder(await getGitDir(), baseSha);
+        GitVersionFinder(await gitDir, baseSha);
     return gitVersionFinder;
   }
 
--- a/script/tool/lib/src/common/plugin_command.dart
+++ b/script/tool/lib/src/common/plugin_command.dart
@@ -20,8 +20,8 @@ abstract class PluginCommand extends Command<void> {
   PluginCommand(
     this.packagesDir, {
     this.processRunner = const ProcessRunner(),
-    this.gitDir,
-  }) {
+    GitDir? gitDir,
+  }) : _gitDir = gitDir {
     argParser.addMultiOption(
       _pluginsArg,
       splitCommas: true,
@@ -76,10 +76,11 @@ abstract class PluginCommand extends Command<void> {
   /// This can be overridden for testing.
   final ProcessRunner processRunner;
 
-  /// The git directory to use. By default it uses the parent directory.
+  /// The git directory to use. If unset, [getGitDir] uses the packages
+  /// directory's enclosing repository.
   ///
   /// This can be mocked for testing.
-  final GitDir? gitDir;
+  GitDir? _gitDir;
 
   int? _shardIndex;
   int? _shardCount;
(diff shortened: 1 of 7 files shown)


script/tool/lib/src/common/plugin_command.dart
```diff
@@ -100,6 +101,26 @@ abstract class PluginCommand extends Command<void> {
     return _shardCount!;
   }
 
+  /// Returns the [GitDir] containing [packagesDir].
+  Future<GitDir> getGitDir() async {
```
gaaclarke, line 105: nit: This should be a property.

script/tool/lib/src/common/plugin_command.dart
```diff
@@ -76,10 +76,11 @@ abstract class PluginCommand extends Command<void> {
   /// This can be overridden for testing.
   final ProcessRunner processRunner;
 
-  /// The git directory to use. By default it uses the parent directory.
+  /// The git directory to use. If unset, [getGitDir] uses the packages
+  /// directory's enclosing repository.
   ///
   /// This can be mocked for testing.
-  final GitDir? gitDir;
+  GitDir? _gitDir;
```
gaaclarke, line 83: This should be a late final non-null field, no?

script/tool/lib/src/version_check_command.dart
```diff
@@ -100,175 +104,175 @@ class VersionCheckCommand extends PluginCommand {
...
 
-    final List<String> changedPubspecs =
-        await gitVersionFinder.getChangedPubSpecs();
+  @override
+  Future<List<String>> runForPackage(Directory package) async {
+    final List<String> errors = <String>[];
 
-    final List<String> badVersionChangePubspecs = <String>[];
+    final Pubspec? pubspec = _tryParsePubspec(package);
+    Pubspec.parse(package.childFile('pubspec.yaml').readAsStringSync());
```
gaaclarke, line 118: Why are you reading a string then throwing it away?  To force a lazy value?  In the original code it used this value.

script/tool/lib/src/version_check_command.dart
```diff
@@ -100,175 +104,175 @@ class VersionCheckCommand extends PluginCommand {
...
+    final Version? currentPubspecVersion = pubspec.version;
+    if (currentPubspecVersion == null) {
+      printError('${indentation}No version found in pubspec.yaml. A package '
+          'that intentionally has no version should be marked '
+          '"publish_to: none".');
+      errors.add('No pubspec.yaml version.');
+      return errors; // No remaining checks make sense.
+    }
+
+    if (!await _checkVersionChange(package, pubspec: pubspec)) {
```
gaaclarke, line 138: nit: The verb "check" doesn't suggest what the return value actually means.  `if (!await _isVersionChangeValid(package, pubspec: pubspec))` would be easier to verify when reading the code.

script/tool/lib/src/version_check_command.dart
```diff
@@ -100,175 +104,175 @@ class VersionCheckCommand extends PluginCommand {
...
+      errors.add('Disallowed version change.');
+    }
+
+    if (!(await _checkVersionsMatch(package, pubspec: pubspec))) {
```
gaaclarke, line 142: Same comment about "check", how about `_doVersionsMatch`?

script/tool/lib/src/version_check_command.dart
```diff
@@ -100,175 +104,175 @@ class VersionCheckCommand extends PluginCommand {
...
+        return Version.none;
+    }
+  }
 
-      if (sourceVersion == headVersion) {
-        print('${indentation}No version change.');
-        continue;
-      }
+  /// Returns the version of [package] from git at the base comparison hash.
+  Future<Version?> _getPreviousVersionFromGit(
```
gaaclarke, line 177: nit: property

gaaclarke (changes requested):
Everything looks good.  I have a few nits.  The only thing I think that should change is the verb "check" in method names.

script/tool/lib/src/common/plugin_command.dart
stuartmorgan: It's used as a cache, so cannot be late final.

script/tool/lib/src/common/plugin_command.dart
stuartmorgan: Changed. I wasn't aware that properties could be async.

script/tool/lib/src/version_check_command.dart
stuartmorgan: I missed removing it when converting the codepath to using _tryParsePubspec. Removed. (Probably it was autoformatted with this as the RHS of an assignment at an interim refactoring stage, then I accidentally deleted the first half and the second half auto-formatted back as a pointless statement.)

script/tool/lib/src/version_check_command.dart
stuartmorgan: Changed to _hasValidVersionChange (the argument is a package, not a version, so it has a change, rather than being a change).

script/tool/lib/src/version_check_command.dart
stuartmorgan: Changed to _hasConsistentVersion (for similar reasons; the arguments are not version being compared).

script/tool/lib/src/version_check_command.dart
```diff
@@ -100,175 +104,175 @@ class VersionCheckCommand extends PluginCommand {
...
+  @override
+  Future<void> completeRun() async {
+    _pubVersionFinder.httpClient.close();
+  }
+
+  /// Returns the previous published version of [package].
+  ///
+  /// [packageName] must be the actual name of the package as published (i.e.,
+  /// the name from pubspec.yaml, not the on disk name if different.)
+  Future<Version?> _getPreviousVersionFromPub(String packageName) async {
```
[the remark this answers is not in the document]
stuartmorgan: Changed to fetch. However:
a) this is not in any way a property of the command, but an attributed of the provided package, and
b) properties can't take arguments.

script/tool/lib/src/version_check_command.dart
stuartmorgan: Same issue.

gaaclarke (approved):
lgtm

PR #4114, opened 2021-06-29
[in_app_purchase] Added priceCurrencySymbol to SkuDetailsWrapper

renefloor:
This PR adds priceCurrencySymbol to the SkuDetailsWrapper. It uses the java function `Currency.getInstance(currencyCode).getSymbol()` to get the symbol.
We could change the platform_interface to also expose this in the ProductDetails.

Part of flutter/flutter#83984 (other part can be the platform_interface).

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Note that unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [X] I signed the [CLA].
- [X] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I [updated pubspec.yaml](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates) with an appropriate new version according to the [pub versioning philosophy].
- [X] I updated CHANGELOG.md to add a description of the change.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[the auto-formatter]: https://github.com/flutter/plugins/blob/master/script/tool/README.md#format-code

PR #4115, opened 2021-06-29
[in_app_purchase] Add currencySymbol to ProductDetails

renefloor:
Adds currencySymbol to ProductDetails.

With PR #4114 both Android and iOS have a currency symbol. This PR adds the currency symbol to the platform interface so it can be exposed in the app facing package.

Part of flutter/flutter#83984

This is a breaking change because it has a new required parameter in the constructor. I could make it optional, but that would be suboptimal for usage. I'm not sure if there is a migration guid needed, but it is stated in the changelog what is changed.

## Pre-launch Checklist

- [X] I signed the [CLA].


BeMacized:
LGTM!

BeMacized (approved):
LGTM!

renefloor:
@stuartmorgan The publishable task fails because it is a breaking change and that is strongly discouraged.
I could make this not a breaking change by making the property optional, but that's not what the user of the package is expecting and this interface is only being used by our 2 platform implementations.

stuartmorgan:
> @stuartmorgan The publishable task fails because it is a breaking change and that is strongly discouraged.

Yes, that's https://github.com/flutter/flutter/issues/85391

> I could make this not a breaking change by making the property optional, but that's not what the user of the package is expecting

Can you elaborate on what that means? If you're just referring to the type of the accessor you could mask that by defaulting it to `''`, and then ensuring that we always fill it in. Not ideal, but it avoids a breaking change throughout the package just to make a single type non-nullable.

> and this interface is only being used by our 2 platform implementations.

This is inconsistent with the statement above. Either this class ends up as part of the public interface of the app-facing packages, in which case this is by definition not true because literally anyone could be using this interface, or it's not, in which case the details of the class are not relevant to clients.

renefloor:
> If you're just referring to the type of the accessor you could mask that by defaulting it to '', and then ensuring that we always fill it in.

Indeed, I meant that it is required in the constructor, but by adding a default it is not required. I'll make it optional that way.

> Either this class ends up as part of the public interface of the app-facing packages, in which case this is by definition not true because literally anyone could be using this interface, or it's not, in which case the details of the class are not relevant to clients.

In the app you (normally) only read the ProductDetails objects and you don't create them, so for reading them it is not breaking. But indeed it is possible to create them as well, so it would be a breaking change for the app facing packages.

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/types/product_details.dart
```diff
@@ -42,4 +43,7 @@ class ProductDetails {
   /// The currency code for the price of the product.
   /// Based on the price specified in the App Store Connect or Sku in Google Play console based on the platform.
   final String currencyCode;
+
+  ///The currency symbol for the locale, e.g. $ for US locale.
```
mvanbeusekom, line 47: Turns out in some cases (applies mainly on Android) the currency symbol cannot be determined. The current Java tooling used (https://docs.oracle.com/javase/7/docs/api/java/util/Currency.html#getSymbol()) returns the ISO 4217 currency code when the symbol cannot be determined. It would be good to mention this in the documentation.

Also a small nit, the documentation is missing a space after the `///`.

Maybe we can change the documentation to something like below. Of course this means that we need to ensure this will work on all platforms in the same way.
```suggestion
  /// The currency symbol for the locale, e.g. $ for US locale.
  ///
  /// When the currency symbol cannot be determined, the ISO 4217 currency code is returned. 
```

mvanbeusekom (changes requested):

packages/in_app_purchase/in_app_purchase_platform_interface/lib/src/types/product_details.dart
renefloor: @mvanbeusekom On iOS I can't find what it does 'if not found', maybe it just supports all currency symbols?
https://developer.apple.com/documentation/foundation/nslocale/1642814-currencysymbol

mvanbeusekom: Maybe we can add a simple check, if the `currencySymbol` is `nil` or empty we replace it with the `currencyCode`?

mvanbeusekom (approved):
LGTM

PR #3194, opened 2020-10-23
[image_picker] fix camera on Android 11

Bubu:
## Description

resolveActivity(intent) might return false on android 11+ even though startActivity(intent) would succeed.
Instead of introducing a <queries> entry we can just try to start the
activity and catch the exception if that doesn't work.

See here for a longer form explanation: https://cketti.de/2020/09/03/avoid-intent-resolveactivity/

## Related Issues
Fixes https://github.com/flutter/flutter/issues/85599
flutter/flutter#62669

Some more info on the android 11 behaviour change: https://github.com/flutter/flutter/issues/63727#issuecomment-690070820

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #4117, opened 2021-06-30
[image_picker] Fixed IOException when cache directory is removed

BeMacized:
Currently, the image picker plugin throws an IOException when trying to pick an image, if the cache directory does not exist. 
This PR makes it so the cache directory is created in case it does not exist.

Relevant issue:
- flutter/flutter#84251

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/image_picker/image_picker/CHANGELOG.md
+++ b/packages/image_picker/image_picker/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.8.1+3
+
+* Fix image picker causing a crash when the cache directory is deleted.
+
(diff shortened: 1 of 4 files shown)


renefloor (commented):
It looks good to me, but I wonder if we shouldn't handle the mockStaticFile in setup and teardown as it is now used in a lot of tests like I suggested in this other PR: https://github.com/flutter/plugins/pull/3194#issuecomment-869391087

renefloor (approved):
LGTM

PR #4118, opened 2021-06-30
[flutter_plugin_tools] Add a summary for successful runs

stuartmorgan:
Add a summary to the end of successful runs for everything using the new looping base command, similar to what we do for summarizing failures. This will make it easy to manually check results for PRs that we know should be changing the set of run packages (adding a new package, adding a new test type to a package, adding a new test type to the tool), as well as spot-checking when we see unexpected results (e.g., looking back and why a PR didn't fail CI when we discover that it should have).

To support better surfacing skips, this restructures the return value of `runForPackage` to have "skip" as one of the options. As a result of it being a return value, packages that used `printSkip` to indicate that *parts* of the command were being skipped have been changed to no longer do that.

Fixes https://github.com/flutter/flutter/issues/85626

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/common/package_looping_command.dart
+++ b/script/tool/lib/src/common/package_looping_command.dart
@@ -212,6 +212,9 @@ abstract class PackageLoopingCommand extends PluginCommand {
     final String skippedWarningSummary =
         runWarningCount > 0 ? ' ($skippedWarningCount with warnings)' : '';
     print('------------------------------------------------------------');
+    if (hasLongOutput) {
+      _printPerPackageRunOverview(packages);
+    }
     print(
         'Ran for ${packages.length - skipCount} package(s)$runWarningSummary');
     if (skipCount > 0) {
@@ -220,6 +223,31 @@ abstract class PackageLoopingCommand extends PluginCommand {
     if (_otherWarningCount > 0) {
       print('$_otherWarningCount warnings not associated with a package');
     }
+  }
+
+  /// Prints a one-line-per-package overview of the run results for each
+  /// package.
+  void _printPerPackageRunOverview(List<Directory> packages) {
+    print('Run overview:');
+    for (final Directory package in packages) {
+      final bool hadWarning = _packagesWithWarnings.contains(package);
+      Colorize summary;
+      if (_skippedPackages.contains(package)) {
+        if (hadWarning) {
+          summary = Colorize('skipped (with warning)')..lightYellow();
+        } else {
+          summary = Colorize('skipped')..darkGray();
+        }
+      } else {
+        if (hadWarning) {
+          summary = Colorize('ran (with warning)')..yellow();
+        } else {
+          summary = Colorize('ran')..green();
+        }
+      }
+      print('  ${getPackageDescription(package)} - $summary');
+    }
+    print('');
   }
 
   /// Prints a summary of all of the failures from [results].
--- a/script/tool/lib/src/common/package_looping_command.dart
+++ b/script/tool/lib/src/common/package_looping_command.dart
@@ -11,6 +11,48 @@ import 'core.dart';
 import 'plugin_command.dart';
 import 'process_runner.dart';
 
+/// Possible outcomes of a command run for a package.
+enum RunState {
+  /// The command succeeded for the package.
+  succeeded,
+
+  /// The command was skipped for the package.
+  skipped,
+
+  /// The command failed for the package.
+  failed,
+}
+
+/// The result of a [runForPackage] call.
+class PackageResult {
+  /// A successful result.
+  PackageResult.success() : this._(RunState.succeeded);
+
+  /// A run that was skipped as explained in [reason].
+  PackageResult.skip(String reason)
+      : this._(RunState.skipped, <String>[reason]);
+
+  /// A run that failed.
+  ///
+  /// If [errors] are provided, they will be listed in the summary, otherwise
+  /// the summary will simply show that the package failed.
+  PackageResult.fail([List<String> errors = const <String>[]])
+      : this._(RunState.failed, errors);
+
+  const PackageResult._(this.state, [this.details = const <String>[]]);
+
+  /// The state the package run completed with.
+  final RunState state;
+
+  /// Information about the result:
+  /// - For `succeeded`, this is empty.
+  /// - For `skipped`, it contains a single entry describing why the run was
+  ///   skipped.
+  /// - For `failed`, it contains zero or more specific error details to be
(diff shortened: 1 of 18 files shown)


script/tool/lib/src/common/package_looping_command.dart
```diff
@@ -167,4 +195,80 @@ abstract class PackageLoopingCommand extends PluginCommand {
...
+    final int skippedWarningCount = _skippedPackagesWithWarnings.length;
+    final int runWarningCount =
+        _packagesWithWarnings.length - skippedWarningCount;
+
+    final String runWarningSummary =
+        runWarningCount > 0 ? ' ($runWarningCount with warnings)' : '';
+    final String skippedWarningSummary =
+        runWarningCount > 0 ? ' ($skippedWarningCount with warnings)' : '';
+    print('------------------------------------------------------------');
+    if (hasLongOutput) {
```
gaaclarke, line 215: Is `hasLongOutput` suppose to mean "verbose" in this case?  We had another place in the code where the same verbiage meant "compress the output because it's long" I thought.

gaaclarke (approved):

script/tool/lib/src/common/package_looping_command.dart
```diff
@@ -22,6 +22,18 @@ abstract class PackageLoopingCommand extends PluginCommand {
...
+  Set<Directory> _skippedPackages = <Directory>{};
+
+  /// Packages that had at least one [logWarning] call.
+  Set<Directory> _packagesWithWarnings = <Directory>{};
+
+  /// Number of warnings that happened outside of a [runForPackage] call.
+  int _otherWarningCount = 0;
+
+  /// The package currently being run by [runForPackage].
+  Directory? _currentPackage;
```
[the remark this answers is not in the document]
stuartmorgan: > It would be better off as a parameter to logSkip and logWarning.

I strongly considered that, since this approach feels a bit sketchy, but in practice that would be worse. With this version, the possible bugs are contained entirely within this one base class, which is short and has unit tests ensuring that these work correctly. With the parameter version, however, it's possible to write things like:

`logWarning(message, example)` // Oops, this was supposed to be `package`, not the `example` subdirectory.

and thus causing the warning not be summarized anywhere.

For `logSkip`, the more I think about the summary, the more I think I should actually make it a return value option instead. Originally I had created printSkip for logging any case where something is being skipped, which for some commands might be, say, a subset of the examples. But with the summary we don't actually want to allow that, so I need to rework the return value. I'll land out the outstanding other conversions, then change the value to a return class with state (pass/skip/error)+error messages.

> Keeping it around could lead to some weird bugs at some point because the coupling between logSkip and _currentPackage isn't obvious to the client.

For `logSkip` I agree. For `logWarning` I think that's actually a feature; a command can just logging warnings whenever it wants to as output, and doesn't have to care about what the base class might be doing with them. How they are tracked and summarized is entirely an implementation detail of the base.

script/tool/lib/src/common/package_looping_command.dart
stuartmorgan: It means that in practice, the output of a given `runForPackage` call is likely to be long. The existing use is that it changes the header format from a single line to the ====-wrapped banner style, to make it easier to skim for a run break.

stuartmorgan: (I deliberately called it "long" rather than "verbose" because I don't want to imply that underlying commands are being called with a `-v` or similar flag; some commands print a lot even when you don't make them explicitly verbose.)

script/tool/lib/src/common/package_looping_command.dart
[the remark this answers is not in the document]
stuartmorgan: This PR has been updated to eliminate `logSkip` by changing the return from a run to a small data class and making "skipped" a first-class result rather than a possibly-partial bit based on a log.

Commands that used to log formal skips for subsets of the command while running other subsets now just use simple print statements to indicate what is and isn't being run, and "skip" is reserved for cases where nothing is run.

script/tool/lib/src/common/package_looping_command.dart
```diff
@@ -45,7 +84,7 @@ abstract class PackageLoopingCommand extends PluginCommand {
   /// be included in the final error summary (e.g., a command that only has a
   /// single failure mode), or strings that should be listed for that package
   /// in the final summary. An empty list indicates success.
-  Future<List<String>> runForPackage(Directory package);
+  Future<PackageResult> runForPackage(Directory package);
```
gaaclarke, line 87: Nice improvement.

stuartmorgan: Yes, the fact that I had to give the magic values `[]` and `['']` special names in the original API should have been more of a red flag to me than it was.

script/tool/lib/src/common/package_looping_command.dart
```diff
@@ -167,4 +222,86 @@ abstract class PackageLoopingCommand extends PluginCommand {
...
+    print('');
+  }
+
+  /// Prints a summary of all of the failures from [results].
+  void _printFailureSummary(
+      List<Directory> packages, Map<Directory, PackageResult> results) {
+    const String indentation = '  ';
+    printError(failureListHeader);
+    for (final Directory package in packages) {
+      final PackageResult result = results[package]!;
```
gaaclarke, line 293: nit: I don't think you need to pass in packages here, you can just loop over `results.values`

script/tool/lib/src/common/package_looping_command.dart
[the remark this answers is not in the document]
gaaclarke: This is a huge improvement.  You could eliminate `_currentPackage` if you created a way for PackageResult to carry the warnings too.  It's not a huge deal, `_currentPackage` bugs me a bit since it's a looping parameter in an instance variable.

gaaclarke (approved):
LGTM, nice update.

script/tool/lib/src/common/package_looping_command.dart
stuartmorgan: Results is a map, and thus doesn't have a guaranteed order; it's important that the summary be in the same order as the packages for ease of referencing up to the details.

script/tool/lib/src/common/package_looping_command.dart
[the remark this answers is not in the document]
stuartmorgan: It's valid to print warnings outside the loop though; I don't want to have two different ways of handling warnings in the subclasses.

PR #4121, opened 2021-06-30
Add `buildViewWithTextDirection` to platform interface

bparrishMines:
Adds `buildViewWithTextDirection` to the platform interface. Originally, this was only a method in the `MethodChannel` implementation.

Tests for this can only be added `google_maps_flutter` once this is published. Currently they're in https://github.com/flutter/plugins/pull/4082.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_flutter_platform.dart
```diff
@@ -356,4 +356,38 @@ abstract class GoogleMapsFlutterPlatform extends PlatformInterface {
...
     throw UnimplementedError('buildView() has not been implemented.');
   }
+
+  /// Returns a widget displaying the map view.
+  ///
+  /// This method is similar to [buildView], but contains a parameter for
+  /// platforms that require a text direction.
+  ///
+  /// Default behavior passes all parameters except `textDirection` to
+  /// [buildView].
```
stuartmorgan, line 366: How about adding "This is for backward compatibility with existing implementations. Platforms that use the text direction should override this as the primary implementation, and delegate to it from `buildView`." That makes it clear why the default is (un-intuitively) to delegate in the direction that discards information, rather than the more usual pattern of the one with fewer arguments calling the other way with a default value.

packages/google_maps_flutter/google_maps_flutter_platform_interface/test/platform_interface/google_maps_flutter_platform_test.dart
```diff
@@ -34,6 +38,23 @@ void main() {
     test('Can be extended', () {
       GoogleMapsFlutterPlatform.instance = ExtendsGoogleMapsFlutterPlatform();
     });
+
+    test(
+      'default implementation buildViewWithTextDirection does not throw an exception',
```
stuartmorgan, line 43: s/does not throw an exception/delegates to buildView/

stuartmorgan, line 43: Typo: missing an "of" after "implementation"

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
```diff
@@ -520,7 +516,7 @@ class MethodChannelGoogleMapsFlutter extends GoogleMapsFlutterPlatform {
         },
       );
     }
-    return buildView(
+    return super.buildView(
```
stuartmorgan, line 519: Why super? Super throws, unless I'm missing something.

I think it would be clearer to just move the current `buildView` implementation here, and then have `buildView` be a simple passthrough to `buildViewWithTextDirection` with a default paramater.

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
bparrishMines: Woops. This was suppose to be `super.buildViewWithTextDirection`. I changed it to your suggestion

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_flutter_platform.dart
```diff
@@ -356,4 +356,40 @@ abstract class GoogleMapsFlutterPlatform extends PlatformInterface {
...
+  ///
+  /// Default behavior passes all parameters except `textDirection` to
+  /// [buildView]. This is for backward compatibility with existing
+  /// implementations. Platforms that use the text direction should override
+  /// this as the primary implementation, and delegate to it from buildView.
+  Widget buildViewWithTextDirection(
+    int creationId,
+    PlatformViewCreatedCallback onPlatformViewCreated, {
+    required CameraPosition initialCameraPosition,
+    TextDirection textDirection = TextDirection.ltr,
```
stuartmorgan, line 373: Optional: Should this be required, so that callers have to think about it (and set a default themselves if they don't have a good value) instead of just not noticing?

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
```diff
@@ -456,16 +456,12 @@ class MethodChannelGoogleMapsFlutter extends GoogleMapsFlutterPlatform {
...
-  /// This method includes a parameter for platforms that require a text
-  /// direction. For example, this should be used when using hybrid composition
-  /// on Android.
+  @override
   Widget buildViewWithTextDirection(
     int creationId,
     PlatformViewCreatedCallback onPlatformViewCreated, {
     required CameraPosition initialCameraPosition,
-    required TextDirection textDirection,
+    TextDirection textDirection = TextDirection.ltr,
```
stuartmorgan, line 464: See comment on parent class about whether this should be `required` (with this default in `buildView` below)

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
```diff
@@ -519,44 +516,7 @@ class MethodChannelGoogleMapsFlutter extends GoogleMapsFlutterPlatform {
...
-      'initialCameraPosition': initialCameraPosition.toMap(),
-      'options': mapOptions,
-      'markersToAdd': serializeMarkerSet(markers),
-      'polygonsToAdd': serializePolygonSet(polygons),
-      'polylinesToAdd': serializePolylineSet(polylines),
-      'circlesToAdd': serializeCircleSet(circles),
-      'tileOverlaysToAdd': serializeTileOverlaySet(tileOverlays),
-    };
-    if (defaultTargetPlatform == TargetPlatform.android) {
+    } else if (defaultTargetPlatform == TargetPlatform.android) {
```
stuartmorgan, line 519: With the new combined structure it would be clearer to do

```dart
  if (useAndroidViewSurface) {
    ...
  } else {
    ...
  }
} else if ...
```

instead of 

```dart
if (defaultTargetPlatform == TargetPlatform.android &&
        useAndroidViewSurface) {
  ...
} else if (defaultTargetPlatform == TargetPlatform.android) {
  ...
} else if ...
```

stuartmorgan (approved):
LGTM with nits

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_flutter_platform.dart
bparrishMines: I think making it required makes more sense. I changed it.

PR #4128, opened 2021-07-01
Exclude arm64 simulators from google_sign_in example project

jmagman:
Workaround https://github.com/flutter/flutter/issues/85713 to unblock the tree.
Exclude `arm64` for all pods, as well as the app itself.

See also https://github.com/flutter/plugins/pull/4127.


stuartmorgan (approved):
LGTM

jmagman:
Merging to green the tree.

PR #4130, opened 2021-07-01
[flutter_plugin_tools] Remove most exitOnError:true usage

stuartmorgan:
The intent of package-looping commands is that we always want to run them for all the targeted packages, accumulating failures, and then report them all at the end, so we don't end up with the problem of only finding errors one at a time. However, some of them were using `exitOnError: true` for the underlying commands, causing the first error to be fatal to the test.

This PR:
- Removes all use of `exitOnError: true` from the package-looping commands. (It's still used in `format`, but fixing that is a larger issue, and less important due to the way `format` is currently structured.)
- Fixes the mock process runner used in our tests to correctly simulate `exitOrError: true`; the fact that it didn't was hiding this problem in test (e.g., `drive-examples` had a test that asserted that all failures were summarized correctly, which passed because in tests it was behaving as if `exitOnError` were false)
- Adjusts the mock process runner to allow setting a list of mock result for a specific executable instead of one result for all calls to anything, in order to fix some tests that were broken by the two changes above and unfixable without this (e.g., a test of a command that had been calling one executable with `exitOnError: true` then another with ``exitOnError: false`, where the test was asserting things about the second call failing, which only worked because the first call's failure wasn't actually checked). To limit the scope of the PR, the old method of setting a single result for all calls is still supported for now as a fallback.
- Fixes the fact that the mock `run` and `runAndStream` had opposite default exit code behavior when no mock process was set (since that caused me a lot of confusion while fixing the above until I figured it out).

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/CHANGELOG.md
+++ b/script/tool/CHANGELOG.md
@@ -3,6 +3,8 @@
 - Modified the output format of many commands
 - **Breaking change**: `firebase-test-lab` no longer supports `*_e2e.dart`
   files, only `integration_test/*_test.dart`.
+- Fixed some cases where a failure in a command for a single package would
+  immidiately abort the test.
 
 ## 0.3.0
 
--- a/script/tool/CHANGELOG.md
+++ b/script/tool/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Add a summary to the end of successful command runs for commands using the
   new output format.
 - Fixed some cases where a failure in a command for a single package would
-  immidiately abort the test.
+  immediately abort the test.
 
 ## 0.3.0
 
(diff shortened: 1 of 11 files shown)


script/tool/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
   files, only `integration_test/*_test.dart`.
 - Add a summary to the end of successful command runs for commands using the
   new output format.
+- Fixed some cases where a failure in a command for a single package would
+  immidiately abort the test.
```
jmagman, line 9:
```suggestion
  immediately abort the test.
```

jmagman (approved):
LGTM with spelling nit

PR #1022, opened 2019-01-02
[camera] Add serial dispatch_queue for camera plugin to avoid blocking the UI

cvolzke4:
See https://github.com/flutter/flutter/issues/25959 for more information.

PR #4140, opened 2021-07-07
[camera] Run iOS methods on UI thread by default

renefloor:
In flutter/flutter#25959 we found that not everything in the iOS camera implementation should be run on the UI thread, as that blocks the app UI. The solution was to run everything on a different thread.

In flutter/flutter#52578 we found that we should not run everything on a different thread, because for example sending the result should be done on the UI thread.

In PR #1022 it is decided that things should run on UI thread by default and exceptions are made for heavy operations.

In flutter/flutter#25959 the author stated that "it's recommended that AVCaptureSession operations be called on a dedicated serial DispatchQueue". That means that `initWithCameraName`, `start` and `stop` need to be run on a different thread, but I'm not sure if these are the only ones blocking the UI.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/ios/Classes/CameraPlugin.m
+++ b/packages/camera/camera/ios/Classes/CameraPlugin.m
@@ -1315,13 +1315,6 @@ NSString *const errorMethod = @"error";
     _dispatchQueue = dispatch_queue_create("io.flutter.camera.dispatchqueue", NULL);
   }
 
-  // Invoke the plugin on another dispatch queue to avoid blocking the UI.
-  dispatch_async(_dispatchQueue, ^{
-    [self handleMethodCallAsync:call result:result];
-  });
-}
-
-- (void)handleMethodCallAsync:(FlutterMethodCall *)call result:(FlutterResult)result {
   if ([@"availableCameras" isEqualToString:call.method]) {
     if (@available(iOS 10.0, *)) {
       AVCaptureDeviceDiscoverySession *discoverySession = [AVCaptureDeviceDiscoverySession
@@ -1358,27 +1351,30 @@ NSString *const errorMethod = @"error";
     NSString *cameraName = call.arguments[@"cameraName"];
     NSString *resolutionPreset = call.arguments[@"resolutionPreset"];
     NSNumber *enableAudio = call.arguments[@"enableAudio"];
-    NSError *error;
-    FLTCam *cam = [[FLTCam alloc] initWithCameraName:cameraName
-                                    resolutionPreset:resolutionPreset
-                                         enableAudio:[enableAudio boolValue]
-                                         orientation:[[UIDevice currentDevice] orientation]
-                                       dispatchQueue:_dispatchQueue
-                                               error:&error];
+    dispatch_async(_dispatchQueue, ^{
+      NSError *error;
+      FLTCam *cam = [[FLTCam alloc] initWithCameraName:cameraName
+                                      resolutionPreset:resolutionPreset
+                                           enableAudio:[enableAudio boolValue]
+                                           orientation:[[UIDevice currentDevice] orientation]
+                                         dispatchQueue:self->_dispatchQueue
+                                                 error:&error];
+      dispatch_async(dispatch_get_main_queue(), ^{
+        if (error) {
+          result(getFlutterError(error));
+        } else {
+          if (self->_camera) {
+            [self->_camera close];
+          }
+          int64_t textureId = [self->_registry registerTexture:cam];
+          self->_camera = cam;
 
-    if (error) {
-      result(getFlutterError(error));
-    } else {
-      if (_camera) {
-        [_camera close];
-      }
-      int64_t textureId = [_registry registerTexture:cam];
-      _camera = cam;
-
-      result(@{
-        @"cameraId" : @(textureId),
+          result(@{
+            @"cameraId" : @(textureId),
+          });
+        }
       });
-    }
+    });
   } else if ([@"startImageStream" isEqualToString:call.method]) {
     [_camera startImageStreamWithMessenger:_messenger];
     result(nil);


packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -363,6 +358,7 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
     _resolutionPreset = getResolutionPresetForString(resolutionPreset);
   } @catch (NSError *e) {
     *error = e;
+    return nil;
   }
```
renefloor, line 362: @mvanbeusekom and @stuartmorgan If this doesn't return nil setting the error is useless. I assume this doesn't give real-life issues as the resolutionPreset is always set to something valid. Can I just include the return in this PR or should I make a separate issue and PR for it?

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
```diff
@@ -0,0 +1,101 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import camera;
+@import XCTest;
+@import AVFoundation;
+#import <OCMock/OCMock.h>
+
+@interface FLTThreadSafeFlutterResult : NSObject
```
mvanbeusekom, line 10: To fix the build, try to make this an extension interface, like so (similar to how it is defined in the `FLTThreadSafeFlutterResult.m` file):
```suggestion
@interface FLTThreadSafeFlutterResult ()
```

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Flutter/Flutter.h>
+
+/**
+ Wrapper object that always delivers the result on the main thread.
```
stuartmorgan, line 8: "Wrapper for FlutterResult that always [...]"

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+@interface FLTThreadSafeFlutterResult : NSObject
+
+/**
+ Initialize with a FlutterResult object.
```
stuartmorgan, line 13: Initializes

(per Google style)

stuartmorgan, line 13: For all comments, Google style is generally:
```
/**
 * Comments here.
 */
```
for C-style comments (with a `*` on every line)

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
...
+ @param result The FlutterResult object that the result will be given to.
+ */
+- (id _Nonnull)initWithResult:(FlutterResult _Nonnull)result;
```
stuartmorgan, line 16: Prefer `nonnull` and `nullable` whenever possible (which should be everywhere in this file)

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ Send a successful result without any data.
```
stuartmorgan, line 19: Sends

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+- (void)success;
+
+/**
+ Send a successful result with data.
```
stuartmorgan, line 24: Sends

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
...
+ @param data Result data that is send to the Flutter Dart side.
+ */
+- (void)successWithData:(id _Nonnull)data;
+
+/**
+ Send an error as result
```
stuartmorgan, line 30: Sends

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
...
+ @param error Error that will be send as FlutterError.
+ */
+- (void)error:(NSError* _Nonnull)error;
+
+/**
+ Send a FlutterError as result.
```
stuartmorgan, line 36: Sends

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
stuartmorgan, line 30: s/error/NSError/

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+- (void)errorWithCode:(NSString* _Nonnull)code
+              message:(NSString* _Nullable)message
+              details:(id _Nullable)details;
+
+/**
+ Send FlutterMethodNotImplemented as result.
```
stuartmorgan, line 43: Sends

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
...
+/**
+ Initialize with a FlutterResult object.
+ @param result The FlutterResult object that the result will be given to.
+ */
+- (id _Nonnull)initWithResult:(FlutterResult _Nonnull)result;
+
+/**
+ Send a successful result without any data.
+ */
+- (void)success;
```
stuartmorgan, line 21: These should all be verbs, so `sendSuccess`, etc.

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.m
```diff
@@ -0,0 +1,59 @@
...
+              details:(id _Nullable)details {
+  FlutterError* flutterError = [FlutterError errorWithCode:code message:message details:details];
+  [self send:flutterError];
+}
+
+- (void)notImplemented {
+  [self send:FlutterMethodNotImplemented];
+}
+
+- (void)send:(id _Nullable)result {
```
stuartmorgan, line 49: Needs a declaration comment here since it's not previously declared. All methods need comments, per style guide.

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
```diff
@@ -0,0 +1,101 @@
...
+  OCMStub([avCaptureSessionMock canSetSessionPreset:[OCMArg any]]).andReturn(YES);
+
+  _resultObject =
+      [[MockFLTThreadSafeFlutterResult alloc] initWithNotificationCenter:_notificationCenter];
+}
+
+- (void)tearDown {
+  // Put teardown code here. This method is called after the invocation of each test method in the
+  // class.
+}
```
stuartmorgan, line 71: Don't leave boilerplate in the test file.

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
```diff
@@ -0,0 +1,101 @@
...
+
+- (void)testCreate_ShouldCallResultOnMainThread {
+  // Setup method call
```
stuartmorgan, line 74: Nit: // Set up method call.

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
```diff
@@ -0,0 +1,101 @@
...
+@import camera;
+@import XCTest;
+@import AVFoundation;
+#import <OCMock/OCMock.h>
+
+@interface FLTThreadSafeFlutterResult ()
+@property(readonly, nonatomic) FlutterResult flutterResult;
+@end
+
+@interface MockFLTThreadSafeFlutterResult : FLTThreadSafeFlutterResult
```
stuartmorgan, line 14: Needs a comment explaining its role.

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
```diff
@@ -0,0 +1,101 @@
...
+@property(nonatomic, nullable) id receivedResult;
+@end
+
+@implementation MockFLTThreadSafeFlutterResult
+/**
+ Initialize with a notification center.
+ */
+- (id)initWithNotificationCenter:(NSNotificationCenter *)notificationCenter {
+  self = [super init];
+  _notificationCenter = notificationCenter;
```
stuartmorgan, line 25: I'm confused about the use of the notification center. Why not just have the mock have a public `expectation` property that it fulfills when any result is called, and that tests can wait on?

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
```diff
@@ -0,0 +1,101 @@
...
+  return self;
+}
+
+/**
+ Called when result is successful. Sends "successWithData" to the notification center.
+ */
+- (void)successWithData:(id)data {
```
stuartmorgan, line 32: Shouldn't there be a thread assertion in this code?

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+- (void)errorWithCode:(NSString* _Nonnull)code
+              message:(NSString* _Nullable)message
+              details:(id _Nullable)details;
+
+/**
+ Send FlutterMethodNotImplemented as result.
+ */
+- (void)notImplemented;
+@end
```
stuartmorgan, line 46: Please unit tests this class.

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
```diff
@@ -0,0 +1,101 @@
...
+  XCTNSNotificationExpectation *notificationExpectation =
+      [[XCTNSNotificationExpectation alloc] initWithName:@"successWithData"
+                                                  object:nil
+                                      notificationCenter:_notificationCenter];
+
+  FlutterMethodCall *call = [FlutterMethodCall
+      methodCallWithMethodName:@"create"
+                     arguments:@{@"resolutionPreset" : @"medium", @"enableAudio" : @(1)}];
+
+  // Call handleMethodCall
```
stuartmorgan, line 84: Don't add comments that describe basic language mechanics.

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
```diff
@@ -0,0 +1,101 @@
...
+  [_camera handleMethodCallWithThreadSafeResult:call result:_resultObject];
+
+  // Don't expect a result yet
+  XCTAssertNil(_resultObject.receivedResult);
```
stuartmorgan, line 88: It's not clear to me why this is an assertion. Why would it be an error if the plugin were to respond synchronously?

stuartmorgan (changes requested):

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
renefloor: The create call does some heavy work which should not be on the UI thread, so if this responds synchronously that means the calls on the background thread are now done on the main thread, or that the function completes before the calls on the background thread are completed.

stuartmorgan: What if it's done on the background thread but the background thread work completes before the main thread is scheduled again to run this assert?

renefloor: Didn't expect that could happen, doesn't this run synchronously? If that can happen I'll remove the assert.

stuartmorgan: Hm, I guess even if the background thread work completes, the part that was kicked back to the main thread won't have been run yet. So it's fine to leave, as long as you add a comment explaining why you're asserting that it *can't* have finished, since that's unusual (usually callbacks can happen at any time, including synchronously).

packages/camera/camera/ios/Classes/CameraPlugin.m
renefloor: Chose to exclude it from this PR

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
renefloor: What do you mean?
"Sends an NSError as result" Or "Sends s/error/NSError/"?

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
renefloor: Can you have a look at my added comment?

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
renefloor: You are right, the notification center is not needed.

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
renefloor: What do you mean with the thread assertion? That we verify that it is not the main thread?

renefloor:
@stuartmorgan I resolved most of your comments and asked for clarifications on the others. Can you have a look at this PR again?

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
stuartmorgan: Looks good, thanks.

packages/camera/camera/example/ios/RunnerTests/CameraMethodChannelTests.m
stuartmorgan: I don't remember what I meant by having that comment in this location unfortunately; maybe I was forgetting the flow and thinking this was after the bounce back to the main thread.

However, the more general questions is still valid: the goal of this PR is to ensure that methods are getting callbacks on the main thread, so why isn't there any code in the test that calls into the camera handler with a result object controlled by the test, and actually asserts that the callback was on the main thread? Like this: https://github.com/flutter/plugins/blob/master/packages/local_auth/example/ios/RunnerTests/FLTLocalAuthPluginTests.m#L62-L64

It is not clear to me what in this test would deterministically fail without the fix from this PR.

renefloor: @stuartmorgan I have now added ThreadSafeFlutterResultTests that test that the result is always called on the main thread. These tests in CameraMethodChannelTests just verify that ThreadSafeFlutterResult is used, not how that is working.

PR #4145, opened 2021-07-08
[flutter_plugin_tools] Improve and test 'format'

stuartmorgan:
- Adds unit tests, as there are currently none.
- Adds more graceful failure handling.
- Adds an internal ignore list to skip files that don't need to be
  formatted that showed up during local testing.
- Adds a note explaining that it's intentially not using the new base
  command due to performance issues.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/format_command.dart
+++ b/script/tool/lib/src/format_command.dart
@@ -14,6 +14,11 @@ import 'common/core.dart';
 import 'common/plugin_command.dart';
 import 'common/process_runner.dart';
 
+const int _exitClangFormatFailed = 3;
+const int _exitFlutterFormatFailed = 4;
+const int _exitJavaFormatFailed = 5;
+const int _exitGitFailed = 6;
+
 final Uri _googleFormatterUrl = Uri.https('github.com',
     '/google/google-java-format/releases/download/google-java-format-1.3/google-java-format-1.3-all-deps.jar');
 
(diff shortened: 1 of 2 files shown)


stuartmorgan:
Originally this was going to be a PR to convert to PackageLoopingCommand, but that turned out to be terrible.

There's more that could be done, like adding flags to control different languages (e.g., to allow someone to format Dart+Java without having clang-format installed), but the more explicit failure handling and the unit testing gives us a better foundation for making future improvements as follow-up later.

script/tool/test/format_command_test.dart
```diff
@@ -0,0 +1,344 @@
...
+    processRunner.mockProcessesForExecutable['java'] = <io.Process>[
+      MockProcess.failing()
+    ];
+    Error? commandError;
+    final List<String> output = await runCapturingPrint(
+        runner, <String>['format'], errorHandler: (Error e) {
+      commandError = e;
+    });
+
+    expect(commandError, isA<ToolExit>());
```
ditman, line 142: Why not also expect that the tool is using retcode 5? (`const int _exitJavaFormatFailed = 5;`)

script/tool/lib/src/format_command.dart
```diff
@@ -79,66 +91,105 @@ class FormatCommand extends PluginCommand {
...
+      return RegExp('(?:^|$s)${p.joinAll(directories)}$s');
+    }
+
+    return files
+        .map((File file) => file.path)
+        .where((String path) =>
+            // Ignore files in build/ directories (e.g., headers of frameworks)
+            // to avoid useless extra work in local repositories.
+            !path.contains(
+                pathFragmentForDirectories(<String>['example', 'build'])) &&
```
ditman, line 181: Can these be implemented with some util from the path library? maybe [isWithin](https://api.flutter.dev/flutter/package-path_path/isWithin.html)? (probably not, it looks like it expects absolute paths)

ditman (approved):
This is great. It's great that you added semantic retcodes (and tests :P)

script/tool/test/format_command_test.dart
stuartmorgan: In practice we never actually use the exit code, so we've never bothered with testing them; I figured we could add tests later if we ever decide to do something with them. I use them in the code because it's easy, and makes the line self-documenting.

script/tool/lib/src/format_command.dart
stuartmorgan: I looked a bit and couldn't find anything that did arbitrary relate path containment.

PR #4150, opened 2021-07-11
[flutter_plugin_tools] Support format on Windows

stuartmorgan:
Allows `format` to run successfully on Windows:
- Ensures that no calls exceed the command length limit.
- Allows specifying a `java` path to make it easier to run without a system Java (e.g., by pointing to the `java` binary in an Android Studio installation).
- Adds clear error messages when `java` or `clang-format` is missing since it's very non-obvious what's wrong otherwise.

Bumps the version, which I intended to do in the previous PR but apparently didn't push to the PR.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/format_command.dart
+++ b/script/tool/lib/src/format_command.dart
@@ -8,7 +8,6 @@ import 'dart:io' as io;
 import 'package:file/file.dart';
 import 'package:http/http.dart' as http;
 import 'package:platform/platform.dart';
-import 'package:quiver/iterables.dart';
 
 import 'common/core.dart';
 import 'common/plugin_command.dart';
@@ -114,18 +113,9 @@ class FormatCommand extends PluginCommand {
         files, <String>{'.h', '.m', '.mm', '.cc', '.cpp'});
     if (clangFiles.isNotEmpty) {
       print('Formatting .cc, .cpp, .h, .m, and .mm files...');
-      final Iterable<List<String>> batches = partition(clangFiles, 100);
-      int exitCode = 0;
-      for (final List<String> batch in batches) {
-        batch.sort(); // For ease of testing; partition changes the order.
-        exitCode = await processRunner.runAndStream(
-            getStringArg('clang-format'),
-            <String>['-i', '--style=Google', ...batch],
-            workingDir: packagesDir);
-        if (exitCode != 0) {
-          break;
-        }
-      }
+      final int exitCode = await runBatched(
+          getStringArg('clang-format'), <String>['-i', '--style=Google'],
+          files: clangFiles);
       if (exitCode != 0) {
         printError(
             'Failed to format C, C++, and Objective-C files: exit code $exitCode.');
@@ -140,9 +130,9 @@ class FormatCommand extends PluginCommand {
         _getPathsWithExtensions(files, <String>{'.java'});
     if (javaFiles.isNotEmpty) {
       print('Formatting .java files...');
-      final int exitCode = await processRunner.runAndStream('java',
-          <String>['-jar', googleFormatterPath, '--replace', ...javaFiles],
-          workingDir: packagesDir);
+      final int exitCode = await runBatched(
+          'java', <String>['-jar', googleFormatterPath, '--replace'],
+          files: javaFiles);
       if (exitCode != 0) {
         printError('Failed to format Java files: exit code $exitCode.');
         throw ToolExit(_exitJavaFormatFailed);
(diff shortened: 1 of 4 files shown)


script/tool/lib/src/format_command.dart
```diff
@@ -212,4 +241,65 @@ class FormatCommand extends PluginCommand {
 
     return javaFormatterPath;
   }
+
+  /// Returns true if [command] can be run successfully.
+  Future<bool> _hasDependency(String command) async {
```
ditman, line 246: I don't know if we have minimum version restrictions for java/clang-formant, but this looks like a cool point to add that validation when required! Good stuff!

script/tool/lib/src/format_command.dart
```diff
@@ -212,4 +241,65 @@ class FormatCommand extends PluginCommand {
...
+        platform.isWindows ? windowsCommandLineMax : nonWindowsCommandLineMax;
+
+    // Compute the max length of the file argument portion of a batch.
+    // Add one to each argument's length for the space before it.
+    final int argumentTotalLength =
+        arguments.fold(0, (int sum, String arg) => sum + arg.length + 1);
+    final int batchMaxTotalLength =
+        commandLineMax - command.length - argumentTotalLength;
+
+    // Divide the list into batches that don't exceed the length.
```
ditman, line 280: Maybe split this to its own method, just to make _runBatched super straight-forward:

final batches = _createBatches(params);
for (batch in batches) ...

script/tool/test/format_command_test.dart
```diff
@@ -46,13 +46,39 @@ void main() {
...
         .toList();
   }
 
+  /// Returns a list of [count] relative paths to pass to [createFakePlugin]
+  /// with name [pluginName] such that each path will be 99 characters long
+  /// relative to [packagesDir].
+  ///
+  /// This is for each of testing batching, since it means each file will
+  /// consume 100 characters of the batch length.
+  List<String> _get99CharacterPathExtraFiles(String pluginName, int count) {
```
ditman, line 67: Fun xD

script/tool/test/format_command_test.dart
```diff
@@ -348,4 +494,95 @@ void main() {
...
+  test('Batches moderately long file lists on Windows', () async {
+    mockPlatform.isWindows = true;
+
+    const String pluginName = 'a_plugin';
+    // -1 since the command itself takes some length.
+    const int batchSize = (windowsCommandLineMax ~/ 100) - 1;
+
+    // Make the file list one file longer than would fit in the batch.
+    final List<String> batch1 =
+        _get99CharacterPathExtraFiles(pluginName, batchSize + 1);
```
ditman, line 507: No need to `-1` batchSize above, then?

script/tool/test/format_command_test.dart
```diff
@@ -348,4 +494,95 @@ void main() {
...
+                '$pluginName\\$extraFile',
+              ],
+              packagesDir.path),
+        ));
+  });
+
+  test('Does not batch moderately long file lists on non-Windows', () async {
+    const String pluginName = 'a_plugin';
+    // -1 since the command itself takes some length.
+    const int batchSize = (windowsCommandLineMax ~/ 100) - 1;
```
ditman, line 537: should this be `nonWindowsCommandLineMax`? Or is this assuming that `windowsCommandLineMax` is always going to be much lower than `nonWindows` one?

script/tool/test/format_command_test.dart
```diff
@@ -348,4 +494,95 @@ void main() {
...
+    );
+
+    await runCapturingPrint(runner, <String>['format']);
+
+    expect(processRunner.recordedCalls.length, 1);
+  });
+
+  test('Batches extremely long file lists on non-Windows', () async {
+    const String pluginName = 'a_plugin';
+    const int commandLineMax = 1000000;
```
ditman, line 556: Unused var `commandLineMax`?

ditman (approved):
Good stuff! A few minor questions/nits, but nothing blocking. 🚀

script/tool/lib/src/format_command.dart
stuartmorgan: I'm not sure for Java; for clang-format we should actually have an exact version per https://github.com/flutter/flutter/issues/39767

I can see about fixing that as a follow-up while I'm thinking of it.

script/tool/lib/src/format_command.dart
stuartmorgan: Good idea, done.

script/tool/test/format_command_test.dart
stuartmorgan: Yes, the tests here were... interesting to write.

script/tool/test/format_command_test.dart
stuartmorgan: I left it this way deliberately, because I think it's easier to understand the intention: figure out the batch size (which requires -1) then overflow it by one (which requires +1). Otherwise it feels like it just sort of works by accident.

It's also easier to see how this would scale. E.g., if we end up needing a test with N batches in the future, having the correct batch size in the code that would be copied from makes that easier.

script/tool/test/format_command_test.dart
stuartmorgan: The latter. Added a comment to the test to make that clear.

script/tool/test/format_command_test.dart
stuartmorgan: Fixed; cruft from before I made the values public for testing.

ditman (approved):
LGTM!

PR #4154, opened 2021-07-13
[flutter_plugin_tools] Improve license-check output

stuartmorgan:
Currently each type of check handles its output in isolation, which
creates confusing output when the last check succeeds but an earlier
check fails, since the end of the output will just be a success message.

This makes the output follow the same basic approach as the package
looper commands, where all failures are collected, and then a final
summary is presented at the end, so the last message will always reflect
the important details.

It also adopts the colorized output now used by most other commands.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/CHANGELOG.md
+++ b/script/tool/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+- Improved `license-check` output.
+
 ## 0.4.0
 
 - Modified the output format of many commands
--- a/script/tool/lib/src/license_check_command.dart
+++ b/script/tool/lib/src/license_check_command.dart
@@ -107,21 +107,65 @@ class LicenseCheckCommand extends PluginCommand {
 
   @override
   Future<void> run() async {
-    final Iterable<File> codeFiles = (await _getAllFiles()).where((File file) =>
+    final Iterable<File> allFiles = await _getAllFiles();
+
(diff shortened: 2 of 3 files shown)


stuartmorgan:
The code diff looks worse than it actually is; it's mostly just moving code.

(This is unrelated to the other recent PRs, it's just to address something I saw cause confusion on another PR.)

ditman (approved):
Very nice!

ditman:
> The code diff looks worse than it actually is; it's mostly just moving code.

Yeah, I noticed this pretty much only handles the return of already existing checks in a slightly different way. Good stuff!

My only comment would be to return a `Map<_ReasonForFailure, List<String>>` instead of `List<File>` when validating (return only the path of the failing file), but since you're already passing in a list with all the files, and that function is only filtering, and no new Files are being created, it doesn't matter!

stuartmorgan:
> My only comment would be to return a `Map<_ReasonForFailure, List<String>>` instead of `List<File>` when validating (return only the path of the failing file), but since you're already passing in a list with all the files, and that function is only filtering, and no new Files are being created, it doesn't matter!

That's actually the signature I wrote when I started reworking the return value, before I realized it was easier in practice to just use `File` given the structure that was already there. 🙂

PR #4157, opened 2021-07-15
[google_sign_in] Add iOS unit tests

jmagman:
- Add google_sign_in iOS unit tests to validate we don't break (much) in preparation for https://github.com/flutter/flutter/issues/86436, which included SDK removals and signature changes without deprecations. Native code plugin coverage increased from 41.4% to 78.4%.
- Add `-[FLTGoogleSignInPlugin initWithSignIn:]` initializer to allow easy injection of `GIDSignIn` into unit tests.  Expose in a new `google_sign_in.Test` explicit clang module.  Include new modulemap and umbrella header to support this.
- Add lightweight generics to the plugin.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_sign_in/google_sign_in/example/ios/Flutter/AppFrameworkInfo.plist
```diff
@@ -25,6 +25,6 @@
     <string>arm64</string>
   </array>
   <key>MinimumOSVersion</key>
-  <string>8.0</string>
+  <string>9.0</string>
```
jmagman, line 28: Automatic from https://github.com/flutter/flutter/pull/85174

packages/google_sign_in/google_sign_in/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -539,7 +539,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
```
jmagman, line 542: Automatic from https://github.com/flutter/flutter/pull/85174

packages/google_sign_in/google_sign_in/example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      codeCoverageEnabled = "YES">
```
jmagman, line 30: Turn on code coverage.

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
```diff
@@ -16,7 +16,7 @@ @interface FLTGoogleSignInPluginTest : XCTestCase
 @property(strong, nonatomic) NSObject<FlutterBinaryMessenger> *mockBinaryMessenger;
 @property(strong, nonatomic) NSObject<FlutterPluginRegistrar> *mockPluginRegistrar;
 @property(strong, nonatomic) FLTGoogleSignInPlugin *plugin;
-@property(strong, nonatomic) GIDSignIn *mockSharedInstance;
+@property(strong, nonatomic) id mockSharedInstance;
```
stuartmorgan, line 19: Can we give this variable a clearer name, especially since it no longer has a type?

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
```diff
@@ -26,62 +26,406 @@ - (void)setUp {
   [super setUp];
   self.mockBinaryMessenger = OCMProtocolMock(@protocol(FlutterBinaryMessenger));
   self.mockPluginRegistrar = OCMProtocolMock(@protocol(FlutterPluginRegistrar));
-  self.mockSharedInstance = [OCMockObject partialMockForObject:[GIDSignIn sharedInstance]];
+
+  id mockSharedInstance = OCMClassMock([GIDSignIn class]);
+  OCMStub([mockSharedInstance sharedInstance]).andReturn(mockSharedInstance);
```
stuartmorgan, line 31: Any chance we could adjust the code to accept an injectable GIDSignIn instance instead of using class mocking?

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
```diff
@@ -26,62 +26,406 @@ - (void)setUp {
...
+  OCMVerify([self.mockSharedInstance signOut]);
+}
+
+- (void)testDisconnect {
+  FlutterMethodCall *methodCall = [FlutterMethodCall methodCallWithMethodName:@"disconnect"
+                                                                    arguments:nil];
+
+  [self.plugin handleMethodCall:methodCall
+                         result:^(id result){
+                         }];
```
stuartmorgan, line 77: Don't we want to assert that it returned a (non-error) result?

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
```diff
@@ -26,62 +26,406 @@ - (void)setUp {
...
+
+- (void)testSignInSilently {
+  OCMExpect([self.mockSharedInstance restorePreviousSignIn]);
+
+  FlutterMethodCall *methodCall = [FlutterMethodCall methodCallWithMethodName:@"signInSilently"
+                                                                    arguments:nil];
+
+  [self.plugin handleMethodCall:methodCall
+                         result:^(id result){
+                         }];
```
stuartmorgan, line 203: Same question here. (And all following cases that don't assert anything about the return value)

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
```diff
@@ -26,62 +26,406 @@ - (void)setUp {
...
+                           XCTAssertEqualObjects(result.code, @"sign_in_canceled");
+                           XCTAssertEqualObjects(result.message, kGIDSignInErrorDomain);
+                           [expectation fulfill];
+                         }];
+  [self waitForExpectationsWithTimeout:5.0 handler:nil];
+}
+
+- (void)testGetTokensURLError {
+  id mockUser = OCMClassMock([GIDGoogleUser class]);
+  OCMStub([self.mockSharedInstance currentUser]).andReturn(mockUser);
```
stuartmorgan, line 340: Don't these need stopMocking calls?

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin.m
```diff
@@ -76,7 +76,8 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
       NSString *path = [[NSBundle mainBundle] pathForResource:@"GoogleService-Info"
                                                        ofType:@"plist"];
       if (path) {
-        NSMutableDictionary *plist = [[NSMutableDictionary alloc] initWithContentsOfFile:path];
+        NSMutableDictionary<NSString *, NSString *> *plist =
+            [[NSMutableDictionary alloc] initWithContentsOfFile:path];
         BOOL hasDynamicClientId =
             [[call.arguments valueForKey:@"clientId"] isKindOfClass:[NSString class]];
```
stuartmorgan, line 82: Maybe use that new-fangled `[]` access here while you are modernizing :)

(We could consider doing a sweep of the repo for old-style dictionary and array reads and writes and updating them, like I did for literals.)

stuartmorgan (commented):
Nice :)

The description says "Turn on code coverage", but I'm not seeing project changes other than deployment target.

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
jmagman: The tests that don't check the result are doing do because the result callback happens from a `GIDSignIn.delegate` method.  
https://github.com/flutter/plugins/blob/b548ae18646d1725a8e1f004c26bf42e4d362aa8/packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin.m#L132-L135
sends the response in the `GIDSignInDelegate` call.
https://github.com/flutter/plugins/blob/b548ae18646d1725a8e1f004c26bf42e4d362aa8/packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin.m#L246-L250
which means I'd have to mock out the delegate calls.  
GoogleSignIn 6.0 (https://github.com/flutter/flutter/issues/86436) removed this delegate and made it a `disconnectWithCallback:` so the above code will become:
```obj-c
    [GIDSignIn.sharedInstance disconnectWithCallback:^(NSError * _Nullable error) {
      [self respondWithAccount:@{} result:result error:nil];
    }];
```
which will be super easy to test.

TL;DR I started to write tests for these results, but it would have been a lot of throwaway work for the next review.  Instead, I wrote all the tests that seemed like the best bang for my buck, and the next review tests diff will be easier to read, and not a wall o'green like this review.

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
jmagman: Same comment as above, in GoogleSignIn 6.0 this became a callback `restorePreviousSignInWithCallback:` instead of a delegate.
https://github.com/flutter/plugins/blob/b548ae18646d1725a8e1f004c26bf42e4d362aa8/packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin.m#L103-L106
is handled in the delegate method:
https://github.com/flutter/plugins/blob/b548ae18646d1725a8e1f004c26bf42e4d362aa8/packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin.m#L208-L214
will become the easily testable:
```obj-c
    [GIDSignIn.sharedInstance restorePreviousSignInWithCallback:^(GIDGoogleUser *user, NSError *error) {
      [self didSignInForUser:user result:result withError:error];
    }];
```

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
jmagman: The teardown calls `[self.mockSharedInstance stopMocking]`.
Only `OCMClassMock` objects with stubbed _class_ methods need `-stopMocking` called on them.  Stubbing on instance method calls `-stopMocking` when the instance is deallocated.  `mockSharedInstance` currently needs it because `[mockSharedInstance sharedInstance]` is mocked (though it won't be needed once I finish your injection suggestion).

jmagman:
> The description says "Turn on code coverage", but I'm not seeing project changes other than deployment target.

Ah sorry I removed that in an intermediate commit since it's not being checked anywhere, so not worth the overhead in CI.  Fixed the description.

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
jmagman: Done.

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
jmagman: Done.

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin.m
jmagman: Done.

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin_Private.h
```diff
@@ -0,0 +1,13 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
jmagman, line 1: Unfortunately I had to make this a public header to make it accessible to the RunnerTest.

jmagman: Maybe I should rename to `FLTGoogleSignInPlugin_Test.h`

jmagman: Just renamed it.

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
stuartmorgan: Yep, I just can't read; I saw `OCMStub` and `sharedInstance` in the same line and thought this was stubbing another class method.

packages/google_sign_in/google_sign_in/example/ios/RunnerTests/GoogleSignInTests.m
stuartmorgan: Makes sense, thanks!

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin_Private.h
stuartmorgan: It would be kind of ugly, but couldn't the test do a path-based include of the private header instead of getting it via the module import?

stuartmorgan: Alternately, you could do the thing that I almost always tell people *not* to do (in favor of making an internal header like you did) and re-declare the method you need at the top of the test file. I've never liked it, but it seems better than actually making the private header public in the module.

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin_Test.h
```diff
@@ -0,0 +1,13 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <google_sign_in/FLTGoogleSignInPlugin.h>
+
+@class GIDSignIn;
+
+@interface FLTGoogleSignInPlugin ()
```
stuartmorgan, line 9: I usually put a comment like "Methods exposed for testing" on these extensions, in case people who aren't familiar with that pattern are trying to understand what the header is for.

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin_Test.h
```diff
@@ -0,0 +1,13 @@
...
+
+- (instancetype)initWithSignIn:(GIDSignIn *)signIn NS_DESIGNATED_INITIALIZER;
```
stuartmorgan, line 11: Maybe comment that this is explicitly for injecting the GIDSignIn for tests?

stuartmorgan (approved):
LGTM with the private header thing adrressed

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin_Private.h
jmagman: How about a `Test` module so it needs to be explicitly imported?
```obj-c
@import google_sign_in;
@import google_sign_in.Test;
```

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin_Test.h
jmagman: Done.

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin_Test.h
jmagman: Done.

packages/google_sign_in/google_sign_in/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Flutter plugin for Google Sign-In, a secure authentication system
   for signing in with a Google account on Android and iOS.
 repository: https://github.com/flutter/plugins/tree/master/packages/google_sign_in/google_sign_in
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+google_sign_in%22
-version: 5.0.4
+version: 5.0.5
```
jmagman, line 6: @stuartmorgan with the module change this seems different enough to warrant a patch bump.

packages/google_sign_in/google_sign_in/ios/google_sign_in.podspec
```diff
@@ -12,8 +12,9 @@ Enables Google Sign-In in Flutter apps.
   s.license          = { :type => 'BSD', :file => '../LICENSE' }
   s.author           = { 'Flutter Team' => 'flutter-dev@googlegroups.com' }
   s.source           = { :http => 'https://github.com/flutter/plugins/tree/master/packages/google_sign_in' }
-  s.source_files = 'Classes/**/*'
+  s.source_files = 'Classes/**/*.{h,m}'
```
jmagman, line 15: `modulemap` shouldn't be included in the source files.

packages/google_sign_in/google_sign_in/ios/Classes/google_sign_in-umbrella.h
```diff
@@ -0,0 +1,10 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
jmagman, line 1: The file was previously generated by CocoaPods with the same name.  Now we need to manage it manually and check it into the repo because we're defining a module map, which takes us out of the magical auto module world.

packages/google_sign_in/google_sign_in/ios/Classes/FLTGoogleSignInPlugin.modulemap
```diff
@@ -0,0 +1,10 @@
+framework module google_sign_in {
+  umbrella header "google_sign_in-umbrella.h"
+
+  export *
+  module * { export * }
+
+  explicit module Test {
```
jmagman, line 7: This isn't a private module since there are some complications for Swift files not being able to import those.

Instead, this explicit module requires that `google_sign_in.Test` be imported to access FLTGoogleSignInPlugin_Test.  So `initWithSignIn:` won't be available in the normal `import google_sign_in` case, won't auto-complete, etc.

stuartmorgan (approved):
LGTM

PR #4161, opened 2021-07-16
[flutter_plugin_tests] Split analyze out of xctest

stuartmorgan:
To prep for making a combined command to run native tests across different platforms, rework `xctest`:
- Split analyze out into a new `xcode-analyze` command:
  - Since the analyze step runs a new build over everything with different flags, this is only a small amount slower than the combined version
  - This makes the logic easier to follow
  - This allows us to meaningfully report skips, to better notice missing tests.
- Add the ability to target specific test bundles (RunnerTests or RunnerUITests)

To share code between the commands, this extracts a new `Xcode` helper class.

Part of https://github.com/flutter/flutter/issues/84392 and https://github.com/flutter/flutter/issues/86489

## Pre-launch Checklist

- [x] I signed the [CLA].


script/tool/test/xctest_command_test.dart
```diff
@@ -16,22 +16,8 @@ import 'package:test/test.dart';
...
-      'bundlePath':
-          '/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 13.0.simruntime',
-      'buildversion': '17A577',
-      'runtimeRoot':
-          '/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 13.0.simruntime/Contents/Resources/RuntimeRoot',
-      'identifier': 'com.apple.CoreSimulator.SimRuntime.iOS-13-0',
-      'version': '13.0',
-      'isAvailable': true,
-      'name': 'iOS 13.0'
-    },
```
stuartmorgan, line 34: This testing of finding the best device among many is now part of `xcode_test.dart`, so what's left here is just one valid device for the test to find.

script/tool/test/xctest_command_test.dart
```diff
@@ -123,13 +88,93 @@ void main() {
...
+                  'GCC_TREAT_WARNINGS_AS_ERRORS=YES',
+                ],
+                pluginExampleDirectory.path),
+          ]));
+    });
+
     group('iOS', () {
       test('skip if iOS is not supported', () async {
-        createFakePlugin('plugin', packagesDir, extraFiles: <String>[
-          'example/test',
```
stuartmorgan, line 129: Opportunistic cleanup; this `extraFiles` was copy/pasted from some initial test into lots of other tests that don't actually need it, and I've been cleaning it up as I notice.

script/tool/test/xctest_command_test.dart
```diff
@@ -157,19 +201,14 @@ void main() {
...
+            'plugin', packagesDir, platformSupport: <String, PlatformSupport>{
           kPlatformIos: PlatformSupport.inline
         });
 
         final Directory pluginExampleDirectory =
             pluginDirectory.childDirectory('example');
 
-        processRunner.processToReturn = MockProcess.succeeding();
-        processRunner.resultStdout =
-            '{"project":{"targets":["bar_scheme", "foo_scheme"]}}';
```
stuartmorgan, line 172: Opportunistic cleanup; I'm assuming that this is from an earlier version of `xctest_command`, because nothing currently uses this output.

script/tool/test/xctest_command_test.dart
```diff
@@ -314,14 +331,13 @@ void main() {
...
-        ], platformSupport: <String, PlatformSupport>{
-          kPlatformMacos: PlatformSupport.federated,
-        });
+        createFakePlugin('plugin', packagesDir,
+            platformSupport: <String, PlatformSupport>{
+              kPlatformMacos: PlatformSupport.federated,
+            });
 
-        final List<String> output = await runCapturingPrint(runner,
-            <String>['xctest', '--macos', _kDestination, 'foo_destination']);
```
stuartmorgan, line 324: Opportunistic cleanup; I accidentally left the targets when initially copy/pasting iOS tests for macOS.

script/tool/lib/src/xctest_command.dart
```diff
@@ -56,9 +59,6 @@ class XCTestCommand extends PackageLoopingCommand {
       'Runs the xctests in the iOS and/or macOS example apps.\n\n'
       'This command requires "flutter" and "xcrun" to be in your path.';
 
-  @override
-  String get failureListHeader => 'The following packages are failing XCTests:';
```
stuartmorgan, line 60: Opportunistic cleanup; I left this when doing the initial migration to the new base command to minimize test diff, but there's no reason we need a custom summary header for this command.

script/tool/lib/src/common/xcode.dart
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  /// Returns the newest available simulator (highest OS version, with ties
+  /// broken in favor of newest device), if any.
+  Future<String?> findBestAvailableIphoneSimulator() async {
+    final List<String> findSimulatorsArguments = <String>[
+      'simctl',
+      'list',
+      '--json'
+    ];
```
jmagman, line 95: I know you just copied this, but if you add `devices available`
```suggestion
    final List<String> findSimulatorsArguments = <String>[
      'simctl',
      'list',
      'devices',
      'available',
      '--json'
    ];
```
The `availabilityError`/`isAvailable` check below can be removed.

script/tool/lib/src/xcode_analyze_command.dart
```diff
@@ -0,0 +1,111 @@
...
+
+    final List<String> failures = <String>[];
+    if (testIos &&
+        !await _analyzePlugin(package, 'iOS', extraFlags: <String>[
+          '-destination',
+          'generic/platform=iOS Simulator'
+        ])) {
+      failures.add('iOS');
+    }
+    if (testMacos && !await _analyzePlugin(package, 'macOS')) {
```
jmagman, line 67: You don't need to worry about this now, but when CI starts using Xcode 13 next year we _may_ need a macOS `-destination` option.
https://github.com/flutter/flutter/issues/86590

jmagman:
> Add the ability to target specific test bundles (RunnerTests or RunnerUITests)

Both types of tests will run on the same build artifacts, so it should be faster to run them at the same time.  It's a fine feature to add, just keep it in mind if you intend to split them into separate CI tests.

jmagman (approved):

stuartmorgan:
> Both types of tests will run on the same build artifacts, so it should be faster to run them at the same time. It's a fine feature to add, just keep it in mind if you intend to split them into separate CI tests.

Yes, I had tested that at the same time I tested splitting out `analyze` and as expected it is much slower. This is intended for local use (and is prep for the new combined PR that will have options to select unit or integration tests for local use, and maybe for use in CI on platforms where combining them doesn't help).

script/tool/lib/src/common/xcode.dart
stuartmorgan: Thanks, done! (I had to add 'runtimes' as well since it's parsing those too.)

jmagman (approved):

PR #4169, opened 2021-07-20
Add `postUrl` method

ydag:
This PR adds `postUrl` implementation to use **POST** request and accept **POST** data.

Related issues :

- Fixes [flutter/flutter#27730](https://github.com/flutter/flutter/issues/27730)

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/android/build.gradle
+++ b/packages/webview_flutter/android/build.gradle
@@ -37,5 +37,6 @@ android {
         implementation 'androidx.annotation:annotation:1.0.0'
         implementation 'androidx.webkit:webkit:1.0.0'
         testImplementation 'junit:junit:4.12'
+        testImplementation 'org.mockito:mockito-core:3.10.0'
     }
 }
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -17,6 +17,7 @@ import android.webkit.WebStorage;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
 import io.flutter.plugin.common.BinaryMessenger;
 import io.flutter.plugin.common.MethodCall;
 import io.flutter.plugin.common.MethodChannel;
@@ -77,6 +78,14 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
     public void onProgressChanged(WebView view, int progress) {
       flutterWebViewClient.onLoadingProgress(progress);
     }
+  }
+
+  @VisibleForTesting
+  FlutterWebView(WebView webView, MethodChannel methodChannel, Handler platformThreadHandler) {
+    this.webView = webView;
+    this.methodChannel = methodChannel;
+    flutterWebViewClient = new FlutterWebViewClient(methodChannel);
+    this.platformThreadHandler = platformThreadHandler;
   }
 
   @TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR1)
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -261,6 +261,14 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
     result.success(null);
   }
 
+  public void postUrl(MethodCall methodCall, Result result) {
+    Map<String, Object> request = (Map<String, Object>) methodCall.arguments;
+    String url = (String) request.get("url");
+    byte[] postData = (byte[]) request.get("postData");
+    webView.postUrl(url, postData);
+    result.success(null);
+  }
+
   private void canGoBack(Result result) {
     result.success(webView.canGoBack());
   }
--- a/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -193,6 +193,9 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
       case "loadUrl":
         loadUrl(methodCall, result);
         break;
+      case "postUrl":
+        postUrl(methodCall, result);
+        break;
       case "updateSettings":
         updateSettings(methodCall, result);
         break;
--- a/packages/webview_flutter/android/src/test/java/io/flutter/plugins/webviewflutter/WebViewTest.java
+++ b/packages/webview_flutter/android/src/test/java/io/flutter/plugins/webviewflutter/WebViewTest.java
@@ -5,11 +5,83 @@
 package io.flutter.plugins.webviewflutter;
 
 import static org.junit.Assert.assertEquals;
+import static org.mockito.ArgumentMatchers.isA;
+import static org.mockito.Mockito.doNothing;
 
+import android.os.Handler;
+import android.webkit.WebView;
 import android.webkit.WebViewClient;
+import io.flutter.plugin.common.MethodCall;
+import io.flutter.plugin.common.MethodChannel;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Before;
 import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.mockito.internal.matchers.Null;
 
 public class WebViewTest {
+  private static final String POST_URL = "postUrl";
+  private static final String URL = "www.example.com";
+  byte[] postData;
+
+  @Mock WebView mockWebView;
+  @Mock MethodChannel mockChannel;
+  @Mock Handler mockHandler;
+  @Mock MethodChannel.Result mockResult;
+
+  FlutterWebView flutterWebView;
+
+  @Before
+  public void setUp() {
+    MockitoAnnotations.openMocks(this);
+    postData = new byte[5];
+  }
+
+  @Test
+  public void testPostUrl_should_call_webView_postUrl_with_correct_url() {
+    MethodCall call = buildMethodCall(POST_URL, URL, postData);
+    flutterWebView = new FlutterWebView(mockWebView, mockChannel, mockHandler);
+
+    ArgumentCaptor<String> valueCapture = ArgumentCaptor.forClass(String.class);
+
+    doNothing().when(mockWebView).postUrl(valueCapture.capture(), isA(byte[].class));
+
+    flutterWebView.postUrl(call, mockResult);
+
+    assertEquals(URL, valueCapture.getValue());
+  }
+
+  @Test
+  public void testPostUrl_should_call_webView_postUrl_with_correct_http_body() {
+    MethodCall call = buildMethodCall(POST_URL, URL, postData);
+    flutterWebView = new FlutterWebView(mockWebView, mockChannel, mockHandler);
+
+    ArgumentCaptor<byte[]> valueCapture = ArgumentCaptor.forClass(byte[].class);
+
+    doNothing().when(mockWebView).postUrl(isA(String.class), valueCapture.capture());
+
+    flutterWebView.postUrl(call, mockResult);
+
+    assertEquals(postData, valueCapture.getValue());
+  }
+
+  @Test
+  public void testPostUrl_should_call_result_success_with_null() {
+    MethodCall call = buildMethodCall(POST_URL, URL, postData);
+    flutterWebView = new FlutterWebView(mockWebView, mockChannel, mockHandler);
+
+    ArgumentCaptor<Null> valueCapture = ArgumentCaptor.forClass(Null.class);
+
+    doNothing().when(mockResult).success(valueCapture.capture());
+
+    flutterWebView.postUrl(call, mockResult);
+
+    assertEquals(null, valueCapture.getValue());
+  }
+
   @Test
   public void errorCodes() {
     assertEquals(
@@ -45,5 +117,13 @@ public class WebViewTest {
     assertEquals(
         FlutterWebViewClient.errorCodeToString(WebViewClient.ERROR_UNSUPPORTED_SCHEME),
         "unsupportedScheme");
+  }
+
+  private MethodCall buildMethodCall(String method, final String url, final byte[] postData) {
+    final Map<String, Object> arguments = new HashMap<>();
+    arguments.put("url", url);
+    arguments.put("postData", postData);
+
+    return new MethodCall(method, arguments);
   }
 }
--- a/packages/webview_flutter/example/ios/RunnerTests/FLTWebViewTests.m
+++ b/packages/webview_flutter/example/ios/RunnerTests/FLTWebViewTests.m
@@ -11,6 +11,63 @@
 
 static bool feq(CGFloat a, CGFloat b) { return fabs(b - a) < FLT_EPSILON; }
 
+@interface FLTWebViewController (Test)
+- (bool)postUrl:(NSString *)url withBody:(FlutterStandardTypedData *)postData;
+- (void)onPostUrl:(FlutterMethodCall *)call result:(FlutterResult)result;
+- (bool)postRequest:(NSDictionary<NSString *, id> *)request;
+@end
+
+@interface MockFLTWebViewControllerForOnPostUrl : FLTWebViewController
+- (instancetype)initWithPostRequest:(BOOL)postRequestResult;
+@end
+
+@implementation MockFLTWebViewControllerForOnPostUrl {
+  bool _postRequestResult;
+}
+
+- (instancetype)initWithPostRequest:(bool)postRequestResult {
+  _postRequestResult = postRequestResult;
+  return self;
+}
+
+- (bool)postRequest:(NSDictionary<NSString *, id> *)request {
+  return _postRequestResult;
+}
+
+@end
+
+@interface MockFLTWebViewControllerForPostRequest : FLTWebViewController
+- (instancetype)initWithPostUrl:(BOOL)postUrlResult;
+@end
+
+@implementation MockFLTWebViewControllerForPostRequest {
+  bool _postUrlResult;
+}
+
+- (instancetype)initWithPostUrl:(bool)postUrlResult {
+  _postUrlResult = postUrlResult;
+  return self;
+}
+
+- (bool)postUrl:(NSString *)url withBody:(FlutterStandardTypedData *)postData {
+  return _postUrlResult;
+}
+
+@end
+
+@interface MockWKWebViewForPostUrl : FLTWKWebView
+@property(nonatomic, nullable) NSMutableURLRequest *receivedResult;
+@end
+
+@implementation MockWKWebViewForPostUrl
+
+- (WKNavigation *)loadRequest:(NSMutableURLRequest *)request {
+  _receivedResult = request;
+  return nil;
+}
+
+@end
+
 @interface FLTWebViewTests : XCTestCase
 
 @property(strong, nonatomic) NSObject<FlutterBinaryMessenger> *mockBinaryMessenger;
--- a/packages/webview_flutter/example/ios/RunnerTests/FLTWebViewTests.m
+++ b/packages/webview_flutter/example/ios/RunnerTests/FLTWebViewTests.m
@@ -72,6 +72,8 @@ static bool feq(CGFloat a, CGFloat b) { return fabs(b - a) < FLT_EPSILON; }
 
 @property(strong, nonatomic) NSObject<FlutterBinaryMessenger> *mockBinaryMessenger;
 
+@property(readonly, nonatomic) MockWKWebViewForPostUrl *resultObject;
+
 @end
 
 @implementation FLTWebViewTests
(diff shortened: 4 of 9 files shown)


packages/webview_flutter/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -79,6 +80,14 @@ public void onProgressChanged(WebView view, int progress) {
...
   }
 
+  @VisibleForTesting
+  FlutterWebView(WebView webView, MethodChannel methodChannel, Handler platformThreadHandler) {
+    this.webView = webView;
+    this.methodChannel = methodChannel;
+    flutterWebViewClient = new FlutterWebViewClient(methodChannel);
+    this.platformThreadHandler = platformThreadHandler;
+  }
+
```
mvanbeusekom, line 90: Creating a separate constructor just for the purpose of testing is considered a bad practise (see also [this answer](https://softwareengineering.stackexchange.com/a/385805/336392)).

I have just merged my [PR](https://github.com/flutter/plugins/pull/4178) which refactors the existing constructor which should help you mock the `methodChannel` and `webView` instances more easily. If you merge with the master branch you should be able to remove this constructor and fix the unit-test.

packages/webview_flutter/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -258,6 +270,14 @@ private void loadUrl(MethodCall methodCall, Result result) {
     result.success(null);
   }
 
+  public void postUrl(MethodCall methodCall, Result result) {
```
mvanbeusekom, line 273: I think this method should be `private`, if there is a reason to keep it `public` it should be documented.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.h
```diff
@@ -7,26 +7,28 @@
...
+
 @interface FLTWebViewController : NSObject <FlutterPlatformView, WKUIDelegate>
 
 - (instancetype)initWithFrame:(CGRect)frame
                viewIdentifier:(int64_t)viewId
                     arguments:(id _Nullable)args
               binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger;
 
+- (instancetype)initWithWebView:(FLTWKWebView*)webView;
+
```
mvanbeusekom, line 26: Same as with Android, creating a constructor just to facilitate unit-testing is a bad practise. On the iOS side I haven't made any refactor changes so have a look if you can solve it in a similar way.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -192,6 +200,17 @@ - (void)onLoadUrl:(FlutterMethodCall*)call result:(FlutterResult)result {
   }
 }
 
+- (void)onPostUrl:(FlutterMethodCall*)call result:(FlutterResult)result {
```
mvanbeusekom, line 203: Missing documentation. All methods need comments, per style guide.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -450,6 +469,33 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
   return true;
 }
 
+- (bool)postRequest:(NSDictionary<NSString*, id>*)request {
```
mvanbeusekom, line 472: Missing documentation.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -450,6 +469,33 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
...
+  if ([url isKindOfClass:[NSString class]]) {
+    id postData = request[@"postData"];
+    if ([postData isKindOfClass:[FlutterStandardTypedData class]]) {
+      return [self postUrl:url withBody:postData];
+    }
+  }
+  return false;
+}
+
+- (bool)postUrl:(NSString*)url withBody:(FlutterStandardTypedData*)postData {
```
mvanbeusekom, line 487: Missing documentation.

mvanbeusekom (changes requested):

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -450,6 +469,33 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
...
 
+- (bool)postRequest:(NSDictionary<NSString*, id>*)request {
+  if (!request) {
+    return false;
+  }
+
+  NSString* url = request[@"url"];
+  if ([url isKindOfClass:[NSString class]]) {
+    id postData = request[@"postData"];
+    if ([postData isKindOfClass:[FlutterStandardTypedData class]]) {
```
stuartmorgan, line 480: Please just use early returns for error handling instead of chaining methods for the sole purpose of reducing nesting.

packages/webview_flutter/webview_flutter/lib/platform_interface.dart
```diff
@@ -187,6 +188,22 @@ abstract class WebViewPlatformController {
...
+  ///
+  /// If `url` is not a network URL, it will be loaded with `loadUrl` instead,
+  /// ignoring the `postData` param on Android.
+  ///
+  /// `url` and `postData` must not be null.
+  ///
+  /// Throws an ArgumentError if `url` is not a valid URL string.
+  Future<void> postUrl(
+    String url,
+    Uint8List postData,
```
stuartmorgan, line 201: Someone is inevitably going to want to pass headers here.

More generally, anything we might add to one of these methods is likely to apply to the other, which makes me thing we should make a new API that operates in terms of a new simple request class, rather than having a bunch of parameters. That will also mean we could add new things to the requests later without it being a breaking change for implementors, since the request object wouldn't be subclassed by each implementation (unlike method arguments, which are breaking to add).

mvanbeusekom: The main reason why we haven't implemented support for headers here is that this is not supported on Android. There is a possible workaround where we do a normal HTTP Post request and load the resulting HTML in the WebView control but this will have the side effect that we need to await for the whole request to be finished before we can load the HTML in the WebView control. Whereas on iOS the WKWebView control can start loading the HTML content directly as it arrives on the response stream. So since there is a difference in behaviour between platforms we choose to stick to the most common implementation (which is indeed a bit limited).

I just wanted to provide this bit of context so you have the full picture on why we choose this direction. I like the idea of adding a new API which operates based on a simple request class. However that also doesn't really solve the issue when dealing with headers on Android. It might even make things worse as it will be more difficult to explain certain fields don't work on certain platforms. 

So maybe good to have a small discussion on this before we pick a certain direction to go on?

stuartmorgan: It's fine to explain that certain features don't work on specific platforms; we do that all the time in plugins. We shouldn't paint ourselves into an API corner here just because we can't currently support a feature on one platform.

packages/webview_flutter/webview_flutter/lib/platform_interface.dart
```diff
@@ -173,15 +174,25 @@ abstract class WebViewPlatformController {
...
+  /// for the request.
+  ///
+  /// - If [WebViewRequest.body] is not null, it will be added as HTTP body
+  /// for the request.
   ///
   /// `url` must not be null.
   ///
   /// Throws an ArgumentError if `url` is not a valid URL string.
   Future<void> loadUrl(
     String url,
```
stuartmorgan, line 194: Why is the URL separate from the request?

packages/webview_flutter/webview_flutter/lib/platform_interface.dart
```diff
@@ -173,15 +174,25 @@ abstract class WebViewPlatformController {
...
-    Map<String, String>? headers,
+    WebViewRequest? request,
```
stuartmorgan, line 195: I don't think this should be nullable; a request must have a method. (If URL gets folded in, it will definitely not be nullable).

packages/webview_flutter/webview_flutter/lib/platform_interface.dart
stuartmorgan, line 194: This seems like a good opportunity to switch from `String` to `Uri`; a lesson of `url_launcher` has been that people pass a lot of invalid URLs to things that accept strings, and this causes issue (e.g., Android working but iOS not working because iOS is stricter about creating URLs from invalid strings.)

packages/webview_flutter/webview_flutter/lib/platform_interface.dart
```diff
@@ -173,15 +174,25 @@ abstract class WebViewPlatformController {
...
+  /// [WebViewRequest.headers] will be added as key value pairs of HTTP headers
+  /// for the request.
+  ///
+  /// - If [WebViewRequest.body] is not null, it will be added as HTTP body
+  /// for the request.
   ///
   /// `url` must not be null.
   ///
   /// Throws an ArgumentError if `url` is not a valid URL string.
   Future<void> loadUrl(
```
stuartmorgan, line 193: I think rather than breaking clients, we should add a `loadRequest` and deprecate `loadUrl`.

packages/webview_flutter/webview_flutter/lib/src/types/webview_request.dart
```diff
@@ -0,0 +1,53 @@
...
+    required this.method,
+    this.headers,
+    this.body,
+  });
+
+  /// HTTP method used to load the page.
+  final WebViewLoadMethod method;
+
+  /// HTTP headers for the request.
+  final Map<String, String>? headers;
```
stuartmorgan, line 42: Generally we prefer default-empty collections rather than nullable collections, so there aren't two ways to express the same thing ("no headers" in this case).

packages/webview_flutter/webview_flutter/lib/src/webview_method_channel.dart
```diff
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:flutter/services.dart';
+import '../src/types/webview_request.dart';
```
stuartmorgan, line 8: This should be grouped with the below, not the `package:` entry.

packages/webview_flutter/webview_flutter/lib/webview_flutter.dart
```diff
@@ -634,19 +635,29 @@ class WebViewController {
...
+  /// - If [WebViewRequest.body] is not null, it will be added as HTTP body
+  /// for the request.
   ///
   /// `url` must not be null.
   ///
   /// Throws an ArgumentError if `url` is not a valid URL string.
   Future<void> loadUrl(
     String url, {
-    Map<String, String>? headers,
+    WebViewRequest? request,
```
stuartmorgan, line 656: All the same comments apply here (I meant to make them on the public interface not the platform interface, but they are applicable in both places.)

stuartmorgan (changes requested):
First pass on the Dart changes; I haven't looked at the implementation yet since getting the API nailed down is the most important part.

BeMacized:
@googlebot I consent.

ydag:
@googlebot I consent.

packages/webview_flutter/webview_flutter/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -294,6 +312,65 @@ private void loadUrl(MethodCall methodCall, Result result) {
...
+                  @Override
+                  public void onComplete(String content) {
+                    if (!webView.isAttachedToWindow()) {
+                      result.error(
+                          "webview_destroyed",
+                          "Could not complete the post request because the webview is destroyed",
+                          null);
+                    } else {
+                      webView.loadDataWithBaseURL(
+                          webViewRequest.getUrl(), content, "text/html", "UTF-8", null);
```
stuartmorgan, line 341: Are we certain that the request URL is always correct? E.g., what happens here in the case of a server-side redirect to a different origin?

packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
```diff
@@ -90,6 +91,7 @@ void main() {
     final Map<String, String> headers = <String, String>{
       'test_header': 'flutter_test_header'
     };
+    // ignore: deprecated_member_use
```
stuartmorgan, line 94: Why are there no new integration tests using the new API?

packages/webview_flutter/webview_flutter/example/ios/RunnerTests/FLTWebViewTests.m
```diff
@@ -11,19 +11,215 @@
...
+}
+
+- (FLTWKWebView *)createFLTWKWebViewWithFrame:(CGRect)frame
+                                configuration:(WKWebViewConfiguration *)configuration
+                           navigationDelegate:(FLTWKNavigationDelegate *)navigationDelegate {
+  mockFLTWKWebView = [MockFLTWKWebView new];
+  return mockFLTWKWebView;
+}
+
+- (MockFLTWKWebView *)getResultObject {
```
stuartmorgan, line 50: ObjC methods should not start with `get`.

packages/webview_flutter/webview_flutter/example/ios/RunnerTests/FLTWebViewTests.m
```diff
@@ -11,19 +11,215 @@
...
+@end
+
+@implementation MockFLTWebViewController {
+  MockFLTWKWebView *mockFLTWKWebView;
+}
+
+- (FLTWKWebView *)createFLTWKWebViewWithFrame:(CGRect)frame
+                                configuration:(WKWebViewConfiguration *)configuration
+                           navigationDelegate:(FLTWKNavigationDelegate *)navigationDelegate {
+  mockFLTWKWebView = [MockFLTWKWebView new];
```
stuartmorgan, line 46: https://google.github.io/styleguide/objcguide.html#do-not-use-new

packages/webview_flutter/webview_flutter/example/ios/RunnerTests/FLTWebViewTests.m
stuartmorgan: It's also not clear to me why this is a "result object" when it's created by the mock itself. Why not just call this `mockWebView` (and make it a property)?

packages/webview_flutter/webview_flutter/example/ios/RunnerTests/FLTWebViewTests.m
```diff
@@ -11,19 +11,215 @@
...
+@end
+
 @implementation FLTWebViewTests
 
 - (void)setUp {
   [super setUp];
   self.mockBinaryMessenger = OCMProtocolMock(@protocol(FlutterBinaryMessenger));
 }
 
+- (void)testbuildNSURLRequest_should_return_nil_when_arguments_is_nil {
```
stuartmorgan, line 86: Please use standard ObjC naming, which does not use snake case.

packages/webview_flutter/webview_flutter/example/lib/main.dart
```diff
@@ -270,6 +270,7 @@ class SampleMenu extends StatelessWidget {
       WebViewController controller, BuildContext context) async {
     final String contentBase64 =
         base64Encode(const Utf8Encoder().convert(kNavigationExamplePage));
+    // ignore: deprecated_member_use
     await controller.loadUrl('data:text/html;base64,$contentBase64');
```
stuartmorgan, line 274: We should use current APIs, not deprecated APIs, in the example.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -94,34 +94,46 @@ - (instancetype)initWithFrame:(CGRect)frame
     [self updateAutoMediaPlaybackPolicy:args[@"autoMediaPlaybackPolicy"]
                         inConfiguration:configuration];
 
-    _webView = [[FLTWKWebView alloc] initWithFrame:frame configuration:configuration];
     _navigationDelegate = [[FLTWKNavigationDelegate alloc] initWithChannel:_channel];
-    _webView.UIDelegate = self;
-    _webView.navigationDelegate = _navigationDelegate;
+    _webView = [self createFLTWKWebViewWithFrame:frame
```
stuartmorgan, line 98: Do not call methods in init: https://google.github.io/styleguide/objcguide.html#avoid-messaging-the-current-object-within-initializers-and--dealloc

I see that there are already violations of this; we should not add more (and should at some point fix those).

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -192,6 +206,29 @@ - (void)onLoadUrl:(FlutterMethodCall*)call result:(FlutterResult)result {
   }
 }
```
stuartmorgan, line 207: The legacy URL-based method should, as much as possible, be refactored to use the same flow as the new request-based method, so we don't have independent codepaths.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -450,6 +487,53 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
...
+ * Parses the arguments and converts them into an NSURLRequest object.
+ *
+ * @param arguments the method call arguments.
+ *
+ * @return NSURLRequest object.
+ */
+- (NSURLRequest*)buildNSURLRequest:(NSDictionary<NSString*, id>*)arguments {
+  if (!arguments) {
+    return nil;
+  }
```
stuartmorgan, line 500: This check isn't necessary.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -450,6 +487,53 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
...
+
+  id requestParameters = arguments[@"request"];
+  if (!(requestParameters && [requestParameters isKindOfClass:[NSDictionary class]])) {
+    return nil;
+  }
+
+  NSString* url = requestParameters[@"url"];
```
stuartmorgan, line 507: `urlString`

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -450,6 +487,53 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
...
+  if (!url) {
+    return nil;
+  }
+
+  NSURL* nsUrl = [NSURL URLWithString:url];
```
stuartmorgan, line 512: `url`

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -450,6 +487,53 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
...
+ *
+ * @return NSURLRequest object.
+ */
+- (NSURLRequest*)buildNSURLRequest:(NSDictionary<NSString*, id>*)arguments {
+  if (!arguments) {
+    return nil;
+  }
+
+  id requestParameters = arguments[@"request"];
+  if (!(requestParameters && [requestParameters isKindOfClass:[NSDictionary class]])) {
```
stuartmorgan, line 503: The first check is unnecessary; sending a message to nil returns nil/NO/0/etc.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -450,6 +487,53 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
...
+
+  NSMutableURLRequest* request = [NSMutableURLRequest requestWithURL:nsUrl];
+
+  NSString* httpMethod = requestParameters[@"method"];
+  if (httpMethod) {
+    [request setHTTPMethod:httpMethod];
+  }
+
+  id httpBody = requestParameters[@"body"];
+  if (httpBody && [httpBody isKindOfClass:[FlutterStandardTypedData class]]) {
```
stuartmorgan, line 525: First check is unnecessary again.

packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -450,6 +487,53 @@ - (bool)loadUrl:(NSString*)url withHeaders:(NSDictionary<NSString*, NSString*>*)
...
+    [request setHTTPBody:[httpBody data]];
+  }
+
+  id headers = requestParameters[@"headers"];
+  if (headers && [headers isKindOfClass:[NSDictionary class]]) {
```
stuartmorgan, line 530: And here.

packages/webview_flutter/webview_flutter/lib/platform_interface.dart
```diff
@@ -187,6 +188,30 @@ abstract class WebViewPlatformController {
         "WebView loadUrl is not implemented on the current platform");
   }
 
+  /// Loads the specified URL.
+  ///
+  /// [WebViewRequest] must not be null.
```
stuartmorgan, line 193: Don't document things that are communicated by the type (and enforced by the compiler)

packages/webview_flutter/webview_flutter/lib/platform_interface.dart
```diff
@@ -187,6 +188,30 @@ abstract class WebViewPlatformController {
...
+  /// in the [WebViewLoadMethod].
+  ///
+  /// If [WebViewRequest.headers] is not empty, the key value pairs in the
+  /// [WebViewRequest.headers] will be added as key value pairs of HTTP headers
+  /// for the request.
+  ///
+  /// If [WebViewRequest.body] is not null, it will be added as HTTP body
+  /// for the request.
+  ///
+  /// [WebViewRequest.uri] must not be null.
```
stuartmorgan, line 205: Same.

packages/webview_flutter/webview_flutter/lib/src/types/webview_request.dart
```diff
@@ -0,0 +1,58 @@
...
+  final Uint8List? body;
+
+  /// Serializes the [WebViewRequest] to JSON.
+  Map<String, dynamic> toJson() => {
+        'url': this.uri.toString(),
+        'method': this.method.serialize(),
+        'headers': this.headers,
+        'body': this.body,
+      };
+}
```
stuartmorgan, line 58: Please add simple unit tests of this class, validating that `toJson` works.

packages/webview_flutter/webview_flutter/lib/webview_flutter.dart
```diff
@@ -648,6 +650,31 @@ class WebViewController {
...
+  /// for the request.
+  ///
+  /// [WebViewRequest.uri] must not be null.
+  ///
+  /// Throws an ArgumentError if [WebViewRequest.uri] has empty scheme.
+  Future<void> loadRequest({
+    required WebViewRequest request,
+  }) async {
+    assert(request != null);
+    _validateUri(request.uri);
```
stuartmorgan, line 674: Why not just inline this? It's a trivial method that's only used once.

packages/webview_flutter/webview_flutter/test/webview_flutter_test.dart
```diff
@@ -894,13 +905,14 @@ void main() {
...
       await controller.loadUrl('https://google.com', headers: headers);
 
       expect(platform.lastUrlLoaded, 'https://google.com');
       expect(platform.lastRequestHeaders, headers);
     });
   });
+
   testWidgets('Set UserAgent', (WidgetTester tester) async {
     await tester.pumpWidget(const WebView(
       initialUrl: 'https://youtube.com',
```
stuartmorgan, line 918: Why are there no new tests for the new API?

stuartmorgan (changes requested):

PR #4176, opened 2021-07-21
[flutter_plugin_tools] Replace xctest and java-test with native-test

stuartmorgan:
Creates a new `native-test` command that will be used to run native unit and UI/integration tests for all platforms over time. This replaces both `xctest` and `java-test`.

For CI we can continue to run each platform separately for clarity, but the combined command makes it easier to use (and remember how to use) for local development, as well as avoiding the need to introduce several new commands for desktop testing as support for that is added to the tool.

Fixes https://github.com/flutter/flutter/issues/84392
Fixes https://github.com/flutter/flutter/issues/86489

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
It looks like the overall change broke the rename detection threshold, so it'll be easier to review the individual commits rather than the whole PR.

Also, if you'd rather I split this into "convert xctest to native-test" and "fold java-test into native-test" as separate PRs I can easily do that. I was going to do it initially, but I thought seeing what it looked like with more than one platform's test might make the overall structure clearer.

script/tool/lib/src/native_test_command.dart
```diff
@@ -0,0 +1,377 @@
...
+        error:
+            hasMissingBuild ? 'Examples must be built before testing.' : null);
+  }
+
+  Future<_PlatformResult> _testIos(Directory plugin, _TestMode mode) {
+    return _runXcodeTests(plugin, 'iOS', mode,
+        extraFlags: _iosDestinationFlags);
+  }
+
+  Future<_PlatformResult> _testMacOs(Directory plugin, _TestMode mode) {
```
jmagman, line 242:
```suggestion
  Future<_PlatformResult> _testMacOS(Directory plugin, _TestMode mode) {
```

script/tool/lib/src/native_test_command.dart
```diff
@@ -0,0 +1,377 @@
...
+    Directory plugin,
+    String platform,
+    _TestMode mode, {
+    List<String> extraFlags = const <String>[],
+  }) async {
+    String? testTarget;
+    if (mode.unitOnly) {
+      testTarget = 'RunnerTests';
+    } else if (mode.integrationOnly) {
+      testTarget = 'RunnerUITests';
```
jmagman, line 261: You don't need to address it in this PR since I know this was copied, but I just remembered about the image_picker `RunnerUITestiOS14` target.  It probably doesn't make sense for that to be a separate target, I wasn't involved in the recent creation of that.

jmagman (approved):
LGTM w/ spelling nit

script/tool/lib/src/native_test_command.dart
stuartmorgan: Yes, I found that when checking the repo to make sure this would work, and discussed it with folks this morning; it should get folded back in in the next day or two.

script/tool/lib/src/native_test_command.dart
stuartmorgan: I never know what to do with iOS and macOS in Dart code :P I hadn't realized there was [an exception for two-letter acronyms](https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words) in Dart style though, thanks for flagging this!

PR #4198, opened 2021-07-27
Remove pubspec.yaml examples from READMEs

stuartmorgan:
Demonstrations of how to add a plugin dependency in pubspec.yaml
inevitably become outdated, causing more confusion than they solve.
pub.dev has an installation tab which shows the same example but using a
dynamic version, so these instructions are almost never necessary; where
specifically referencing installation is useful, link to that tab
instead of an inline, hard-coded example.

Many of these examples were in implementation packages that were not
endorsed, and had never been updated to indicate that they are now
endorsed, so this PR also corrects those READMEs to indicate that they
are in fact endorsed.

Fixes https://github.com/flutter/flutter/issues/87056

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/e2e/README.md
```diff
@@ -1,195 +1,3 @@
 # e2e (deprecated)
 
-## DEPRECATED
-
```
stuartmorgan, line 4: This package is several deprecations old, and only this README exists in the tree (in case people have links to the package), so rather than do a targeted update I just nuked everything but the pointer.

ditman (approved):
LGTM, I like the new verbiage much better. Thanks for doing this!

PR #4210, opened 2021-07-31
[camera_web] Recording Video

ABausG:
Adds ability to record `webm` of the camera platform interface.

Implements calls `startVideoRecording`, `pauseVideoRecording`, `resumeVideoRecording` and `stopVideoRecording`

Part of flutter/flutter#45297.

## Pre-launch Checklist

- [x] I signed the [CLA].


ABausG:
@bselwe @felangel @ditman I saw that you did some work on camera_web recently. Could you take a look at this?

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -171,4 +174,69 @@ class Camera {
...
+  Future<XFile> stopVideoRecording() async {
+    if (_mediaRecorder == null || _mediaRecorder!.state == 'inactive') {
+      throw html.DomException.INVALID_STATE;
+    }
+    final availableData = Completer<XFile>();
+    _mediaRecorder!.addEventListener('dataavailable', (event) {
+      final blob = (event as html.BlobEvent).data;
+      availableData.complete(XFile(html.Url.createObjectUrl(blob)));
+    });
+    _mediaRecorder?.stop();
```
ditman, line 238: There seems to be some code duplication here vs startVideoRecording (with a max length).

Could you unify these two codepaths so the implementation of the`dataavailable` event on `startVideoRecording` uses the same logic as "stop"?

I also wonder why the start with max length emits a `VideoRecordedEvent`, but `stopVideoRecording` doesn't?

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -171,4 +174,69 @@ class Camera {
...
+
+    if (maxVideoDuration != null) {
+      _mediaRecorder!.addEventListener('dataavailable', (event) {
+        final blob = (event as html.BlobEvent).data;
+        final file = XFile(html.Url.createObjectUrl(blob));
+        _videoRecorderController
+            .add(VideoRecordedEvent(this.textureId, file, maxVideoDuration));
+        _mediaRecorder!.stop();
+      });
+      _mediaRecorder!.start(maxVideoDuration.inMilliseconds);
```
ditman, line 203: What happens if `maxVideoDuration` is `0 milliseconds`? Should that be disallowed?

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -171,4 +174,69 @@ class Camera {
...
+    }
+    _mediaRecorder?.pause();
+  }
+
+  /// Resumes a video Recording
+  /// Throws a [html.DomException.INVALID_STATE] if there is no active Recording
+  Future<void> resumeVideoRecording() async {
+    if (_mediaRecorder == null || _mediaRecorder!.state == 'inactive') {
+      throw html.DomException.INVALID_STATE;
+    }
```
ditman, line 223: (These exceptions are already thrown by the browser, no need to assert for _mediaRecorder.state)

ditman (commented):
Thanks @ABausG for the contribution, this is great!

I've left some questions about the current approach (mostly about cross-browser compat).

BTW, @bselwe is currently OOTO (until the end of the week), but he was working in normalizing the exceptions thrown from the code.

In this PR, for example, we're manually throwing DomExceptions from the low-level Camera object, but I think @bselwe was working on getting some standardization in *what* types of exceptions to throw *from where*. The guiding lines here would be the core `camera` plugin, seeing what types of Exceptions are expected there, so everything works transparently.

I'll defer to @bselwe for more comments in that regard!

Other than that, this is looking amazing! Thanks again for the contribution!

bselwe (commented):
Awesome, thanks for the contribution @ABausG and your comments @ditman! 🙌

I've added some clarification regarding which exceptions to throw and when. The general rule would be to try to match the exceptions thrown in [`MethodChannelCamera`](https://github.com/flutter/plugins/blob/08c9b40d020bf249f346322ec395220e36889b42/packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart#L23) (an implementation of the camera platform used for iOS and Android platforms) and handled in [`CameraController`](https://github.com/flutter/plugins/blob/08c9b40d020bf249f346322ec395220e36889b42/packages/camera/camera/lib/src/camera_controller.dart#L204). Let me know if there's anything unclear about this approach.

I think it would also be nice to include `Camera` tests. That could include:
- For `startVideoRecording`:
  - Throws an exception when the media recorder is in an invalid state.
  - Initializes a media recorder instance from the video element stream.
  - Uses an appropriate mime type for the video recording based on `MediaRecorder.isTypeSupported()`.
  - Uses `maxVideoDuration` to record a video.
  - Emits a `VideoRecordedEvent` when the video recording has been stopped.
- For `pauseVideoRecording`:
  - Throws an exception when the media recorder is null.
  - Calls pause on the media recorder.
- And similar tests for other methods.

Thanks again!

bselwe (commented):
Should we maybe remove throwing an unimplemented error in [`prepareForVideoRecording`](https://github.com/flutter/plugins/pull/4210/files#diff-95b0d214cebd76a613f66fbe94ad5a3cbccd21e0084d34b83b8589d2a14b5ed3R355)?

ABausG:
> Should we maybe remove throwing an unimplemented error in [`prepareForVideoRecording`]

I was thinking about that method as well. I'm seeing two options:
1. Make it no-op
2. Perform initialisation in the actual camera implementation like setting up the media recorder.

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -171,4 +174,69 @@ class Camera {
...
+      _videoRecorderController.stream;
+
+  /// Starts a new Video Recording using [html.MediaRecorder]
+  /// /// Throws a [html.DomException.INVALID_STATE] if there already is an active Recording
+  Future<void> startVideoRecording({Duration? maxVideoDuration}) async {
+    if (_mediaRecorder != null && _mediaRecorder!.state != 'inactive') {
+      throw html.DomException.INVALID_STATE;
+    }
+    _mediaRecorder ??= html.MediaRecorder(
+        videoElement.captureStream(), {'mimeType': 'video/webm'});
```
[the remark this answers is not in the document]
ABausG: I've added a List of possible mimeTypes and changed to use the first supported one. Right now I've added `video/mp4` and `video/webm` not sure what formats/order to add them. Any advice/suggestion?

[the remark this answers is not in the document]
ABausG: Also using the `srcObject` as suggested by @bselwe

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -171,4 +174,69 @@ class Camera {
...
+
+  /// Returns a Stream that emits when a video Recodring with a defined maxVideoDuration was created
+  Stream<VideoRecordedEvent> get onVideoRecordedEventStream =>
+      _videoRecorderController.stream;
+
+  /// Starts a new Video Recording using [html.MediaRecorder]
+  /// /// Throws a [html.DomException.INVALID_STATE] if there already is an active Recording
+  Future<void> startVideoRecording({Duration? maxVideoDuration}) async {
+    if (_mediaRecorder != null && _mediaRecorder!.state != 'inactive') {
+      throw html.DomException.INVALID_STATE;
```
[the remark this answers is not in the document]
ABausG: Throwing a `PlatformException` now and not checking fo the `state` of the mediaRecorder`

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -171,4 +174,69 @@ class Camera {
...
+    }
+    _mediaRecorder ??= html.MediaRecorder(
+        videoElement.captureStream(), {'mimeType': 'video/webm'});
+
+    if (maxVideoDuration != null) {
+      _mediaRecorder!.addEventListener('dataavailable', (event) {
+        final blob = (event as html.BlobEvent).data;
+        final file = XFile(html.Url.createObjectUrl(blob));
```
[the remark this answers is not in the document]
ABausG: Added mimeType and Filename. For now using the `hashcode` of the blob as filename

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -171,4 +174,69 @@ class Camera {
...
+      _mediaRecorder!.start(maxVideoDuration.inMilliseconds);
+    } else {
+      _mediaRecorder!.start();
+    }
+  }
+
+  /// Pauses the current video Recording
+  /// Throws a [html.DomException.INVALID_STATE] if there is no active Recording
+  Future<void> pauseVideoRecording() async {
+    if (_mediaRecorder == null || _mediaRecorder!.state == 'inactive') {
```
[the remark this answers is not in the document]
ABausG: See above. Now throwing PlatformExceptions

packages/camera/camera_web/lib/src/camera.dart
ABausG: Unified it by registering the listener on `dataavailable` only in the startVideoRecording.

If the listener is triggered it will do the following things:
- emit a `VideoRecordedEvent`
- Complete the Completer used to obtain the `XFile` in `stopVideoRecording`
- remove the listener
- stop the MediaRecorder
   - This stop is necessary to only get data once if a `maxVideoDuration` is provided as `MediaRecorder` only takes in splices instead of a maxDuration

packages/camera/camera_web/lib/src/camera.dart
ABausG: Throwing a PlatformException

packages/camera/camera_web/lib/src/camera.dart
ABausG: See above

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -171,4 +174,69 @@ class Camera {
...
+  /// /// Throws a [html.DomException.INVALID_STATE] if there already is an active Recording
+  Future<void> startVideoRecording({Duration? maxVideoDuration}) async {
+    if (_mediaRecorder != null && _mediaRecorder!.state != 'inactive') {
+      throw html.DomException.INVALID_STATE;
+    }
+    _mediaRecorder ??= html.MediaRecorder(
+        videoElement.captureStream(), {'mimeType': 'video/webm'});
+
+    if (maxVideoDuration != null) {
+      _mediaRecorder!.addEventListener('dataavailable', (event) {
```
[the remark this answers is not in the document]
ABausG: Yes. Removing it now

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -171,4 +174,69 @@ class Camera {
       ..objectFit = 'cover'
       ..transform = 'scaleX(-1)';
   }
+
+  html.MediaRecorder? _mediaRecorder;
+  final StreamController<VideoRecordedEvent> _videoRecorderController =
+      StreamController();
+
+  /// Returns a Stream that emits when a video Recodring with a defined maxVideoDuration was created
+  Stream<VideoRecordedEvent> get onVideoRecordedEventStream =>
```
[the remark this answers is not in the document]
ABausG: I renamed it to `onVideoRecordedEvent` to match the naming of the getter in the `camera_plugin_interface`

packages/camera/camera_web/lib/src/camera.dart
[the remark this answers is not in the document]
ditman: /cc @bselwe look, YUV 4:2:0 and friends are starting to show up here: https://developer.mozilla.org/en-US/docs/Web/Media/Formats/codecs_parameter#av1 (I don't think we should mess with the `codecs` value, unless strictly necessary)

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -297,5 +298,170 @@ void main() {
...
+            camera.resumeVideoRecording,
+            throwsA(predicate<PlatformException>((ex) =>
+                ex.code ==
+                CameraErrorCode.mediaRecordingNotStarted.toString())));
+      });
+    });
+
+    group('stopVideoRecording', () {
+      testWidgets('stops a video recording and returns a File', (tester) async {
+        // TODO(abausg) There seems to be an issue with removing the listener and stopping the recorder
```
ditman, line 430: Are you still tracking this down @ABausG?

ABausG: I've managed to get this test to work 🙌

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -297,5 +298,175 @@ void main() {
...
+
+        await camera.startVideoRecording();
+
+        expect('recording', camera.mediaRecorder!.state);
+      });
+
+      testWidgets(
+          'starts a video recording with a given maxDuration '
+          'emits a VideoRecordedEvent', (tester) async {
+        // TODO: MediaRecorder with a Timeslice does not seem to call the dataavailable Listener in a Test
```
ABausG, line 320: I could not get this test to work. As it does not seem to trigger the datavailable event of the media recorder.
I was able to manually test the behaviour with setting a max duration (I tested with 4 seconds) however this test seems to timeout waiting for the first element of the event stream.
Could you take a look at what's might be going on? 
Also is there a (nice) way to debug the integration tests?

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -297,5 +298,175 @@ void main() {
...
+          textureId: 1,
+          cameraSettings: cameraSettings,
+        );
+
+        await camera.initialize();
+        await camera.play();
+
+        await camera.startVideoRecording();
+
+        expect('recording', camera.mediaRecorder!.state);
```
bselwe, line 314:
```suggestion
        expect(camera.mediaRecorder!.state, equals('recording`));
```

packages/camera/camera_web/example/integration_test/camera_test.dart
bselwe: I will have a look at this test.

There is a way to debug the integration tests in the browser. You'll need to execute this in `camera_web/example`:

```
flutter run -d web-server --target integration_test/camera_test.dart --debug
```

It will produce a link that you should open in the browser, open inspection tools (CMD+Option+J in Chrome), find a dart file you want to debug in the "Source" tab and mark a breakpoint. Refresh the web page and it should stop at the breakpoint.

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -297,5 +298,175 @@ void main() {
...
+          textureId: 1,
+          cameraSettings: cameraSettings,
+        );
+
+        await camera.initialize();
+        await camera.play();
+        expect(
+            () => camera.startVideoRecording(maxVideoDuration: Duration.zero),
+            throwsA(predicate<PlatformException>(
+                (ex) => ex.code == CameraErrorCode.notSupported.toString())));
```
bselwe, line 354: nit 
```suggestion
            throwsA(predicate<PlatformException>(
                (ex) => ex.code == CameraErrorCode.notSupported.toString(),),),);
```

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -297,5 +298,175 @@ void main() {
...
+          cameraSettings: cameraSettings,
+        );
+
+        await camera.initialize();
+        await camera.play();
+        await camera.startVideoRecording();
+
+        await camera.pauseVideoRecording();
+
+        expect('paused', camera.mediaRecorder!.state);
```
bselwe, line 371:
```suggestion
        expect(camera.mediaRecorder!.state, equals('paused'));
```

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -297,5 +298,175 @@ void main() {
...
+        await camera.initialize();
+        await camera.play();
+
+        await camera.startVideoRecording();
+
+        await camera.pauseVideoRecording();
+
+        await camera.resumeVideoRecording();
+
+        expect('recording', camera.mediaRecorder!.state);
```
bselwe, line 409:
```suggestion
        expect(camera.mediaRecorder!.state, equals('recording'));
```

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -297,5 +298,175 @@ void main() {
...
+
+        await camera.initialize();
+        await camera.play();
+
+        expect(
+            camera.stopVideoRecording,
+            throwsA(predicate<PlatformException>((ex) =>
+                ex.code ==
+                CameraErrorCode.mediaRecordingNotStarted.toString())));
+      });
```
bselwe, line 469: nit 
```suggestion
            throwsA(predicate<PlatformException>((ex) =>
                ex.code ==
                CameraErrorCode.mediaRecordingNotStarted.toString(),),),);
      });
```

packages/camera/camera_web/example/integration_test/camera_web_test.dart
```diff
@@ -720,35 +720,225 @@ void main() {
...
+
+        // Save the camera in the camera plugin.
+        (CameraPlatform.instance as CameraPlugin).cameras[cameraId] = camera;
+
+        final video =
+            await CameraPlatform.instance.stopVideoRecording(cameraId);
+
+        verify(camera.stopVideoRecording).called(1);
+
+        expect(video, capturedVideo);
```
bselwe, line 793: nit
```suggestion
        expect(video, equals(capturedVideo));
```

bselwe:
Hey @ABausG, could you give me push access to your fork so I could continue working on this?

bselwe:
@ABausG Is there anything that you think might be missing for the video recording feature?

ABausG:
Apart from the one test I think it should be complete

bselwe:
@googlebot I consent.

ditman:
I've tried to update this with the latest master to look at the test that is commented out, but I think there's other problem with the branch; I think there might be a problem regarding `camera.dispose()` (and I might have been the cause, of course :P)

`camera_test.dart` wouldn't finish, and it seems to me that the culprit were the two tests that have `await camera.dispose()`?

(not entirely sure it has to do with the tests that are failing in this PR)

bselwe:
Yeah, thanks for checking @ditman, I'll fix that. It seems that after adding `await` to  `_videoRecorderController.close` in `Camera.dispose` tests fail. Updating `_videoRecorderController` to a broadcast stream controller fixes this issue.

ditman:
> Updating `_videoRecorderController` to a broadcast stream controller fixes this issue.

Ahhh, very good catch there @bselwe!

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -365,23 +390,155 @@ class Camera {
...
+
+    // Remove a data listener before stopping the recorder.
+    mediaRecorder!.removeEventListener(
+      'dataavailable',
+      _videoDataAvailableListener,
+    );
+
+    // Stop the recorder if the video has a maxVideoDuration
+    // and the recording was not stopped manually.
+    if (maxVideoDuration != null && mediaRecorder!.state == 'recording') {
```
ditman, line 452: This should also check if we have recorded > time than maxVideoDuration, right?

ditman (changes requested):
I have 2 general comments about this PR; I don't understand very well how these features are currently implemented:

* How does this respect the max length of a video recording passed by a user? I don't think we ever check or use the value of max length (only null checks).
* How does this combine several `ondataavailable` events with video "chunks" to produce one single, large video file that can be used later? It seems this code emits the File as soon as the first `ondataavailable` event fires.
  * See the how they use `ondataavailable` and `stop` events [here](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/ondataavailable#example) (it's for audio, but the API is the same).
  * Is this even supported by XFile, do we need to change anything there? Do we need to create a File from an html.File, or a bunch of Blobs or something other than "bytes"?
* @bselwe noticed that video on Chrome/Edge is missing some metadata, and the duration is listed as `Infinity`; have you noticed this? It seems to be [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=642012). [SO1](https://stackoverflow.com/questions/38443084/how-can-i-add-predefined-length-to-audio-recorded-from-mediarecorder-in-chrome/39971175#39971175), [SO2](https://stackoverflow.com/questions/50586612/seeking-is-not-working-in-recorded-video-with-mediarecorder-api). (Is this something that can be fixed when putting together the XFile from all the "chunks"? Does the mp4 codec have this problem?)

For now, I'm going to publish the camera_web plugin as is, and when this PR matures a little bit more, we can release it in a new version. This is a great start, but I'm a little concerned about the comments above :S

ditman:
> How does this respect the max length of a video recording passed by a user? I don't think we ever check or use the value of max length (only null checks).

We discussed this, and it seems that we're leveraging the `timeslice` feature (that I had misunderstood), and we only use one blob of video, until we call stop, or the `timeslice` is done.

> How does this combine several ondataavailable events with video "chunks" to produce one single, large video file

This is already achieved by the behavior of the `ondataavailable` event, and @bselwe just made it so that we use the stop event to collect the recorded video, closer to the example.

ditman:
@googlebot I consent.

ditman (approved):
LGTM! All tests pass locally, let's go with this one!

PR #4218, opened 2021-08-05
[path_provider_linux] Using TMPDIR env as a primary temporary path

proninyaroslav:
TMPDIR is a standard variable on UNIX/Linux systems, and is often used in containers such as Flatpak to redirect to a temporary folder inside a sandbox. This allows not to make hard bindings to the /tmp directory

Fixes https://github.com/flutter/flutter/issues/87742

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
+++ b/packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
@@ -4,6 +4,7 @@
 
 import 'dart:io';
 
+import 'package:flutter/foundation.dart';
 import 'package:path/path.dart' as path;
 import 'package:path_provider_platform_interface/path_provider_platform_interface.dart';
 import 'package:xdg_directories/xdg_directories.dart' as xdg;
@@ -12,15 +13,36 @@ import 'package:xdg_directories/xdg_directories.dart' as xdg;
 ///
 /// This class implements the `package:path_provider` functionality for linux
 class PathProviderLinux extends PathProviderPlatform {
+  /// Constructs an instance of [PathProviderLinux]
+  PathProviderLinux() : _environment = Platform.environment;
+
+  /// Constructs an instance of [PathProviderLinux] with the given [environment]
+  @visibleForTesting
+  PathProviderLinux.private({
+    required Map<String, String> environment,
+  }) : _environment = environment;
+
+  final Map<String, String> _environment;
+
   /// Registers this class as the default instance of [PathProviderPlatform]
   static void registerWith() {
     PathProviderPlatform.instance = PathProviderLinux();
   }
 
+  /// Registers this class as the default instance of [PathProviderPlatform]
+  /// with the given [environment]
+  @visibleForTesting
+  static void privateRegisterWith({
+    required Map<String, String> environment,
+  }) {
+    PathProviderPlatform.instance = PathProviderLinux.private(
+      environment: environment,
+    );
+  }
+
   @override
   Future<String?> getTemporaryPath() {
-    final Map<String, String> env = Platform.environment;
-    final String? tmpdir = env['TMPDIR'];
+    final String? tmpdir = _environment['TMPDIR'];
     return Future<String?>.value(
       tmpdir == null || tmpdir.isEmpty ? '/tmp' : tmpdir,
     );
--- a/packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
+++ b/packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
@@ -20,7 +20,8 @@ class PathProviderLinux extends PathProviderPlatform {
   @override
   Future<String?> getTemporaryPath() {
     final Map<String, String> env = Platform.environment;
-    return Future<String?>.value(env['TMPDIR'] ?? '/tmp');
+    final String? tmpdir = env['TMPDIR'];
+    return Future<String?>.value(tmpdir == null || tmpdir.isEmpty ? '/tmp' : tmpdir);
   }
 
   @override
--- a/packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
+++ b/packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
@@ -29,17 +29,6 @@ class PathProviderLinux extends PathProviderPlatform {
     PathProviderPlatform.instance = PathProviderLinux();
   }
 
-  /// Registers this class as the default instance of [PathProviderPlatform]
-  /// with the given [environment]
-  @visibleForTesting
-  static void privateRegisterWith({
-    required Map<String, String> environment,
-  }) {
-    PathProviderPlatform.instance = PathProviderLinux.private(
-      environment: environment,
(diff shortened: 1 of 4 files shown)


stuartmorgan:
Thanks for the submission! This will need the remaining items in the checklist addressed before it moves forward with review. If you have any questions about completing those steps that aren’t addressed in the linked documentation, please let me know.

proninyaroslav:
@stuartmorgan 
I have questions about this: 
```
I updated pubspec.yaml with an appropriate new version according to the pub versioning philosophy.
I updated CHANGELOG.md to add a description of the change.
```
Do I have to make a version bump in this case?

proninyaroslav:
@stuartmorgan 
About tests: the problem is that the `Platform` class from the `dart:io` library is not mockable, so I can't insert the required environment variable value inside the test (`Platform.environment`). The solution would be to use the mockable version of `Platform`, some plugins use this https://pub.dev/packages/platform

stuartmorgan:
> I have questions about this:
> 
> ```
> I updated pubspec.yaml with an appropriate new version according to the pub versioning philosophy.
> I updated CHANGELOG.md to add a description of the change.
> ```
> 
> Do I have to make a version bump in this case?

The section you are quoting links to an explanation of the versioning policy for flutter/plugins. Could you elaborate on what your are seeing there that suggests you would not need to so that I can clarify it?

> About tests: the problem is that the `Platform` class from the `dart:io` library is not mockable, so I can't insert the required environment variable value inside the test (`Platform.environment`). The solution would be to use the mockable version of `Platform`, some plugins use this https://pub.dev/packages/platform

Since all you need in the `environment`, a much simpler solution at this point would just be to make `PathProviderLinux` take an injectable-by-tests environment map that would replace the call to `Platform.environment`. We generally try to minimize package dependencies in the plugins here.

proninyaroslav:
@stuartmorgan 
I added tests and also bumped the plugin version to 2.1.0 as I believe the API changes are backwards compatible and are not just a patch.

packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
```diff
@@ -12,14 +13,39 @@ import 'package:xdg_directories/xdg_directories.dart' as xdg;
...
+
   /// Registers this class as the default instance of [PathProviderPlatform]
   static void registerWith() {
     PathProviderPlatform.instance = PathProviderLinux();
   }
 
+  /// Registers this class as the default instance of [PathProviderPlatform]
+  /// with the given [environment]
+  @visibleForTesting
+  static void privateRegisterWith({
```
stuartmorgan, line 35: Why does this method exist? `registerWith` is called automatically; nothing will call this automatically, so it's not clear what value this is providing over the tests just setting `instance` directly, and it seems actively confusing as it's named in the same style as something that *is* called automatically.

packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
```diff
@@ -12,14 +13,39 @@ import 'package:xdg_directories/xdg_directories.dart' as xdg;
...
+  }) {
+    PathProviderPlatform.instance = PathProviderLinux.private(
+      environment: environment,
+    );
+  }
+
   @override
   Future<String?> getTemporaryPath() {
-    return Future<String?>.value('/tmp');
+    final String? tmpdir = _environment['TMPDIR'];
```
stuartmorgan, line 45: How about `environmentTmpDir` as the name here, to make its role clearer. This current name doesn't communicate that it doesn't already include the fallback.

packages/path_provider/path_provider_linux/lib/path_provider_linux.dart
```diff
@@ -12,14 +13,39 @@ import 'package:xdg_directories/xdg_directories.dart' as xdg;
...
+    return Future<String?>.value(
+      tmpdir == null || tmpdir.isEmpty ? '/tmp' : tmpdir,
```
stuartmorgan, line 47: You could simplify this slightly by making the variable above non-nullable with ` = _environment['TMPDIR'] ?? ''`, then just doing an `isEmpty` here.

packages/path_provider/path_provider_linux/test/path_provider_linux_test.dart
```diff
@@ -7,7 +7,15 @@ import 'package:path_provider_platform_interface/path_provider_platform_interfac
 
 void main() {
   TestWidgetsFlutterBinding.ensureInitialized();
-  PathProviderLinux.registerWith();
```
stuartmorgan, line 10: You are removing critical test coverage here; this was ensuring that `registerWith` worked.

packages/path_provider/path_provider_linux/test/path_provider_linux_test.dart
```diff
@@ -7,7 +7,15 @@ import 'package:path_provider_platform_interface/path_provider_platform_interfac
...
+
+  late Map<String, String> environment;
+
+  setUp(() {
+    environment = <String, String>{};
+    PathProviderLinux.privateRegisterWith(
+      environment: environment,
+    );
+  });
```
stuartmorgan, line 18: `setUp` should only be used when something is needed in multiple tests, which isn't the case here. You can just make a local `environment` and a local `PathProviderLinux` instance in a test specifically of environment variable handling.

packages/path_provider/path_provider_linux/test/path_provider_linux_test.dart
```diff
@@ -16,6 +24,12 @@ void main() {
   test('getTemporaryPath', () async {
     final PathProviderPlatform plugin = PathProviderPlatform.instance;
     expect(await plugin.getTemporaryPath(), '/tmp');
```
stuartmorgan, line 26: Unless I'm missing something, this will fail if someone runs it on a machine with TMPDIR set to something non-default. If you want to explicitly test null, you should be using an environment created by the test that doesn't contain TMPDIR, rather than assuming the host environment doesn't have it.

stuartmorgan (changes requested):

packages/path_provider/path_provider_linux/test/path_provider_linux_test.dart
proninyaroslav: We pass `Map<String, String>` as environment , which is initially empty. Therefore, there can be no conflict.

proninyaroslav:
@stuartmorgan 
All comments have been resolved.

proninyaroslav:
@stuartmorgan 
Can you review my changes again?

stuartmorgan (approved):
Sorry for the delay. Looking again, the tests would be much clearer as three distinct tests rather than one test that alternates between changing the environment and then making new assertions. Rather than do another round of review, I just made those change in the PR.

LGTM with that. Thanks!

stuartmorgan:
Oops, this got caught by the Windows bot change so it was never auto-landed. Landing manually, since this can't affect Windows.

PR #4221, opened 2021-08-09
[url_launcher] Fixed missing # in links href when opening in new tab on the web

TheOneWithTheBraid:
When using the `Link` widget, it's theoretically possible to open the link in a new tab, e.g. by dragging the widget in the tab bar of modern browsers. In This case, up to now, an invalid route was formed, as the leading `#` from Flutters named routes is not included in the Anchor node's `href`.

- Fixes https://github.com/flutter/flutter/issues/87941

*I'm not exactly sure whether it's possible to cover a behavior like drag outside of Flutter to a new tab in tests. As I don't think so, I left out such a test.*

## Pre-launch Checklist

- [x] I signed the [CLA].


TheOneWithTheBraid:
@googlebot I signed it!

stuartmorgan:
Thanks for the submission.

> _I'm not exactly sure whether it's possible to cover a behavior like drag outside of Flutter to a new tab in tests. As I don't think so, I left out such a test._

Just because an end-to-end test may not be feasible does not mean a change can't have tests. You are transforming a property in a setter, which should be easily unit testable.

In the future, please don't check things in the checklist that you have not actually done. Asserting yourself that you don't think it needs a test isn't one of the options described here.

TheOneWithTheBraid:
You are fully right, @stuartmorgan. I definitely should not tick a checkbox then there is an unanswered question on it, and you are right with the getter too. Thank you for the reminder. I just added a corresponding test, and it's passing (at least on my local machine).

packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -220,7 +219,13 @@ class LinkViewController extends PlatformViewController {
     if (uri == null) {
       _element.removeAttribute('href');
     } else {
-      _element.setAttribute('href', uri.toString());
+      String href = uri.toString();
+      // in case an internal uri is given, a leading `#` must be added to avoid
+      // invalid links when opening in a new tab
+      if (!uri.hasScheme) {
+        href = '#' + href;
+      }
```
ditman, line 227: I am not entirely sure this is going to be true forever, for example, there's a new Uri Strategy that doesn't use the `#` and instead uses the browser history API.

/cc @mdebbar can you take a look at this fix and see if it can be applied as is?

mdebbar: You are right @ditman. This works only with the [`HashUrlStrategy`](https://github.com/flutter/flutter/blob/3216e9f1690fdf822a312eaadce5378e751ad601/packages/flutter_web_plugins/lib/src/navigation/url_strategy.dart#L86). If someone uses the [`PathUrlStrategy`](https://github.com/flutter/flutter/blob/3216e9f1690fdf822a312eaadce5378e751ad601/packages/flutter_web_plugins/lib/src/navigation/url_strategy.dart#L172) or any other custom url strategy, it won't work.

We need to figure out a way to grab the current url strategy used by the app, and call [`prepareExternalUrl`](https://github.com/flutter/flutter/blob/3216e9f1690fdf822a312eaadce5378e751ad601/packages/flutter_web_plugins/lib/src/navigation/url_strategy.dart#L46) to obtain the correct path that should go in the link href.

ditman: @TheOneWithTheBraid: at the plugins level we can't know (for now) what's the URL strategy picked by the current app (even if we exposed that information from the engine, it'd take a few months for it to trickle down to the `stable` channel), to add a `#` or not.

However, programmers of web apps *do know* what type of links their web app needs.

As an alternative solution, have you tried setting your link widget with a URL similar to:

`#/another/internal/route` instead of `/another/internal/route` (I think this was the example in the originally reported issue)?

Is the problem then that when clicking on the URL it doesn't work?

I think it might be easier to teach the Link widget/Flutter framework to recognize those URLs when dealing with the onClick of the Link element, rather than attempting to "fix" the URL that the programmer is setting in links.

(Note that now we have 2 URL strategies, but if more end up being added, we might end up needing some extra behavior for each of them!)

Apologies if I'm misunderstanding the problem, but as far as I'm currently understanding this, I think we're attempting a fix in the wrong component!

TheOneWithTheBraid: Yes, I see the point. It definitely takes very long, but anyway, at some point in the future, there should be a way to check the current URL strategy.

> Is the problem then that when clicking on the URL it doesn't work?

No, it's not when clicking (as this is handled by Flutter internally in Dart and preventing the browser-level handling of links) but when *dragging* the Anchor element to e.g. to the address bar to open in a new tab, *opening a context menu* and copy the link etc.

If I correctly see the current behavior, setting the link to `#/myRoute` is not working right now but ending up in `##/myRoute for Flutter-internal handling but perfectly working for browser-level actions as described above.

As this issue is really not urgent and only occurs in edge cases, it's maybe worth the waiting to implement the corresponding handlers in the engine as sometime in the Future, it might be useful to access to current URL strategy anyway, right?

ditman: > it might be useful to access to current URL strategy anyway, right?

Yes, I don't have any problem with some of that information being exposed (or maybe a Link generator component that is compatible with the current URL strategy). The biggest problem of exposing these web-only APIs (for now) is that there's not a good place to expose them in a consistent way. (this is not a problem exclusive of this particular API, it's a flutter web issue in general!)

The purpose of the link is to work as a "normal" `a href`, with regards to clicking, copy-pasting and dragging to the address bar; we need to review that flutter is not doing "too much" with links for them to work (because the browser will *not* do that much)

mdebbar: > May you tell me whether my idea is right:
> 
> Would it be possible to somehow publicly make available `_customUrlStrategy` in `engine:lib/web_ui/lib/src/engine/window.dart:38`?
> 
> That way, the platform interface could check the present url strategy, right?
> 
> Would you mind if I create a corresponding getter in the file?


Yes, this should work.


> However, programmers of web apps _do know_ what type of links their web app needs.

The idea is that inside your app, you don't need to know how the URL is represented in the browser. It could be a simple path `/route` or a hash path `#/route` or something completely different, depending on the url strategy. But your app shouldn't care about this difference. Only the url strategy needs to know about this (and you could implement your own url strategy if you want to customize this).

The url strategy has two methods to handle this:

- `prepareExternalUrl`: converts an app route name to the path that goes in the browser url bar and `<a>` elements, etc. For example, `HashUrlStrategy.prepareExternalUrl` adds a `#` to the beginning of the route name to make it a hash path.

- `getPath`: converts a path from the browser url bar to a route name for the app to consume. For example, `HashUrlStrategy.getPath` gets the hash part of the browser location, and strips the `#` from it.


> (Note that now we have 2 URL strategies, but if more end up being added, we might end up needing some extra behavior for each of them!)

It doesn't matter how many url strategies there are (actually, users can create their own custom url strategies today, so we already have more than two). The Link widget shouldn't do any `#`-specific behavior/logic. It should just call `UrlStrategy.prepareExternalUrl` and let it handle the hash if necessary.

ditman: > maybe a Link generator component that is compatible with the current URL strategy

Ahh, there you go @mdebbar, the `prepareExternalUrl` from each strategy!

TheOneWithTheBraid: I now opened https://github.com/flutter/flutter/pull/90708 exposing a getter required to call `prepareExternalUrl`. Let's wait for it to trickle down, I'd propose. Right?

stuartmorgan:
Marking as a draft to remove it from our review queue; please remove the draft designation once this can proceed on stable.

stuartmorgan:
Was the necessary change in 2.8? I see that the commit has some 2.8.0-*.pre tags, which suggest yes, but I know there were some version oddities on master.

TheOneWithTheBraid:
It seems like the [`urlStrategy`](https://api.flutter.dev/flutter/flutter_web_plugins/urlStrategy.html) getter trickled down to stable. Could you check whether this PR is now eligible for merging?

TheOneWithTheBraid:
The PR is now rebased on the current HEAD. It would be wonderful to have it merged before new conflicts occur.

stuartmorgan:
> Could you check whether this PR is now eligible for merging?

> It would be wonderful to have it merged before new conflicts occur.

We don't merge unreviewed PRs; this PR has not been approved by any reviewer.

packages/url_launcher/url_launcher_web/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.9
+
+- Fixed invalid routes when opening a `Link` in a new tab
```
stuartmorgan, line 3: This needs to be updated to follow [our CHANGELOG style](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style)

TheOneWithTheBraid: Here you are.

packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -11,7 +11,8 @@ import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/services.dart';
-
+// used to access the current [UrlStrategy]
+import 'package:flutter_web_plugins/flutter_web_plugins.dart';
```
mdebbar, line 15: I don't think you need to explain why the import exists.

You can do the following to make it clear that you only need one thing from this package:
```suggestion
import 'package:flutter_web_plugins/flutter_web_plugins.dart' show urlStrategy;
```

mdebbar (approved):
Looks good to me! I would love for @ditman to take a look too in case there's some plugins-specific stuff that I missed.

ditman (approved):
If @mdebbar is happy, I'm happy. I checked that the minimum version of flutter is >=2.10.0 so this only resolves for users in a modern enough version of the stable channel.

Ship it!

PR #4228, opened 2021-08-11
[image_picker] Fix pickImage not returning on iOS when dismissing the PHPicker view by swiping down.

BeMacized:
When the PHPicker view is used for picking images on iOS, methods like `pickImage` do not return a value when dismissing the view by swiping down. It only sends a result when explicitly pressing the cancel action, or when any content is picked.

This PR fixes this by also sending a result when the view is dismissed. 

Relevant issue: 
- flutter/flutter#87330

## Pre-launch Checklist

- [x] I signed the [CLA].
  - None updated/added as no public methods were altered.
  - This is a simple bug fix that should not affect any current tests. If this is something that has to be tested, let me know.


nikitadol (approved):

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -373,6 +375,16 @@ - (NSNumber *)getDesiredImageQuality:(NSNumber *)imageQuality {
...
 
+- (void)presentationControllerDidDismiss:(UIPresentationController *)presentationController {
+  dispatch_queue_t backgroundQueue =
+      dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0);
+  dispatch_async(backgroundQueue, ^{
+    self.result(nil);
+    self.result = nil;
+    self->_arguments = nil;
+  });
+}
```
renefloor, line 386: Results should be called on the main thread.

I would also check if `self.result` is not nil.

renefloor (changes requested):

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
BeMacized: Fixed

renefloor (approved):

qualitattechs:
can you check this one please ?
https://github.com/flutter/flutter/issues/78391

PR #4246, opened 2021-08-16
[flutter_plugin_tool] Don't allow NEXT on version bumps

stuartmorgan:
The special "NEXT" entry in a CHANGELOG should never be present in a
commit that bumped the version. This validates that this is true even if
the CHANGELOG would be correct for a non-version-change state, to catch
someone incorrectly resolving a merge conflict by leaving both parts of
the conflict, rather than folding the 'NEXT' entry's list into the new
version's notes.

Fixes https://github.com/flutter/flutter/issues/85584

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/version_check_command.dart
+++ b/script/tool/lib/src/version_check_command.dart
@@ -155,11 +155,12 @@ class VersionCheckCommand extends PackageLoopingCommand {
 
     final List<String> errors = <String>[];
 
-    bool versionChanged = false;
+    bool versionChanged;
     final _CurrentVersionState versionState =
         await _getVersionState(package, pubspec: pubspec);
     switch (versionState) {
       case _CurrentVersionState.unchanged:
+        versionChanged = false;
         break;
       case _CurrentVersionState.validChange:
         versionChanged = true;
@@ -169,12 +170,13 @@ class VersionCheckCommand extends PackageLoopingCommand {
         errors.add('Disallowed version change.');
         break;
       case _CurrentVersionState.unknown:
+        versionChanged = false;
         errors.add('Unable to determine previous version.');
         break;
     }
 
-    if (!(await _hasConsistentVersion(package,
-        pubspec: pubspec, versionChanged: versionChanged))) {
+    if (!(await _validateChangelogVersion(package,
+        pubspec: pubspec, pubspecVersionChanged: versionChanged))) {
       errors.add('pubspec.yaml and CHANGELOG.md have different versions');
     }
 
@@ -302,12 +304,15 @@ ${indentation}HTTP response: ${pubVersionFinderResponse.httpResponse.body}
     return _CurrentVersionState.validChange;
   }
 
-  /// Returns whether or not the pubspec version and CHANGELOG version for
-  /// [plugin] match.
-  Future<bool> _hasConsistentVersion(
+  /// Checks whether or not [package]'s CHANGELOG's versioning is correct,
+  /// both that it matches [pubspec] and that NEXT is used correctly, printing
+  /// the results of its checks.
+  ///
+  /// Returns false if the CHANGELOG fails validation.
+  Future<bool> _validateChangelogVersion(
     Directory package, {
     required Pubspec pubspec,
-    required bool versionChanged,
+    required bool pubspecVersionChanged,
   }) async {
     // This method isn't called unless `version` is non-null.
     final Version fromPubspec = pubspec.version!;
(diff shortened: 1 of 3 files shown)


script/tool/lib/src/version_check_command.dart
```diff
@@ -140,11 +155,26 @@ class VersionCheckCommand extends PackageLoopingCommand {
...
     final List<String> errors = <String>[];
 
-    if (!await _hasValidVersionChange(package, pubspec: pubspec)) {
-      errors.add('Disallowed version change.');
+    bool versionChanged = false;
+    final _CurrentVersionState versionState =
+        await _getVersionState(package, pubspec: pubspec);
+    switch (versionState) {
+      case _CurrentVersionState.unchanged:
+        break;
```
gaaclarke, line 163: nit: I'd put versionChanged = false here since setting it above isn't clearly associated with the unchanged branch.

script/tool/lib/src/version_check_command.dart
```diff
@@ -268,16 +297,17 @@ ${indentation}HTTP response: ${pubVersionFinderResponse.httpResponse.body}
...
-    return true;
+    return _CurrentVersionState.validChange;
   }
 
   /// Returns whether or not the pubspec version and CHANGELOG version for
   /// [plugin] match.
   Future<bool> _hasConsistentVersion(
     Directory package, {
     required Pubspec pubspec,
+    required bool versionChanged,
```
gaaclarke, line 310: Can `versionChanged` be calculated inside of this function?  The first 2 arguments make sense to me, this one isn't clear.  It seems like you wouldn't ask "has consistent version?" if you already know the "version hasn't changed".

We should at least update the docstring.  Also, based on the name it seems like a pure function but it is printing out stuff.

script/tool/lib/src/version_check_command.dart
```diff
@@ -140,11 +155,26 @@ class VersionCheckCommand extends PackageLoopingCommand {
...
+        versionChanged = true;
+        errors.add('Disallowed version change.');
+        break;
+      case _CurrentVersionState.unknown:
+        errors.add('Unable to determine previous version.');
+        break;
     }
 
-    if (!(await _hasConsistentVersion(package, pubspec: pubspec))) {
+    if (!(await _hasConsistentVersion(package,
```
gaaclarke, line 176: Should we be calling hasConsistentVersion when _CurrentVersionState.unknown?

gaaclarke (commented):
Looks good other than some niggles about `_hasConsistentVersion`.  It's not a huge deal, let me know what you think.

script/tool/lib/src/version_check_command.dart
stuartmorgan: Changed to no initialization of the variable so that the compiler forces setting it in each branch, for better clarity/future-proofing.

script/tool/lib/src/version_check_command.dart
stuartmorgan: In general, my experience has been that unless it's expensive (which isn't the case here) it's much better to have CI checks fail but continue, so that we find more errors per run. It's really frustrating to fix a CI failure, push a PR, and wait for checks to run again, only to discover that there's another error now that could have been found before but wasn't. That's a pattern I've been following in most of the tool overhaul work, and it's already made things better.

In this case, errors in the CHANGELOG are almost entirely unrelated to whether the version actually changed (only the one edge case this test now covers); almost all of the time a failure here will just be due to a mismatch between what's in pubspec.yaml and what's in CHANGELOG.md, so surfacing that (so the PR author can fix it in parallel) even if something weird happens with the version check is useful.

script/tool/lib/src/version_check_command.dart
stuartmorgan: I changed the name and docstring, so hopefully it's clearer what this is doing now (it's unfortunately hard to do pure functions in these checks a lot of the time, because verbose logging of everything that's happening is a feature in CI checks; failures where you don't get easily and clearly enough detail about why are a pain). I also changed the parameter name to be more specific about what version changed (the pubspec version) since that was, as evidenced by your question, ambiguous.

For an example of why we'd call this even when that hasn't changed: if a PR author forgets to actually update the version in `pubspec.yaml` but remembers to update `CHANGELOG.md` as if they had, we want that to fail CI. In that case the parameter here will be false, but the version consistency check will fail.

gaaclarke (approved):
LGTM!

PR #4248, opened 2021-08-18
[flutter_plugin_tools] Fix build-examples for packages

stuartmorgan:
The build-examples command was filtering what it attempted to build by plugin platform, which means it never does anything for non-plugin packages. flutter/packages has steps that run this command, which suggests it used to work and regressed at some point, but nobody noticed; this will re-enable those builds so that we are getting CI coverage that the examples in flutter/packages build.

Mostly fixes https://github.com/flutter/flutter/issues/88435 (needs a flutter/packages tool pin roll to pick this up)

## Pre-launch Checklist

- [x] I signed the [CLA].


script/tool/lib/src/build_examples_command.dart
```diff
@@ -99,38 +99,60 @@ class BuildExamplesCommand extends PackageLoopingCommand {
...
+        ? requestedPlatforms
+            .where((_PlatformDetails platform) =>
+                pluginSupportsPlatform(platform.pluginPlatform, package))
+            .toSet()
+        : requestedPlatforms.toSet();
     if (buildPlatforms.isEmpty) {
       final String unsupported = requestedPlatforms.length == 1
           ? '${requestedPlatforms.first.label} is not supported'
-          : 'None of [${requestedPlatforms.map((_PlatformDetails p) => p.label).join(',')}] are supported';
+          : 'None of [${requestedPlatforms.map((_PlatformDetails p) => p.label).join(', ')}] are supported';
```
gaaclarke, line 119: nit: I ran into this too, it probably would be better to pull the join argument to a const or wrap that behavior in a function so it could be updated in one place.

script/tool/lib/src/build_examples_command.dart
```diff
@@ -99,38 +99,60 @@ class BuildExamplesCommand extends PackageLoopingCommand {
...
       for (final _PlatformDetails platform in buildPlatforms) {
+        // Repo policy is that a plugin must have examples configured for all
+        // supported platforms. For packages, just log and skip any requested
+        // platform that a package doesn't have set up.
+        if (!isPlugin &&
+            !example
+                .childDirectory(platform.flutterPlatformDirectory)
+                .existsSync()) {
+          print('Skipping ${platform.label} for $packageName; not supported.');
+          continue;
```
gaaclarke, line 151: nit: Instead of using `continue` do you think an `else` clause would be preferable?

script/tool/lib/src/common/plugin_utils.dart
```diff
@@ -16,6 +16,25 @@ enum PlatformSupport {
...
+        loadYaml(pubspecFile.readAsStringSync()) as YamlMap;
+    final YamlMap? flutterSection = pubspecYaml['flutter'] as YamlMap?;
+    if (flutterSection == null) {
+      return false;
+    }
+    final YamlMap? pluginSection = flutterSection['plugin'] as YamlMap?;
+    return pluginSection != null;
+  } on FileSystemException {
+    return false;
+  } on YamlException {
```
gaaclarke, line 33: Wouldn't it be better to use explicit checks instead of relying on exceptions?  It seems like a malformed yaml file would result in a `isFlutterPlugin == false` instead of crashing the tool which I would have expected.

gaaclarke (commented):
LGTM modulo the question about logic using exceptions in isFlutterPlugin

script/tool/lib/src/build_examples_command.dart
stuartmorgan: I pulled out a local function for the two lines that were doing the same display list transform.

script/tool/lib/src/build_examples_command.dart
stuartmorgan: I generally don't like to nest non-trivial primary logic in an else clause.

script/tool/lib/src/common/plugin_utils.dart
stuartmorgan: > Wouldn't it be better to use explicit checks instead of relying on exceptions?

Removed the `FileSystemException` in favor of an explicit check. (I'm not sure why the code was originally written that way; this was just a move refactor).

> It seems like a malformed yaml file would result in a `isFlutterPlugin == false` instead of crashing the tool which I would have expected.

In practice that wouldn't actually be a problem since we have a different CI test that validates the pubspecs, but I agree we shouldn't rely on it (and in fact, that's much more recent than this pubspec parsing code). I removed the exception handler for `YamlException`, and I've added a catch-all to `PackageLoopingCommand`'s `runForPackage`. That covers most of the code the tool runs at this point, and is a good idea to add in general since a central goal of that command base class is to always keep going, and then report all errors at the end.

script/tool/lib/src/build_examples_command.dart
gaaclarke: SGTM, not my preference but not in violation of our style.  Academics have argued about it for decades: https://en.wikipedia.org/wiki/Structured_programming#Early_exit

gaaclarke (approved):
LGTM!

PR #4249, opened 2021-08-19
[webview] Fix typos in README

wytesk133:


stuartmorgan:
Thanks for the submission!

In the future, please do not delete the checklist that is in the PR template; it is there for a reason. This PR is missing required elements described in the checklist (in particular, a version update and the corresponding changelog update), which needs to be addressed before it moves forward with review. If you have any questions about completing those steps that aren’t addressed documentation, please let me know.

wytesk133:
Updated :)

stuartmorgan:
Thanks, but that's only half of the change:

> in particular, a version update and the corresponding changelog update

Please see FAQ linked from the checklist item about updating the version.

PR #4251, opened 2021-08-20
[google_sign_in] adds option to re-authenticate on google silent sign in

ThetaSinner:
Adds an optional paramter to the signInSilently method which allows the application using the plugin to decide whether re-authentication should be attempted.

Fixes [#88556](https://github.com/flutter/flutter/issues/88556) 

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/packages/google_sign_in/google_sign_in/CHANGELOG.md
+++ b/packages/google_sign_in/google_sign_in/CHANGELOG.md
@@ -1,4 +1,6 @@
-## NEXT
+## 5.1.0
+
+* Add reAuthenticate option to signInSilently to allow re-authentication to be requested
 
 * Updated Android lint settings.
 
--- a/packages/google_sign_in/google_sign_in/lib/google_sign_in.dart
+++ b/packages/google_sign_in/google_sign_in/lib/google_sign_in.dart
@@ -314,11 +314,13 @@ class GoogleSignIn {
   /// successful sign in or `null` if there is no previously authenticated user.
   /// Use [signIn] method to trigger interactive sign in process.
   ///
-  /// Authentication process is triggered only if there is no currently signed in
+  /// Authentication is triggered if there is no currently signed in
   /// user (that is when `currentUser == null`), otherwise this method returns
   /// a Future which resolves to the same user instance.
   ///
-  /// Re-authentication can be triggered only after [signOut] or [disconnect].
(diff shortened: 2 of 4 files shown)


ditman:
Thanks @ThetaSinner! As I said, in web I'm not sure that calling signInSilently multiple times will cause the same effect as in android, but that's a gap that can be filled later. We just need to confirm it, so we can document the difference across platforms.

ThetaSinner:
Looks like you're right about the web behaviour, the existing login is [used ](https://github.com/flutter/plugins/blob/master/packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart#L126)to build the response to signInSilently.

The underlying [JavaScript library](https://github.com/google/google-api-javascript-client) does not discuss ID tokens (that I have seen!). The signIn [method](https://github.com/google/google-api-javascript-client/blob/e296cd6faf2adcb2c9ee2dfc3b05f1aae92ff731/docs/samples.md#authorizing-and-making-authorized-requests) it provides is interactive.

It doesn't look like working quietly  with ID tokens on web will work without a change to the google-api-javascript-client library. I don't need it so I'm happy with documented differences. Would you like me to find somewhere for that to go and add it to this PR?

ditman:
The web version is going to need a bunch of rewriting [soonish](https://github.com/flutter/flutter/issues/88084), that might change things. In the meantime, let's do this. I think the web version re-authenticates automatically, so as long as one always requests the current identity from the JS code, the token will always be valid (at least, that's my operating assumption :P)

ditman (approved):
LGTM! Let me find somebody from the iOS team to verify that this won't hurt there!

cyanglaz (approved):
LGTM on iOS

ditman:
Thanks for the super-late review @cyanglaz !!

Bharathh-Raj:
Hey @ThetaSinner, I wanted fresh idToken whenever i sign in. So I used this signInSilently method with reAuthenticate param as true. But I keep on getting the same idToken even after using this.

Bharathh-Raj:
So the OAuth API itself won't give us new idtoken unless it is about to/did expire? Is there any way to force the api to respond with new id token whenever I request? I'm using a service which have just 60 seconds window time to validate.

PR #4252, opened 2021-08-20
[flutter_plugin_tools] Introduce a class for packages

stuartmorgan:
Packages are the primary conceptual object in the tool, but currently they are represented simply as Directory (or occasionally a path string). This introduces an object for packages and:
- moves a number of existing utility methods into it
- sweeps the code for the obvious cases of using `Directory` to represent a package, especially in method signatures and migrates them
- notes a few places where we should migrate later, to avoid ballooning the size of the PR

There are no doubt other cases not caught in the sweep, but this gives us a foundation both for new code, and to migrate incrementally toward as we find existing code that was missed.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/common/package_looping_command.dart
+++ b/script/tool/lib/src/common/package_looping_command.dart
@@ -13,6 +13,7 @@ import 'package:platform/platform.dart';
 import 'core.dart';
 import 'plugin_command.dart';
 import 'process_runner.dart';
+import 'repository_package.dart';
 
 /// Possible outcomes of a command run for a package.
 enum RunState {
@@ -84,7 +85,7 @@ abstract class PackageLoopingCommand extends PluginCommand {
   int _otherWarningCount = 0;
 
   /// The package currently being run by [runForPackage].
-  PackageEnumerationEntry? _currentPackage;
+  PackageEnumerationEntry? _currentPackageEntry;
 
   /// Called during [run] before any calls to [runForPackage]. This provides an
   /// opportunity to fail early if the command can't be run (e.g., because the
@@ -97,7 +98,7 @@ abstract class PackageLoopingCommand extends PluginCommand {
   /// be included in the final error summary (e.g., a command that only has a
   /// single failure mode), or strings that should be listed for that package
   /// in the final summary. An empty list indicates success.
-  Future<PackageResult> runForPackage(Directory package);
+  Future<PackageResult> runForPackage(RepositoryPackage package);
 
   /// Called during [run] after all calls to [runForPackage]. This provides an
   /// opportunity to do any cleanup of run-level state.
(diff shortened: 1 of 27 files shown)


script/tool/lib/src/common/package_looping_command.dart
```diff
@@ -219,50 +202,50 @@ abstract class PackageLoopingCommand extends PluginCommand {
   Future<bool> _runInternal() async {
     _packagesWithWarnings.clear();
     _otherWarningCount = 0;
-    _currentPackage = null;
+    _currentPackageEntry = null;
 
     await initializeRun();
 
-    final List<PackageEnumerationEntry> packages = includeSubpackages
+    final List<PackageEnumerationEntry> packageEnumeration = includeSubpackages
```
gaaclarke, line 209: nit: s/packageEnumeration/targetPackages  I wouldn't call something an enumeration unless its purpose is to assign a number to items, plus targetPackages matches the verbiage of the function.

script/tool/lib/src/common/plugin_command.dart
```diff
@@ -14,15 +14,16 @@ import 'package:yaml/yaml.dart';
...
 /// An entry in package enumeration for APIs that need to include extra
 /// data about the entry.
 class PackageEnumerationEntry {
   /// Creates a new entry for the given package directory.
-  PackageEnumerationEntry(this.directory, {required this.excluded});
+  PackageEnumerationEntry(this.package, {required this.excluded});
 
-  /// The package's location.
-  final Directory directory;
+  /// The package.
```
gaaclarke, line 25: nit: not really a helpful comment.

script/tool/lib/src/common/pub_version_finder.dart
```diff
@@ -27,10 +27,10 @@ class PubVersionFinder {
 
   /// Get the package version on pub.
   Future<PubVersionFinderResponse> getPackageVersion(
-      {required String package}) async {
-    assert(package.isNotEmpty);
+      {required String packageName}) async {
+    assert(packageName.isNotEmpty);
```
gaaclarke, line 31: fyi: It would be nice if Dart could statically assert this.  In many places in my code I just traded null checks for isEmpty checks.

script/tool/lib/src/common/repository_package.dart
```diff
@@ -0,0 +1,80 @@
...
+// TODO(stuartmorgan): Add more package-related info here, such as an on-demand
+// cache of the parsed pubspec.
+class RepositoryPackage {
+  /// Creates a representation of the package at [directory].
+  ///
+  ///
+  RepositoryPackage(this.directory);
+
+  /// The location of the package.
+  Directory directory;
```
gaaclarke, line 21: Make this `final` please.

gaaclarke (commented):
Slick, this really cleans stuff up.  LGTM modulo I think RepositoryPackage should be immutable.

script/tool/lib/src/common/pub_version_finder.dart
stuartmorgan: > I just traded null checks for isEmpty checks

Well, not really; you traded code that was wrong without you noticing for code that is correct 🙂. Because if an empty string wasn't valid, but the check you had before was `foo != null` rather than `foo != null && !foo.isEmpty`, then the old code was wrong.

script/tool/lib/src/common/package_looping_command.dart
stuartmorgan: Done. (I actually made this change initially because you'd expressed concern about referring to a `PackageEnumerationEntry` as a package in the earlier review discussion; I like `targetPackages` better myself.)

As for enumeration, I was using it in what I think of as the common CS sense (Wikipedia: "An enumeration is a complete, ordered listing of all the items in a collection. The term is commonly used in mathematics and computer science to refer to a listing of all of the elements of a set.")

script/tool/lib/src/common/plugin_command.dart
stuartmorgan: I added a warning about `excluded` to this, as making sure that's not missed seemed like the only useful thing to say here. (I have to say something, because the repo has the lint requiring comments on public fields.

script/tool/lib/src/common/repository_package.dart
stuartmorgan: Oops, fixed.

stuartmorgan (commented):
> I think RepositoryPackage should be immutable.

I was going to do that when I first wrote it, but one of the things I want to add to this later (per the TODO) is a lazy cached accessor for the parsed pubspec, as that's something that's used in several places (and caching it would improve some APIs that are currently passing around the package and its parsed pubspec to avoid re-parsing). That doesn't work if it's immutable, I don't want to thrash on adding and then later removing all the `const`s at creation sites when removing immutability at that point.

gaaclarke:
> immutable

Yea makes sense.  When I saw those methods I thought memoizing them could be helpful, but ultimately I'm not sure how many times you call them.  Immutable minus whatever is needed for memoization sounds good to me.

gaaclarke (approved):
LGTM!

PR #4170, opened 2021-07-20
[camera] Expand CameraImage DTO with properties for lens aperture, exposure time and ISO.

BeMacized:
This PR adds additional properties to the CameraImage DTO:

- `lensAperture`
- `sensorExposureTime`
- `sensorSensitivity`

Relevant issue:
- flutter/flutter#73250

This PR adds no breaking changes. 

As this PR depends on changes made by the Android refactor (#4059), this PR will remain in draft state until those changes have been merged. 

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Note that unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I [updated pubspec.yaml](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates) with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [x] All existing and new tests are passing.

PR #4256, opened 2021-08-24
[camera] Expand CameraImage DTO with properties for lens aperture, exposure time and ISO. 

BeMacized:
(Resubmission of #4170)

This PR adds additional properties to the CameraImage DTO:

- `lensAperture`
- `sensorExposureTime`
- `sensorSensitivity`

Relevant issue:
- flutter/flutter#73250

This PR adds no breaking changes. 

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/CameraCaptureCallback.java
+++ b/packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/CameraCaptureCallback.java
@@ -76,15 +76,9 @@ class CameraCaptureCallback extends CaptureCallback {
     Float lensAperture = result.get(CaptureResult.LENS_APERTURE);
     Long sensorExposureTime = result.get(CaptureResult.SENSOR_EXPOSURE_TIME);
     Integer sensorSensitivity = result.get(CaptureResult.SENSOR_SENSITIVITY);
-    if (lensAperture != null) {
-      this.captureProps.setLastLensAperture(lensAperture);
-    }
-    if (sensorExposureTime != null) {
-      this.captureProps.setLastSensorExposureTime(sensorExposureTime);
-    }
-    if (sensorSensitivity != null) {
-      this.captureProps.setLastSensorSensitivity(sensorSensitivity);
-    }
+    this.captureProps.setLastLensAperture(lensAperture);
+    this.captureProps.setLastSensorExposureTime(sensorExposureTime);
+    this.captureProps.setLastSensorSensitivity(sensorSensitivity);
 
     if (cameraState != CameraState.STATE_PREVIEW) {
       Log.d(
--- a/packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/CameraCaptureCallback.java
+++ b/packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/CameraCaptureCallback.java
@@ -11,6 +11,7 @@ import android.hardware.camera2.CaptureResult;
 import android.hardware.camera2.TotalCaptureResult;
 import android.util.Log;
 import androidx.annotation.NonNull;
+import io.flutter.plugins.camera.types.CameraCaptureProperties;
 import io.flutter.plugins.camera.types.CaptureTimeoutsWrapper;
 
 /**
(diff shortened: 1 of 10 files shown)


packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/CameraCaptureCallback.java
```diff
@@ -67,6 +72,20 @@ private void process(CaptureResult result) {
     Integer aeState = result.get(CaptureResult.CONTROL_AE_STATE);
     Integer afState = result.get(CaptureResult.CONTROL_AF_STATE);
 
+    // Update capture properties
+    Float lensAperture = result.get(CaptureResult.LENS_APERTURE);
+    Long sensorExposureTime = result.get(CaptureResult.SENSOR_EXPOSURE_TIME);
+    Integer sensorSensitivity = result.get(CaptureResult.SENSOR_SENSITIVITY);
+    if (lensAperture != null) {
+      this.captureProps.setLastLensAperture(lensAperture);
+    }
```
mvanbeusekom, line 81: Quick check, is the `null` check on purpose? I can imagine that `null` indicates the value was not available for this particular capture result and you want to communicate this back to the Dart side. 

Don't you now have the risk that you communicate a (invalid) previous value with the current image?

(this question also goes for the `sensorExposureTime` and `sensorSensitivity` values)

packages/camera/camera/lib/src/camera_image.dart
```diff
@@ -125,4 +128,15 @@ class CameraImage {
...
+  /// The aperture settings for this image.
+  ///
+  /// Represented as an f-stop value.
+  final double lensAperture;
+
+  /// The sensor exposure time for this image in nanoseconds.
+  final int sensorExposureTime;
+
+  /// The sensor sensitivity in standard ISO arithmetic units.
+  final double sensorSensitivity;
```
mvanbeusekom, line 141: Shouldn't these be nullable?

From what I can see (I might be missing something) on Android these values can potentially be `null`. When the JAVA `CameraCaptureProperties` class is instantiated the initial values will all be `null`, now if the `CaptureResult` object also returns a `null` value the properties are not set which still results in a `null` value communicated to Flutter.

mvanbeusekom (commented):
I went through the code and noticed a potential `null` error which I wanted to double check.

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/CameraCaptureCallback.java
BeMacized: The way I had it in my head, is that in case you have some frames which do not supply these values, there's a very good chance the values will have been the same as the one from the frame(s) before it. Not sure if this is actually the case, but if you have a scenario where the camera implementation only reports back these values every X frames, there wouldn't be any issues.

I do agree that this might be confusing and potentially unwanted behavior though. I'm not sure either whether my concerns actually have any ground.

packages/camera/camera/lib/src/camera_image.dart
BeMacized: They should definitely be nullable yes, will make a fix.

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/CameraCaptureCallback.java
BeMacized: That would certainly be a simpler solution. I guess I'll just do that, if it turns out to give some trouble we can always add something like it back in later.

BeMacized: Looking at it now I just remembered why I implemented it this way in the first place:

These properties all have to be sent back in the handler given to [`imageStreamReader.setOnImageAvailableListener`](https://github.com/flutter/plugins/pull/4256/files#diff-97f0364e631a66ce678d82ff9906a3fa2f6fc75cfa6c59511d7d848f8c33544eR1049). 
However, I ran into the problem here that I cannot actually access the CaptureResult which contains these values here. 
Instead, the capture result is only available in `CameraCaptureCallback`, which is where I save these values. 

This is why I opted to cache these, as I don't really see any way of associating a specific image with a specific capture result.

mvanbeusekom (approved):
LGTM

PR #4264, opened 2021-08-25
Update README.md

KevinBeckers:
Documented the use of `shareddocuments` to open de iOS files app.


KevinBeckers:
@googlebot I signed it!

stuartmorgan:
Thanks for the submission, but the README is not intended to be a comprehensive list of schemes available on each platform, it just highlights a few common examples.

PR #4266, opened 2021-08-25
[flutter_plugin_tool] Fix CHANGELOG validation failure summary

stuartmorgan:
The error summary for a CHANGELOG validation failure was written when
the only thing being checked was that the versions matched, but now
there are other ways to fail as well (i.e., leaving NEXT). This fixes
the summary message to be more generic so that it doesn't mislead people
who hit validation failures.

While adding the test for this message, I discovered that almost all of
the tests were actually talking to pub.dev, causing their behavior to in
some cases depend on whether a package with that name happened to have
been published, and if so what its version was. In order to make the
tests hermetic and predictable, this fixes that by making all tests use
a mock HTTP client.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/version_check_command.dart
+++ b/script/tool/lib/src/version_check_command.dart
@@ -178,7 +178,7 @@ class VersionCheckCommand extends PackageLoopingCommand {
 
     if (!(await _validateChangelogVersion(package,
         pubspec: pubspec, pubspecVersionChanged: versionChanged))) {
-      errors.add('pubspec.yaml and CHANGELOG.md have different versions');
+      errors.add('CHANGELOG.md failed validation.');
     }
 
     return errors.isEmpty
--- a/script/tool/test/version_check_command_test.dart
+++ b/script/tool/test/version_check_command_test.dart
@@ -54,11 +54,15 @@ void main() {
     late List<List<String>> gitDirCommands;
     Map<String, String> gitShowResponses;
     late MockGitDir gitDir;
+    // Ignored if mockHttpResponse is set.
+    int mockHttpStatus;
+    Map<String, dynamic>? mockHttpResponse;
 
     setUp(() {
       fileSystem = MemoryFileSystem();
       mockPlatform = MockPlatform();
       packagesDir = createPackagesDirectory(fileSystem: fileSystem);
+
       gitDirCommands = <List<String>>[];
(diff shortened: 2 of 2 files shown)


script/tool/lib/src/version_check_command.dart
```diff
@@ -178,7 +178,7 @@ class VersionCheckCommand extends PackageLoopingCommand {
 
     if (!(await _validateChangelogVersion(package,
         pubspec: pubspec, pubspecVersionChanged: versionChanged))) {
-      errors.add('pubspec.yaml and CHANGELOG.md have different versions');
+      errors.add('CHANGELOG.md failed validation.');
```
cyanglaz, line 181: Do we have a place to explain what are the failing scenarios? If not, maybe we should list them here so it's easier for debugging. (Especially when none-team-member contributor hits this error)

script/tool/test/version_check_command_test.dart
```diff
@@ -54,11 +54,15 @@ void main() {
...
     late MockGitDir gitDir;
+    // Ignored if mockHttpResponse is set.
+    int mockHttpStatus;
+    Map<String, dynamic>? mockHttpResponse;
 
     setUp(() {
       fileSystem = MemoryFileSystem();
       mockPlatform = MockPlatform();
       packagesDir = createPackagesDirectory(fileSystem: fileSystem);
+
```
cyanglaz, line 65: nits: extra line

cyanglaz (approved):
LGTM % nits

script/tool/lib/src/version_check_command.dart
stuartmorgan: The specific error is output inline, this is just the line that goes in the final summary. The summary block always says to see the full log for details. See https://cirrus-ci.com/task/4541043946291200 for an example (that led to this PR).

script/tool/test/version_check_command_test.dart
stuartmorgan: I added this on purpose; the next section is a big block of setup for the mock GitDir and having it separated out slightly makes it easier to see the high-level flow vs it being a wall of text.

(I'll likely replace the git setup here with something based on processRunner like I did with the new `publish` tests, which would help, but didn't want to get into that in this PR.)

PR #4277, opened 2021-08-26
[video_player] upgraded pigeon to 1.0

gaaclarke:
change to `video_player_platform_interface` that must merge first: https://github.com/flutter/plugins/pull/4291

## Pre-launch Checklist

- [ ] I signed the [CLA].


gaaclarke:
There is one issue with the Android generated code:
```
/Users/aaclarke/dev/plugins/packages/video_player/video_player/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java:146: error: incompatible types: Map<Object,Object> cannot be converted to Map<String,String>
      Map<String, String> httpHeaders = arg.getHttpHeaders();
                                                          ^
1 error
```

Looks like the template parameters didn't get brought over, the generated code has that method returning `Map<Object, Object>`.

packages/video_player/video_player/example/ios/Flutter/AppFrameworkInfo.plist
```diff
@@ -25,6 +25,6 @@
     <string>arm64</string>
   </array>
   <key>MinimumOSVersion</key>
-  <string>8.0</string>
+  <string>9.0</string>
```
stuartmorgan, line 28: Please don't change this.

packages/video_player/video_player/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -274,6 +274,7 @@
 				TargetAttributes = {
 					97C146ED1CF9000F007C117D = {
 						CreatedOnToolsVersion = 7.3.1;
+						DevelopmentTeam = S8QB4VV633;
```
stuartmorgan, line 277: None of the changes in this file should be checked in either.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
```diff
@@ -511,7 +511,7 @@ - (FLTTextureMessage*)onPlayerSetup:(FLTVideoPlayer*)player
   return result;
 }
 
-- (void)initialize:(FlutterError* __autoreleasing*)error {
+- (void)initialize:(FlutterError* _Nullable __autoreleasing*)error {
```
stuartmorgan, line 514: Use `nullable` in ObjC method signatures.

stuartmorgan: Also, why is this method named incorrectly? This isn't Pigeon-generated presumably.

stuartmorgan (changes requested):
> Am I able to land this update atomically or am I going to have to separate the video_player_platform_interface changes into a separate PR that I land first?

The latter; there is no such thing as an atomic change to multiple packages because they have to be published individually. We also have a repo policy against publishing any untested configuration, and it's impossible to have the dependent packages go through CI as they would be published until the dependencies are published.

gaaclarke:
@stuartmorgan This is a WIP.  I'll ping you when it's ready.  I'm waiting for 1.0 to land before cleaning it up for review.

gaaclarke:
I've tested the example app locally on android now too and it works.

packages/video_player/video_player/example/ios/Flutter/AppFrameworkInfo.plist
gaaclarke: done

packages/video_player/video_player/example/ios/Runner.xcodeproj/project.pbxproj
gaaclarke: done

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
gaaclarke: I think you have to use _Nullable since we are passing in a handle to a FlutterError.  It's not named incorrectly, I've used `@ObjcSelector` to make all the selectors backwards compatible with how the code previously worked.

stuartmorgan:
This will need to be updated to move the platform channel into the implementation, per [the new platform channel plan](https://docs.google.com/document/d/1BSrDA_FjLU-3T0H774RocJ4IYyc0ytMl937rAimGJdg/edit?usp=sharing). That's probably best done a precursor PR to move in a copy of the existing implementation, then updating this PR to update that version.

PR #3264, opened 2020-11-11
[image_picker] add forceFullMetadata for iOS

cpboyd:
## Description

Originally #3090

Make iOS 11+ permissions requests optional per https://developer.apple.com/forums/thread/653414.
PHAssets aren't actually required and there was already some fallback code in place.

To keep this as a non-breaking change, I added an optional parameter of `forceFullMetaData` with a default of `true`.
Setting this to false will disable any permissions checking on iOS and simply display the image picker.

I still need to update the tests.

## Related Issues

Fixes https://github.com/flutter/flutter/issues/65995

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors (See [Contributor Guide]).
- [x] All existing and new tests are passing.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] The analyzer (`flutter analyze`) does not report any problems on my PR.
- [x] I read and followed the [Flutter Style Guide].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. [shared_preferences]
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I signed the [CLA].
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate a breaking change in CHANGELOG.md and increment major revision).
- [x] No, this is *not* a breaking change.

[issue database]: https://github.com/flutter/flutter/issues
[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning
[CLA]: https://cla.developers.google.com/

PR #4288, opened 2021-08-27
[image_picker] add forceFullMetadata to interface

cpboyd:
Splitting interface portion from #3264 per https://github.com/flutter/plugins/pull/3264#issuecomment-907328067

## Description

Make iOS 11+ permissions requests optional per https://developer.apple.com/forums/thread/653414.
PHAssets aren't actually required and there was already some fallback code in place.

To keep this as a non-breaking change, I added an optional parameter of `forceFullMetadata` with a default of `true`.
Setting this to false will disable any permissions checking on iOS and simply display the image picker.

## Checklist

- [x] I signed the [CLA].

## Breaking Change

### DIFF
--- a/packages/image_picker/image_picker_platform_interface/CHANGELOG.md
+++ b/packages/image_picker/image_picker_platform_interface/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 2.4.0
+
+* Add `forceFullMetadata` option to `pickImage`.
+  * To keep this non-breaking `forceFullMetadata` defaults to `true`, so the plugin tries
+   to get the full image metadata which may require extra permission requests on certain platforms.
+  * If `forceFullMetadata` is set to `false`, the plugin fetches the image in a way that reduces
+   permission requests from the platform (e.g on iOS the plugin won’t ask for the `NSPhotoLibraryUsageDescription` permission).
+
 ## 2.3.0
 
 * Updated `LostDataResponse` to include a `files` property, in case more than one file was recovered.
--- a/packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
+++ b/packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
@@ -58,6 +58,11 @@ abstract class ImagePickerPlatform extends PlatformInterface {
   /// image types such as JPEG. If compression is not supported for the image that is picked,
   /// a warning message will be logged.
   ///
+  /// `forceFullMetadata` defaults to `true`, so the plugin tries to get the full image metadata which may require
+  /// extra permission requests on certain platforms.
+  /// If `forceFullMetadata` is set to `false`, the plugin fetches the image in a way that reduces permission requests
+  /// from the platform (e.g. on iOS the plugin won’t ask for the `NSPhotoLibraryUsageDescription` permission).
+  ///
   /// Use `preferredCameraDevice` to specify the camera to use when the `source` is [ImageSource.camera].
   /// The `preferredCameraDevice` is ignored when `source` is [ImageSource.gallery]. It is also ignored if the chosen camera is not supported on the device.
   /// Defaults to [CameraDevice.rear]. Note that Android has no documented parameter for an intent to specify if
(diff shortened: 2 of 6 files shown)


packages/image_picker/image_picker_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 2.3.1
```
cyanglaz, line 1: should be 2.4.0 as there are non-breaking changes to the public API, same thing in pubspec.yaml

cyanglaz (changes requested):

packages/image_picker/image_picker_platform_interface/CHANGELOG.md
cpboyd: I've updated this as well as the version & changelog for web

cyanglaz: I think the web change needs to be in the other PR

cpboyd: Since web directly `extends ImagePickerPlatform`, it's required to match the function signatures for `ImagePickerPlatform`

As a result, the automated web tests fail if it's not also changed to match. Granted, there's no changes to web beyond adding `bool forceFullMetadata = true,` to the function signatures.

cyanglaz: hmm. This is weird. I would think it is the other way around. Since the platform_interface change hasn't been published, the web shouldn't pick up the newest API, thus the override should report a compile error.

cyanglaz: @cpboyd Do you mind double checking again. The test for web should not fail because the web package is depending on an old version of platform_interface, where the method signature hasn't been changed.

cpboyd: @cyanglaz I just pushed a clean commit without the web changes

cyanglaz (approved):
LGTM!

packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
```diff
@@ -73,6 +78,7 @@ abstract class ImagePickerPlatform extends PlatformInterface {
     double? maxWidth,
     double? maxHeight,
     int? imageQuality,
+    bool forceFullMetadata = true,
```
stuartmorgan, line 81: This is a breaking change. This needs to be reverted, as this version is broken (and we need to figure out how to make sure this kind of thing doesn't pass CI in the future).

packages/image_picker/image_picker_platform_interface/CHANGELOG.md
stuartmorgan: > Since web directly `extends ImagePickerPlatform`, it's required to match the function signatures for `ImagePickerPlatform`
> 
> As a result, the automated web tests fail if it's not also changed to match. Granted, there's no changes to web beyond adding `bool forceFullMetadata = true,` to the function signatures.

This is a description of why this is a breaking change, so I'm not clear on why this was landed. You cannot add parameters to methods to an interface without breaking all implementations.

You also cannot just ignore that something is a breaking change by also pushing a change in another package. I don't understand how this was approved with *unpublished* code changes to a separate package.

stuartmorgan: > @cyanglaz I just pushed a clean commit without the web changes

Part of the problem here is that this did not actually happen :(

stuartmorgan:
Now that the fire is hopefully out due to the revert: please see https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-platform-interface-method-parameters for the right way to re-implement this.

cyanglaz:
My fault, I completely forgot how platform_interface breaking change works. We should bump the version to 3.0.0 for this change then adopt the new version (3.0.0) in the mobile and web packages.
@cpboyd

stuartmorgan:
See the my link above; I would strongly prefer we not do breaking changes every time we need to add a param to a method.

PR #4289, opened 2021-08-28
Reland "[ci] update wait-on-check version and set verbose to false"

cyanglaz:
Relands https://github.com/flutter/plugins/pull/4262 as the white list has been updated to allow the newer wait-on-check version.

Fixes https://github.com/flutter/flutter/issues/85127#issuecomment-907557132


stuartmorgan (approved):
LGTM

stuartmorgan:
Landing this to fix the tree; it looks like this wasn't *added* to the whitelist, but instead *replaced* the existing whitelist entry, so now it's broken without this.

PR #4292, opened 2021-08-31
[tool] Add a way to opt a file out of formatting

Hixie:
Per discussion on Discord, this adds a way to opt-out of the formatting.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/CHANGELOG.md
+++ b/script/tool/CHANGELOG.md
@@ -1,6 +1,8 @@
-## NEXT
+## 0.7.0
 
 - `native-test` now supports `--linux` for unit tests.
+- Formatting now skips Dart files that contain a line that exactly
+  matches the string `// This file is hand-formatted.`.
 
 ## 0.6.0+1
 
--- a/script/tool/lib/src/format_command.dart
+++ b/script/tool/lib/src/format_command.dart
@@ -206,7 +206,27 @@ class FormatCommand extends PluginCommand {
 
     final String fromPath = relativeTo.path;
 
+    // Dart files are allowed to have a pragma to disable auto-formatting. This
+    // was added because Hixie hurts when dealing with what dartfmt does to
+    // artisanally-formatted Dart, while Stuart gets really frustrated when
+    // dealing with PRs from newer contributors who don't know how to make Dart
+    // readable. After much discussion, it was decided that files in the plugins
+    // and packages repos that really benefit from hand-formatting (e.g. files
+    // with large blobs of hex literals) could be opted-out of the requirement
+    // that they be autoformatted, so long as the code's owner was willing to
+    // bear the cost of this during code reviews.
+    // In the event that code ownership moves to someone who does not hold the
+    // same views as the original owner, the pragma can be removed and the file
+    // auto-formatted.
+    const String handFormattedExtension = '.dart';
+    const String handFormattedPragma = '// This file is hand-formatted.';
+
     return files
+        .where((File file) {
+          // See comment above near [handFormattedPragma].
+          return path.extension(file.path) != handFormattedExtension ||
+              !file.readAsLinesSync().contains(handFormattedPragma);
+        })
         .map((File file) => path.relative(file.path, from: fromPath))
         .where((String path) =>
             // Ignore files in build/ directories (e.g., headers of frameworks)
--- a/script/tool/test/format_command_test.dart
+++ b/script/tool/test/format_command_test.dart
@@ -8,6 +8,7 @@ import 'package:args/command_runner.dart';
 import 'package:file/file.dart';
 import 'package:file/memory.dart';
 import 'package:flutter_plugin_tools/src/common/core.dart';
+import 'package:flutter_plugin_tools/src/common/file_utils.dart';
 import 'package:flutter_plugin_tools/src/format_command.dart';
 import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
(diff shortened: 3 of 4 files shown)


Hixie:
I made this apply to all files rather than just Dart, mostly because it was easier to plug into the code where `File` objects were still around. I considered pushing the File-to-String conversion lower down the codepaths, but that ended up being pretty invasive and not obviously better. I also considered just having the Dart code go back from String to Files but that ended up being pretty ugly too.

script/tool/lib/src/format_command.dart
```diff
@@ -206,7 +206,21 @@ class FormatCommand extends PluginCommand {
 
     final String fromPath = relativeTo.path;
 
+    const String handFormattedPragma = '// This file is hand-formatted.';
+
     return files
+        .where((File file) {
```
stuartmorgan, line 212: > I made this apply to all files rather than just Dart, mostly because it was easier to plug into the code where `File` objects were still around.

I would much rather this apply only to Dart, as this is strictly worse than `clang-format off` for `clang-format`-formatted files because the latter will prevent people with editors that are configured to auto-format from accidentally formatting the file when editing it. (I really, really wish `dartfmt` would support this for the same reason.)

What's the issue with just adding a `file.basename.endsWith('.dart')` check inside this lambda?

script/tool/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
   `--no-push-flags`. Releases now always tag and push.
 - **Breaking change**: `publish`'s `--package` flag has been replaced with the
   `--packages` flag used by most other packages.
+- Formatting now skips files that contain a line that exactly
+  matches the string `// This file is hand-formatted`.
```
stuartmorgan, line 14: The period needs to be inside the \`s; it's part of the string being searched for.

script/tool/lib/src/format_command.dart
```diff
@@ -206,7 +206,21 @@ class FormatCommand extends PluginCommand {
 
     final String fromPath = relativeTo.path;
 
+    const String handFormattedPragma = '// This file is hand-formatted.';
```
stuartmorgan, line 209: I think it's worth adding a short explanatory comment here about why this exists to save someone finding the PR to understand the reason.

script/tool/test/format_command_test.dart
```diff
@@ -85,12 +86,21 @@ void main() {
...
+      extraFiles: <String>[
+        ...files,
+        fileWithOptOut,
+      ],
     );
 
+    final p.Context posixContext = p.posix;
+    childFileWithSubcomponents(pluginDir, posixContext.split(fileWithOptOut))
+        .writeAsStringSync(
+            '// copyright bla bla\n// This file is hand-formatted.\ncode...');
```
stuartmorgan, line 102: This is a distinct enough thing from the basic "does the formatter get called" that I think it would be much better to have a separate "skips files with the opt-out" test, so that a test failure would be more specific.

stuartmorgan (changes requested):

script/tool/lib/src/format_command.dart
```diff
@@ -206,7 +206,21 @@ class FormatCommand extends PluginCommand {
...
     return files
+        .where((File file) {
+          try {
+            return !file.readAsLinesSync().contains(handFormattedPragma);
+          } on FileSystemException {
+            // If we can't open the file or can't read it as UTF-8, then we just
+            // consider it included. It should get filtered out later when we
+            // deal with file extensions. (We want the formatters to see the
+            // files that we can't read, in case they have opinions about such
+            // things as character encodings, etc.)
```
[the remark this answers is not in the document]
Hixie: please update the style guide to mention this

script/tool/lib/src/format_command.dart
Hixie: it seemed weird to have this code know about file extensions too, but i'm happy to add it if you like

script/tool/lib/src/format_command.dart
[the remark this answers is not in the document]
Hixie: since i added the check for the extension, i was able to remove all this error handling logic and thus the comment is gone too

Hixie:
PTAL

script/tool/lib/src/format_command.dart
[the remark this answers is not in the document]
stuartmorgan: Do you still want me to add this to the style guide? I'm happy to do so, but it's also a position that not everyone agrees with.

script/tool/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
   `--packages` flag used by most other packages.
 - **Breaking change** Passing both `--run-on-changed-packages` and `--packages`
   is now an error; previously it the former would be ignored.
+- Formatting now skips files that contain a line that exactly
+  matches the string `// This file is hand-formatted.`.
```
stuartmorgan, line 16: This got merged into the wrong section, it needs to be in a new one since 0.6.0 is published.

You should actually do a version bump in this PR since in order to use it in flutter/packages it'll need to be published (and then the flutter/packages pin rolled to pick it up)

script/tool/lib/src/format_command.dart
```diff
@@ -206,7 +206,27 @@ class FormatCommand extends PluginCommand {
 
     final String fromPath = relativeTo.path;
 
+    // Dart files are allowed to have a pragma to disable auto-formatting. This
+    // was added because Hixie hurts when dealing with what dartfmt does to
+    // artisanally-formatted Dart, while Stuart gets really frustrated when
+    // dealing with PRs from newer contributors who don't know how to make Dart
+    // readable. After much discussion, it was decided that files in the plugins
```
stuartmorgan, line 213: 🤣 Well, I did ask.

(I was just thinking something like "Allow package maintainers to opt out of Dart autoformatting in order to use Flutter's manual formatting style instead.", but this certainly works.)

stuartmorgan (approved):
LGTM modulo adding a version change.

script/tool/CHANGELOG.md
stuartmorgan: Oh, and it should be changed to "skips Dart files".

script/tool/lib/src/format_command.dart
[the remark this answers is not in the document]
Hixie: I don't feel strongly about it (but then I get a guilty pleasure in writing sentences that use the same pronoun to mean different people), but if there's an argument to be made that it makes things clearer, we should document it somewhere people can see it.

script/tool/CHANGELOG.md
Hixie: Done.

PR #4294, opened 2021-09-01
[flutter_plugin_tools] Add Linux support to native-test

stuartmorgan:
- Adds a minimal unit test to url_launcher_linux as a proof of concept. This uses almost exactly the same CMake structure as the Windows version that was added recently.
- Adds Linux support for unit tests to `native-test`, sharing almost all of the existing Windows codepath.
- Fixes the fact that it it was running the debug version of the unit tests, but `build-examples` only builds release. (On other platforms we run debug unit tests, but on those platforms the test command internally builds the requested unit tests, so the mismatch doesn't matter.)
- Enables the new test in CI.

Also opportunistically fixes some documentation in `native_test_command.dart` that wasn't updated as more platform support was added.

Linux portion of https://github.com/flutter/flutter/issues/82445

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/url_launcher/url_launcher_linux/linux/url_launcher_plugin_private.h
+++ b/packages/url_launcher/url_launcher_linux/linux/url_launcher_plugin_private.h
@@ -1,0 +1,13 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "include/url_launcher_linux/url_launcher_plugin.h"
+
+#include <flutter_linux/flutter_linux.h>
+
+// Handles the canLaunch method call.
+//
+// Temporarily exposed for testing due to
+// https://github.com/flutter/flutter/issues/88724
+FlMethodResponse* can_launch(FlUrlLauncherPlugin* self, FlValue* args);
--- a/packages/url_launcher/url_launcher_linux/linux/url_launcher_plugin_private.h
+++ b/packages/url_launcher/url_launcher_linux/linux/url_launcher_plugin_private.h
@@ -2,9 +2,9 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include "include/url_launcher_linux/url_launcher_plugin.h"
-
 #include <flutter_linux/flutter_linux.h>
+
+#include "include/url_launcher_linux/url_launcher_plugin.h"
 
 // Handles the canLaunch method call.
 //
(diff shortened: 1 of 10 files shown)


stuartmorgan:
Hrm, I'll need to debug why it's not finding the tests when running in CI tomorrow; it worked for me locally.

stuartmorgan:
> Hrm, I'll need to debug why it's not finding the tests when running in CI tomorrow; it worked for me locally.

Should be fixed now; `build-examples` builds release, and `native-test` was only running debug, so the combination used by CI didn't actually work. PR description updated with the added change.

packages/url_launcher/url_launcher_linux/linux/url_launcher_plugin_private.h
```diff
@@ -0,0 +1,13 @@
...
+// found in the LICENSE file.
+
+#include <flutter_linux/flutter_linux.h>
+
+#include "include/url_launcher_linux/url_launcher_plugin.h"
+
+// Handles the canLaunch method call.
+//
+// Temporarily exposed for testing due to
+// https://github.com/flutter/flutter/issues/88724
```
cbracken, line 12: Nit consider phrasing as a TODO to aid in tech-debt greppability.

script/tool/lib/src/native_test_command.dart
```diff
@@ -442,10 +468,11 @@ this command.
           .whereType<File>()
           .where(isTestBinary)
           .where((File file) {
-        // Only run the debug build of the unit tests, to avoid running the
-        // same tests multiple times.
+        // Only run the release build of the unit tests, to avoid running the
+        // same tests multiple times. Release is used rather than debug since
+        // `build-examples` builds release versions.
```
cbracken, line 473: Was going to say perhaps we should be changing to generate a debug build of build-examples given that it's typically more convenient to debug debug builds. That said, in the end you want to verify the release build of the library code in the plugin, and people shouldn't be relying on `#ifdef`'ed code in their tests, so this seems fine.

If this ever gets problematic, we could add a buildmode flag.

? Debugging debug builds of unit tests is typically more convenient, though

cbracken (approved):
lgtm

script/tool/lib/src/native_test_command.dart
stuartmorgan: I just realized that there's actually a more annoying problem, which is that the current behavior isn't going to be good for local use. Filed https://github.com/flutter/flutter/issues/89303.

I'm going to do that as a follow-up since it applies to the already-landed Windows flow in the same way, but I'll do it shortly.

packages/url_launcher/url_launcher_linux/linux/url_launcher_plugin_private.h
stuartmorgan: Good point, done.

PR #4295, opened 2021-09-01
[video_player]fix video play() not update VideoPlayerValue when isInitialized is false

jerryzhoujw:
Before  `VideoPlayerController.initialize()` finished. 
Client maybe call `VideoPlayerController.play()`.
It's better to notify the `VideoPlayerValue` to client.

To make this work:

```dart
  Future<void> play() async {
    if (value.position == value.duration) {
      await seekTo(const Duration());              // <--- so seeTo() need to avoid exception before initialize() finished. 
    }
    value = value.copyWith(isPlaying: true);  // <--- this value better to update and notify. so 
    await _applyPlayPause();
  }
```

The exception on my android devices is:

      PlatformException (PlatformException(NullPointerException, java.lang.NullPointerException: Attempt to invoke virtual method 'void io.flutter.plugins.videoplayer.VideoPlayer.seekTo(int)' on a null object reference, null, null))

https://github.com/flutter/flutter/issues/89259

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -499,7 +499,12 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     } else if (position < const Duration()) {
       position = const Duration();
     }
-    await _videoPlayerPlatform.seekTo(_textureId, position);
+
+    try {
+      await _videoPlayerPlatform.seekTo(_textureId, position);
+    } catch (e) {
+      //
+    }
     _updatePosition(position);
   }
 


packages/video_player/video_player/lib/video_player.dart
```diff
@@ -499,7 +499,12 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     } else if (position < const Duration()) {
       position = const Duration();
     }
-    await _videoPlayerPlatform.seekTo(_textureId, position);
+
+    try {
+      await _videoPlayerPlatform.seekTo(_textureId, position);
+    } catch (e) {
```
stuartmorgan, line 505: Silently discarding all exceptions is not a fix, it's hiding the problem—and a variety of other possible problems as well, which is not okay.

`seekTo` isn't doing the same up-front `isInitialized` check that other methods are doing, which seems like the actual bug here. It looks like this flow throwing is a regression from https://github.com/flutter/plugins/pull/3727 (@cc @KyleFin)

stuartmorgan (changes requested):
Thanks for the submission! I've left an initial review comment, but this will also need the missing items in the checklist addressed before moving forward.

packages/video_player/video_player/lib/video_player.dart
KyleFin: Thanks @stuartmorgan! Yes, this issue was mentioned in https://github.com/flutter/plugins/pull/3727#issuecomment-876879494. @jerryzhoujw let me know if there's anything I can do to help. Thanks!

KyleFin: Instead of this change I believe you could just change line 397 to be
`if (value.position == value.duration && value.isInitialized) {`

KyleFin: And unit tests could be similar to those added in https://github.com/flutter/plugins/pull/3727/files#diff-b67d4b0ee52dfea4c9c71082e12eca5459a174693c8d40577b0191d8ee498e34

jerryzhoujw: Thanks

jerryzhoujw: close this since https://github.com/flutter/plugins/pull/4300 much better.

jerryzhoujw:
close this since https://github.com/flutter/plugins/pull/4300 much better.

PR #4296, opened 2021-09-01
when fast lock screen and unlock can crash

xfans:
camera 0.9.2+1
flutter 2.0.6

In Camera.java startBackgroundThread  and stopBackgroundThread 
In dart didChangeAppLifecycleState inactive and resumed
not matching call


xfans:
@googlebot I signed it!

PR #4298, opened 2021-09-02
[google_maps_flutter_web] Fix getScreenCoordinate, Circle zIndex

ditman:
This PR passes the `zIndex` attribute to `Circle` geometry objects.

Fixes https://github.com/flutter/flutter/issues/89374 (simple)

This PR **ensures that the `getScreenCoordinate` method works as expected** on the web platform.

Fixes https://github.com/flutter/flutter/issues/80710

The `getScreenCoordinate(LatLng)` method (along with its inverse `getLatLng(ScreenCoordinate)` require a fully rendered map to be tested, because they need the physical size in pixels, current center and zoom level of a map to work.

So far, these two methods weren't tested, but **this PR adds happy-case tests for both (see `projection_test.dart`)**

In order to test those methods, the `Controller` of the rendered map must be available on the test, so we can retrieve information from the map. While doing this, I noticed that the map Controller was being sent to the programmer before it was truly ready (while the map was still initializing).

Instead of littering the test with imprecise timeouts that may make these tests slower (or flakier) than needed, **this PR also changes the initialization process of a GMap** slightly so when its Controller is returned to the user (`onPlatformViewCreated` method call), it is truly ready.

For this, **we now call `init` on the Controller immediately as it is created, then wait for the first `onTilesloaded` event coming from the JS SDK, and only then we return the Controller to the user.**

This change happens within "private" sections of the plugin, so programmers using the plugin "normally" shouldn't notice any difference whatsoever (only that the GMap might load slightly faster, and the `onPlatformViewCreated` callback might be firing a few hundred milliseconds later).

_(This PR also triggered changes in the mock files, that can be safely ignored, since those are auto-generated.)_

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
```diff
@@ -305,11 +308,16 @@ class GoogleMapsPlugin extends GoogleMapsFlutterPlatform {
       polylines: polylines,
       circles: circles,
       mapOptions: mapOptions,
-    );
+    )..init(); // Initialize the controller
```
ditman, line 311: This is the crux of the early initialization. Before the flow was:

buildView -> onPlatformViewCreated (plugin) -> init (plugin) -> init (controller) -> JS SDK

but now, it's:

buildView + init (controller) -> JS SDK -> onPlatformViewCreated (plugin) -> init (plugin, noop)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
```diff
@@ -35,7 +35,10 @@ class GoogleMapsPlugin extends GoogleMapsFlutterPlatform {
 
   @override
   Future<void> init(int mapId) async {
-    _map(mapId).init();
+    // The internal instance of our controller is initialized eagerly in `buildView`,
+    // so we don't have to do anything in this method, which is left intentionally
+    // blank.
+    assert(_map(mapId) != null, 'Must call buildWidget before init!');
```
ditman, line 41: This assertion is added to preserve some of the validation that happened before, mainly that calling init before buildWidget would cause quite a lot of breakage.

This method used to call `init` in the controller, but no longer. We do that immediately after creating the controller in buildView.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
```diff
@@ -163,6 +185,10 @@ class GoogleMapController {
 
   // Funnels map gmap events into the plugin's stream controller.
   void _attachMapEvents(gmaps.GMap map) {
+    map.onTilesloaded.first.then((event) {
```
ditman, line 188: There's [no `map.onReady` event](https://developers.google.com/maps/documentation/javascript/reference/map#Map-Events), so we synthesize it from the first `onTilesloaded` event 🤷 (_[years of discussion](https://stackoverflow.com/questions/832692/how-can-i-check-whether-google-maps-is-fully-loaded)_)

packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
```diff
@@ -13,8 +13,10 @@ dependencies:
     sdk: flutter
 
 dev_dependencies:
-  build_runner: ^1.11.0
-  google_maps: ^5.1.0
+  build_runner: ^2.1.1
+  google_maps: ^5.2.0
+  google_maps_flutter: # Used for projection_test.dart
+    path: ../../google_maps_flutter
```
stuartmorgan, line 19: This feels sketchy, but I don't have a better suggestion offhand :(

(Maybe long term some of the widgets should be in the interface package, not the app-facing package?)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -441,3 +442,33 @@ gmaps.LatLng _pixelToLatLng(gmaps.GMap map, int x, int y) {
 
   return projection.fromPointToLatLng!(point)!;
 }
+
+// original JS by: Krasimir (https://krasimirtsonev.com/blog/article/google-maps-api-v3-convert-latlng-object-to-actual-pixels-point-object)
```
stuartmorgan, line 446: Have you talked to OSS licensing folks about this (and the similarly commented code above)?

packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
ditman: > (Maybe long term some of the widgets should be in the interface package, not the app-facing package?)

Yes, I'm not a fan of the current widget. There's a circular dependency in the code that didn't allow me to completely separate it from the GoogleMapsController that also lives in the core plugin, [here](https://github.com/flutter/plugins/blob/master/packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart#L27) (I'm not a fan of the controller accessing the map state directly, but never had the time to make it better :/).

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: Let me send them an email, thanks for the reminder.

ditman: Resolved properly, this is now added as a `third_party` import.

script/tool/lib/src/license_check_command.dart
```diff
@@ -49,12 +49,17 @@ const Set<String> _ignoredFullBasenameList = <String>{
...
-// Third-party code used in url_launcher_web.
+  // Third-party code used in url_launcher_web.
   RegExp(
       r'^// Copyright 2017 Workiva Inc\..*'
       r'^// Licensed under the Apache License, Version 2\.0',
       multiLine: true,
       dotAll: true),
+  // Third-party code used in google_maps_flutter_web.
+  RegExp(
+    r'^// The MIT License [^C]+ Copyright \(c\) 2008 Krasimir Tsonev',
```
ditman, line 60: needs dart format

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -264,16 +264,17 @@ gmaps.MarkerOptions _markerOptionsFromMarker(
...
     ..fillColor = _getCssColor(circle.fillColor)
     ..fillOpacity = _getCssOpacity(circle.fillColor)
     ..center = gmaps.LatLng(circle.center.latitude, circle.center.longitude)
     ..radius = circle.radius
-    ..visible = circle.visible;
-  return populationOptions;
+    ..visible = circle.visible
+    ..zIndex = circle.zIndex;
+  return circleOptions;
 }
```
stuartmorgan, line 278: Out of curiosity, where aren't there unit tests for this file? Currently the unit test dir just point to integration tests, but a utility file like this one should be trivially unit testable in a normal Dart test context, no?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
```diff
@@ -305,11 +308,16 @@ class GoogleMapsPlugin extends GoogleMapsFlutterPlatform {
...
-    );
+    )..init(); // Initialize the controller
 
     _mapById[creationId] = mapController;
 
-    onPlatformViewCreated.call(creationId);
+    mapController.events.whereType<WebMapReadyEvent>().first.then((event) {
+      assert(creationId == event.mapId,
+          'Received WebMapReadyEvent for the wrong map');
+      // *now* we can notify the plugin that there's a fully working controller.
```
stuartmorgan, line 318: go/avoidwe

"Notify the plugin now that there's a fully working controller."

packages/google_maps_flutter/google_maps_flutter_web/lib/src/third_party/to_screen_location/to_screen_location.dart
```diff
@@ -0,0 +1,57 @@
+// The MIT License (MIT)
+//
```
stuartmorgan, line 2: I don't think these first two lines should be here; I've never seen an actual MIT license file that was labelled here like this.

See also https://opensource.org/licenses/MIT

stuartmorgan (approved):
A comment nit and license format nit, but otherwise LGTM

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: We can't have VM unit tests on this file because 99% of it is converting between "Plugin Entities" <-> "`gmaps` Entities" (for the JS Interop, which uses `dart:html` extensively).

There's a few minor functions that could be extracted to VM tests, but anything that touches the `gmaps` namespace is tainted.

We could still write a convert_test.dart that unit-tested each of those methods in better isolation (but still as a chromedriver integration test), I just never got the time to dive that deep (most of these methods have a single entrypoint, and *that* is going to be have tests for sure, however, I wish I could prove this with some coverage metrics :P).

stuartmorgan: > We can't have VM unit tests on this file because 99% of it is converting between "Plugin Entities" <-> "`gmaps` Entities" (for the JS Interop, which uses `dart:html` extensively).

Ah, right. I should have remembered that those objects were not from this plugin.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/third_party/to_screen_location/to_screen_location.dart
ditman: D'oh, copied verbatim from here: https://krasimirtsonev.com/license, didn't notice that "The MIT License (MIT)" was decoration and not part of the license itself. I'll fix.

ditman: There's division about this. Left the title in, respecting the original author's license text.

ditman:
I'll manually squash and merge when master recovers, I want to have a proper commit message for this one.

PR #4305, opened 2021-09-02
[plugin_tools] build-examples .pluginToolsConfig.yaml support

Hixie:
## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/CHANGELOG.md
+++ b/script/tool/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.1
+
+- Add support for `.pluginToolsConfig.yaml` in the `build-examples` command.
+
 ## 0.7.0
 
 - `native-test` now supports `--linux` for unit tests.
--- a/script/tool/lib/src/build_examples_command.dart
+++ b/script/tool/lib/src/build_examples_command.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 
 import 'package:file/file.dart';
 import 'package:platform/platform.dart';
+import 'package:yaml/yaml.dart';
 
 import 'common/core.dart';
 import 'common/package_looping_command.dart';
@@ -16,7 +17,19 @@ import 'common/repository_package.dart';
 /// Key for APK.
 const String _platformFlagApk = 'apk';
 
+const String _pluginToolsConfigFileName = '.pluginToolsConfig.yaml';
+const String _pluginToolsConfigBuildFlagsKey = 'buildFlags';
+const String _pluginToolsConfigGlobalKey = 'global';
+
+const String _pluginToolsConfigExample = '''
+$_pluginToolsConfigBuildFlagsKey:
+  $_pluginToolsConfigGlobalKey:
+    - "--no-tree-shake-icons"
+    - "--dart-define=buildmode=testing"
+''';
+
 const int _exitNoPlatformFlags = 3;
+const int _exitInvalidPluginToolsConfig = 4;
 
 // Flutter build types. These are the values passed to `flutter build <foo>`.
 const String _flutterBuildTypeAndroid = 'apk';
@@ -99,7 +112,13 @@ class BuildExamplesCommand extends PackageLoopingCommand {
   @override
   final String description =
       'Builds all example apps (IPA for iOS and APK for Android).\n\n'
-      'This command requires "flutter" to be in your path.';
+      'This command requires "flutter" to be in your path.\n\n'
+      'A $_pluginToolsConfigFileName file can be placed in an example app '
+      'directory to specify additional build arguments. It should be a YAML '
+      'file with a top-level map containing a single key '
+      '"$_pluginToolsConfigBuildFlagsKey" containing a map containing a '
+      'single key "$_pluginToolsConfigGlobalKey" containing a list of build '
+      'arguments.';
 
   @override
   Future<void> initializeRun() async {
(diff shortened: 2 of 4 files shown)


script/tool/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.8.0
```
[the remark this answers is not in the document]
Hixie: I mean, it's a breaking change (if someone already has a file with the `pluginToolsConfig.yaml` name in a directory, then it'll almost certainly fail now)...

script/tool/lib/src/build_examples_command.dart
```diff
@@ -16,7 +17,10 @@ import 'common/repository_package.dart';
 /// Key for APK.
 const String _platformFlagApk = 'apk';
 
+const String _pluginToolsConfigFileName = '.pluginToolsConfig.yaml';
```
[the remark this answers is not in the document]
Hixie: I don't think we can know what format to pick to avoid a format break, but ok. (For example, what if we discover that actually what we want is for this to be a list of buildFlags, so that you can try running with several configurations? So the root should be a list rather than a map? Or, maybe the root should be platforms, with buildFlags as the second level. Or...)

Hixie:
Ok, updated as requested.

script/tool/lib/src/build_examples_command.dart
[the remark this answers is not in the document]
stuartmorgan: But when your root is a map you can *always* change things without it being a hard breaking change, because you can just use a different root key for the new format. Then eventually you can remove support for the old key.

(In this case the cost of the breaking change would be pretty low since we only guarantee that the tool works for two repos that we control; maybe I'm over-engineering based on experiences with things where it's much harder when you can't do the incremental change via a map.)

script/tool/lib/src/build_examples_command.dart
```diff
@@ -225,6 +244,58 @@ class BuildExamplesCommand extends PackageLoopingCommand {
...
+            printError(_pluginToolsConfigExample);
+            throw ToolExit(_exitInvalidPluginToolsConfig);
+          }
+          extraBuildFlags = <String>[
+            ...extraBuildFlags,
+            ...globalBuildFlagsConfiguration.cast<String>()
+          ];
+        }
+      }
+    }
```
stuartmorgan, line 297: Optional nit: with all the error checking this is long enough that I would extract it to a `List<String> _getPackageSpecificBuildFlags(RepositoryPackage package)` directory.

script/tool/lib/src/build_examples_command.dart
```diff
@@ -225,6 +244,58 @@ class BuildExamplesCommand extends PackageLoopingCommand {
...
+          final Object? globalBuildFlagsConfiguration =
+              buildFlagsConfiguration[_pluginToolsConfigGlobalKey];
+          if (globalBuildFlagsConfiguration is! YamlList) {
+            printError(
+                'The $_pluginToolsConfigFileName file\'s "$_pluginToolsConfigBuildFlagsKey" key must be a map');
+            printError('whose "$_pluginToolsConfigGlobalKey" key is a list.');
+            printError(
+                'That list must contain a list of arguments to pass to the flutter tool.');
+            printError(
+                'For example, the $_pluginToolsConfigFileName file could look like:');
```
stuartmorgan, line 287: Optional nit: are the slight variations of error messages actually buying us anything here over just having one constant (like the example) for any invalid structure, that describes the correct structure? It doesn't seem like someone is going to have a hard time figuring out which part they've done wrong from a catch-all message and example.

script/tool/test/build_examples_command_test.dart
```diff
@@ -379,7 +379,6 @@ void main() {
         ]),
       );
 
-      print(processRunner.recordedCalls);
```
stuartmorgan, line 382: Thanks, I thought I'd caught all of these during all the refactoring/cleanup work. Apparently it spread to more places than I'd noticed.

stuartmorgan (approved):
LGTM; all comments optional.

script/tool/lib/src/build_examples_command.dart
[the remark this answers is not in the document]
Hixie: i think things are a lot less clear-cut in practice than this (e.g. changing from a list to a map is non-breaking too since you can just keep on supporting the old list mode, and sometimes a map's keyspace is complete, e.g. the semantics are that you do something for each key, in which case the old key would conflict with the new keyspace, etc), but it's not a big deal either way.

script/tool/lib/src/build_examples_command.dart
Hixie: ok, extracted to function

script/tool/lib/src/build_examples_command.dart
Hixie: in my experience, more detail in error messages (if it's accurate) is always more helpful. :-)

script/tool/test/build_examples_command_test.dart
Hixie: there's still some prints somewhere in the tests but i couldn't track them down in the 10 seconds i spent trying. It says "Determine diff with base sha:" sometimes, and "null" once.

stuartmorgan: > "Determine diff with base sha:" sometimes

I bet that's from a test of the helper directly, not going through a command runner that captures output. That helper really shouldn't log anyway, I just haven't gotten around to fixing it.

PR #4306, opened 2021-09-02
[ci.yaml] Add builders to recipes cq

CaseyHillers:
This enables testing flutter/recipes changes to these builders

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -39,6 +39,7 @@ targets:
     recipe: plugins/plugins
     timeout: 30
     properties:
+      add_recipes_cq: "true"
       target_file: windows_build_and_platform_tests.yaml
       dependencies: >
         [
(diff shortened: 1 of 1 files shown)


.ci.yaml
```diff
@@ -39,6 +39,7 @@ targets:
     recipe: plugins/plugins
     timeout: 30
     properties:
+      add_recipes_cq: "true"
```
stuartmorgan, line 42: We want the string true, not the value `true`?

stuartmorgan (approved):
RSLGTM (with one question), since this key isn't documented at https://github.com/flutter/cocoon/blob/master/CI_YAML.md so I'm not entirely sure what it does. (Could it be added there?)

.ci.yaml
CaseyHillers: Thanks for pointing this out! I'll add it to the docs.

This is a limitation of yaml files backed by proto (we define properties to be Map<String, String> as there's no Map<String, dynamic> option (granted we could do a JsonString of sorts)

PR #4311, opened 2021-09-03
Bring up new Windows test targets

stuartmorgan:
Uses the new generic plugins recipe to add a number of new Windows tests
that have recently gained tool support, but not yet added to the LUCI
builds:
- New targets:
  - The "build all plugins" test (master and stable)
  - UWP build tests (master only, since UWP is not on stable yet)
  - Tool tests (master only with existing tool tests on Linux)
  - Replacement versions of the existing builders but with names that
    match the Cirrus naming to make the parallels in testing easier
    to understand in the GitHub UI and configs.
- Modification of existing targets:
  - Adds Windows native unit tests to the existing script.

All but the modification change are in bringup mode, so won't affect the repo until a follow-up
removes that flag.

Part of https://github.com/flutter/flutter/issues/86113
Part of https://github.com/flutter/flutter/issues/82729

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -62,7 +62,7 @@ targets:
         ]
     scheduler: luci
 
-  - name: windows-build+platform-tests master
+  - name: Windows win32_build+platform-tests master
     recipe: plugins/plugins
     bringup: true
     timeout: 30
@@ -76,7 +76,7 @@ targets:
         ]
     scheduler: luci
 
-  - name: windows-build+platform-tests stable
+  - name: Windows win32_build+platform-tests stable
     recipe: plugins/plugins
     bringup: true
     timeout: 30
@@ -91,7 +91,7 @@ targets:
         ]
     scheduler: luci
 
-  - name: build_all_plugins_windows master
+  - name: Windows build_all_plugins master
     recipe: plugins/plugins
     bringup: true
     timeout: 30
@@ -105,7 +105,7 @@ targets:
         ]
     scheduler: luci
 
-  - name: build_all_plugins_windows stable
+  - name: Windows build_all_plugins stable
     recipe: plugins/plugins
     bringup: true
     timeout: 30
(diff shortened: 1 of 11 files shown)


.ci.yaml
```diff
@@ -60,6 +62,88 @@ targets:
         ]
     scheduler: luci
 
+  - name: windows-build+platform-tests master
```
CaseyHillers, line 65: target names must follow the format $platform $name

So `Windows build+platform-tests master`. I'll send a PR to update this in the CI_YAML.md

.ci.yaml
```diff
@@ -60,6 +62,88 @@ targets:
...
+    recipe: plugins/plugins
+    bringup: true
+    timeout: 30
+    properties:
+      # TODO(stuartmorgan): Uncomment when removing bringup.
+      #add_recipes_cq: "true"
```
CaseyHillers, line 71: Is this necessary? My expectation is this could land in this PR

.ci/targets/windows_build_and_platform_tests.yaml
```diff
@@ -1,8 +1,9 @@
-
 tasks:
   - name: "prepare tool"
     script: .ci/scripts/prepare_tool.sh
-  - name: "build examples"
+  - name: "build examples (Win32)"
```
CaseyHillers, line 4: I don't think these target yaml files need strings around the name

.ci.yaml
stuartmorgan: Probably not for this particular one since it's a copy-paste, but in general I would rather make sure a new target actually works in the wild before adding it to the recipe CQ since I really don't want to bring the recipes repo to a standstill by throwing untested things into its pipeline for landing changes.

.ci/targets/windows_build_and_platform_tests.yaml
stuartmorgan: Removed.

.ci.yaml
stuartmorgan: Fixed.

.ci.yaml
CaseyHillers: This is a really good insight! I'll have to add it to the documentation (and later as a test)

CaseyHillers (approved):

PR #4314, opened 2021-09-05
Revert "[image_picker] add forceFullMetadata to interface"

zanderso:
Reverts flutter/plugins#4288

FYI @cpboyd 

This commit appears to be breaking the framework tree somehow: https://ci.chromium.org/ui/p/flutter/builders/prod/Linux%20flutter_plugins/3301/overview

```
Running command: "dart analyze --fatal-infos" in /b/s/w/ir/x/t/flutter_plugins.DYUYDK/packages/image_picker/image_picker_for_web
Analyzing image_picker_for_web...

  error - lib/image_picker_for_web.dart:51:22 - 'ImagePickerPlugin.pickImage' ('Future<PickedFile> Function({int? imageQuality, double? maxHeight, double? maxWidth, CameraDevice preferredCameraDevice, required ImageSource source})') isn't a valid override of 'ImagePickerPlatform.pickImage' ('Future<PickedFile?> Function({bool forceFullMetadata, int? imageQuality, double? maxHeight, double? maxWidth, CameraDevice preferredCameraDevice, required ImageSource source})'). - invalid_override
  error - lib/image_picker_for_web.dart:113:17 - 'ImagePickerPlugin.getImage' ('Future<XFile> Function({int? imageQuality, double? maxHeight, double? maxWidth, CameraDevice preferredCameraDevice, required ImageSource source})') isn't a valid override of 'ImagePickerPlatform.getImage' ('Future<XFile?> Function({bool forceFullMetadata, int? imageQuality, double? maxHeight, double? maxWidth, CameraDevice preferredCameraDevice, required ImageSource source})'). - invalid_override

2 issues found.
```

Fixes https://github.com/flutter/flutter/issues/89517

### DIFF
--- a/packages/image_picker/image_picker_platform_interface/pubspec.yaml
+++ b/packages/image_picker/image_picker_platform_interface/pubspec.yaml
@@ -4,7 +4,7 @@ repository: https://github.com/flutter/plugins/tree/master/packages/image_picker
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+image_picker%22
 # NOTE: We strongly prefer non-breaking changes, even at the expense of a
 # less-clean API. See https://flutter.dev/go/platform-interface-breaking-changes
-version: 2.3.0
+version: 2.4.1
 
 environment:
   sdk: ">=2.12.0 <3.0.0"
--- a/packages/image_picker/image_picker_platform_interface/pubspec.yaml
+++ b/packages/image_picker/image_picker_platform_interface/pubspec.yaml
@@ -4,7 +4,7 @@ repository: https://github.com/flutter/plugins/tree/master/packages/image_picker
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+image_picker%22
 # NOTE: We strongly prefer non-breaking changes, even at the expense of a
 # less-clean API. See https://flutter.dev/go/platform-interface-breaking-changes
-version: 2.4.0
+version: 2.3.0
 
 environment:
   sdk: ">=2.12.0 <3.0.0"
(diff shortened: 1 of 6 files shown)


packages/image_picker/image_picker_platform_interface/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ repository: https://github.com/flutter/plugins/tree/master/packages/image_picker
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+image_picker%22
 # NOTE: We strongly prefer non-breaking changes, even at the expense of a
 # less-clean API. See https://flutter.dev/go/platform-interface-breaking-changes
-version: 2.4.0
+version: 2.3.0
```
stuartmorgan, line 7: You can't revert the version, because 2.3.0 and 2.4.0 are already published. This should be 2.4.1.

stuartmorgan:
> This commit appears to be breaking the framework tree somehow

It published a breaking change as if it weren't breaking.

stuartmorgan:
The reasons things are red is because it's using the published, broken version of the interface package to analyze web. I think we'll need to force-land this and break the tree temporarily, then publish on red, to fix it. This is quite a mess.

I will take a look later.

packages/image_picker/image_picker_platform_interface/pubspec.yaml
zanderso: Done.

stuartmorgan:
I think I may be able to fix this forward in a way that won't require bypassing all of our testing safeties; let me see if that will work.

stuartmorgan:
> I think I may be able to fix this forward in a way that won't require bypassing all of our testing safeties; let me see if that will work.

No, I forgot that Dart doesn't allow C-style function overloading. That makes what I had in mind impossible.

I think I just have to push this through a bunch of red. I'll manually check all the failures here and make sure they are exactly what I expect from this situation.

stuartmorgan:
`publishable` was a legitimate failure, which I've fixed. The rest is in the clients of the interface package as expected for this scenario. I'll let publishable re-run to completion, then force this through if that's green.

stuartmorgan (approved):
LGTM. I will force land this, which will in turn turn the tree red. I'll then manually publish `image_picker_platform_interface`, which should fix everything (including flutter/flutter) and then re-run the post-submit tests to green the tree.

(Actually, if I publish fast enough I might be able to beat at least some of the post-submit tests...)

PR #2224, opened 2019-10-19
[video_player] Add onComplete callback

smkhalsa:
## Description

Adds an onComplete callback to the video_player plugin.

PR #4318, opened 2021-09-07
[video_player] adding onComplete callback

omartinma:
PR to add onComplete callback when the video has finished playing. Following same changes were applied in this [PR](https://github.com/flutter/plugins/pull/2224) but it was never merged.

Fixes https://github.com/flutter/flutter/issues/21929

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,525 +1,527 @@
+## 2.1.16
+
+- Add on complete callback
+
 ## 2.1.15
 
-* Ensured seekTo isn't called before video player is initialized. Fixes [#89259](https://github.com/flutter/flutter/issues/89259).
```
stuartmorgan, line 3: Please do not rewrite the entire CHANGELOG format.

stuartmorgan:
Thanks for the contribution!

> [x] I listed at least one issue that this PR fixes in the description above.

Could you please list the issue that explains the use case for this API addition? You've checked the box here, but there's no issue link in the PR.

packages/video_player/video_player/CHANGELOG.md
omartinma: I just revert the change

omartinma:
> Thanks for the contribution!
> 
> > [x] I listed at least one issue that this PR fixes in the description above.
> 
> Could you please list the issue that explains the use case for this API addition? You've checked the box here, but there's no issue link in the PR.

I linked in the description to a previous PR, I thought was enough. I will unmark this field then!

stuartmorgan:
> > Thanks for the contribution!
> > > [x] I listed at least one issue that this PR fixes in the description above.
> > 
> > 
> > Could you please list the issue that explains the use case for this API addition? You've checked the box here, but there's no issue link in the PR.
> 
> I linked in the description to a previous PR, I thought was enough. I will unmark this field then!

The previous PR doesn't link to an issue either, or provide the context I was requesting.

omartinma:
> > > Thanks for the contribution!
> > > > [x] I listed at least one issue that this PR fixes in the description above.
> > > 
> > > 
> > > Could you please list the issue that explains the use case for this API addition? You've checked the box here, but there's no issue link in the PR.
> > 
> > 
> > I linked in the description to a previous PR, I thought was enough. I will unmark this field then!
> 
> The previous PR doesn't link to an issue either, or provide the context I was requesting.

What should I do then?

stuartmorgan:
Please file an issue (per [the contributing guide](https://github.com/flutter/flutter/wiki/Tree-hygiene#overview)) describing the use case for the new API you want to add, and link to it from the PR description.

omartinma:
> Please file an issue (per [the contributing guide](https://github.com/flutter/flutter/wiki/Tree-hygiene#overview)) describing the use case for the new API you want to add, and link to it from the PR description.

I just created the [issue](https://github.com/flutter/flutter/issues/90030) @stuartmorgan . Hope is ok with that, thanks for helping!

stuartmorgan (changes requested):
Based on the discussion in the two related issues, it seems like it's currently possible to determine this via the existing value listener, but it's not as easy as it could be.

I think a better approach would be to look at adding more information to the value to make it easier to determine that (e.g., maybe a "transition type" that could indicate special cases like started, ended, etc.), rather than adding a one-off, completely separate way of listening to one particular event.

ditman:
I agree with @stuartmorgan in this one, the `onComplete` looks a little bit forced into the API. It is a shame that the VideoPlayerController doesn't expose a `Stream<VideoPlayerEvent>`, similar to the one listened to here:

https://github.com/flutter/plugins/blob/master/packages/video_player/video_player/lib/video_player.dart#L388-L389

Everything would be quite a bit simpler IMO with a Stream rather than with a ValueListenable + a complex value object (but I understand this is my personal taste, and probably not an absolute truth)

omartinma:
Thanks both for you feedback! @stuartmorgan @ditman 
I like also the idea to expose `Stream<VideoPlayerEvent>` so it will give more flexibility, for instance maybe you want to know when it starts, paused, etc.
Should i update my PR or should we create new issue and create a new PR for that?

stuartmorgan:
I think the next step would be to propose one or more new events in https://github.com/flutter/flutter/issues/21929 that would cover the use case, and we can discuss further there.

I'm going to close this pending a resolution of a discussion there. Thanks!

PR #4319, opened 2021-09-07
[camera] Fix IllegalStateException being thrown in Android implementation when switching activities.

BeMacized:
This PR removes the lifecycle hooks from the Android implementation. I have a feeling these were added by accident during the recent refactor, as the example app already [puts this responsibility on the user of the plugin](https://github.com/flutter/plugins/blob/master/packages/camera/camera/example/lib/main.dart#L116). 
This is also described in the readme: https://github.com/flutter/plugins/tree/master/packages/camera/camera#handling-lifecycle-states

This should fix the issue with the IllegalStateException being thrown when switching to a different activity, for example when using `camera` in combination with `image_picker`, as can be seen in the issue linked below:

Relevant issues:
 - Fixes flutter/flutter#89352
 - Fixes flutter/flutter#89815
 - Fixes flutter/flutter#89431

## Pre-launch Checklist

- [x] I signed the [CLA].
  - No new public members to document, or existing ones to update.

### DIFF
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.9.4+1
+
+* Fixed Android implementation throwing IllegalStateException when switching to a different activity.
+
 ## 0.9.4
 
 * Add web support by endorsing `package:camera_web`.
--- a/packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
+++ b/packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
@@ -574,15 +574,17 @@ class Camera
 
   /** Starts a background thread and its {@link Handler}. */
   public void startBackgroundThread() {
-    if (backgroundHandlerThread == null) {
-      backgroundHandlerThread = HandlerThreadFactory.create("CameraBackground");
-      try {
-        backgroundHandlerThread.start();
-      } catch (IllegalThreadStateException e) {
-        // Ignore exception in case the thread has already started.
-      }
-      backgroundHandler = HandlerFactory.create(backgroundHandlerThread.getLooper());
+    if (backgroundHandlerThread != null) {
+      return;
     }
+
+    backgroundHandlerThread = HandlerThreadFactory.create("CameraBackground");
+    try {
+      backgroundHandlerThread.start();
+    } catch (IllegalThreadStateException e) {
+      // Ignore exception in case the thread has already started.
+    }
+    backgroundHandler = HandlerFactory.create(backgroundHandlerThread.getLooper());
   }
 
   /** Stops the background thread and its {@link Handler}. */
--- a/packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
+++ b/packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
@@ -35,9 +35,6 @@ import android.view.Display;
 import android.view.Surface;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-import androidx.lifecycle.Lifecycle;
-import androidx.lifecycle.LifecycleObserver;
-import androidx.lifecycle.OnLifecycleEvent;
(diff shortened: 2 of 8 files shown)


packages/camera/camera/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.9.3+1
+
+* Fixed Android implementation throwing IllegalStateException when switching to a different activity.
```
AlexV525, line 3:
```suggestion
* Fixed Android implementation throwing IllegalStateException when switching to a different activity.

```

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -576,19 +573,19 @@ private Display getDefaultDisplay() {
...
 
   /** Starts a background thread and its {@link Handler}. */
-  @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
   public void startBackgroundThread() {
-    backgroundHandlerThread = new HandlerThread("CameraBackground");
-    try {
-      backgroundHandlerThread.start();
-    } catch (IllegalThreadStateException e) {
-      // Ignore exception in case the thread has already started.
+    if (backgroundHandlerThread == null) {
```
AlexV525, line 577: Could we add a lifecycle test to avoid re-involved with the same issue?

mvanbeusekom: There is a unit-test which tests for the `if` statement. However I am not sure how we could add a test to verify that we don't add implementation of listening to lifecycle events in the future. 

Do you have some insights on how this could be accomplished?

AlexV525: Add an observer `WidgetsBindingObserver` in a Flutter App, then override the `didChangeAppLifecycleState` might be a solution.

stuartmorgan: Capturing from a separate discussion, verifying that the core classes (`Camera`, `CameraPlugin`) are not `instanceof LifecycleObserver` seems like a possible approach, even if it's not foolproof (e.g., a composed object could implement it)

BeMacized:
@googlebot I consent.

mvanbeusekom:
@googlebot I consent.

stuartmorgan:
@blasten @bparrishMines Could one of you review this? It fixes a P2 regression.

packages/camera/camera/android/src/main/java/io/flutter/plugins/camera/Camera.java
blasten: alternatively, it could just return earlier.

blasten:
```java
if (backgroundHandlerThread  != null) {
  return;
}
```

packages/camera/camera/android/src/test/java/io/flutter/plugins/camera/CameraTest.java
```diff
@@ -773,6 +800,22 @@ public void resumePreview_shouldSendErrorEventOnCameraAccessException()
...
+    verify(mockHandlerThread, times(1)).start();
+    assertEquals(mockHandler, TestUtils.getPrivateField(camera, "backgroundHandler"));
+  }
+
+  @Test
+  public void startBackgroundThread_shouldNotStartNewThreadWhenAlreadyCreated() {
+    camera.startBackgroundThread();
+    camera.startBackgroundThread();
+
+    verify(mockHandlerThread, times(1)).start();
```
blasten, line 816: verify(mockHandlerThread, times(0)).start();  first. before you call `camera.startBackgroundThread();`.  It verifies the state before `startBackgroundThread()` is called.

packages/camera/camera/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+
```
blasten, line 1: no need for this line

blasten (approved):
lgtm

PR #4324, opened 2021-09-08
[video_player] Added new field allowBackgroundPlayback to enable background playing of video

IlyaMax:
*Description*
This PR introduces a new option to VideoPlayerOptions, which enables user to decide on whether user wants to keep playing video in the background or not.
This is a much requested feature, in many issues.

*Related issues*
63836 - flutter/flutter#63836
63735 - flutter/flutter#63735
55673 - flutter/flutter#55673
48008 - flutter/flutter#48008
40722 - flutter/flutter#40722
32571 - flutter/flutter#32571

## Pre-launch Checklist

- [x] I signed the [CLA].


IlyaMax:
@googlebot I signed it!

stuartmorgan:

ZhangZhiH:
@IlyaMax 
hello
All I see is an observeAppLifecycle
But I don't see background playback implemented there
How to play video in background

IlyaMax:
@ZhangZhiH audio playback is active when you close application. if you mean picture in picture, it is not related.

ZhangZhiH:
> @ZhangZhiH audio playback is active when you close application or lock screen. if you mean picture in picture, it is not related.

This problem now finds a solution
Use audio player when switching lock screen
Change back to the video player when cutting back to the page and synchronize the progress

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -350,6 +350,11 @@ class DurationRange {
 
 /// [VideoPlayerOptions] can be optionally used to set additional player settings
 class VideoPlayerOptions {
+  /// Set this to true to let plugin manage play/pause with app lifecycle changes,
+  /// Set this to false to keep playing video in background, when app goes in background
+  /// The default value is true
+  final bool observeAppLifecycle;
```
stuartmorgan, line 356: I think it would make more sense to call this `pauseInBackground` or similar; observing the app lifecycle is how it's implemented, not the actual user intent of the option.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -737,6 +737,14 @@ void main() {
     await controller.initialize();
     expect(controller.videoPlayerOptions!.mixWithOthers, true);
   });
+
+  test('observeAppLifecycle', () async {
+    final VideoPlayerController controller = VideoPlayerController.file(
+        File(''),
+        videoPlayerOptions: VideoPlayerOptions(observeAppLifecycle: true));
+    await controller.initialize();
+    expect(controller.videoPlayerOptions!.observeAppLifecycle, true);
```
stuartmorgan, line 746: This is testing that the options struct isn't modified when it's passed it, but that's all; if your changes to `initialize()` that honor this option were reverted the test would still pass. This needs a real test that the option is being honored.

stuartmorgan (changes requested):
Overall this change seems reasonable, it'll just need more meaningful testing.

ZhangZhiH:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
IlyaMax: maybe `needBackgroundPlayback`?

stuartmorgan: I'm not sure about "need", but `allowBackgroundPlayback` sounds good to me.

packages/video_player/video_player/test/video_player_test.dart
IlyaMax: You mean I need to validate `_VideoAppLifeCycleObserver` doesn't work  if flag is specified?

stuartmorgan: The purpose of a test is to verify that a change works. If I can remove the key part of a change and the test would still pass (which is true here), then the test isn't actually testing the feature.

I haven't tried this, but I believe you could call `WidgetsBinding.instance!.handleAppLifecycleStateChanged(AppLifecycleState.paused)` in the test and assert that it doesn't pause. You should similarly add a test that does that with the option set the other way and ensure that it does.

blessedman:
It looks like background playback will be supported soon

blessedman:
any update?

IlyaMax:
@blessedman I need some time to fix these comments. I hope I have time today. But if not I will definitely complete this next week

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
IlyaMax: fixed

packages/video_player/video_player/test/video_player_test.dart
IlyaMax: fixed

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.3.0
+
+* Add `allowBackgroundPlayback` to `VideoPlayerOptions`
```
stuartmorgan, line 3: Nit: Adds

(Per new repo CHANGELOG style guide)

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -95,6 +95,22 @@ class _FakeClosedCaptionFile extends ClosedCaptionFile {
 }
 
 void main() {
+  void verifyAppLifeCycle(
+    VideoPlayerController controller, {
+    bool isObserving = true,
```
stuartmorgan, line 100: Please make this required instead of having a default value; it's better for tests to be as explicit as possible.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -189,6 +205,14 @@ void main() {
...
     group('initialize', () {
+      test('started app lifecycle observing', () async {
+        final VideoPlayerController controller = VideoPlayerController.network(
+          'https://127.0.0.1',
+        );
+        await controller.initialize();
+        await controller.play();
+        verifyAppLifeCycle(controller);
+      });
       test('asset', () async {
```
stuartmorgan, line 216: Nit: blank line between tests please.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -775,6 +799,32 @@ void main() {
...
+  group('VideoPlayerOptions', () {
+    test('setMixWithOthers', () async {
+      final VideoPlayerController controller = VideoPlayerController.file(
+          File(''),
+          videoPlayerOptions: VideoPlayerOptions(mixWithOthers: true));
+      await controller.initialize();
+      expect(controller.videoPlayerOptions!.mixWithOthers, true);
+    });
+
+    group('allowBackgroundPlayback', () {
```
stuartmorgan, line 811: Excessive nesting of groups makes IDE UIs harder to use; this can be removed, and the test name below updated to include this string in the name.

packages/video_player/video_player_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.1.0
+
+* Add `allowBackgroundPlayback` to `VideoPlayerOptions`
```
stuartmorgan, line 3: Nit: Adds
Nit: Missing period at end of sentence

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -329,6 +329,11 @@ class DurationRange {
 
 /// [VideoPlayerOptions] can be optionally used to set additional player settings
 class VideoPlayerOptions {
+  /// Set this to false to let plugin manage play/pause with app lifecycle changes,
```
stuartmorgan, line 332: Remove this sentence; you don't generally need to document both options for a bool since they are inverse of each other.

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -329,6 +329,11 @@ class DurationRange {
...
+  /// Set this to true to keep playing video in background, when app goes in background
```
stuartmorgan, line 333: Nit: missing period.

packages/video_player/video_player_web/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.6
+
+* Bump `video_player_platform_interface` version
```
stuartmorgan, line 3: Why?

stuartmorgan (changes requested):
Some nits, but this looks good overall.

@gaaclarke for secondary review before we split out the platform interface change.

ZhangZhiH:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

packages/video_player/video_player_web/CHANGELOG.md
IlyaMax: wasn't sure about this change. I had problems with building example without this. But maybe that because there weren't versions constraints when I published pr.

IlyaMax:
@stuartmorgan What do you think about removing allowBackgroundPlayback option from VideoPlayerOptions to simplify this pr? As possible solution I think we can rename VideoPlayerOptions to something like NativeVideoPlayerOptions, because allowBackgroundPlayback setting configure dart behavior and create VideoPlayerOptions for dart side to avoid code duplication. For example I have idea also to provide option not to recover playing state after `AppLifecycleState.resumed`

stuartmorgan:
> As possible solution I think we can rename `VideoPlayerOptions` to something like `NativeVideoPlayerOptions`

That would be a good option, except that it would be a breaking change for the platform interface package, which we only do when necessary.

We could potentially use the same name and some combination of `hide` and `import as` to avoid collisions, but I think that's probably more confusing than just dealing with needing to make Dart-only additions in the platform interface.

IlyaMax:
Yes, it seems that breaking change is not reasonable for now. There are no options to configure on dart apart from these two, at least I can't think of them.

stuartmorgan:
(We could theoretically a new secondary options class—e.g., `VideoPlayerControllerOptions`—and make it an additional optional argument, but that would make the public-facing API really confusing, so isn't something I'd actually want to do.)

stuartmorgan:
@gaaclarke Ping on secondary review

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -299,8 +299,12 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
 
   /// Attempts to open the given [dataSource] and load metadata about the video.
   Future<void> initialize() async {
-    _lifeCycleObserver = _VideoAppLifeCycleObserver(this);
-    _lifeCycleObserver.initialize();
+    final allowBackgroundPlayback =
```
gaaclarke, line 302: nit: The linter isn't flagging this for omitting the datatype?

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -95,6 +95,22 @@ class _FakeClosedCaptionFile extends ClosedCaptionFile {
 }
 
 void main() {
+  void verifyAppLifeCycle(
```
gaaclarke, line 98: nit: this name isn't very helpful.  Something like verifyPlayingWhenAppLifeCyclePaused?

gaaclarke (approved):
lgtm

packages/video_player/video_player/lib/video_player.dart
stuartmorgan: We haven't converted video_player off of the legacy options yet unfortunately.

gaaclarke:
> @stuartmorgan is it ok if I close this pr and open a new one? I've deleted old repo, so I'm not sure how I can fix nits without opening new pr.

The code still exists on github and you can check it out at `surfstudio:add-option-to-not-observe-lifecycle-final-pr`.

Some thing like:
```
git clone git@github.com:flutter/plugins.git
cd plugins
git checkout origin/add-option-to-not-observe-lifecycle-final-pr -b add-option-to-not-observe-lifecycle-final-pr
```

Alternatively you could try using github's built in editor so you can edit the files without downloading them locally.

IlyaMax:
@gaaclarke this doesn't work unfortunately. In both my new fork and clone this is the list of available branches. I can't checkout pr branches. Is this so critical if I create a new pr and link this pr there? 
image

stuartmorgan:
Given that you've deleted the *remote* repository, I don't think there's anything you can do other than make a new PR. AFAIK GitHub doesn't allow changing the source of a PR after the fact.

(In the future, please don't delete the repository hosting an active PR. Changing PRs is disruptive to the review process since we can't see incremental diffs and can't easily reference previous comments.)

PR #3324, opened 2020-12-11
Merge null-safety plugins

blasten:

PR #4327, opened 2021-09-09
[camera] Bump minimum Flutter version and iOS deployment target

jmagman:
Flutter apps run on 2.4.0-0.0.pre and later [will be upgraded to a minimum of iOS 9.0](https://github.com/flutter/flutter/pull/85174).  Now that Flutter 2.5 has hit stable and [iOS 8 support has been dropped](https://flutter.dev/go/rfc-ios8-deprecation), change the plugin minimum iOS version to `9.0`.  

Bump the Flutter and dart version constraints to 2.5 and 2.14 respectively to enforce that the iOS 9.0 migration has happened on the app side, because an older 8.0 app won't build with a 9.0 plugin.  I believe the last time this was done was a few months ago for null safety adoption https://github.com/flutter/plugins/pull/3324 (there's recent precedent).

Users on Flutter versions lower than the current stable 2.5 will not be able to upgrade this plugin past this version.  

Camera part of https://github.com/flutter/flutter/issues/84198

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -504,7 +504,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
```
jmagman, line 507: Same, done automatically via https://github.com/flutter/flutter/pull/85174.

jmagman:
@stuartmorgan Do you think this change is too soon considering 2.5 stable only dropped this week?

stuartmorgan (approved):
> @stuartmorgan Do you think this change is too soon considering 2.5 stable only dropped this week?

No, this is fine; we only guarantee compatibility with the latest stable version for 1P plugins. And the plugin updates will also only drop this week 🙂 

One small nit: for clarity's sake we should probably update the `example` app's minimum Flutter SDK version too. Looks like we forgot that during NNBD migration, but this is a good opportunity to clean it up. (It's not a big deal if we don't obviously, but I assume the error message will be clearer when trying to run the example with an older version if we do.)

jmagman:
> One small nit: for clarity's sake we should probably update the `example` app's minimum Flutter SDK version too. Looks like we forgot that during NNBD migration, but this is a good opportunity to clean it up.

Done.

stuartmorgan (approved):
Still LGTM

PR #4328, opened 2021-09-09
[webview_flutter] Fix platform exception on iOS 14+ when evaluating any JavaScript that evaluates to null or undefined.

BeMacized:
Currently, a PlatformException is thrown whenever `evaluateJavascript` is run for any JavaScript that evaluates to `null` or `undefined`. 

This is due to (I assume to be) a regression in webkit's [evaluateJavascript](https://developer.apple.com/documentation/webkit/wkwebview/1415017-evaluatejavascript?language=objc) function, which no longer supports `null` or `undefined` for the given JavaScript to evaluate to, as of iOS 14. (Still tested working on iOS 11.4 and 12.4) 

This PR attempts to fix this problem by instead evaluating JavaScript using [callAsyncJavaScript](https://developer.apple.com/documentation/webkit/wkwebview/3656355-callasyncjavascript?language=objc) on iOS 14+. 

Relevant issue:
- flutter/flutter#66318

## Pre-launch Checklist

- [x] I signed the [CLA].
  - **NOTE:** I have marked this as a _patch_. However, iOS will now always return `"null"` for both `null` and `undefined` to match the Android side's behavior, rather than `"(null)"` for `null` and `"<null>"` for `undefined`. I am not entirely certain if this should be considered a breaking change.


packages/webview_flutter/webview_flutter/ios/Classes/FlutterWebView.m
```diff
@@ -230,18 +238,31 @@ - (void)onEvaluateJavaScript:(FlutterMethodCall*)call result:(FlutterResult)resu
...
+                            stringWithFormat:@"JavaScript string was: '%@'\n%@", jsString, error]]);
+    } else if (evaluateResult == /*null*/ [NSNull null] || evaluateResult == /*undefined*/ nil) {
+      result(@"null");
+    } else {
+      result([NSString stringWithFormat:@"%@", evaluateResult]);
+    }
+  };
+  if (@available(iOS 14, *)) {
+    jsString = [NSString
+        stringWithFormat:@"return eval(\"%@\");", [self escapeStringForJavaScript:jsString]];
```
stuartmorgan, line 257: What's the reason for wrapping this in an `eval`? That seems like something that could cause side-effects. E.g., I believe CSP could completely break this.

BeMacized: The reason for this is that unlike `evaluateJavascript`, `callAsyncJavaScript` executes the given string as the body of an async function, rather than directly evaluating it. As such, it will only return any value that is explicitly returned, not the result of the string being evaluated. I use `eval` here to make sure the behavior is identical to `evaluateJavascript` on iOS 13 and below. 

Personally I have not been able to find any edge cases or side effects, but I've also not tried anything to do with CSP. If you have any examples of this maybe we can find a way around it?

The only other way I can think of to avoid these side effects, assuming they're a thing, is to fix this the other way around:(make `evaluateJavaScript` behave like `callAsyncJavaScript`), forcing everyone to explicitly return a value in their JS strings, and consider it a breaking change.

Please let me know what you think. In the mean time I'll update this PR to move the fix to the separated iOS implementation.

stuartmorgan: > (make `evaluateJavaScript` behave like `callAsyncJavaScript`), forcing everyone to explicitly return a value in their JS strings, and consider it a breaking change.

How would that be better than the current behavior?

stuartmorgan: I mean master; isn't the issue you are trying to fix that a void return causes an error? Can't people already avoid that error right now by explicitly returning a value?

BeMacized: On master the platform exception can indeed be avoided by ensuring the given string evaluates to a non-null/non-undefined value.

E.g:
`Snackbar.postMessage("User Agent: " + navigator.userAgent)` Throws.
`Snackbar.postMessage("User Agent: " + navigator.userAgent);1` Does not.

BeMacized:
Superseded by #4401

PR #4331, opened 2021-09-10
[video_player] Fix playback speed resetting on iOS

alexagat:
This stabilizes the behavior of the `video_player` package such that calling `seekTo` does not reset the playback speed to `1.0` on the iOS platform.

* https://github.com/flutter/flutter/issues/71264

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
+++ b/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
@@ -315,7 +315,7 @@ static inline CGFloat radiansToDegrees(CGFloat radians) {
     [_player play];
 
     // fix to always set playback speed accurately
-    [self setPlaybackSpeed: _internalPlaybackSpeed];
+    [self setPlaybackSpeed:_internalPlaybackSpeed];
   } else {
     [_player pause];
   }
--- a/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
+++ b/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
@@ -40,6 +40,7 @@
 @property(nonatomic, readonly) bool disposed;
 @property(nonatomic, readonly) bool isPlaying;
 @property(nonatomic) bool isLooping;
+@property(nonatomic) double internalPlaybackSpeed;
 @property(nonatomic, readonly) bool isInitialized;
 - (instancetype)initWithURL:(NSURL*)url
                frameUpdater:(FLTFrameUpdater*)frameUpdater
@@ -47,6 +48,7 @@
 - (void)play;
 - (void)pause;
 - (void)setIsLooping:(bool)isLooping;
+- (void)setInternalPlaybackSpeed:(double)speed;
 - (void)updatePlayingState;
 @end
 
@@ -242,6 +244,7 @@ static inline CGFloat radiansToDegrees(CGFloat radians) {
 
   _player = [AVPlayer playerWithPlayerItem:item];
   _player.actionAtItemEnd = AVPlayerActionAtItemEndNone;
+  [self setInternalPlaybackSpeed:1.0f];
 
   [self createVideoOutputAndDisplayLink:frameUpdater];
 
@@ -310,6 +313,9 @@ static inline CGFloat radiansToDegrees(CGFloat radians) {
   }
   if (_isPlaying) {
     [_player play];
+
+    // fix to always set playback speed accurately
+    [self setPlaybackSpeed: _internalPlaybackSpeed];
   } else {
     [_player pause];
   }
@@ -369,6 +375,10 @@ static inline CGFloat radiansToDegrees(CGFloat radians) {
   _isLooping = isLooping;
 }
 
+- (void)setInternalPlaybackSpeed:(double)speed {
+  _internalPlaybackSpeed = speed;
+}
+
 - (void)setVolume:(double)volume {
   _player.volume = (float)((volume < 0.0) ? 0.0 : ((volume > 1.0) ? 1.0 : volume));
 }
@@ -394,6 +404,7 @@ static inline CGFloat radiansToDegrees(CGFloat radians) {
     return;
   }
 
+  [self setInternalPlaybackSpeed:speed];
   _player.rate = speed;
 }
 
--- a/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
+++ b/packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
@@ -48,7 +48,6 @@
 - (void)play;
 - (void)pause;
 - (void)setIsLooping:(bool)isLooping;
-- (void)setInternalPlaybackSpeed:(double)speed;
 - (void)updatePlayingState;
 @end
 
@@ -244,7 +243,6 @@ static inline CGFloat radiansToDegrees(CGFloat radians) {
 
   _player = [AVPlayer playerWithPlayerItem:item];
   _player.actionAtItemEnd = AVPlayerActionAtItemEndNone;
-  [self setInternalPlaybackSpeed:1.0f];
 
   [self createVideoOutputAndDisplayLink:frameUpdater];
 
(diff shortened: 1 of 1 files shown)


alexagat:
@googlebot I fixed it

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
```diff
@@ -242,6 +244,7 @@ - (instancetype)initWithPlayerItem:(AVPlayerItem*)item frameUpdater:(FLTFrameUpd
 
   _player = [AVPlayer playerWithPlayerItem:item];
   _player.actionAtItemEnd = AVPlayerActionAtItemEndNone;
+  [self setInternalPlaybackSpeed:1.0f];
```
stuartmorgan, line 247: Why is this needed on init?

Also, per style guide, this should not call self: https://google.github.io/styleguide/objcguide.html#avoid-messaging-the-current-object-within-initializers-and--dealloc
(There are existing violations, but they are wrong, and we shouldn't add more.)

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
```diff
@@ -310,6 +313,9 @@ - (void)updatePlayingState {
   }
   if (_isPlaying) {
     [_player play];
+
+    // fix to always set playback speed accurately
+    [self setPlaybackSpeed:_internalPlaybackSpeed];
```
stuartmorgan, line 318: Is this always needed here? My impression from comments in the issue was that calling it from the completion handler of the seek method would likely be enough.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
```diff
@@ -369,6 +375,10 @@ - (void)setIsLooping:(bool)isLooping {
   _isLooping = isLooping;
 }
 
+- (void)setInternalPlaybackSpeed:(double)speed {
```
stuartmorgan, line 378: Why are you overriding a property setter with the default implementation of a property setter.

stuartmorgan (changes requested):
Thanks for the contribution!

Comments inline, but this will also need tests (and a version update), per the checklist and contribution guide.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
alexagat: Good point, this is not needed and I will remove.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
alexagat: I was copying the pattern I saw above, but after referring to [Apple docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjectiveC/Chapters/ocProperties.html), I see that property declarations give us the getter/setters for free. Nice! Will remove

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
alexagat: @stuartmorgan which completion handler are you referring to? I only see a `seekTo` method in the file.

stuartmorgan: See https://github.com/flutter/flutter/issues/71264#issuecomment-734656335

alexagat: @stuartmorgan I see. I tried moving the call inside the completionHandler block of the `seekTo` method as follows:

```
[_player seekToTime:CMTimeMake(location, 1000)
      toleranceBefore:kCMTimeZero
       toleranceAfter:kCMTimeZero
    completionHandler:^(BOOL finished) {
    [self setPlaybackSpeed:_internalPlaybackSpeed];
  }];
```

but the issue of the playback speed resetting to `1.0` came back. Therefore that is not equivalent.

stuartmorgan: Interesting. So presumably then it's the `playbackLikelyToKeepUpContext` and/or `AVPlayerItemStatusReadyToPlay` codepath that's resolving this? In which case perhaps the underlying issue is that AVPlayer automatically resets the playback speed when the buffer is insufficient.

alexagat:
@stuartmorgan  / @creativecreatorormaybenot I've realized that while this modification fixes the playback speed issue it causes a negative side-effect. When an unbuffered range of the video stream is played back, the video player can lose volume (go mute) and become unstable. Once it gets in this state it does not recover. Therefore looks like there's still more work, unfortunately.

creativecreatorormaybenot:
@alexagat if you take a look at my PR (#3404), it _does_ fix the issue. Can we not just perform the same action on the native side?

Because at the end of the day, I am only making a method channel call..

stuartmorgan:
> When an unbuffered range of the video stream is played back, the video player can lose volume (go mute) and become unstable.

Does setting it back only for `playbackLikelyToKeepUpContext` work better, or is that already the only place it was being called in practice?

> @alexagat if you take a look at my PR (#3404), it _does_ fix the issue.

It is very likely that your approach would racily have the same behavior. There's no guarantee that some fraction of 500ms is enough time to create sufficient buffer to avoid an overrun.

alexagat:
@creativecreatorormaybenot you are correct your PR does fix this issue on the iOS platform and does not add any of the negative side-effects the native fix above introduced. I played around for a bit but didn't come up with a native-only solution here. For now I've added your code to my fork so we can patch this issue for users.

creativecreatorormaybenot:
@alexagat yeah, that is what we are also doing in production, using our own fork of the plugin.

stuartmorgan:
If there's no way to automatically handle this in the plugin that doesn't have the potential to cause playback issues, the other option would be to add an event that would report playback speed changes back to the Dart side, so that UIs could update, and users could change it back themselves.

But I would expect that there's a way to determine the relevant state to handle this correctly. Does observing `timeControlStatus` give any indication as to what's happening?

creativecreatorormaybenot:
@stuartmorgan I have one concern with this part:

> If there's no way to automatically handle this in the plugin that doesn't have the potential to cause playback issues, the other option would be to add an event that would report playback speed changes back to the Dart side, so that UIs could update, and users could change it back themselves.

You were saying that my fix was a hack. But your proposal is not? If I look at this from a different perspective, this means that users of the problem would need to integrate the hack (changing the playback speed after the change has been reported to the Dart side).  
I see how this would be better in some odd way, but then again I really feel like this is even worse than just using the not-so-great fix because it also makes for a very bad API IMHO.

> But I would expect that there's a way to determine the relevant state to handle this correctly.

That would be my preferred choice as well..

stuartmorgan:
> You were saying that my fix was a hack. But your proposal is not?

There are two possibilities here:
1. There is a reliable programatic way to determine when it is safe to automatically set the playback rate back to the desired value. If there is, we should do that.
2. There is no reliable way to do it automatically, such that any attempt to do so may break playback in the way described above. In this case I don't see any alternative to simply communicating the state back to the UI so that users can control it.

Your PR was sort of attempting to do 1, but in a very hacky way. I'm saying that either we need to do 1 correctly, or if 1 is impossible we should do 2. If 2 is the only option it's not a hack, it a reflection of the limitations of the platform that the plugin is wrapping.

alexagat:
@stuartmorgan I am not planning to update this PR and it should probably be closed. I never got the iOS implementation to work as desired and ended up using the code change from https://github.com/flutter/plugins/pull/3404 to resolve the bug.

creativecreatorormaybenot:
Which is why is still think it should be merged 👍 It is not *perfect* but the best from a cost <> value perspective.

stuartmorgan:
Closing per the above; thanks for the investigation so far. This should serve as a good starting point for future attempts to fix the issue.

PR #4333, opened 2021-09-11
[google_maps_flutter] Bump minimum Flutter version and iOS deployment target

jmagman:
Flutter apps run on 2.4.0-0.0.pre and later [will be upgraded to a minimum of iOS 9.0](https://github.com/flutter/flutter/pull/85174).  Now that Flutter 2.5 has hit stable and [iOS 8 support has been dropped](https://flutter.dev/go/rfc-ios8-deprecation), change the plugin minimum iOS version to `9.0`.  

Bump the Flutter and dart version constraints to 2.5 and 2.14 respectively to enforce that the iOS 9.0 migration has happened on the app side, because an older 8.0 app won't build with a 9.0 plugin.  I believe the last time this was done was a few months ago for null safety adoption https://github.com/flutter/plugins/pull/3324 (there's recent precedent).

Users on Flutter versions lower than the current stable 2.5 will not be able to upgrade this plugin past this version.  

google_maps_flutter part of https://github.com/flutter/flutter/issues/84198

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
+++ b/packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
@@ -2,11 +2,11 @@ name: google_maps_flutter
 description: A Flutter plugin for integrating Google Maps in iOS and Android applications.
 repository: https://github.com/flutter/plugins/tree/master/packages/google_maps_flutter/google_maps_flutter
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+maps%22
-version: 2.0.9
+version: 2.0.10
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
-  flutter: ">=2.0.0"
+  sdk: ">=2.14.0 <3.0.0"
+  flutter: ">=2.5.0"
 
 flutter:
   plugin:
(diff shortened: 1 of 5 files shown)


packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
```diff
@@ -2,11 +2,11 @@ name: google_maps_flutter
 description: A Flutter plugin for integrating Google Maps in iOS and Android applications.
 repository: https://github.com/flutter/plugins/tree/master/packages/google_maps_flutter/google_maps_flutter
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+maps%22
-version: 2.0.8
+version: 2.1.0
```
cyanglaz, line 5: I don't think there's a public API change here so a minor version bump should be enough.

jmagman: Requiring a new Flutter stable isn't cause enough for a minor bump? @stuartmorgan what do you think?

stuartmorgan: I don't think what we're doing really fits into the semver scheme, so I think it doesn't especially matter. Since the resolver will prevent someone for whom this won't work from ever getting it in the first place, and for anyone else it's basically a bugfix (it won't get modified on the fly, and for some packages will stop having so many deprecation warnings) I think a bugfix bump is fine.

But I'm also fine with saying "it's not exactly a bugfix either so lets call it a minor bump". 🤷🏻

(If I'm the tiebreaker here, I think I'd lean slightly toward bugfix bump rather than minor bump).

jmagman: Done.

stuartmorgan (approved):
LGTM

PR #4339, opened 2021-09-11
[webview_flutter] - Added WebViewSetting option for iOS

GauteHaugen:
Added limitsNavigationsToAppBoundDomains that is an optional setting for iOS

Added `limitsNavigationsToAppBoundDomains` property to `WebViewSettings` that can be used on iOS. This will give the developer access to some JavaScript APIs that are disabled by default. 

List which issues are fixed by this PR. You must list at least one issue.
([flutter/issues/89530](https://github.com/flutter/flutter/issues/89530))

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Thanks for the submission! As explained in the contributing guide, all behavioral changes need tests. Please let us know when you’ve updated the PR with tests and we can take a look.

stuartmorgan:
@GauteHaugen Are you planning on updating this PR to add tests?

packages/webview_flutter/webview_flutter/README.md
```diff
@@ -92,3 +92,28 @@ android {
...
+            <array>
+                <string>example1.com</string>
+                <string>example2.org</string>
+                ...
+            </array>
+          </dict>
+        ```
+  * `Array` of App-Bound Domains is limited to 10 domains
+  * Content supplied by the app through local files, data URLs and HTML String are always treated
+    as App-Bound Domains, and do not need to be listed.
```
stuartmorgan, line 119: Please link to authoritative documentation of the OS feature rather than duplicating it here (where it could easily become outdated).

packages/webview_flutter/webview_flutter_android/example/lib/web_view.dart
```diff
@@ -178,6 +179,32 @@ class WebView extends StatefulWidget {
...
+  ///             <array>
+  ///                 <string>example1.com</string>
+  ///                 <string>example2.org</string>
+  ///                 ...
+  ///             </array>
+  ///           </dict>
+  ///         ```
+  ///   * `Array` of App-Bound Domains is limited to 10 domains
+  ///   * Content supplied by the app through local files, data URLs and HTML String are always treated
+  ///     as App-Bound Domains, and do not need to be listed.
```
stuartmorgan, line 205: Same here; just link to the documentation.

packages/webview_flutter/webview_flutter_wkwebview/example/lib/web_view.dart
```diff
@@ -156,6 +157,32 @@ class WebView extends StatefulWidget {
...
+  ///             <array>
+  ///                 <string>example1.com</string>
+  ///                 <string>example2.org</string>
+  ///                 ...
+  ///             </array>
+  ///           </dict>
+  ///         ```
+  ///   * `Array` of App-Bound Domains is limited to 10 domains
+  ///   * Content supplied by the app through local files, data URLs and HTML String are always treated
+  ///     as App-Bound Domains, and do not need to be listed.
```
stuartmorgan, line 183: You don't need to re-document the feature in the example. If people want to know what the feature does, they should look at the API documentation.

packages/webview_flutter/webview_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.1.3
+
+* Added `limitsNavigationsToAppBoundDomains` to `WebSettings` ([flutter/issues/89530](https://github.com/flutter/flutter/issues/89530))
```
stuartmorgan, line 3: We don't generally reference issues in CHANGELOGs

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m
```diff
@@ -348,6 +348,12 @@ - (NSString*)applySettings:(NSDictionary<NSString*, id>*)settings {
     } else if ([key isEqualToString:@"userAgent"]) {
       NSString* userAgent = settings[key];
       [self updateUserAgent:[userAgent isEqual:[NSNull null]] ? nil : userAgent];
+    } else if ([key isEqualToString:@"limitsNavigationsToAppBoundDomains"]) {
+      if (@available(iOS 14.0, *)) {
+        NSNumber* limitsNavigationsToAppBoundDomains = settings[key];
+        _webView.configuration.limitsNavigationsToAppBoundDomains =
+            [limitsNavigationsToAppBoundDomains boolValue];
+      }
```
stuartmorgan, line 356: There's no test coverage of this part of the change; if this were accidentally removed in the future, the feature would stop working but no tests would fail. The best option here would be a native unit test that sends a message with this enabled, and ensures (probably via an exposed-for-testing accessor to give you the `_webView`) that the configuration is set correctly.

stuartmorgan-g (changes requested):
It'll need to be split into three PRs; see https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins

But there's current a large rewrite happening in the Android implementation that will eliminate the need for the Android change here, so I would suggest starting with splitting out just the platform interface PR; once that's reviewed and landed, hopefully the Android change will have landed, and then you can split out just the iOS change.

(I left a few initial comments; it wasn't a full review, but generally the overall change looks good to proceed with splitting it.)

GauteHaugen:
@stuartmorgan, I have now followed the steps in the "Changing federated packages" guide. The first PR #4464 is up with the changes for flutter_webview_platform_interface. I have also as the guide explained changed this PR to include dependency_overrides until the other PRs have been merged

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m
GauteHaugen: I understand why this needs a unit test, but I'm not sure how to implement testing for it. I tried looking at the other properties, and to me, it looks like none of the properties would be caught in a test if native code were altered. So I'm not entirely sure how to proceed with the unit test.

Some of the questions I have regarding the native unit test. 

- Should it be in the flutter_webview_wkwebview package or the flutter_webview package?
- Should it be all swift code, or is there some way to test from flutter with flutter code sending the message over the message channel?
- Also if the test is added to the ios folder, is there anything special I need to do to make it work with the GitHub workflows/checks?

stuartmorgan: > it looks like none of the properties would be caught in a test if native code were altered.

That's true, but: https://github.com/flutter/flutter/wiki/Plugin-Tests#faq-do-i-need-to-add-tests-even-if-the-part-of-the-code-im-changing-isnt-already-tested

> * Should it be in the flutter_webview_wkwebview package or the flutter_webview package?

The former. Unit tests should live in the same package as the code they are testing.

> * Should it be all swift code, or is there some way to test from flutter with flutter code sending the message over the message channel?

It should be Objective-C; testing from Dart would require an integration test, which would be much more involved for this case.

> * Also if the test is added to the ios folder, is there anything special I need to do to make it work with the GitHub workflows/checks?

Native unit tests are already configured for the package, you just need to add a test to those existing test files (or if you make a new file, add it to the same target).

GauteHaugen:
@stuartmorgan, while writing the iOS native unit test. I found out that `limitsNavigationsToAppBoundDomains` is not a setting that can change after the `WKWebView` has been initialized. So I have rewritten the implementation to be set before the `WKWebView` is initialized. I have also changed the flutter unit test to reflect the changes, as well as added iOS native unit tests. I have also updated #4464 with the changes

packages/webview_flutter/webview_flutter/lib/src/webview.dart
```diff
@@ -94,6 +96,7 @@ class WebView extends StatefulWidget {
     this.initialMediaPlaybackPolicy =
         AutoMediaPlaybackPolicy.require_user_action_for_all_media_types,
     this.allowsInlineMediaPlayback = false,
+    this.limitsNavigationsToAppBoundDomains = false,
```
stuartmorgan, line 99: Adding platform-specific arguments to this list is not a direction I want to take if we can possibly avoid it; it could get out of hand very quickly.

@bparrishMines @mvanbeusekom You're both more familiar than I am with the overall API of this plugin; do either of you have thoughts on how to introduce platform-specific extensions here?

I'm wondering if it would make sense to add a creation parameters class that's created by a factory, implemented by the implementing packages, so that they could return subclasses of the settings object that have extensions. Then clients could do platform-specific casting to access settings like this one, similar to the IAP extensions.

stuartmorgan-g (changes requested):

packages/webview_flutter/webview_flutter/lib/src/webview.dart
bparrishMines: In the long term, I believe we could use casting to add cleaner support for very specific platform features like this. For example, the API would look something like:

```dart
class MyWidget extends StatelessElement {
  MyWidget(StatelessWidget widget) : super(widget);

  Future<void> additionalIosSetup(IosWebViewController controller) {
    controller.configuration.setLimitsNavigationsToAppBoundDomains(false);
  }

  Future<void> additionalAndroidSetup(AndroidWebViewController controller) {
    ...
  }

  @override
  Widget build() {
    return WebView(
        if (Platform.isIOS) {
          additionalIosSetup(controller as IosWebViewController);
        } else if (Platform.isAndroid) {
          additionalAndroidSetup(controller as AndroidWebViewController);
        }
      },
    );
  }
}
```

However, this can only be done after the iOS platform implementation is transitioned to pigeon and we update the platform interface. After https://github.com/flutter/plugins/pull/4503 lands, this should be possible for Android. So we may want to hold off on adding this feature until iOS is also transitioned.

GauteHaugen: Not sure if I'm missing something with your example @bparrishMines, but `limitsNavigationsToAppBoundDomains` is not a setting that can be changed after the `WebView` has been initialized. It needs to be configured during the initialize process of the `FLTWebViewController` object. And if I understand it correctly `onWebViewCreated` is called after both the WebView Configuration and the WebView is done initializing

bparrishMines: Ah, I hadn't realized it was part of initialization. Then, I agree with @stuartmorgan that we could expose a `CreationParams` class that each platform can add additional fields too. A solution that could look something like this in the platform interface: 

```dart
class CreationParams {
  factory CreationParams() {
    return WebView.platform.createCreationParams();
  }
}

abstract class WebViewPlatform {
.
.
.
  void createCreationParams() {
    throw UnimplementedError();
  }
}
```

@stuartmorgan Is this what you were considering?

stuartmorgan: That looks pretty much like I was thinking (except that we'd probably want the platform interface class to return the base class rather than throw unimplemented, so that only platforms that want extra parameters need to implement it).

GauteHaugen: @stuartmorgan is this something I should implement, or should I wait for it to be implemented before adding my changes? Also, how would this look in an app where I might target multiple platforms and want toggle options for multiple platforms?

stuartmorgan: It's not something there's an existing plan to implement, so you should go ahead and implement it. I would start with trying it as part of this PR; we could do it as a separate PR that lands first, but it's probably useful to build (and review) in conjunction with this use as a test case to make sure it's working as we want it to end to end. (If it turns out to be more complex than anticipated for some reason, it could be split out later.)

> Also, how would this look in an app where I might target multiple platforms and want toggle options for multiple platforms?

Basically yike [the example shown above](https://github.com/flutter/plugins/pull/4339#discussion_r749486889), but with the settings object being cast, not the controller.

stuartmorgan: > so you should go ahead and implement it

Oh, but as @bparrishMines mentioned, it should wait a couple of weeks since the restructuring that's happening in the webview implementation will probably interact with this.

GauteHaugen: @stuartmorgan Sounds good, is there an active issue or PR for the iOS transition to pigeon that I can follow to know when this change is landed?

stuartmorgan: There wasn't, but really should be. I filed https://github.com/flutter/flutter/issues/93732 (Android is in progress now and nearing completion, iOS will follow after).

stuartmorgan:
Status update from triage: The iOS conversion to Dart is still in progress, so this is still on hold.

lllyyy:
跳转其他url无法触发onPageFinished

stuartmorgan:
https://github.com/flutter/flutter/issues/93732 is complete, so this is unblocked; you'll just need to update the implementation for the new Dart-based implementation.

stuartmorgan:
@GauteHaugen Are you still planning on updating this to the new implementation?

GauteHaugen:
@stuartmorgan I can take a look at it over the weekend. Is the plan to still implement it as discussed in the comments above?

stuartmorgan:
I believe so. @bparrishMines Is the thread above still the model for the v4 structure?

bparrishMines:
I think this feature is still blocked because it requires the feature to set a platform specific `CreationParams`. This is done in the v4 api being worked on in https://github.com/flutter/flutter/issues/94051. But, it's still not available in the current version.

GauteHaugen:
> I think this feature is still blocked because it requires the feature to set a platform specific `CreationParams`. This is done in the v4 api being worked on in [flutter/flutter#94051](https://github.com/flutter/flutter/issues/94051). But, it's still not available in the current version.

That was going to be my next question. When I was looking at the code, I could not find an easy way to override CreationsParams. Then I will wait for https://github.com/flutter/flutter/issues/94051 to land before implementing

stuartmorgan:
Sorry about that; I forgot that platform-specific creation params were not supported before, and conflated it with just changing `CreationParams` in general (which was one of the things we could still actually change in v3).

zbuhler:
Any update on this issue since September? Really would love this option as PWAs / Service Workers cannot work inside of a flutter WebView it seems. Looking at v4 of flutter WebView seems like I should be able to set it via WebKitWebViewControllerCreationParams and seems to be what is being referred to in the last comment.

stuartmorgan:
webview_flutter v4 has finally been completed, which included fixing the platform interface to that we can make additions to it without breaking changes. That means that this is finally unblocked! Thanks for your patience while we worked through the restructuring.

webview_flutter v4 uses a Dart-first implementation, where the platform channel boundary is as close as possible to the underlying host APIs, and plugin logic is written primarily in Dart rather than the host language. That means that this PR will need to be updated to follow that model. Please let us know if you have questions about updating to v4 (and https://github.com/flutter/plugins/pull/6881 may be useful as an example of adding a feature in the v4 structure).

stuartmorgan:
I'm going to go ahead and mark this as a draft since it needs to be reworked for v4, just to get it off of our review queue. Please mark it as ready for review once that's been done. Thanks!

stuartmorgan-g:

Please see [our instructions](https://github.com/flutter/flutter/wiki/Migrating-Plugins-repository-PRs-to-Packages) for an explanation of how to move your PR, and if you have any issues moving your PR please don't hesitate to reach out in the #hackers-ecosystem channel in [Discord](https://github.com/flutter/flutter/wiki/Chat).

Our apologies that your PR was caught in this one-time transition. We're aware that it's disruptive in the short term, and appreciate your help in getting us to a better long-term state!

PR #4340, opened 2021-09-13
Renew cirrus gcp credentials

keyonghan:
GCP account key expires every three months. This CL updates the GCP credential based on the renewed key.


stuartmorgan (approved):
LGTM

PR #4365, opened 2021-09-21
[url_launcher] Error handling when URL cannot be parsed with Uri.parse

Baw-Appie:
image

Uri.parse cannot parse some app scheme URLs.
Use try-catch to handle errors.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/url_launcher/url_launcher/lib/url_launcher.dart
+++ b/packages/url_launcher/url_launcher/lib/url_launcher.dart
@@ -75,8 +75,8 @@ Future<bool> launch(
   try {
     final Uri url = Uri.parse(urlString.trimLeft());
     isWebURL = url.scheme == 'http' || url.scheme == 'https';
-  } on FormatException {}
-  
+  } on FormatException catch (_) {}
+
   if ((forceSafariVC == true || forceWebView == true) && !isWebURL) {
     throw PlatformException(
         code: 'NOT_A_WEB_SCHEME',
--- a/packages/url_launcher/url_launcher/lib/url_launcher.dart
+++ b/packages/url_launcher/url_launcher/lib/url_launcher.dart
@@ -71,11 +71,9 @@ Future<bool> launch(
   Brightness? statusBarBrightness,
   String? webOnlyWindowName,
 }) async {
-  bool isWebURL = false;
-  try {
-    final Uri url = Uri.parse(urlString.trimLeft());
-    isWebURL = url.scheme == 'http' || url.scheme == 'https';
-  } on FormatException catch (_) {}
+  final Uri? url = Uri.tryParse(urlString.trimLeft());
+  final bool isWebURL =
+      url != null && (url.scheme == 'http' || url.scheme == 'https');
 
   if ((forceSafariVC == true || forceWebView == true) && !isWebURL) {
     throw PlatformException(
--- a/packages/url_launcher/url_launcher/lib/url_launcher.dart
+++ b/packages/url_launcher/url_launcher/lib/url_launcher.dart
@@ -74,15 +74,15 @@ Future<bool> launch(
   bool isWebURL = false;
   try {
     final Uri url = Uri.parse(urlString.trimLeft());
-
     isWebURL = url.scheme == 'http' || url.scheme == 'https';
-    if ((forceSafariVC == true || forceWebView == true) && !isWebURL) {
-      throw PlatformException(
-          code: 'NOT_A_WEB_SCHEME',
-          message: 'To use webview or safariVC, you need to pass'
-              'in a web URL. This $urlString is not a web URL.');
-    }
-  } catch (e) {}
+  } on FormatException {}
+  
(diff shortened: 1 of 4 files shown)


stuartmorgan:
Thanks for the submission!

> Uri.parse cannot parse some app scheme URLs.

Please file an issue describing this problem in detail, with a specific example. The PR as submitted is not correct, since it will always catch the deliberately thrown exception, so we need to better understand what the problem is to figure out how to move forward.

Baw-Appie:
There is an issue with Uri.parse not correctly parsing URLs containing ":"

For example, Microsoft Remote Desktop uses a custom app scheme URL like the one below, but Uri.parse throws an FormatException error.
```
rdp://full address=s:example.com:3389&username=s:Administrator
```

Dart recognizes everything after the ":" as a port, but it seems to be giving an error because it's a string.

Should I file an issue to dart-lang/sdk instead of submitting this PR?

stuartmorgan:
> Should I file an issue to dart-lang/sdk instead of submitting this PR?

It turns out that's not actually a valid URI per RFC 3986, which is what `Uri` uses, so it's not actually a Dart bug. Given that, a more targeted version of the change you've proposed here is reasonable.

packages/url_launcher/url_launcher/lib/url_launcher.dart
```diff
@@ -71,14 +71,18 @@ Future<bool> launch(
...
+    final Uri url = Uri.parse(urlString.trimLeft());
+
+    isWebURL = url.scheme == 'http' || url.scheme == 'https';
+    if ((forceSafariVC == true || forceWebView == true) && !isWebURL) {
+      throw PlatformException(
+          code: 'NOT_A_WEB_SCHEME',
+          message: 'To use webview or safariVC, you need to pass'
+              'in a web URL. This $urlString is not a web URL.');
+    }
+  } catch (e) {}
```
stuartmorgan, line 85: This should only be wrapping `parse`, not everything (especially not the `throw`) and should only catch the specific exception it is intended to handle.

stuartmorgan (changes requested):
In addition to the specific comment below, this will need the rest of the items on the PR checklist addressed. Most importantly, this needs a unit test that validates behavior using a sample RDP URL.

stuartmorgan:
> rdp://full address=s:example.com:3389&username=s:Administrator

When you write the test, please properly encode the space (see https://docs.microsoft.com/en-us/windows-server/remote/remote-desktop-services/clients/remote-desktop-uri for example); as written this is invalid for reasons unrelated to the actual problem with rdp: URLs.

packages/url_launcher/url_launcher/lib/url_launcher.dart
```diff
@@ -71,8 +71,12 @@ Future<bool> launch(
   Brightness? statusBarBrightness,
   String? webOnlyWindowName,
 }) async {
-  final Uri url = Uri.parse(urlString.trimLeft());
-  final bool isWebURL = url.scheme == 'http' || url.scheme == 'https';
+  bool isWebURL = false;
+  try {
+    final Uri url = Uri.parse(urlString.trimLeft());
+    isWebURL = url.scheme == 'http' || url.scheme == 'https';
+  } on FormatException catch (_) {}
```
stuartmorgan, line 78: Actually, you can just use `Uri.tryParse` instead of this, and simplify the code. I forgot to see if there was a `try` version of this API before my last review comments.

```
  final Uri? url = Uri.tryParse(urlString.trimLeft());
  final bool isWebURL = url != null && (url.scheme == 'http' || url.scheme == 'https');
```

packages/url_launcher/url_launcher/test/url_launcher_test.dart
```diff
@@ -281,6 +281,42 @@ void main() {
       await launchResult;
       expect(binding.renderView.automaticSystemUiAdjustment, true);
     });
+
+    test('open remote desktop url', () async {
```
stuartmorgan, line 285: Let's call this 'open non-parseable URL' so it's clear what the real purpose of this test is.

packages/url_launcher/url_launcher/test/url_launcher_test.dart
```diff
@@ -281,6 +281,42 @@ void main() {
...
+          webOnlyWindowName: null,
+        )
+        ..setResponse(true);
+      expect(
+          await launch(
+              'rdp://full%20address=s:mypc:3389&audiomode=i:2&disable%20themes=i:1'),
+          isTrue);
+    });
+
+    test('cannot open remote desktop url with forceSafariVC: true', () async {
```
stuartmorgan, line 305: Same here and below; s/remove desktop/non-parseable/

stuartmorgan (changes requested):
Small nits, but otherwise looks good!

stuartmorgan (approved):
LGTM, thanks!

PR #4366, opened 2021-09-21
[webview_flutter] Migrate main package to fully federated architecture.

mvanbeusekom:
This PR migrates the main "app facing" webview_flutter package to depend on the new `webview_flutter_platform_interface`, `webview_flutter_android` and `webview_flutter_wkwebview` packages. This finalises the migration to the fully federated architecture as described in issue flutter/flutter#86286.

Relevant issue:
 - Resolves flutter/flutter#86286

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan (commented):
Code changes LG, but the iOS test failure looks real since it's across both channels. Any idea what's going on there?

mvanbeusekom:
> Code changes LG, but the iOS test failure looks real since it's across both channels. Any idea what's going on there?

That was indeed a mistake from my side. The `SurfaceAndroidWebView` integration tests were also run on iOS because I forgot to switch the `_skipDueToIssue86757` back to `true` (I had set it to `false` to run tests locally). I have made skipping these tests a bit more robuust and added an additional condition so they only run when `Platform.isAndroid` is `true`.

stuartmorgan (approved):
LGTM

packages/webview_flutter/webview_flutter/lib/platform_interface.dart
```diff
@@ -2,547 +2,23 @@
...
+export 'package:webview_flutter_platform_interface/webview_flutter_platform_interface.dart'
+    show
+        AutoMediaPlaybackPolicy,
+        CreationParams,
+        JavascriptChannel,
+        JavascriptChannelRegistry,
+        JavascriptMessage,
+        JavascriptMode,
+        JavascriptMessageHandler,
+        WebViewPlatform,
```
stuartmorgan, line 17: For posterity: we accidentally shipped a breaking change to external implementors here without noticing. There's nothing we can reasonably do about it at this point, and AFAIK it only affected one client (now fixed), but what happened was:
- https://github.com/flutter/plugins/pull/4302/files#diff-bbee1da5da193a072dca86d16467e75802852826099a7b0db2e6c497c71a3f31R54 changed `WebViewPlatform`. In isolation that change seemed reasonable because it was a new package, so wasn't a breaking change.
- This switched from the old local `WebViewPlatform` to the new one, which has a new argument to `build`, which is breaking for any implementations of `WebViewPlatform`. The fact that it wasn't a direct copy isn't apparent here, so we forgot to consider that.

Because in this repository we previously had only the one combined package, and we were thus not eating our own dogfood in an important way with respect to the platform interface, nothing in CI alerted us to the mistake. The good news is that being fully federated gives us much more protection against making that mistake in the future (which is validation of the need for us to follow the intended policy of having all of our federated plugins be fully federated precisely so that we *are* eating our own dogfood).

So ironically the change that will save us from this bug in the future caused exactly the bug it's meant to avoid. But the bug was one-time, and the protection is persistent for all future changes, so it's definitely a net win despite that.

PR #4369, opened 2021-09-21
[flutter_plugin_tools] Allow overriding breaking change check

stuartmorgan:
Adds the ability to override the current prohibition against breaking changes to platform interfaces with a justification (following a set pattern, which will be documented in the wiki) in the PR description. This fixes the incorrect hard limit that's currently in place, but ensures that there is enough extra process that:
- PR authors won't be able to change major versions without realizing that it's strongly discouraged and prominently explaining why they believe it's necessary, and
- PR reviewers are much less likely to miss in review that a strongly discouraged change is being made.

Fixes https://github.com/flutter/flutter/issues/85391

## Pre-launch Checklist

- [x] I signed the [CLA].


script/tool/lib/src/version_check_command.dart
```diff
@@ -292,16 +315,14 @@ ${indentation}HTTP response: ${pubVersionFinderResponse.httpResponse.body}
...
-    // TODO(stuartmorgan): Relax this check. See
-    // https://github.com/flutter/flutter/issues/85391
-    if (isPlatformInterface &&
-        allowedNextVersions[currentVersion] == NextVersionType.BREAKING_MAJOR) {
+    if (allowedNextVersions[currentVersion] == NextVersionType.BREAKING_MAJOR &&
+        !_validateBreakingChange(package)) {
       printError('${indentation}Breaking change detected.\n'
-          '${indentation}Breaking changes to platform interfaces are strongly discouraged.\n');
+          '${indentation}Breaking changes to platform interfaces are not '
+          'allowed without explicit justification.\n');
```
ditman, line 322: Maybe add a link to some documentation here, so this error message is more "self-service" for contributors?

script/tool/lib/src/version_check_command.dart
```diff
@@ -398,4 +419,40 @@ ${indentation}The first version listed in CHANGELOG.md is $fromChangeLog.
...
+  }
+
+  /// Returns the contents of the file pointed to by [_changeDescriptionFile],
+  /// or an empty string if that flag is not provided.
+  String _getChangeDescription() {
+    final String path = getStringArg(_changeDescriptionFile);
+    if (path.isEmpty) {
+      return '';
+    }
+    return packagesDir.fileSystem.file(path).readAsStringSync();
```
ditman, line 456: Check if file exists and `return '';` if it doesn't?

ditman (approved):
LGTM!

I would have loved to be able to override even more the version number (to jump from 0.2.0 to 0.6.0 :P) but this is more to the point, and keeps the processes saner. 🚀

stuartmorgan:
> I would have loved to be able to override even more the version number (to jump from 0.2.0 to 0.6.0 :P) but this is more to the point, and keeps the processes saner. 🚀

I was keeping that use case in mind while writing this; it would be relatively straightforward to extend this to cover that if we find another time where we feel that we need it.

script/tool/lib/src/version_check_command.dart
stuartmorgan: I actually didn't do that deliberately; if someone thinks they are passing in a description but has, e.g., typo'd the path, they should get an error rather than have it just silently not work.

I'll make it an explicit ToolExit though, so that it's obvious.

script/tool/lib/src/version_check_command.dart
stuartmorgan: Yep, makes sense. I'll link to the contribution wiki page, and describe this there.

script/tool/lib/src/version_check_command.dart
```diff
@@ -453,6 +458,11 @@ ${indentation}The first version listed in CHANGELOG.md is $fromChangeLog.
     if (path.isEmpty) {
       return '';
     }
-    return packagesDir.fileSystem.file(path).readAsStringSync();
+    final File file = packagesDir.fileSystem.file(path);
+    if (!file.existsSync()) {
+      printError('${indentation}No such file: $path');
+      throw ToolExit(_exitMissingChangeDescriptionFile);
```
ditman, line 464: Yeah, since this file (and its path) are managed by the cirrus.yaml, it makes more sense that this is a tool exit error, good call!

PR #4364, opened 2021-09-20
[in_app_purchase] Ensure the `introductoryPriceMicros` field is populated correctly.

mvanbeusekom:
Due to a typo the `SkuDetailsWrapper.introductoryPriceMicros` field wasn't populated correctly with the values send from the native side. 

The first [commit](https://github.com/flutter/plugins/commit/1b1ec27222107eee976faae18476bf088502b1d0) of this PR updates the existing unit tests to prove the existence of the issue. The second [commit](https://github.com/flutter/plugins/commit/77e2c6ddaae95600b1b93f9d51350fb155a045f9) fixes the implementation of the `SkuDetailsWrapper` class so the tests succeed and populate the field correctly.

Fixes flutter/flutter#90244

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Note that unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I [updated pubspec.yaml](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates) with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[...]

PR #4370, opened 2021-09-22
[in_app_purchase] Ensure the introductoryPriceMicros field is transported as a String.

mvanbeusekom:
In the current version the `introductoryPriceMicros` field is transported from JAVA to Dart as a `long` type. This is causing a runtime error on the Dart side as it is expected to be of type `String` (see comment [here](https://github.com/flutter/plugins/pull/4364#issuecomment-924116386)). 

This PR will resolve the problem by making sure the `introductoryPriceMicros` field is converted to a `String` type before it is transported to Dart. 

Ideally we would change the Dart API to expect the field to be a `int`, however this would introduce a breaking change. Therefore I opted for the conversion from `long` to `String` on the JAVA side.

Fixes the problem discussed here: https://github.com/flutter/plugins/pull/4364#issuecomment-924116386

## Pre-launch Checklist

- [x] I signed the [CLA].


781flyingdutchman:
I understand the desire to return a String to avoid a breaking change, but the `priceAmountMicros` and `originalPriceAmountMicros` fields are both type `int` so having `introductoryPriceAmountMicros` now suddenly be a String is very inconsistent.  My understanding is that the `introductoryPriceAmountMicros` field was never populated correctly due to the wrong key being used to extract it from JSON, so couldn't we argue that the field effectively wasn't supported until the commit from two days ago (which turned out to not work)?  In that case I would suggest changing the field to be an `int` and thus keeping it consistent with the other `...priceAmountMicros` fields.  If there ever was a moment to make this change, it is now, before the field is actually being used.

stuartmorgan:
> My understanding is that the `introductoryPriceAmountMicros` field was never populated correctly due to the wrong key being used to extract it from JSON, so couldn't we argue that the field effectively wasn't supported until the commit from two days ago (which turned out to not work)?

There's a difference between code that compiles but returned nothing at runtime and code that doesn't compile. It appears from blame that people have had four months, not two days, to write code that would fail to build if this changed. It's a breaking change.

@mvanbeusekom This isn't forwarded through the cross-platform API, right; i.e., someone must directly use `in_app_purchase_android` to interact with this? If so, then can't we do a breaking change to just the android package, and pick that up in a minor update to the app-facing package?

(We should still do a bugfix in a bugfix update first though, so that the last version people will get that's resolvable with existing dependencies isn't actively broken.)

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
```diff
@@ -26,7 +26,10 @@
     info.put("description", detail.getDescription());
     info.put("freeTrialPeriod", detail.getFreeTrialPeriod());
     info.put("introductoryPrice", detail.getIntroductoryPrice());
-    info.put("introductoryPriceAmountMicros", detail.getIntroductoryPriceAmountMicros());
+    // Make sure the `introductoryPriceAmountMicros` field is returned as a `String` value as the
+    // Dart side expects it as `String`. Changing the type in Dart would introduce a breaking change.
+    info.put(
+        "introductoryPriceAmountMicros", String.valueOf(detail.getIntroductoryPriceAmountMicros()));
```
stuartmorgan, line 32: Changing this on the Java side will make a better fix later harder to do. Can't we change the transport name, and have a String getter on the dart side that returns the underlying int value as a String?

781flyingdutchman: @stuartmorgan I understand changing to int would technically be a breaking change, but only for people who read and use the field - and that field would *always* have returned an empty string. I therefore expect the number of builds that break to be very very small.

Is a compromise to change the field to an `int` and also add a getter `introductoryPriceAmountMicrosAsString` that does the conversion so the very few people who read the original field and expect a string can simply change their field to the new getter?

stuartmorgan: > @stuartmorgan I understand changing to int would technically be a breaking change, but only for people who read and use the field 

I'm well aware that it's only a breaking change for people who reference the field; please re-read my comment.

> I therefore expect the number of builds that break to be very very small.

That may be true (or may not; you don't really know how many people did things like try to use it, find that it didn't seem to work for them, and added a fallback but left the code in place), but the definition of a non-breaking change in semantic versioning is not "I don't think it'll break many people".

> Is a compromise to change the field to an `int` and also add a getter `introductoryPriceAmountMicrosAsString` that does the conversion so the very few people who read the original field and expect a string can simply change their field to the new getter?

No, because that's a breaking change.

There is no way to change the type of the existing getter without it being a breaking change; this is not a matter of discussion, it's an objective fact. And we're not going to make an obviously breaking change without adjusting the version accordingly, because we do not deliberately violate semantic versioning in our plugins.

781flyingdutchman: Got it - didn't mean to come across as pushing, was just trying to help. I have absolutely no skin in this game (I came across the original error, I don't use the field at all).

If I may make one comment, it would be that the impact of the original change (that caused the error) was quite severe, in that it rendered the entire package unusable for Android users. Ideally, this would have been caught by tests before bumping the version. I am saying this because I had the same thing happen with the wakelock package just a few days ago, and if you have a flutter app with many package dependencies then these bugs as a result of version updates are very disruptive and difficult to discover and trace back to the offending package.  It leads me to consider locking packages to a specific version (instead of using the caret notation to benefit from upgrades), and that is not a good developer experience.

stuartmorgan: > If I may make one comment, it would be that the impact of the original change (that caused the error) was quite severe, in that it rendered the entire package unusable for Android users. Ideally, this would have been caught by tests before bumping the version.

Yes, it's not clear to me yet why that PR passed our automated tests (of which we have quite a few). That's something we'll be looking into.

> I am saying this because I had the same thing happen with the wakelock package just a few days ago

That was a third-party package, and as far as I can tell from the nature of that bug, that plugin:
- has no automated testing of any kind, and
- was never launched on iOS after the change

We cannot control what third-party package authors do, but it was a very different scenario. That bug could not have passed our CI.

cyanglaz:
I would suggest the following (cooperating with what @stuartmorgan has commented):
1. Introduce a new field in dart: `final int introductoryPriceAmountMicros`, which gets the value directly from java side, and also has the same name as in java.
2. Mark introductoryPriceMicros as deprecated, make it return the string value of `introductoryPriceAmountMicros`
3. This will be a minor bump. 

Then open another PR, making a breaking change (major version bump). This PR removes the deprecated `introductoryPriceMicros` field.

As for testing, I'm not aware that we have anything specific setup to make sure method channels work end to end (I might have missed something?). In iOS, we have [StoreKit Tests](https://developer.apple.com/documentation/storekittest?language=objc), which is a UITests. It does cover the end to end case. However, StoreKit Test equivalent does not exist in Play Billing Library just yet.

mvanbeusekom:
@stuartmorgan, @cyanglaz, I have updated the PR to introduce the `introductoryPriceAmountMicros` field using the correct type and deprecating the `introductoryPriceMicros` field which used the wrong type. I would appreciate it if you could have another look.

packages/in_app_purchase/in_app_purchase_android/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
-## 0.1.4+7
```
stuartmorgan, line 1: This was published; it shouldn't be removed from the changelog.

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/sku_details_wrapper.dart
```diff
@@ -68,8 +72,17 @@ class SkuDetailsWrapper {
...
+  final int introductoryPriceAmountMicros;
+
+  /// String representation of [introductoryPrice] in micro-units 990000
+  @Deprecated('Use `introductoryPriceAmountMicros` instead.')
+  @JsonKey(ignore: true)
+  String get introductoryPriceMicros => _introductoryPriceMicros.isEmpty
+      ? introductoryPriceAmountMicros != 0
+          ? introductoryPriceAmountMicros.toString()
+          : ''
+      : _introductoryPriceMicros;
```
stuartmorgan, line 86: Needs a no-op, deprecated setter to avoid being a breaking change, per discussion.

packages/in_app_purchase/in_app_purchase_android/test/billing_client_wrappers/sku_details_wrapper_deprecated_test.dart
```diff
@@ -0,0 +1,68 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:in_app_purchase_android/billing_client_wrappers.dart';
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
stuartmorgan, line 5: This should be before imports (apparently the license check could be a little stricter).

packages/in_app_purchase/in_app_purchase_android/test/billing_client_wrappers/sku_details_wrapper_deprecated_test.dart
```diff
@@ -0,0 +1,68 @@
...
+      originalPrice: 'originalPrice',
+      originalPriceAmountMicros: 1000,
+    );
+
+    expect(skuDetails, isNotNull);
+    expect(skuDetails.introductoryPriceAmountMicros, 990000);
+    // ignore: deprecated_member_use_from_same_package
+    expect(skuDetails.introductoryPriceMicros, '990000');
+  });
+}
```
stuartmorgan, line 68: Let's add a setter test to cover the missed breaking change. It's fine if it's testing that it's a no-op, it just needs to build.

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/sku_details_wrapper.dart
```diff
@@ -32,7 +32,9 @@ class SkuDetailsWrapper {
     required this.description,
     required this.freeTrialPeriod,
     required this.introductoryPrice,
-    required this.introductoryPriceMicros,
+    @Deprecated('Use `introductoryPriceAmountMicros` parameter instead')
+        String introductoryPriceMicros = '',
+    this.introductoryPriceAmountMicros = 0,
```
stuartmorgan, line 37: I think the better option here is to make this `int introductoryPriceAmountMicros,`, and have the constructor body do:
```
introductoryPriceAmountMicros = introductoryPriceAmountMicros ?? int.tryParse(introductoryPriceMicros) ?? 0
```

Then you don't need all the complicated logic in the getter, or `_introductoryPriceMicros`.

stuartmorgan (changes requested):

packages/in_app_purchase/in_app_purchase_android/test/billing_client_wrappers/sku_details_wrapper_deprecated_test.dart
mvanbeusekom: As per previous comment, I don't think setter is necessary after all.

packages/in_app_purchase/in_app_purchase_android/CHANGELOG.md
mvanbeusekom: You are right and fixed it.

packages/in_app_purchase/in_app_purchase_android/test/billing_client_wrappers/sku_details_wrapper_deprecated_test.dart
mvanbeusekom: Done.

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/sku_details_wrapper.dart
```diff
@@ -68,8 +72,17 @@ class SkuDetailsWrapper {
...
-  @JsonKey(name: 'introductoryPriceAmountMicros', defaultValue: '')
-  final String introductoryPriceMicros;
+  @JsonKey(name: 'introductoryPriceAmountMicros', defaultValue: 0)
+  final int introductoryPriceAmountMicros;
+
+  /// String representation of [introductoryPrice] in micro-units 990000
+  @Deprecated('Use `introductoryPriceAmountMicros` instead.')
+  @JsonKey(ignore: true)
+  String get introductoryPriceMicros => _introductoryPriceMicros.isEmpty
+      ? introductoryPriceAmountMicros != 0
```
cyanglaz, line 82: I didn't get why we needed the nested part. If introductoryPriceAmountMicros is 0, can't introductoryPriceMicros just be '0'?
Wouldn't the below work?
```dart
String get introductoryPriceMicros => _introductoryPriceMicros.isEmpty ? 
                                                                  introductoryPriceAmountMicros.toString() :
                                                                  ''
```

mvanbeusekom: Yes it would work, but the original behavior (although wrong) was to return an empty string as default value. I tried to stick as close as possible to that using the nested condition.

cyanglaz: If returning empty string is wrong, we don't need to worry about the wrong behavior. This PR suppose to "fix" it :)

mvanbeusekom: Yes that makes sense. In that case the inner condition can indeed be removed.

mvanbeusekom: Done.

cyanglaz (approved):
LGTM, let's wait for @stuartmorgan's approval on the things he commented.

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/sku_details_wrapper.dart
```diff
@@ -68,8 +72,15 @@ class SkuDetailsWrapper {
   final String introductoryPrice;
 
   /// [introductoryPrice] in micro-units 990000
```
cyanglaz, line 74: Can we add a new comment saying something similar to the android doc:
/// Returns 0 if the SKU is not a subscription or doesn't have an introductory period.

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/sku_details_wrapper.dart
stuartmorgan: Oh, right; good call.

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/sku_details_wrapper.dart
stuartmorgan: We'll be changing it to required at that point presumably to match everything else, and it will already be a breaking change in general, so that doesn't seem like a significant issue. But it's not a big deal either way.

stuartmorgan (approved):
LGTM

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/sku_details_wrapper.dart
mvanbeusekom: Done, good catch. I think that is a valuable addition.

PR #4383, opened 2021-09-24
[webview_flutter] Adjust integration test domains

stuartmorgan:
Tests that load https://www.google.com and then assert that that's the
current URL are failing, because Google is redirecting our tests to a
"suspicious activity" page. Switch to example.com, as a domain that is
less likely to do traffic-based redirection.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
+++ b/packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
@@ -58,9 +58,9 @@ void main() {
       ),
     );
     final WebViewController controller = await controllerCompleter.future;
-    await controller.loadUrl('https://www.google.com/');
+    await controller.loadUrl('https://www.example.com/');
     final String? currentUrl = await controller.currentUrl();
-    expect(currentUrl, 'https://www.google.com/');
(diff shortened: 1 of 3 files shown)


packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
```diff
@@ -58,9 +58,9 @@ void main() {
       ),
     );
     final WebViewController controller = await controllerCompleter.future;
-    await controller.loadUrl('https://www.google.com/');
+    await controller.loadUrl('https://www.example.com/');
```
ditman, line 61: Should this URL be a `const String` at the top of the test files, so it can be changed more easily in the future?

ditman (approved):

packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
stuartmorgan: Since there's no actual relation between its use in the different tests (i.e., we could make it different in every test with no issue) I would lean away from extracting a shared constant. Maybe within each test, but that doesn't change things much.

stuartmorgan:

PR #4389, opened 2021-09-27
[WIP] [image_picker_for_web] Added support for maxWidth, maxHeight and imageQuality 

balvinderz:
Adds support for maxWidth, maxHeight and imageQuality in image_picker_for_web

https://github.com/flutter/flutter/issues/78170

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
+++ b/packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
@@ -121,6 +121,9 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     List<XFile> files = await getFiles(
       accept: _kAcceptImageMimeType,
       capture: capture,
+      maxWidth: maxWidth,
+      maxHeight: maxHeight,
+      imageQuality: imageQuality,
     );
     return files.first;
   }
@@ -158,7 +161,12 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     double? maxHeight,
     int? imageQuality,
   }) {
-    return getFiles(accept: _kAcceptImageMimeType, multiple: true);
+    return getFiles(
+        accept: _kAcceptImageMimeType,
+        multiple: true,
+        imageQuality: imageQuality,
+        maxHeight: maxHeight,
+        maxWidth: maxWidth);
   }
 
   /// Injects a file input with the specified accept+capture attributes, and
@@ -171,11 +179,13 @@ class ImagePickerPlugin extends ImagePickerPlatform {
   ///
   /// See https://caniuse.com/#feat=html-media-capture
   @visibleForTesting
-  Future<List<XFile>> getFiles({
-    String? accept,
-    String? capture,
-    bool multiple = false,
-  }) {
+  Future<List<XFile>> getFiles(
+      {String? accept,
+      String? capture,
+      bool multiple = false,
+      double? maxWidth,
+      double? maxHeight,
+      int? imageQuality}) {
     html.FileUploadInputElement input = createInputElement(
       accept,
       capture,
@@ -183,7 +193,8 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     ) as html.FileUploadInputElement;
     _injectAndActivate(input);
 
-    return _getSelectedXFiles(input);
+    return _getSelectedXFiles(input,
+        maxWidth: maxWidth, maxHeight: maxHeight, imageQuality: imageQuality);
   }
 
   // DOM methods
@@ -238,23 +249,29 @@ class ImagePickerPlugin extends ImagePickerPlatform {
   }
 
   /// Monitors an <input type="file"> and returns the selected file(s).
-  Future<List<XFile>> _getSelectedXFiles(html.FileUploadInputElement input) {
+  Future<List<XFile>> _getSelectedXFiles(html.FileUploadInputElement input,
+      {double? maxWidth, double? maxHeight, int? imageQuality}) {
     final Completer<List<XFile>> _completer = Completer<List<XFile>>();
     // Observe the input until we can return something
-    input.onChange.first.then((event) {
+    input.onChange.first.then((event) async {
       final files = _handleOnChangeEvent(event);
       if (!_completer.isCompleted && files != null) {
-        _completer.complete(files
-            .map((file) => XFile(
-                  html.Url.createObjectUrl(file),
-                  name: file.name,
-                  length: file.size,
-                  lastModified: DateTime.fromMillisecondsSinceEpoch(
-                    file.lastModified ?? DateTime.now().millisecondsSinceEpoch,
-                  ),
-                  mimeType: file.type,
-                ))
-            .toList());
+        //Convert the file  if one of the parameter is not null
+        if (imageQuality != null || maxWidth != null || maxHeight != null) {
+          final convertedFileUris = await Future.wait(files.map((e) async =>
+              await _getCompressedUri(e, imageQuality, maxHeight, maxWidth)));
+
+          int index = 0;
+          _completer.complete(files.map((file) {
+            index += 1;
+            final xFile = file.toXFile(path: convertedFileUris[index - 1]);
+            return xFile;
+          }).toList());
+        } else {
+          _completer.complete(files.map((file) {
+            return file.toXFile();
+          }).toList());
+        }
       }
     });
     input.onError.first.then((event) {
@@ -310,6 +327,72 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     _target.children.add(element);
     element.click();
   }
+
+  Future<String> _getCompressedUri(html.File file, int? imageQuality,
+      double? maxHeight, double? maxWidth) async {
+    //max width, max height, imageQuality are not supported for gif
+    if (file.type == "image/gif") {
+      return html.Url.createObjectUrl(file);
+    }
+
+    final Completer<String> _completer = Completer<String>();
+    final blobUrl = html.Url.createObjectUrl(file);
+    final image = html.ImageElement(src: blobUrl);
+    image.onLoad.listen((event) async {
+      html.Url.revokeObjectUrl(blobUrl);
+      final canvas = html.CanvasElement();
+      final size = _calculateSize(image, maxWidth ?? image.width!.toDouble(),
+          maxHeight ?? image.height!.toDouble());
+      canvas.width = size[0];
+      canvas.height = size[1];
+      final ctx = canvas.context2D;
+      if (maxHeight == null && maxWidth == null) {
+        ctx.drawImage(image, 0, 0);
+      } else {
+        ctx.drawImageScaled(image, 0, 0, canvas.width!, canvas.height!);
+      }
+      final blob =
+          await canvas.toBlob(file.type, (imageQuality ?? 100) / 100.0); // Image quality only works for jpeg images
+      _completer.complete(html.Url.createObjectUrlFromBlob(blob));
+    });
+    image.onError.listen((event) {
+      //Return the original image if error comes
+      _completer.complete(html.Url.createObjectUrl(file));
+    });
+    return _completer.future;
+  }
+
+  List<int> _calculateSize(
+      html.ImageElement img, double maxWidth, double maxHeight) {
+    var width = img.width!;
+    var height = img.height!;
+
+    // calculate the width and height, constraining the proportions
+    if (width > height) {
+      if (width > maxWidth) {
+        height = ((height * maxWidth) / width).round();
+        width = maxWidth.toInt();
+      }
+    } else {
+      if (height > maxHeight) {
+        width = ((width * maxHeight) / height).round();
+        height = maxHeight.toInt();
+      }
+    }
+    return [width, height];
+  }
+}
+
+extension on html.File {
+  XFile toXFile({String? path}) => XFile(
+        path ?? html.Url.createObjectUrl(this),
+        name: this.name,
+        length: this.size,
+        lastModified: DateTime.fromMillisecondsSinceEpoch(
+          this.lastModified ?? DateTime.now().millisecondsSinceEpoch,
+        ),
+        mimeType: this.type,
+      );
 }
 
 // Some tools to override behavior for unit-testing
--- a/packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
+++ b/packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:html' as html;
 
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
+import 'package:image/image.dart';
 import 'package:meta/meta.dart';
 import 'package:image_picker_platform_interface/image_picker_platform_interface.dart';
 
@@ -334,7 +335,6 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     if (file.type == "image/gif") {
       return html.Url.createObjectUrl(file);
     }
-
     final Completer<String> _completer = Completer<String>();
     final blobUrl = html.Url.createObjectUrl(file);
     final image = html.ImageElement(src: blobUrl);
--- a/packages/image_picker/image_picker/pubspec.yaml
+++ b/packages/image_picker/image_picker/pubspec.yaml
@@ -24,7 +24,8 @@ dependencies:
   flutter:
     sdk: flutter
   flutter_plugin_android_lifecycle: ^2.0.1
-  image_picker_for_web: ^2.1.0
+  image_picker_for_web:
+    path: ../image_picker_for_web
   image_picker_platform_interface: ^2.3.0
 
 dev_dependencies:
--- a/packages/image_picker/image_picker_for_web/CHANGELOG.md
+++ b/packages/image_picker/image_picker_for_web/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.1.4
+
+* Implemented `maxWidth`, `maxHeight` and `imageQuality`
+
 ## 2.1.3
 
 * Add `implements` to pubspec.
--- a/packages/image_picker/image_picker_for_web/pubspec.yaml
+++ b/packages/image_picker/image_picker_for_web/pubspec.yaml
@@ -2,7 +2,7 @@ name: image_picker_for_web
 description: Web platform implementation of image_picker
 repository: https://github.com/flutter/plugins/tree/master/packages/image_picker/image_picker_for_web
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+image_picker%22
-version: 2.1.3
+version: 2.1.4
 
 environment:
   sdk: ">=2.12.0 <3.0.0"


balvinderz:
Implementation is based on this : https://labs.madisoft.it/javascript-image-compression-and-resizing/

There are some limitations in this.
imageQuality only works for jpeg images.
I could not find a way to make this work for gif images.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -121,6 +121,9 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     List<XFile> files = await getFiles(
       accept: _kAcceptImageMimeType,
       capture: capture,
+      maxWidth: maxWidth,
+      maxHeight: maxHeight,
+      imageQuality: imageQuality,
```
ditman, line 126: Please, do not pass the image quality parameters to the getFiles method, otherwise you have to drill those down through several layers of the plugin.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -121,6 +121,9 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
     );
     return files.first;
```
ditman, line 128: Instead, determine if resizing is needed, and then `return imageResizer.resizeImage(files.first, maxWidth, maxHeight, imageQuality);`

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -158,7 +161,12 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     double? maxHeight,
     int? imageQuality,
   }) {
-    return getFiles(accept: _kAcceptImageMimeType, multiple: true);
```
ditman, line 161: Same here, determine if resizing is needed, and then return imageResizer.resizeImages(getFiles(...), maxWidth, maxHeight, imageQuality);

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -171,19 +179,22 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
-    String? capture,
-    bool multiple = false,
-  }) {
+  Future<List<XFile>> getFiles(
+      {String? accept,
+      String? capture,
+      bool multiple = false,
+      double? maxWidth,
+      double? maxHeight,
+      int? imageQuality}) {
```
ditman, line 188: (Do not accept the resize image parameters any further down than the getImage/getImages method)

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -238,23 +249,29 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
-                  lastModified: DateTime.fromMillisecondsSinceEpoch(
-                    file.lastModified ?? DateTime.now().millisecondsSinceEpoch,
-                  ),
-                  mimeType: file.type,
-                ))
-            .toList());
+        //Convert the file  if one of the parameter is not null
+        if (imageQuality != null || maxWidth != null || maxHeight != null) {
+          final convertedFileUris = await Future.wait(files.map((e) async =>
+              await _getCompressedUri(e, imageQuality, maxHeight, maxWidth)));
```
ditman, line 262: This `_getCompressedUri` method should be a method from a separate object that only deals with resizing images in `XFile` objects.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -310,6 +327,73 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     _target.children.add(element);
     element.click();
   }
+
+  Future<String> _getCompressedUri(html.File file, int? imageQuality,
```
ditman, line 331: Extract to a new `ImageResizer` object. Also, try to use: Future<XFile> as the input format for the file, and the output for the resized image.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -310,6 +327,73 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
+    });
+    image.onError.listen((event) {
+      //Return the original image if error comes
+      _completer.complete(html.Url.createObjectUrl(file));
+    });
+    return _completer.future;
+  }
+
+  List<int> _calculateSize(
+      html.ImageElement img, double maxWidth, double maxHeight) {
```
ditman, line 367: Move this helper to the `ImageResizer` object as well. Also, maybe return a `Size` object [docs](https://api.flutter.dev/flutter/dart-ui/Size-class.html), instead of a list of ints?

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -310,6 +327,73 @@ class ImagePickerPlugin extends ImagePickerPlatform {
...
+extension on html.File {
+  XFile toXFile({String? path}) => XFile(
+        path ?? html.Url.createObjectUrl(this),
+        name: this.name,
+        length: this.size,
+        lastModified: DateTime.fromMillisecondsSinceEpoch(
+          this.lastModified ?? DateTime.now().millisecondsSinceEpoch,
+        ),
+        mimeType: this.type,
+      );
```
ditman, line 396: Let's not apply this extension on a dart:html class. This will cause people to some times have a `.toXFile` method on their `html.File`, and some times **not** have it, and the cause is whether they use the image_picker_for_web or not.

(This might be more appropriate for the cross_file package, but I think the best solution is still a XFile constructor for web only that takes an html.File input file.)

ditman (changes requested):
Thanks for the contribution @balvinderz! I think that this one is a little bit too green, and it needs some changes to the approach to make this more modular and testable!

In my opinion, the image resizing code should be extracted from the `ImagePickerPlugin` class completely. The only thing the plugin knows about resizing images is that it has an instance of an `ImageResizer` object that accepts and returns `XFile` instances.

I think the following steps are needed:

- [ ] Create a `ImageResizer` component that deals with images in `XFile` instances, and resizes them according to the parameters passed by the user.
  - 99% of the code of this component already exists in the `_getCompressedUri` and `_calculateSize` methods. Having your own object will allow you to design it more carefully. 
- [ ] Have an instance of that `ImageResizer` inside of the `ImagePickerPlugin`, so it can be:
  - Used when needed to resize images (normal operation)
  - Injectable from tests so it can be mocked out if needed (testability)
- [ ] Use the `ImageResizer` in the `pickImage`/`pickImages` methods, if the user wants a resized output
  - **Do not pass the image resizing parameters (`maxWidth`, `maxHeight`, `imageQuality`) to the generic methods that pick files and convert them to a XFile!**
- [ ] Unit tests for the new `ImageResizer` component
- [ ] Unit tests for the new resize images (when the correct parameters are passed) feature.

balvinderz:
I think i have 

> Thanks for the contribution @balvinderz! I think that this one is a little bit too green, and it needs some changes to the approach to make this more modular and testable!
> 
> In my opinion, the image resizing code should be extracted from the `ImagePickerPlugin` class completely. The only thing the plugin knows about resizing images is that it has an instance of an `ImageResizer` object that accepts and returns `XFile` instances.
> 
> I think the following steps are needed:
> 
> * [X]  Create a `ImageResizer` component that deals with images in `XFile` instances, and resizes them according to the parameters passed by the user.
>   
>   * 99% of the code of this component already exists in the `_getCompressedUri` and `_calculateSize` methods. Having your own object will allow you to design it more carefully.
> * [X]  Have an instance of that `ImageResizer` inside of the `ImagePickerPlugin`, so it can be:
>   
>   * Used when needed to resize images (normal operation)
>   * Injectable from tests so it can be mocked out if needed (testability)
> *  [X]  Use the `ImageResizer` in the `pickImage`/`pickImages` methods, if the user wants a resized output
>   
>   * **Do not pass the image resizing parameters (`maxWidth`, `maxHeight`, `imageQuality`) to the generic methods that pick files and convert them to a XFile!**
> * [X]  Unit tests for the new `ImageResizer` component
> * [X]  Unit tests for the new resize images (when the correct parameters are passed) feature.

I think I have done all the steps that are mentioned in this.

ditman:
> I think I have done all the steps that are mentioned in this.

Amazing, thanks! I'll take a look later tonight!

balvinderz:
I have found a way to scale gif images using package:image. 
https://pub.dev/packages/image
But it is way too slow and freezes the ui.

packages/image_picker/image_picker_for_web/lib/image_resizer_utils.dart
```diff
@@ -0,0 +1,73 @@
...
+bool isImageQualityValid(int imageQuality) {
+  return (imageQuality >= 0 && imageQuality <= 100);
+}
+
+/// a functions that calculates the size of the scaled image.
+/// imageWidth is the width of the image
+/// imageHeight is the height of  the image
+/// maxWidth is the maximum width of the scaled image
+/// maxHeight is the maximum height of the scaled image
+Size calculateSizeOfScaledImage(double imageWidth, double imageHeight,
```
ditman, line 33: If you pass `Size imageSize` in, you can use the tools in the Size object to do the resizing (see [operator *](https://api.flutter.dev/flutter/dart-ui/Size/operator_multiply.html))

packages/image_picker/image_picker_for_web/lib/image_resizer_utils.dart
```diff
@@ -0,0 +1,73 @@
...
+    double? maxWidth, double? maxHeight) {
```
ditman, line 34: What is the behavior here when both maxWidth and maxHeight are passed? Constrain to fit? Stretch to fit?

ditman: If we never stretch, but constrain to the smallest side maitaining aspect ratio, the code can be simplified a lot, I think!

packages/image_picker/image_picker_for_web/lib/image_resizer_utils.dart
```diff
@@ -0,0 +1,73 @@
...
+  double originalWidth = imageWidth;
+  double originalHeight = imageHeight;
+
+  bool hasMaxWidth = maxWidth != null;
```
ditman, line 38: If we're not stretching the output of the image, I think most of this method can be simplified a little bit:

```dart
Size calculateSizeOfScaledImage(Size imageSize, double? maxWidth, double? maxHeight) {
  double widthFactor = maxWidth != null ? imageSize.width / maxWidth : 0;
  double heightFactor = maxHeight != null ? imageSize.height / maxHeight : 0;
  double resizeFactor = Math.max(widthFactor, heightFactor);
  return imageSize ~/ resizeFactor;
}
```

ditman (commented):
I have some general comments. I think the image_resizer_utils can be revisited with much simpler algorithms (note that it's only Github code, so I haven't tested that my suggestions have 100% perfect syntax or cover all the corner cases that your code might do!), please take them as inspiration to cut down the amount of LOC, if possible! ;)

packages/image_picker/image_picker_for_web/lib/image_resizer_utils.dart
```diff
@@ -0,0 +1,73 @@
...
+  if (resizeNeeded) {
+    if (imageQuality != null) {
+      return isImageQualityValid(imageQuality);
+    } else {
+      return true;
+    }
+  } else {
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
balvinderz: yes

packages/image_picker/image_picker_for_web/lib/image_resizer_utils.dart
balvinderz: We are never stretching. My implementation of this function was based on how  the ios side of this plugin scales the image.

packages/image_picker/image_picker_for_web/lib/image_resizer_utils.dart
balvinderz: This does work but it scales the image above its original width and height . I think I should have used the word downScaled instead of scaled in this function .  The android and IOS implementation of this function always downScales the  image and never upscales the image.  so I think, I will keep the same implementation as the others to be consistent in all platforms . What do you think?

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
```diff
@@ -0,0 +1,63 @@
...
+import 'dart:math';
+import 'package:image_picker_for_web/image_resizer_utils.dart';
+import 'package:image_picker_platform_interface/image_picker_platform_interface.dart';
+import 'dart:html' as html;
+
+/// Helper class that resizes image
+class ImageResizer {
+  /// Resizes the image if needed
+  /// Does not support gif image
+  Future<XFile> resizeImageIfNeeded(XFile file, double? maxWidth,
```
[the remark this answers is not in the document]
balvinderz: I have now  made three functions . Would they need to be individually tested ?.

packages/image_picker/image_picker_for_web/lib/image_resizer_utils.dart
ditman: > The android and IOS implementation of this function always downScales the image and never upscales the image.

Absolutely, yes!

packages/image_picker/image_picker_for_web/README.md
```diff
@@ -43,7 +43,8 @@ Each browser may implement `capture` any way they please, so it may (or may not)
 difference in your users' experience.
 
 ### pickImage()
-The arguments `maxWidth`, `maxHeight` and `imageQuality` are not supported on the web.
+The arguments `maxWidth`, `maxHeight` and `imageQuality` are not supported for gif images.
+THe argument `imageQuality` only works for jpeg and webp images.
```
ditman, line 47:
```suggestion
The argument `imageQuality` only works for jpeg and webp images.
```

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -157,8 +163,11 @@ class ImagePickerPlugin extends ImagePickerPlatform {
     double? maxWidth,
     double? maxHeight,
     int? imageQuality,
-  }) {
-    return getFiles(accept: _kAcceptImageMimeType, multiple: true);
+  }) async {
+    return await Future.wait(
+        (await getFiles(accept: _kAcceptImageMimeType, multiple: true)).map(
+            (e) => _imageResizer.resizeImageIfNeeded(
+                e, maxWidth, maxHeight, imageQuality)));
```
ditman, line 170: Do you need the top-level await Future.wait? Isn't this enough?

```dart
return (await getFiles(...)).map( (e) => _imageResizer.resize(...) );
```

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -241,20 +250,20 @@ class ImagePickerPlugin extends ImagePickerPlatform {
   Future<List<XFile>> _getSelectedXFiles(html.FileUploadInputElement input) {
     final Completer<List<XFile>> _completer = Completer<List<XFile>>();
     // Observe the input until we can return something
-    input.onChange.first.then((event) {
+    input.onChange.first.then((event) async {
```
ditman, line 253: There's no `await` inside this event function, so does this need to be `async`?

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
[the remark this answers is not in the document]
ditman: > Would they need to be individually tested ?

I think that while keeping the tests that you already have for happy-case resizing, having these methods individually tested would help adding coverage for extra corner cases, yes! :)

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
```diff
@@ -0,0 +1,85 @@
...
+    final imageLoadCompleter = Completer<html.ImageElement>();
+    final imageElement = html.ImageElement();
+    imageElement.src = blobUrl;
+
+    imageElement.onLoad.listen((event) {
+      html.Url.revokeObjectUrl(blobUrl);
+      imageLoadCompleter.complete(imageElement);
+    });
+    imageElement.onError.listen((event) {
+      final excpetion = Exception(
```
ditman:
```suggestion
      final exception = Exception(
```

(small typo)

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
```diff
@@ -0,0 +1,85 @@
...
+        file.mimeType == "image/gif") {
+      //TODO Implement maxWidth and maxHeight for image/gif
+      return file;
+    }
+    try {
+      final imageElement = await loadImage(file.path);
+      final canvas = resizeImageElement(imageElement, maxWidth, maxHeight);
+      return writeCanvasToFile(file, canvas, imageQuality);
+    } catch (e) {
+      return file;
```
ditman, line 28: Note that you're revoking the blobUrl in the onLoad event of the `loadImage` method. This `return` is going to return a bad file in case `resizeImageElement` or `writeCanvasToFile` fail in any way.

I think you should `revokeObjectUrl` only after successfully calling `writeCanvasToFile` to ensure that the old blob URL is revoked after nobody else is going to use it.

packages/image_picker/image_picker_for_web/lib/image_resizer_utils.dart
ditman: > it scales the image above its original width and height

only if `resizeFactor` is < 1, I think the upscaling can be avoided by:

```
return (resizeFactor > 1 ? imageSize ~/ resizeFactor : imageSize);
```

Right? No need for the super complex algorithm to compute the resized dimensions?

ditman (commented):
Some extra comments, I think this is very close to being done!

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
balvinderz: _imageResizer.resize() returns a future. so if we use this then return type is `Future<List<Future<Xfile>>>`  but we need `Future<List<XFIile>> so we need the future.wait function.

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
balvinderz: done

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
balvinderz: done

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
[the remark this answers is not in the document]
balvinderz: I have added tests for each function.

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /// function that loads the blobUrl into an imageElement
+  Future<html.ImageElement> loadImage(String blobUrl) {
+    final imageLoadCompleter = Completer<html.ImageElement>();
+    final imageElement = html.ImageElement();
+    imageElement.src = blobUrl;
+
+    imageElement.onLoad.listen((event) {
+      html.Url.revokeObjectUrl(blobUrl);
```
ditman, line 39: This is a potential bug. If we revokeObjectUrl here, and then `resizeImageElement` or `writeCanvasToFile` fail, we're returning a `file` object (line 28) that has a revoked blob URL, so its data can't be accessed.

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
ditman: (This is unresolved, I explained the issue in another comment)

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
```diff
@@ -0,0 +1,82 @@
...
+
+import 'dart:async';
+import 'dart:math';
+import 'dart:ui';
+import 'package:image_picker_for_web/image_resizer_utils.dart';
+import 'package:image_picker_platform_interface/image_picker_platform_interface.dart';
+import 'dart:html' as html;
+
+/// Helper class that resizes image
+class ImageResizer {
```
ditman, line 13: Please, move this image_resizer and its utils to `lib/src`, so people can't `import 'package:image_picker_for_web/image_resizer.dart';` from outside of this package.

balvinderz: done

packages/image_picker/image_picker_for_web/lib/image_resizer.dart
balvinderz: Forgot to push this change. done now.

packages/image_picker/image_picker_for_web/lib/image_resizer_utils.dart
balvinderz: yes

ditman:
@googlebot I consent.

ditman:
I've rebased this branch with the latest master, there was a small breakage in CI that required the latest code to pass.

ditman:
(It's too late to land this. I'll merge on monday.)

balvinderz:
Thanks for the dash @ditman 😄

PR #4414, opened 2021-10-05
Prepare Android Implementation for Pigeon

bparrishMines:
## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -117,23 +117,23 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
 
     Map<String, Object> settings = (Map<String, Object>) params.get("settings");
     if (settings != null) {
-      applySettings(settings);
+      applySettings(webView, settings);
     }
 
     if (params.containsKey(JS_CHANNEL_NAMES_FIELD)) {
       List<String> names = (List<String>) params.get(JS_CHANNEL_NAMES_FIELD);
       if (names != null) {
-        registerJavaScriptChannelNames(names);
+        registerJavaScriptChannelNames(webView, names);
       }
     }
 
     Integer autoMediaPlaybackPolicy = (Integer) params.get("autoMediaPlaybackPolicy");
     if (autoMediaPlaybackPolicy != null) {
-      updateAutoMediaPlaybackPolicy(autoMediaPlaybackPolicy);
+      updateAutoMediaPlaybackPolicy(webView, autoMediaPlaybackPolicy);
     }
     if (params.containsKey("userAgent")) {
       String userAgent = (String) params.get("userAgent");
-      updateUserAgent(userAgent);
+      updateUserAgent(webView, userAgent);
     }
     if (params.containsKey("initialUrl")) {
       String url = (String) params.get("initialUrl");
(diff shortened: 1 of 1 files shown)


packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -220,113 +220,121 @@ public void onFlutterViewDetached() {
...
       case "getScrollY":
-        getScrollY(result);
+        result.success(getScrollY(webView));
         break;
       default:
         result.notImplemented();
     }
   }
 
   @SuppressWarnings("unchecked")
```
mvanbeusekom, line 302: I believe this suppressions can now be removed as the casting of method call arguments has moved from this method to the `onMethodCall` method.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -341,65 +349,49 @@ public void onReceiveValue(String value) {
...
-  private void addJavaScriptChannels(MethodCall methodCall, Result result) {
-    List<String> channelNames = (List<String>) methodCall.arguments;
-    registerJavaScriptChannelNames(channelNames);
-    result.success(null);
+  private void addJavaScriptChannels(WebView webView, List<String> channelNames) {
+    registerJavaScriptChannelNames(webView, channelNames);
   }
 
   @SuppressWarnings("unchecked")
```
mvanbeusekom, line 356: This suppression can also be removed.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -341,65 +349,49 @@ public void onReceiveValue(String value) {
...
   }
 
-  private void getScrollY(Result result) {
-    result.success(webView.getScrollY());
+  private int getScrollY(WebView webView) {
+    return webView.getScrollY();
   }
 
-  private void applySettings(Map<String, Object> settings) {
+  private void applySettings(WebView webView, Map<String, Object> settings) {
```
mvanbeusekom, line 388: When marking the `updateJsMode` method as `static`, this method can also be marked `static`. This would have a small performance advantage and also guards against usage of instance members.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -434,7 +426,7 @@ private void applySettings(Map<String, Object> settings) {
     }
   }
 
-  private void updateJsMode(int mode) {
+  private void updateJsMode(WebView webView, int mode) {
```
mvanbeusekom, line 429: This method can be marked as `static`. This would have a small performance advantage and also guards against usage of instance members.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -447,7 +439,7 @@ private void updateJsMode(int mode) {
     }
   }
 
-  private void updateAutoMediaPlaybackPolicy(int mode) {
+  private void updateAutoMediaPlaybackPolicy(WebView webView, int mode) {
```
mvanbeusekom, line 442: This method can also be marked as `static`.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -456,14 +448,14 @@ private void updateAutoMediaPlaybackPolicy(int mode) {
     }
   }
 
-  private void registerJavaScriptChannelNames(List<String> channelNames) {
+  private void registerJavaScriptChannelNames(WebView webView, List<String> channelNames) {
```
mvanbeusekom, line 451: This method can also be marked as `static`.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -456,14 +448,14 @@ private void updateAutoMediaPlaybackPolicy(int mode) {
...
     for (String channelName : channelNames) {
       webView.addJavascriptInterface(
           new JavaScriptChannel(methodChannel, channelName, platformThreadHandler), channelName);
     }
   }
 
-  private void updateUserAgent(String userAgent) {
+  private void updateUserAgent(WebView webView, String userAgent) {
```
mvanbeusekom, line 458: This method can also be marked as `static`.

mvanbeusekom (commented):
The code changes look good and I can clearly see how this would make the transition to Pigeon easier. 

I did notice some `@SuppressWarnings` attributes that I think can now be removed and I believe some of the methods can now be marked as `static` which would yield a small performance gain and also protects against using class members.

I also found a few `if` statements that are missing curly braces, which is required by convention.

bparrishMines:
@mvanbeusekom In response to the comments about making methods static, [pigeon](https://pub.dev/packages/pigeon) will generate an abstract class that looks something like:

```java
abstract class WebViewApi {
  void loadUrl(WebView webView, String url, Map<String, String> headers);
  void goBack(WebView webView);
  void reload(WebView webView);
}
```

And my intention in the next PR was to covert the `FlutterWebView.java` to be an implementation of this class. So all the `static` modifiers would be removed.

mvanbeusekom:
> @mvanbeusekom In response to the comments about making methods static, [pigeon](https://pub.dev/packages/pigeon) will generate an abstract class that looks something like:
> 
> 
> 
> ```java
> 
> abstract class WebViewApi {
> 
>   void loadUrl(WebView webView, String url, Map<String, String> headers);
> 
>   void goBack(WebView webView);
> 
>   void reload(WebView webView);
> 
> }
> 
> ```
> 
> 
> 
> And my intention in the next PR was to covert the `FlutterWebView.java` to be an implementation of this class. So all the `static` modifiers would be removed.

Aaa ok yes that makes a lot of sense. Please ignore all these comments.

stuartmorgan:
> I'm basically just winging each step to get there :/

FWIW, they way I would do this if it were me is to pick a platform, map out what I think it would roughly look like, and then build it as a series of local commits start to finish, where I created new chained branches for each checkpoint of what seems like it would make sense as an incremental PR. When I found a step that was wrong, I would jump back to the relevant earlier branch, fix it, and then rebase all the later pieces. Once I had something I was happy with, I would go back and turn each branch point into an incremental PR. (In this case, I would expect it's only 2 or 3 PRs?) And then do the same for the second platform, with the structure already mapped out successfully from the first one, making it easier.

It's slightly fiddly to do the backtracking to older branches, but it means that by the time you are sending things for review you already know that each piece makes sense, and where it's going, which makes the reviews much easier. And doing the backtracking locally first avoids unnecessary churn in the actual landed code (like adding a parameter to a bunch of methods, only to realize after landing it that they aren't actually a good idea). I've done a lot of refactorings this way over the years, and found it very effective.

PR #4420, opened 2021-10-09
[camera] Add front camera mirroring update on iOS

trK54Ylmz:
This PR intents to update mirroring on front camera on iOS devices

https://github.com/flutter/flutter/issues/91491

The current and updated videos, [example.zip](https://github.com/flutter/plugins/files/7315257/example.zip)
 
## Pre-launch Checklist

* [x] I read the [Contributor Guide](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
* [x] I read the [Tree Hygiene](https://github.com/flutter/flutter/wiki/Tree-hygiene) wiki page, which explains my responsibilities.
* [x] I read and followed the [relevant style guides](https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#style) and ran [the auto-formatter](https://github.com/flutter/plugins/blob/master/script/tool/README.md#format-code). (Note that unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
* [x] I signed the [CLA](https://cla.developers.google.com/).
* [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
* [x] I listed at least one issue that this PR fixes in the description above.
* [x] I [updated pubspec.yaml](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates) with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning).
* [x] I updated CHANGELOG.md to add a description of the change.
* [ ] I updated/added relevant documentation (doc comments with `///`).
* [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
* [x] All existing and new tests are passing.


stuartmorgan:
> I have tried to revert the mirroring effect (currently videos are horizontally mirrored) on iOS when front camera is active. The video mirror enabled in [CameraPlugin.m#L394-L396](https://github.com/flutter/plugins/blob/master/packages/camera/camera/ios/Classes/CameraPlugin.m#L394-L396).

What is the purpose of having one part of the plugin code enable mirroring only to manually undo that mirroring elsewhere in the same plugin?

trK54Ylmz:
> > I have tried to revert the mirroring effect (currently videos are horizontally mirrored) on iOS when front camera is active. The video mirror enabled in [CameraPlugin.m#L394-L396](https://github.com/flutter/plugins/blob/master/packages/camera/camera/ios/Classes/CameraPlugin.m#L394-L396).
> 
> What is the purpose of having one part of the plugin code enable mirroring only to manually undo that mirroring elsewhere in the same plugin?

When we remove mirroring from [CameraPlugin.m#L394-L396](https://github.com/flutter/plugins/blob/master/packages/camera/camera/ios/Classes/CameraPlugin.m#L394-L396) the camera preview does not show the video correctly (which is video flipped horizontally in the preview and this is not expected behavior from camera) but output is quite right.

I believe the maintainers had 2 options, showing the correct video in the preview (`videoMirrored = YES`) or saving the correct video for the output (`videoMirrored = NO` or removing the lines 394:396), I think camera preview has higher priority in this case.

trK54Ylmz:
for the initial work, I didn't want to change the main logic so I only applied flip operation on native side. I have tried to do flip operation on Dart side in here [camera_preview.dart#L47](https://github.com/flutter/plugins/blob/master/packages/camera/camera/lib/src/camera_preview.dart#L47) by using `Matrix4` transformation, seems working properly. 

Should I update the Dart code or keep it on native side @stuartmorgan? /cc @bparrishMines

Tested on iPhone 11, iPhone 7 and Redmi 8A with camera plugin example application.

The test code added in [camera_preview.dart#L47](https://github.com/flutter/plugins/blob/master/packages/camera/camera/lib/src/camera_preview.dart#L47),

```dart
return Transform(
    alignment: Alignment.center,
    transform: Matrix4.rotationY(
        controller.description.lensDirection == CameraLensDirection.front ? pi : 0,
    ),
    child: child,
);
```

stuartmorgan:
@bparrishMines Any input on the question above? It seems like it would make much more sense to record unflipped and only flip the preview display itself, but I'm not sure if there's context for the original design that I'm missing.

bparrishMines:
@trK54Ylmz You may also need to handle the fact that [Transform](https://api.flutter.dev/flutter/widgets/Transform-class.html) rotates prior to painting and not prior to layout. Which could have some consequences in how the Widget is laid out.

@stuartmorgan I believe the Dart solution provided by @trK54Ylmz would definitely be the quickest and easiest solution to this problem. But, I'll provide some context why that is below.

The TLDR is that preview and video recording share the same [capture output](https://developer.apple.com/documentation/avfoundation/avcaptureoutput?language=objc) because it didn't require preview to use a `PlatformView` (we currently render it to a `Texture`). And this implementation was written before `PlatformView`s existed.

The longer explanation is that preview, video Recording, and byte streaming all use the same [AVCaptureVideoDataOutput] because:
1. An [AVCaptureSession](https://developer.apple.com/documentation/avfoundation/avcapturesession) can only have one `AVCaptureVideoDataOutput` at a time. And since all three required this output, they had to share the output frames.
2. Previews can be streamed to an [AVCaptureVideoPreviewLayer](https://developer.apple.com/documentation/avfoundation/avcapturevideopreviewlayer?language=objc). However, as far as I have been able to research, this would require using `PlatformView`s to show the frames. (e.g. with a `UIView`). And `PlatformView`s didn't exist during the implementation. 
3. Video recording could be done with [AVCaptureMovieFileOutput](https://developer.apple.com/documentation/avfoundation/avcapturemoviefileoutput?language=objc). However, `AVCaptureMovieFileOutput` and `AVCaptureVideoDataOutput` can't be used with the same `AVCaptureSession`.
4. Video recording also couldn't use `AVCaptureMovieFileOutput` because there is no support for pausing and resuming recording on iOS, only MacOS. And this feature was [required by a major customer](https://github.com/flutter/plugins/pull/1370), so we can't really remove it either. (Side note: support for this on iOS required a complex solution I implemented with `AVCaptureVideoDataOutput`.

So assuming this is all still true and I didn't forget anything, we could swap out the preview frames to render to a `UIView` instead. Then the mirroring could be handled separately on the native side.

stuartmorgan:
> So assuming this is all still true and I didn't forget anything, we could swap out the preview frames to render to a `UIView` instead. Then the mirroring could be handled separately on the native side.

We have the same issue on Android though, as I understand it, which is part of why I was suggesting a Dart-side mirroring solution.

bparrishMines:
I only really know the context of the native side of the API. Most of the Dart code is relatively new to me. But it makes sense from my understanding of how the frames are obtained.

I don't think Android had the same issue since mirroring is handled automatically by `Camera2` and there is no option to change it. I think the issue with Android had to do with a rotation on the z-axis: https://github.com/flutter/flutter/issues/76210

@mvanbeusekom Do you have thoughts on this solution since you worked on rotation issues for camera before?

stuartmorgan:
@bparrishMines @mvanbeusekom What's the status on deciding how we want to move forward here?

mvanbeusekom:
I think @bparrishMines is correct and this issue only occurs on iOS. 

I have been studying the `AVFoundation` API a bit and I had the impression we could resolve the mirroring problem using the  `AVCaptureConnection.videoMirrored()` method. Something along the lines of (which is a small extension of the current code at [CameraPlugin.m#L414-416](https://github.com/flutter/plugins/blob/main/packages/camera/camera/ios/Classes/CameraPlugin.m#L414-416)):

```objc
  if ([_captureDevice position] == AVCaptureDevicePositionFront && connection.supportsVideoMirroring) {
    connection.automaticallyAdjustsVideoMirroring = NO;
    connection.videoMirrored = YES;
  }
```

I didn't have the time to fully test this, but it seems the suggested solution on the [Apple developer forums](https://developer.apple.com/forums/thread/126936).

stuartmorgan:
> I think @bparrishMines is correct and this issue only occurs on iOS.

Why are people in https://github.com/flutter/flutter/issues/27650 saying it's happening on Android as well though?

jgoyvaerts:
There's also a mirroring issue on windows. Camera preview is mirrored (wrong), actual picture taken is not mirrored (correct). Tested both with an integrated laptop webcam and an external usb webcam

stuartmorgan:
> There's also a mirroring issue on windows. Camera preview is mirrored (wrong), actual picture taken is not mirrored (correct).

That sounds like correct behavior to me; please file a new issue for discussion.

hellohuanlin:
@trK54Ylmz are you planning on creating an issue for further discussion? Implementation wise @mvanbeusekom's approach sounds very promising IMO.

jmagman:
Thanks for your contribution! Since I haven't seen a reply, I'm going to reluctantly close this pull request for now, but if you get time to address the comments, we'd be more than happy to take a look at a new patch. If you do send a new patch, it would be very useful to reference this one in the description in order to help reviewers see the previous context.

reiko-dev:
Any update on this? This is happening on Android too @jmagman. 
Reversing the mirror is easy for images, but for videos we can't do much without making the user wait, wait and then Hate our app.

jmagman:
This PR is closed, the issue is tracked at https://github.com/flutter/flutter/issues/27650.  We had a big backlog of PRs, which is why this one took so long to be reviewed.  We've cleared out most of that backlog and would happily review a PR similar to this one with the feedback addressed.

PR #4430, opened 2021-10-13
[camera] Add camera test module map

jmagman:
Create a test header for https://github.com/flutter/plugins/pull/4426.

Create `camera.Test` explicit clang module. Include new modulemap and umbrella header to support this.


jmagman:
Closing, this is included in https://github.com/flutter/plugins/pull/4426.

PR #4434, opened 2021-10-15
[in_app_purchase] Update to the latest pkg:json_serializable

kevmoo:
- Remove unneeded class now that enum helpers can be generated without one
- Code reorganization
- Remove superflous JsonKey values


kevmoo:
Help here, @cyanglaz ?

kevmoo:
@LHLL – do we need your approval, too?

PR #4448, opened 2021-10-23
[image_picker] Support reading WebP images for iOS

TrGiLong:
Hi, 

This PR adds separate handling for WebP images. The reason is that the `loadObjectOfClass:completionHandler:` method of `NSItemProvider` class doesn't support direct reading of WebP images and returns `Could not coerce an item to class UIImage` error. 

Before adding this handling, the platform code doesn't emit any WebP images to the flutter app. After adding this handling, the platform emits WebP images.

Fixes [#88094](https://github.com/flutter/flutter/issues/88094)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/image_picker/image_picker/example/ios/RunnerTests/PickerSaveImageToPathOperationTests.m
```diff
@@ -0,0 +1,56 @@
...
+
+@implementation PickerSaveImageToPathOperationTests
+
+- (void)testSaveWebPImage API_AVAILABLE(ios(14));  {
+    // Read item from bundle
+    NSItemProvider *itemProvider = [[NSItemProvider alloc] initWithContentsOfURL:[[NSBundle mainBundle] URLForResource:@"webpImage" withExtension:@"webp"]];
+    
+    PHPickerResult *result = [self createPickerResult:itemProvider withIdentifier:UTTypeWebP.identifier];
+    
+    [self testOperationWithPickerResult:result];
```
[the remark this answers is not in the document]
TrGiLong: My mistake 😅. I want to create a general method that tests the saving process by giving it `PHPickerResult `. The `result` instance here is a result of picking of jpg, png, gif or webp files.

packages/image_picker/image_picker/example/ios/RunnerTests/PickerSaveImageToPathOperationTests.m
```diff
@@ -0,0 +1,56 @@
...
+    XCTestExpectation *pathExpectation = [self expectationWithDescription:@"Path was created"];
+    
+    FLTPHPickerSaveImageToPathOperation *operation = [[FLTPHPickerSaveImageToPathOperation alloc]
+                                                      initWithResult:result
+                                                      maxHeight:@100
+                                                      maxWidth:@100
+                                                      desiredImageQuality:@100
+                                                      savedPathBlock:^(NSString *savedPath) {
+        if (savedPath != nil) {
+            [pathExpectation fulfill];
```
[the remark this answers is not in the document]
TrGiLong: I check it now by using `[[NSFileManager defaultManager] fileExistsAtPath:savedPath]`

packages/image_picker/image_picker/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -19,6 +19,12 @@
...
 		680049392280F2B9006DD6AB /* jpgImage.jpg in Resources */ = {isa = PBXBuildFile; fileRef = 680049362280F2B8006DD6AB /* jpgImage.jpg */; };
 		6801C8392555D726009DAF8D /* ImagePickerFromGalleryUITests.m in Sources */ = {isa = PBXBuildFile; fileRef = 6801C8382555D726009DAF8D /* ImagePickerFromGalleryUITests.m */; };
+		86E9A890272747C00017E6E0 /* webpImage.webp in Resources */ = {isa = PBXBuildFile; fileRef = 86E9A88F272747B90017E6E0 /* webpImage.webp */; };
+		86E9A89127274A360017E6E0 /* webpImage.webp in Resources */ = {isa = PBXBuildFile; fileRef = 86E9A88F272747B90017E6E0 /* webpImage.webp */; };
+		86E9A893272754860017E6E0 /* PickerSaveImageToPathOperationTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 86E9A892272754860017E6E0 /* PickerSaveImageToPathOperationTests.m */; };
+		86E9A894272754A30017E6E0 /* webpImage.webp in Resources */ = {isa = PBXBuildFile; fileRef = 86E9A88F272747B90017E6E0 /* webpImage.webp */; };
+		86E9A895272769130017E6E0 /* pngImage.png in Resources */ = {isa = PBXBuildFile; fileRef = 680049352280F2B8006DD6AB /* pngImage.png */; };
```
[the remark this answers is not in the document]
TrGiLong: They are already existing in the project. These lines will appear when we set a project target for them. 😄

TrGiLong:
Hi @stuartmorgan, @cyanglaz 

The PR is ready to be reviewed. Can you please check it? 😅

TrGiLong:
Hi @cyanglaz,

Are you still working on this PR? 😅
Should anything else be done by me?

cyanglaz:
@TrGiLong I apologize for the delay. I will make sure to squeeze in some time this week to take a look at it. Thank you for being patient.

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
```diff
@@ -82,51 +84,63 @@ - (void)start {
   }
   if (@available(iOS 14, *)) {
     [self setExecuting:YES];
+
+    if ([self.result.itemProvider hasItemConformingToTypeIdentifier:UTTypeWebP.identifier]) {
+      [self.result.itemProvider
+          loadDataRepresentationForTypeIdentifier:UTTypeWebP.identifier
+                                completionHandler:^(NSData *_Nullable data,
+                                                    NSError *_Nullable error) {
+                                  UIImage *image = [[UIImage alloc] initWithData:data];
```
cyanglaz, line 93: Can we also add error handling here? What if `data` is nil? (Should also test this)

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
```diff
@@ -82,51 +84,63 @@ - (void)start {
...
-                                                        maxHeight:self.maxHeight
-                                                     imageQuality:self.desiredImageQuality];
-                               [self completeOperationWithPath:savedPath];
-                             }];
-            }
+            [self processImage:image];
           }
         }];
   } else {
     [self setFinished:YES];
```
cyanglaz, line 108: I know it's not your code, should probably add error handling here too.

cyanglaz (changes requested):

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
TrGiLong: You are right, we have to add error handling here. The reason why I didn't do it because I thought it's outside of my scope :D But I will add it anyway.

TrGiLong: @cyanglaz 

I think it's not easy to add error handling according the code that we have. Let see the signature of the function in flutter:
```
Future<List<PickedFile>?> pickMultiImage({
    double? maxWidth,
    double? maxHeight,
    int? imageQuality,
  })
```

It's not clear how developer can handle the error because there are many scenarios:
1. All selected images return error.
2. Many of selected images return error.

In the first case we throw just error. But in second case throwing the error doesn't make sense, because I think the devs still want to get successfully picked image, and then show the error for those images, that throw error. 

Passing a successfully picked image through error can solve the problem, but I think it's not a good design.

My suggestion is creating something like that:
```
class PickedFileInfo {
   final PickedFile file;
   final Error? error;
}

Future<List<PickedFileInfo>?> pickMultiImage({
    double? maxWidth,
    double? maxHeight,
    int? imageQuality,
  })
```

or

```
Future<List<PickedFileInfo>?> pickMultiImage({
    double? maxWidth,
    double? maxHeight,
    int? imageQuality,
    void function(error)? onError,
  })
```

cyanglaz: This is a good point. Can we instead add a new field called `error` in PickeFile?

TrGiLong: Ok, I will try. I let you know later.

wenboLee:
what's new on the PR? 😄

TrGiLong:
Hi @cyanglaz,

I have tried to improve error handling for image picking but it makes this PR enormous. I think we have to do it in separate PR. There reasons are:
1. It requires to change the way, how we send the data back to flutter. Currently we receive a list of path of successful picking. Adding an error information to the list requires to change all data handling in Flutter side.
2. It isn't clear how to send back error to invoker. Currently we use `XFile` and deprecated `PickedFile` classes. Adding the `error` field solves the problem, but I don't think it's a good idea from design perspective. I assume that `error` in `File` means a reading or writing problem. The picking error shouldn't been there. I still prefer to show a list of picking error in callback to minimise an affect to other projects. 
3. We also should apply those changes to Android. I'm not an expert there.

TrGiLong:
> what's new on the PR? 😄

Hi, @wenboLee 😀

The PR is ready, but now we discuss about error handling of picking.

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
cyanglaz: Since this PR already improves the current plugin and error handling is out of scope, I'm ok with not adding it in this PR. I create an issue tracking the error handling here: https://github.com/flutter/flutter/issues/95268

packages/image_picker/image_picker/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.8.4+4
+
+* Support adding WebP for iOS
```
cyanglaz, line 3:
```suggestion
* iOS: allow picking images with WebP format.
```

packages/image_picker/image_picker/example/ios/RunnerTests/PickerSaveImageToPathOperationTests.m
```diff
@@ -0,0 +1,85 @@
...
+                                                             withExtension:@"webp"];
+  NSItemProvider *itemProvider = [[NSItemProvider alloc] initWithContentsOfURL:imageURL];
+  PHPickerResult *result = [self createPickerResult:itemProvider
+                                     withIdentifier:UTTypeWebP.identifier];
+
+  [self saveImageToPath:result];
+}
+
+- (void)testSavePNGImage API_AVAILABLE(ios(14)) {
+  NSURL *imageURL = [[NSBundle bundleForClass:[self class]] URLForResource:@"pngImage"
```
cyanglaz, line 28: Thanks for also adding this test for all other formats!

packages/image_picker/image_picker/example/ios/RunnerTests/PickerSaveImageToPathOperationTests.m
```diff
@@ -0,0 +1,85 @@
...
+                        withIdentifier:(NSString *)identifier API_AVAILABLE(ios(14)) {
+  PHPickerResult *result = OCMClassMock([PHPickerResult class]);
+
+  OCMStub([result itemProvider]).andReturn(itemProvider);
+  OCMStub([result assetIdentifier]).andReturn(identifier);
+
+  return result;
+}
+
+- (void)saveImageToPath:(PHPickerResult *)result API_AVAILABLE(ios(14)) {
```
cyanglaz, line 67: nits:
This method does more than just "saveImageToPath", this is actually testing the save image operation.
How about rename it to something like: `testSaveImageOperationWithResult:`.

cyanglaz (approved):
LGTM % nits

CC @stuartmorgan for secondary review.

packages/image_picker/image_picker/CHANGELOG.md
stuartmorgan: Nit: should be "Allows", not "Allow", in the suggested change.

packages/image_picker/image_picker/example/ios/RunnerTests/PickerSaveImageToPathOperationTests.m
stuartmorgan: It can't start with "test", but something similar with "verify" or "validate" would be good.

stuartmorgan, line 67: This needs a declaration comment, per the ObjC style guide.

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
```diff
@@ -82,51 +84,63 @@ - (void)start {
...
   }
 }
 
+- (void)processImage:(UIImage *)localImage API_AVAILABLE(ios(14)) {
```
stuartmorgan, line 112: Needs a declaration comment.

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
```diff
@@ -82,51 +84,63 @@ - (void)start {
...
+  PHAsset *originalAsset = [FLTImagePickerPhotoAssetUtil getAssetFromPHPickerResult:self.result];
+
+  if (self.maxWidth != (id)[NSNull null] || self.maxHeight != (id)[NSNull null]) {
```
stuartmorgan, line 115: I know it's pre-existing code, but are these casts actually necessary?

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
```diff
@@ -82,51 +84,63 @@ - (void)start {
...
+    localImage = [FLTImagePickerImageUtil scaledImage:localImage
+                                             maxWidth:self.maxWidth
+                                            maxHeight:self.maxHeight
+                                  isMetadataAvailable:originalAsset != nil];
+  }
+  if (!originalAsset) {
```
stuartmorgan, line 121: Since you're already moving this, please flip the condition and the blocks. This:
```
if (foo) {
  // foo case
} else {
  // !foo case
}
```

is generally much easier to read than:
```
if (!foo) {
  // !foo case
} else {
  // !!foo case, which is foo case
}
```

packages/image_picker/image_picker/example/ios/RunnerTests/PickerSaveImageToPathOperationTests.m
```diff
@@ -0,0 +1,85 @@
...
+  NSURL *imageURL = [[NSBundle bundleForClass:[self class]] URLForResource:@"gifImage"
+                                                             withExtension:@"gif"];
+  NSItemProvider *itemProvider = [[NSItemProvider alloc] initWithContentsOfURL:imageURL];
+  PHPickerResult *result = [self createPickerResult:itemProvider
+                                     withIdentifier:UTTypeWebP.identifier];
+
+  [self saveImageToPath:result];
+}
+
+- (PHPickerResult *)createPickerResult:(NSItemProvider *)itemProvider
```
stuartmorgan, line 57: Needs a declaration comment per style guide.

stuartmorgan, line 57: In ObjC naming, the thing right before the argument should name the argument; the argument isn't a result here. This should be called something like `createPickerResultWithProvider:identifier:`

stuartmorgan (changes requested):

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
TrGiLong: I don't want to touch this code because it works. But I have analysed the flow.  

1. The flutter side initialises the call.
```dart
Future<String?> _getImagePath({
    required ImageSource source,
    double? maxWidth,
    double? maxHeight,
    int? imageQuality,
    CameraDevice preferredCameraDevice = CameraDevice.rear,
  }) {
/// ....
return _channel.invokeMethod<String>(
      'pickImage',
      <String, dynamic>{
        'source': source.index,
        'maxWidth': maxWidth,
        'maxHeight': maxHeight,
        'imageQuality': imageQuality,
        'cameraDevice': preferredCameraDevice.index
      },
    );
}
```
2. Base on the information of channel mechanism. Dart nil will be convert to objective-c nil or NSNull if it's nested ([link](https://docs.flutter.dev/development/platform-integration/platform-channels?tab=type-mappings-obj-c-tab))
3. The plugin use just parses those arguments.
```objective-c
NSNumber *maxWidth = [_arguments objectForKey:@"maxWidth"];
NSNumber *maxHeight = [_arguments objectForKey:@"maxHeight"];
```

Does it mean that `maxWidth` and `maxHeight` are nil? If yes then the code should crash because we some nested method will try to access to the value. And as far as I know [NSNull null] is not equal to nil.

packages/image_picker/image_picker/example/ios/RunnerTests/PickerSaveImageToPathOperationTests.m
TrGiLong: I completely forgot this rule. I have switched to swift for a long time ago and didn't write any objective-c code :D

TrGiLong:
Hi @stuartmorgan

Can you review my code? I have tried to fix all your request as soon as possible :D

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
stuartmorgan: My question wasn't whether the checks were necessary, it was whether the *casts* (to `id`) were necessary.

TrGiLong: Oh, I think is not necessary.
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/NumbersandValues/Articles/Null.html

stuartmorgan: Either there's a compiler warning without the casts, or there isn't; it's not an opinion-based question. If there isn't, we should remove them.

(The real problem is that `NSNull` should never be assigned to `NSNumber` variables in the first place, since it's a type violation; once that's fixed these would be `nil` checks instead. @cyanglaz Perhaps you could fix that in a small separate PR?)

TrGiLong:
@stuartmorgan 

 I have checked your review. I have to say sorry because I have submitted a wrong code. The part with error handling in `handleSavedPathList` belongs to my attempt to improve plugin error handling. But we have decided to create a separated branch for this task. I will revert the last commit.

packages/image_picker/image_picker/example/ios/RunnerTests/PickerSaveImageToPathOperationTests.m
```diff
@@ -0,0 +1,99 @@
...
+                                                             withExtension:@"gif"];
+  NSItemProvider *itemProvider = [[NSItemProvider alloc] initWithContentsOfURL:imageURL];
+  PHPickerResult *result = [self createPickerResultWithProvider:itemProvider
+                                     withIdentifier:UTTypeWebP.identifier];
+
+  [self verifySavingImageWithPickerResult:result];
+}
+
+/**
+  * Create a mock picker result using NSItemProvider.
```
stuartmorgan, line 58: Nit: Creates

Per style guide.

packages/image_picker/image_picker/example/ios/RunnerTests/PickerSaveImageToPathOperationTests.m
```diff
@@ -0,0 +1,99 @@
...
+  PHPickerResult *result = OCMClassMock([PHPickerResult class]);
+
+  OCMStub([result itemProvider]).andReturn(itemProvider);
+  OCMStub([result assetIdentifier]).andReturn(identifier);
+
+  return result;
+}
+
+/**
+  * Validate a saving process of FLTPHPickerSaveImageToPathOperation.
```
stuartmorgan, line 74: Validates

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
```diff
@@ -82,51 +84,66 @@ - (void)start {
...
+            [self processImage:image];
           }
         }];
   } else {
     [self setFinished:YES];
   }
 }
 
+/*
+ * This method processes the image and save it to the file.
```
stuartmorgan, line 113: Nit: s/This method processes/Processes/
Nit: s/save/saves/

stuartmorgan, line 113: What is "the file" here? There's no file parameter, so its not clear as written what file is being referred to.

stuartmorgan (changes requested):
This mostly looks good, just small nits. You'll also need to merge in the latest `main` so that CI can run; this brach is now old enough that the repository keys have expired which is why everything is red.

packages/image_picker/image_picker/ios/Classes/FLTPHPickerSaveImageToPathOperation.m
stuartmorgan: I fixed the use of `NSNull` in this plugin, so these check should be nil checks now (as they are in the current version of the code you are moving).

stuartmorgan (changes requested):
You have [formatting issues](https://github.com/flutter/plugins/pull/4448/checks?check_run_id=5116962762) that need to be fixed.

You also need to update your unit tests for recent changes in the test structure of this package, as they are currently failing in CI.

stuartmorgan (approved):
LGTM!

stuartmorgan:
Filed https://github.com/flutter/flutter/issues/99033 to investigate why the bot won't land this. (Other team members: please don't land it manually for now.)

PR #4450, opened 2021-10-25
[webview_flutter] Add platform interface method `loadRequest`.

BeMacized:
This PR is a partial resubmission of my colleague's PR: #4169. 

It adds the `loadRequest` method to the webview plugin's platform interface, in order to support executing more specific requests when loading web pages. 

Related issue:
- flutter/flutter#27730

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_platform_interface/test/src/method_channel/webview_method_channel_test.dart
```diff
@@ -101,6 +101,30 @@ void main() {
...
+                'uri': 'https://test.url',
+                'method': 'get',
+                'headers': {},
+                'body': null,
+              }
+            },
+          ),
+        ],
+      );
+    });
```
stuartmorgan, line 126: Please add a second test that sets the optional parameters and ensures they come through as expected as well.

stuartmorgan (approved):
LGTM with nit

packages/webview_flutter/webview_flutter_platform_interface/test/src/method_channel/webview_method_channel_test.dart
BeMacized: @stuartmorgan Would that not just be testing WebViewRequest's `toJson` method? We can of course add a second test, but as that `toJson` is already being tested [here](https://github.com/flutter/plugins/pull/4450/files#diff-e82bfc76d3c600958391b3c90f53616eb0e68aea4f93cf76bb83633c8436fc2dR28), including the optional parameters, I feel like we would be testing for the same thing twice.

stuartmorgan: How do you know that `loadRequest` calls `WebViewRequest`'s `toJson` method? Or to be more precise: how do you know that it will continue to do so?

I guarantee you that I could write a PR that breaks the test I am asking you to write without changing `WebViewRequest.toJson`.

BeMacized: That is indeed a fair point, I now see it wouldn't. I've added the test for the optional parameters.

Just for my own curiosity: Would it not be possible to avoid testing `toJson` by instead mocking the WebViewRequest and checking if the return value of `toJson` gets passed to `invokeMethod`?

stuartmorgan: Tests should assert the behavior you want to guarantee. Which is a better description of the thing you want to guarantee is true:
1. "The method channel is sent the following specific keys with the expected values", or
2. "The call to the method channel is implemented by calling `toJson` on `invokeMethod`"?

(As an example of why only one of these is a good description: if someone later decides that the JSON keys aren't very descriptive, and sends a PR that updates `toJson` and its test to use different keys, what do you want to happen?)

PR #4457, opened 2021-10-28
[WIP] Delete deprecated v1 code

GaryQian:
We plan on deleting the v1 Android embedding, so we should no longer use the deprecated code.

Part of https://github.com/flutter/flutter/pull/92643 which is an experimental run of deleting v1 embedding.

This change is breaking as it drops support for old non-migrated apps.


GaryQian:
go/flutter-delete-v1-android

PR #4459, opened 2021-10-29
[webview_flutter_android] Fix bugs in Java Implementation of the Pigeon API

bparrishMines:

The pigeon implementation and [current implementation](https://github.com/flutter/plugins/blob/7588dd9cfe0b8e6ca5292389723682ab26258149/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java#L164) has potential leaking callbacks when a `WebView` is disposed. This is a possible cause of flaky tests for WebView, so this PR attempts to prevent callbacks when the corresponding `WebView` object is disposed. In https://github.com/flutter/plugins/pull/4455, the integration tests now seem to pass consistently. Along with fixing the resize test: https://github.com/flutter/plugins/pull/4460

This also fixes some trivial bugs found though the integration and widget tests in https://github.com/flutter/plugins/pull/4455.

Dart Portion: https://github.com/flutter/plugins/pull/4461

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/DownloadListenerHostApiImpl.java
```diff
@@ -5,40 +5,63 @@
 package io.flutter.plugins.webviewflutter;
 
 import android.webkit.DownloadListener;
-import io.flutter.plugins.webviewflutter.GeneratedAndroidWebView.DownloadListenerFlutterApi;
 
 class DownloadListenerHostApiImpl implements GeneratedAndroidWebView.DownloadListenerHostApi {
```
stuartmorgan, line 9: Please make sure all public classes have comments, per style.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/DownloadListenerHostApiImpl.java
```diff
@@ -5,40 +5,63 @@
...
   private final InstanceManager instanceManager;
   private final DownloadListenerCreator downloadListenerCreator;
-  private final GeneratedAndroidWebView.DownloadListenerFlutterApi downloadListenerFlutterApi;
+  private final DownloadListenerFlutterApiImpl flutterApi;
+
+  static class DownloadListenerImpl implements DownloadListener, Releasable {
+    private final Long instanceId;
+    private final DownloadListenerFlutterApiImpl flutterApi;
+    private boolean ignoreCallbacks = false;
```
stuartmorgan, line 17: I think this would benefit from a comment explaining what it's for, so people don't have to go read the implementation to understand it.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/InstanceManager.java
```diff
@@ -18,13 +18,24 @@ void addInstance(Object instance, long instanceId) {
     instanceIdsToInstances.append(instanceId, instance);
   }
 
-  /** Remove the instance from the manager. */
-  void removeInstance(long instanceId) {
+  /** Remove the instance with instanceId from the manager. */
```
stuartmorgan, line 21: Please describe the return value in the comment.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/InstanceManager.java
```diff
@@ -18,13 +18,24 @@ void addInstance(Object instance, long instanceId) {
...
+  Object removeInstanceWithInstanceId(long instanceId) {
     final Object instance = instanceIdsToInstances.get(instanceId);
     if (instance != null) {
       instanceIdsToInstances.remove(instanceId);
       instancesToInstanceIds.remove(instance);
     }
+    return instance;
+  }
+
+  /** Remove the instance from the manager. */
```
stuartmorgan, line 31: Same.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/InstanceManager.java
```diff
@@ -18,13 +18,24 @@ void addInstance(Object instance, long instanceId) {
     instanceIdsToInstances.append(instanceId, instance);
   }
 
-  /** Remove the instance from the manager. */
-  void removeInstance(long instanceId) {
+  /** Remove the instance with instanceId from the manager. */
+  Object removeInstanceWithInstanceId(long instanceId) {
```
stuartmorgan, line 22: Nit: How about just `removeInstanceWithId`; the fact that the ID is the instance's ID seems self-explanatory.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/JavaScriptChannel.java
```diff
@@ -33,6 +36,28 @@
     this.methodChannel = methodChannel;
     this.javaScriptChannelName = javaScriptChannelName;
     this.platformThreadHandler = platformThreadHandler;
+    this.instanceId = null;
+    this.flutterApi = null;
+  }
+
+  /**
```
stuartmorgan, line 43: This should have a [summary fragment](https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment)

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/JavaScriptChannelFlutterApiImpl.java
```diff
@@ -0,0 +1,27 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package io.flutter.plugins.webviewflutter;
+
+import io.flutter.plugin.common.BinaryMessenger;
+
+public class JavaScriptChannelFlutterApiImpl
```
stuartmorgan, line 9: Needs a comment.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/Releasable.java
```diff
@@ -0,0 +1,13 @@
...
+// found in the LICENSE file.
+
+package io.flutter.plugins.webviewflutter;
+
+/**
+ * Represents a resource, or a holder of resources, which may be released once they are no longer
+ * needed.
+ */
+interface Releasable {
+  void release();
```
stuartmorgan, line 12: Comment.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewClientHostApiImpl.java
```diff
@@ -63,12 +62,206 @@
     return requestData;
   }
 
+  @RequiresApi(Build.VERSION_CODES.N)
+  static class WebViewClientImpl extends WebViewClient implements Releasable {
+    private final Long instanceId;
+    private final InstanceManager instanceManager;
+    private final Boolean shouldOverrideUrlLoading;
+    private final WebViewClientFlutterApiImpl flutterApi;
+    private boolean ignoreCallbacks = false;
```
stuartmorgan, line 71: Would it make sense for all of these wrappers to null out flutterApi instead of having a second variable, so that if someone forgets to check it's an immediate crash?

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
```diff
@@ -85,6 +171,76 @@ public void onInputConnectionLocked() {
...
       unlockInputConnection();
     }
+
+    @Override
+    public void setWebViewClient(WebViewClient webViewClient) {
+      super.setWebViewClient(webViewClient);
+      if (currentWebViewClient instanceof Releasable) {
+        ((Releasable) currentWebViewClient).release();
+      }
+      currentWebViewClient = (WebViewClient) webViewClient;
```
stuartmorgan, line 181: Does Java have the ability to do subclass-restricted templates? You have this repeated pattern; it would be nice if you could instead have a simple smart pointer class that can only be created for things that inherit from Releasable (but is still typed with the correct specific type), that would handle the call-release-if-necessary-then-assign logic internally.

stuartmorgan (changes requested):

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/DownloadListenerHostApiImpl.java
bparrishMines: Done. All pigeon api impl should be fully documented now

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/InstanceManager.java
bparrishMines: done

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/InstanceManager.java
bparrishMines: done

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewClientHostApiImpl.java
bparrishMines: I agree. I changed flutter api to be nullable and made it null in release

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
bparrishMines: I wasn't sure what you meant by this. But, my assumption is that you were thinking something like this

```java
private static class ReleasableChild<T> {
    @Nullable private T child;

    private ReleasableChild(@Nullable T child) {
      this.child = child;
    }

    private void release() {
      if (child instanceof Releasable) {
        ((Releasable) child).release();
      }
      child = null;
    }
  }
```

I replaced all the variables with this value.

stuartmorgan: (It looks like the Java name for what I was looking for support for is "bounded types".)

This still relies on `release` being called manually on every assignment, which is error-prone. I was looking for something like (probably doesn't compile as written since it's freehand):
```
  private static class ReleasableValue<T extends Releasable> {
    @Nullable private T value;

    ReleasableValue(@Nullable T value) {
      this.value = value;
    }

    void set(@Nullable T newValue) {
      release();
      value = newValue;
    }

    @Nullable T get() {
      return value;
    }

    void release() {
      if (value != null) {
        value.release();
      }
      value = null;
    }

    void finalize() {
      release();
    }
  }
```

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
```diff
@@ -71,27 +71,40 @@ public void setWebContentsDebuggingEnabled(boolean enabled) {
     }
   }
 
-  private static class ReleasableChild<T> {
-    @Nullable private T child;
+  private static class ReleasableValue<T> {
```
stuartmorgan, line 74: Why isn't this using bounded types?

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
```diff
@@ -47,11 +125,79 @@ public View getView() {
...
+      super.setWebChromeClient(client);
+      currentWebChromeClient.set(client);
+    }
+
+    @SuppressLint("JavascriptInterface")
+    @Override
+    public void addJavascriptInterface(Object object, String name) {
+      super.addJavascriptInterface(object, name);
+      if (object instanceof JavaScriptChannel) {
+        javaScriptInterfaces.put(name, new ReleasableValue<>((JavaScriptChannel) object));
```
stuartmorgan, line 157: If this replaces a value, the old values won't get `release()` called until the next GC cycle; is that okay, or does any previous value need to be explicitly found and `release`d here?

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
bparrishMines: ah ok that makes sense. I changed it to:

```java
private static class ReleasableValue<T> {
    @Nullable private T value;

    ReleasableValue() {}

    ReleasableValue(@Nullable T value) {
      this.value = value;
    }

    void set(@Nullable T newValue) {
      release();
      value = newValue;
    }

    @Nullable
    T get() {
      return value;
    }

    void release() {
      if (value instanceof Releasable) {
        ((Releasable) value).release();
      }
      value = null;
    }
  }
```

I removed `finalize` since it was only calling `release`. I also changed `ReleasableValue<T extends Releasable>` to `ReleasableValue<T>` because the `WebView.setWebViewClient` will take either a `WebViewClientImpl` or `WebViewClientCompatImpl` and their only shared interface is `WebViewClient`; which doesn't extend `Releasable`.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
stuartmorgan: (Which would avoid the need for the type check in `release`, in addition to making sure incorrect use of ReleasableValue wouldn't compile.)

stuartmorgan (approved):
LGTM modulo the two comments above.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
bparrishMines: But I suppose I could just cast `WebViewClient` to `Releasable` as I set it. Maybe that is a better solution?

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
bparrishMines: Good catch. Release actually needs to be called because the Dart Object would still need to be disposed.

stuartmorgan: This is why I think `ReleasableValue` should have a `finalize` that calls `release`.

PR #4460, opened 2021-10-29
[webview_flutter] Fix Flaky Resize Test

bparrishMines:

~~After reading through https://flutter.dev/docs/cookbook/testing/widget/introduction. It seems that this test isn't using a Widget test in the expected way.~~ After more testing, the current solution does work in updating the widget, but still has a false positive for the 2nd reason below.

~~1. It's attempting to change the size of a `WebView` by pumping a new `Widget`.~~
2. I was able to make this test fail by adding a  `await Future.delayed(Duration(seconds: 2));` before the `boolean` check. This shows that resize is called regardless of changing the width and height.

This PR follows the Flutter cookbook example of a Widget test and creates a widget that changes the WebView on a button press. It also waits for the initial resize of the WebView from the initial load.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
```diff
@@ -162,91 +163,24 @@ void main() {
...
-              width: 200,
-              height: 200,
-              child: webView,
-            ),
-          ],
-        ),
-      ),
-    );
+    ));
+    await tester.pump(Duration(seconds: 3));
```
bparrishMines, line 175: This is required to wait for the initial resize of the `WebView`. The line `await pageFinishedCompleter.future;` can finish before this.

stuartmorgan: Please capture this in a code comment, as it's non-obvious.

packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
```diff
@@ -162,91 +163,24 @@ void main() {
...
-              width: 400,
-              height: 400,
-              child: webView,
-            ),
-          ],
-        ),
-      ),
-    );
+    await tester.tap(find.byKey(const ValueKey('resizeButton')));
+    await tester.pump(Duration(seconds: 3));
```
stuartmorgan, line 181: How were the 3-second values chosen? It seems like a very arbitrary number.

stuartmorgan (approved):
LGTM with nits.

(Related though: do we need three copies of this test? It's not clear to me where exactly the functionality that it's intended to test lives; are there important parts that live in both native an app-facing-package-dart?)

packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
bparrishMines: It is mostly arbitrarily chosen :/. I just chose a safe value to wait for. I considered just waiting for the first call to resize, but I wasn't 100% sure it gets called every time on every platform.

stuartmorgan: Wait, so it this always waiting 3 seconds? I'm not all that familiar with widget tests; I assumed this was a timeout. If it's a static 3 second wait, that's still going to be a source of flake.

bparrishMines: That's a good point. I changed the tests to wait for resize completers.

stuartmorgan (approved):
LGTM

PR #4461, opened 2021-10-29
Fix bugs in Dart Implementation of the Pigeon API 

bparrishMines:

The pigeon implementation and [current implementation](https://github.com/flutter/plugins/blob/7588dd9cfe0b8e6ca5292389723682ab26258149/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java#L164) has potential leaking callbacks when a `WebView` is disposed. This is a possible cause of flaky tests for WebView, so this PR attempts to prevent callbacks when the corresponding `WebView` object is disposed. In https://github.com/flutter/plugins/pull/4455, the integration tests now seem to pass consistently. Along with fixing the resize test: https://github.com/flutter/plugins/pull/4460

This also fixes some trivial bugs found though the integration and widget tests in https://github.com/flutter/plugins/pull/4455

Java Portion: https://github.com/flutter/plugins/pull/4459

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -332,7 +309,7 @@ class WebSettings {
   ///
   /// The default is false.
   Future<void> setSupportMultipleWindows(bool support) {
-    return api.setSupportZoomFromInstance(this, support);
+    return api.setSupportMultipleWindowsFromInstance(this, support);
```
stuartmorgan, line 312: This seems like a change that should have an associated test.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -423,12 +400,19 @@ class WebSettings {
...
+      JavaScriptChannelFlutterApi.setup(JavaScriptChannelFlutterApiImpl());
+      _flutterApisHaveBeenSetup = true;
+    }
+  }
 
   /// Pigeon Host Api implementation for [JavaScriptChannel].
   @visibleForTesting
   static JavaScriptChannelHostApiImpl api = JavaScriptChannelHostApiImpl();
 
+  static bool _flutterApisHaveBeenSetup = false;
```
stuartmorgan, line 414: Nit: `_flutterApisHaveBeenSetUp`. (Our code shouldn't use the wrong name even though Pigeon does.)

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -439,55 +423,62 @@ abstract class JavaScriptChannel {
 /// Receive various notifications and requests for [WebView].
 abstract class WebViewClient {
   /// Constructs a [WebViewClient].
-  WebViewClient({this.shouldOverrideUrlLoading = true});
+  WebViewClient({this.shouldOverrideUrlLoading = true}) {
+    if (!_flutterApisHaveBeenSetup) {
+      WebViewClientFlutterApi.setup(WebViewClientFlutterApiImpl());
+      _flutterApisHaveBeenSetup = true;
```
stuartmorgan, line 429: Same.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -573,6 +564,16 @@ abstract class WebViewClient {
...
 abstract class DownloadListener {
+  /// Constructs a [DownloadListener].
+  DownloadListener() {
+    if (!_flutterApisHaveBeenSetup) {
+      DownloadListenerFlutterApi.setup(DownloadListenerFlutterApiImpl());
+      _flutterApisHaveBeenSetup = true;
+    }
+  }
+
+  static bool _flutterApisHaveBeenSetup = false;
```
stuartmorgan, line 575: Same.

Would it make sense for all this setup to be done in one central location, perhaps?

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -439,55 +423,62 @@ abstract class JavaScriptChannel {
...
+      WebViewClientFlutterApi.setup(WebViewClientFlutterApiImpl());
+      _flutterApisHaveBeenSetup = true;
+    }
+  }
+
+  static bool _flutterApisHaveBeenSetup = false;
 
   /// User authentication failed on server.
-  static const int errorAuthentication = 0xfffffffc;
+  static const int errorAuthentication = -4;
```
stuartmorgan, line 436: Are these Java constants? If these must match something else, the comments should explain what that something is (e.g., with a docs link)

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -589,6 +590,16 @@ abstract class DownloadListener {
...
 abstract class WebChromeClient {
+  /// Constructs a [WebChromeClient].
+  WebChromeClient() {
+    if (!_flutterApisHaveBeenSetup) {
+      WebChromeClientFlutterApi.setup(WebChromeClientFlutterApiImpl());
+      _flutterApisHaveBeenSetup = true;
+    }
+  }
+
+  static bool _flutterApisHaveBeenSetup = false;
```
stuartmorgan, line 601: Same.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_api_impls.dart
```diff
@@ -239,7 +239,7 @@ class WebSettingsHostApiImpl extends WebSettingsHostApi {
     WebSettings instance,
     bool flag,
   ) {
-    return setJavaScriptCanOpenWindowsAutomatically(
+    return setJavaScriptEnabled(
```
stuartmorgan, line 242: This also seems like it should have a test.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -627,7 +638,7 @@ class WebResourceRequest {
   final String method;
 
   /// Gets the headers associated with the request.
-  final Map<String, String> requestHeaders;
+  final Map<String, String>? requestHeaders;
```
stuartmorgan, line 641: Collections should generally be non-nullable, otherwise there are two ways of representing the same thing (null and empty). Callers can use `??` to pass an empty collection when they have null.

stuartmorgan (changes requested):

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
bparrishMines: I added a test for this and the rest of the methods in the `android_webview.dart` file.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
bparrishMines: Done

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
bparrishMines: I wasn't sure where to do it at. Options that I was considering were:

1. In the constructor of `WebView` since it is the central class of the plugin. However, this pattern wouldn't be reproducible for plugins where there wasn't a central class.
2. In the constructor of `AndroidWebView`. However, a user may want to use the `android_webview` API without using the shared platform interface.
3. It could be done in the plugin registration when it's ready, so this would just be temporary. I could add a `TODO` if this should be the case.
4. The current solution. I like this one because it only sets the Flutter API when a user starts to use a WebView. I remember that there is a common complaint that Flutter startup is slow because plugin initialization is done for every plugin at startup. This solution follows the idea of initializing only once it is being used.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
bparrishMines: Done. I added the java links to each one

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_api_impls.dart
bparrishMines: done

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
bparrishMines: That's a good point. I changed it back to being non-null.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
stuartmorgan: Is there any real benefit for the setup to be incremental at a per-class-within-the-plugin-level though?

Rather than having the same static and `if` in every single class, you could have a `ensurePigeonApis` helper in a separate file that does that check and then calls *all* of the API setup calls. Then each class could call it (unconditionally) where you currently have all of the logic replicated in each class.

bparrishMines: That's a good point. At this small of a plugin, it probably has little to no benefit. I created `AndroidWebViewFlutterApis.ensureSetUp`.

stuartmorgan (approved):
LGTM

PR #4463, opened 2021-11-01
[webview_flutter] annotate deprecated functions to suppress build warnings

juliansteenbakker:
This PR annotates deprecated functions which already have been replaced on newer API's.

Fixes https://github.com/flutter/flutter/issues/92826

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/webview_flutter_android/CHANGELOG.md
+++ b/packages/webview_flutter/webview_flutter_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.2.1
+
+* Annotate deprecated functions to suppress build warnings.
+
 ## 2.2.0
 
 * Implemented new `runJavascript` and `runJavascriptReturningResult` methods in platform interface.
--- a/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
+++ b/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
@@ -36,7 +36,7 @@ class FlutterCookieManager implements MethodCallHandler {
   void dispose() {
     methodChannel.setMethodCallHandler(null);
   }
-  
+
   @SuppressWarnings("deprecation")
   private static void clearCookies(final Result result) {
     CookieManager cookieManager = CookieManager.getInstance();
--- a/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
+++ b/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
@@ -36,7 +36,8 @@ class FlutterCookieManager implements MethodCallHandler {
   void dispose() {
     methodChannel.setMethodCallHandler(null);
   }
-
+  
+  @SuppressWarnings("deprecation")
   private static void clearCookies(final Result result) {
     CookieManager cookieManager = CookieManager.getInstance();
     final boolean hasCookies = cookieManager.hasCookies();
--- a/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
+++ b/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
@@ -57,6 +57,7 @@ public class FlutterWebView implements PlatformView, MethodCallHandler {
               return true;
             }
 
+            @SuppressWarnings("deprecation")
             @Override
             public boolean shouldOverrideUrlLoading(WebView view, String url) {
               if (!flutterWebViewClient.shouldOverrideUrlLoading(
--- a/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
+++ b/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java
@@ -215,6 +215,7 @@ class FlutterWebViewClient {
         }
       }
 
+      @SuppressWarnings("deprecation")
       @Override
       public void onReceivedError(
           WebView view, int errorCode, String description, String failingUrl) {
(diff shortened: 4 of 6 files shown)


packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
```diff
@@ -37,6 +37,7 @@ void dispose() {
     methodChannel.setMethodCallHandler(null);
   }
 
+  @SuppressWarnings("deprecation")
```
stuartmorgan, line 40: Why are you suppressing deprecation warnings for the entire function, rather than just the specific call? If the new code path were to become deprecated, we would want to know.

stuartmorgan (changes requested):

stuartmorgan:
@blasten for second review

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
juliansteenbakker: As far as i know it is only possible to suppress the deprecation warning for the whole class or a function, but not a specific call. Please tell me how it's done otherwise.

stuartmorgan: There are a number of examples of statement-level suppressions in the repository: https://github.com/flutter/plugins/search?q=SuppressWarnings

What line or lines are you trying to suppress warnings on, exactly? The issue you've linked from this PR doesn't have any warnings for this method, so it seems like this PR is not for the errors described in the issue.

juliansteenbakker: `cookieManager.removeAllCookie();` gives a deprecation warning. I didn't got this warning with my personal project which is where the log is coming from in the issue. I did however get this warning when building the example app. 

I tried suppressing the call only, however it doesn't let me do that. 

image

stuartmorgan: @blasten Is there a way to suppress an individual line that doesn't declare a variable?

(Failing that, what we should do is extract the legacy part of the if/else to a dedicated function, and suppress warnings on it.)

blasten: no, it only works at the class/method level

blasten: and field level

blasten (changes requested):
Any chance we could migrate to the new API instead of suppressing warnings?

There are some instructions in the docs. For example: https://developer.android.com/reference/android/webkit/WebViewClient#shouldOverrideUrlLoading(android.webkit.WebView,%20java.lang.String)

juliansteenbakker:
> Any chance we could migrate to the new API instead of suppressing warnings?

The new function is already implemented. See https://github.com/flutter/plugins/blob/47192b979a51c1ac1cba637c33cfdab79b0e5a50/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java#L51

The deprecated function is still there for backwards compatibility. It still however produces a warning in the log. That's why i annotated it since it's already been replaced.

packages/webview_flutter/webview_flutter_android/CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
+## 2.2.2
+
+* Annotate deprecated functions to suppress build warnings.
+
 ## 2.2.1
 
 * Fix `NullPointerException` from a race condition when changing focus. This only affects `WebView`
 when it is created without Hybrid Composition.
-
+
```
stuartmorgan, line 9: Please don't add whitespace to blank lines.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
stuartmorgan: Okay, good to know. In that case we should extract the legacy codepath to a function per my alternate suggestion above.

stuartmorgan (changes requested):

stuartmorgan:
@juliansteenbakker Are you still planning on updating this PR based on the last round of reviews?

juliansteenbakker:
> @juliansteenbakker Are you still planning on updating this PR based on the last round of reviews?

Yes, i will sort this out next week.

stuartmorgan:

juliansteenbakker:
I tried merging it yesterday but lots has changed in the base code of this plugin so i have to recheck it all. If i do i'll create a new PR. sorry to keep you waiting so lang.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
bswhite1: Why not fix the problem rather than hiding it?

https://developer.android.com/reference/android/webkit/CookieManager#public-constructors
says to use this:
https://developer.android.com/reference/android/webkit/CookieManager#getInstance()

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebView.java
```diff
@@ -57,6 +57,7 @@ public boolean shouldOverrideUrlLoading(
               return true;
             }
 
+            @SuppressWarnings("deprecation")
```
bswhite1, line 60: Why not fix the problem rather than hiding it? This is a global statement for this entire pull request

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterCookieManager.java
stuartmorgan: > Why not fix the problem rather than hiding it?

Supporting older versions of Android is a feature, not a problem. The only problem here is the warning on the legacy branch.

This has in fact already been discussed in this PR.

> https://developer.android.com/reference/android/webkit/CookieManager#public-constructors says to use this: https://developer.android.com/reference/android/webkit/CookieManager#getInstance()

Please read the code in question before leaving review comments.

PR #4465, opened 2021-11-01
[path_provider] Federate mobile implementations

stuartmorgan:
Fully federates the plugin by moving the existing mobile implementations to their own packages, per planned repo structure.

Temporarily marks `path_provider` as unpublishable to allow the implementations to be moved, rather than copied and deleted, in order to better preserve git history. A follow-up PR will restore it to publishable form.

Part of https://github.com/flutter/flutter/issues/68498

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
For ease of review, the individual commits are:
1. Copy `path_provider` as-is to the `_android` and `_ios` versions, deleting the files that are no longer relevant from each package.
2. Fix all of the pubspecs and CHANGELOGs
3. Fix the Android and iOS tests to use the platform interface instead of the app-facing package (per standard repo practice, to avoid dependencies on a higher layer)
4. Adjust the podspec and gradle files for the new names

packages/path_provider/path_provider_android/README.md
```diff
@@ -0,0 +1,29 @@
+# path_provider
```
bparrishMines, line 1: This `README` is still the exact same as the one in `path_provider`. Is the plan to fix this in the followup PR?

packages/path_provider/path_provider_ios/README.md
```diff
@@ -0,0 +1,29 @@
+# path_provider
```
bparrishMines, line 1: Same as Android implementation README

packages/path_provider/path_provider_android/example/README.md
```diff
@@ -0,0 +1,8 @@
+# path_provider_example
```
bparrishMines, line 1: nit: The typical pattern of the example apps of plugins would name this `path_provider_android_example`

bparrishMines: I don't think its that important to keep the pattern unless one of our tests uses this pattern to look for test

stuartmorgan: There shouldn't be anything that depends on it. I'm going to leave this as-is since I'm not 100% sure that changing it to something that doesn't match the existing native code won't cause issues. (Normally the name of everything would be auto-generated to be right, but I wanted to leave exactly the same native runner that we already know works rather than re-generate, as this is intended to be as close to a file move as possible.)

packages/path_provider/path_provider_android/README.md
stuartmorgan: Oops, that's just copy-pasta. Thanks for catching that!

packages/path_provider/path_provider_ios/README.md
stuartmorgan: Fixed.

packages/path_provider/path_provider_android/README.md
stuartmorgan: Fixed. It's now using exactly the same format as the existing implementation package READMEs.

(It says they are endorsed, which isn't technically true at the moment they are published, but will be immediately after since they aren't going to have a deliberately "available as unendorsed" period as we've sometimes done with plugins (e.g., on web), so having them say it now avoids pointless churn.)

bparrishMines (approved):
LGTM

PR #4469, opened 2021-11-03
[url_launcher] Safari Modal Presentation Style

ABausG:
Adding support for `UIModalPresentationStyle` for url_launcher to change the transition and presentation style for Safari ViewControllers 

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan (commented):
Thanks for the submission! Per my comment below this requires non-trivial changes to the plugin API before we would go forward with it, so I'm going to add this to the backlog for now. If you are interested in driving that work forward, let me know, otherwise this will need to wait until that happens.

ABausG:
@stuartmorgan Thanks for taking a look! I understand and agree with the idea of restructuring the api. 
Looking at [your current proposal](https://github.com/stuartmorgan/plugins/blob/url-launcher-uri/packages/url_launcher/url_launcher/lib/url_launcher.dart) would you think the functionality this PR adds as an option in something like an extra `iOSConfiguration` Object similar to `WebConfiguration` or rather add a `LaunchMode` for presenting something modally?

stuartmorgan:
> @stuartmorgan Thanks for taking a look! I understand and agree with the idea of restructuring the api. Looking at [your current proposal](https://github.com/stuartmorgan/plugins/blob/url-launcher-uri/packages/url_launcher/url_launcher/lib/url_launcher.dart) would you think the functionality this PR adds as an option in something like an extra `iOSConfiguration` Object similar to `WebConfiguration` or rather add a `LaunchMode` for presenting something modally?

I think it would fit best as an entry in `WebConfiguration`, since that's already a configuration object that's specific to web URLs and, in most cases, to inline display.

It would probably be best for it to be translated to less OS-specific terms, since there's no reason another OS implementation couldn't support one or more of these display styles. (It's also not clear to me that actually supporting all of these is particularly useful; how many of these presentation styles do people actively want to use for `url_launcher`?)

ABausG:
> It's also not clear to me that actually supporting all of these is particularly useful; how many of these presentation styles do people actively want to use for `url_launcher`?

I guess it might be sufficient to only provide 2 options that show the SafariVC either as a modal popover or as is sliding in from the side (current Implementation when `forceSafariVC = false`) So might be possible to have it as a boolean option in the direction of `presentAsModal`. I am not 100% what the actual differences in the PresentationModes on iOS are but they seem pretty similar to me visually

stuartmorgan:
Since this would need to be significantly restructured once the plugin's API is reworked, I don't think there's any reason to leave this PR open in the meantime. Given that I'm going to close this, but please do submit a new version once `url_launcher`'s API is reworked.

I would strongly encourage you to file an issue requesting this feature to track interest, and to refer to this PR as a starting point so it's easy for someone else who might be interested in this functionality to find this code and discussion again in the future.

PR #4472, opened 2021-11-03
[url_launcher] Improving unit tests on mac

igorpolt:
Add unit tests for url_launcher canLaunch and launch methods on mac with success and failure results.

Fixes #[92969]

## Pre-launch Checklist

- [x] I signed the [CLA].
- [x] I updated CHANGELOG.md to add a description of the change, following repository CHANGELOG style.

[92969]: https://github.com/flutter/flutter/issues/92969


packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,106 @@ import FlutterMacOS
...
+
+}
+
 class RunnerTests: XCTestCase {
-  func testCanLaunch() throws {
-    let plugin = UrlLauncherPlugin()
+
+  var workspace: StubWorkspace! = nil
+  var pluginWithStubWorkspace: UrlLauncherPlugin! = nil
+  var plugin: UrlLauncherPlugin! = nil
```
stuartmorgan, line 27: Please don't make stateful test fixtures without a compelling reason, as they make individual tests harder to understand. Making the object under test stateful is in particular an anti-pattern. (The fact that there are two different versions of it and each test only uses one makes this particularly problematic.)

It is much better to have a couple of extra lines of boilerplate than to make people go look elsewhere to understand the core setup of the test.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,106 @@ import FlutterMacOS
 import XCTest
 import url_launcher_macos
 
+class StubWorkspace: URLLauncher {
+  var isSuccessful = true
+  var url = URL.init(string: "https://flutter.dev")
```
stuartmorgan, line 11: Tests would be clearer if there were no default value for this, so that it's clear in the context of a test what the result will be. Also, the fact that you've made this non-nullable means it's impossible to test the (common) case where there's no handler for a URL.

(Also, this is a very strange value to use, since an application's URL is never going to be a website.)

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,106 @@ import FlutterMacOS
...
+  func testCanLaunchSuccessTrue() throws {
     let call = FlutterMethodCall(
       methodName: "canLaunch",
       arguments: ["url": "https://flutter.dev"])
-    var canLaunch: Bool?
+
+    plugin.handle(
+      call,
+      result: { (result: Any?) -> Void in
+        XCTAssertNotNil(result)
```
stuartmorgan, line 49: You're already asserting that the cast result isn't nil due to the use of `as?`; why not cast to `Bool?` instead and avoid this check?

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,106 @@ import FlutterMacOS
...
+        let canLaunch = result! as? Bool
+        XCTAssertNotNil(canLaunch)
+        XCTAssertTrue(canLaunch!)
+      })
+  }
+
+  func testCanLaunchFailureFalse() throws {
```
stuartmorgan, line 56: I'm not sure what these names are meant to indicate exactly. What is the distinction between "Success/Failure" and "True/False"? The fact that you only have "SuccessTrue" and "FailureFalse" suggests that there isn't one.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,106 @@ import FlutterMacOS
...
+    let call = FlutterMethodCall(
+      methodName: "canLaunch",
+      arguments: ["url": "brokenUrl"])
```
stuartmorgan, line 59: The naming also hides the fact that the coverage here is missing important cases. If you called this test method something more descriptive like "testCanLaunchInvalidUrl", then it would be more obvious that you are missing the much more common "false" case of "testCanLaunchNoHandler".

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,106 @@ import FlutterMacOS
...
+      call,
+      result: { (result: Any?) -> Void in
+        XCTAssertNotNil(result)
+        let canLaunch = result! as? Bool
+        XCTAssertNotNil(canLaunch)
+        XCTAssertFalse(canLaunch!)
+      })
+  }
+
+  func testCanLaunchFailureError() throws {
```
stuartmorgan, line 71: Similarly, this suggests there can only ever be one possible way to get an error.

Test case names should describe the conditions they are testing, not (or at least, not just) the expected result. `testCanLaunchMissingArgumentReturnsError` is a much better name, and leaves the potential to test other error conditions.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,106 @@ import FlutterMacOS
...
+
+    pluginWithStubWorkspace.handle(
+      call,
+      result: { (result: Any?) -> Void in
+        XCTAssertNotNil(result)
+        let launch = result! as? Bool
+        XCTAssertNotNil(launch)
       })
+  }
+
```
stuartmorgan, line 97: Per the comment above, there's a test case missing here where there's no handler. (That's a much more important case than "the arguments are missing", which can only ever happen if we break the Dart side of the plugin and thus should never actually run.)

packages/url_launcher/url_launcher_macos/macos/Classes/UrlLauncherPlugin.swift
```diff
@@ -5,7 +5,21 @@
 import FlutterMacOS
 import Foundation
 
+public protocol URLLauncher {
```
stuartmorgan, line 8: This needs a comment.

packages/url_launcher/url_launcher_macos/macos/Classes/UrlLauncherPlugin.swift
```diff
@@ -5,7 +5,21 @@
...
+  func open(_ url: URL) -> Bool
+  func urlForApplication(toOpen: URL) -> URL?
```
stuartmorgan, line 10: As do these.

packages/url_launcher/url_launcher_macos/macos/Classes/UrlLauncherPlugin.swift
stuartmorgan, line 8: This is a potentially confusing name given that it's the name of the plugin. How about `SystemURLHandler`?

stuartmorgan (changes requested):
Thanks for the tests! Some comments inline, but overall the approach here looks good.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
igorpolt: Thank you, fixed.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
igorpolt: Test case names fixed

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
igorpolt: fixed, now its checking that result should be FlutterMethodNotImplemented

packages/url_launcher/url_launcher_macos/macos/Classes/UrlLauncherPlugin.swift
igorpolt: added

packages/url_launcher/url_launcher_macos/macos/Classes/UrlLauncherPlugin.swift
igorpolt: added

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,106 @@ import FlutterMacOS
 import XCTest
 import url_launcher_macos
 
+class StubWorkspace: URLLauncher {
```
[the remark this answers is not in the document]
igorpolt: added

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
igorpolt: renamed according to the tested conditions.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,106 @@ import FlutterMacOS
...
+  var isSuccessful = true
+  var url = URL.init(string: "https://flutter.dev")
+
+  func open(_ url: URL) -> Bool {
```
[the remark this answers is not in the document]
igorpolt: sorry, i don't understand, i don't override anything here, it's just implementation.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
igorpolt: removed all the NotNil asserts.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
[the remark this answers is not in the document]
stuartmorgan: Sorry, I'm mixing up languages; I do a lot more C++ than Swift, and the distinction doesn't exist there.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,105 @@ import FlutterMacOS
...
+}
+
 class RunnerTests: XCTestCase {
-  func testCanLaunch() throws {
+
+  func testNoHandlerReturnsFlutterMethodNotImplemented() throws {
+    let plugin = UrlLauncherPlugin()
+
+    let call = FlutterMethodCall(
+      methodName: "noHandler",
```
stuartmorgan, line 29: It looks like there was a misunderstanding here. The test I was suggesting was not calling an invalid method (that's not a particularly useful test, since it should never happen), it's the common case of calling `canLaunch` with a valid URL for which there is no handler registered. E.g., a `whatsapp:` URL on a device without WhatsApp installed.

The whole point of the `canLaunch` method is that not every URL can be launched on every device. If the only possible return value for a valid URL were `true` (as the current test suite here suggests), then the method wouldn't exist.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,105 @@ import FlutterMacOS
...
+    let call = FlutterMethodCall(
+      methodName: "launch",
+      arguments: ["url": "https://flutter.dev"])
+
+    pluginWithStubWorkspace.handle(
+      call,
+      result: { (result: Any?) -> Void in
+        XCTAssertEqual(result as? Bool, true)
       })
+  }
```
stuartmorgan, line 95: Similarly, `launch` can fail in exactly the same way. You've set up the mock handler to have the ability to return false, but you're not ever using it to write a test for launch failure.

stuartmorgan (changes requested):
Comments below.

I've added specific comments about missing tests, but as a general note: the primary purpose of plugin native unit tests is to cover, as much as possible, the cases that *clients of the plugin* can create.

Tests of cases that can only happen if *we* break the Dart side of the plugin have very limited value; it's not that we shouldn't have them, but they are not the point, and should not be the focus of the test suite.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
igorpolt: yes, it was misunderstanding from my side. I've fixed that.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
igorpolt: Thank you, now the stub is in use for testing when `launch` cannot open the URL.

packages/url_launcher/url_launcher_macos/macos/Classes/UrlLauncherPlugin.swift
```diff
@@ -5,7 +5,26 @@
 import FlutterMacOS
 import Foundation
 
+/// A handler that can launch other apps, check if any app is able to open the URL.
+public protocol SystemURLHandler {
+
+  /// Opens the location at the specified URL.
```
stuartmorgan, line 11: This needs to document the return value. https://google.github.io/swift/#parameter-returns-and-throws-tags

packages/url_launcher/url_launcher_macos/macos/Classes/UrlLauncherPlugin.swift
```diff
@@ -5,7 +5,26 @@
...
+  func open(_ url: URL) -> Bool
+
+  /// Returns the URL to the default app that would be opened.
```
stuartmorgan, line 14: And this needs to document the argument.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,121 @@ import FlutterMacOS
...
     let call = FlutterMethodCall(
       methodName: "canLaunch",
       arguments: ["url": "https://flutter.dev"])
-    var canLaunch: Bool?
+
+    plugin.handle(
+      call,
+      result: { (result: Any?) -> Void in
+        XCTAssertEqual(result as? Bool, true)
+      })
```
stuartmorgan, line 36: Removing `result` calls from the implementation would not fail any of these tests as you have written them, because the only assertion is in the callback.

The usual pattern for our tests is that you create [an expectation](https://developer.apple.com/documentation/xctest/asynchronous_tests_and_expectations/testing_asynchronous_operations_with_expectations?language=objc) that is `fulfill`ed in the `result` callback.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -6,19 +6,121 @@ import FlutterMacOS
...
+  }
+
+  func testCanLaunchNoAppIsAbleToOpenUrlReturnsFalse() throws {
+    let plugin = UrlLauncherPlugin()
+
+    let call = FlutterMethodCall(
+      methodName: "canLaunch",
+      arguments: ["url": "example://flutter.dev"])
```
stuartmorgan, line 44: It doesn't seem impossible that someone could have such a scheme when running tests locally. If you're not going to use the fake here, use a much more distinct scheme like `schemethatdoesnotexist`.

stuartmorgan (changes requested):

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
igorpolt: The URL scheme has been changed.

packages/url_launcher/url_launcher_macos/example/macos/RunnerTests/RunnerTests.swift
igorpolt: Thank you for the advice. All the tests now should fulfill an expectation in the completion handler.

packages/url_launcher/url_launcher_macos/macos/Classes/UrlLauncherPlugin.swift
igorpolt: Added.

packages/url_launcher/url_launcher_macos/macos/Classes/UrlLauncherPlugin.swift
igorpolt: Added.

stuartmorgan (changes requested):
All the tests look good now, this just needs a version bump since there are changes to actual (non-test) plugin logic. In theory it should be a no-op functionally, but if there were a bug that caused unexpected effects having it in its own version would simplify things.

stuartmorgan (changes requested):
Part of a version bump is updating the CHANGELOG, which is why `publishable` is failing.

packages/url_launcher/url_launcher_macos/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 ## NEXT
 
 * Updates code for new analysis options.
+* Update unit tests.
+
+## 2.0.3
+
+* Update unit tests.
```
stuartmorgan, line 8: Per the bot failure:
>  When bumping the version for release, the NEXT section should be incorporated into the new version's release notes.

You need to change `NEXT` to `2.0.3`, and add the new line, not add a new section. Everything that was in `NEXT` is by definition part of this version bump, because this is the next version.

Also, it's "Updates", not "Update", per the repo style guide for CHANGELOGs.

(I would just fix this, but the branch is in your org rather than a personal fork, which prevents me from committing to it)

stuartmorgan (changes requested):

stuartmorgan:
@cbracken for secondary review, per Flutter policy.

packages/url_launcher/url_launcher_macos/CHANGELOG.md
igorpolt: Thank you very much for your help.

stuartmorgan (approved):
LGTM

cbracken (approved):
LGTM stamp from a Japanese personal seal

stuartmorgan:
GitHub has gotten very aggressive about requiring `main` to be merged in before landing even when there are no conflicts (I'm not sure why), so this needs another merge. If you ping me once that's done I can try to land it before it gets stale again.

igorpolt:
ping

PR #4473, opened 2021-11-04
[ci] Add 'main' builders in bringup mode

stuartmorgan:
Creates `main` versions of the existing `master` CI builders, as the
first half of switching LUCI tests over to the `main` branch.

Part of https://github.com/flutter/flutter/issues/90476

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -48,6 +48,20 @@ targets:
         ]
     scheduler: luci
 
+  - name: Windows win32-platform_tests main
+    bringup: true
+    recipe: plugins/plugins
+    timeout: 30
+    properties:
+      add_recipes_cq: "true"
+      target_file: windows_build_and_platform_tests.yaml
+      channel: main
+      dependencies: >
+        [
+          {"dependency": "vs_build"}
+        ]
+    scheduler: luci
+
   - name: Windows win32-platform_tests stable
     recipe: plugins/plugins
     timeout: 30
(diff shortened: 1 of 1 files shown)


stuartmorgan:
I'll wait on landing this until we actually have the branch mirroring, but I was updating the tool and though I'd go ahead and prep this at the same time.

CaseyHillers:
\cc @godofredoc whose leading the migrations.

I think instead of adding new builders, we would want to convert the existing ones to main once we have mirroring set up.

stuartmorgan:
How would we validate that there aren't problems (e.g., repo tooling bugs) with running on `main` without doing a two-phase conversion where we bring up bots using `main`, ensure they work, then eliminate the `master` versions?

CaseyHillers:
Since the mirrors are pretty close to instant, the way we validate these tooling issues is by doing scheduled outages where we rename master to master_archived for short periods of time. In practice, the parallel runs catch very few issues compared to the outage approach.

godofredoc:
> he way we validate these tooling issues is by doing scheduled outages where we rename master to master_archived for short periods of time. In practice, the parallel runs catch very few issues compared to the outage approach.

@stuartmorgan do the luci builders generate artifacts that can clash when using the same commit and builders coming from two different branches?

stuartmorgan:
> @stuartmorgan do the luci builders generate artifacts that can clash when using the same commit and builders coming from two different branches?

No, our LUCI builders do not generate artifacts. (Nor do our Cirrus builders, actually; the closest thing we have to artifacts is the GitHub Action auto-publisher; we can do that one as a single switch)

godofredoc:
> lly; the closest thing we have to artifacts is the GitHub Action auto-publisher; we can do that one as a single switch)

If that is the case we can enable parallel builders. The only reason we are doing a one time swap in the engine is because they are generating artifacts that get duplicated when running parallel builds.

CaseyHillers (approved):
flutter-lgtm

PR #4474, opened 2021-11-04
[flutter_plugin_tools] Add 'main' support

stuartmorgan:
Treat `main` the same as `master` for branch-based switching, in
preparation for switching the branch names in Flutter repositories.

Also updates all of the tests that used `master` as the explicit base to
use `main` instead; what the tests use is arbitrary, so they can be
switched now even though the repo itself hasn't switched.

Part of https://github.com/flutter/flutter/issues/90476

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/version_check_command.dart
+++ b/script/tool/lib/src/version_check_command.dart
@@ -262,7 +262,9 @@ ${indentation}HTTP response: ${pubVersionFinderResponse.httpResponse.body}
     // This method isn't called unless `version` is non-null.
     final Version currentVersion = pubspec.version!;
     Version? previousVersion;
+    String previousVersionSource;
     if (getBoolArg(_againstPubFlag)) {
+      previousVersionSource = 'pub';
       previousVersion = await _fetchPreviousVersionFromPub(pubspec.name);
       if (previousVersion == null) {
         return _CurrentVersionState.unknown;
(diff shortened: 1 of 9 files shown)


script/tool/lib/src/version_check_command.dart
```diff
@@ -273,6 +275,7 @@ ${indentation}HTTP response: ${pubVersionFinderResponse.httpResponse.body}
       }
     } else {
       final GitVersionFinder gitVersionFinder = await retrieveVersionFinder();
+      previousVersionSource = await gitVersionFinder.getBaseSha();
```
gaaclarke, line 278: This is a logical change since getBaseSha() isn't necessarily `master` (or `main`), right?  This sounds like the correct thing to do.  It might be worth mentioning this a bit more explicitly in the description in case it does cause some problems.

gaaclarke (approved):

script/tool/lib/src/version_check_command.dart
stuartmorgan: > This is a logical change since getBaseSha() isn't necessarily `master` (or `main`), right?

In that the error logging was wrong before, and now isn't, yes. `getBaseSha()` is what the code that populates `previousVersion` actually uses; the error logging on failure (which is all this is used for) was incorrectly hard-coding that to `master`. In practice, in presubmit, it's the same thing, so it hadn't been noticed.

> It might be worth mentioning this a bit more explicitly in the description in case it does cause some problems.

It would only be a problem if something were parsing human-targeted error output (which we aren't, and we only guarantee this tool's behavior for our use currently), and relying on the *exact* output, and were using a base revision other than master.

PR #4485, opened 2021-11-08
[path_provider] Fix the comment of `getApplicationDocumentsDirectory` method

PollexLee:
https://github.com/flutter/flutter/issues/93245

`getDataDirectory` should be `getDir`.

Code in the plugin:
https://github.com/flutter/plugins/blob/5bf7fb0360aa70571c46d5b0ac64949fc81a0dfd/packages/path_provider/path_provider/android/src/main/java/io/flutter/plugins/pathprovider/PathProviderPlugin.java#L79
```kotlin
  private String getPathProviderApplicationDocumentsDirectory() {
    return PathUtils.getDataDirectory(context);
  }
```

PathUtils code in Flutter engine:
https://github.com/flutter/engine/blob/1509b7b6b8313ebe0f7e216ddd59f9ccae399ea1/shell/platform/android/io/flutter/util/PathUtils.java#L15
```kotlin
  public static String getDataDirectory(Context applicationContext) {
    return applicationContext.getDir("flutter", Context.MODE_PRIVATE).getPath();
  }
```

The `getDataDirectory` in the comment is inaccurate, the correct one should be `getDir`.

## Pre-launch Checklist

- [x] I signed the [CLA].


Hixie:
test-exempt: only affects comments

stuartmorgan (changes requested):
Thanks for the submission! The real issue here, as demonstrated by the fact that it has become incorrect, is that this is documenting specific implementation details of code it doesn't control. Instead of repeating the same mistake again, which will likely lead to this becoming wrong again in the future, this should be changed to say something more generic about it using application-scoped storage.

Also:
> * [ ]  I [updated pubspec.yaml](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates) with an appropriate new version according to the [pub versioning philosophy].
> * [ ]  I updated CHANGELOG.md to add a description of the change.

These steps are missing.

PollexLee:
@stuartmorgan 
In my opinion, it is precisely because this is documenting specific implementation details of code it doesn't control, Therefore, the correctness of this part of the document is more important.

The document here only describe the implementation on the Android side, and there are more detailed comments on it.

I deleted the `Pre-launch Checklist` because I only corrected the document.

stuartmorgan (changes requested):
> @stuartmorgan thank you for your reply, in my opinion, it is precisely because this is documenting specific implementation details of code it doesn't control,

Sorry, I'm not following what you are arguing here. The purpose of documentation is to provide accurate information; a comment in this plugin about what the engine's implementation is **cannot** reliably do that in this case, because what it is describing is an internal implementation detail of another codebase.

Comments asserting things about code outside of the code that they are commenting is an anti-pattern. The need for this PR is a clear illustration of exactly why this is a bad idea. We should not perpetuate that anti-pattern, we should fix it.

> Therefore, the correctness of this part of the document is more important.

How do you know the engine implementation won't change next week, rendering your proposed text inaccurate again?

> I deleted the `Pre-launch Checklist` because I only corrected the document.

The checklist applies to all PRs, even if not every item on it needs to be done. If you click the link in the part I quoted, you'll see that this PR is not exempt from those steps. (I'm not sure why you are assuming that those steps are unnecessary when I explicitly pointed out that they were.)

PollexLee:
@stuartmorgan 
Thank you for your reply, you are correct, we cannot guarantee whether the engine implementation will change frequently.

I updated CHANGELOG.md to add a description of this change.

packages/path_provider/path_provider/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## NEXT
```
stuartmorgan, line 1: Please read the link I pointed you to again, and re-read my comments about the missing steps. This needs a version change.

packages/path_provider/path_provider/lib/path_provider.dart
```diff
@@ -128,7 +128,7 @@ Future<Directory> getLibraryDirectory() async {
 /// On iOS, this uses the `NSDocumentDirectory` API. Consider using
 /// [getApplicationSupportDirectory] instead if the data is not user-generated.
 ///
-/// On Android, this uses the `getDataDirectory` API on the context. Consider
+/// On Android, this uses the `getDir` API on the context. Consider
```
stuartmorgan, line 131: > Thank you for your reply, you are right, we cannot guarantee whether the engine implementation will change frequently.

I'm glad we've reached agreement, but you have not actually addressed my feedback about changing this comment accordingly.

packages/path_provider/path_provider/CHANGELOG.md
PollexLee: OK, i updated changelog.md and pubspace.yaml, add a new version, please review.

packages/path_provider/path_provider/lib/path_provider.dart
PollexLee: Here are general instructions for this method:   
https://github.com/PollexLee/plugins/blob/aa242b45a1c50c212a4e85e8f4def67ee4f7384d/packages/path_provider/path_provider/lib/path_provider.dart#L125

I think the implementation method of the native side in the comment is to facilitate the development of native programmers.

About how to fix the anti-pattern, i have no idea. 😂

stuartmorgan: As I said in my initial review:

> this should be changed to say something more generic about it using application-scoped storage.

stuartmorgan: Are you planning on updating the PR to address this feedback?

PollexLee: Sorry, I am not sure how to address this feedback.

stuartmorgan: I'm saying that instead of saying "this uses the `getDir` API on the context", it should say something like "this uses application-scoped storage". That provides context for the recommendation in the next sentence, without tying it to implementation details that aren't actually controlled by the plugin.

stuartmorgan:

PR #4492, opened 2021-11-09
[flutter_plugin_tools] Build gtest unit tests

stuartmorgan:
Builds Windows and Linux unit tests before running them in
`native-test`, in order to match the behavior of other platforms and
significantly improve the UX of running the command locally while
changing tests.

Fixes https://github.com/flutter/flutter/issues/89303

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/url_launcher/url_launcher/example/linux/flutter/CMakeLists.txt
```diff
@@ -78,7 +78,8 @@ add_custom_command(
   COMMAND ${CMAKE_COMMAND} -E env
     ${FLUTTER_TOOL_ENVIRONMENT}
     "${FLUTTER_ROOT}/packages/flutter_tools/bin/tool_backend.sh"
-      linux-x64 ${CMAKE_BUILD_TYPE}
+      ${FLUTTER_TARGET_PLATFORM} ${CMAKE_BUILD_TYPE}
+  VERBATIM
```
stuartmorgan, line 82: These two lines are auto-migrations by `flutter` that have just never been checked in.

packages/url_launcher/url_launcher/example/windows/flutter/generated_plugin_registrant.cc
```diff
@@ -6,9 +6,9 @@
...
 #include "generated_plugin_registrant.h"
 
-#include <url_launcher_windows/url_launcher_plugin.h>
+#include <url_launcher_windows/url_launcher_windows.h>
 
 void RegisterPlugins(flutter::PluginRegistry* registry) {
-  UrlLauncherPluginRegisterWithRegistrar(
-      registry->GetRegistrarForPlugin("UrlLauncherPlugin"));
+  UrlLauncherWindowsRegisterWithRegistrar(
+      registry->GetRegistrarForPlugin("UrlLauncherWindows"));
```
stuartmorgan, line 13: Apparently I forgot to check in the updated generated registrant when I restructured the plugin for unit testability.

script/tool/lib/src/common/cmake.dart
```diff
@@ -0,0 +1,112 @@
...
+  /// The platform that commands are being run on.
+  final Platform platform;
+
+  /// The build mode (e.g., Debug, Release).
+  ///
+  /// This is a constructor paramater because on Linux many properties depend
+  /// on the build mode since it uses a single-configuration generator.
+  final String buildMode;
+
+  String? _cmakeCommand;
```
gaaclarke, line 45: nit: I don't believe you need an instance variable for this.  The code would be easier to maintain if it was just defined in `getCMakeCommand`

script/tool/lib/src/common/cmake.dart
```diff
@@ -0,0 +1,112 @@
...
+    return buildDir;
+  }
+
+  File get _cmakeCacheFile => buildDirectory.childFile('CMakeCache.txt');
+
+  /// Returns the CMake command to run build commands for this project.
+  ///
+  /// Assumes the project has been built at least once, such that the CMake
+  /// generation step has run.
+  String getCMakeCommand() {
```
gaaclarke, line 72: In dart you should use properties instead of having a `get` method.

script/tool/test/native_test_command_test.dart
```diff
@@ -53,6 +55,16 @@ final Map<String, dynamic> _kDeviceListMap = <String, dynamic>{
   }
 };
 
+const String _fakeCmakeCommand = 'path/to/cmake';
+
+void _createFakeCmakeCache(Directory pluginDir, Platform platform) {
```
gaaclarke, line 60: Inconsistent camel casing, above it is "CMakeProject" here it is "CmakeCache".

gaaclarke (approved):
LGTM with just a few minor observations.

script/tool/lib/src/common/cmake.dart
stuartmorgan: The Flutter style guide [disagrees](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#getters-feel-faster-than-methods). Loading and parsing a file isn't a trivial operation.

script/tool/lib/src/common/cmake.dart
stuartmorgan: I'm not sure I follow; the purpose of a cache isn't simplicity, it's efficiency. Are you saying it's bad to cache the results of parsing a file?

script/tool/test/native_test_command_test.dart
stuartmorgan: Yep, I'll fix this. I keep messing this up because I'm mostly interacting with the `cmake` command so usually don't think about it being capitalized.

script/tool/lib/src/common/cmake.dart
gaaclarke: Oof seems like a bad choice mixing implementation details with semantics, sgtm.

script/tool/lib/src/common/cmake.dart
gaaclarke: Yep, mutable data and caching can lead to bugs and should be avoided.  If you want to cache for performance I'd recommend a memoization pattern:

```dart
lazy int _memoizedGetHash => _calculateHash();
int getHash() {
  return _memoizedGetHash;
}
```

stuartmorgan: I'm not familiar with `lazy`, and DartPad doesn't think it's valid Dart. Do you have more info on this?

gaaclarke: Sorry, got my syntax wrong, lazy values are 'late' in Dart:
```dart
late final int _memoizedGetHash => _calculateHash();
int getHash() {
  return _memoizedGetHash;
}
```

stuartmorgan: Oh, nice! I had no idea `late` had that second meaning.

script/tool/test/native_test_command_test.dart
stuartmorgan: Fixed; the only inconsistent version left should be where I'm specifically referring to the command `cmake`, where the lack of capitalization is deliberate.

PR #4502, opened 2021-11-11
Bump plugin Android compileSdkVersions to 31

camsim99:
Bumps compileSdkVersion of the following plugin example apps to 31: `android_alarm_manager`, `android_intent`, `battery`, `connectivity`, `device_info`, `espresso`, `flutter_plugin_android_lifecycle`, `google_maps_flutter`, `google_sign_in`, `image_picker`, `in_app_purchase`, `local_auth`, `package_info`, `path_provider`,  `quick_actions`, `sensors`, `share`, `share_preferences`, `url_launcher`,  `video_player`, `webview_flutter`, `wifi_info_flutter`.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/quick_actions/quick_actions/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
```diff
@@ -49,7 +52,18 @@ public void onMethodCall(MethodCall call, MethodChannel.Result result) {
       case "setShortcutItems":
         List<Map<String, String>> serializedShortcuts = call.arguments();
         List<ShortcutInfo> shortcuts = deserializeShortcuts(serializedShortcuts);
-        shortcutManager.setDynamicShortcuts(shortcuts);
+        ThreadPoolExecutor executor =
```
camsim99, line 55: **Note:** `setDynamicShortcuts(...)` call moved to worker thread to fix CI failure.

packages/path_provider/path_provider_android/android/src/main/java/io/flutter/plugins/pathprovider/PathProviderPlugin.java
```diff
@@ -243,7 +243,7 @@ private String getPathProviderStorageDirectory() {
   }
 
   private List<String> getPathProviderExternalCacheDirectories() {
-    final List<String> paths = new ArrayList<>();
+    final List<String> paths = new ArrayList<String>();
```
camsim99, line 246: **Note:** Type added to fix warnings given by `flutter build apk`.

camsim99:
### A couple of notes
- The following deprecated plugins have been bumped (currently) and have unfixed deprecation warnings: `android_alarm_manager`, `wifi_info_flutter`, `battery` (relies on v1 Android embedding).
-  The following deprecated plugins have also been bumped (currently): `android_intent`, `connectivity`, `device_info`, `package_info`, `sensors`, `share`.

Piinks (approved):
This LGTM, but I don't frequent this neck of the woods, so another reviewer may have better insight. :)

packages/device_info/device_info/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.4
```
stuartmorgan, line 1: `## NEXT`

We shouldn't be pushing updates to deprecated plugins unless it's critical.

packages/flutter_plugin_android_lifecycle/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.5
```
stuartmorgan, line 1: `## NEXT`

packages/in_app_purchase/in_app_purchase/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.10
```
stuartmorgan, line 1: Where the change is only to `example` (so, app-facing plugins with federated implementations), the change is exempt from version updates, because it doesn't affect clients.

packages/quick_actions/quick_actions/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.6.0+8
```
stuartmorgan, line 1: `## NEXT`

packages/shared_preferences/shared_preferences/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## NEXT
 
 * Fixes newly enabled analyzer options.
+* Updated Android compileSdkVersion to 31.
```
stuartmorgan, line 4: When a change that should be published happens to a plugin that already has `NEXT`, the `NEXT` should be updated to be a version bump.

As a high-level note, since I think there's some confusion here based on the edits in this PR: whether or not a PR bumps the version or uses `NEXT` is never a function of what's already in the CHANGELOG, it's a determination based on the current PR: If the change is exempt from version updates (per the list linked from the PR checklist), then you should use `NEXT`, and if not, you must bump the version (updating any existing `NEXT` entry).

packages/webview_flutter/webview_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.3.2
```
stuartmorgan, line 1: `## NEXT`

packages/quick_actions/quick_actions/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+* Updated Android compileSdkVersion to 31.
```
stuartmorgan, line 3: This should note the code change.

stuartmorgan (changes requested):
I left inline comments, but once I got to the end I realized they are probably not all correct. Aren't *all* of the gradle changes version-exempt? All the changes to `example`s are, but my understanding is that the `compileSdkVersion` that's in the actual plugin's `build.gradle` actually has no effect on clients. (/cc @blasten to sanity-check me here).

If that's the case, then only changes to the Java files need version bumps (and one of those doesn't, because it's to a deprecated plugin where almost everything is currently version-exempt).

stuartmorgan:
This will need a test exemption from @Hixie. I think this should qualify, because the motivation for this is a Flutter change that broke our tree and was temporarily reverted, so the test will be that we aren't broken when that re-lands.

blasten:
>  my understanding is that the compileSdkVersion that's in the actual plugin's build.gradle actually has no effect on clients.

That is right. Although, Camille added a warning to the Flutter tool, so now it will show something.

packages/android_alarm_manager/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Remove support for the V1 Android embedding.
 * Updated Android lint settings.
 * Removed `-Werror` in Android builds.
+* Updated Android compileSdkVersion to 31.
```
blasten, line 6: looks like present tense is preferred in changelogs. https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style

stuartmorgan:
> > my understanding is that the compileSdkVersion that's in the actual plugin's build.gradle actually has no effect on clients.
> 
> That is right. Although, Camille added a warning to the Flutter tool, so now it will show something.

Hm, do we actually want to bump those versions then? Or only the examples? Because if we change all the plugins to 31, that means we'll be pointlessly forcing plugin clients to update their SDK version, right?

Although I guess the counter-argument is that if our plugin lists, say, version 29, and the example builds with 31, then we could accidentally add a dependency on 30 or 31 in the build without clients getting a warning, right? (In which case we should have a check in the repo tooling that enforces that plugin and example versions match.)

blasten:
Once https://github.com/flutter/engine/pull/29585 lands, every app will need to bump the `compileSdkVersion`. 

I agree the bumping `compileSdkVersion` in the plugin's build.gradle isn't necessarily relevant, but I also don't see issues with bumping it.

packages/quick_actions/quick_actions/CHANGELOG.md
camsim99: Shouldn't this require a version bump because of the changes I made [here](https://github.com/flutter/plugins/pull/4502/files#r750511289)?

Hixie:
test-exempt: already tested per @stuartmorgan's comment above

packages/quick_actions/quick_actions/CHANGELOG.md
stuartmorgan: Yes, sorry; I confused this with a deprecated plugin.

packages/path_provider/path_provider_android/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## NEXT
```
stuartmorgan, line 1: You should go ahead and version bump this, since plugin clients using newer SDK versions will hit the warning as well; we get issues filed about build warnings, so people do notice.

packages/android_alarm_manager/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Remove support for the V1 Android embedding.
 * Updated Android lint settings.
 * Removed `-Werror` in Android builds.
+* Update example app Android compileSdkVersion to 31.
```
stuartmorgan, line 6: These should all be "Updates", per the new style guide ("Update" is imperative, rather than indicative).

Hopefully once we've been following the new style for a while it'll get easier, since all the recent entries will start to become consistent over time.

packages/path_provider/path_provider_android/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+
+* Update example app Android compileSdkVersion to 31.
+* Fixed typing build warning.
```
stuartmorgan, line 4: Fixes

packages/quick_actions/quick_actions/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.6.0+8
+
+* Update example app Android compileSdkVersion to 31.
+* Move method call to background thread to fix CI failure.
```
stuartmorgan, line 4: Moves

packages/quick_actions/quick_actions/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
```diff
@@ -49,7 +52,18 @@ public void onMethodCall(MethodCall call, MethodChannel.Result result) {
...
+            new ThreadPoolExecutor(0, 1, 1, TimeUnit.SECONDS, new LinkedBlockingQueue<Runnable>());
+
+        executor.execute(
+            new Runnable() {
+              @Override
+              public void run() {
+                if (Build.VERSION.SDK_INT > Build.VERSION_CODES.N_MR1) {
```
stuartmorgan, line 62: Shouldn't this check be around all of the new code? Presumably we don't want to start a thread and send a task to it just to no-op on older versions of Android.

stuartmorgan (approved):
LGTM with nits. Sorry there ended up being a lot of review churn in the minor details here.

packages/quick_actions/quick_actions/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
blasten: could this use the new pattern for background platform channels?

gaaclarke: I talked to @camsim99 and we can't use Background Platform Channels unless we do a breaking change to the platform_interface (since it would require splitting up the channels between a platform thread handler channel and a background thread handler channel).  I'm going to file a bug for implementing TaskQueues for methods inside of MethodChannels.  @stuartmorgan this is related to the issue we were talking about for moving channels out of the platform_interface.

packages/quick_actions/quick_actions/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
```diff
@@ -49,7 +52,18 @@ public void onMethodCall(MethodCall call, MethodChannel.Result result) {
...
+                  shortcutManager.setDynamicShortcuts(shortcuts);
```
gaaclarke, line 63: Right now you are calling result after you've dispatched the message to the background thread.  You should call the result after the background task has executed (calling `result` from the background thread should be thread-safe now, no need to dispatch back to the main thread).

packages/quick_actions/quick_actions/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
gaaclarke: I filed an issue for methodchannel method specific taskqueues: https://github.com/flutter/flutter/issues/93828

I think it's better just to remove platform channels from the platform_interface though.

packages/quick_actions/quick_actions/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
stuartmorgan: > calling `result` from the background thread should be thread-safe now

Does "now" include stable? Plugins support at least back to last stable (often more, but currently for this plugin it's 2.5)

gaaclarke: No, it doesn't I corrected myself over chat.  We can't call the result directly until background platform channels is on stable.

packages/quick_actions/quick_actions/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
```diff
@@ -47,10 +53,36 @@ public void onMethodCall(MethodCall call, MethodChannel.Result result) {
...
+                      "Exception thrown when setting dynamic shortcuts",
+                      null);
+                }
+
+                final boolean didSucceed = dynamicShortcutsSet;
+                uiThreadExecutor.execute(
+                    () -> {
+                      if (didSucceed) result.success(null);
+                    });
+              });
```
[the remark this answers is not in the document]
camsim99: No, this was [corrected](https://github.com/flutter/plugins/pull/4502#discussion_r751787557)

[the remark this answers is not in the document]
blasten: ah gotcha. I missed the last part of the conversation

blasten (approved):
LGTM

packages/quick_actions/quick_actions/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
```diff
@@ -47,10 +53,36 @@ public void onMethodCall(MethodCall call, MethodChannel.Result result) {
...
+              () -> {
+                boolean dynamicShortcutsSet = false;
+                try {
+                  shortcutManager.setDynamicShortcuts(shortcuts);
+                  dynamicShortcutsSet = true;
+                } catch (Exception e) {
+                  result.error(
+                      "quick_action_setshortcutitems_failure",
+                      "Exception thrown when setting dynamic shortcuts",
+                      null);
```
stuartmorgan, line 75: This is on the wrong thread; any interaction with `result` needs to be (on current stable) on the main thread.

stuartmorgan (changes requested):

packages/quick_actions/quick_actions/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
```diff
@@ -69,16 +69,18 @@ public void onMethodCall(MethodCall call, MethodChannel.Result result) {
...
-                      "Exception thrown when setting dynamic shortcuts",
-                      null);
+                  // Leave dynamicShortcutsSet as false
                 }
 
                 final boolean didSucceed = dynamicShortcutsSet;
                 uiThreadExecutor.execute(
                     () -> {
                       if (didSucceed) result.success(null);
+                      else
```
stuartmorgan, line 79: Nit: The `if` and `else` here need braces [per style](https://google.github.io/styleguide/javaguide.html#s4.1-braces). (Sorry I missed that on the `if` in the previous review 🙁 )

Since you have to edit this file again anyway, maybe also mention that this re-dispatch can be removed once Flutter 2.8 or later is on the stable channel.

stuartmorgan (approved):
One last nit, otherwise LGTM

PR #4503, opened 2021-11-11
[webview_flutter_android] Implementation of Android WebView widget using pigeon

bparrishMines:
Replaces `MethodChannelPlatformController` with `WebViewAndroidPlatformController` and `WebViewAndroidWidget`.

Android portion of https://github.com/flutter/flutter/issues/93732

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
```diff
@@ -38,36 +50,100 @@ public WebViewFlutterPlugin() {}
...
       return;
     }
 
     flutterCookieManager.dispose();
     flutterCookieManager = null;
   }
+
+  @Override
+  public void onAttachedToActivity(@NonNull ActivityPluginBinding activityPluginBinding) {
+    setUp(
```
stuartmorgan, line 127: Can't `onAttachedToActivity` be called multiple times? What are the implications of moving platform view registration from the engine attachment to the activity attachment?

And similarly, are the pigeon setup APIs safe to call repeatedly?

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
```diff
@@ -38,36 +50,100 @@ public WebViewFlutterPlugin() {}
...
+        pluginBinding.getBinaryMessenger(),
+        pluginBinding.getPlatformViewRegistry(),
+        activityPluginBinding.getActivity(),
+        null);
+  }
+
+  @Override
+  public void onDetachedFromActivityForConfigChanges() {}
```
stuartmorgan, line 135: The activity was passed to `WebViewHostApi`; doesn't that need to be disconnected somehow at this point?

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -0,0 +1,632 @@
...
+      useHybridComposition: useHybridComposition,
+    );
+
+    webView.settings.setDomStorageEnabled(true);
+    webView.settings.setJavaScriptCanOpenWindowsAutomatically(true);
+    webView.settings.setSupportMultipleWindows(true);
+    webView.settings.setLoadWithOverviewMode(true);
+    webView.settings.setUseWideViewPort(true);
+    webView.settings.setDisplayZoomControls(false);
+    webView.settings.setBuiltInZoomControls(true);
```
stuartmorgan, line 106: Do we know what the performance overhead of all of these being individual platform channel calls is?

This doesn't need to block proceeding, but it would be good to evaluate that to figure out if we should adjust the Dart/native boundary in a follow up for performance reasons (e.g., by adding a 'settings delta' struct that we could use to batch changes).

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -0,0 +1,632 @@
...
+
+  @override
+  Future<void> updateSettings(WebSettings settings) {
+    return Future.wait(<Future<void>>[
+      _trySetHasProgressTracking(settings.hasProgressTracking),
+      _trySetHasNavigationDelegate(settings.hasNavigationDelegate),
+      _trySetJavaScriptMode(settings.javascriptMode),
+      _trySetDebuggingEnabled(settings.debuggingEnabled),
+      _trySetUserAgent(settings.userAgent),
+      _trySetZoomEnabled(settings.zoomEnabled),
```
stuartmorgan, line 189: Same here.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -0,0 +1,632 @@
...
+
+    final AutoMediaPlaybackPolicy autoMediaPlaybackPolicy =
+        creationParams.autoMediaPlaybackPolicy;
+    switch (autoMediaPlaybackPolicy) {
+      case AutoMediaPlaybackPolicy.always_allow:
+        webView.settings.setMediaPlaybackRequiresUserGesture(false);
+        break;
+      default:
+        webView.settings.setMediaPlaybackRequiresUserGesture(true);
+    }
```
stuartmorgan, line 283: Why not just:
`webView.settings.setMediaPlaybackRequiresUserGesture(autoMediaPlaybackPolicy != AutoMediaPlaybackPolicy.always_allow)`?

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -0,0 +1,632 @@
...
+
+    addJavascriptChannels(creationParams.javascriptChannelNames);
+  }
+
+  Future<void> _trySetHasProgressTracking(bool? hasProgressTracking) {
```
stuartmorgan, line 288: Can we call all of these something more like `setFooOrIgnoreNull`? I wasn't sure what "try" meant at the call sites.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -0,0 +1,632 @@
...
+    if (hasProgressTracking == true) {
+      webChromeClient._onProgress = callbacksHandler.onProgress;
+    } else if (hasProgressTracking == false) {
+      webChromeClient._onProgress = null;
+    }
+
+    return Future<void>.sync(() => null);
```
stuartmorgan, line 295: All the other functions use an early return for null; please be consistent to make understanding the batch of methods easier.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
stuartmorgan: Alternately, you could make these not take nullables and use conditionals at the call site:

```
return Future.wait(<Future<void>>[
  if (settings.hasProgressTracking != null)
    _setHasProgressTracking(settings.hasProgressTracking),
  if (settings.hasNavigationDelegate)
    _trySetHasNavigationDelegate(settings.hasNavigationDelegate),
  ...
```

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -0,0 +1,632 @@
...
+  Future<void> _trySetUserAgent(WebSetting<String?> userAgent) {
+    if (userAgent.isPresent && userAgent.value != null) {
+      return webView.settings.setUserAgentString(userAgent.value!);
+    }
+
+    return webView.settings.setUserAgentString('');
+  }
+
+  Future<void> _trySetZoomEnabled(bool? zoomEnabled) async {
+    if (zoomEnabled == null) return Future<void>.sync(() => null);
```
stuartmorgan, line 348: This inline condition+statement reminds me: since you're rewriting ~all of the Dart, how hard would it be to flip the analysis options of this package to the new options? It would substantially cut down on later churn.

(To do just this one, you'd move packages/webview_flutter/analysis_option.yaml into all of the subpackages except for this one.)

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -0,0 +1,632 @@
...
+  /// Manages named JavaScript channels and forwarding incoming messages on the correct channel.
+  final JavascriptChannelRegistry javascriptChannelRegistry;
+
+  @override
+  void postMessage(String message) {
+    javascriptChannelRegistry.onJavascriptChannelMessage(channelName, message);
+  }
+}
+
+/// Receives callbacks when content can not be handled by the rendering engine for [WebViewAndroidPlatformController], and should be downloaded instead.
```
stuartmorgan, line 370: Please wrap to 80

stuartmorgan:
Adding @mvanbeusekom for a second review, since this is a substantial rewrite, so would benefit from more eyes.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
bparrishMines: We had a similar problem with doing this with [google_mobile_ads](https://github.com/googleads/googleads-mobile-flutter/pull/381). It mostly only affect loading views with cached engines. The solution was to set everything up in `onAttachedToEngine` with `binding.applicationContext` and then switch the `context` when the plugin was attached to an `Activity`. I went ahead and changed this plugin to do the same. I don't believe there are any other problems this implementation causes.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
bparrishMines: I switch the context for `WebViewHostApi` to `FlutterPluginBinding.applicationContext()` for the explanation I gave above for `onAttachedToActivity`.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
bparrishMines: I prefer to use `switch` statements for enums because the Dart code will throw a lint error that one of the values wasn't listed. We had a bug once when we didn't support all the enum values.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
bparrishMines: Done. I decided to go with this solution.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
stuartmorgan: `default` covers all cases, so you'll never get a linter warning here no matter what is added to the enum.

> We had a bug once when we didn't support all the enum values.

If someone is adding new functionality, they should be adding test coverage, which is how we should catch this. Relying on the lint (by removing `default` to enable it) would make adding new enum values repository-breaking. We've already had a sever issue with this; see https://github.com/flutter/flutter/issues/89866

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
bparrishMines: So I did some testing with the with loading the `initialUrl` and the results I got were: (milliseconds)

```
WebView Widget (this implementation):
3016
2856
2882
2866
2925
3023
2803
2758
2934
2801

Avg: 2887.3

WebView Master (method channel implementation)
3702
3115
4877
3257
4581
3276
2998
2953
2789
2778

Avg excluding outliers: 3023
```

I haven't been able to find any noticeable differences between the implementations. But, I was also curious with the speed check when there were other `MethodCall`s being made while the `WebView` was loading so I added a constant `MethodCall` during the test:

```
Future<void> testCall() async {
    const MethodChannel('testChannel').invokeMethod<void>('');
    if (!stop) {
      await Future<void>.delayed(const Duration(milliseconds: 10));
      testCall();
    }
  }
```

And I received similar results.

The code for the test: 
```dart
testWidgets('initialUrl', (WidgetTester tester) async {
    final DateTime now = DateTime.now();
    final Completer<void> onPageFinished = Completer<void>();
    await tester.pumpWidget(
      MaterialApp(
        home: Directionality(
          child: WebView(
            initialUrl: primaryUrl,
            onPageFinished: (_) {
              final int millis = DateTime.now().millisecondsSinceEpoch -
                  now.millisecondsSinceEpoch;
              print('$millis');
              onPageFinished.complete();
            },
          ),
        ),
      ),
    );
    await onPageFinished.future;
  });
```

stuartmorgan: Awesome, good to know that overhead of crossing the boundary is minimal in general!

stuartmorgan:
Is this ready for re-review?

mvanbeusekom (commented):
I am wondering if it would be possible to replace the `create` method channel call with the default structure that is offered by the `AndroidView` widget. See also my comment for the `FlutterWebViewFactory.java` file.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
bparrishMines: Done. I updated the analysis options for this plugin

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
bparrishMines: I don't think flutter/flutter#89866 actually provides a recommendation on the implementation. Are you suggesting that we keep the `default` value? If the platform interface adds a new `enum` value, wouldn't only an integration test in the main plugin (e.g. `webview_flutter`) catch a change like that? The unit tests in the platform interface and main plugin should be platform agnostic. 

I think I prefer option 3 from flutter/flutter#89866. That way there is a process to locate everywhere enum used in the implementations. And we can add `TODO`s to update at a later point or in a followup PR.

stuartmorgan: > I don't think [flutter/flutter#89866](https://github.com/flutter/flutter/issues/89866) actually provides a recommendation on the implementation. Are you suggesting that we keep the `default` value?

I think in cases where we can have a reasonable `default`, we should do that, because it reduces the chances of cross-package breakage. (We don't yet have a way to detect that in CI.)

> If the platform interface adds a new `enum` value, wouldn't only an integration test in the main plugin (e.g. `webview_flutter`) catch a change like that? The unit tests in the platform interface and main plugin should be platform agnostic.

I'm not sure I follow. If someone is adding a new enum value, it's presumably to add new functionality based on that new value. That would generally require a change to each platform implementation package to actually use the new enum value, and those changes should have tests. So it should be something that would be caught by tests both at the platform implementation level and potentially at the app-facing level.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewFactory.java
```diff
@@ -5,29 +5,24 @@
...
   public PlatformView create(Context context, int id, Object args) {
-    Map<String, Object> params = (Map<String, Object>) args;
-    MethodChannel methodChannel = new MethodChannel(messenger, "plugins.flutter.io/webview_" + id);
-    return new FlutterWebView(context, methodChannel, params, containerView);
+    final PlatformView view = (PlatformView) instanceManager.getInstance((Integer) args);
+    if (view == null) {
+      throw new IllegalStateException("Unable to find WebView instance: " + args);
+    }
+    return view;
   }
```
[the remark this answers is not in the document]
bparrishMines: The main reason I didn't want create to the WebView in the `PlatformViewFactory` was that I wanted to eventually have an update for https://github.com/flutter/flutter/issues/51875. Where a user would be able to load a URL for a `WebView` and make calls to scrolling without needing to add the `Widget` to the widget tree. We probably should actually have the `PlatformView` API provide support to create the View without adding it to the tree, but I think this is the only way around it at the moment.

Also, the Pigeon API is built to mimic the [android.webkit](https://developer.android.com/reference/android/webkit/WebView) API. My goal was to avoid significant deviation from this because it can make it harder to handle changes to the API later or run into conflicts when adding features we weren't aware of before.

In the long term, it should become a standard to always call `create` in the constructor of a class and `dispose` would be handled automatically by garbage collection. This is probably at least 2 quarters away, but the changes to `webview_flutter` are essentially the stepping stones to get there.

bparrishMines:
@stuartmorgan This is ready for another review!

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
bparrishMines: Ah, ok. I assumed someone would be able to make a change to a platform interface `enum` without needing to update the implementations. If that's the case, then yea we should be able to catch all the uses if we require this.

bparrishMines: I changed it to your suggestion

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
```diff
@@ -38,36 +53,113 @@ public WebViewFlutterPlugin() {}
...
     }
 
     flutterCookieManager.dispose();
     flutterCookieManager = null;
   }
+
+  @Override
+  public void onAttachedToActivity(@NonNull ActivityPluginBinding activityPluginBinding) {
+    final Activity activity = activityPluginBinding.getActivity();
+    webViewHostApi.setContext(activityPluginBinding.getActivity());
```
stuartmorgan, line 140: Shouldn't this be `activity`?

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
```diff
@@ -38,36 +53,113 @@ public WebViewFlutterPlugin() {}
...
+    final Activity activity = activityPluginBinding.getActivity();
+    webViewHostApi.setContext(activity);
+    javaScriptChannelHostApi.setPlatformThreadHandler(new Handler(activity.getMainLooper()));
+  }
+
+  @Override
+  public void onDetachedFromActivity() {
+    final Context context = pluginBinding.getApplicationContext();
+    webViewHostApi.setContext(context);
+    javaScriptChannelHostApi.setPlatformThreadHandler(new Handler(context.getMainLooper()));
```
stuartmorgan, line 163: This pattern is repeated four times; it seems like changing this (e.g., adding another thing that needs to be adjusted) would be very error-prone. How about extracting a helper that takes the `Context`, and then each of these can be a one-liner to call that helper with the right context for the scenario.

stuartmorgan (approved):
LGTM with nits

(With the caveat that with a full rewrite like this, it's of course very hard to have confidence that nothing regressed.)

mvanbeusekom (approved):
LGTM

PR #4508, opened 2021-11-14
[battery] Recreate Android example

stuartmorgan:
Re-creates example/android using `flutter create`, preserving only the
integration test harness files and the gradle settings, since the
example was using the v1 embedding, which is now an error on Flutter
master.

Fixes out-of-band tree breakage due to https://github.com/flutter/flutter/pull/93566

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/battery/battery/example/android/app/src/main/AndroidManifest.xml
+++ b/packages/battery/battery/example/android/app/src/main/AndroidManifest.xml
@@ -5,7 +5,7 @@
         android:name="${applicationName}"
         android:icon="@mipmap/ic_launcher">
         <activity
-            android:name=".MainActivity"
+            android:name="io.flutter.embedding.android.FlutterActivity"
             android:exported="true"
             android:launchMode="singleTop"
             android:theme="@style/LaunchTheme"
--- a/packages/battery/battery/example/android/app/src/main/AndroidManifest.xml
+++ b/packages/battery/battery/example/android/app/src/main/AndroidManifest.xml
@@ -1,7 +1,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="dev.flutter.plugins.example">
+    package="io.flutter.plugins.batteryexample">
    <application
-        android:label="example"
+        android:label="batteryexample"
         android:name="${applicationName}"
         android:icon="@mipmap/ic_launcher">
         <activity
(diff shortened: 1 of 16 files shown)


GaryQian (approved):
Lgtm!

Thanks for bringing this back into compatibility.

I'll definitely be more proactive in manually validating changes like this.

stuartmorgan:
@GaryQian @blasten Any idea what I've done wrong to get
```
/Users/stuartmorgan/src/plugins/packages/battery/battery/example/android/app/src/main/AndroidManifest.xml:5:9-42 Error:
	Attribute application@name at AndroidManifest.xml:5:9-42 requires a placeholder substitution but no value for <applicationName> is provided.
```
on stable (but not master)?

I'm assuming something about the way I partially recreated the project but left a few small bits intact is problematic, but I'm not sure how.

GaryQian:
The ${applicationName} variable is only used for newer versions of flutter as the feature was added after stable. 

This can be resolved to work on both stable and master by just deleting the android:name property. It is used to automatically add support for things like multidex which these plugins/examples will never need.

packages/battery/battery/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -2,11 +2,9 @@
     package="io.flutter.plugins.batteryexample">
    <application
         android:label="batteryexample"
-        android:name="${applicationName}"
         android:icon="@mipmap/ic_launcher">
         <activity
             android:name="io.flutter.embedding.android.FlutterActivity"
-            android:exported="true"
```
GaryQian: Nit: You should include the exported="true" as this will need to be update din future versions anyways

stuartmorgan:
> The ${applicationName} variable is only used for newer versions of flutter as the feature was added after stable.

Thanks, I figured it out by recreating the example on stable (obvious in retrospect; sorry for pinging you). I was confused before because I was checking against the v2 migration instructions, which showed making that change, and I wasn't expecting that doc to reference things that aren't available in current stable.

packages/battery/battery/example/android/app/src/main/AndroidManifest.xml
stuartmorgan: Presumably it'll have an auto-migration at the point where it becomes required though. This is removed relative to the previous upload because 2.5 doesn't have it in the register, and this is now a 2.5-created example.

Hixie:
test-exempt: fixes test

stuartmorgan:

PR #4509, opened 2021-11-15
[webview_flutter] Add onUrlChanged callback to platform interface.

BeMacized:
This PR adds a `onUrlChanged` event to the platform interface that is supposed to be fired whenever the url for a webview changes. In contrast to the already existing `onPageStarted`, `onProgress`, and `onPageFinished` events, `onUrlChanged` also fires for url changes without a full page load, like when navigating around in a single page application. 

Relevant issue: 
- flutter/flutter#27729

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_platform_interface/lib/src/platform_interface/webview_platform_callbacks_handler.dart
```diff
@@ -24,9 +24,16 @@ abstract class WebViewPlatformCallbacksHandler {
...
   /// Invoked by [WebViewPlatformController] when a page is loading.
-  /// /// Only works when [WebSettings.hasProgressTracking] is set to `true`.
+  /// Only works when [WebSettings.hasProgressTracking] is set to `true`.
   void onProgress(int progress);
 
+  /// Invoked by [WebViewPlatformController] when the webview's URL has changed.
+  ///
+  /// Unlike [onPageStarted], [onProgress], and [onPageFinished],
+  /// [onUrlChanged] also fires when navigating without a full page load
+  /// e.g. when using a single page application.
```
mvanbeusekom, line 34: nit: saying "using a single page application" doesn't describe when the `onUrlChanged` callback is called and the `onPageStarted`, `onProgress` and `onPageFinished` not. Changing it too "when navigating within a single page application."  adds a bit more context and makes the example easier to comprehend.

```suggestion
  /// e.g. when navigating within a single page application.
```

mvanbeusekom (approved):
LGTM with one small nit.

stuartmorgan:
This was a breaking change :(

Because implementations use `implements` rather than `extends`, it can't be trivially fixed within this package, so I'm going to have to revert while we figure out how to do this differently.

BeMacized:
> This was a breaking change :(

I was not aware this would lead to a breaking change, absolutely my bad. 
Currently I do not see a way to do this differently without marking it as a breaking change.
I'll discuss this with @mvanbeusekom tomorrow to see if we can find a better solution.

stuartmorgan:
> I was not aware this would lead to a breaking change, absolutely my bad.

Yes, this is a structural issue in our CI; we need to implement https://github.com/flutter/flutter/issues/89862

stuartmorgan:
> Currently I do not see a way to do this differently without marking it as a breaking change.
> I'll discuss this with @mvanbeusekom tomorrow to see if we can find a better solution.

This interface is "internal" to the implementations (i.e., it's not something that clients of the app-facing package would implement), right?

If so, the non-breaking change would likely be something like:
- Add a new, second callbacks object. (To future-proof this one, prominently document that anything implementing it should `extend`, not `implement`, as more things may be added in the future. And that any method added to it must have a default implementation.)
- Make a new alternate `build` method in the platform interface that includes that second object as an argument; give it a default implementation that calls the old build method (discarding the new object) so that's not breaking.

Then the implementations can implement the new version of build to wire things up using that new object, and the app-facing package can switch to calling the new build instead of the old one. It's not great from an API perspective, but if that's contained to the implementing packages it's livable.

BeMacized:
@stuartmorgan The problem I'm running into with this solution is that the WebViewCupertino and WebViewAndroid classes from the native packages both `implement` WebViewPlatform. As far as I can see, this solution would still result in a breaking change.

stuartmorgan:
> The problem I'm running into with this solution is that the WebViewCupertino and WebViewAndroid classes from the native packages both `implement` WebViewPlatform.

Well that's not good.

I'll need to take a look in detail at the class structure here and see if there's a point where we could set up a new opt-in path, or if we're stuck doing a breaking change to fix the overuse of `implements`

NTA-trongpq:
How to use, bro?

PR #4520, opened 2021-11-18
[camera] Fixed a crash when streaming on iOS

zuvola:
Fixes [flutter/flutter#44436](https://github.com/flutter/flutter/issues/44436)

Streaming large images on older devices seems to cause the call stack to fill up as the processing cannot keep up.
Added a confirmation of reception for each frame, and skipped frames that could not be processed.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera/example/integration_test/camera_test.dart
```diff
@@ -240,4 +240,34 @@ void main() {
...
+
+      await controller.initialize();
+      int _frame = 0;
+
+      await controller.startImageStream((CameraImage image) {
+        _frame++;
+      });
+
+      await Future.delayed(Duration(seconds: 5));
+      expect(_frame > 30, true);
```
stuartmorgan, line 266: It seems like this is testing round-trip speed on the device it's running on, rather than anything about the code itself. Where is the value of `30` coming from? How do you know that this will always pass?

packages/camera/camera/example/integration_test/camera_test.dart
```diff
@@ -240,4 +240,34 @@ void main() {
     },
     skip: !Platform.isAndroid,
   );
+
+  testWidgets(
+    'Image streaming persistence test on iOS',
```
stuartmorgan, line 245: Would this test fail if the rest of the PR were reverted? It doesn't seem that way, unless I'm missing something about the test.

packages/camera/camera/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * Fixes bug where calling a method after the camera was closed resulted in a Java `IllegalStateException` exception.
 * Fixes integration tests.
+* Fixes crash when streaming in iOS.
```
stuartmorgan, line 5: Changes cannot be added to versions that have already been published.

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -438,6 +438,13 @@ class CameraController extends ValueNotifier<CameraValue> {
     _imageStreamSubscription =
         cameraEventChannel.receiveBroadcastStream().listen(
       (dynamic imageData) {
+        if (defaultTargetPlatform == TargetPlatform.iOS) {
+          try {
+            _channel.invokeMethod<void>('receivedImageStreamData');
```
stuartmorgan, line 443: Given that this is asynchronous in both directions I would expect this could cause significant frame loss even on a reasonably powerful device. How much investigation have you done into specifically what the parameters of the crash are? Is it really necessary to have a maximum of a single frame in flight?

It seems likely that this needs a more nuanced approach (maybe a hard limit still but higher than one, maybe something that's tunable by plugin clients). It's hard to determine the right fix without more details about what the failure condition actually is.

stuartmorgan (commented):
Thanks for the contribution!

packages/camera/camera/example/integration_test/camera_test.dart
zuvola: Surely 30 is not important, so I removed it.
I wanted to test if I could keep it running for a certain amount of time without crashing.
I'm not sure if 5 seconds is a reasonable number in this case.
Actually, I don't know how to create such environment-dependent tests.

packages/camera/camera/example/integration_test/camera_test.dart
zuvola: It is a device dependent test, so if you test it on an old device, it will crash.

packages/camera/camera/CHANGELOG.md
zuvola: Fixed it.

packages/camera/camera/lib/src/camera_controller.dart
zuvola: I modified the plugin to allow you to set a limit on the maximum number of frames pending processing.
And I tested it on my devices to see how many frames can be processed in 5 seconds. The results are shown in the table below.
The value is the number of frames processed at max quality and is the average of three times.
The horizontal axis is the maximum number of pending frames (frameStack), where 0 means no limit.
With no limit, iPhone6 and iPadAir crashed.
In this case, if frameStack is set to about 4, frame loss can be reduced to some extent.

  | 0 | 1 | 2 | 3 | 4 | 5 | 6
-- | -- | -- | -- | -- | -- | -- | --
iPhoneX | 143 | 76 | 108 | 118 | 133 | 138 | 136
iPhone6 | 13(crash) | 40 | 53 | 69 | 70 | 55 | 56
iPadAir-1st | 14(crash) | 44 | 62 | 53 | 63 | 52 | 45

I don't have that many devices, but according to [flutter/flutter#44436](https://github.com/flutter/flutter/issues/44436), it seems to be crashing on iPhone 6, 7, 8, etc. These devices still have a large market share that cannot be ignored. Also, for image analysis processes such as OCR, low or medium quality is not always satisfactory.

packages/camera/camera/example/integration_test/camera_test.dart
stuartmorgan: This means that it wouldn't actually fail in our CI, which means there's nothing preventing this from breaking again.

A test for something like this should be creating the conditions to allow for testing that the behavior is what we want regardless of the device, so that it has predictable output. If the goal is to prevent more than N frames from being in flight without response, for instance, then a test should deliberately respond very slowly and assert that N isn't exceeded.

packages/camera/camera/lib/src/camera_controller.dart
stuartmorgan: Thanks for this testing. Given this table, I'm fine with starting from setting a fixed value—4 looks like a good place to start—with an explanation in the code of where that number came from, and then if it turns out that causes issues in some cases we can add an API to tune it later.

stuartmorgan (changes requested):
(Updating review status for the comments above)

zuvola:
Fixed the points you pointed out.
Please review again.

renanyoy:
would be nice to just have an option skip pending frame.. for OCR usages, etc..

packages/camera/camera/example/ios/RunnerTests/StreamingTest.m
```diff
@@ -0,0 +1,95 @@
...
+
+@interface FLTImageStreamHandler : NSObject <FlutterStreamHandler>
+@property FlutterEventSink eventSink;
+@end
+
+@interface FLTCam (Private)
+@property(assign, nonatomic) int maxStreamingPendingFrames;
+@property(assign, nonatomic) BOOL isStreamingImages;
+@property(nonatomic) FLTImageStreamHandler *imageStreamHandler;
+@property(retain, nonatomic) FLTCam *camera;
```
stuartmorgan, line 20: You're declaring a non-existent method here.

packages/camera/camera/example/ios/RunnerTests/StreamingTest.m
```diff
@@ -0,0 +1,95 @@
...
+  // Initial value
+  XCTAssertEqual(cam.streamingPendingFrames, 0);
+
+  // Emulate receiving a video frame
+  [camera.camera captureOutput:nil didOutputSampleBuffer:sampleBuffer fromConnection:nil];
+  XCTAssertEqual(cam.streamingPendingFrames, 1);
+
+  // ReceivedCall reduces streamingPendingFrames
+  [camera handleMethodCallAsync:receivedCall result:nil];
+  XCTAssertEqual(cam.streamingPendingFrames, 0);
```
stuartmorgan, line 86: Why do you need to do all of these assertions of internal counts? Isn't what matters here the externally observable behavior, which is that you don't get more than N callbacks to your handler if you don't acknowledge receipt? (And that once you do, you start getting more callbacks again.)

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -158,6 +158,8 @@ - (void)handleMethodCallAsync:(FlutterMethodCall *)call
   } else if ([@"stopImageStream" isEqualToString:call.method]) {
     [_camera stopImageStream];
     [result sendSuccess];
+  } else if ([@"receivedImageStreamData" isEqualToString:call.method]) {
+    _camera.streamingPendingFrames--;
```
stuartmorgan, line 162: This should be an internal detail; the public API for FLTCam should be a method that acknowledges receipt, not a counter. Incrementing and decrementing a variable shouldn't be split between two classes.

stuartmorgan (changes requested):

stuartmorgan:
@jmagman @hellohuanlin One of you should also take a look at this PR.

jmagman:
> @jmagman @hellohuanlin One of you should also take a look at this PR.

Will do.  I'll review after the current requested changes are fixed.

hellohuanlin (commented):
Thank you for your effort on this. I think it makes sense. Left a question on the camera API change. 

Besides dropping the frames, another improvement I was thinking (proposed [here](https://docs.google.com/document/d/1Znjp-BduXIb1Ix3JOE1YKyJi1Z0d6rB-2q6Lt7j8xSo/edit#), under "_System Pressure_" section) is to **reduce the frame rate upon system pressure**, which should help with both streaming and disk writing. The combination of dropping frames and reducing frame rate should address this kind of issues

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -158,6 +158,8 @@ - (void)handleMethodCallAsync:(FlutterMethodCall *)call
   } else if ([@"stopImageStream" isEqualToString:call.method]) {
     [_camera stopImageStream];
     [result sendSuccess];
+  } else if ([@"receivedImageStreamData" isEqualToString:call.method]) {
```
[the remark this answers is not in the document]
zuvola: Only iOS is called when calling from camera_controller. So other platforms should not be affected. Does this answer your question?

zuvola:
I think I have corrected all of your comments.
Please review again.

packages/camera/camera/ios/Classes/FLTCam.m
```diff
@@ -68,7 +60,8 @@ @interface FLTCam () <AVCaptureVideoDataOutputSampleBufferDelegate,
 @property(assign, nonatomic) BOOL videoIsDisconnected;
 @property(assign, nonatomic) BOOL audioIsDisconnected;
 @property(assign, nonatomic) BOOL isAudioSetup;
-@property(assign, nonatomic) BOOL isStreamingImages;
+@property(assign, nonatomic) int streamingPendingFramesCount;
+@property(assign, nonatomic) int maxstreamingPendingFramesCount;
```
hellohuanlin, line 64: capital `S`

hellohuanlin (changes requested):

packages/camera/camera/example/ios/RunnerTests/StreamingTest.m
```diff
@@ -0,0 +1,85 @@
...
+
+  id handlerMock = OCMClassMock([FLTImageStreamHandler class]);
+  OCMStub([handlerMock eventSink]).andReturn(^(id event) {
+    [streamingExpectation fulfill];
+  });
+
+  id messenger = OCMProtocolMock(@protocol(FlutterBinaryMessenger));
+  [_camera startImageStreamWithMessenger:messenger imageStreamHandler:handlerMock];
+
+  [self waitForStart];
```
hellohuanlin, line 49: i highly recommend not to create this helper function since it's making the test a bit harder to read. Unit tests themselves are not unit tested, so it's better to make it as simple as possible.

packages/camera/camera/example/ios/RunnerTests/StreamingTest.m
```diff
@@ -0,0 +1,85 @@
...
+
+- (void)tearDown {
+  CFRelease(_sampleBuffer);
+}
+
+- (void)waitForStart {
+  NSPredicate *predicate = [NSPredicate predicateWithFormat:@"isStreamingImages == YES"];
+  XCTNSPredicateExpectation *expectation =
+      [[XCTNSPredicateExpectation alloc] initWithPredicate:predicate object:_camera];
+  XCTWaiterResult result = [XCTWaiter waitForExpectations:@[ expectation ] timeout:3];
```
hellohuanlin, line 33: i'm curious will it fail if we use 1 second? due to predicate parsing? if so please document it out. otherwise let's keep 1 second.

hellohuanlin: also i think you can do `[self expectationForPredicate...]` and `[self waitForExpectations:]`

hellohuanlin (commented):
It's looking good

packages/camera/camera/example/ios/RunnerTests/StreamingTest.m
zuvola: I see. Fixed.

packages/camera/camera/example/ios/RunnerTests/StreamingTest.m
zuvola: Using XCTNSPredicateExpectation, it seems to time out with a 1 second wait.
I tried using KVO this time and it seems to work well.

hellohuanlin (approved):

packages/camera/camera/ios/Classes/FLTCam.h
```diff
@@ -60,6 +60,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)setExposureModeWithResult:(FLTThreadSafeFlutterResult *)result mode:(NSString *)modeStr;
 - (void)setFocusModeWithResult:(FLTThreadSafeFlutterResult *)result mode:(NSString *)modeStr;
 - (void)applyFocusMode;
+- (void)receivedImageStreamData;
```
stuartmorgan, line 63: This needs a declaration comment, per style guide.

(I know almost nothing else in this header current has one, but existing style violations aren't a reason to perpetuate new style violations.)

packages/camera/camera/ios/Classes/FLTCam.m
```diff
@@ -68,7 +60,8 @@ @interface FLTCam () <AVCaptureVideoDataOutputSampleBufferDelegate,
 @property(assign, nonatomic) BOOL videoIsDisconnected;
 @property(assign, nonatomic) BOOL audioIsDisconnected;
 @property(assign, nonatomic) BOOL isAudioSetup;
-@property(assign, nonatomic) BOOL isStreamingImages;
+@property(assign, nonatomic) int streamingPendingFramesCount;
+@property(assign, nonatomic) int maxStreamingPendingFramesCount;
```
stuartmorgan, line 64: Same for these new properties.

packages/camera/camera/ios/Classes/FLTCam_Test.h
```diff
@@ -5,6 +5,14 @@
 #import "FLTCam.h"
 #import "FLTSavePhotoDelegate.h"
 
+@interface FLTImageStreamHandler : NSObject <FlutterStreamHandler>
+// The queue on which `eventSink` property should be accessed
```
stuartmorgan, line 9: Nit: These are declaration comments, so should follow declaration comment style (which is `///` for this file).

packages/camera/camera/ios/Classes/FLTCam_Test.h
```diff
@@ -5,6 +5,14 @@
...
+@property(nonatomic, strong) dispatch_queue_t captureSessionQueue;
+// `eventSink` property should be accessed on `captureSessionQueue`.
```
stuartmorgan, line 11: Blank line between declarations please.

stuartmorgan (changes requested):
Looks good! Just a few minor style issues to resolve.

packages/camera/camera/ios/Classes/FLTCam.h
zuvola: Done.
But, my English isn't very good, so please point out any oddities.

packages/camera/camera/ios/Classes/FLTCam.h
```diff
@@ -60,6 +60,10 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)setExposureModeWithResult:(FLTThreadSafeFlutterResult *)result mode:(NSString *)modeStr;
 - (void)setFocusModeWithResult:(FLTThreadSafeFlutterResult *)result mode:(NSString *)modeStr;
 - (void)applyFocusMode;
+
+/// Sets that a streaming image has been received.
```
stuartmorgan, line 64: "Sets" implies that there's a boolean toggle here, which isn't the case. And the last sentence of this comment isn't relevant to clients using the method.

I wold replace this whole declaration comment with something like:
```
/**
 * Acknowledges the receipt of one image stream frame.
 *
 * This should be called each time a frame is received. Failing to call it may
 * cause later frames to be dropped instead of streamed.
 */
```

(This file uses /** ... */ style. Ideally we'd be consistent within the whole project, but we should at least have file-level consistency.)

packages/camera/camera/ios/Classes/FLTCam_Test.h
```diff
@@ -6,11 +6,14 @@
 #import "FLTSavePhotoDelegate.h"
 
 @interface FLTImageStreamHandler : NSObject <FlutterStreamHandler>
-// The queue on which `eventSink` property should be accessed
+
+/// The queue on which `eventSink` property should be accessed
```
stuartmorgan, line 10: Nit: missing period.

packages/camera/camera/ios/Classes/FLTCam.m
```diff
@@ -135,6 +133,11 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
   _videoFormat = kCVPixelFormatType_32BGRA;
   _inProgressSavePhotoDelegates = [NSMutableDictionary dictionary];
 
+  // To prevent memory consumption, limit the number of frames pending processing.
```
stuartmorgan, line 136: Nit: change "prevent" to "limit".

packages/camera/camera/ios/Classes/FLTCam_Test.h
```diff
@@ -5,6 +5,17 @@
 #import "FLTCam.h"
 #import "FLTSavePhotoDelegate.h"
 
+@interface FLTImageStreamHandler : NSObject <FlutterStreamHandler>
+
+/// The queue on which `eventSink` property should be accessed
+@property(nonatomic, strong) dispatch_queue_t captureSessionQueue;
+
+/// `eventSink` property should be accessed on `captureSessionQueue`.
+/// The block itself should be invoked on the main queue.
```
stuartmorgan, line 14: A declaration comment should start wit a description of what something is, before getting into how to use it. So:

```
/// The event sink to stream camera events to Dart.
///
/// The property should only be accessed on `captureSessionQueue`.
/// The block itself should be invoked on the main queue.
```

stuartmorgan (changes requested):
Just a few more minor nits on the comments. Thanks for adding them!

packages/camera/camera/ios/Classes/FLTCam.h
zuvola: I learned a lot. Thank you.
I ended up using your suggestion in its entirety.

stuartmorgan (approved):
LGTM, thanks!

renanyoy:
is it fixed in flutter 3.0 ?

jmagman:
This was fixed in camera plugin version 0.9.4+18, not a version of Flutter.

AnimeshHestabit:
**I'm getting crash on video recording on camera v0.9.7+1 and the latest one (camera v0.9.8+1)** 

*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: '*** -[AVAssetWriterInputPixelBufferAdaptor appendPixelBuffer:withPresentationTime:] A pixel buffer cannot be appended when readyForMoreMediaData is NO.'
*** First throw call stack:
(0x185349cac 0x19c3b8758 0x18def7a48 0x18def516c 0x18defa8ec 0x10383851c 0x1a7e897ac 0x1a7e74a20 0x1a80a79c8 0x1a823fc3c 0x1850162f0 0x184fb89c8 0x184fca304 0x184fbc41c 0x184fbcff0 0x184fc6ae4 0x1deccef38 0x1decceaa4)
libc++abi: terminating with uncaught exception of type NSException
* thread #25, queue = 'io.flutter.camera.captureSessionQueue', stop reason = signal SIGABRT
    frame #0: 0x00000001be89ea48 libsystem_kernel.dylib`__pthread_kill + 8
libsystem_kernel.dylib`__pthread_kill:
->  0x1be89ea48 <+8>:  b.lo   0x1be89ea64               ; <+36>
    0x1be89ea4c <+12>: stp    x29, x30, [sp, #-0x10]!
    0x1be89ea50 <+16>: mov    x29, sp
    0x1be89ea54 <+20>: bl     0x1be89a6d0               ; cerror_nocancel
Target 0: (Runner) stopped.
Lost connection to device.

PR #4530, opened 2021-11-19
[flutter_plugin_tools] Check for missing version and CHANGELOG updates

stuartmorgan:
The currently documented repository policy is to:
- require version updates for packages changes that don't meet specific exemptions, and
- require CHANGELOG changes for essentially all changes.

This adds tooling that enforces that policy, with a mechanism for overriding it via PR descriptions, to avoid cases where they are accidentally omitted without reviewers catching it.

In order to facilitate testing (which require mocking another `git` command), this also updates the existing `version-check` tests:
- Replaces the custom git result injection/validating with the newer bind-to-process-mocks approach that is now used in the rest of the tool tests.
- Fixes some tests that were only checking for `ToolExit` to also check the error output, in order to ensure that failure tests are not accidentally passing for the wrong reason (as is being done in general as tests in the tooling are updated).

Fixes https://github.com/flutter/flutter/issues/93790

## Pre-launch Checklist

- [x] I signed the [CLA].


script/tool/lib/src/version_check_command.dart
```diff
@@ -141,8 +150,26 @@ class VersionCheckCommand extends PackageLoopingCommand {
...
+  /// to allow skipping a CHANGELOG change for a PR that would normally require
+  /// one.
+  static const String _missingChangelogChangeJustificationMarker =
+      'No CHANGELOG change:';
+
   final PubVersionFinder _pubVersionFinder;
 
+  late GitVersionFinder _gitVersionFinder;
+  late String _mergeBase;
+  late List<String> _changedFiles;
```
gaaclarke, line 169: These can be late final since they are only set once, no?

script/tool/lib/src/version_check_command.dart
```diff
@@ -156,7 +183,11 @@ class VersionCheckCommand extends PackageLoopingCommand {
   bool get hasLongOutput => false;
 
   @override
-  Future<void> initializeRun() async {}
+  Future<void> initializeRun() async {
+    _gitVersionFinder = await retrieveVersionFinder();
+    _mergeBase = await _gitVersionFinder.getBaseSha();
+    _changedFiles = await _gitVersionFinder.getChangedFiles();
```
gaaclarke, line 189: nit: These could be parallelized, right?  No need to serialize them. 

```
mergeBase = _gitVersionFinder.getBaseSha().then((x) { _mergeBase = x; });
...
await Futures.wait([mergeBase, changedFiles])
```

gaaclarke (approved):

script/tool/lib/src/version_check_command.dart
stuartmorgan: Good call, done.

script/tool/lib/src/version_check_command.dart
stuartmorgan: In theory yes, but `getChangedFiles` needs the base SHA, so it would make the code more complex without actually changing anything.

stuartmorgan:
submit-queue incorrectly showing red; landing.

PR #4537, opened 2021-11-23
[webview_flutter] Adjust getTitle test

stuartmorgan:
The getTitle test has been flaking on iOS, for reasons that are not
clear. This is a highly speculative attempted fix, in case the issue is
that the futere is completing too early (e.g., if 'about:blank' is
triggering a completion).

Fixes https://github.com/flutter/flutter/issues/94117

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
+++ b/packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
@@ -748,7 +748,6 @@ void main() {
     final Completer<WebViewController> controllerCompleter =
         Completer<WebViewController>();
 
-
     final testUrl = 'data:text/html;charset=utf-8;base64,$getTitleTestBase64';
     await tester.pumpWidget(
       Directionality(
@@ -759,6 +758,8 @@ void main() {
             controllerCompleter.complete(controller);
           },
           onPageStarted: (String url) {
+            // Ensure that this update is for the test page, in case any other
+            // URL (e.g., about:blank) fires earlier notifications.
             if (url == testUrl) {
               pageStarted.complete(null);
             }
--- a/packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
+++ b/packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
@@ -748,19 +748,25 @@ void main() {
     final Completer<WebViewController> controllerCompleter =
         Completer<WebViewController>();
 
+
+    final testUrl = 'data:text/html;charset=utf-8;base64,$getTitleTestBase64';
     await tester.pumpWidget(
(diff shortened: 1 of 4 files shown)


stuartmorgan:
I don't this this will actually work, but I thought it was worth trying first.

My other theory is that the premise of the test—that the title is available as soon as the page loads—is actually wrong. `title` is KVO-compliant, but we don't have a way to use that from the Dart side (maybe long term we should consider having an event stream for various page attributes), so if we still see flake I was thinking of adding a step where we run JS before checking the title, to ensure that the page has been fully parsed and handled.

packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
```diff
@@ -748,19 +748,25 @@ void main() {
...
         textDirection: TextDirection.ltr,
         child: WebView(
-          initialUrl: 'data:text/html;charset=utf-8;base64,$getTitleTestBase64',
+          initialUrl: testUrl,
           onWebViewCreated: (WebViewController controller) {
             controllerCompleter.complete(controller);
           },
           onPageStarted: (String url) {
-            pageStarted.complete(null);
+            if (url == testUrl) {
```
jmagman, line 762: Can you add a comment for why this is needed? And maybe log when `url != testUrl` to confirm your speculative fix?

packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
```diff
@@ -748,19 +748,25 @@ void main() {
     final Completer<WebViewController> controllerCompleter =
         Completer<WebViewController>();
 
+
```
jmagman, line 751:
```suggestion
```

stuartmorgan: Done.

packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart
stuartmorgan: Added a comment. I had thought about logging, but didn't want to since we generally don't want extra logging in test runs, and if this theory is right this would be permanent code.

I don't think logging would help us much with confirming since it was flaky; we'd never know if a pass without logging meant the theory was wrong, or that it's a run that wouldn't have failed in the first place. We'll know pretty soon whether it worked or not regardless, by whether the flaking stops.

jmagman (approved):
LGTM

stuartmorgan:
Landing on red since the red on master is due to this test flaking.

ditman:
(Reverted by https://github.com/flutter/plugins/pull/4539)

PR #4544, opened 2021-11-25
[webview_flutter] Android implementation of `loadFile` and `loadHtmlString` methods

mvanbeusekom:
Adds implementations of the `loadFile` and `loadHtmlString` method channel calls for the webview_flutter_android package. 

The `loadFile` and `loadHtmlString` methods have been added to the platform interface by PR #4446 and this PR will add the Android implementation. 

Related to issues:
- flutter/flutter#27086
- flutter/flutter#25162
- flutter/flutter#34920
- flutter/flutter#63392

NOTE: as per offline discussion with @stuartmorgan the loading of Flutter assets (assets registered within the pubspec.yaml) should be handled via the `loadFile` method and retrieve the path to the physical asset location the path_provider should be used (or something similar). At the moment the path_provider doesn't support returning the path to the physical location of Flutter assets there is however an issue (flutter/flutter#34026) to track this functionality. 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -66,6 +66,91 @@ class WebView {
...
+  ///   <html><body>'%28' is the code for '('</body></html>
+  /// ''';
+  /// final encodedHtml = base64.encode(utf8.encode(unencodedHtml));
+  /// print(encodedHtml);
+  /// ```
+  ///
+  /// The [mimeType] parameter specifies the format of the data. If WebView
+  /// can't handle the specified MIME type, it will download the data. If
+  /// `null`, defaults to 'text/html'.
+  Future<void> loadData(String data, String? mimeType, String? encoding) {
```
mvanbeusekom, line 93: I choose to make the `mimeType` and `encoding` parameters nullable positional parameters to match the original JAVA API as close as possible. Alternatively I could also make these optional named parameters which might make the Dart API a bit more user friendly. @bparrishMines, @stuartmorgan what would you prefer?

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
```diff
@@ -477,4 +501,12 @@ public void setWebChromeClient(Long instanceId, Long clientInstanceId) {
     final WebView webView = (WebView) instanceManager.getInstance(instanceId);
     webView.setWebChromeClient((WebChromeClient) instanceManager.getInstance(clientInstanceId));
   }
+
+  private static String parseNullStringIdentifier(String value) {
```
bparrishMines, line 505: Can you add a `@Nullable` annotation to this?

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
bparrishMines: I think optional named parameters should work fine here. I agree it makes the code more *Darty* without changing the method.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -66,6 +66,91 @@ class WebView {
...
+    return api.loadDataFromInstance(
+      this,
+      data,
+      mimeType ?? _nullStringIdentifier,
+      encoding ?? _nullStringIdentifier,
+    );
+  }
+
+  /// Loads the given data into this WebView, using baseUrl as the base URL for
```
bparrishMines, line 102: This should all be on one line. I personally don't mind if the first line exceeds the line limit, but `...using baseUrl as the base URL for the content.` could probably be separated to avoid line limit if that was your intention.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
```diff
@@ -66,6 +66,91 @@ class WebView {
...
+  ///
+  /// If a valid HTTP or HTTPS base URL is not specified in [baseUrl], then
+  /// content loaded using this method will have a `window.origin` value of
+  /// `"null"`. This must not be considered to be a trusted origin by the
+  /// application or by any JavaScript code running inside the WebView (for
+  /// example, event sources in DOM event handlers or web messages), because
+  /// malicious content can also create frames with a null origin. If you need
+  /// to identify the main frame's origin in a trustworthy way, you should use a
+  /// valid HTTP or HTTPS base URL to set the origin.
+  Future<void> loadDataWithBaseUrl(
```
bparrishMines, line 137: This method could also use optional named parameters with `data` being `@required`.

bparrishMines (commented):
This looks great! I left just a few comments

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewHostApiImpl.java
mvanbeusekom: Nice catch, thank update.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
mvanbeusekom: Fixed.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview.dart
mvanbeusekom: Thanks for thinking along, updated as you suggested.

mvanbeusekom:
Hi @bparrishMines, thank you for the review. I have processed your feedback and would appreciate it if you could have another look.

bparrishMines (approved):
LGTM

PR #4555, opened 2021-11-29
[webview_flutter] Add supporting interfaces for setting cookies to platform interface.

BeMacized:
This PR adds the following:

- A CookieManager interface for cookie related operations. `clearCookies` was moved here, and `setCookie` was added.
  - Currently, adding new functionality to `WebViewPlatform` causes breaking changes due to flutter/flutter#94051. Having a separate cookie manager should prevent this.
  - Having cookie functionality separated should allow the implementating class to be more accessible. It allows the cookie manager, for example, to be accessed from the WebViewController in the android package, to set initial cookies from the creation params. Previously, the location of `clearCookies` made this rather difficult.
- `CreationParams` has been expanded to also contain initial cookies to be set upon creation of the webview. 
- `clearCookies` in `WebViewPlatform` has been deprecated in favour of `WebViewCookieManager.clearCookies`.

Relevant issue:
- flutter/flutter#27597

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_platform_interface/lib/src/platform_interface/webview_platform.dart
```diff
@@ -59,6 +59,7 @@ abstract class WebViewPlatform {
   /// Clears all cookies for all [WebView] instances.
   ///
   /// Returns true if cookies were present before clearing, else false.
+  @Deprecated('Use `WebViewCookieManagerPlatform.clearCookies` instead.')
```
BeMacized, line 62: @Reviewer Just want to double check if it's ok to deprecate platform interface methods like this?

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/webview_cookie.dart
```diff
@@ -0,0 +1,49 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Configuration to use when creating a new [WebViewPlatformController].
+///
+/// The `autoMediaPlaybackPolicy` parameter must not be null.
```
mvanbeusekom, line 7: This comment is incorrect (probably forgotten to update after copy/paste action).

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/webview_cookie.dart
```diff
@@ -0,0 +1,49 @@
...
+class WebViewCookie {
+  /// Construct a new [WebViewCookie].
```
mvanbeusekom, line 9: nit: Construct -> Constructs (present tense is now recommended for consistency)
```suggestion
  /// Constructs a new [WebViewCookie].
```

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/webview_cookie.dart
```diff
@@ -0,0 +1,49 @@
...
+      required this.domain,
+      this.path = '/'});
+
+  /// The name of the cookie.
+  ///
+  /// Its value should match "cookie-name" in RFC6265bis:
+  /// https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-02#section-4.1.1
+  final String name;
+
+  /// The value of the cookie.
```
mvanbeusekom, line 22: The first paragraph of the dart docs for the `value`, `domain` and `path` fields are exactly the same (namely `The value of the cookie.`) which is a bit confusing. It would be better to describe this a bit more distinctive. A minimal example would be: 

```
/// The cookie-value of the cookie.

/// The domain-value of the cookie.

/// The path-value of the cookie.
```

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/webview_cookie.dart
```diff
@@ -0,0 +1,49 @@
...
+  /// https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-02#section-4.1.1
+  final String domain;
+
+  /// The value of the cookie.
+  ///
+  /// Its value should match "path-value" in RFC6265bis:
+  /// https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-02#section-4.1.1
+  final String path;
+
+  /// Serialize a [WebViewCookie] to a Map<String, String>.
```
mvanbeusekom, line 40: nit: Serialize -> Serializes (present tense is now recommended for consistency)
```suggestion
  /// Serializes the [WebViewCookie] to a Map<String, String>.
```

Note, I also changed `a` into `the` as the method will serialise that particular instance of the `WebCookie` class and not just a random one.

mvanbeusekom (changes requested):
This is looking good, I just added some comments regarding some of the DartDocs.

mvanbeusekom (approved):
LGTM

packages/webview_flutter/webview_flutter_platform_interface/lib/src/platform_interface/webview_platform.dart
stuartmorgan: You can't deprecate methods we're still using; it breaks the tree. (I should have tooling soon that will catch this kind of thing in CI.)

They can be commented as deprecated, and then actually deprecated later once the uses have been removed.

Unless you can fix the call sites right away, we need to revert this part, as this will close not only our tree, but the flutter/flutter tree as well.

BeMacized: @stuartmorgan Alright, I'll quickly submit a PR that changes this annotation to a comment.

PR #4547, opened 2021-11-26
[path_provider] Switch macOS to an internal method channel

stuartmorgan:
Eliminates the path_provider_macos reliance on the default, shared method channel implementation, in favor of an in-package implementation.

Now that it's trivial to do so, also moves the creation of directories when necessary to the Dart side, and unit tests it there.

Part of https://github.com/flutter/flutter/issues/94224

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[the auto-formatter]: https://github.com/flutter/plugins/blob/master/script/tool/README.md#format-code
[...]

PR #4559, opened 2021-11-30
[path_provider] Android implementation of `getDownloadsDirectory`

guillermin:
This PR adds an Android implementation for the getDownloadsDirectory method in the path_provider interface, returning the public Android downloads directory path.

Resolves issue #93198: [path_provider] getDownloadsDirectory() doesn't work for iOS and Android.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/path_provider/path_provider_android/android/src/main/java/io/flutter/plugins/pathprovider/PathProviderPlugin.java
+++ b/packages/path_provider/path_provider_android/android/src/main/java/io/flutter/plugins/pathprovider/PathProviderPlugin.java
@@ -9,6 +9,7 @@ import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
 import android.os.Handler;
 import android.os.Looper;
+import android.os.Environment;
 import android.util.Log;
 import androidx.annotation.NonNull;
 import com.google.common.util.concurrent.FutureCallback;
@@ -55,6 +56,8 @@ public class PathProviderPlugin implements FlutterPlugin, MethodCallHandler {
 
     void getStorageDirectory(@NonNull Result result);
 
+    void getDownloadsDirectory(@NonNull Result result);
+
     void getExternalCacheDirectories(@NonNull Result result);
 
     void getExternalStorageDirectories(@NonNull String directoryName, @NonNull Result result);
@@ -82,6 +85,10 @@ public class PathProviderPlugin implements FlutterPlugin, MethodCallHandler {
 
     public void getStorageDirectory(@NonNull Result result) {
       executeInBackground(() -> getPathProviderStorageDirectory(), result);
+    }
+
+    public void getDownloadsDirectory(@NonNull Result result) {
+      executeInBackground(() -> getPathProviderDownloadsDirectory(), result);
     }
 
     public void getExternalCacheDirectories(@NonNull Result result) {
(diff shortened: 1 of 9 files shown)


packages/path_provider/path_provider_android/android/src/main/java/io/flutter/plugins/pathprovider/PathProviderPlugin.java
```diff
@@ -242,6 +256,10 @@ private String getPathProviderStorageDirectory() {
     return dir.getAbsolutePath();
   }
 
+  private String getPathProviderDownloadsDirectory() {
+    return Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS).getAbsolutePath();
```
stuartmorgan, line 260: This is deprecated; we deliberately do not use it for that reason.

packages/path_provider/path_provider_platform_interface/lib/src/method_channel_path_provider.dart
```diff
@@ -84,8 +84,8 @@ class MethodChannelPathProvider extends PathProviderPlatform {
 
   @override
   Future<String?> getDownloadsPath() {
-    if (!_platform.isMacOS) {
-      throw UnsupportedError('Functionality only available on macOS');
+    if (_platform.isIOS) {
```
stuartmorgan, line 87: This is actually a breaking change. These packages aren't atomic, and the layering here is problematic (these checks should really never have been in the platform interface), so it would be possible to have a version of the Android package that doesn't support it, but the new version of the platform interface that removed the check, and as a result get a different error type.

stuartmorgan (changes requested):
Making the change this way adds complexity, and a breaking change we don't want. The better option is to:
1. Wait until 2.8 or later is on `stable`
2. Switch Android to Dart+internal-method-channel implementation (see https://github.com/flutter/plugins/pull/4547)
3. Make getDownloadsDirectory redirect at the Dart level to `getExternalStorageDirectories` (which despite the legacy name no longer calls that Android method) with the downloads directory constant.

stuartmorgan:
2.8 has been released to stable, so this could proceed.

Also, you'll need to fix your commits to use the same email address as the one you used to sign the CLA (which appears to be the one associated with your GitHub account).

guillermin:
> 2.8 has been released to stable, so this could proceed as described above.
> 
> Also, you'll need to fix your commits to use the same email address as the one you used to sign the CLA (which appears to be the one associated with your GitHub account).

I created a new PR (from a new GitHub account) focusing only on switching Android to an internal method channel: https://github.com/flutter/plugins/pull/4617

I do not consider it by any means "done", I just copied the same steps you did on your MacOS PR (except for the directory creation, which I imagine can't be done on Android). It is supposed to be a starting point for people who know more about this to help finish it off: for example, @gaaclarke could review the TODO that he left in for when https://github.com/flutter/engine/pull/29147 was in stable (which I believe it is).

I imagine a first step I could take in the right direction is review which methods are supported on Android and make the pertinent changes in PathProviderAndroid and the tests.

Just wanted you to know I'm on it and if anybody is interested in participating they can do so through this PR instead of creating a new one.

stuartmorgan:
> I just copied the same steps you did on your MacOS PR (except for the directory creation part). What else do you think it would need?

Probably nothing; I'll take a look.

> Perhaps @gaaclarke could review the TODO that he left in for when [flutter/engine#29147](https://github.com/flutter/engine/pull/29147) was in stable (which I believe it is).

That's unrelated to the Dart side of the method channel, so would be best handled in a different PR.

stuartmorgan:
Are you planning on updating this to do step 3 of https://github.com/flutter/plugins/pull/4559#pullrequestreview-819322914?

PR #4562, opened 2021-12-01
[webview_flutter] Adds the `loadFlutterAsset` method to the interface.

mvanbeusekom:
Adds the `loadFlutterAsset` method to the platform interface. This method, when implemented, is responsible for correctly loading HTML content that  is part of the Flutter asset bundle.

flutter/flutter#27086

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart
+++ b/packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart
@@ -84,13 +84,31 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
   @override
   Future<void> loadFile(String absoluteFilePath) async {
     assert(absoluteFilePath != null);
-    return _channel.invokeMethod<void>('loadFile', absoluteFilePath);
+
+    try {
+      return await _channel.invokeMethod<void>('loadFile', absoluteFilePath);
+    } on PlatformException catch (ex) {
+      if (ex.code == 'loadFile_failed') {
+        throw ArgumentError(ex.message);
+      }
+
+      rethrow;
+    }
   }
 
   @override
   Future<void> loadFlutterAsset(String key) async {
     assert(key.isNotEmpty);
-    return _channel.invokeMethod<void>('loadFlutterAsset', key);
+
+    try {
+      return await _channel.invokeMethod<void>('loadFlutterAsset', key);
+    } on PlatformException catch (ex) {
+      if (ex.code == 'loadFlutterAsset_failed') {
+        throw ArgumentError(ex.message);
+      }
+
+      rethrow;
+    }
   }
 
   @override
--- a/packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart
+++ b/packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart
@@ -103,7 +103,7 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
     try {
       return await _channel.invokeMethod<void>('loadFlutterAsset', key);
     } on PlatformException catch (ex) {
-      if (ex.code == 'loadFlutterAsset_failed') {
+      if (ex.code == 'loadFlutterAsset_invalidKey') {
         throw ArgumentError(ex.message);
       }
 
--- a/packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart
+++ b/packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart
@@ -88,6 +88,12 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
   }
 
   @override
+  Future<void> loadFlutterAsset(String key) async {
+    assert(key.isNotEmpty);
+    return _channel.invokeMethod<void>('loadFlutterAsset', key);
+  }
+
+  @override
   Future<void> loadHtmlString(
     String html, {
     String? baseUrl,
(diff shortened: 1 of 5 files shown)


packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart
```diff
@@ -87,6 +87,12 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
     return _channel.invokeMethod<void>('loadFile', absoluteFilePath);
   }
 
+  @override
+  Future<void> loadFlutterAsset(String key) async {
+    assert(key.isNotEmpty);
+    return _channel.invokeMethod<void>('loadFlutterAsset', key);
```
stuartmorgan, line 93: I missed this before with `loadFile`; for both of these methods, the docs claim that they throw `ArgumentError`s in specific cases, but there's no mechanism for that here. Doesn't this method channel implementation need to have conversion logic that translates a specific error (e.g., by string matching on the error code) into an `ArgumentError`?

stuartmorgan (approved):
LGTM with one comment.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart
mvanbeusekom: That is a good catch, I missed that completely. 

Actually I need to look into this as I tested on Android where is maps to the Android `WebView.loadUrl` method which will simply load a 404 page not found page and doesn't throw any exception at all. If the behaviour on iOS is similar maybe it would be better to remove the documentation on throwing an `ArgumentException`.

stuartmorgan: On iOS we'd know that the resource lookup didn't return a path. Could we do that check on Android before passing the URL to the webview?

Changing the documentation would be okay, but I'd rather not have wildly different behavior on Android vs other platforms.

mvanbeusekom: I ran some tests and on iOS an error is reported through the `WebViewPlatformCallbacksHandler.onWebResourceError` callback. On Android no error is reported and it just loads the default 404 page. 

I guess for the `loadFile` which excepts a local path we can simply check the file system if the file exists using `dart:io`. For the `loadFlutterAsset` method this will be more difficult unless there is a way we can verify (on the Dart side) that the supplied asset exists. I am not sure there is a good solution otherwise.

stuartmorgan: > unless there is a way we can verify (on the Dart side)

Why does it have to be on the Dart side? I meant querying the engine API to resolve a resource. We can expose that capability to the Dart code via method channel.

mvanbeusekom: The problem is that on the Android the native side only exposes the `loadUrl` and `loadData` methods (as it matches the native WebView API). This means on the Android side we cannot distinguish between the Dart `loadFile`, `loadFlutterAsset` and `loadUrl` methods as they all call the `loadUrl` method. 

How would we go about querying the engine API? I don't have any experience with that.

stuartmorgan: Hm, it looks like the engine always gives an answer on Android by constructing the path that it *would* be at, without actually checking that it's there. But I assume [this](https://developer.android.com/reference/android/content/res/AssetManager#list(java.lang.String)) would return `null` or `[]` for an asset that doesn't exist?

mvanbeusekom: That makes sense, I will try to confirm this tonight. If this works, would it then be appropriate to create a separate method channel call to expose this to Dart within the webview_flutter_android package?

stuartmorgan: Yes, that seems like the right approach given the structure of all the other calls in the current implementation.

mvanbeusekom: I have updated the current implementation to correctly convert the `PlatformException` into a `ArgumentError` as mentioned in the documentation. Not sure if you want to give it another look since this change adds some additional code (including tests).

packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart
```diff
@@ -84,7 +84,31 @@ class MethodChannelWebViewPlatform implements WebViewPlatformController {
...
+  }
+
+  @override
+  Future<void> loadFlutterAsset(String key) async {
+    assert(key.isNotEmpty);
+
+    try {
+      return await _channel.invokeMethod<void>('loadFlutterAsset', key);
+    } on PlatformException catch (ex) {
+      if (ex.code == 'loadFlutterAsset_failed') {
```
stuartmorgan, line 106: Since we don't have to be compatible with an existing implementation (unlike loadFile) let's give this a specific code like `loadFlutterAsset_invalidKey`. If we ever have failure cases that aren't due to a bad key for some reason, we would't want them mapped to `ArgumentError`.

packages/webview_flutter/webview_flutter_platform_interface/test/src/method_channel/webview_method_channel_test.dart
```diff
@@ -73,6 +90,52 @@ void main() {
...
+        () => webViewPlatform.loadFlutterAsset('invalid key'),
+        throwsA(
+          isA<ArgumentError>().having(
+            (ArgumentError error) => error.message,
+            'message',
+            'Failed loading asset.',
+          ),
+        ),
+      );
+    });
```
stuartmorgan, line 137: Please also add tests for both that errors with different codes aren't remapped to `ArgumentError`s.

stuartmorgan (approved):
LGTM with nits

PR #4479, opened 2021-11-05
[webview_flutter] Implement `loadRequest` in Android package.

BeMacized:
This PR is a partial resubmission of my colleague's PR: #4169. 

It implements the `loadRequest` method added in #4450 for the Android package.

As the Android package is currently being refactored to make use of pigeon, this PR still requires some additional refactoring. It is still missing some tests, but I'll delay writing these until this has taken place. I'll keep this PR in draft state until this has all been done. 

Related issue:
- flutter/flutter#27730

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Note that unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I [updated pubspec.yaml](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates) with an appropriate new version according to the [pub versioning philosophy].
- [ ] I updated CHANGELOG.md to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
  - Mostly complete, still some extra tests required.
- [x] All existing and new tests are passing.

PR #4563, opened 2021-12-01
[webview_flutter] Implement loadRequest in Android package.

BeMacized:
This PR supersedes #4479. 

It implements the `loadRequest` method added in #4450 for the Android package.

Related issue:
- flutter/flutter#27730

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_android/pubspec.yaml
```diff
@@ -19,6 +19,7 @@ flutter:
 dependencies:
   flutter:
     sdk: flutter
+  http: ^0.13.4
```
stuartmorgan, line 22: We try to limit dependencies as much as possible. It seems like we should strongly consider having this just be a documented limitation on Android, and list this as a workaround. Someone could even publish it as a separate package.

BeMacized: _Just for context: The native [`postUrl`](https://developer.android.com/reference/android/webkit/WebView#postUrl(java.lang.String,%20byte[])) method for webviews on Android lacks the ability to set http headers. As such, when headers are to be set, the request has to be made manually and its response be loaded in with [`loadDataWithBaseURL`](https://developer.android.com/reference/android/webkit/WebView#loadDataWithBaseURL(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String))._

@stuartmorgan I'm not sure documenting this issue instead of providing a working implementation would be the right way to go:

Personally I'd believe not being able to use `loadRequest` for making post requests with headers on Android is a pretty major oversight, as it would prevent users from, for example, passing tokens through Authorization headers without doing their own workaround. 

Coupled with the fact that the `http` package is not just any random package, but an official dart-lang package, I'd urge you to consider having it as a dependency an acceptable tradeoff for users not having to deal with a platform specific shortcoming.

stuartmorgan: > Personally I'd believe not being able to use `loadRequest` for making post requests with headers on Android is a pretty major oversight, as it would prevent users from, for example, passing tokens through Authorization headers without doing their own workaround.

I understand the use case, but I'm not seeing what compelling advantage is coming from having the workaround be embedded into the package rather than in another package.

I'm also concerned about unexpected side-effects here: I suspect that there are cases where loading data isn't going to have identical behavior to loading a site normally would, and with this approach there's no indication to clients of this package that those things would happen. If they are using a separate package for the workaround, they will be explicitly opting into those differences.

> Coupled with the fact that the `http` package is not just any random package, but an official dart-lang package, I'd urge you to consider it an acceptable tradeoff

My concern in this case with the dependency is not about random vs. official, but about compatibility. Version conflicts in transitive dependencies with `http` is a very real possibility.

> having it as a dependency for users not having to deal with a platform specific shortcoming. 

Embedding the workaround into the package forces *every* user of `webview_flutter`—most of whom will never do post requests with headers—to suffer the potential consequences of version conflicts with `http`.

stuartmorgan: Also: we can easily add this later if it turns out that having it be a separate package is extremely onerous for lots of users. Removing it if we add it now and regret it, on the other hand, is a breaking change.

BeMacized: > I understand the use case, but I'm not seeing what compelling advantage is coming from having the workaround be embedded into the package rather than in another package.

@stuartmorgan The advantage here is usability. As a user, finding out I have to resort to workarounds or separate packages to just basically "patch" functionality that I expected to work right out of the box, isn't a great experience. Having it embedded would therefore (to me) be a compelling advantage.

That said, I do understand your concerns now, so I've removed the workaround and documented the limitation + workaround in the method doc, as well as for the method and readme for the app facing package (#4573).

BeMacized: Just an extra thought: 

The need for this dependency stemmed from our move to pigeon. Before mapping the native api's one to one, Android's native HTTP client was used for this workaround. 

Maybe a possibility for the future could be to also map parts of Android's bundled HttpClient so that this workaround can still be implemented without requiring a new dependency. (Although this would neither account for differences between making a request and doing a "regular" page load)

stuartmorgan: > Maybe a possibility for the future could be to also map parts of Android's bundled HttpClient so that this workaround can still be implemented without requiring a new dependency.

That would definitely be much better than adding `http`.

> We could potentially get around this by making the request manually for every call to loadRequest?

My concern is that there are going to be subtle things that are broken; unnecessarily breaking those things all the time for consistency doesn't address my core concern, it makes it much worse.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -174,6 +174,43 @@ class WebViewAndroidPlatformController extends WebViewPlatformController {
     return webView.loadUrl(url, headers ?? <String, String>{});
   }
 
+  @override
+
+  /// Makes a specific HTTP request ands loads the response in the webview.
```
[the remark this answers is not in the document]
stuartmorgan: Although actually, everything here that's copy/pasted from the main comment should be removed. @overrides shouldn't re-comment the interface.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -174,6 +174,43 @@ class WebViewAndroidPlatformController extends WebViewPlatformController {
...
+  ///
+  /// If [WebViewRequest.headers] is not empty, its key-value pairs will be
+  /// added as the headers for the request.
+  ///
+  /// If [WebViewRequest.body] is not null, it will be added as the body
+  /// for the request.
+  ///
+  /// Throws an ArgumentError if [WebViewRequest.uri] has empty scheme.
+  ///
+  /// Android only:
```
stuartmorgan, line 192: This paragraph is the only part that should stay. This is the Android package though, so you don't need "Android only".

packages/webview_flutter/webview_flutter_android/test/webview_android_widget_test.dart
```diff
@@ -342,6 +343,98 @@ void main() {
...
+              method: WebViewRequestMethod.post,
+              body: body));
+
+          verify(mockWebView.postUrl(
+            'https://www.google.com',
+            body,
+          ));
+        });
+
+        testWidgets('POST ignores headers', (WidgetTester tester) async {
```
stuartmorgan, line 420: Why are we testing that something we wished worked doesn't work?

BeMacized: Test removed

stuartmorgan (approved):
LGTM!

stuartmorgan:
Overriding incorrect submit-queue status.

PR #4565, opened 2021-12-02
[webview_flutter] Proof of concept web implementation

BeMacized:
This PR adds a very limited proof of concept for a web implementation of the webview_flutter plugin in a separate package named webview_flutter_web.

Due to various limits on the web platform, this version currently only supports a few features.

Because of its limited functionality, it currently is not compatible with the example of the app-facing plugin, only its own included example. 

~~The provided shims for dart:ui are still triggering analysis warnings that I'm not sure how to resolve.
(Something to do with flutter/flutter#55000). I cannot seem to disable all of the warnings that occur for the shim files, and I have not been able to figure out if providing a custom `analysis_options.yaml` that excludes the shim directory is a possibility here.~~

Likewise a check fails due to there being no integration tests. As far as I know these are not a thing yet for the web platform, but I'm not sure how to disable the check.

Relevant issue:
- flutter/flutter#45295

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_web/AUTHORS
```diff
@@ -0,0 +1,69 @@
...
+Eitan Schwartz <eshvartz@gmail.com>
+Chris Rutkowski <chrisrutkowski89@gmail.com>
+Juan Alvarez <juan.alvarez@resideo.com>
+Aleksandr Yurkovskiy <sanekyy@gmail.com>
+Anton Borries <mail@antonborri.es>
+Alex Li <google@alexv525.com>
+Rahul Raj <64.rahulraj@gmail.com>
+Maurits van Beusekom <maurits@baseflow.com>
+Nick Bradshaw <nickalasb@gmail.com>
+
```
stuartmorgan, line 69: New packages can start with a clean list, containing just Google and yourselves.

packages/webview_flutter/webview_flutter_web/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+## 1.0.0
```
stuartmorgan, line 1: Let's not call this 1.0.0. 0.1.0 maybe?

packages/webview_flutter/webview_flutter_web/example/pubspec.yaml
```diff
@@ -0,0 +1,32 @@
...
+  sdk: ">=2.14.0 <3.0.0"
+
+dependencies:
+  flutter:
+    sdk: flutter
+  flutter_web_plugins:
+    sdk: flutter
+  webview_flutter_web:
+    # When depending on this package from a real application you should use:
+    #   webview_flutter: ^x.y.z
```
stuartmorgan, line 15: webview_flutter_web

packages/webview_flutter/webview_flutter_web/lib/shims/dart_ui_fake.dart
```diff
@@ -0,0 +1,30 @@
...
+      String viewTypeId, html.Element Function(int viewId) viewFactory) {}
+}
+
+/// Shim for web_ui engine.AssetManager.
+/// https://github.com/flutter/engine/blob/master/lib/web_ui/lib/src/engine/assets.dart#L12
+class webOnlyAssetManager {
+  /// Shim for getAssetUrl.
+  /// https://github.com/flutter/engine/blob/master/lib/web_ui/lib/src/engine/assets.dart#L45
+  static getAssetUrl(String asset) {}
+}
```
stuartmorgan, line 27: See https://github.com/flutter/plugins/blob/2ab3044cfbb4ee2f3dcbc5741628da8f4e0eb437/packages/url_launcher/url_launcher_web/lib/src/shims/dart_ui_fake.dart#L15-L30 for a version of these that passes analysis; I just fixed this for url_launcher.

packages/webview_flutter/webview_flutter_web/pubspec.yaml
```diff
@@ -0,0 +1,33 @@
...
+  webview_flutter_platform_interface: ^1.7.0
+
+dev_dependencies:
+  flutter_driver:
+    sdk: flutter
+  flutter_test:
+    sdk: flutter
+  mockito: ^5.0.0
+  pedantic: ^1.10.0
+  pigeon: 1.0.7
```
stuartmorgan, line 33: Not needed

packages/webview_flutter/webview_flutter_web/pubspec.yaml
```diff
@@ -0,0 +1,33 @@
...
+    sdk: flutter
+  webview_flutter_platform_interface: ^1.7.0
+
+dev_dependencies:
+  flutter_driver:
+    sdk: flutter
+  flutter_test:
+    sdk: flutter
+  mockito: ^5.0.0
+  pedantic: ^1.10.0
```
stuartmorgan, line 32: Not needed.

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
```diff
@@ -0,0 +1,182 @@
...
+  Future<void> loadHtmlString(
+    String html, {
+    String? baseUrl,
+  }) async {
+    throw UnimplementedError();
+  }
+
+  @override
+  Future<void> loadRequest(WebViewRequest request) {
+    throw UnimplementedError();
```
stuartmorgan, line 180: Shouldn't this at least trivial `get` requests, and only throw for the rest, so that this package doesn't stop working once the app-facing package changes over to loadRequest?

packages/webview_flutter/webview_flutter_web/README.md
```diff
@@ -0,0 +1,3 @@
+# webview\_flutter\_web
+
+The Web implementation of [`webview_flutter`][1].
```
stuartmorgan, line 3: This definitely needs some discussion of the (many) caveats. E.g., explicitly list the supported methods, and say that nothing else works.

It should also a mention of the fact that it's unendorsed, like https://github.com/flutter/plugins/tree/master/packages/google_maps_flutter/google_maps_flutter_web#depend-on-the-package

stuartmorgan (changes requested):
A few notes from a quick skim; I'll leave the full review to @ditman

packages/webview_flutter/webview_flutter_web/CHANGELOG.md
BeMacized: That would be more accurate yes.

packages/webview_flutter/webview_flutter_web/lib/shims/dart_ui_fake.dart
BeMacized: Ah thank you, that seems to do the trick.

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
BeMacized: Yes. When I wrote this implementation this was not yet part of the platform interface, so I had not yet gotten around to it. It's now implemented together with `loadHtmlString` as well.

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
```diff
@@ -0,0 +1,282 @@
...
+    final HttpRequest httpReq = await _httpRequestFactory.request(request.uri.toString(),
+        method: request.method.serialize(), requestHeaders: request.headers, sendData: request.body);
+    final String contentType = httpReq.getResponseHeader('content-type') ?? 'text/html';
+    print(contentType);
+    _element.src = 'data:$contentType,' + Uri.encodeFull(httpReq.responseText ?? '');
+  }
+}
+
+/// Factory class for creating [HttpRequest] instances.
+class HttpRequestFactory {
```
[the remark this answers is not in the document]
ditman: (This is used to be mocked in tests, to ensure that the `WebViewRequest` is correctly parsed; I have a comment about this approach below)

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
```diff
@@ -0,0 +1,282 @@
...
+  ///         // Do something with the response.
+  ///     });
+  ///
+  /// Note that requests for file:// URIs are only supported by Chrome extensions
+  /// with appropriate permissions in their manifest. Requests to file:// URIs
+  /// will also never fail- the Future will always complete successfully, even
+  /// when the file cannot be found.
+  ///
+  /// See also: [authorization headers](http://en.wikipedia.org/wiki/Basic_access_authentication).
+  Future<HttpRequest> request(String url,
```
ditman, line 265: I think this would be more useful if it was:

```dart
Future<HttpRequest> fromWebViewRequest(WebViewRequest webViewRequest) {
  ...
  return HttpRequest.request(webViewRequest.url,
    ...
    );
```

then have the `HttpRequestFactory` class (or maybe the bare `fromWebViewRequest` function) unit-tested with as many weird WebViewRequest examples as needed, rather than having to make this an overrideable property just to see that the `loadRequest` method is extracting the correct values from the WebViewRequest object, and passing the expected values to this method.

What do you think?

(This is a completely private API, so this can be done later, if wanted)

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
```diff
@@ -0,0 +1,282 @@
...
+        ));
+      },
+    );
+  }
+
+  @override
+  Future<bool> clearCookies() async => false;
+
+  /// Gets called when the plugin is registered.
+  static void registerWith(Registrar registrar) {}
```
ditman, line 60: Shouldn't this method be registering the `instance` of the plugin to the WebWebViewPlatform class? There's no global `WebViewPlugin.instance` anymore? Or is this the new way of doing federation?

PS: let's say we want to add web support to the webview (core) example [here](https://github.com/flutter/plugins/tree/master/packages/webview_flutter/webview_flutter/example). Can it be done?

packages/webview_flutter/webview_flutter_web/example/lib/web_view.dart
```diff
@@ -0,0 +1,385 @@
...
+/// A web view widget for showing html content.
+///
+/// The [WebView] widget wraps around the [WebWebViewPlatform].
+///
+/// The [WebView] widget is controlled using the [WebViewController] which is
+/// provided through the `onWebViewCreated` callback.
+///
+/// In this example project it's main purpose is to facilitate integration
+/// testing of the `webview_flutter_web` package.
+class WebView extends StatefulWidget {
```
ditman, line 25: I find it a little bit odd that an example app needs all this boilerplate, shouldn't this be provided by the core plugin? Or is this here to prevent a circular dependency?

Is this a reduced copy of https://github.com/flutter/plugins/blob/master/packages/webview_flutter/webview_flutter/lib/src/webview.dart ?

ditman (approved):
This LGTM, I've only left a bunch of minor nits, and some confused questions about the New Way of federating things :)

stuartmorgan:
> and some confused questions about the New Way of federating things :)

`webview_flutter` doesn't have a new way of federating things, it just doesn't have the modern way. It was (IIUC) a sort of prototype of the platform interface, and we federated it without my realizing that—and thus without us cleaning it up first. See https://github.com/flutter/flutter/issues/94051 (the second bullet point I think is where some of your confusion is coming from)

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
```diff
@@ -0,0 +1,282 @@
...
+  ///
+  /// Throws an ArgumentError if [WebViewRequest.uri] has empty scheme.
+  Future<void> loadRequest(WebViewRequest request) async {
+    if (!request.uri.hasScheme) {
+      throw ArgumentError('WebViewRequest#uri is required to have a scheme.');
+    }
+    final HttpRequest httpReq = await _httpRequestFactory.request(request.uri.toString(),
+        method: request.method.serialize(), requestHeaders: request.headers, sendData: request.body);
+    final String contentType = httpReq.getResponseHeader('content-type') ?? 'text/html';
+    print(contentType);
```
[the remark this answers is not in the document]
BeMacized: Nope sorry, it was left from debugging. Removing

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
```diff
@@ -0,0 +1,282 @@
...
+
+  @override
+  Future<void> loadHtmlString(
+    String html, {
+    String? baseUrl,
+  }) async {
+    _element.src = 'data:text/html,' + Uri.encodeFull(html);
+  }
+
+  @override
```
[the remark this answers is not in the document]
BeMacized: Indeed, this was sloppy

packages/webview_flutter/webview_flutter_web/example/lib/web_view.dart
BeMacized: That's correct. The other examples from the android and ios packages have a similar amount of boilerplate, as they all have a reduced copy of this class.

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
BeMacized: I have to confess that I'm not entirely sure. Afaik the federation is like Stuart mentioned not very standard on this plugin. I assumed this method would not be needed, but the `generated_plugin_registrant.dart` in the example does require it to be here. @mvanbeusekom maybe you have some extra info?

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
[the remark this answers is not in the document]
BeMacized: The only reason this factory exists, is so that `HttpRequest.request` can be mocked in the tests. 
Ideally we wouldn't have this class at all, but I'm not sure how else I'd mock it. 

Having the user supply the HttpRequest object is not really an option, because as far as I can see, the http request is fired immediately when `HttpRequest.request(...)` is called. 

If you know of a better way to do this, I'd be very interested as it feels like a bit much just for a mock.

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
BeMacized: As mentioned above, this factory only exists for directly mocking `HttpRequest.request`. If this is something we need to unit test please let me know, but I assumed it would not be the case as it's nothing more than a direct pass-through.

Whether the correct request is generated from the `WebViewRequest` in `loadRequest` should already be unit tested.

ditman:
> See flutter/flutter#94051 (the second bullet point I think is where some of your confusion is coming from)

Ah, yes @stuartmorgan, that issue helped to understand what's going on in this PR, thanks for the link!

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
ditman: Yep, this is normally what we use to set the "instance" of the platform-specific plugin in other plugins (see [url_launcher](https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart#L46-L50) for example). However this plugin is not federated exactly the same as others, so that's why this method is empty (see [comment](https://github.com/flutter/plugins/pull/4565#issuecomment-987194281)).

ditman:
```
Running "flutter pub get" in webview_flutter_web...              2,228ms
org-dartlang-app:///webview_flutter_web_test.mocks.dart:735:43: Error: The parameter 'value' of the method 'MockIFrameElement.setAttribute' has type 'String?', which does not match the corresponding type, 'Object', in the overridden method, 'Element.setAttribute'.
 - 'Object' is from 'dart:core'.
Change to a supertype of 'Object', or, for a covariant parameter, a subtype.
  void setAttribute(String? name, String? value) =>
                                          ^
org-dartlang-sdk:///dart-sdk/lib/html/dart2js/html_dart2js.dart: Context: This is the overridden method ('setAttribute').
org-dartlang-app:///webview_flutter_web_test.mocks.dart:739:67: Error: The parameter 'value' of the method 'MockIFrameElement.setAttributeNS' has type 'String?', which does not match the corresponding type, 'Object', in the overridden method, 'Element.setAttributeNS'.
 - 'Object' is from 'dart:core'.
Change to a supertype of 'Object', or, for a covariant parameter, a subtype.
  void setAttributeNS(String? namespaceURI, String? name, String? value) =>
                                                                  ^
org-dartlang-sdk:///dart-sdk/lib/html/dart2js/html_dart2js.dart: Context: This is the overridden method ('setAttributeNS').
Failed to compile
```

Is this a problem with the generated mock?

BeMacized:
> Is this a problem with the generated mock?

@ditman Seems like it. As this is from a generated file, I'm not really sure how to go about fixing it. I've regenerated the mocks just to be sure, but it doesn't seem to have an effect.

packages/webview_flutter/webview_flutter_web/AUTHORS
```diff
@@ -0,0 +1,9 @@
+# Below is a list of people and organizations that have contributed
+# to the Flutter project. Names should be added to the list like so:
+#
+#   Name/Organization <email address>
+
+Google Inc.
+The Chromium Authors
```
stuartmorgan, line 7: This isn't Chromium code, so this can go :)

packages/webview_flutter/webview_flutter_web/README.md
```diff
@@ -0,0 +1,20 @@
...
+- `loadHTMLString` (Without `baseUrl`)
+- Setting the `initialUrl` through `CreationParams`.
+
+Nothing else is currently supported.
+
+## Usage
+
+### Depend on the package
+
+This package is not an endorsed implementation of the google_maps_flutter plugin yet, so you'll need to
```
stuartmorgan, line 19: s/google_maps_flutter/`webview_flutter`/ (with backticks around the corrected name for consistency with the first line of the README).

packages/webview_flutter/webview_flutter_web/lib/webview_flutter_web.dart
```diff
@@ -0,0 +1,292 @@
...
+  /// [WebViewRequest.method] must be one of the supported HTTP methods
+  /// in [WebViewRequestMethod].
+  ///
+  /// If [WebViewRequest.headers] is not empty, its key-value pairs will be
+  /// added as the headers for the request.
+  ///
+  /// If [WebViewRequest.body] is not null, it will be added as the body
+  /// for the request.
+  ///
+  /// Throws an ArgumentError if [WebViewRequest.uri] has empty scheme.
```
stuartmorgan, line 198: Here and below (as with my comment in the other PR) please remove anything that's just duplicated from the method it overrides; overrides comments should only cover things specific to the override.

stuartmorgan (changes requested):
A few minor things

stuartmorgan:
> > Is this a problem with the generated mock?
> 
> @ditman Seems like it. As this is from a generated file, I'm not really sure how to go about fixing it. I've regenerated the mocks just to be sure, but it doesn't seem to have an effect.

Are there non-web shims for the class being mocked? If so, perhaps those shims have the wrong types.

If minor edits to the generated file work, then let's just do that (with comments), and then follow up with an issue to try to find a better solution. If minor edits aren't enough, can you make a manual fake that does what you need? It looks like you're not using that many methods/properties.

ditman:
In this case, the class being problematic seems to be `dart:html IFrameElement`; maybe you can use a new `IFrameElement()` and pass it around, instead of a mock?

ditman:
Looks like this change: https://github.com/dart-lang/sdk/commit/017850fe5be47bd5e1be7331329db2cae5a4d99b

stuartmorgan:
I had the same thought; I'm locally testing changing to Object? to see if stable is happy.

stuartmorgan:
It's also missing integration tests; I'm trying to figure out if we can useful test anything without being able to inspect the iframe. It's probably useful to at least have something that builds the whole plugin and checks that calling `loadUrl` doesn't explode even if we can't see that it worked...

stuartmorgan:
> I had the same thought; I'm locally testing changing to Object? to see if stable is happy.

Yep. I'll push that shortly once I figure out what to do with integration tests.

stuartmorgan:
Wait, why does the example only have post when post doesn't work on web? @BeMacized do you have an updated example already, or should I fix it now?

ditman:
@stuartmorgan it seems that POST does work on web, as long as CORS/same domain is initialized. The only thing POST seems to do is to read the response body from the request, and render it as text on the iframe via a data: URL (if I understood the code correctly)

stuartmorgan:
Oh, I missed that; I thought the only thing we were supporting was setting the `src`. Maybe I'll just add a simple `loadUrl` option then, rather than changing it.

BeMacized:
> it seems that POST does work on web, as long as CORS/same domain passes.

This is correct.

> It's also missing integration tests

I left these out, as when I tried writing these, I got an error message telling me that integrations tests were not yet supported on web.

stuartmorgan:
> Oh, I missed that; I thought the only thing we were supporting was setting the `src`. Maybe I'll just add a simple `loadUrl` option then, rather than changing it.

I see, this is just me being unfamiliar with the webview_flutter examples. The main page does the standard simple load, and then there's a menu for other things. This is the subset of that standard example that works here. It's a bit odd in the context of this package that doesn't have many options, but that's fine.

I'll look at a simple integration test though.

> I left these out, as when I tried writing these, I got an error message telling me that integrations tests were not yet supported on web.

Where you using `flutter test` rather than `flutter drive` maybe?

BeMacized:
> I'll look at a simple integration test though.
>
> Where you using `flutter test` rather than `flutter drive` maybe?

This could very well have been the case yes. I can take another look at writing some integration tests tomorrow morning if needed.

stuartmorgan:
@BeMacized I have very minimal integration tests ready, and a fix for the unit test. Can you open this branch to pushes from Flutter contributors so I can push them?

stuartmorgan:
Closing in favor of https://github.com/flutter/plugins/pull/4594, which fixed/added tests.

PR #4575, opened 2021-12-03
[flutter_plugin_tools] Add a new 'make-deps-path-based' command

stuartmorgan:
Adds a new command that adds `dependency_overrides` to any packages in the repository that depend on a list of target packages, including an option to target packages that will publish a non-breaking change in a given diff.

Adds a new CI step that uses the above in conjunction with a new `--run-on-dirty-packages` to adjust the dependencies of anything in the repository that uses a to-be-published package and then re-run analysis on just those packages. This will allow us to catch in presubmit any changes that are not breaking from a semver standpoint, but will break us due to our strict analysis in CI.

Fixes https://github.com/flutter/flutter/issues/89862

## Pre-launch Checklist

- [x] I signed the [CLA].


gaaclarke (commented):
I think "pathify" is is a bit difficult to grok.  I know it's a pain to change in the PR but I think "convert_depenencies_to_relative_paths" would be more clear.

script/tool/lib/src/pathify_command.dart
```diff
@@ -0,0 +1,244 @@
...
+    GitDir? gitDir,
+  }) : super(packagesDir, gitDir: gitDir) {
+    argParser.addMultiOption(_targetPackagesArg,
+        help: 'The names of the packages to convert to dependencies.\n'
+            'Ignored if --$_targetNonBreakingUpdatePackagesArg is passed.',
+        valueHelp: 'some_package');
+    argParser.addFlag(
+      _targetNonBreakingUpdatePackagesArg,
+      help: 'Causes all pacakges that have non-breaking version changes '
+          'relative to the git base to be treated as target packages.',
```
gaaclarke, line 39: "Relative to the git base" is a bit confusing for me.  Maybe "when compared against the git base".  "Relative" is less precise and a loaded term when operating on paths.

script/tool/lib/src/pathify_command.dart
```diff
@@ -0,0 +1,244 @@
...
+/// plugin would cause post-publish analyzer failures in another package of that
+/// plugin.
+class PathifyCommand extends PluginCommand {
+  /// Creates an instance of the pathify command.
+  PathifyCommand(
+    Directory packagesDir, {
+    GitDir? gitDir,
+  }) : super(packagesDir, gitDir: gitDir) {
+    argParser.addMultiOption(_targetPackagesArg,
+        help: 'The names of the packages to convert to dependencies.\n'
```
gaaclarke, line 33: This is recursive, right?  We should mention that.

script/tool/test/common/git_version_finder_test.dart
```diff
@@ -88,6 +88,17 @@ file2/file2.cc
...
+
+  test('include uncommitted files if requested', () async {
+    const String customBaseSha = 'aklsjdcaskf12312';
+    gitDiffResponse = '''
+file1/pubspec.yaml
+file2/file2.cc
+''';
+    final GitVersionFinder finder = GitVersionFinder(gitDir, customBaseSha);
+    await finder.getChangedFiles(includeUncommitted: true);
+    verify(gitDir.runCommand(<String>['diff', '--name-only', customBaseSha]));
```
gaaclarke, line 100: A comment that you are looking for the absence of "HEAD" in the args would help make this more clear.

script/tool/test/pathify_command_test.dart
```diff
@@ -0,0 +1,283 @@
...
+
+    final List<String> output =
+        await runCapturingPrint(runner, <String>['pathify']);
+
+    expect(
+      output,
+      containsAllInOrder(<Matcher>[
+        contains('No target packages'),
+      ]),
+    );
```
gaaclarke, line 73: Instead of looking at the tools output we should be looking at the pubspec file changes, no?

script/tool/lib/src/pathify_command.dart
```diff
@@ -0,0 +1,244 @@
...
+    argParser.addFlag(
+      _targetNonBreakingUpdatePackagesArg,
+      help: 'Causes all pacakges that have non-breaking version changes '
+          'relative to the git base to be treated as target packages.',
+    );
+  }
+
+  static const String _targetPackagesArg = 'target-packages';
+  static const String _targetNonBreakingUpdatePackagesArg =
+      'target-non-breaking-update-packages';
```
gaaclarke, line 45: What are "update packages"?  Does this mean "updated packages"?

.cirrus.yml
```diff
@@ -109,13 +109,24 @@ task:
...
-      script:
+      analyze_script:
         # DO NOT change the custom-analysis argument here without changing the Dart repo.
         # See the comment in script/configs/custom_analysis.yaml for details.
         - ./script/tool_runner.sh analyze --custom-analysis=script/configs/custom_analysis.yaml
+      pathified_analyze_script:
+        # Re-run analysis with path-based dependencies to ensure that publishing
+        # the changes won't break analysis of other packages in the respository
+        # that depend on it.
+        - ./script/tool_runner.sh pathify --target-non-breaking-update-packages
```
gaaclarke, line 123: Related to my other comment, unfortunately I think this is harder to grok than it could be.

`./script/tool_runner.sh convert_dependencies_to_paths --convert-non-breaking-updated-packages` is slightly better.

gaaclarke (approved):
The code LGTM.  I read the change to the yml file first and didn't really understand what was happening until I read through the PR.  I have the feeling that we are increasing the complexity more so we should do our best to make sure the documentation / naming is as clear as we can make it.

script/tool/lib/src/pathify_command.dart
stuartmorgan: This was hard to name; I tried a few versions. The goal is to target packages which have non-breaking updates. Maybe `target-packages-with-non-breaking-updates` is the best option.

script/tool/lib/src/pathify_command.dart
stuartmorgan: It's not recursive; only direct dependencies are handled. There's no need for it to be recursive because the whole repository is updated; any transitive dependencies will be through other packages in the repository.

script/tool/lib/src/pathify_command.dart
```diff
@@ -0,0 +1,244 @@
...
+    Directory packagesDir, {
+    GitDir? gitDir,
+  }) : super(packagesDir, gitDir: gitDir) {
+    argParser.addMultiOption(_targetPackagesArg,
+        help: 'The names of the packages to convert to dependencies.\n'
+            'Ignored if --$_targetNonBreakingUpdatePackagesArg is passed.',
+        valueHelp: 'some_package');
+    argParser.addFlag(
+      _targetNonBreakingUpdatePackagesArg,
+      help: 'Causes all pacakges that have non-breaking version changes '
```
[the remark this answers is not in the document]
stuartmorgan: Fixed

script/tool/lib/src/pathify_command.dart
stuartmorgan: Done.

script/tool/lib/src/pathify_command.dart
stuartmorgan: Changed.

script/tool/test/common/git_version_finder_test.dart
stuartmorgan: Done.

script/tool/test/pathify_command_test.dart
stuartmorgan: The output matters too, but added a second package with a dependency on the first and validate that its pubspec is unchanged as well.

stuartmorgan (commented):
> I think "pathify" is is a bit difficult to grok. I know it's a pain to change in the PR but I think "convert_depenencies_to_relative_paths" would be more clear.

I liked `pathify`, but you're right that clarity is more important than something that tries to be clever.

I went with `make-deps-path-based`; the important thing is that they are path-based, not that they are relative (I could just have easily have done absolute paths since people don't generally move their repos while testing a PR), and it's significantly shorter than `convert-dependencies-to-path-based` without sacrificing clarity.

stuartmorgan:
I also changed `target-packages` to `target-dependencies` (and similarly for the other flag) since I realized that using `packages` was ambiguous. I.e., "target packages" could just as easily mean the packages whose pubspecs are being rewritten, which isn't the case.

PR #4581, opened 2021-12-06
[webview_flutter] Android implementation of `loadFlutterAsset` method.

mvanbeusekom:
Android implementation of the `loadFlutterAsset` method of the platform interface.

Related issue: flutter/flutter#27086

## Pre-launch Checklist

- [x] I signed the [CLA].


bparrishMines (approved):
LGTM

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
```diff
@@ -120,7 +126,9 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding binding) {
         binding.getBinaryMessenger(),
         binding.getPlatformViewRegistry(),
         binding.getApplicationContext(),
-        null);
+        null,
+        new FlutterAssetManager.PluginBindingFlutterAssetManager(
+            binding.getApplicationContext().getAssets(), binding.getFlutterAssets()));
```
stuartmorgan, line 131: If you're using the application context, doesn't this need updating in `updateContext`?

packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/PluginBindingFlutterAssetManagerTest.java
```diff
@@ -0,0 +1,54 @@
...
+    try {
+      when(mockAssetManager.list("test/path"))
+          .thenReturn(new String[] {"index.html", "styles.css"});
+      String[] actualFilePaths = tesPluginBindingFlutterAssetManager.list("test/path");
+      verify(mockAssetManager).list("test/path");
+      assertArrayEquals(new String[] {"index.html", "styles.css"}, actualFilePaths);
+    } catch (IOException ex) {
+      fail();
+    }
+  }
```
stuartmorgan, line 47: There should be a test ensuring that a null return from `AssetManager.list` is handled correctly. Especially since that will need to be converted to an empty list at the Pigeon API boundary.

packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/RegistrarFlutterAssetManagerTest.java
```diff
@@ -0,0 +1,55 @@
...
+    try {
+      when(mockAssetManager.list("test/path"))
+          .thenReturn(new String[] {"index.html", "styles.css"});
+      String[] actualFilePaths = testRegistrarFlutterAssetManager.list("test/path");
+      verify(mockAssetManager).list("test/path");
+      assertArrayEquals(new String[] {"index.html", "styles.css"}, actualFilePaths);
+    } catch (IOException ex) {
+      fail();
+    }
+  }
```
stuartmorgan, line 48: Same.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
```diff
@@ -134,6 +143,11 @@ class WebViewAndroidPlatformController extends WebViewPlatformController {
   /// This should only be changed for testing purposes.
   final WebViewProxy webViewProxy;
 
+  /// Manages access to Flutter assets that are part of the Android App bundle.
+  ///
+  /// This should only be changed for testing purposes.
```
stuartmorgan, line 148: Sounds like it should be annotated `@visibleForTesting` then

stuartmorgan (changes requested):

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
mvanbeusekom: I thought this would only be the case for the `Activity` context and the application context stays the same during the live time of the App.

packages/webview_flutter/webview_flutter_android/lib/webview_android_widget.dart
mvanbeusekom: My reasoning was that since the member is marked `final` it can only be set through the constructor and the constructor parameter is already set to `@visibleForTesting` it wouldn't be necessary to mark this as visible for testing (to be consistent with the `webViewProxy` field).

packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/PluginBindingFlutterAssetManagerTest.java
mvanbeusekom: I have added the test `` to the FlutterAssetManagerHostApiImplTest.java in my latest commit (
08023b2). The `PluginBindingFlutterAssetManager` and `RegistrarFlutterAssetManager` implementations simply relays the result of the `AssetManager.list` method to the `FlutterAssetManagerHostApiImpl` class. The later is responsible for preparing the result in a way that it can be send on the method channel.

This was a good catch though, it found a bug that I overlooked. Many thanks.

packages/webview_flutter/webview_flutter_android/pubspec.yaml
```diff
@@ -2,7 +2,10 @@ name: webview_flutter_android
 description: A Flutter plugin that provides a WebView widget on Android.
 repository: https://github.com/flutter/plugins/tree/master/packages/webview_flutter/webview_flutter_android
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+webview%22
-version: 2.4.0
+version: 2.5.0
+
+# TODO(mvanbeusekom): Remove when webview_flutter_platform_interface no longer referenced through path reference.
+publish_to: none
```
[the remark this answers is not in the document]
mvanbeusekom: Old habit, before we had the `publishable` rule I remember we used the `publish_to: none` to ensure things were not published by accident. You are right, I should not do this anymore. Removed the lines.

packages/webview_flutter/webview_flutter_android/pubspec.yaml
```diff
@@ -19,7 +22,9 @@ flutter:
 dependencies:
   flutter:
     sdk: flutter
-  webview_flutter_platform_interface: ^1.5.2
+  # TODO(mvanbeusekom): Update to pub.dev dependency once version 1.8.0 is released.
```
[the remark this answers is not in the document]
mvanbeusekom: Done.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
stuartmorgan: I'm still an Android n00b; sorry for the noise!

stuartmorgan (approved):
LGTM

PR #4601, opened 2021-12-09
[webview_flutter] Enable setAllowFileAccess on Android setting when loading files

mvanbeusekom:
This makes two changes:

1. It adds the `WebSettings.setAllowFileAccess()` method to provide the option to enable/ disable this feature on Android;
2. Ensures that the `WebSettings.setAllowFileAccess()` is called from the `WebViewAndroidWidget.loadFile()` method to ensure this option is enabled. Before Android API 30 this option is enabled by default, however on Android API 30 and up this behaviour is flipped and the option is now disabled by default which means the `loadFile` method doesn't work on these Android versions.  This bug was not detected before because the webview_flutter_android example app is configured to target Android SDK 28 (details can be found here: https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean)).

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan (approved):
LGTM with the format issues fixed. I was wondering when closing that other PR last night how this worked 🙂 I missed the fact that the example was using an older version than the switch though.

I don't love the automatic permanent change of security setting when loading a file, but I didn't have a suggestion that doesn't feel more annoying for clients. We should do a follow-up to document that side effect in the app-facing method fix comment though.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/GeneratedAndroidWebView.java
```diff
@@ -1,7 +1,3 @@
-// Copyright 2013 The Flutter Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE file.
```
stuartmorgan, line 3: This still needs to be re-added for CI to pass.

mvanbeusekom: Sorry completely missed that. Just submitted an update.

stuartmorgan:
Merging in master will help this pass Android tests

PR #4606, opened 2021-12-10
[path_provider] Fix handling of null application ID

stuartmorgan:
The lookup of application ID was handling `null` (Dart null), but not
`nullptr` (Dart representation of a C null pointer), so was throwing an
exception in applications without an application ID.

Refactors the FFI calls into a wrapper so that they can be faked for
unit testing.

This includes the `shared_preferences_linux` example application, so
this fixes tree breakage.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/path_provider/path_provider_linux/pubspec.yaml
+++ b/packages/path_provider/path_provider_linux/pubspec.yaml
@@ -2,7 +2,7 @@ name: path_provider_linux
 description: Linux implementation of the path_provider plugin
 repository: https://github.com/flutter/plugins/tree/master/packages/path_provider/path_provider_linux
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+path_provider%22
-version: 2.1.3
+version: 2.1.4
 
 environment:
   sdk: ">=2.12.0 <3.0.0"
@@ -19,6 +19,7 @@ dependencies:
   ffi: ^1.1.2
   flutter:
     sdk: flutter
+  meta: ^1.3.0
   path: ^1.8.0
   path_provider_platform_interface: ^2.0.0
   xdg_directories: ^0.2.0
--- a/packages/path_provider/path_provider_linux/test/get_application_id_test.dart
+++ b/packages/path_provider/path_provider_linux/test/get_application_id_test.dart
@@ -1,0 +1,61 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:path_provider_linux/src/get_application_id_real.dart';
+
+class FakeGioUtils implements GioUtils {
+  int? application = 0;
+  Pointer<Utf8>? applicationId = nullptr;
+
+  @override
+  bool libraryIsPresent = false;
+
+  @override
+  int gApplicationGetDefault() => application!;
+
+  @override
+  Pointer<Utf8> gApplicationGetApplicationId(int app) => applicationId!;
+}
+
+void main() {
+  final FakeGioUtils fakeGio = FakeGioUtils();
+
+  setUp(() {
+    gioUtilsOverride = fakeGio;
+  });
+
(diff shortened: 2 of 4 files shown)


stuartmorgan:
> It looks like this pull request may not have tests.

I forgot to `git add` the test file; fixed.

packages/path_provider/path_provider_linux/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: path_provider_linux
 description: Linux implementation of the path_provider plugin
 repository: https://github.com/flutter/plugins/tree/master/packages/path_provider/path_provider_linux
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+path_provider%22
-version: 2.1.3
+version: 2.1.4
```
cyanglaz, line 5: I remember even for @visibleForTesting API updates, we need to bump minor? (I might be wrong)

packages/path_provider/path_provider_linux/test/get_application_id_test.dart
```diff
@@ -0,0 +1,61 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:path_provider_linux/src/get_application_id_real.dart';
+
+class FakeGioUtils implements GioUtils {
+  int? application = 0;
```
cyanglaz, line 11:
```suggestion
  int application = 0;
```

packages/path_provider/path_provider_linux/test/get_application_id_test.dart
```diff
@@ -0,0 +1,61 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:path_provider_linux/src/get_application_id_real.dart';
+
+class FakeGioUtils implements GioUtils {
```
cyanglaz, line 10:
```suggestion
class _FakeGioUtils implements GioUtils {
```

packages/path_provider/path_provider_linux/test/get_application_id_test.dart
```diff
@@ -0,0 +1,61 @@
...
+
+  @override
+  int gApplicationGetDefault() => application!;
+
+  @override
+  Pointer<Utf8> gApplicationGetApplicationId(int app) => applicationId!;
+}
+
+void main() {
+  final FakeGioUtils fakeGio = FakeGioUtils();
```
cyanglaz, line 25: Should this be in setup as well? We probably should avoid having this object stated in tests.

cyanglaz (changes requested):

packages/path_provider/path_provider_linux/pubspec.yaml
stuartmorgan: The file that I added the new content to shouldn't be exported. I just wanted to be extra-obvious about it being test-only.

packages/path_provider/path_provider_linux/test/get_application_id_test.dart
stuartmorgan: Fixed the other way, by removing the default, which is what I meant to do. Originally I made them non-nullable with default values but then it's possible to have tests that are expecting things based on setup that was implicit elsewhere, which makes tests harder to understand. I meant to change them to nullable without defaults to prevent that possibility, but forgot to remove the defaults.

packages/path_provider/path_provider_linux/test/get_application_id_test.dart
stuartmorgan: Done.

packages/path_provider/path_provider_linux/test/get_application_id_test.dart
stuartmorgan: Indeed! Done.

cyanglaz (approved):
LGTM

cyanglaz:
Landing on red to fix tree

PR #4592, opened 2021-12-07
Renew cirrus gcp credentials

keyonghan:
GCP account key expires every three months. This PR updates the GCP credential based on the renewed key.

PR #4608, opened 2021-12-10
[camera]Fix crash due to calling engine APIs from background thread

hellohuanlin:
The engine APIs used in camera plugin are required to be run on platform thread. Some of these APIs have explicit asserts (e.g. `MethodChannel`) and some have warning instructions in documentation (e.g. `TextureRegistry`). However, in camera plugin we are currently calling these APIs in background thread, causing the crash. 

There's already a `FLTThreadSafeFlutterResult` wrapper to invoke `FlutterResult` in main thread. This PR created similar "thread safe wrappers" for `EventChannel`, `MethodChannel` and `TextureRegistry`. 

This PR should fix the crash, but we should start looking into simplifying the threading logic in this plugin when we get time. I also wrote up [some ideas here](https://github.com/flutter/plugins/pull/4619).  

Note that there's a separate crash due to race condition, which is [fixed here](https://github.com/flutter/plugins/pull/4619). 

No version change: 
I want to update the pubspec.yaml file in next PR so that we can batch the change. 

## Issues: 

https://github.com/flutter/flutter/issues/94723
https://github.com/flutter/flutter/issues/52578

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,4 +1,8 @@
-##  0.9.4+5
+## NEXT
+
+* Fixed a crash in iOS when using image stream due to threading issue. 
+
+## 0.9.4+5
 
 * Fixes bug where calling a method after the camera was closed resulted in a Java `IllegalStateException` exception.
 * Fixes integration tests.
--- a/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
+++ b/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
@@ -1,0 +1,44 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import camera;
+@import XCTest;
+#import <OCMock/OCMock.h>
+
+@interface ThreadSafeEventChannelTests : XCTestCase
+@end
+
+@implementation ThreadSafeEventChannelTests {
+  FLTThreadSafeEventChannel *_channel;
+  XCTestExpectation *_mainThreadExpectation;
+}
+
+- (void)setUp {
+  [super setUp];
+  id mockEventChannel = OCMClassMock([FlutterEventChannel class]);
+
+  _mainThreadExpectation =
+      [[XCTestExpectation alloc] initWithDescription:@"invokeMethod must be called in main thread"];
+  _channel = [[FLTThreadSafeEventChannel alloc] initWithEventChannel:mockEventChannel];
+
+  OCMStub([mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
+    if (NSThread.isMainThread) {
+      [self->_mainThreadExpectation fulfill];
+    }
+  });
+}
+
+- (void)testSetStreamHandler_shouldStayOnMainThreadIfCalledFromMainThread {
+  [_channel setStreamHandler:nil];
+  [self waitForExpectations:@[ _mainThreadExpectation ] timeout:1];
+}
+
+- (void)testSetStreamHandler_shouldDispatchToMainThreadIfCalledFromBackgroundThread {
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    [self->_channel setStreamHandler:nil];
+  });
+  [self waitForExpectations:@[ _mainThreadExpectation ] timeout:1];
+}
+
+@end
--- a/packages/camera/camera/example/ios/RunnerTests/ThreadSafeMethodChannelTests.m
+++ b/packages/camera/camera/example/ios/RunnerTests/ThreadSafeMethodChannelTests.m
@@ -1,0 +1,46 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import camera;
+@import XCTest;
+#import <OCMock/OCMock.h>
+
+@interface ThreadSafeMethodChannelTests : XCTestCase
+@end
+
+@implementation ThreadSafeMethodChannelTests {
+  FLTThreadSafeMethodChannel *_channel;
+  XCTestExpectation *_mainThreadExpectation;
+}
+
+- (void)setUp {
+  [super setUp];
+  id mockMethodChannel = OCMClassMock([FlutterMethodChannel class]);
+
+  _mainThreadExpectation =
+      [[XCTestExpectation alloc] initWithDescription:@"invokeMethod must be called in main thread"];
+  _channel = [[FLTThreadSafeMethodChannel alloc] initWithMethodChannel:mockMethodChannel];
+
+  OCMStub([mockMethodChannel invokeMethod:[OCMArg any] arguments:[OCMArg any]])
+      .andDo(^(NSInvocation *invocation) {
+        if (NSThread.isMainThread) {
+          [self->_mainThreadExpectation fulfill];
+        }
+      });
+}
+
+- (void)testInvokeMethod_shouldStayOnMainThreadIfCalledFromMainThread {
+  [_channel invokeMethod:@"foo" arguments:nil];
+
+  [self waitForExpectations:@[ _mainThreadExpectation ] timeout:1];
+}
+
+- (void)testInvokeMethod__shouldDispatchToMainThreadIfCalledFromBackgroundThread {
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    [self->_channel invokeMethod:@"foo" arguments:nil];
+  });
+  [self waitForExpectations:@[ _mainThreadExpectation ] timeout:1];
+}
+
+@end
--- a/packages/camera/camera/ios/Classes/FLTThreadSafeEventChannel.m
+++ b/packages/camera/camera/ios/Classes/FLTThreadSafeEventChannel.m
@@ -1,0 +1,29 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FLTThreadSafeEventChannel.h"
+
+@implementation FLTThreadSafeEventChannel {
+  FlutterEventChannel *_channel;
+}
+
+- (instancetype)initWithEventChannel:(FlutterEventChannel *)channel {
+  self = [super init];
+  if (self) {
+    _channel = channel;
+  }
+  return self;
+}
+
+- (void)setStreamHandler:(NSObject<FlutterStreamHandler> *)handler {
+  if (!NSThread.isMainThread) {
+    dispatch_async(dispatch_get_main_queue(), ^{
+      [self->_channel setStreamHandler:handler];
+    });
+  } else {
+    [_channel setStreamHandler:handler];
+  }
+}
+
+@end
--- a/packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.h
+++ b/packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.h
@@ -1,0 +1,51 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Flutter/Flutter.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Wrapper for FlutterTextureRegistry that always sends events on the main thread
+ */
+@interface FLTThreadSafeTextureRegistry : NSObject
+
+/**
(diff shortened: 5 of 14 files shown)


packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.h
```diff
@@ -0,0 +1,51 @@
...
+ * Creates a FLTThreadSafeTextureRegistry by wrapping an object conforming to
+ * FlutterTextureRegistry.
+ * @param registry The FlutterTextureRegistry object to be wrapped.
+ */
+- (instancetype)initWithTextureRegistry:(NSObject<FlutterTextureRegistry> *)registry;
+
+/**
+ * Registers a `FlutterTexture` for usage in Flutter and returns an id that can be used to reference
+ * that texture when calling into Flutter with channels. Textures must be registered on the
+ * platform thread. On success returns the pointer to the registered texture, else returns 0.
```
cyanglaz, line 24: We should use the word "main thread" here as plugin is not part of engine, "platform thread" might not make much sense to plugin devs or users.

ditto other places

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.m
```diff
@@ -0,0 +1,53 @@
...
+  }
+  return self;
+}
+
+- (int64_t)registerTexture:(NSObject<FlutterTexture> *)texture {
+  if (!NSThread.isMainThread) {
+    __block int64_t textureId;
+    // Use dispatch_sync to keep FlutterTextureRegistry's sychronous API, so that we don't introduce new potential race conditions.
+    // We do not break priority inversion here since it's the background thread waiting for main thread.
+    dispatch_sync(dispatch_get_main_queue(), ^{
```
cyanglaz, line 24: dispatch_sync made this method wait for main thread. It is obvious to the caller.
Maybe we introduce a completion handler and use dispatch_async instead so the wait is more obivous?

jmagman:
```
Failed to stop: Error getting access token for service account: 400 Bad Request
```
You should rebase onto #4592 to pick up the new credentials.

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.m
hellohuanlin: I guess what I wanted to say in my comment is that, if we do not wait for main thread to complete (use `dispatch_async`), we effectively change the semantic of the wrapped function (from a sync API to an async one), which may introduce a potential race condition. But looks like my comment was not clearly phrased.

hellohuanlin: Just in case miscommunication, consider this dummy code: 

```
var _foo = 0
func setFooTo2() {
  _foo = 2
}
func whatever() {
  setFooTo2()
  // since `setFooTo2()` is synchronous API, the subsequent code can safely assume `_foo` is updated
  let bar = 1 / foo
}
```

However, if we create a wrapper and change the API to asynchronous one, we can't make that assumption anymore: 
```
func whatever() {
  setFooTo2Asynchronously() // async API
  // race condition here
  let bar = 1 / foo
}
```

So back to our example, after calling this texture API, there could be a potential code execution assuming that this API is already taken effect (this code execution may or may not be the code immediately following this texture API). Since we have like 10+ registered methods, there could be exponential amount of combination that we need to check for race, which can easily get out of hand. That's why it's important to keep the sync API here, and changing it to async may introduce new hard-to-find races.

cyanglaz: I think your comment is fine. 
 
The race condition could happen on `textureId`. if we use a completion block, the `textureId` would only be available inside the block. So the caller should be aware that the `textureId` is only accessible after the block is executed.

Your point makes sense too, the caller in this plugin always need to wait for the textureId to continue anyway. Maybe we can rename the method to make it clearer to the caller that this method will block the current thread and wait on main thread? Maybe something like: 

```objc
/**
 * Registers a `FlutterTexture` for usage in Flutter and returns an id that can be used to reference
 * that texture when calling into Flutter with channels. Textures must be registered on the
 * platform thread. On success returns the pointer to the registered texture, else returns 0.
 *
 * The registration happens on Main Thread. If the calling thread is not Main Thread, the current thread will be blocked until the registration is successful.
 */
- (int64_t)registerTextureSync:(NSObject<FlutterTexture> *)texture;

```

hellohuanlin: > if we use a completion block, the textureId would only be available inside the block. So the caller should be aware that the textureId is only accessible after the block is executed.

We can use completion block for the code in the same function for sure, but we can't use it for separate functions. (e.g. `dispose` and `create` functions are separate). The issue here is that, it's hard to tell if the developers had made implicit assumption of the synchronous nature when implementing this plugin. If we were to rewrite this plugin from the beginning, I would be fine with either approaches. 

I am cool with the new name as you suggested.

packages/camera/camera/example/ios/RunnerTests/ThreadSafeMethodChannelTests.m
```diff
@@ -0,0 +1,46 @@
...
+}
+
+- (void)testInvokeMethod__shouldDispatchToMainThreadIfCalledFromBackgroundThread {
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    [self->_channel invokeMethod:@"foo" arguments:nil];
+  });
+  [self waitForExpectations:@[ _mainThreadExpectation ] timeout:1];
+}
+
+@end
```
hellohuanlin, line 46: A note to myself: need to check-in the project file change.

packages/camera/camera/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: A Flutter plugin for controlling the camera. Supports previewing
   Dart.
 repository: https://github.com/flutter/plugins/tree/master/packages/camera/camera
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+camera%22
-version: 0.9.4+5
+version: 0.9.4+6
```
hellohuanlin, line 7: Note to myself: should not bump the version in this PR. bump it in next PR.

hellohuanlin: @jmagman I removed this one but it's giving me error saying changelog and yaml versions does not match. Maybe i should not update the changelog in this PR?

packages/camera/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.9.4+6
```
cyanglaz, line 1:
```suggestion
## NEXT
```
If you want to publish this change in a later PR,

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.m
```diff
@@ -0,0 +1,58 @@
...
+
+- (instancetype)initWithTextureRegistry:(NSObject<FlutterTextureRegistry> *)registry {
+  self = [super init];
+  if (self) {
+    _registry = registry;
+  }
+  return self;
+}
+
+- (int64_t)registerTexture:(NSObject<FlutterTexture> *)texture {
```
cyanglaz, line 19: Reminder to rename.

packages/camera/camera/pubspec.yaml
hellohuanlin: synced with @cyanglaz offline that we can use `## NEXT` in the changelog

cyanglaz (approved):
LGTM

stuartmorgan:
> No version change:
> I want to update the pubspec.yaml file in next PR so that we can batch the change.

I know it's too late now, but I only just saw this: Why? We explicitly don't want to batch changes in general (see the linked docs), and you also have no control in general over what it will be batched with—there's no guarantee that the next `camera` change to land will be the one you have in mind.

What was the advantage of doing this?

jmagman:
I hadn't reviewed this yet!

jmagman:
@hellohuanlin let's revert this at https://github.com/flutter/plugins/pull/4629 until Stuart and I can review it.

hellohuanlin:
> > No version change:
> > I want to update the pubspec.yaml file in next PR so that we can batch the change.
> 
> I know it's too late now, but I only just saw this: Why? We explicitly don't want to batch changes in general (see the linked docs), and you also have no control in general over what it will be batched with—there's no guarantee that the next `camera` change to land will be the one you have in mind.
> 
> What was the advantage of doing this?

@stuartmorgan I wanted to combine the 2 fixes because I wanted to avoid too frequent publishes, so that developers don't have to bump the version twice. But from your comment it seems this is fine. Then I am ok with bumping the version.

hellohuanlin:
> I hadn't reviewed this yet!

@jmagman Ah sorry. I did not intend to leave you out. I thought I just needed 1 stamp iirc. 

The land button turned green though, maybe some bug with the CI landing requirement check.

jmagman:
> Ah sorry. I did not intend to leave you out. I thought I just needed 1 stamp iirc. The land button turned green though, maybe some bug with the CI landing requirement check?

By policy you generally do only need 1 LGTM from a flutter-hacker, I had just mentioned that I would get a chance to review on Monday. 🙂   You should probably add @stuartmorgan as a reviewer to non-trivial plugin changes, particularly when he has a lot of context for particular patterns (or failures to fix those patterns, in this case).

jmagman (commented):
Adding stylistic nits since I know you're putting this out for review again.

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.m
jmagman: I understand with the `create` `dispose` example the data should be sent before `create` returns, but the caller in `handleMethodCallAsync` can make it synchronous in that case and wait to return in the completion block, instead of enforcing its synchronicity in the `-registerTexture:` API.

packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
```diff
@@ -0,0 +1,44 @@
...
+@import camera;
+@import XCTest;
+#import <OCMock/OCMock.h>
+
+@interface ThreadSafeEventChannelTests : XCTestCase
+@end
+
+@implementation ThreadSafeEventChannelTests {
+  FLTThreadSafeEventChannel *_channel;
+  XCTestExpectation *_mainThreadExpectation;
```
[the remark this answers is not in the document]
hellohuanlin: Very interesting. My preference was the opposite (use ivar unless I need extra functionalities in property). But not a strong preference so I can change it to properties.

packages/camera/camera/example/ios/RunnerTests/ThreadSafeMethodChannelTests.m
```diff
@@ -0,0 +1,46 @@
...
+- (void)testInvokeMethod_shouldStayOnMainThreadIfCalledFromMainThread {
+  [_channel invokeMethod:@"foo" arguments:nil];
+
+  [self waitForExpectations:@[ _mainThreadExpectation ] timeout:1];
+}
+
+- (void)testInvokeMethod__shouldDispatchToMainThreadIfCalledFromBackgroundThread {
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    [self->_channel invokeMethod:@"foo" arguments:nil];
+  });
```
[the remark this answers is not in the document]
hellohuanlin: I am cool with either, though are we concerned about inconsistency in the same plugin?

packages/camera/camera/ios/Classes/FLTThreadSafeEventChannel.m
```diff
@@ -0,0 +1,29 @@
...
+}
+
+- (void)setStreamHandler:(NSObject<FlutterStreamHandler> *)handler {
+  if (!NSThread.isMainThread) {
+    dispatch_async(dispatch_get_main_queue(), ^{
+      [self->_channel setStreamHandler:handler];
+    });
+  } else {
+    [_channel setStreamHandler:handler];
+  }
```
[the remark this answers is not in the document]
hellohuanlin: Good call

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.h
```diff
@@ -0,0 +1,51 @@
...
+// found in the LICENSE file.
+
+#import <Flutter/Flutter.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Wrapper for FlutterTextureRegistry that always sends events on the main thread
+ */
+@interface FLTThreadSafeTextureRegistry : NSObject
```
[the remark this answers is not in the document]
hellohuanlin: Yes I was following the convention for consistency. But I can change both if you prefer.

stuartmorgan:
> so that developers don't have to bump the version twice

Developers don't have to bump the version twice; they can update it however often they want to, regardless of how often we release. Releasing more often gives people more options, not fewer. (Also, manually updating versions of individual is not generally how most projects get new bugfix versions of plugins anyway, because of the way the constraint system works.)

PR #4615, opened 2021-12-14
Update firebase test key

keyonghan:
The old cirrus GCP key expired. We have updated the key in https://github.com/flutter/plugins/pull/4592, but the firebase test lab also needs an update.


godofredoc (approved):

stuartmorgan (approved):
LGTM

stuartmorgan:
Landing on red to fix redness.

PR #4616, opened 2021-12-15
update firebase key

keyonghan:
Follow up of https://github.com/flutter/plugins/pull/4615.

Task level key should not be the same as the org level GCP key: https://cirrus-ci.org/guide/writing-tasks/#encrypted-variables.


godofredoc (approved):

keyonghan:
Manually land to unblock the tree.

PR #4617, opened 2021-12-15
[path_provider] Switch Android to an internal method channel

guille-ideotec:
Eliminates the path_provider_android reliance on the default, shared method channel implementation, in favor of an in-package implementation.

Part of https://github.com/flutter/flutter/issues/94224

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/path_provider/path_provider_android/lib/path_provider_android.dart
+++ b/packages/path_provider/path_provider_android/lib/path_provider_android.dart
@@ -30,7 +30,7 @@ class PathProviderAndroid extends PathProviderPlatform {
 
   @override
   Future<String?> getLibraryPath() {
-    return methodChannel.invokeMethod<String>('getLibraryDirectory');
+    throw UnsupportedError('getLibraryPath is not supported on Android');
   }
 
   @override
@@ -41,25 +41,25 @@ class PathProviderAndroid extends PathProviderPlatform {
 
   @override
   Future<String?> getExternalStoragePath() async {
-    throw UnsupportedError(
-        'getExternalStoragePath is not supported on Android');
+    return methodChannel.invokeMethod<String>('getStorageDirectory');
   }
 
   @override
   Future<List<String>?> getExternalCachePaths() async {
-    throw UnsupportedError('getExternalCachePaths is not supported on Android');
+    return methodChannel
+        .invokeMethod<List<String>>('getExternalCacheDirectories');
   }
 
   @override
   Future<List<String>?> getExternalStoragePaths({
     StorageDirectory? type,
   }) async {
-    throw UnsupportedError(
-        'getExternalStoragePaths is not supported on Android');
+    return methodChannel
+        .invokeMethod<List<String>>('getExternalStorageDirectories');
   }
 
   @override
   Future<String?> getDownloadsPath() {
-    return methodChannel.invokeMethod<String>('getDownloadsDirectory');
+    throw UnsupportedError('getDownloadsPath is not supported on Android');
   }
 }
--- a/packages/path_provider/path_provider_android/lib/path_provider_android.dart
+++ b/packages/path_provider/path_provider_android/lib/path_provider_android.dart
@@ -1,0 +1,65 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/services.dart';
+import 'package:meta/meta.dart';
+import 'package:path_provider_platform_interface/path_provider_platform_interface.dart';
+
+/// The Android implementation of [PathProviderPlatform].
+class PathProviderAndroid extends PathProviderPlatform {
+  /// The method channel used to interact with the native platform.
+  @visibleForTesting
+  MethodChannel methodChannel =
+      const MethodChannel('plugins.flutter.io/path_provider_android');
+
+  /// Registers this class as the default instance of [PathProviderPlatform]
+  static void registerWith() {
+    PathProviderPlatform.instance = PathProviderAndroid();
+  }
+
+  @override
+  Future<String?> getTemporaryPath() {
+    return methodChannel.invokeMethod<String>('getTemporaryDirectory');
+  }
+
+  @override
+  Future<String?> getApplicationSupportPath() {
+    return methodChannel.invokeMethod<String>('getApplicationSupportDirectory');
+  }
+
+  @override
+  Future<String?> getLibraryPath() {
+    return methodChannel.invokeMethod<String>('getLibraryDirectory');
+  }
+
+  @override
+  Future<String?> getApplicationDocumentsPath() {
+    return methodChannel
+        .invokeMethod<String>('getApplicationDocumentsDirectory');
+  }
+
+  @override
+  Future<String?> getExternalStoragePath() async {
+    throw UnsupportedError(
+        'getExternalStoragePath is not supported on Android');
+  }
+
+  @override
+  Future<List<String>?> getExternalCachePaths() async {
+    throw UnsupportedError('getExternalCachePaths is not supported on Android');
+  }
+
+  @override
+  Future<List<String>?> getExternalStoragePaths({
+    StorageDirectory? type,
+  }) async {
+    throw UnsupportedError(
+        'getExternalStoragePaths is not supported on Android');
+  }
+
+  @override
+  Future<String?> getDownloadsPath() {
+    return methodChannel.invokeMethod<String>('getDownloadsDirectory');
+  }
+}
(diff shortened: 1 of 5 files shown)


packages/path_provider/path_provider_android/lib/path_provider_android.dart
```diff
@@ -0,0 +1,65 @@
...
+
+  @override
+  Future<String?> getExternalStoragePath() async {
+    return methodChannel.invokeMethod<String>('getStorageDirectory');
+  }
+
+  @override
+  Future<List<String>?> getExternalCachePaths() async {
+    return methodChannel
+        .invokeMethod<List<String>>('getExternalCacheDirectories');
```
stuartmorgan, line 50: Why did you change the implementations of these list methods? The change isn't correct, which is why the unit tests are failing.

packages/path_provider/path_provider_android/test/path_provider_android_test.dart
```diff
@@ -0,0 +1,119 @@
...
+        log,
+        <Matcher>[
+          isMethodCall('getExternalStorageDirectories', arguments: null)
+        ],
+      );
+      expect(result, kExternalStoragePaths);
+    });
+
+    test('getLibraryPath throws', () async {
+      expect(pathProvider.getLibraryPath(), throwsA(isUnsupportedError));
```
stuartmorgan, line 112: These are failing; you either need to make the functions async, or use `() => pathProvider.foo()` in the `expect`.

packages/path_provider/path_provider_android/lib/path_provider_android.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:path_provider_platform_interface/path_provider_platform_interface.dart';
+
+/// The Android implementation of [PathProviderPlatform].
+class PathProviderAndroid extends PathProviderPlatform {
+  /// The method channel used to interact with the native platform.
+  @visibleForTesting
+  MethodChannel methodChannel =
+      const MethodChannel('plugins.flutter.io/path_provider_android');
+
+  /// Registers this class as the default instance of [PathProviderPlatform]
```
stuartmorgan, line 16: Missing `.` at the end.

(I see this is a copy-paste, but we shouldn't perpetuate the typo.)

packages/path_provider/path_provider_android/lib/path_provider_android.dart
```diff
@@ -0,0 +1,65 @@
...
+    return methodChannel
+        .invokeMethod<List<String>>('getExternalCacheDirectories');
+  }
+
+  @override
+  Future<List<String>?> getExternalStoragePaths({
+    StorageDirectory? type,
+  }) async {
+    return methodChannel
+        .invokeMethod<List<String>>('getExternalStorageDirectories');
```
stuartmorgan, line 58: Here you not only changed the type, but you discarded the `type`.

The calls should be identical to the versions in the original method channel. I get the impression that you copied the macOS version of this file and then wrote the parts it was missing for Android from scratch; I would strongly recommend instead copying the shared method channel implementation, and then removing the platform checks (which is how I made the macOS version).

stuartmorgan (changes requested):
Thanks! Some issues below, but this is a great start.

packages/path_provider/path_provider_android/test/path_provider_android_test.dart
```diff
@@ -0,0 +1,147 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:path_provider_android/path_provider_android.dart';
+import 'package:path_provider_platform_interface/src/enums.dart';
```
guille-ideotec, line 8: Not too sure about importing path_provider_platform_interface/src/enums.dart to reference StorageDirectory. Since this enum is only applicable to Android, should I move it into the path_provider_android package?

packages/path_provider/path_provider_android/lib/path_provider_android.dart
guille-ideotec: You're absolutely right: I blindly copied the macOS version and modified it just enough so that it looked right for Android. Now that I know how you did the macOS version I realize what I was supposed to do, and so I copied from the original method channel instead of your macOS version.

I'm sorry for such shoddy work, the truth is I didn't really invest the time to understand what I was doing because I wanted to get something out there as soon as possible to force myself to commit to it (if not I would have just forgotten about it and ended up never doing it).

Thanks for your feedback! Let me know if there are still any issues (I left a comment on something I think might have room for improvement).

packages/path_provider/path_provider_android/test/path_provider_android_test.dart
stuartmorgan: > Not too sure about importing path_provider_platform_interface/src/enums.dart to reference StorageDirectory.

It's re-exported from the main package header (`package:path_provider_platform_interface.dart`), which is what you should include instead.

> Since this enum is only applicable to Android, should I move it into the path_provider_android package?

You can't, because the platform interface uses it. The API can't use a type that's not defined anywhere.

If path_provider were being written from scratch as a federated plugin, the Android-specific API would be an extension that only exited on the Android implementation, in which case the enum could move, but moving it now would be a breaking change that's not worth the effect that would have on the ecosystem.

guille-ideotec: Got it. I've changed it to import from the main package header. Any other issues?

stuartmorgan (approved):
LGTM

@bparrishMines for second review per policy.

packages/path_provider/path_provider_android/pubspec.yaml
```diff
@@ -15,11 +15,13 @@ flutter:
       android:
         package: io.flutter.plugins.pathprovider
         pluginClass: PathProviderPlugin
+        dartPluginClass: PathProviderAndroid
 
 dependencies:
   flutter:
     sdk: flutter
-  path_provider_platform_interface: ^2.0.0
+  meta: ^1.3.0
```
bparrishMines, line 23: `meta` shouldn't be necessary if you are just using it for `@visibleForTesting`. You should be able to just import `'package:flutter/foundation.dart'`. I believe the Flutter SDK provides a version of `meta` that we don't want to have version conflicts with later.

bparrishMines (changes requested):

bparrishMines (approved):
LGTM

packages/path_provider/path_provider_android/pubspec.yaml
stuartmorgan: TIL! We import `meta` in a bunch of plugins, probably mostly for this. We should do a pass to clean that up where possible.

stuartmorgan: Filed https://github.com/flutter/flutter/issues/95658

packages/path_provider/path_provider_android/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: path_provider_android
 description: Android implementation of the path_provider plugin.
 repository: https://github.com/flutter/plugins/tree/master/packages/path_provider/path_provider_android
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+path_provider%22
-version: 2.0.9
+version: 2.0.10
 
 environment:
   sdk: ">=2.14.0 <3.0.0"
```
stuartmorgan, line 8: I just realized that I forgot we needed to bump the Flutter SDK requirement to 2.8 as part of this since it uses Dart plugin registration :(

I don't think it's really fixable unfortunately, unless we:
- revert + publish as 2.0.11
- reland with the SDK bump + publish as 2.0.12

because if we just bump the SDK requirement this incorrect version will be the last one 2.5 users will resolve to. We can wait and see if there's non-trivial fallout from this first. (If it's only a small number of people, we can tell them to temporarily pin a path_provider_android dependency.)

stuartmorgan: We're getting a fair number of reports of this already (https://github.com/flutter/flutter/issues/95706), so we'll need to fix this.

Instead of a full revert, I think we could temporarily change the channel name back since nothing about the channel has changed yet, then change it back with an SDK bump.

guille-ideotec: I created a new PR reverting the channel name. I would need to know if it requires a version bump (with corresponding CHANGELOG.md line) and I guess I also need an exception for new tests.

guille-ideotec: https://github.com/flutter/plugins/pull/4636

PR #4619, opened 2021-12-16
[camera]fix crash due to race condition in dispatch queue

hellohuanlin:
This PR fixed a crash due to race condition of `_captureSessionQueue` ivar. This kind of race condition is annoying and hard to fix because it takes tons of luck to reproduce (I accidentally touched iPhone's home bar, causing the app to become active and then immediately inactive, forming a race condition). 

**Crash**
```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[AVCaptureAudioDataOutput setSampleBufferDelegate:queue:] NULL queue passed'
terminating with uncaught exception of type NSException
```

**Reproduce**
(1) In sample app, start video recording then stop it. 
(2) Swipe home bar just a bit on iPhone X, so the app becomes inactive and immediately regain active. 
(3) Tap the video button again and observe this crash. 

**Root Cause**
`_captureSessionQueue` is checked in main thread, but set to nil in background thread, which is open to race condition: 
```
  // main thread: 
  if (_captureSessionQueue == nil) {
    _captureSessionQueue = dispatch_queue_create("io.flutter.camera.captureSessionQueue", NULL);
  } 

  // background thread 
 _captureSessionQueue = nil
```

**Explain**
In the above reproduce step, (1) is to assign a new `_captureSessionQueue`, (2) is to trigger "dispose" and "create" method calls consecutively, resulting in a race condition, and (3) will pass in the `_captureSessionQueue` which has been nil'ed out asynchronously by the `dispose` call in previous step. 

**Solution**
(1) It looks like the camera plugin has pretty complicated threading logic and it's getting hard to maintain. There have been multiple related issues. If we get time I am interested in refactoring the threading model in the future. 
(2) For now an easy fix to this race condition is not to nil out this `_captureSessionQueue`. DispatchQueues are quite lightweight as they are not actual threads. 
(3) Or we can dispatch to main thread again before setting nil, which should fix `_captureSessionQueue` race for this particular case, but may still open for other unknown cases

**Issues**
https://github.com/flutter/flutter/issues/59124
https://github.com/flutter/flutter/issues/52578#issuecomment-797688303

# Next Step

* This section is outdated. We wrote up [a formal proposal](https://docs.google.com/document/d/1UtKfFoHrQEr9Ehsx9oPGLqjcPT9IG3KjX0DSStkqHwU/edit#) to address various threading related issues. 

This PR should fix the crash, but we should start looking at simplifying the threading logic in this plugin. Ideally, a `DispatchQueue` should have its lifecycle scoped to the resource that the queue manages, instead of manually nil'ing out the reference. HOWEVER, let's step back a bit: **should this even be a plugin problem?** 

If we trace the execution from UI to plugin and back to UI, we get these thread hops: 
```
Flutter's UI thread ->
iOS platform thread ->
camera background thread -> 
iOS platform thread -> 
Flutter's UI thread
```
The iOS platform thread is involved twice, just to dispatch to another thread, which means the whole process could have been as simple as: 
```
Flutter's UI thread ->
camera background thread -> 
Flutter's UI thread
```

I think engine should support invoking plugin methods in background. The benefits are: 
- Avoid unnecessary thread hops, obviously;
- Plugin developers (e.g. camera) don't have to work on threading problems (if they simply want to run code in background thread). 
- Avoid crashes like this one. 

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/ios/Classes/CameraPlugin.m
+++ b/packages/camera/camera/ios/Classes/CameraPlugin.m
@@ -1322,9 +1322,7 @@ NSString *const errorMethod = @"error";
 @property(readonly, nonatomic) FLTThreadSafeMethodChannel *deviceEventMethodChannel;
 @end
 
-@implementation CameraPlugin {
-  dispatch_queue_t _captureSessionQueue;
-}
+@implementation CameraPlugin
 
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
   FlutterMethodChannel *channel =
@@ -1341,6 +1339,7 @@ NSString *const errorMethod = @"error";
   NSAssert(self, @"super init cannot be nil");
   _registry = [[FLTThreadSafeTextureRegistry alloc] initWithTextureRegistry:registry];
   _messenger = messenger;
+  _captureSessionQueue = dispatch_queue_create("io.flutter.camera.captureSessionQueue", NULL);
   [self initDeviceEventMethodChannel];
   [self startOrientationListener];
   return self;
@@ -1385,10 +1384,6 @@ NSString *const errorMethod = @"error";
 }
 
 - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result {
-  if (_captureSessionQueue == nil) {
-    _captureSessionQueue = dispatch_queue_create("io.flutter.camera.dispatchqueue", NULL);
-  }
-
   // Invoke the plugin on another dispatch queue to avoid blocking the UI.
   dispatch_async(_captureSessionQueue, ^{
     FLTThreadSafeFlutterResult *threadSafeResult =
@@ -1507,7 +1502,6 @@ NSString *const errorMethod = @"error";
     } else if ([@"dispose" isEqualToString:call.method]) {
       [_registry unregisterTexture:cameraId];
       [_camera close];
-      _captureSessionQueue = nil;
       [result sendSuccess];
     } else if ([@"prepareForVideoRecording" isEqualToString:call.method]) {
       [_camera setUpCaptureSessionForAudio];
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,6 +1,7 @@
-## NEXT
+## 0.9.4+7
 
-* Minor internal code cleanup.
+* Fixes a crash in iOS when passing null queue pointer into AVFoundation API due to race condition.  
+* Minor iOS internal code cleanup.
 
 ## 0.9.4+6
 
--- a/packages/camera/camera/example/ios/Runner.xcodeproj/project.pbxproj
+++ b/packages/camera/camera/example/ios/Runner.xcodeproj/project.pbxproj
@@ -21,6 +21,7 @@
 		97C146FE1CF9000F007C117D /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 97C146FD1CF9000F007C117D /* Assets.xcassets */; };
 		97C147011CF9000F007C117D /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 97C146FF1CF9000F007C117D /* LaunchScreen.storyboard */; };
 		E01EE4A82799F3A5008C1950 /* QueueHelperTests.m in Sources */ = {isa = PBXBuildFile; fileRef = E01EE4A72799F3A5008C1950 /* QueueHelperTests.m */; };
+		E032F250279F5E94009E9028 /* CameraCaptureSessionQueueRaceConditionTests.m in Sources */ = {isa = PBXBuildFile; fileRef = E032F24F279F5E94009E9028 /* CameraCaptureSessionQueueRaceConditionTests.m */; };
 		E0C6E2002770F01A00EA6AA3 /* ThreadSafeMethodChannelTests.m in Sources */ = {isa = PBXBuildFile; fileRef = E0C6E1FD2770F01A00EA6AA3 /* ThreadSafeMethodChannelTests.m */; };
 		E0C6E2012770F01A00EA6AA3 /* ThreadSafeTextureRegistryTests.m in Sources */ = {isa = PBXBuildFile; fileRef = E0C6E1FE2770F01A00EA6AA3 /* ThreadSafeTextureRegistryTests.m */; };
 		E0C6E2022770F01A00EA6AA3 /* ThreadSafeEventChannelTests.m in Sources */ = {isa = PBXBuildFile; fileRef = E0C6E1FF2770F01A00EA6AA3 /* ThreadSafeEventChannelTests.m */; };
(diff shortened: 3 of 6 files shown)


stuartmorgan:
> (1) It looks like the camera plugin has pretty complicated threading logic and it's getting hard to maintain. There have been multiple related issues.

Yes, this is if anything an understatement. It was until recently *very* wrong, and some small steps to make it somewhat less wrong were taken, but it needs to be overhauled.

packages/camera/camera/ios/Classes/CameraPlugin.h
```diff
@@ -5,4 +5,5 @@
 #import <Flutter/Flutter.h>
 
 @interface CameraPlugin : NSObject <FlutterPlugin>
+@property(nonatomic, strong, nullable) dispatch_queue_t dispatchQueue;
```
stuartmorgan, line 8: If this is only public for tests, please use a `_Test.h` header instead.

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -1470,7 +1468,6 @@ - (void)handleMethodCallAsync:(FlutterMethodCall *)call
     } else if ([@"dispose" isEqualToString:call.method]) {
       [_registry unregisterTexture:cameraId];
       [_camera close];
-      _dispatchQueue = nil;
```
stuartmorgan, line 1473: If you've never going to nil it out, then it can be created in `init` instead of lazily as it currently is.

packages/camera/camera/example/ios/RunnerTests/CameraDispatchQueueRaceConditionTests.m
```diff
@@ -0,0 +1,41 @@
+//
+//  CameraDispatchQueueRaceConditionTests.m
+//  RunnerTests
+//
+//  Created by Huan Lin on 12/16/21.
+//  Copyright © 2021 The Flutter Authors. All rights reserved.
+//
```
stuartmorgan, line 7: This needs to be fixed.

packages/camera/camera/example/ios/RunnerTests/CameraDispatchQueueRaceConditionTests.m
```diff
@@ -0,0 +1,41 @@
...
+                      [disposeExpectation fulfill];
+                    }];
+  [camera handleMethodCall:createCall
+                    result:^(id _Nullable result){
+                    }];
+  [self waitForExpectations:@[ disposeExpectation ] timeout:1];
+  XCTAssertNotNil(camera.dispatchQueue, @"dispatchQueue must not be nil after create method. ");
+}
+
+@end
```
stuartmorgan, line 41: You've added a new test file, but not added it to the project, so it's not actually going to be run.

stuartmorgan (changes requested):
Could you merge in the latest version of master so that the tests run? A key recently expired, which I think is why everything is red.

hellohuanlin:
@stuartmorgan thanks for the review. I think we are having code freeze until next year, so I will fix them next year when we come back.

packages/camera/camera/example/ios/RunnerTests/CameraDispatchQueueRaceConditionTests.m
```diff
@@ -0,0 +1,41 @@
+//
+//  CameraDispatchQueueRaceConditionTests.m
+//  RunnerTests
+//
+//  Created by Huan Lin on 12/16/21.
+//  Copyright © 2021 The Flutter Authors. All rights reserved.
+//
+
```
cyanglaz, line 8: license

packages/camera/camera/example/ios/RunnerTests/CameraDispatchQueueRaceConditionTests.m
```diff
@@ -0,0 +1,41 @@
...
+                     arguments:@{@"resolutionPreset" : @"medium", @"enableAudio" : @(1)}];
+  // Mimic a dispose call followed by a create call, which can be triggered by slightly dragging the
+  // home bar, causing the app to be inactive, and immediately regain active.
+  [camera handleMethodCall:disposeCall
+                    result:^(id _Nullable result) {
+                      [disposeExpectation fulfill];
+                    }];
+  [camera handleMethodCall:createCall
+                    result:^(id _Nullable result){
+                    }];
```
cyanglaz, line 36: The old code might also pass this test (since it was a race condition), right?
Maybe we don't even need to call `createCall` here and make the test more like:
"dispatchQueue should not be nil after dispose"

stuartmorgan:
> I think we are having code freeze until next year

There's no code freeze for this repository (although people should make sure not to land publish-triggering changes unless they will be around to deal with any fallout, as is generally the case here).

packages/camera/camera/ios/Classes/CameraPlugin.m
cyanglaz, line 1473: nits:
Now that we don't nullify `_dispatchQueue`, `_dispatchQueue`'s creation can be moved to the init method of the plugin to better indicate its lifecycle.

packages/camera/camera/ios/Classes/CameraPlugin.h
cyanglaz, line 8: Make this readonly and comment it is for testing only.

cyanglaz (changes requested):
Thank for the fix! Left some comments.

hellohuanlin:
@stuartmorgan gotcha. thanks for the clarification. I think I will wait till https://github.com/flutter/plugins/pull/4608 is landed, because I want to bump version and changelog in both PRs.

packages/camera/camera/ios/Classes/CameraPlugin.h
cyanglaz: actually, do this instead : https://github.com/flutter/plugins/pull/4619#discussion_r770928353

packages/camera/camera/ios/Classes/CameraPlugin.m
cyanglaz: Oops, my page is out of sync. Didn't see Stuart's comment when I was reviewing. I'm going to close this comment as it says the same thing Stuart has said.

packages/camera/camera/example/ios/RunnerTests/CameraDispatchQueueRaceConditionTests.m
hellohuanlin: The old code fails the test. I added `createCall` immediately following `dispose` to illustrate the race, otherwise a single `dispose` was ok to nil out the reference.

cyanglaz (approved):
LGTM!

stuartmorgan:
Is this obsoleted by your other changes, or should it be rebased and cleaned up for landing?

hellohuanlin:
> Is this obsoleted by your other changes, or should it be rebased and cleaned up for landing?

will rebase and clean up

hellohuanlin:
@stuartmorgan updated. PTAL. Thanks.

jmagman (approved):
LGTM.  Can you update this PR's description since this doesn't touch anything named `_dispatchQueue`?

stuartmorgan (approved):

PR #4623, opened 2021-12-17
[video_player] Support for simultaneous playback of audio sources on iOS

mitghi:
This PR adds support for  simultaneous playback of audio sources such as background music
along with video and other sources in iOS.

The _mixWithOthers_ option works as expected, as long as the video player is the **single**
audio source within the app. Therefore, the _mixWithOthers_ option alone, does not 
provide the flexibility of using the video player along with background music and other
audio sources within the app ( e.g. muted video + text-to-speech + background music ).

The proposed solution is to pass another argument to **VideoPlayerOptions** to setup ambient mode:
```dart
VideoPlayerOptions(mixWithOthers: true, ambient: true)
```

_Ambient_ mode allows mixing of audio sources within app with background sources.

This PR fixes issue [#94328](https://github.com/flutter/flutter/issues/94328).

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Thanks for the contribution! This will need native unit tests testing the actual change in functionality; currently the only test coverage is that the option is passed through the method channel, not that it's used correctly.

Also, you've indicated that you've signed the CLA, but the check indicates that neither of the email addresses associated with the commits have CLA entries, so that will need to be resolved before we could review this.

mitghi:
@stuartmorgan Thank you for the feedback. I will write the platform test cases and resubmit the PR. I have corrected the email address as well, but should resubmit the PR. Previously, the wrong email address was associated with git.

I have a question, this PR changes the platform interface slightly in order to add additional argument to _VideoPlayerOptions_. Should the changes to platform interface be separately submitted first? Because otherwise the video_player depend on a version of video_player_platform_interface that does not exist in repo.

stuartmorgan:
See https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins

stuartmorgan:
> but should resubmit the PR

You're welcome to do a new PR if you want to, but you could also just `amend` the commits locally to have the right email address, and then force-push to this PR.

PR #4626, opened 2021-12-20
[webview_flutter_wkwebview] Start work on new ios implementation

bparrishMines:
Start of iOS side of: https://github.com/flutter/flutter/issues/93732

Starts the Dart side of the platform specific implemention for iOS. Created `WebViewCupertinoWidget`and added support for `WebSettings.allowsInlineMediaPlayback` and `AutoMediaPlaybackPolicy`. This currently has no effect on the working iOS implementation.

The branch that contained all the changes for Dart came out to about `9300` lines of code. This is the first PR to add all the api piece by piece.

No version change:
This is only the initial commit to https://github.com/flutter/flutter/issues/93732 and doesn't make any changes to the current implementation.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
```diff
@@ -0,0 +1,227 @@
...
+    switch (autoMediaPlaybackPolicy) {
+      case AutoMediaPlaybackPolicy.require_user_action_for_all_media_types:
+        requiresUserAction = true;
+        break;
+      case AutoMediaPlaybackPolicy.always_allow:
+        requiresUserAction = false;
+        break;
+    }
+
+    if (_isGreaterOrEqualVersion(await systemVersion, '10.0')) {
```
bparrishMines, line 180: An alternative to having this version check logic would be to only support `mediaTypesRequiringUserActionForPlayback` and handle calling the other methods with platform code. However, I wanted to try a solution of handling version conflicts in Dart. A future feature may require a more complex solution than a 1 liner like this one.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -0,0 +1,151 @@
...
+/// * Custom scripts to inject into the webpage
+/// * Custom rules that determine how to render content
+///
+/// You create a [WebViewConfiguration] object in your code, configure its
+/// properties, and pass it to the initializer of your [WebView] object. The web
+/// view incorporates your configuration settings only at creation time; you
+/// cannot change those settings dynamically later.
+class WebViewConfiguration {
+  /// Indicates whether HTML5 videos play inline or use the native full-screen controller.
+  set allowsInlineMediaPlayback(bool allow) {
```
bparrishMines, line 48: Relevant code from current implementation https://github.com/flutter/plugins/blob/master/packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m#L548

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -0,0 +1,151 @@
...
+    throw UnimplementedError();
+  }
+
+  /// The media types that require a user gesture to begin playing.
+  ///
+  /// Use [AudiovisualMediaType.none] to indicate that no user gestures are
+  /// required to begin playing media.
+  set mediaTypesRequiringUserActionForPlayback(
```
bparrishMines, line 56: Relevant code from current implementation: https://github.com/flutter/plugins/blob/master/packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m#L568

packages/webview_flutter/webview_flutter_wkwebview/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## NEXT
+
+* Starts the Dart side of the platform specific implemention for iOS. Created
+  `WebViewCupertinoWidget`and added support for `WebSettings.allowsInlineMediaPlayback` and
+  `AutoMediaPlaybackPolicy`. This currently has no effect on the working iOS implementation.
```
stuartmorgan, line 5: I think we shouldn't put this in the CHANGELOG. If we end up with bugfix releases to this package interspersed with the work I think it would make for a very confusing changelog. Let's do the PR description exemption for all of these until it's actually turned on.

This would all be good detail to add to the PR description though.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -0,0 +1,151 @@
...
+    assert(types.isNotEmpty);
+    throw UnimplementedError();
+  }
+
+  /// Indicates whether HTML5 videos require the user to start playing them.
+  ///
+  /// Deprecated for iOS >= 10.0.
+  ///
+  /// Passing false indicates the videos can play automatically.
+  @Deprecated(
```
stuartmorgan, line 68: Isn't this code going to be entirely internal to the package? We shouldn't deprecate our internal wrappers when we know we'll be using them.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -0,0 +1,151 @@
...
+    'Deprecated for iOS 10.0+. '
+    'Please use WebViewConfiguration.mediaTypesRequiringUserActionForPlayback.',
+  )
+  set requiresUserActionForMediaPlayback(bool required) {
+    throw UnimplementedError();
+  }
+
+  /// Deprecated for iOS >= 9.0.
```
stuartmorgan, line 76: That's all we support now, so this is dead code. We just never cleaned up the implementation after changing the min version (I guess we missed it in the scrub).

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -0,0 +1,151 @@
...
+/// extends the [NavigationDelegate] class. Use your navigation delegate to
+/// modify the web view’s navigation behavior, or to track the loading progress
+/// of new content.
+///
+/// You can also use the methods of [WebView] to navigate programmatically
+/// through your content, or to trigger navigation from other parts of your
+/// app’s interface. For example, if your UI includes forward and back buttons,
+/// connect those buttons to the [goBack] and [goForward] methods of your web
+/// view to trigger the corresponding web navigation. Use the [canGoBack] and
+/// [canGoForward] fields to determine when to enable or disable your buttons.
```
stuartmorgan, line 135: We should not be duplicating docs wholesale from Apple code.

I would much prefer we say something like:
"Wraps WKWebView. See insert-link-here for documentation".

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
```diff
@@ -0,0 +1,227 @@
...
+  final JavascriptChannelRegistry javascriptChannelRegistry;
+
+  /// A collection of properties that you use to initialize a web view.
+  ///
+  /// If null, a default configuration is used.
+  final web_kit.WebViewConfiguration? configuration;
+
+  /// The system version of the iOS device.
+  ///
+  /// If null, this value is retrieved asynchronously.
```
stuartmorgan, line 45: This documentation seems confusing; AFAICT is is describing what clients of this code might do, not anything this class does.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
```diff
@@ -0,0 +1,227 @@
...
+  final FutureOr<String>? systemVersion;
+
+  /// The handler for constructing [web_kit.WebView]s.
```
stuartmorgan, line 48: If it's for creating webviews, isn't it a `Factory` rather than a `Proxy`?

packages/webview_flutter/webview_flutter_wkwebview/pubspec.yaml
```diff
@@ -16,13 +16,17 @@ flutter:
         pluginClass: FLTWebViewFlutterPlugin
 
 dependencies:
+  device_info_plus: ^3.1.1
```
stuartmorgan, line 19: I definitely don't want to do this. We should build the specific functionality we need into the plugin.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
```diff
@@ -0,0 +1,227 @@
...
+    } else {
+      configuration.mediaPlaybackRequiresUserAction = requiresUserAction;
+    }
+  }
+
+  bool _isGreaterOrEqualVersion(String first, String second) {
+    return _parseVersion(first).compareTo(_parseVersion(second)) >= 0;
+  }
+
+  Version _parseVersion(String versionStr) {
```
stuartmorgan, line 197: `pub_semver` is overkill here; you're already writing manual parsing code anyway, and the set of things it needs to support is smaller than `pub_semver` (there's no `iOS 10.0.7+3-dev` for instance). You can eliminate the dependency and just parse to a three-element array then compare in a loop.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
```diff
@@ -0,0 +1,227 @@
...
+        break;
+    }
+
+    if (_isGreaterOrEqualVersion(await systemVersion, '10.0')) {
+      configuration.mediaTypesRequiringUserActionForPlayback =
+          <web_kit.AudiovisualMediaType>{
+        if (requiresUserAction) web_kit.AudiovisualMediaType.all,
+        if (!requiresUserAction) web_kit.AudiovisualMediaType.none
+      };
+    } else if (_isGreaterOrEqualVersion(await systemVersion, '9.0')) {
```
stuartmorgan, line 186: This is always true, so should just be the `else`.

stuartmorgan (changes requested):

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
bparrishMines: I was back and forth about to what extent we should document these classes. I think this is a good pattern moving forward. I think this would also pair well with testing that links are up to date. I created an issue for it: https://github.com/flutter/flutter/issues/96974https://github.com/flutter/flutter/issues/96974

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
bparrishMines: On Android, this class is also used to call static methods. So there is a possibility that it could be used for more than constructing `WebView`s. I updated the documentation.

bparrishMines:
@stuartmorgan This is ready for another review. I decided to remove the logic for checking the version since it can be easily handled with platform code: https://github.com/flutter/plugins/blob/main/packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m#L568

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -0,0 +1,71 @@
...
+    assert(types.isNotEmpty);
+    throw UnimplementedError();
+  }
+}
+
+/// Object that displays interactive web content, such as for an in-app browser.
+///
+/// Wraps
+/// [WKWebView](https://developer.apple.com/documentation/webkit/wkwebview?language=objc).
+class WebView {
```
stuartmorgan, line 56: I wonder if we should actually keep the WK prefix on these? It's not standard Dart, but it would make it very clear that these are just mirroring the corresponding ObjC class.

I could go either way on it, but I thought it was worth raising.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
```diff
@@ -0,0 +1,157 @@
...
+import 'dart:async';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
+import 'package:webview_flutter_platform_interface/webview_flutter_platform_interface.dart';
+
+import 'web_kit/web_kit.dart' as web_kit;
+
+/// A [Widget] that displays a [web_kit.WebView].
+class WebViewCupertinoWidget extends StatefulWidget {
```
stuartmorgan, line 14: We should probably not call this Cupertino since we'll likely add macOS support to this package at some point, and in Flutter terminology `Cupertino` is specifically iOS. `WKWebViewWidget`? Although that could be confusing, especially if we put `WK` on the other classes. How about `WebViewWebKitWidget`?

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
```diff
@@ -0,0 +1,157 @@
...
+  /// The initial parameters used to setup the WebView.
+  final CreationParams creationParams;
+
+  /// The handler of callbacks made made by [web_kit.NavigationDelegate].
+  final WebViewPlatformCallbacksHandler callbacksHandler;
+
+  /// Manager of named JavaScript channels and forwarding incoming messages on the correct channel.
+  final JavascriptChannelRegistry javascriptChannelRegistry;
+
+  /// A collection of properties that you use to initialize a web view.
```
stuartmorgan, line 34: s/that you use/used/

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return widget.onBuildWidget(controller);
+  }
+}
+
+/// An implementation of [WebViewPlatformController] with the WebKit api.
+class WebViewCupertinoPlatformController extends WebViewPlatformController {
```
stuartmorgan, line 74: Same name change here.

stuartmorgan (approved):
Sorry this took forever to re-review :( LGTM with some small nits.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
bparrishMines: I like this idea. I added `WK` to all the webkit classes.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
bparrishMines: I used `Cupertino` because it is used as the name of the [WebViewPlatform implementation](https://github.com/flutter/plugins/blob/main/packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino.dart#L18), but I think we should also update that with the new plugin interface. What are your thoughts on `WebKitWebViewWidget`? I was also, coincidently, updating the Android implementation to `AndroidWebKitWebViewWidget`.

This also made me consider that this plugin is called `webview_flutter_wkwebview`, but the Apple library is called [WebKit](https://developer.apple.com/documentation/webkit?language=objc) and [WKWebView](https://developer.apple.com/documentation/webkit/wkwebview?language=objc) is only a class in the library. However, the Android library could technically be called [WebKit](https://developer.android.com/reference/android/webkit/package-summary) too,  so I could understand why are plugin naming scheme wouldn't be consistent.

But all in all, I was considering

| Platform      | Platform Implementation Prefix |   Library File Name   |
| :---:        |    :----:   |          :---: |
| iOS/MacOS      | WebKit...       | web_kit.dart   |
| Android   | AndroidWebKit...        | android_web_kit.dart  |

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
bparrishMines: conversation continued in https://github.com/flutter/plugins/pull/4626#discussion_r799729739

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webview_cupertino_widget.dart
stuartmorgan: I like WKWebView because it's less ambiguous; webkit doesn't really say much if you don't know its intended to be a library name. UIWebView was also WebKit-based, for instance. While WKWebView is only one class, it's the core class, with all the others being supporting classes.

PR #4628, opened 2021-12-20
[script]update the version check error output to include information …

hellohuanlin:
…about how to trigger PR description upload

I had this issue when I was trying to land flutter/plugins#4608. I got this error:

```
No version change found, but the change to this package could not be verified to be exempt from version changes according to repository policy. If this is a false positive, please add a line starting with
No version change:
to your PR description with an explanation of why it is exempt.

```

So I updated the PR description and re-run the check, but still not working.

I suspected that it's because the PR description is not uploaded after edited for some reason. In order to verify this assumption, I rehashed the commit, which will most likely trigger the whole pipeline, and the error is gone.

For now, I figured that it may be helpful to include this as temp workaround in the error output.

https://github.com/flutter/flutter/issues/95594

No version change:
This is only CI script. 

## Pre-launch Checklist

- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]` *
(I used [script] since it's not a plugin specific change)

### DIFF
--- a/script/tool/lib/src/version_check_command.dart
+++ b/script/tool/lib/src/version_check_command.dart
@@ -577,7 +577,9 @@ ${indentation}The first version listed in CHANGELOG.md is $fromChangeLog.
             'repository policy. If this is a false positive, please '
             'add a line starting with\n'
             '$_missingVersionChangeJustificationMarker\n'
-            'to your PR description with an explanation of why it is exempt.');
+            'to your PR description with an explanation of why it is exempt.'
+            'If editing PR description does not work, try rehashing your '
+            'commit to trigger the upload.\n');
         return 'Missing version change';
       }
     }


script/tool/lib/src/version_check_command.dart
```diff
@@ -577,7 +577,9 @@ ${indentation}The first version listed in CHANGELOG.md is $fromChangeLog.
             'repository policy. If this is a false positive, please '
             'add a line starting with\n'
             '$_missingVersionChangeJustificationMarker\n'
-            'to your PR description with an explanation of why it is exempt.');
+            'to your PR description with an explanation of why it is exempt.'
+            'If editing PR description does not work, try rehashing your '
```
stuartmorgan, line 581: What does "rehashing" mean? I can maybe guess, but it's not a standard git term that I've ever encountered, so it's not clear that this would clarify for most readers.

I don't think we wast to say "if editing does not work", since that could also happen if, say, someone typos the marker; we should more clearly state that triggerring a re-run won't work.

stuartmorgan: We should also add this to the `!hasChangelogChange` message.

stuartmorgan (changes requested):
This should include test changes.

stuartmorgan:
@hellohuanlin Are you still planning on finishing this PR?

hellohuanlin:
@stuartmorgan Yes. I plan to do it after camera work is done, as I may need to look into how to run this script.

hellohuanlin:
Close it for now. Will get back to it when I get a chance

PR #4629, opened 2021-12-20
Revert "[camera]Fix crash due to calling engine APIs from background thread"

jmagman:
Reverts flutter/plugins#4608

I don't think this is the best way to solve this, let's revert until @stuartmorgan and I can review.

No version change:
That PR didn't bump the version, so there's no publication issue.

PR #4630, opened 2021-12-20
Revert "[camera]Fix crash due to calling engine APIs from background thread"

jmagman:
Reverts flutter/plugins#4608

I don't think this is the best way to solve this, let's revert until @stuartmorgan and I can review.

No version change:
That PR didn't bump the version, so there's no publication issue.


jmagman:
Recreation of https://github.com/flutter/plugins/pull/4629

hellohuanlin (approved):

PR #4638, opened 2022-01-03
[image_picker] add requestFullMetadata for iOS (optional permissions)

PiotrMitkowski:
## Description
Originally #3264 made by @cpboyd 

Make iOS 11+ permissions requests optional per https://developer.apple.com/forums/thread/653414.
PHAssets aren't actually required and there was already some fallback code in place.

## Related issues
Fixes flutter/flutter#65995

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/image_picker/image_picker/lib/image_picker.dart
```diff
@@ -191,6 +191,13 @@ class ImagePicker {
   /// the front or rear camera should be opened, this function is not guaranteed
   /// to work on an Android device.
   ///
+  /// Use `forceFullMetadata` (defaults to `true`) to control, how much additional information the plugin tries to get on iOS.
```
stuartmorgan, line 194: Nit: remove the comma after "control".

stuartmorgan, line 194: Nit: This shouldn't specifically say it's for iOS, as it might apply to other platforms in the future. I would remove iOS here, and add it to the sentence below (e.g., "... which may require extra permission requests on some platforms, such as iOS").

packages/image_picker/image_picker_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.5.0
+
+* Re-adding `forceFullMetadata` option, but as a parameter of a new `pickImageFromSource`
```
stuartmorgan, line 3: Re-adds

(Per style guide)

packages/image_picker/image_picker_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.5.0
+
+* Re-adding `forceFullMetadata` option, but as a parameter of a new `pickImageFromSource` 
+  method (non-breaking change approach).
```
stuartmorgan, line 4: You don't need to explain in the changelog that it's non-breaking; that's indicated by the version.

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
```diff
@@ -270,4 +272,24 @@ class MethodChannelImagePicker extends ImagePickerPlatform {
       files: pickedFileList,
     );
   }
+
+  @override
+  Future<XFile?> getImageFromSource({
```
stuartmorgan, line 277: Please put this next to `getImage`

packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
```diff
@@ -146,6 +146,8 @@ abstract class ImagePickerPlatform extends PlatformInterface {
     throw UnimplementedError('retrieveLostData() has not been implemented.');
   }
 
+  /// This method is deprecated in favor of [getImageFromSource] and will be removed in a future update
```
stuartmorgan, line 149: Nit: Missing final period.

stuartmorgan (changes requested):
Where are the iOS changes? This PR adds all the plumbing, but has no effect on any platform.

packages/image_picker/image_picker_platform_interface/CHANGELOG.md
PiotrMitkowski: Okay, I only wanted to explicitly show the difference from previous implementation

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -97,7 +97,12 @@ - (void)pickImageWithPHPicker:(int)maxImagesAllowed API_AVAILABLE(ios(14)) {
 
   self.maxImagesAllowed = maxImagesAllowed;
 
-  [self checkPhotoAuthorizationForAccessLevel];
+  BOOL usePhaAsset = [[_arguments objectForKey:@"requestFullMetadata"] boolValue];
```
stuartmorgan, line 100: The type is `PHAsset`, not `PHAAsset`, and also ObjC style is to capitalize abbreviations even in variable names, so this, and the similar variables below, should be `usePHAsset` rather than `usePhaAsset`.

packages/image_picker/image_picker/lib/image_picker.dart
```diff
@@ -191,6 +191,10 @@ class ImagePicker {
   /// the front or rear camera should be opened, this function is not guaranteed
   /// to work on an Android device.
   ///
+  /// Use `requestFullMetadata` (defaults to `true`) to control how much additional information the plugin tries to get on iOS.
```
stuartmorgan, line 194: Remove the "on iOS" here.

stuartmorgan (changes requested):
The iOS changes need unit tests that the parameter is actually honored (since this flow can't use integration tests). Right now all of the iOS code here could be broken or removed, and no tests would fail, which means there's no meaningful testing.

PiotrMitkowski:
@stuartmorgan can you give me a little hint on the iOS tests? I see two expected results of having the `requestFullMetadata` flag set to `false`:

1. when running `pickImage` method call, plugin's method  `checkPhotoAuthorization` should not be called
2. after picking an image, plugin's method `saveImageWithPickerInfo` should be called

For now I don't see a way to check if any of those happened in the iOS unit tests.

stuartmorgan:
> I see two expected results of having the `requestFullMetadata` flag set to `false`:

Those aren't expected results, they are internal implementation details. Expected results here would be expressed in terms of the iOS APIs that are or are not called, which would be checked by mocking them.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -142,14 +147,19 @@ - (void)launchUIImagePickerWithSource:(int)imageSource {
...
   self.maxImagesAllowed = 1;
 
   switch (imageSource) {
     case SOURCE_CAMERA:
       [self checkCameraAuthorizationWithImagePicker:imagePickerController];
       break;
     case SOURCE_GALLERY:
-      [self checkPhotoAuthorizationWithImagePicker:imagePickerController];
+      if (usePHAsset) {
+        [self checkPhotoAuthorizationWithImagePicker:imagePickerController];
```
stuartmorgan, line 159: According to the thread you linked to, this change is only correct on iOS 11+, and iOS 9 and 10 still need permissions.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -169,11 +179,11 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
 
   self.result = result;
   _arguments = call.arguments;
+  BOOL usePHAsset = [[_arguments objectForKey:@"requestFullMetadata"] boolValue];
 
   if ([@"pickImage" isEqualToString:call.method]) {
     int imageSource = [call.arguments[@"source"] intValue];
-
-    if (imageSource == SOURCE_GALLERY) {  // Capture is not possible with PHPicker
+    if (usePHAsset && imageSource == SOURCE_GALLERY) {  // Capture is not possible with PHPicker
```
stuartmorgan, line 186: I'm confused by this change. Why is forcing `UIImagePicker` desirable, given that `PHPicker` also shouldn't require permission prompting? Isn't the permission issue specifically related to accessing metadata, not the top-level API used?

stuartmorgan (changes requested):

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
PiotrMitkowski: You're right, for `PHPicker` it's enough to prevent from calling util method `getAssetFromImagePickerInfo`. I'll fix that part

cedvdb:
is the default to not request full metadata ? It should

packages/image_picker/image_picker/lib/image_picker.dart
```diff
@@ -206,13 +210,15 @@ class ImagePicker {
     double? maxHeight,
     int? imageQuality,
     CameraDevice preferredCameraDevice = CameraDevice.rear,
+    bool requestFullMetadata = true,
```
cedvdb, line 213: The default should be false, it makes more sense even if it is a breaking change for some.

stuartmorgan: > Imo, the default should be false, it makes more sense

If this were a new API, I would agree it should default to false, but it's not. We wouldn't make a breaking change to the plugin just for this.

Feel free to file an issue to track changing the default the next time we decide to make a breaking change for other reasons.

> even if it is a breaking change for some.

There's no such thing as a breaking change "for some" in a plugin. A breaking change will change the major version, which affects every client of the plugin whether they would be broken by the specific change or not.

packages/image_picker/image_picker_platform_interface/lib/src/types/image_picker_options.dart
```diff
@@ -0,0 +1,46 @@
...
+  /// the original quality will be returned. Compression is only supported for certain
+  /// image types such as JPEG. If compression is not supported for the image that is picked,
+  /// a warning message will be logged.
+  final int? imageQuality;
+
+  /// Used to specify the camera to use when the `source` is [ImageSource.camera].
+  /// The `preferredCameraDevice` is ignored when `source` is [ImageSource.gallery]. It is also ignored if the chosen camera is not supported on the device.
+  /// Defaults to [CameraDevice.rear]. Note that Android has no documented parameter for an intent to specify if
+  /// the front or rear camera should be opened, this function is not guaranteed
+  /// to work on an Android device.
```
stuartmorgan, line 38: The wrapping on this comment is very uneven; please wrap everything to 80 chars.

stuartmorgan (changes requested):
It looks like you've broken one of the existing iOS tests:
```
Failing tests:
	RunnerUITests:
		-[ImagePickerFromGalleryUITests testPickingFromGallery]
```

packages/image_picker/image_picker_android/pubspec.yaml
```diff
@@ -26,3 +26,9 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   mockito: ^5.0.0
+
+# FOR TESTING ONLY. DO NOT MERGE.
+dependency_overrides:
+  image_picker_platform_interface:
+    path:
+      ../image_picker_platform_interface
```
cyanglaz, line 34: add a line to the end of file

packages/image_picker/image_picker_ios/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -180,6 +180,27 @@ - (void)testPickMultiImageShouldUseUIImagePickerControllerOnPreiOS14 {
...
+  if (@available(iOS 14, *)) {
+    return;
+  }
+  id mockUIImagePicker = OCMClassMock([UIImagePickerController class]);
+  FLTImagePickerPlugin *plugin = [FLTImagePickerPlugin new];
+  [plugin setImagePickerControllerOverrides:@[ mockUIImagePicker ]];
+  FlutterMethodCall *call = [FlutterMethodCall methodCallWithMethodName:@"pickImage"
+                                                              arguments:@{
+                                                                @"source" : @(1),
+                                                                @"requestFullMetadata" : @(false),
```
cyanglaz, line 193: nits: @(NO)

packages/image_picker/image_picker_ios/example/pubspec.yaml
```diff
@@ -27,3 +27,9 @@ dev_dependencies:
 
 flutter:
   uses-material-design: true
+
+# FOR TESTING ONLY. DO NOT MERGE.
+dependency_overrides:
+  image_picker_platform_interface:
+    path:
+      ../../image_picker_platform_interface
```
cyanglaz, line 35: line EOF

packages/image_picker/image_picker_ios/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -133,7 +133,12 @@ - (void)pickImageWithPHPicker:(int)maxImagesAllowed API_AVAILABLE(ios(14)) {
 
   self.maxImagesAllowed = maxImagesAllowed;
 
-  [self checkPhotoAuthorizationForAccessLevel];
+  BOOL usePHAsset = [[_arguments objectForKey:@"requestFullMetadata"] boolValue];
```
cyanglaz, line 136: Should the BOOL name to be `requestFullMetadata`? Or explain why `usePHAsset` is the same as `requestFullMetadata` in a comment

packages/image_picker/image_picker_ios/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -142,14 +147,24 @@ - (void)launchUIImagePickerWithSource:(int)imageSource {
   imagePickerController.delegate = self;
   imagePickerController.mediaTypes = @[ (NSString *)kUTTypeImage ];
 
+  BOOL usePHAsset = [[_arguments objectForKey:@"requestFullMetadata"] boolValue];
```
cyanglaz, line 150: ditto

packages/image_picker/image_picker_ios/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -206,12 +220,17 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
       imagePickerController.videoMaximumDuration = max;
     }
 
+    BOOL usePHAsset = [[_arguments objectForKey:@"requestFullMetadata"] boolValue];
```
cyanglaz, line 223: ditto

cyanglaz (changes requested):

packages/image_picker/image_picker_android/pubspec.yaml
stuartmorgan: @cyanglaz Path-based dependency overrides are always temporary lines; this whole block will be reverted before anything can land, in step 4 of [the instructions](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins). Given that, we shouldn't worry about whitespace on those bits.

packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
```diff
@@ -251,4 +253,33 @@ abstract class ImagePickerPlatform extends PlatformInterface {
...
+  /// Where iOS supports HEIC images, Android 8 and below doesn't. Android 9 and above only support HEIC images if used
+  /// in addition to a size modification, of which the usage is explained in [ImagePickerOptions].
+  ///
+  /// In Android, the MainActivity can be destroyed for various reasons. If that happens, the result will be lost
+  /// in this call. You can then call [getLostData] when your app relaunches to retrieve the lost data.
+  ///
+  /// If no images were picked, the return value is null.
+  Future<XFile?> getImageFromSource({
+    required ImageSource source,
+    ImagePickerOptions? options,
```
cyanglaz, line 274: ditto: can we make `options` non-null?

cyanglaz (changes requested):

packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
```diff
@@ -251,4 +253,33 @@ abstract class ImagePickerPlatform extends PlatformInterface {
...
+  ///
+  /// Where iOS supports HEIC images, Android 8 and below doesn't. Android 9 and above only support HEIC images if used
+  /// in addition to a size modification, of which the usage is explained in [ImagePickerOptions].
+  ///
+  /// In Android, the MainActivity can be destroyed for various reasons. If that happens, the result will be lost
+  /// in this call. You can then call [getLostData] when your app relaunches to retrieve the lost data.
+  ///
+  /// If no images were picked, the return value is null.
+  Future<XFile?> getImageFromSource({
+    required ImageSource source,
```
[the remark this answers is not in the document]
PiotrMitkowski: This parameter is needed whenever the picking method is called. All parameters inside `ImagePickerOptions ` are more optional, so the whole options object can be skipped if a caller doesn't need any specific adjustments. It was @stuartmorgan's suggestion in this PR that I've agreed with, as calling the picking method doesn't require creating the options object for every use

cyanglaz (commented):
LGTM, @stuartmorgan Do you mind taking another look before we update the path dependency?

packages/image_picker/image_picker_platform_interface/lib/src/types/image_picker_options.dart
```diff
@@ -0,0 +1,48 @@
...
+  /// [referredCameraDevice] and [requestFullMetadata]. Any of the params may be null.
+  const ImagePickerOptions({
+    this.maxHeight,
+    this.maxWidth,
+    this.imageQuality,
+    this.preferredCameraDevice = CameraDevice.rear,
+    this.requestFullMetadata = true,
+  });
+
+  /// If specified, the image will be at most `maxWidth` wide. Otherwise,
```
stuartmorgan, line 19: You don't need to include the name of a field in the field comment name (as you do in the context of a method comment, where this was taken from). Also, the second sentence isn't true once you split it apart into two different comments for the two fields, because if one maximum is specified, it'll be scaled to fit that max which changes the other dimension. So it should be:

> The maximum width of the image, in pixels. If `null`, the image will only be resized if `maxHeight` is specified.

packages/image_picker/image_picker_platform_interface/lib/src/types/image_picker_options.dart
```diff
@@ -0,0 +1,48 @@
...
+  /// the image will be returned at its original width.
+  final double? maxWidth;
+
+  /// If specified, the image will be at most`maxHeight` tall.
+  /// Otherwise the image will be returned at its original height.
```
stuartmorgan, line 24: Same here.

packages/image_picker/image_picker_platform_interface/lib/src/types/image_picker_options.dart
```diff
@@ -0,0 +1,48 @@
...
+  final double? maxHeight;
+
+  /// Modifies the quality of the image, ranging from 0-100 where 100 is the original/max
+  /// quality. If `imageQuality` is null, the image with the originalquality will
```
stuartmorgan, line 28: "If null, the image will be returned at its original quality."

stuartmorgan: Also, lets put a blank comment line after the first sentence here, per https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

packages/image_picker/image_picker_platform_interface/lib/src/types/image_picker_options.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:image_picker_platform_interface/src/types/types.dart';
+
+/// Specifies options for picking a single image from the device's camera or gallery.
+class ImagePickerOptions {
+  /// Creates an instance with the given [maxHeight], [maxWidth], [imageQuality],
+  /// [referredCameraDevice] and [requestFullMetadata]. Any of the params may be null.
```
stuartmorgan, line 10: Remove "Any of the params may be null"; nullability is part of types now, and doesn't need to be commented.

packages/image_picker/image_picker_platform_interface/lib/src/types/image_picker_options.dart
```diff
@@ -0,0 +1,48 @@
...
+  /// be returned. Compression is only supported for certain image types such as
+  /// JPEG. If compression is not supported for the image that is picked, a warning
+  /// message will be logged.
+  final int? imageQuality;
+
+  /// Used to specify the camera to use when the `source` is [ImageSource.camera].
+  /// The `preferredCameraDevice` is ignored when `source` is [ImageSource.gallery].
+  /// It is also ignored if the chosen camera is not supported on the device.
+  /// Defaults to [CameraDevice.rear]. Note that Android has no documented parameter
+  /// for an intent to specify if the front or rear camera should be opened, this
```
stuartmorgan, line 38: We try to avoid commenting about specific platforms in the platform interface (other than as illustrative examples). You can just remove this whole sentence since it's a minor detail that callers at the platform interface level aren't going to need to know.

packages/image_picker/image_picker_platform_interface/lib/src/types/image_picker_options.dart
```diff
@@ -0,0 +1,48 @@
...
+
+  /// Modifies the quality of the image, ranging from 0-100 where 100 is the original/max
+  /// quality. If `imageQuality` is null, the image with the originalquality will
+  /// be returned. Compression is only supported for certain image types such as
+  /// JPEG. If compression is not supported for the image that is picked, a warning
+  /// message will be logged.
+  final int? imageQuality;
+
+  /// Used to specify the camera to use when the `source` is [ImageSource.camera].
+  /// The `preferredCameraDevice` is ignored when `source` is [ImageSource.gallery].
```
stuartmorgan, line 35: Remove this sentence, and add a blank line.

packages/image_picker/image_picker_platform_interface/lib/src/types/image_picker_options.dart
```diff
@@ -0,0 +1,48 @@
...
+  /// It is also ignored if the chosen camera is not supported on the device.
```
stuartmorgan, line 36: "Ignored if the source is not [ImageSource.camera], or the chosen camera is not supported on the device."

packages/image_picker/image_picker_platform_interface/lib/src/types/image_picker_options.dart
```diff
@@ -0,0 +1,48 @@
...
+  /// Defaults to [CameraDevice.rear]. Note that Android has no documented parameter
+  /// for an intent to specify if the front or rear camera should be opened, this
+  /// function is not guaranteed to work on an Android device.
+  final CameraDevice preferredCameraDevice;
+
+  /// `requestFullMetadata` defaults to `true`, so the plugin tries to get the
```
stuartmorgan, line 42: A field comment should lead with what it's for, not what the default is. How about:

> If true, requests full image metadata, which may require extra permissions on some platforms, (e.g., `NSPhotoLibraryUsageDescription` on iOS).
>
> Defaults to true.

packages/image_picker/image_picker_ios/pigeons/messages.dart
```diff
@@ -35,9 +35,9 @@ class SourceSpecification {
 @HostApi(dartHostTestHandler: 'TestHostImagePickerApi')
 abstract class ImagePickerApi {
   @async
-  @ObjCSelector('pickImageWithSource:maxSize:quality:')
-  String? pickImage(
-      SourceSpecification source, MaxSize maxSize, int? imageQuality);
+  @ObjCSelector('pickImageWithSource:maxSize:quality:requestFullMetadata:')
```
stuartmorgan, line 38: Nit: this param would be more idiomatic in ObjC as just `fullMetadata:`; the "with" is implied to go with following arguments, and "with full metadata" reads much better than "with request full metadata".

stuartmorgan (changes requested):
I have some documentation feedback and minor nits, but otherwise this looks good. @PiotrMitkowski please go ahead and open the platform interface PR (with the platform interface parts of my comments below addressed) and we can get that a quick final review and land it to start unwinding the full PR.

packages/image_picker/image_picker/lib/image_picker.dart
```diff
@@ -206,13 +210,17 @@ class ImagePicker {
...
   }) {
-    return platform.getImage(
+    return platform.getImageFromSource(
```
stuartmorgan, line 215: You need to require the version of the platform interface that has this method in pubspec.yaml.

packages/image_picker/image_picker/lib/image_picker.dart
```diff
@@ -206,13 +210,17 @@ class ImagePicker {
...
-      maxWidth: maxWidth,
-      maxHeight: maxHeight,
-      imageQuality: imageQuality,
-      preferredCameraDevice: preferredCameraDevice,
+      options: ImagePickerOptions(
+        maxWidth: maxWidth,
+        maxHeight: maxHeight,
+        imageQuality: imageQuality,
+        preferredCameraDevice: preferredCameraDevice,
+        requestFullMetadata: requestFullMetadata,
```
stuartmorgan, line 222: It would also be better to pull the iOS changes into a separate PR and land it first so that you can require that version, otherwise someone who only updates this package will not actually get a working implementation of the new parameter.

packages/image_picker/image_picker/test/image_picker_deprecated_test.dart
```diff
@@ -57,14 +57,16 @@ void main() {
                 'maxWidth': null,
                 'maxHeight': null,
                 'imageQuality': null,
-                'cameraDevice': 0
+                'cameraDevice': 0,
+                'requestFullMetadata': true,
```
stuartmorgan, line 61: You'll need to reconcile the test changes against https://github.com/flutter/plugins/pull/5706 once that lands (which is needed to resolve the current tree breakage, so will be before this PR can land).

packages/image_picker/image_picker_ios/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: image_picker_ios
 description: iOS implementation of the video_picker plugin.
 repository: https://github.com/flutter/plugins/tree/main/packages/image_picker/image_picker_ios
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+image_picker%22
-version: 0.8.5+2
+version: 0.8.6
```
stuartmorgan, line 6: This also needs to require the right version of the platform interface.

stuartmorgan (changes requested):

PiotrMitkowski:
I closed this PR by a mistake today, so I've created a new one: https://github.com/flutter/plugins/pull/5915

PR #4639, opened 2022-01-04
[video_player] Validate size only when assets contain video tracks on iOS

AlexV525:
By adding a more specific type check, we can expand the support range of the video_player on iOS from videos to all supported media types. This is from the documentation of `AVSimplePlayer`: [AVSimplePlayer/AVSPDocument.m#139](https://developer.apple.com/library/archive/samplecode/AVSimplePlayerOSX/Listings/AVSimplePlayer_AVSPDocument_m.html#:~:text=tracksWithMediaType%3AAVMediaTypeVideo)

Resolves flutter/flutter#38480, flutter/flutter#90429, or more issues related to iOS audio playing.
Related to flutter/flutter#47105, flutter/flutter#75636.

## Pre-launch Checklist

- [x] I signed the [CLA].


AlexV525:
Seems the duration for audios on iOS will always be `0:00:00.001000`. Drafting for further investigations.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
```diff
@@ -375,7 +377,7 @@ - (int64_t)position {
 }
 
 - (int64_t)duration {
-  return FLTCMTimeToMillis([[_player currentItem] duration]);
+  return FLTCMTimeToMillis([[[_player currentItem] asset] duration]);
```
AlexV525, line 380: https://openradar.appspot.com/radar?id=4968600712511488

AlexV525:
Hmm...1 millisecond of differences with the duration between platforms make tests fail.

packages/video_player/video_player/example/integration_test/video_player_test.dart
```diff
@@ -240,4 +240,73 @@ void main() {
...
+    testWidgets('can be initialized', (WidgetTester tester) async {
+      await _controller.initialize();
+
+      expect(_controller.value.isInitialized, true);
+      expect(_controller.value.position, const Duration(seconds: 0));
+      expect(_controller.value.isPlaying, false);
+      // Due to the duration calculation accurancy between platforms,
+      // the milliseconds on Web will be a slightly different from natives.
+      // The audio was made with 44100 Hz, 192 Kbps CBR, and 32 bits.
+      if (kIsWeb) {
```
AlexV525, line 258: It's 5.041625 sec. The web takes `5.042` and natives take `5.041`.

AlexV525: I've done some debug steps, and the root cause of this difference is from here:
https://github.com/flutter/plugins/blob/a10ac2e0d8a8cd878fd0d0a54aa9a87c44c7d7c5/packages/video_player/video_player_web/lib/video_player_web.dart#L301
If we want to make them equal, we might use `toInt()` here.

stuartmorgan:
I don't think this actually resolves flutter/flutter#47105 as claimed in the PR description; using `video_player` to play audio isn't actually a good idea, and will be even less so when if (most likely when) we migrate it to platform views.

I'm conflicted about this PR. The change itself is simple, but adding tests asserting that audio works sets a problematic precedent. What if a new platform we want to support has different APIs for supporting video and audio; are we now saying we require audio support in that implementation too? When there are inevitably new issues or feature requests that are specific to audio, will those all suddenly be valid?

I think where I come down on this is that because the mismatch with Android is currently a source of confusion, and the incremental cost is extremely low here, I'm inclined to take this PR, but:
- The official position should still be that we don't officially support audio
- We should still do https://github.com/flutter/flutter/issues/75636, and explicitly clarify in the README that future versions and/or platforms may not support audio (I can do this in a follow-up)
- The test group being added here should have a comment saying that audio playback is tested to prevent *accidental* regression, but could be removed in the future if there's a technical reason to do so.

@jmagman Does that sound good to you?

jmagman:
> @jmagman Does that sound good to you?

SGTM

AlexV525:
> using video_player to play audio isn't actually a good idea

That's more like it used to be supported implicitly, but someday it becomes invalid. I'll update the above suggestions.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
jmagman: You may want to put this link in a comment over this line.

jmagman (approved):
This LGTM with comment nit.

packages/video_player/video_player/ios/Classes/FLTVideoPlayerPlugin.m
AlexV525: Done.

PR #4641, opened 2022-01-05
[camera_windows] Support camera on windows desktop platform

jokerttu:
This PR adds a Windows Desktop implementation of [camera plugin](https://pub.dev/packages/camera) using Media Foundation and CaptureEngine interface.

The implementation supports multiple cameras, video preview, and capturing photos and videos. Works with camera plugin and [camera example](https://pub.dev/packages/camera/example).

The current implementation uses FlutterDesktopPixelBuffer, but could be switched to use shared textures or Platform views after those are supported.

Unit tests are partially done; for the CaptureControllerImpl class all falilure code paths are not yet covered.

Still missing some features, like exposure and focus controls. These can be done later with interfaces IAMCameraControl and IAMVideoProcAmp,

Resolves flutter/flutter#41709: [camera] Add Windows support

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,454 @@
...
+          ExposureMode.auto,
+          false,
+          FocusMode.auto,
+          false,
+        ),
+      );
+    } else {
+      throw CameraException(
+        'INITIALIZATION_FAILED',
+        'The platform "$defaultTargetPlatform" did not return valid data when reporting success. The platform should always return a valid data or report an error.',
```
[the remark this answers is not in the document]
stuartmorgan: Although in this case, it sounds like it's describing an implementation bug in the native code; providing an exception with an error message isn't useful in that case since the developer who would get it can't act on this.

Just use `!` (which will throw an exception if there's a bug in the plugin).

This applies below as well.

packages/camera/camera_windows/pubspec.yaml
```diff
@@ -0,0 +1,30 @@
+name: camera_windows
+description: A Flutter plugin for getting information about and controlling the camera on Windows.
+version: 0.0.1
+# TODO: Update next lines to match proper repository and issue_tracker url
```
stuartmorgan, line 4: Why is this a TODO? You just need to remove the commenting AFAICT.

packages/camera/camera_windows/pubspec.yaml
```diff
@@ -0,0 +1,30 @@
...
+#repository: https://github.com/flutter/plugins/tree/master/packages/camera/camera_windows
+#issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+camera%22
+# TODO: Remove next line after plugin is ready to be published
```
stuartmorgan, line 7: What changes are planned before publishing?

packages/camera/camera_windows/pubspec.yaml
```diff
@@ -0,0 +1,30 @@
...
+publish_to: 'none'
+
+environment:
+  sdk: ">=2.12.0 <3.0.0"
+  flutter: ">=2.5.3"
```
stuartmorgan, line 12: Just use 2.5.0 unless you really need .3 for some reason.

packages/camera/camera_windows/windows/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+)
+
+apply_standard_settings(${PLUGIN_NAME})
+set_target_properties(${PLUGIN_NAME} PROPERTIES
+  CXX_VISIBILITY_PRESET hidden)
+target_compile_definitions(${PLUGIN_NAME} PRIVATE FLUTTER_PLUGIN_IMPL)
+target_include_directories(${PLUGIN_NAME} INTERFACE
+  "${CMAKE_CURRENT_SOURCE_DIR}/include")
+target_link_libraries(${PLUGIN_NAME} PRIVATE flutter flutter_wrapper_plugin)
+target_link_libraries(${PLUGIN_NAME} PRIVATE mf mfplat mfreadwrite mfuuid d3d11)
```
stuartmorgan, line 38: Do these limit the version of Windows that is supported? If so, let's mention that in the README.

packages/camera/camera_windows/windows/camera.cpp
```diff
@@ -0,0 +1,271 @@
...
+}
+
+}  // namespace
+
+CameraImpl::CameraImpl(const std::string &device_id)
+    : device_id_(device_id),
+      capture_controller_(nullptr),
+      messenger_(nullptr),
+      camera_id_(-1),
+      Camera(device_id) {}
```
stuartmorgan, line 34: Prefer initializing at declaration rather than in the constructor.

packages/camera/camera_windows/windows/camera.cpp
```diff
@@ -0,0 +1,271 @@
...
+  assert(!device_id_.empty());
+  messenger_ = messenger;
+  capture_controller_ =
+      capture_controller_factory->CreateCaptureController(this);
+  capture_controller_->InitCaptureDevice(texture_registrar, device_id_,
+                                         enable_audio, resolution_preset);
+}
+
+// Adds pending result to the pending_results map.
+// If result already exists, call result error handler
```
stuartmorgan, line 66: calls

(Please audit declaration comments for descriptive verb form, per style guide)

packages/camera/camera_windows/windows/camera_plugin.h
```diff
@@ -0,0 +1,93 @@
...
+
+  // Disallow copy and move.
+  CameraPlugin(const CameraPlugin &) = delete;
+  CameraPlugin &operator=(const CameraPlugin &) = delete;
+
+  // Called when a method is called on plugin channel;
+  void HandleMethodCall(const flutter::MethodCall<> &method_call,
+                        std::unique_ptr<MethodResult<>> result);
+
+ protected:
```
stuartmorgan, line 46: Why are there protected elements? Plugins aren't meant to be subclassed.

packages/camera/camera_windows/windows/camera.h
```diff
@@ -0,0 +1,155 @@
...
+  void OnStopRecordFailed(const std::string &error) override;
+  void OnTakePictureSucceeded(const std::string &filepath) override;
+  void OnTakePictureFailed(const std::string &error) override;
+  void OnVideoRecordSucceeded(const std::string &filepath,
+                              int64_t video_duration) override;
+  void OnVideoRecordFailed(const std::string &error) override;
+  void OnCaptureError(const std::string &error) override;
+
+  // From Camera
+
```
stuartmorgan, line 84: Please remove all blank lines from this block of overrides (as above).

packages/camera/camera_windows/windows/camera.h
```diff
@@ -0,0 +1,155 @@
...
+  camera_windows::CaptureController *GetCaptureController() override {
+    return capture_controller_.get();
+  };
+
+  void InitCamera(flutter::TextureRegistrar *texture_registrar,
+                  flutter::BinaryMessenger *messenger, bool enable_audio,
+                  ResolutionPreset resolution_preset) override;
+
+  void InitCamera(
```
stuartmorgan, line 106: All non-override methods need declaration comments, per style guide. (In at least some cases you already have declaration-sytle comments, they just need to be moved from the implementation file to here.)

Applies to all files in the PR.

packages/camera/camera_windows/windows/camera.cpp
```diff
@@ -0,0 +1,271 @@
...
+    return nullptr;
+  }
+  auto result = std::move(it->second);
+  pending_results_.erase(it);
+  return result;
+}
+
+bool CameraImpl::HasPendingResultByType(PendingResultType type) {
+  if (pending_results_.empty()) {
+    return nullptr;
```
stuartmorgan, line 97: nullptr is not a bool type.

packages/camera/camera_windows/windows/camera.cpp
```diff
@@ -0,0 +1,271 @@
...
+using flutter::EncodableMap;
+using flutter::EncodableValue;
+
+namespace {
+// Camera channel events
+const char kCameraMethodChannelBaseName[] = "flutter.io/cameraPlugin/camera";
+const char kVideoRecordedEvent[] = "video_recorded";
+const char kErrorEvent[] = "error";
+
+// Helper function for creating messaging channel for camera
```
stuartmorgan, line 18: Why are you constantly creating new method channels instead of keeping them around?

packages/camera/camera_windows/windows/camera_plugin.cpp
```diff
@@ -0,0 +1,593 @@
...
+    return ResolutionPreset::RESOLUTION_PRESET_VERY_HIGH;
+  } else if (resolution_preset.compare(kResolutionPresetValueUltraHigh) == 0) {
+    return ResolutionPreset::RESOLUTION_PRESET_ULTRA_HIGH;
+  } else if (resolution_preset.compare(kResolutionPresetValueMax) == 0) {
+    return ResolutionPreset::RESOLUTION_PRESET_MAX;
+  }
+  return ResolutionPreset::RESOLUTION_PRESET_AUTO;
+}
+
+bool HasCurrentTextureId(int64_t current_camera_id, const EncodableMap &args) {
```
stuartmorgan, line 89: All these private helpers need declaration comments.

packages/camera/camera_windows/windows/camera_plugin.cpp
```diff
@@ -0,0 +1,593 @@
...
+    UINT32 id_size;
+    hr = device->GetAllocatedString(
+        MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE_VIDCAP_SYMBOLIC_LINK, &id, &id_size);
+
+    if (SUCCEEDED(hr)) {
+      device_info->display_name = Utf8FromUtf16(std::wstring(name, name_size));
+      device_info->device_id = Utf8FromUtf16(std::wstring(id, id_size));
+    }
+
+    ::CoTaskMemFree(id);
```
stuartmorgan, line 118: Please replace manual memory management with smart pointers whenever possible; there's a fair amount of non-trivial logic around memory in this code. I believe `CComHeapPtr` will cover many of the cases.

packages/camera/camera_windows/windows/camera_plugin.h
```diff
@@ -0,0 +1,93 @@
...
+  Camera *GetCameraByCameraId(int64_t camera_id);
+  void DisposeCameraByCameraId(int64_t camera_id);
+
+  bool EnumerateVideoCaptureDeviceSources(IMFActivate ***devices,
+                                          UINT32 *count) override;
+
+ private:
+  std::unique_ptr<CameraFactory> camera_factory_;
+  flutter::TextureRegistrar *texture_registrar_;
+  flutter::BinaryMessenger *messenger_;
```
stuartmorgan, line 59: https://google.github.io/styleguide/cppguide.html#Declaration_Order

packages/camera/camera_windows/windows/camera_plugin.h
```diff
@@ -0,0 +1,93 @@
...
+
+  // Method handlers
+
```
stuartmorgan, line 62: Please remove the blank lines from this block. (It's fine to not document them individually given the rote structure.)

packages/camera/camera_windows/windows/capture_controller.cpp
```diff
@@ -0,0 +1,1185 @@
...
+  HRESULT hr = MFCreateAttributes(&attributes, 1);
+
+  if (SUCCEEDED(hr)) {
+    hr = attributes->SetGUID(MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE,
+                             MF_DEVSOURCE_ATTRIBUTE_SOURCE_TYPE_VIDCAP_GUID);
+  }
+
+  if (SUCCEEDED(hr)) {
+    hr = MFEnumDeviceSources(attributes.Get(), devices, count);
+  }
```
stuartmorgan, line 42: This is a weird pattern. Why not:
```
if (!SUCCEEDED(hr)) {
  return hr;
}
```

so that all of the normal logic is the main body, instead of appearing to be conditional?

In addition to being odd to read; it seems dangerous; it would be very easy to accidentally have one stray line of code later that wasn't gated and ran even for failures.

packages/camera/camera_windows/windows/capture_controller.h
```diff
@@ -0,0 +1,285 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef PACKAGES_CAMERA_CAMERA_WINDOWS_WINDOWS_CAPTURE_CONTROLLER_H_
+#define PACKAGES_CAMERA_CAMERA_WINDOWS_WINDOWS_CAPTURE_CONTROLLER_H_
+
+#include <d3d11.h>
+//#include <d3d12.h>
```
stuartmorgan, line 9: ?

packages/camera/camera_windows/windows/capture_controller.h
```diff
@@ -0,0 +1,285 @@
...
+#include <memory>
+#include <string>
+
+#include "capture_controller_listener.h"
+#include "capture_engine_listener.h"
+
+namespace camera_windows {
+using Microsoft::WRL::ComPtr;
+
+struct FlutterDesktop_Pixel {
```
stuartmorgan, line 28: Class and struct names don't generally use _

packages/camera/camera_windows/windows/capture_controller.h
```diff
@@ -0,0 +1,285 @@
...
+  uint64_t recording_duration_us_ = 0;
+  int64_t max_video_duration_ms_ = -1;
+
+  uint32_t capture_frame_width_ = 0;
+  uint32_t capture_frame_height_ = 0;
+  ComPtr<IMFMediaType> base_capture_media_type_;
+  ComPtr<IMFCapturePhotoSink> photo_sink_;
+  ComPtr<IMFCaptureRecordSink> record_sink_;
+  std::string pending_picture_path_ = "";
+  std::string pending_record_path_ = "";
```
stuartmorgan, line 222: This is a *lot* of ivars for one class. If you have this many things, and they have their own groupings and headings, that's a strong signal that you need to break out responsibilities into helper classes.

stuartmorgan (changes requested):
First pass comments below. I haven't reviewed most of the actual logic (@cbracken will be the primary reviewer on the actual Windows APIs), but I did a pass for things that looked like general style issues or structure/pattern issues.

packages/camera/camera_windows/README.md
```diff
@@ -0,0 +1,53 @@
...
+*Note*: This plugin is under development. See [missing implementation](#missing-implementation).
+
+## Usage
+
+### Depend on the package
+
+This package is not an [endorsed](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#endorsed-federated-plugin) implementation of the [`camera`][camera] plugin yet, so you'll need to 
+[add it explicitly](https://pub.dev/packages/camera_windows/install)
+
+## Example
```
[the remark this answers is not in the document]
jokerttu: Example removed, cannot point to https://pub.dev/packages/camera#example either as this plugin is not endorsed.

packages/camera/camera_windows/windows/CMakeLists.txt
jokerttu: This is good point.
These libraries are available from Windows 7 [Media Foundation Headers and Libraries](https://docs.microsoft.com/en-us/windows/win32/medfound/media-foundation-headers-and-libraries)
[Minimum supported client for used IMFCaptureEngine](https://docs.microsoft.com/en-us/windows/win32/api/mfcaptureengine/nn-mfcaptureengine-imfcaptureengine) is Windows 8
I have tested camera plugin with Windows 10 and with Windows 11.
This should be tested with Win8 and Win7 as well, but I haven't been able to do so yet.

packages/camera/camera_windows/example/integration_test/camera_test.dart
```diff
@@ -0,0 +1,96 @@
...
+import 'package:async/async.dart';
+import 'package:camera_platform_interface/camera_platform_interface.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  // TODO(jokerttu): write rest of the integration tests after camera support is available on windows test framework or native plugin code is mockable
```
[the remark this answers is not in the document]
jokerttu: The comment was poorly written. It was only intended to point out that the remaining functionalities can only be tested if one or more cameras are available in the test environment (or if plugin registration can be mocked somehow, but that won't of course be possible)

Converted it as a comment before main function:
```
// Note that these integration tests do not currently cover
// all features and code paths, as they can only be tested if
// one or more cameras are available in the test environment.
```

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,454 @@
...
+      const MethodChannel('plugins.flutter.io/camera');
+
+  /// Registers the Windows implementation of CameraPlatform.
+  static void registerWith() {
+    CameraPlatform.instance = CameraWindows();
+  }
+
+  final Map<int, MethodChannel> _channels = <int, MethodChannel>{};
+
+  /// The controller we need to broadcast the different events coming
```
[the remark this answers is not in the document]
jokerttu: This was actually copied directly from [method_channel_camera.dart](https://github.com/flutter/plugins/blob/main/packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart#L33) and should be fixed there as well?
Fixed

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  @override
+  Future<void> initializeCamera(
+    int cameraId, {
+    ImageFormatGroup imageFormatGroup = ImageFormatGroup.unknown,
+  }) async {
+    final int requestedCameraId = cameraId;
+
+    //Create channel for camera events
```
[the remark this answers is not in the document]
jokerttu: Done, but I am not closing this yet as I need to go them through once more to be sure.
Leaving this open for recheck.

packages/camera/camera_windows/lib/camera_windows.dart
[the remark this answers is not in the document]
jokerttu: This should be fixed here as well: [method_channel_camera.dart](https://github.com/flutter/plugins/blob/main/packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart#L209-L212)

FIXED

packages/camera/camera_windows/pubspec.yaml
jokerttu: Ok, I added this here only to avoid possible accidental publishing on situations unknown for me...
Removed.

packages/camera/camera_windows/windows/camera.cpp
```diff
@@ -0,0 +1,271 @@
...
+    return nullptr;
+  }
+  auto it = pending_results_.find(type);
+  if (it == pending_results_.end()) {
+    return false;
+  }
+  return it->second != nullptr;
+}
+
+void CameraImpl::SendErrorForPendingResults(const std::string &error_id,
```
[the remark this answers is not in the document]
jokerttu: Seems like `dart run ./script/tool/bin/flutter_plugin_tools.dart format` does not fix these and my IDE currently does not use right style either.
I still need to fix this by configuring right formatter/linter to handle these proper way.

packages/camera/camera_windows/windows/capture_controller.h
jokerttu: Yes, I'll refactor these. Some of the logic can be moved to helper classes, like texture handling.
Some of the states can be done with enum states instead of having multiple booleans, to be more error tolerable as well.
This has grown on development and definitely needs some refactoring.

packages/camera/camera_windows/windows/capture_controller.cpp
jokerttu: You are right, this pattern is left over from goto pattern which was used on original MediaFoundation documentation. After I moved to smart pointers (ComPtr) these should have been rewritten as well.
I'll fix these.

packages/camera/camera_windows/windows/camera.cpp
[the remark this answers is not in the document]
stuartmorgan: I believe clang-format defaults to "prevailing style" for this, and we don't have an explicit setting. (It used to be impossible to have different settings for C++ and ObjC due to `.h` files, but I think that may be fixed now, in which case we could explicitly set it to type-align for C++ and name-align for ObjC.)

packages/camera/camera_windows/windows/camera_plugin.h
stuartmorgan: The standard way to access internals—when it is absolutely necessary—in C++ tests is `friend` annotations granting access to a test fixture.

packages/camera/camera_windows/windows/camera.cpp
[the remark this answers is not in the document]
jokerttu: Ok, I found out the reason why my clang-format was aligning pointers as name-align.
This is because clang-setting DerivePointerAlignment is set to true, and it just uses most common aligment of the file.
I'll fix all files with temporary .clang-format settings as:
```
BasedOnStyle: Google

Language: Cpp
# Force pointers to the type for C++.
DerivePointerAlignment: false
PointerAlignment: Left
```

cbracken (commented):
Sending out a few early comments, will keep going on this through the day.

packages/camera/camera_windows/windows/camera_plugin.cpp
jokerttu: I added CComHeapPtr from atlbase.h and these are working great on local environment as I have ATL support installed. 
But as ATL support is not available on test environment, build fails: 
`camera_plugin.cpp(7,10): fatal error C1083: Cannot open include file: 'atlbase.h': No such file or directory`
Requiring ATL support for developers might not be a good idea either.

jokerttu:
One problematic area is fatal errors caused by capture engine. For example situation where camera is disconnected while it is on use. Current implementation sends Error to all pending requests and as well send separate error to the camera channel. These errors can be listened and the behaviour follows same logic as on other platforms (afaik).
In most of the cases camera is un-usable after CaptureEngine fails.

I see few solutions:
- We could just destroy camera automatically at camera_plugin if camera reports error state. After that any requests for camera via platform channel would just inform that camera does not exists anymore. At flutter code camera still exists.
- Mark camera to _error state_, and return this error state each time any request is send over method channel.
- As these errors already flows to the flutter via camera channel, we could just inform plugin users to listen these error events and dispose camera at flutter end to clear the state.

packages/camera/camera_windows/windows/camera_plugin.cpp
stuartmorgan: Sorry, I forgot that it was from ATL. A custom `unique_ptr` sounds like a good option.

jokerttu:
Fixed huge issue of mediatype selection for preview and video record.
From now on video and preview capture prefers best frame_rate available, not only preferred resolution.
This is because some cameras could give multiple supported media types for same resolutions with different frame_rates, and in most cases biggest resolutions might have framerates like 5fps; and these are not really usable for preview or video capture, and should be only used for photo capture.
This change fixes a lot of issues found on some cameras where wrong media type was selected by default.

opiumfive:
Thanks for your hard work. 
It will be very cool if it supports streaming frames (like `cameraController.startImageStream(..)` in camera plugin) or at least taking picture into CameraImage not just XFile. Need this to implement object recognition in the webcamera realtime. Thanks!

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,437 @@
...
+
+      if (cameras == null) {
+        return <CameraDescription>[];
+      }
+
+      return cameras.map((Map<dynamic, dynamic> camera) {
+        return CameraDescription(
+          name: camera['name'] as String,
+          lensDirection:
+              parseCameraLensDirection(camera['lensFacing'] as String),
```
cbracken, line 59: Is this ever used anywhere else? Consider making it a private function in this file.

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,437 @@
...
+    ResolutionPreset? resolutionPreset, {
+    bool enableAudio = false,
+  }) async {
+    try {
+      final Map<String, dynamic>? reply = await _pluginChannel
+          .invokeMapMethod<String, dynamic>('create', <String, dynamic>{
+        'cameraName': cameraDescription.name,
+        'resolutionPreset': resolutionPreset != null
+            ? _serializeResolutionPreset(resolutionPreset)
+            : null,
```
cbracken, line 80: What does it mean to have a `null` resolution preset? It doesn't look like it's documented on the interface. I assume the camera default?

We should document this in the parent interface.

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,437 @@
...
+  }
+
+  @override
+  Stream<CameraInitializedEvent> onCameraInitialized(int cameraId) {
+    return _cameraEvents(cameraId).whereType<CameraInitializedEvent>();
+  }
+
+  @override
+  Stream<CameraResolutionChangedEvent> onCameraResolutionChanged(int cameraId) {
+    /// Windows camera plugin does not support resolution changed events.
```
cbracken, line 161: Document why. e.g. Windows APIs don't support resolution changing vs we haven't yet implemented this (in which case please file a bug and rephrase as a TODO).

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,437 @@
...
+  }
+
+  @override
+  Stream<VideoRecordedEvent> onVideoRecordedEvent(int cameraId) {
+    return _cameraEvents(cameraId).whereType<VideoRecordedEvent>();
+  }
+
+  @override
+  Stream<DeviceOrientationChangedEvent> onDeviceOrientationChanged() {
+    /// Windows camera plugin does not support capture orientations.
```
cbracken, line 182: Similar comment/question to the above. Document why this is -- i.e. no Win32 API support for this, vs we haven't yet had time to implement.

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,437 @@
...
+      const DeviceOrientationChangedEvent(DeviceOrientation.landscapeRight),
+    );
+  }
+
+  @override
+  Future<void> lockCaptureOrientation(
+    int cameraId,
+    DeviceOrientation orientation,
+  ) async {
+    throw UnimplementedError('lockCaptureOrientation() is not implemented.');
```
cbracken, line 194: Here and elsewhere: file a bug to implement this and add a TODO.

Alternatively, if this will never be supported, use `UnsupportedError` instead.

https://api.flutter.dev/flutter/dart-core/UnsupportedError-class.html

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,437 @@
...
+    } on PlatformException catch (e) {
+      throw CameraException(e.code, e.message);
+    }
+
+    return XFile(path!);
+  }
+
+  @override
+  Future<void> pauseVideoRecording(int cameraId) async {
+    throw UnimplementedError('pauseVideoRecording() is not implemented.');
```
cbracken, line 262: Similar here and below. Add TODOs that point to a bug or use Unsupported if we intend never to support this.

cbracken (changes requested):
Additional comments on Dart side.

packages/camera/camera_windows/example/lib/main.dart
```diff
@@ -0,0 +1,417 @@
...
+  /// Fetches list of available cameras from camera_windows plugin.
+  Future<void> _getAvailableCameras() async {
+    String cameraInfo;
+    List<CameraDescription> cameras = <CameraDescription>[];
+
+    int cameraIndex = _cameraIndex;
+    try {
+      cameras = await CameraPlatform.instance.availableCameras();
+      cameraIndex = cameraIndex % cameras.length;
+      if (cameras.isEmpty) {
```
cbracken, line 59: In the case where `cameras.isEmpty` is true, the modulo operation on the line above will throw `IntegerDivisionByZeroException`. Consider doing that operation conditionally in the `else` block of the conditional. 

In the `if` block you could reset to some obviously invalid index. -1 would make it incredibly obvious that the index is invalid, but you'd need to account for it in the modulo expression in the else, since you'd (probably?) want to reset to index 0 when you finally get a list of cameras.

Since technically every index is invalid, 0 seems like an equally fair choice, and behaves nicely with the modulo call the next time this is called.

packages/camera/camera_windows/example/lib/main.dart
```diff
@@ -0,0 +1,417 @@
...
+                  DropdownButton<ResolutionPreset>(
+                    value: _resolutionPreset,
+                    onChanged: (ResolutionPreset? value) =>
+                        _onResolutionChange(value),
+                    items: resolutionItems,
+                  ),
+                  const SizedBox(width: 20),
+                  const Text(
+                    'Audio:',
+                  ),
```
cbracken, line 334: nit: Consider deleting the trailing comma and making this a one-liner.

packages/camera/camera_windows/example/lib/main.dart
```diff
@@ -0,0 +1,417 @@
...
+
+  Future<void> _switchCamera() async {
+    _cameraIndex = (_cameraIndex + 1) % _cameras.length;
+    if (_initialized && _cameraId >= 0) {
+      await _disposeCurrentCamera();
+      if (_cameras.isNotEmpty) {
+        await _initializeCamera();
+      }
+    } else {
+      await _getAvailableCameras();
```
cbracken, line 248: Do we also need to call _initialize here conditionally on the camera list being non-empty? If so, move lines 244-246 down below this if/else block.

packages/camera/camera_windows/example/lib/main.dart
```diff
@@ -0,0 +1,417 @@
...
+    }
+  }
+
+  Future<void> _onResolutionChange(ResolutionPreset? newValue) async {
+    if (newValue == null) {
+      return;
+    }
```
cbracken, line 255: It seems like null resolution presets are nonsensical here. Consider declaring `newValue` non-null in the parameters list and not calling this method when we have null values. That keeps all the null handling near the outer API surface of this plugin rather than handling them throughout.

cbracken (commented):
Additional comments.

packages/camera/camera_windows/example/pubspec.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  integration_test:
+    sdk: flutter
+  pedantic: ^1.10.0
+
+flutter:
+  uses-material-design: true
```
cbracken, line 29: nit: add a trailing newline

packages/camera/camera_windows/example/windows/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
+cmake_minimum_required(VERSION 3.14)
```
cbracken, line 1: I don't think it's necesary to do for this PR, but once this has landed we should probably regenerate with the latest template. There have been some (minor) changes, mostly just comments, recently.

packages/camera/camera_windows/lib/camera_windows.dart
jokerttu: Currently plugin uses best available resolution if resolutionPreset value is null. Current plugin implementation loops cameras resolutions (media_types) through and selects best resolution fitting to the resolutionPreset. If resolution preset is null, best available resolution is selected. Using cameras default media type is one option, but would need separate implementation.
In most cases users use CameraController from main camera plugin which requires resolutionPreset value as non-nullable. But this of course should be documented somewhere.

jokerttu: Not sure should this value ever be nullable for any platform as it seems to throw error on both Android and iOS if null value is given.

packages/camera/camera_windows/example/lib/main.dart
jokerttu: No, if camera is not initialized, we are not initializing it after switch.
We just fetch current camera list from the plugin to reset current camera to new index and make sure there is enough cameras available.
Fixed the module error for this function as well.

packages/camera/camera_windows/lib/camera_windows.dart
jokerttu: @cbracken just added TODOs for each of these for now.
Is is ok to file a bugs before this source is landed, as I see it a bit problematic to have loose issues without source to point to.

packages/camera/camera_windows/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Usage
+
+### Depend on the package
+
+This package is not an [endorsed][endorsed-federated-plugin] 
+implementation of the [`camera`][camera] plugin, so you'll need to
+[add it explicitly][install]
+
+## Limitations on the Windows platform
```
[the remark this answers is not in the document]
jokerttu: @cbracken
Same as above,
Is is ok to file a bugs before this source is landed, as I see it a bit problematic to have loose issues without source to point to.

packages/camera/camera_windows/lib/camera_windows.dart
stuartmorgan: It's fine to file issues without the source landed; that's how almost all TODO issue links work. You can reference the PR instead (e.g., "https://github.com/flutter/plugins/pull/4641 adds support for Windows but <foo> isn't yet implemented. [...]")

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,424 @@
...
+  @override
+  Stream<VideoRecordedEvent> onVideoRecordedEvent(int cameraId) {
+    return _cameraEvents(cameraId).whereType<VideoRecordedEvent>();
+  }
+
+  @override
+  Stream<DeviceOrientationChangedEvent> onDeviceOrientationChanged() {
+    // TODO(jokerttu): Implement device orientation detection.
+    //
+    // Returns empty stream.
```
stuartmorgan, line 180: If the comment is just the single line of code it's commenting reworded as English, it's generally better to just omit the comment.

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,424 @@
...
+      case 'error':
+        _cameraEventStreamController.add(
+          CameraErrorEvent(
+            cameraId,
+            call.arguments['description'] as String,
+          ),
+        );
+        break;
+      default:
+        throw MissingPluginException();
```
stuartmorgan, line 409: Why MissingPlugin rather than Unimplemented?

stuartmorgan (changes requested):
A few minor comments on the metadata and Dart code. (I didn't look at the C++ code again since cbracken is reviewing it already.)

packages/camera/camera_windows/lib/camera_windows.dart
jokerttu: No real reason; following same patters used on Android and iOS platforms at [here](https://github.com/flutter/plugins/blob/main/packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart#L534)
and also [here](https://github.com/flutter/plugins/blob/main/packages/webview_flutter/webview_flutter_platform_interface/lib/src/method_channel/webview_method_channel.dart#L79)

I can change it to `UnimplementedError `if that suits better.

jokerttu:
@cbracken 
Required flutter version is now set to `flutter: ">=2.8.0"` from previous `flutter: ">=2.5.0"`

This is because the need of FlutterPixelBuffer support for `release_context `and `release_callback`.

These are included to the engine with this commit: [3ba3eb5424a96be](https://github.com/flutter/engine/commit/3ba3eb5424a96be1e7484d0a9ea91341721fdb08)

I am not sure what was the first flutter version where this was in. Should we set the needed version as low as possible.

packages/camera/camera_windows/README.md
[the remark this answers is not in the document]
jokerttu: Readme is now updated and contains links to the feature issues.

packages/camera/camera_windows/lib/camera_windows.dart
stuartmorgan: Strange, I'm not sure why it does that. UnimplementedError would be much less confusing (the plugin definitely isn't missing, and a missing plugin error message can easily lead people down the wrong path), so we should use it here.

stuartmorgan:
> @cbracken Required flutter version is now set to `flutter: ">=2.8.0"` from previous `flutter: ">=2.5.0"`
> 
> This is because the need of FlutterPixelBuffer support for `release_context `and `release_callback`.
> 
> These are included to the engine with this commit: [3ba3eb5424a96be](https://github.com/flutter/engine/commit/3ba3eb5424a96be1e7484d0a9ea91341721fdb08)
> 
> I am not sure what was the first flutter version where this was in. Should we set the needed version as low as possible?

It landed between the 2.5 and 2.6 branch cuts, so 2.8 is the first stable release it was in. It's not worth setting a lower bound any lower than the lowest stable version once something is on stable, so 2.8 is the way to go.

cbracken:
@jokerttu here's the error in the error handling I mentioned. It seems to still repro at the latest commit. The error looks to be an issue with the encoding of the error string message in the response to the method invocation to initialise the camera maybe.
```
An Observatory debugger and profiler on Windows is available at: http://127.0.0.1:58023/iAyDsXb9ZS0=/
The Flutter DevTools debugger and profiler on Windows is available at: http://127.0.0.1:9100?uri=http://127.0.0.1:58023/iAyDsXb9ZS0=/
[ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: FormatException: Missing extension byte (at offset 24)
#0      _Utf8Decoder.convertSingle (dart:convert-patch/convert_patch.dart:1789:7)
#1      Utf8Decoder.convert (dart:convert/utf.dart:351:42)
#2      StandardMessageCodec.readValueOfType (package:flutter/src/services/message_codecs.dart:463:29)
#3      StandardMessageCodec.readValue (package:flutter/src/services/message_codecs.dart:438:12)
#4      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:603:47)
#5      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:177:18)
<asynchronous suspension>
#6      MethodChannel.invokeMapMethod (package:flutter/src/services/platform_channel.dart:377:43)
<asynchronous suspension>
#7      CameraWindows.initializeCamera (package:camera_windows/camera_windows.dart:107:15)
<asynchronous suspension>
#8      _MyAppState._initializeCamera (package:camera_windows_example/main.dart:119:7)
<asynchronous suspension>


══╡ EXCEPTION CAUGHT BY SERVICES LIBRARY ╞══════════════════════════════════════════════════════════
The following FormatException was thrown during a platform message callback:
Missing extension byte (at offset 24)

#0      _Utf8Decoder.convertSingle (dart:convert-patch/convert_patch.dart:1789:7)
#1      Utf8Decoder.convert (dart:convert/utf.dart:351:42)
#2      StandardMessageCodec.readValueOfType (package:flutter/src/services/message_codecs.dart:463:29)
#3      StandardMessageCodec.readValue (package:flutter/src/services/message_codecs.dart:438:12)
#4      StandardMessageCodec.readValueOfType (package:flutter/src/services/message_codecs.dart:489:39)
#5      StandardMessageCodec.readValue (package:flutter/src/services/message_codecs.dart:438:12)
#6      StandardMethodCodec.decodeMethodCall (package:flutter/src/services/message_codecs.dart:569:44)
#7      MethodChannel._handleAsMethodCall (package:flutter/src/services/platform_channel.dart:412:35)
#8      MethodChannel.setMethodCallHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:407:34)
#9      _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:377:35)
#10     _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:374:46)
#11     _invoke2.<anonymous closure> (dart:ui/hooks.dart:190:15)
#15     _invoke2 (dart:ui/hooks.dart:189:10)
#16     _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:42:5)
#17     _Channel.push (dart:ui/channel_buffers.dart:132:31)
#18     ChannelBuffers.push (dart:ui/channel_buffers.dart:329:17)
#19     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:583:22)
#20     _dispatchPlatformMessage (dart:ui/hooks.dart:89:31)
(elided 3 frames from dart:async)
════════════════════════════════════════════════════════════════════════════════════════════════════

```

To repro:
1. Launch Chrome and visit https://meet.google.com
2. Start a new meeting so the camera is active
3. In the demo app click the "Create camera" button.

packages/camera/camera_windows/windows/device_info.h
```diff
@@ -0,0 +1,25 @@
...
+#include <string>
+
+namespace camera_windows {
+
+struct CaptureDeviceInfo {
+  std::string display_name;
+  std::string device_id;
+};
+
+std::string GetUniqueDeviceName(std::unique_ptr<CaptureDeviceInfo> device_info);
```
cbracken, line 18: Since this is just getting a device name and not transferring ownership or consuming the unique_ptr, consider just passing a const ref to CaptureDeviceInfo.

Alternatively, I wonder if it might make sense to make CaptureDeviceInfo a class and make this into a getter on it?

cbracken (commented):
Getting closer to the end!

packages/camera/camera_windows/windows/capture_engine_listener.h
```diff
@@ -0,0 +1,62 @@
...
+  // Updates texture buffer
+  virtual bool UpdateBuffer(uint8_t* data, uint32_t new_length) = 0;
+
+  // Handles capture timestamps updates.
+  // Used to stop timed recordings when recorded time is exceeded.
+  virtual void UpdateCaptureTime(uint64_t capture_time) = 0;
+};
+
+class CaptureEngineListener : public IMFCaptureEngineOnSampleCallback,
+                              public IMFCaptureEngineOnEventCallback {
```
cbracken, line 33: Add class documentation. Since it looks like null observers are allowable, it's definitely worth touching on that usecase in the doc comment.

packages/camera/camera_windows/windows/camera_plugin.cpp
```diff
@@ -0,0 +1,590 @@
...
+}
+
+// Builds datetime string from current time.
+// Used as part of the filenames for captured pictures and videos.
+std::string GetCurrentTimeString() {
+  std::chrono::system_clock::duration now =
+      std::chrono::system_clock::now().time_since_epoch();
+
+  auto s = std::chrono::duration_cast<std::chrono::seconds>(now).count();
+  auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(now).count();
```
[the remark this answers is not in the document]
jokerttu: I changed this as it is a bit more readable as well.
I am not really sure is this a good format for file timestamp, but at least there is no collisions and it is sortable by date.

packages/camera/camera_windows/windows/capture_controller.h
```diff
@@ -0,0 +1,290 @@
...
+  bool media_foundation_started_ = false;
+  bool record_audio_ = false;
+  uint32_t preview_frame_width_ = 0;
+  uint32_t preview_frame_height_ = 0;
+  UINT dx_device_reset_token_ = 0;
+  std::unique_ptr<RecordHandler> record_handler_ = nullptr;
+  std::unique_ptr<PreviewHandler> preview_handler_ = nullptr;
+  std::unique_ptr<PhotoHandler> photo_handler_ = nullptr;
+  std::unique_ptr<TextureHandler> texture_handler_ = nullptr;
+  CaptureControllerListener* capture_controller_listener_ = nullptr;
```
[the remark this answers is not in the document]
jokerttu: capture_controller_listener_ is quaranteed to exists the whole lifetime of the capture_controller. but weak_ptr would be better and more safe patterns for sure.

packages/camera/camera_windows/windows/device_info.h
jokerttu: Converted to a class

packages/camera/camera_windows/windows/camera_plugin.cpp
```diff
@@ -0,0 +1,590 @@
...
+// Looks for |key| in |map|, returning the associated int64 value if it is
+// present, or a nullptr if not.
+const int64_t* Int64OrNull(const EncodableMap& map, const char* key) {
+  auto value = ValueOrNull(map, key);
+  if (!value) {
+    return nullptr;
+  }
+
+  if (std::holds_alternative<int32_t>(*value)) {
+    return (int64_t*)std::get_if<int32_t>(value);
```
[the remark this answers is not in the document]
jokerttu: as these were pointers, there was actually bug there. and as they were const pointers casting was not so straight forward either. as this function is not used so ofter, I just ended up copying the value instead of having pointer to it.

jokerttu:
@cbracken @stuartmorgan 
Areas pointed out are now checked and fixed.

At the moment I am a bit worried about memory leak on camera disposal.
This memory leak happens only if texture data is passed to flutter engine, not if textures are handled only on plugin code.
I was not able to build custom engine with debug symbols on time to figure out where this really happens, but I as the size of the leak is about half of the pixel buffer size, it might be packed opengl pixelbuffer surface or something similar.
With bigger capture resolutions we are talking about few megabytes per camera disposal.
With memory profiler it seems that leak is happening on `flutter_windows.dll`.
Or maybe `texture_registrar_->UnregisterTexture` is called on wrong time. It has mutex lock inside, so I don't think this is an issue either.

packages/camera/camera_windows/example/lib/main.dart
```diff
@@ -0,0 +1,460 @@
...
+      _cameraClosingStreamSubscription = CameraPlatform.instance
+          .onCameraClosing(cameraId)
+          .listen(_onCameraClosing);
+
+      unawaited(CameraPlatform.instance
+          .onCameraInitialized(cameraId)
+          .first
+          .then((CameraInitializedEvent event) {
+        _initializeCompleter.complete(event);
+      }));
```
cbracken, line 117: You can avoid using a `Completer` here by doing:
```dart
Future initialized = CameraPlatform.instance
    .onCameraInitialized(cameraId)
    .first;

...

CameraInitializedEvent event = await initialized;
_previewSize = Size(event.previewWidth, event.previewHeight);
```

packages/camera/camera_windows/windows/texture_handler.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+// Marks texture frame available after buffer is updated.
+void TextureHandler::OnBufferUpdated() {
+  if (TextureRegistered()) {
+    texture_registrar_->MarkTextureFrameAvailable(texture_id_);
+  }
+}
+
+const FlutterDesktopPixelBuffer* TextureHandler::ConvertPixelBufferForFlutter(
+    size_t target_width, size_t target_height) {
```
cbracken, line 66: We're currently sizing the destination buffer to the size of the source data from the camera. One optimisation that might make sense down the road is to restrict the destination buffer to the smaller of the source data or the target buffer. 

Might be worth adding a TODO unless you can think of a reason why we wouldn't want to do that.

cbracken:
Alright a bit more on the error message bug. In this code in capture_controller.cpp where we fetch the error message:
```cc
     if (FAILED(event_hr)) {
       // Reads system error
       error = std::system_category().message(event_hr);
     }
```

The error message returned doesn't appear to be UTF-8 encoded. There's a [blog post](https://blogs.msmvps.com/gdicanio/2017/08/16/what-is-the-encoding-used-by-the-error_code-message-string/) that explains that the encoding depends on the system language -- in my case fr-CA. Here's what I see for the contents:

> La table de fichiers ma�tres mat�rielle n�a pas pu lancer une diffusion en continu en
raison d�un manque de ressources mat�rielles.

That should be formatted as:

> La table de fichiers maîtres matérielle n'a pas pu lancer une diffusion en continu en raison d'un manque de ressources matérielles.

Rough translation: System master file table could not create a stream due to a lack of system resources.

Instead of using `std::system_category().message(event_hr)` you can use [_com_error](https://docs.microsoft.com/en-us/cpp/cpp/com-error-class?redirectedfrom=MSDN&view=msvc-170) and use the following code:

```cc
     if (FAILED(event_hr)) {
       // Reads system error
       _com_error err(event_hr);
       error = Utf8FromUtf16(err.ErrorMessage());
     }
```

When I patched that in locally, it fixed the issue entirely for me and I see the error in the snackbar at the bottom:

image

packages/camera/camera_windows/windows/capture_controller.cpp
```diff
@@ -0,0 +1,860 @@
...
+    std::string error;
+
+    HRESULT event_hr;
+    if (FAILED(event->GetStatus(&event_hr))) {
+      return;
+    }
+
+    if (FAILED(event_hr)) {
+      // Reads system error
+      error = std::system_category().message(event_hr);
```
cbracken, line 670: Replace this line with:
```cc
_com_error err(event_hr);
error = Utf8FromUtf16(err.ErrorMessage());
```

You'll need to add this header at the top:
```cc
#include <comdef.h>
```

cbracken (commented):
Overall LGTM modulo comments added.

I wasn't able to narrow down completely the source of the memory leak you mention. I see between a 300kB and 2MB difference between creating the camera / registering the texture and disposing the camera / releasing the texture. 

I haven't narrowed down which objects are involved and whether they're native objects or part of the dart heap.

packages/camera/camera_windows/windows/capture_controller.cpp
jokerttu: Oh, yes, didn't thought about different languages at all. These errors should always be printed in english for now.
I would prefer having set of common camera error codes and messages across all platforms, at the moment there is no 
such list of error codes. With these users need to write custom error handling for each platform they are implementing their app. At the moment errors from all platforms might produce different error codes for a same problem. This might mean breaking change for all platforms though.
Currelty these errors are sen to ui with capture_error code.

jokerttu: Fixed now as suggested.

packages/camera/camera_windows/example/lib/main.dart
jokerttu: Yes, this is much better approach... 
This should be changed at camera plugin as well: [reference](https://github.com/flutter/plugins/blob/main/packages/camera/camera/lib/src/camera_controller.dart#L280-L325)
Fixed and working.

packages/camera/camera_windows/windows/texture_handler.cpp
jokerttu: I was not keen on writing software scaling for this, and just passed all pixels through as original capture size.
But now when I hacked and tested this a bit, it is possible to set target framesize to the preview_sink and let the OS do the scaling for us, this would optimize the amount of pixels processed. 
This would need a bit more re-factoring and I won't implement this on this timeframe, as being a bit risky move.
Users can already optimize the image size by giving proper resolution preset on camera initialization.
If this target scaling is implemented, given resolution preset scale should not be exceeded, as it might affect the capture framerate.

jokerttu: I wrote a TODO line to the code

packages/camera/camera_windows/example/pubspec.yaml
```diff
@@ -0,0 +1,29 @@
...
+    path: ../
+  flutter:
+    sdk: flutter
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  integration_test:
+    sdk: flutter
+  pedantic: ^1.10.0
```
stuartmorgan, line 26: Nit that I missed before: This should be removed. It's only needed in plugins that are still using legacy analysis options.

stuartmorgan (changes requested):
I did another pass over the metadata and Dart and skimmed the C++ headers. I found a few things that I'd missed before; mostly they are very minor but the main things are:
- The Dart code not having unit tests
- The C++ code not having class comments

Both are maintainability issues, so definitely need to be addressed.

stuartmorgan:
Having also tried out the example: something we should file an issue to track for follow-up if it's not already done is preview mirroring. This has been an issue on mobile with front-facing cameras, and on desktop that's going to be the normal case, so we'll want to handle it.

packages/camera/camera_windows/windows/texture_handler.h
```diff
@@ -0,0 +1,89 @@
...
+  uint32_t bytes_per_pixel_ = 4;
+  uint32_t source_buffer_size_ = 0;
+  uint32_t preview_frame_width_ = 0;
+  uint32_t preview_frame_height_ = 0;
+
+  std::vector<uint8_t> source_buffer_;
+  std::vector<uint8_t> dest_buffer_;
+  std::unique_ptr<flutter::TextureVariant> texture_;
+  std::unique_ptr<FlutterDesktopPixelBuffer> flutter_desktop_pixel_buffer_ =
+      nullptr;
```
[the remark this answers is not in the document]
jokerttu: True, this is last minute change, and missed this... I'll fix this.

jokerttu:
@stuartmorgan mirror mode is actually already supported (by software flip) but by default set to `false`.
It could be option on camera initialization but its not supported by the camera_platform_interface so its not really usable attribute.
We should detect the camera facing, this is possible by reading registry or via api for windows10+ devices and use mirror mode as default for front-facing cameras.
Or we can set the default to true, and have mirrored camera by default for now. Not sure for what people are going to use this for, so this is a bit weird to put it as mirror by default.

packages/camera/camera_windows/windows/camera.cpp
```diff
@@ -0,0 +1,264 @@
...
+#include "camera.h"
+
+namespace camera_windows {
+using flutter::EncodableList;
+using flutter::EncodableMap;
+using flutter::EncodableValue;
+
+// Camera channel events.
+constexpr char kCameraMethodChannelBaseName[] =
+    "flutter.io/cameraPlugin/camera";
```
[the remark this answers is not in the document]
jokerttu: It uses the same naming as [here](https://github.com/flutter/plugins/blob/main/packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart#L111).
What is the correct channel name for this?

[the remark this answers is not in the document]
jokerttu: this is a bit different having camera at the ending... it is for prefix camera specific channels... named now as:
"plugins.flutter.io/camera_windows/camera"

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,423 @@
...
+/// An implementation of [CameraPlatform] for Windows.
+class CameraWindows extends CameraPlatform {
+  /// Registers the Windows implementation of CameraPlatform.
+  static void registerWith() {
+    CameraPlatform.instance = CameraWindows();
+  }
+
+  /// The method channel used to interact with the native platform.
+  final MethodChannel _pluginChannel =
+      const MethodChannel('plugins.flutter.io/camera');
```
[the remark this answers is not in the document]
jokerttu: changed to plugins.flutter.io/camera_windows

packages/camera/camera_windows/lib/camera_windows.dart
```diff
@@ -0,0 +1,423 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
```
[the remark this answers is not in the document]
jokerttu: @stuartmorgan tests are done basing on `method_channel_camera_test.dart`

packages/camera/camera_windows/windows/capture_controller.h
```diff
@@ -0,0 +1,277 @@
...
+  kVeryHigh,
+  // 2160p (4096x2160)
+  kUltraHigh,
+  // The highest resolution available.
+  kMax,
+};
+
+enum class CaptureEngineState { kNotInitialized, kInitializing, kInitialized };
+
+class VideoCaptureDeviceEnumerator {
```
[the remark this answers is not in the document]
jokerttu: @stuartmorgan done

cbracken:
Overall LGTM modulo a few minor fixes that I've prepared [in a branch](https://github.com/cbracken/flutter_plugins/commits/camera_windows) to make them easier to patch in.

That branch includes minor fixes for three categories of issues:
1. Adds/tweaks documentation comments.
2. Makes some C++ methods const, where possible.
3. Switches a few C++ string parameters to pass by const reference, where they were previously passing by value.

You should be able to patch them in via:
```
git remote add cbracken git@github.com:cbracken/flutter_plugins.git
git fetch cbracken
git cherry-pick 32880a3e25e6cfbc9b28e2e2d7d847c558bf6bf3
git cherry-pick c559f582ccecd5bc4c2abc4b258ef09de6e51cd1
git cherry-pick d4804e9d53813d6c7ce605eb972cacdd77ea79af
```

Apologies if the above is old-hat to you, it's always hard to tell how much git-related pain any given contributor has suffered through in the past :) 

Once those commits are merged, I think we're LGTM!

jokerttu:
@cbracken thanks, good tweaks.
Those are now cherry picked and committed.

stuartmorgan (approved):
Changes LGTM (I didn't do a full re-review)

cbracken (approved):
LGTM stamp from a Japanese personal seal

HosseinMohebbikhah:
Hi,
I don't know about Error "The video recording device is no longer present." in this project.
Please help me.

PR #4643, opened 2022-01-06
Update all plugins that use `verifyToken` to use `verify` instead

collinjackson:
#4640 (reviewed by @stuartmorgan) updated `plugin_plugin_interface` to 2.1.0, soft-deprecating `verifyToken` and introducing its replacement, `verify`.

Now that 2.1.0 has been [published](https://pub.dev/packages/plugin_platform_interface/versions/2.1.0), this PR updates all of the first-party plugins that used `verifyToken` to use `verify` and depend on `^2.1.0`:
- camera
- file_selector
- google_maps_flutter
- image_picker
- in_app_purchase
- path_provider
- quick_actions
- url_launcher
- video_player
- webview_flutter

It is a follow-on fix for https://github.com/flutter/flutter/issues/96178

## Test plan

This PR has no new tests because it's not possible to test; there was no semantic change to the above plugins. They are not using `const Object()` as their token so a test cannot cause the new `AssertionError` to be thrown. I believe this means @Hixie has to sign off on it and I've posted in #hackers-ecosystem about that.

However, this PR does contain new tests in the sense that there are new tests in https://github.com/flutter/plugins/pull/4640 and this is a follow-on PR that could only be landed after that one did because of the `plugin_platform_interface` dependency. In some sense this repo is a giant test case and we're testing https://github.com/flutter/plugins/pull/4640 by creating lots of sample apps that depend on it.

## Pre-launch Checklist

- [x] I signed the [CLA].
- [too many plugins to list] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [need approval] I added new tests to check the change I am making, or this PR is [test-exempt].

### DIFF
--- a/packages/camera/camera_platform_interface/pubspec.yaml
+++ b/packages/camera/camera_platform_interface/pubspec.yaml
@@ -4,7 +4,7 @@ repository: https://github.com/flutter/plugins/tree/master/packages/camera/camer
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+camera%22
 # NOTE: We strongly prefer non-breaking changes, even at the expense of a
 # less-clean API. See https://flutter.dev/go/platform-interface-breaking-changes
-version: 2.1.2
+version: 2.1.3
 
 environment:
   sdk: '>=2.12.0 <3.0.0'
@@ -15,7 +15,7 @@ dependencies:
   flutter:
     sdk: flutter
   meta: ^1.3.0
-  plugin_platform_interface: ^2.0.0
+  plugin_platform_interface: ^2.1.0
   stream_transform: ^2.0.0
 
 dev_dependencies:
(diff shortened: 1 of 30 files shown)


packages/camera/camera_platform_interface/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
   flutter:
     sdk: flutter
   meta: ^1.3.0
-  plugin_platform_interface: ^2.0.0
+  plugin_platform_interface: '>=2.1.0 <4.0.0'
```
stuartmorgan, line 18: Since the final version of the platform interface chance doesn't require any breaking changes (the deprecation after this all lands is just a minor bump per semver and standard Dart package practice), these should all be `^2.1.0` instead.

stuartmorgan (changes requested):
LG other than version issue

packages/camera/camera_platform_interface/pubspec.yaml
collinjackson: That is what I had in a previous version of this PR, and I've reverted back to it.

stuartmorgan (approved):
LGTM. Thanks for doing all this cleanup!

Hixie:
test-exempt: code refactor with no semantic change

ditman:
Thanks for the cleanup @collinjackson! Long time no see!

PR #4648, opened 2022-01-06
[url_launcher] Fix minor memory leak in Linux url-launcher tests

robert-ancell:
Fix minor memory leak in Linux url-launcher tests

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan (commented):
Oops! Thanks for the fix, LG other than needing a CHAGELOG update (using `## NEXT`) to make `publishable` happy.

robert-ancell:
@stuartmorgan I updated the description to avoid the need for a CHANGELOG entry - should I add one for a change like this?

stuartmorgan:
The policy we'd settled on for now is to generally require a CHANGELOG even for non-production changes because it's convenient for us to see what changed (especially since occasionally we make changes we think aren't production-affecting but were wrong, although obviously that wouldn't apply here). We may revisit if it's annoying after living with it for a while.

I'm fine with not having one here, but you'll need to push something to get `publish` green; rerunning unfortunately doesn't work (I guess Cirrus is caching the description it had or something?)

stuartmorgan (approved):
LGTM

PR #4655, opened 2022-01-07
Revert "Swap the mirroring order."

godofredoc:
Reverts flutter/plugins#4650

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -6,7 +6,7 @@
 on:
   push:
     branches:
-      - 'master'
+      - 'main'
 
 jobs:
   mirror_job:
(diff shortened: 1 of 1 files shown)


stuartmorgan:
I looked at the implementation of the mirror script a bit, and it has:
```
            result = await octokit.git.updateRef({
                ...github.context.repo,
                ref: `heads/${destBranchName}`,
                sha: sourceBranchSha,
                force: true,
            });
```

I haven't looked at the internals of that, but the `force: true` makes me think we don't want to do this. I'm concerned that this will destroy `main`.

stuartmorgan:
If it is indeed doing a force-push to the target branch, then all we need to do is land https://github.com/flutter/plugins/pull/4654 and it'll stomp `master`.

stuartmorgan:
> If it is indeed doing a force-push to the target branch, then all we need to do is land #4654 and it'll stomp `master`.

This was the case, so this PR isn't needed.

PR #4661, opened 2022-01-11
[camera]call engine API in main thread to fix a crash

hellohuanlin:
This PR is part of a larger project to fix the thread model for iOS camera plugin. The formal proposal can be found [here](https://docs.google.com/document/d/1Znjp-BduXIb1Ix3JOE1YKyJi1Z0d6rB-2q6Lt7j8xSo/edit#)

The engine APIs used in camera plugin are required to be run on platform thread. Some of these APIs have explicit asserts (e.g. MethodChannel) and some have warning instructions in documentation (e.g. TextureRegistry). However, in camera plugin we are currently calling these APIs in background thread, causing the crash.

There's already a FLTThreadSafeFlutterResult wrapper to invoke FlutterResult in main thread. This PR created similar "thread safe wrappers" for EventChannel, MethodChannel and TextureRegistry.

## Issues

flutter/flutter#94723
flutter/flutter#52578

## Pre-launch Checklist

- [x] I signed the [CLA].
- [] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].

### DIFF
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.9.4+6
 
-* Fixed a crash in iOS when using image stream due to calling Flutter engine API on non-main thread. 
+* Fixes a crash in iOS when using image stream due to calling Flutter engine API on non-main thread. 
 
 ## 0.9.4+5
 
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,4 +1,8 @@
-##  0.9.4+5
+## 0.9.4+6
+
+* Fixed a crash in iOS when using image stream due to calling Flutter engine API on non-main thread. 
+
+## 0.9.4+5
 
 * Fixes bug where calling a method after the camera was closed resulted in a Java `IllegalStateException` exception.
 * Fixes integration tests.
--- a/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
+++ b/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
@@ -7,57 +7,56 @@
 #import <OCMock/OCMock.h>
 
 @interface ThreadSafeEventChannelTests : XCTestCase
-@property(nonatomic, strong) FlutterEventChannel *mockEventChannel;
-@property(nonatomic, strong) FLTThreadSafeEventChannel *threadSafeEventChannel;
 @end
 
 @implementation ThreadSafeEventChannelTests
 
-- (void)setUp {
-  [super setUp];
-  _mockEventChannel = OCMClassMock([FlutterEventChannel class]);
-  _threadSafeEventChannel =
-      [[FLTThreadSafeEventChannel alloc] initWithEventChannel:_mockEventChannel];
-}
-
 - (void)testSetStreamHandler_shouldStayOnMainThreadIfCalledFromMainThread {
+  FlutterEventChannel *mockEventChannel = OCMClassMock([FlutterEventChannel class]);
+  FLTThreadSafeEventChannel *threadSafeEventChannel =
+      [[FLTThreadSafeEventChannel alloc] initWithEventChannel:mockEventChannel];
+
   XCTestExpectation *mainThreadExpectation = [[XCTestExpectation alloc]
       initWithDescription:@"setStreamHandler must be called on the main thread"];
   XCTestExpectation *mainThreadCompletionExpectation = [[XCTestExpectation alloc]
       initWithDescription:@"setStreamHandler's completion block must be called on the main thread"];
-  OCMStub([self.mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
+  OCMStub([mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
     if (NSThread.isMainThread) {
       [mainThreadExpectation fulfill];
     }
   });
 
-  [self.threadSafeEventChannel setStreamHandler:nil
-                                     completion:^{
-                                       if (NSThread.isMainThread) {
-                                         [mainThreadCompletionExpectation fulfill];
-                                       }
-                                     }];
+  [threadSafeEventChannel setStreamHandler:nil
+                                completion:^{
+                                  if (NSThread.isMainThread) {
+                                    [mainThreadCompletionExpectation fulfill];
+                                  }
+                                }];
   [self waitForExpectations:@[ mainThreadExpectation, mainThreadCompletionExpectation ] timeout:1];
 }
 
 - (void)testSetStreamHandler_shouldDispatchToMainThreadIfCalledFromBackgroundThread {
+  FlutterEventChannel *mockEventChannel = OCMClassMock([FlutterEventChannel class]);
+  FLTThreadSafeEventChannel *threadSafeEventChannel =
+      [[FLTThreadSafeEventChannel alloc] initWithEventChannel:mockEventChannel];
+
   XCTestExpectation *mainThreadExpectation = [[XCTestExpectation alloc]
       initWithDescription:@"setStreamHandler must be called on the main thread"];
   XCTestExpectation *mainThreadCompletionExpectation = [[XCTestExpectation alloc]
       initWithDescription:@"setStreamHandler's completion block must be called on the main thread"];
-  OCMStub([self.mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
+  OCMStub([mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
     if (NSThread.isMainThread) {
       [mainThreadExpectation fulfill];
     }
   });
 
   dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-    [self.threadSafeEventChannel setStreamHandler:nil
-                                       completion:^{
-                                         if (NSThread.isMainThread) {
-                                           [mainThreadCompletionExpectation fulfill];
-                                         }
-                                       }];
+    [threadSafeEventChannel setStreamHandler:nil
+                                  completion:^{
+                                    if (NSThread.isMainThread) {
+                                      [mainThreadCompletionExpectation fulfill];
+                                    }
+                                  }];
   });
   [self waitForExpectations:@[
     mainThreadExpectation,
--- a/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
+++ b/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
@@ -1,0 +1,63 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import camera;
+@import XCTest;
+#import <OCMock/OCMock.h>
+
+@interface ThreadSafeEventChannelTests : XCTestCase
+@property(nonatomic, strong) FLTThreadSafeEventChannel *channel;
+@property(nonatomic, strong) XCTestExpectation *mainThreadExpectation;
+@property(nonatomic, strong) XCTestExpectation *mainThreadCompletionExpectation;
+@end
+
+@implementation ThreadSafeEventChannelTests
+
+- (void)setUp {
+  [super setUp];
+  id mockEventChannel = OCMClassMock([FlutterEventChannel class]);
+
+  _mainThreadExpectation = [[XCTestExpectation alloc]
+      initWithDescription:@"setStreamHandler must be called on the main thread"];
+  _mainThreadCompletionExpectation = [[XCTestExpectation alloc]
+      initWithDescription:@"setStreamHandler's completion block must be called on the main thread"];
+  _channel = [[FLTThreadSafeEventChannel alloc] initWithEventChannel:mockEventChannel];
+
+  OCMStub([mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
+    if (NSThread.isMainThread) {
+      [self.mainThreadExpectation fulfill];
+    }
+  });
+}
+
+- (void)testSetStreamHandler_shouldStayOnMainThreadIfCalledFromMainThread {
+  __weak XCTestExpectation *mainThreadCompletionExpectation = self.mainThreadCompletionExpectation;
+  [self.channel setStreamHandler:nil
+                      completion:^{
+                        if (NSThread.isMainThread) {
+                          [mainThreadCompletionExpectation fulfill];
+                        }
+                      }];
+  [self waitForExpectations:@[ self.mainThreadExpectation, self.mainThreadCompletionExpectation ]
+                    timeout:1];
+}
+
+- (void)testSetStreamHandler_shouldDispatchToMainThreadIfCalledFromBackgroundThread {
+  __weak XCTestExpectation *mainThreadCompletionExpectation = self.mainThreadCompletionExpectation;
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    [self.channel setStreamHandler:nil
+                        completion:^{
+                          if (NSThread.isMainThread) {
+                            [mainThreadCompletionExpectation fulfill];
+                          }
+                        }];
+  });
+  [self waitForExpectations:@[
+    self.mainThreadExpectation,
+    self.mainThreadCompletionExpectation,
+  ]
+                    timeout:1];
+}
+
+@end
--- a/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
+++ b/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
@@ -16,10 +16,11 @@
   FLTThreadSafeEventChannel *threadSafeEventChannel =
       [[FLTThreadSafeEventChannel alloc] initWithEventChannel:mockEventChannel];
 
-  XCTestExpectation *mainThreadExpectation = [[XCTestExpectation alloc]
-      initWithDescription:@"setStreamHandler must be called on the main thread"];
-  XCTestExpectation *mainThreadCompletionExpectation = [[XCTestExpectation alloc]
-      initWithDescription:@"setStreamHandler's completion block must be called on the main thread"];
+  XCTestExpectation *mainThreadExpectation =
+      [self expectationWithDescription:@"setStreamHandler must be called on the main thread"];
+  XCTestExpectation *mainThreadCompletionExpectation =
+      [self expectationWithDescription:
+                @"setStreamHandler's completion block must be called on the main thread"];
   OCMStub([mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
     if (NSThread.isMainThread) {
       [mainThreadExpectation fulfill];
@@ -32,7 +33,7 @@
                                     [mainThreadCompletionExpectation fulfill];
                                   }
                                 }];
-  [self waitForExpectations:@[ mainThreadExpectation, mainThreadCompletionExpectation ] timeout:1];
+  [self waitForExpectationsWithTimeout:1 handler:nil];
 }
 
 - (void)testSetStreamHandler_shouldDispatchToMainThreadIfCalledFromBackgroundThread {
@@ -40,10 +41,11 @@
   FLTThreadSafeEventChannel *threadSafeEventChannel =
       [[FLTThreadSafeEventChannel alloc] initWithEventChannel:mockEventChannel];
 
-  XCTestExpectation *mainThreadExpectation = [[XCTestExpectation alloc]
-      initWithDescription:@"setStreamHandler must be called on the main thread"];
-  XCTestExpectation *mainThreadCompletionExpectation = [[XCTestExpectation alloc]
-      initWithDescription:@"setStreamHandler's completion block must be called on the main thread"];
+  XCTestExpectation *mainThreadExpectation =
+      [self expectationWithDescription:@"setStreamHandler must be called on the main thread"];
+  XCTestExpectation *mainThreadCompletionExpectation =
+      [self expectationWithDescription:
+                @"setStreamHandler's completion block must be called on the main thread"];
   OCMStub([mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
     if (NSThread.isMainThread) {
       [mainThreadExpectation fulfill];
@@ -58,11 +60,7 @@
                                     }
                                   }];
   });
-  [self waitForExpectations:@[
-    mainThreadExpectation,
-    mainThreadCompletionExpectation,
-  ]
-                    timeout:1];
+  [self waitForExpectationsWithTimeout:1 handler:nil];
 }
 
 @end
--- a/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
+++ b/packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
@@ -7,55 +7,61 @@
 #import <OCMock/OCMock.h>
 
 @interface ThreadSafeEventChannelTests : XCTestCase
-@property(nonatomic, strong) FLTThreadSafeEventChannel *channel;
-@property(nonatomic, strong) XCTestExpectation *mainThreadExpectation;
-@property(nonatomic, strong) XCTestExpectation *mainThreadCompletionExpectation;
+@property(nonatomic, strong) FlutterEventChannel *mockEventChannel;
+@property(nonatomic, strong) FLTThreadSafeEventChannel *threadSafeEventChannel;
 @end
 
 @implementation ThreadSafeEventChannelTests
 
 - (void)setUp {
   [super setUp];
-  id mockEventChannel = OCMClassMock([FlutterEventChannel class]);
-
-  _mainThreadExpectation = [[XCTestExpectation alloc]
-      initWithDescription:@"setStreamHandler must be called on the main thread"];
-  _mainThreadCompletionExpectation = [[XCTestExpectation alloc]
-      initWithDescription:@"setStreamHandler's completion block must be called on the main thread"];
-  _channel = [[FLTThreadSafeEventChannel alloc] initWithEventChannel:mockEventChannel];
-
-  OCMStub([mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
-    if (NSThread.isMainThread) {
-      [self.mainThreadExpectation fulfill];
-    }
-  });
+  _mockEventChannel = OCMClassMock([FlutterEventChannel class]);
+  _threadSafeEventChannel =
+      [[FLTThreadSafeEventChannel alloc] initWithEventChannel:_mockEventChannel];
 }
 
 - (void)testSetStreamHandler_shouldStayOnMainThreadIfCalledFromMainThread {
-  __weak XCTestExpectation *mainThreadCompletionExpectation = self.mainThreadCompletionExpectation;
-  [self.channel setStreamHandler:nil
-                      completion:^{
-                        if (NSThread.isMainThread) {
-                          [mainThreadCompletionExpectation fulfill];
-                        }
-                      }];
-  [self waitForExpectations:@[ self.mainThreadExpectation, self.mainThreadCompletionExpectation ]
-                    timeout:1];
+  XCTestExpectation *mainThreadExpectation = [[XCTestExpectation alloc]
+      initWithDescription:@"setStreamHandler must be called on the main thread"];
+  XCTestExpectation *mainThreadCompletionExpectation = [[XCTestExpectation alloc]
+      initWithDescription:@"setStreamHandler's completion block must be called on the main thread"];
+  OCMStub([self.mockEventChannel setStreamHandler:[OCMArg any]]).andDo(^(NSInvocation *invocation) {
+    if (NSThread.isMainThread) {
+      [mainThreadExpectation fulfill];
+    }
+  });
+
+  [self.threadSafeEventChannel setStreamHandler:nil
+                                     completion:^{
+                                       if (NSThread.isMainThread) {
+                                         [mainThreadCompletionExpectation fulfill];
+                                       }
+                                     }];
+  [self waitForExpectations:@[ mainThreadExpectation, mainThreadCompletionExpectation ] timeout:1];
 }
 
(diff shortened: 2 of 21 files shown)


packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.h
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Registers a `FlutterTexture` for usage in Flutter and returns an id that can be used to reference
+ * that texture when calling into Flutter with channels. Textures must be registered on the
+ * main thread.
+ *
+ * On success the completion block completes with the pointer to the registered texture, else with
+ * 0.
+ */
+- (void)registerTexture:(NSObject<FlutterTexture> *)texture
+             completion:(void (^)(int64_t))completion;
```
hellohuanlin, line 30: this is updated to use `dispatch_async`

packages/camera/camera/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-##  0.9.4+5
+## 0.9.4+6
```
hellohuanlin, line 1: version bump

packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
```diff
@@ -0,0 +1,63 @@
...
+// found in the LICENSE file.
+
+@import camera;
+@import XCTest;
+#import <OCMock/OCMock.h>
+
+@interface ThreadSafeEventChannelTests : XCTestCase
+@property(nonatomic, strong) FLTThreadSafeEventChannel *channel;
+@property(nonatomic, strong) XCTestExpectation *mainThreadExpectation;
+@property(nonatomic, strong) XCTestExpectation *mainThreadCompletionExpectation;
```
hellohuanlin, line 12: changed from ivars to properties

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -1115,11 +1118,16 @@ - (void)startImageStreamWithMessenger:(NSObject<FlutterBinaryMessenger> *)messen
...
     _imageStreamHandler = [[FLTImageStreamHandler alloc] init];
-    [eventChannel setStreamHandler:_imageStreamHandler];
-
-    _isStreamingImages = YES;
+    [threadSafeEventChannel setStreamHandler:_imageStreamHandler
+                                  completion:^{
+                                    dispatch_async(self->_dispatchQueue, ^{
+                                      self->_isStreamingImages = YES;
+                                    });
+                                  }];
```
hellohuanlin, line 1130: added completion block

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -1417,11 +1427,13 @@ - (void)handleMethodCallAsync:(FlutterMethodCall *)call
...
       _camera = cam;
-      [result sendSuccessWithData:@{
-        @"cameraId" : @(textureId),
-      }];
+      [self.registry registerTexture:cam
+                          completion:^(int64_t textureId) {
+                            [result sendSuccessWithData:@{
+                              @"cameraId" : @(textureId),
+                            }];
+                          }];
```
hellohuanlin, line 1436: added completion block

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.h
hellohuanlin, line 30: added completion block

packages/camera/camera/ios/Classes/FLTThreadSafeEventChannel.h
```diff
@@ -0,0 +1,29 @@
...
+ * @param channel The FlutterEventChannel object to be wrapped.
+ */
+- (instancetype)initWithEventChannel:(FlutterEventChannel *)channel;
+
+/*
+ * Registers a handler for stream setup requests from the Flutter side on main thread.
+ # Completion block runs on main thread.
+ */
+- (void)setStreamHandler:(nullable NSObject<FlutterStreamHandler> *)handler
+              completion:(void (^)(void))completion;
```
hellohuanlin, line 25: added completion block

hellohuanlin:
Hi @jmagman @cyanglaz and @stuartmorgan, I re-created this PR from last year. 

I added comments for what's updated from the last PR. The biggest change is to use `completion` block which both @cyanglaz and @jmagman suggested.  

PTAL again, thank you!

packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
stuartmorgan: Please don't make expectations a stateful part of the fixture; that makes tests much harder to understand.

I don't see a need for these properties or `setUp` to exist at all. It would be preferable for all of the code to be in each test.

packages/camera/camera/example/ios/RunnerTests/ThreadSafeMethodChannelTests.m
```diff
@@ -0,0 +1,45 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import camera;
+@import XCTest;
+#import <OCMock/OCMock.h>
+
+@interface ThreadSafeMethodChannelTests : XCTestCase
+@property(nonatomic, strong) FLTThreadSafeMethodChannel *channel;
+@property(nonatomic, strong) XCTestExpectation *mainThreadExpectation;
```
stuartmorgan, line 11: Same comments here.

packages/camera/camera/example/ios/RunnerTests/ThreadSafeTextureRegistryTests.m
```diff
@@ -0,0 +1,97 @@
...
+@import camera;
+@import XCTest;
+#import <OCMock/OCMock.h>
+
+@interface ThreadSafeTextureRegistryTests : XCTestCase
+@property(nonatomic, strong) FLTThreadSafeTextureRegistry *registry;
+@property(nonatomic, strong) XCTestExpectation *registerTextureExpectation;
+@property(nonatomic, strong) XCTestExpectation *registerTextureCompletionExpectation;
+@property(nonatomic, strong) XCTestExpectation *unregisterTextureExpectation;
+@property(nonatomic, strong) XCTestExpectation *textureFrameAvailableExpectation;
```
stuartmorgan, line 14: And here.

stuartmorgan (changes requested):

packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
hellohuanlin: @stuartmorgan Very interesting. I did a bit search and found out that we generally prefer declarative coding style in unit test, where readability is more important than DRY principle. Today I learned.

packages/camera/camera/ios/Classes/FLTThreadSafeEventChannel.h
```diff
@@ -0,0 +1,29 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Flutter/Flutter.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Wrapper for FlutterEventChannel that always sends events on the main thread
```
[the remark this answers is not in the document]
hellohuanlin: Yes these objects are thread safe, though it'd be a bit odd to me to add a duplicate comment saying a "FLTThreadSafeFoo" is thread safe. The class name commented itself.

packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
hellohuanlin: I think I still want the `setUp` function to setup the test subject (the wrapper and the mock channel dependency).

packages/camera/camera/ios/Classes/FLTThreadSafeEventChannel.h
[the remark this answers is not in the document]
stuartmorgan: But the description just says that they always call things on the main thread. The fact that its purpose is to make *other code* threadsafe leaves a lot of ambiguity here.

If you think the current naming is completely self-explanatory: what would you call a class whose job is to always funnel `EventChannel` calls to the main thread, but was itself thread-hostile for some reason?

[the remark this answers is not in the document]
hellohuanlin: I'm thinking `A thread safe wrapper for FlutterEventChannel that can be called from any threads`. The "dispatching to main thread" part is more like implementation details that the consumer does not need to know.  What do you think?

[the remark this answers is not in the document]
stuartmorgan: I don't think it's really an implementation detail; it's the reason for the existence of the class. If at some point the need to do that were to change (e.g., exactly what is happening with `FlutterResult` right now), the whole class could be removed.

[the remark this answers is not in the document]
hellohuanlin: To clarify, I think "dispatching to main" is the implementation detail that *the caller of the wrapper* should not care about. The caller only cares the wrapper API is thread-safe, and does not care about the fact that the underlying engine api must happen on main and the thread-safety is achieved by dispatching to main. 

In the future, when the original API itself becomes thread-safe, we can remove this wrapper. In that world, how the original API achieves thread-safety is also implementation detail that the caller should not care about. It only cares that the new API can be called from any thread. As to how this is achieved? By dispatching to a the same serial queue like what we do here? or by using other mechanisms like atomic CAS? or by other means? it doesn't matter.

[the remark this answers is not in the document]
hellohuanlin: In the current world, the consumer of the engine API is the wrapper *implementation*, and the consumer of the wrapper API is in `CameraPlugin.m`.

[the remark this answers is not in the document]
stuartmorgan: > To clarify, I think "dispatching to main" is the implementation detail that _the consumer of the wrapper_ should not care about.

I understand who we're speaking in reference to when we say that something is or is not an implementation detail.

> The consumer only cares the wrapper API is thread-safe, and does not care about the fact that the underlying engine api must happen on main and the thread-safety is achieved by dispatching to main.

I strongly disagree. The consumer is a plugin. A plugin is written against the engine APIs, and is responsible for ensuring that the APIs are used correctly, including calling main-thread-APIs on the main thread. This class is a tool for doing that, it's not a complete abstraction that eliminates the plugin code's knowledge of the Flutter engine.

**How** it dispatches to the main thread is an implementation detail. The fact that it does so is integral to the API contract of the class itself.

> In the future, when the original API itself becomes thread-safe, we can remove this wrapper. In that world, how the original API achieves thread-safety is also implementation detail that the consumer of the API should not care about.

The difference is that the engine is a complete API surface that sets threading requirements. This is a class that is **not part of that API** whose purpose is to make it easier for code to meet certain threading restrictions on an object **from that API surface**.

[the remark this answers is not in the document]
hellohuanlin: I think there are 2 parts discussed here - the wrapper, and the consumer of the wrapper. Only the wrapper's implementation cares about the knowledge that engine API must be run on main. This knowledge is abstracted away from the consumer of the wrapper. 

Yes the wrapper is not part of the engine, but it can still encapsulate implementation details and hide it from its consumers. Having the wrapper to be inside the same plugin directory as its consumers does not mean its consumers must know the internal of this wrapper class.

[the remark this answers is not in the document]
hellohuanlin: > I strongly disagree. The consumer is a plugin. 

Looks like miscommunication here. To clarify, the "consumer" quoted here is to refer to the "consumer of the wrapper" in the previous sentence, and not the "consumer of the engine API". 

So let me emphasize it again because it's very important: 
- The consumer of the engine API is the plugin code, or more precisely, the wrapper part of the plugin code.
- The consumer of the wrapper API is the `CameraPlugin.m` part of the plugin code. 

The wrapper and the consumer of the wrapper API are both under the same plugin directory, but it does not mean that the wrapper's implementation detail has to be exposed to the consumer of the wrapper. The wrapper encapsulated its internal implementation details of how to achieve the thread-safety.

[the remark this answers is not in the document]
stuartmorgan: > Looks like miscommunication here. To clarify, the "consumer" quoted here is to refer to the "consumer of the wrapper" in the previous sentence, and not the "consumer of the engine API".

I understand how this code works, and who the consumers of each API are. It would be helpful if you could stop assuming that any time I disagree with your conclusion I must not understand the code.

> The wrapper and the consumer of the wrapper API are both under the same plugin directory, but it does not mean that the wrapper's implementation detail has to be exposed to the consumer of the wrapper.

I understand the concept that directories and API surfaces are not the same thing.

> The wrapper encapsulated its internal implementation details of how to achieve the thread-safety.

I have already said I agree with this statement. Whether it uses GCD, operation queues, low-level posix primitives, etc., is an implementation detail that is, correctly, encapsulated.

> * Does the `CameraPlugin.m` part of the plugin know that engine API has to be run on the main thread?
>   No, `CameraPlugin.m` only cares that the wrapper is thread safe, so can be called from any threads.

I don't understand the reason for the assertion that the code that **creates the event channel** has no responsibility for knowing anything about the threading model of the object that it is creating and handing off to an object that is explicitly for multi-threaded use. And I don't agree with it for reasons that I've explained.

Specifically, I disagree with this characterization of the code:

> So let me emphasize it again because it's very important:
>
> * The consumer of the engine API is the plugin code, or more precisely, the wrapper part of the plugin code.

`CameraPlugin.m` is a direct consumer of engine APIs, including a direct consumer of `EventChannel` (when it creates it). It is not a consumer of the the specific calls that are wrapped, that's true, but it is deeply, integrally intertwined with the engine API.

Here's a plausible hypothetical for you: imagine that at some point `EventChannel` were extended to have an `eventChannelWithName:binaryMessenger:queue:`, whose semantics were that it was still not thread-safe, but that when initialized that way all calls to it would need to be made from that queue rather than the main queue. This first line of this code **in `CameraPlugin.m`**:
```
    FlutterEventChannel *eventChannel =
        [FlutterEventChannel eventChannelWithName:@"plugins.flutter.io/camera/imageStream"
                                  binaryMessenger:messenger];
    FLTThreadSafeEventChannel *threadSafeEventChannel =
        [[FLTThreadSafeEventChannel alloc] initWithEventChannel:eventChannel];
```

is changed to:

```
    FlutterEventChannel *eventChannel =
        [FlutterEventChannel eventChannelWithName:@"plugins.flutter.io/camera/imageStream"
                                  binaryMessenger:messenger
                                            queue:someOtherQueue];
```

The threading of `FLTThreadSafeEventChannel` is now completely broken by a change in its consumer. If this were a true abstraction of the threading model of the Flutter engine, that would be impossible.

It would be possible to build a complete abstraction layer between all the real logic of the camera plugin and the Flutter engine, but these utility classes **do not do that**.

> Can we break this encapsulation? Well I don't recommend it, but sure.

My point is that I do not agree that this level of encapsulation exists here in the first place. `CameraPlugin.m` is not a higher level of abstraction that is blissfully unaware that it is interacting with the Flutter engine, such that all of the Flutter engine's API requirements are irrelevant to it. It is the central interface with the engine APIs in general.

If I were new to this plugin but familiar with plugin threading, and reading `CameraPlugin.m` to understand how it works (e.g., to debug an issue that could be related to threading), and I saw that a class I know is not safe to use from other threads was being handed to another class that is clearly intended for use from other threads, it would be a red flag. I would then go look at that class's header. If it said that it guarantees that the object is always called from the main thread, then I would immediately understand what it was doing and go back to the main logic. If it did not, then I would need to take a detour into reading the implementation of that class to see if the plugin had been correctly designed in the first place.

> If you have any thoughts on what that line of comment should be, I'm more than happy to update them.

Most of this thread started from your question here:

> The "dispatching to main thread" part is more like implementation details that the consumer does not need to know. What do you think?

I've said what I think: that we shouldn't remove that part, because I don't think its an implementation detail that is irrelevant to the consuming code. But you don't seem to be happy to make that change (or rather, not make the change of removing it) as evidenced by the length of this thread.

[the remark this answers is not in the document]
hellohuanlin: > But you don't seem to be happy to make that change (or rather, not make the change of removing it) as evidenced by the length of this thread.

That's not the case at all. I am **extremely happy** to update the comment. It's just I wasn't clear what you wanted. I will take a look and digest your explanation above tomorrow. But just to avoid further churn, I will update it to - `A thread safe wrapper for FlutterEventChannel that can be called from any threads, by dispatching the underlying engine API call to the main thread. `.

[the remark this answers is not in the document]
hellohuanlin: Just sent a meeting invite for a quick chat.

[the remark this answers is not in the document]
hellohuanlin: Discussed offline. The key thing I overlooked is that, the `CameraPlugin.m` **is** a direct consumer of the engine API (e.g. creation of the thread-unsafe type to be wrapped). I was imagining a world where the `CameraPlugin.m` only interact with the wrapper, and the wrapper interact with the engine API. But that's not the case.

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.h
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Notifies Flutter that the content of the previously registered texture has been updated.
+ *
+ * This will trigger a call to `-[FlutterTexture copyPixelBuffer]` on the raster thread.
+ *
+ * Runs on main thread.
```
stuartmorgan, line 38: Same concerns here about threading documentation. This statement is actually wrong no matter how it's read:
- If it's read as descriptive, it's false: the method (which is what a declaration comment documents) runs on whatever thread it is called from, not the main thread. **Parts of its implementation** run on the main thread, but not the method itself.
- If it's read as prescriptive (i.e., that this must be called on the main thread), it defeats the purpose of the class.

I would change this whole declaration comment to:

"Notifies the Flutter engine on the main thread that the given texture has been updated."

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.h
```diff
@@ -0,0 +1,54 @@
...
+ * On success the completion block completes with the pointer to the registered texture, else with
+ * 0.
+ */
+- (void)registerTexture:(NSObject<FlutterTexture> *)texture
+             completion:(void (^)(int64_t))completion;
+
+/**
+ * Notifies Flutter that the content of the previously registered texture has been updated.
+ *
+ * This will trigger a call to `-[FlutterTexture copyPixelBuffer]` on the raster thread.
```
stuartmorgan, line 36: This code has no control over what the engine will do, so should not be documenting it.

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.h
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Runs on main thread.
+ */
+- (void)textureFrameAvailable:(int64_t)textureId;
+
+/**
+ * Unregisters a `FlutterTexture` that has previously regeistered with `registerTexture:`. Textures
+ * must be unregistered on the main thread.
```
stuartmorgan, line 44: Same here and below; I would suggest a similar rewording.

packages/camera/camera/ios/Classes/FLTThreadSafeFlutterResult.h
```diff
@@ -5,7 +5,8 @@
 #import <Flutter/Flutter.h>
 
 /**
- * Wrapper for FlutterResult  that always delivers the result on the main thread.
+ * A thread safe wrapper for FlutterResult that can be called from any thread, by dispatching its
+ * underlying engine APIs to the main thread.
```
stuartmorgan, line 9: Nit for all of these class comments: "by dispatching underlying engine calls to the main thread". These classes don't have engine APIs, they only have calls to the engine's APIs.

stuartmorgan (changes requested):

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -1115,11 +1118,16 @@ - (void)startImageStreamWithMessenger:(NSObject<FlutterBinaryMessenger> *)messen
...
         [FlutterEventChannel eventChannelWithName:@"plugins.flutter.io/camera/imageStream"
                                   binaryMessenger:messenger];
+    FLTThreadSafeEventChannel *threadSafeEventChannel =
+        [[FLTThreadSafeEventChannel alloc] initWithEventChannel:eventChannel];
 
     _imageStreamHandler = [[FLTImageStreamHandler alloc] init];
-    [eventChannel setStreamHandler:_imageStreamHandler];
-
-    _isStreamingImages = YES;
+    [threadSafeEventChannel setStreamHandler:_imageStreamHandler
```
stuartmorgan, line 1125: I missed this until it came up in a discussion with @gaaclarke : You're setting the stream handler safely now, but what about the calls to the event sink that's set up in the handler's callback? Are those already on the main thread?

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -654,6 +667,7 @@ - (void)captureOutput:(AVCaptureOutput *)output
 
         [planes addObject:planeBuffer];
       }
+      CVPixelBufferUnlockBaseAddress(pixelBuffer, kCVPixelBufferLock_ReadOnly);
```
hellohuanlin, line 670: Moved up this unlock call. Already done using `pixelBuffer` at this point

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -21,19 +21,31 @@ @interface FLTSavePhotoDelegate : NSObject <AVCapturePhotoCaptureDelegate>
...
+    self.eventSink = nil;
+  });
   return nil;
 }
 
 - (FlutterError *_Nullable)onListenWithArguments:(id _Nullable)arguments
                                        eventSink:(nonnull FlutterEventSink)events {
-  _eventSink = events;
+  dispatch_async(self.dispatchQueue, ^{
+    self.eventSink = events;
```
hellohuanlin, line 47: event sink needs to be invoked on the main queue, but the pointer `_eventSink` itself should only be accessed on the session queue.

packages/camera/camera/ios/Classes/CameraPlugin.m
hellohuanlin: Very good callout! I missed the event sink calls in my original proposal (Just updated it and tagged you).

packages/camera/camera/example/ios/RunnerTests/ThreadSafeEventChannelTests.m
```diff
@@ -0,0 +1,68 @@
...
+    [threadSafeEventChannel setStreamHandler:nil
+                                  completion:^{
+                                    if (NSThread.isMainThread) {
+                                      [mainThreadCompletionExpectation fulfill];
+                                    }
+                                  }];
+  });
+  [self waitForExpectations:@[
+    mainThreadExpectation,
+    mainThreadCompletionExpectation,
```
stuartmorgan, line 63: If you remove the comma here, will this all go on one line? The autoformatter output here is... not great.

packages/camera/camera/example/ios/RunnerTests/ThreadSafeTextureRegistryTests.m
```diff
@@ -0,0 +1,118 @@
...
+                                    }
+                                  }];
+  [threadSafeTextureRegistry textureFrameAvailable:0];
+  [threadSafeTextureRegistry unregisterTexture:0];
+  [self waitForExpectations:@[
+    registerTextureExpectation,
+    unregisterTextureExpectation,
+    textureFrameAvailableExpectation,
+    registerTextureCompletionExpectation,
+  ]
```
stuartmorgan, line 61: Since that won't work here, maybe pull this array out into a local variable.

packages/camera/camera/example/ios/RunnerTests/ThreadSafeTextureRegistryTests.m
```diff
@@ -0,0 +1,118 @@
...
+                                    }];
+    [threadSafeTextureRegistry textureFrameAvailable:0];
+    [threadSafeTextureRegistry unregisterTexture:0];
+  });
+  [self waitForExpectations:@[
+    registerTextureExpectation,
+    unregisterTextureExpectation,
+    textureFrameAvailableExpectation,
+    registerTextureCompletionExpectation,
+  ]
```
stuartmorgan, line 114: Same.

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -10,27 +10,42 @@
...
+#import "FLTThreadSafeMethodChannel.h"
+#import "FLTThreadSafeTextureRegistry.h"
 
 @interface FLTSavePhotoDelegate : NSObject <AVCapturePhotoCaptureDelegate>
 @property(readonly, nonatomic) NSString *path;
 @property(readonly, nonatomic) FLTThreadSafeFlutterResult *result;
 @end
 
 @interface FLTImageStreamHandler : NSObject <FlutterStreamHandler>
+@property(nonatomic, strong) dispatch_queue_t dispatchQueue;
```
stuartmorgan, line 24: This should have a declaration comment. (The existing code that doesn't have it is violating style.)

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -10,27 +10,42 @@
...
 @property FlutterEventSink eventSink;
```
stuartmorgan, line 25: Please annotate in a comment here what queue this is allowed to be accessed from.

stuartmorgan (approved):
LGTM with nits

packages/camera/camera/example/ios/RunnerTests/ThreadSafeMethodChannelTests.m
```diff
@@ -0,0 +1,54 @@
...
+
+@implementation ThreadSafeMethodChannelTests
+
+- (void)testInvokeMethod_shouldStayOnMainThreadIfCalledFromMainThread {
+  FlutterMethodChannel *mockMethodChannel = OCMClassMock([FlutterMethodChannel class]);
+  FLTThreadSafeMethodChannel *threadSafeMethodChannel =
+      [[FLTThreadSafeMethodChannel alloc] initWithMethodChannel:mockMethodChannel];
+
+  XCTestExpectation *mainThreadExpectation = [[XCTestExpectation alloc]
+      initWithDescription:@"invokeMethod must be called on the main thread"];
```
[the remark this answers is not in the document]
hellohuanlin: Nice. I didn't know about this API before. Looks handy.

packages/camera/camera/ios/Classes/FLTThreadSafeEventChannel.m
```diff
@@ -0,0 +1,35 @@
...
+
+- (void)setStreamHandler:(NSObject<FlutterStreamHandler> *)handler
+              completion:(void (^)(void))completion {
+  void (^block)(void) = ^{
+    [self.channel setStreamHandler:handler];
+    completion();
+  };
+
+  if (!NSThread.isMainThread) {
+    dispatch_async(dispatch_get_main_queue(), block);
```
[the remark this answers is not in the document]
hellohuanlin: Oh I used `dispatch_sync` last time because I intended to just swap the call. But since I already read through the whole plugin code now and understand how it works, I switched it to `dispatch_async` as suggested last year. 

I still plan to use `dispatch_sync` as a replacement to the CAS spin-lock. So to keep the consistency I use GCD everywhere.

packages/camera/camera/example/ios/RunnerTests/ThreadSafeMethodChannelTests.m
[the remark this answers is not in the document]
jmagman: Actually you don't need the local:
```suggestion
  [self expectationWithDescription:@"invokeMethod must be called on the main thread"];
```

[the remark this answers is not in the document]
hellohuanlin: Oh i think i still need this local variable so that I can fulfill it.

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -1115,11 +1132,16 @@ - (void)startImageStreamWithMessenger:(NSObject<FlutterBinaryMessenger> *)messen
...
-
-    _isStreamingImages = YES;
+    FLTThreadSafeEventChannel *threadSafeEventChannel =
+        [[FLTThreadSafeEventChannel alloc] initWithEventChannel:eventChannel];
+
+    _imageStreamHandler = [[FLTImageStreamHandler alloc] initWithDispatchQueue:_dispatchQueue];
+    [threadSafeEventChannel setStreamHandler:_imageStreamHandler
+                                  completion:^{
+                                    dispatch_async(self->_dispatchQueue, ^{
+                                      self->_isStreamingImages = YES;
```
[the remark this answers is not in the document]
hellohuanlin: `_dispatchQueue` is currently defined as an ivar and not property so I will keep it for now; Will update `isStreamingImages`

packages/camera/camera/example/ios/RunnerTests/ThreadSafeMethodChannelTests.m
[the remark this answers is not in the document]
jmagman: Oh, true for anything you need to `fulfill`.

packages/camera/camera/ios/Classes/FLTThreadSafeTextureRegistry.m
```diff
@@ -0,0 +1,52 @@
...
+
+- (void)registerTexture:(NSObject<FlutterTexture> *)texture
+             completion:(void (^)(int64_t))completion {
+  if (!NSThread.isMainThread) {
+    dispatch_async(dispatch_get_main_queue(), ^{
+      completion([self.registry registerTexture:texture]);
+    });
+  } else {
+    completion([self.registry registerTexture:texture]);
+  }
```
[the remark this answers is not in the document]
hellohuanlin: Great idea!

hellohuanlin:
FYI I've addressed all nits and will merge in a few hours.

PR #4672, opened 2022-01-17
[image_picker] Improves documentation around the destruction of the MainActivity.

mvanbeusekom:
Improves the documentation around the destruction of the MainActivity by the Android OS while the `Intent.ACTOIN_GET_CONTENT` or `MediaStore.ACTION_IMAGE_CAPTURE` is executing as there is still some confusion regarding this phenomenon (see flutter/flutter#48016).

Resolves flutter/flutter#93807.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/image_picker/image_picker/README.md
```diff
@@ -26,7 +26,7 @@ Add the following keys to your _Info.plist_ file, located in `<project root>/ios
 
 Starting with version **0.8.1** the Android implementation support to pick (multiple) images on Android 4.3 or higher.
 
-No configuration required - the plugin should work out of the box.
+No configuration required - the plugin should work out of the box. It is however highly recommended to prepare for Android killing the application when low on memory. How to prepare for this is discussed in the [Handling MainActivity destruction on Android](#handling-mainactivity-destruction-on-android) chapter.
```
stuartmorgan, line 29: s/chapter/section below/

packages/image_picker/image_picker/README.md
```diff
@@ -55,7 +55,7 @@ import 'package:image_picker/image_picker.dart';
 
 ### Handling MainActivity destruction on Android
 
-Android system -- although very rarely -- sometimes kills the MainActivity after the image_picker finishes. When this happens, we lost the data selected from the image_picker. You can use `retrieveLostData` to retrieve the lost data in this situation. For example:
+When under high memory pressure the Android system -- although very rarely -- sometimes kills the MainActivity of the application using the image_picker. On Android the image_picker plugin makes use of the default `Intent.ACTOIN_GET_CONTENT` or `MediaStore.ACTION_IMAGE_CAPTURE` intents. This means that while the intent is executing the source application is moved to the background and becomes eligable for cleanup when the system is low on memory. When the intent finished executing, Android will restart the source application, however when this happens, we lost the data selected from the image_picker. In this situation it is possible to use the `ImagePicker.retrieveLostData()` method to retrieve the lost data. For example:
```
stuartmorgan, line 58: I think we should probably remove the "although very rarely" since on some lower-end devices it seems pretty common based on issue reports. So:
"When under high memory pressure the Android system may the MainActivity of the application using `image_picker`."

stuartmorgan, line 58: Typo: Intent.**ACTION**_GET_CONTENT

stuartmorgan, line 58: s/the image_picker plugin/`image_picker`/

packages/image_picker/image_picker/README.md
```diff
@@ -55,7 +55,7 @@ import 'package:image_picker/image_picker.dart';
...
 
 ```dart
 Future<void> getLostData() async {
```
stuartmorgan, line 61: Could you fix the missing space in `for(final Xfile` below while you're updating this section?

packages/image_picker/image_picker/README.md
stuartmorgan, line 58: We should fix the weird "we lost" construction while we are here (and the tense issues). And actually we can shorten this a bit since the first part about losing the return data seems redundant. How about:
"When the intent finishes executing, Android will restart the application. Since the data is never returned to the original call in that case, use the `ImagePicker.retrieveLostData()` method to retrieve the lost data. For example:"

packages/image_picker/image_picker/README.md
```diff
@@ -74,7 +74,7 @@ Future<void> getLostData() async {
 }
 ```
 
-There's no way to detect when this happens, so calling this method at the right place is essential. We recommend to wire this into some kind of start up check. Please refer to the example app to see how we used it.
+There's no way to detect when this happens, so calling this method at the right place is essential. We recommend to wire this into some kind of start up check. Please refer to the [example](https://github.com/flutter/plugins/blob/main/packages/image_picker/image_picker/example/lib/main.dart) app to see how we used it.
```
stuartmorgan, line 77: While I know what it's trying to say, I find the first part of this section really confusing; there is a way to detect it, because the code above does exactly that. How about replacing this whole paragraph with:

```
This check should always be run at startup in order to detect and handle
this case. Please refer to the
[example app](https://pub.dev/packages/image_picker/example)
for a more complete example of handling this flow.
```

(I think it's best to link to the pub view of the example, since that's where people are likely to be reading the README)

stuartmorgan (changes requested):
I'm not sure why this README uses super long lines; could you fix the sections you are touching to wrap to 80 characters?

stuartmorgan:
I forgot to say: thanks for taking this :) It's been on my to-do list but I hadn't gotten to it.

packages/image_picker/image_picker/README.md
mvanbeusekom: Fixed.

packages/image_picker/image_picker/README.md
mvanbeusekom: Fixed.

mvanbeusekom: Fixed.

mvanbeusekom: Fixed.

mvanbeusekom: I like that, thank you for the suggestion. I only removed the "in that case" part from the sentence turning it into:

>When the intent finishes executing, Android will restart the application. Since the data is never returned to the original call, use the ImagePicker.retrieveLostData() method to retrieve the lost data. For example:

packages/image_picker/image_picker/README.md
mvanbeusekom: Fixed.

packages/image_picker/image_picker/README.md
mvanbeusekom: Yes that is much better indeed, thanks.

mvanbeusekom (commented):
Wrapped all "changed" lines to 80 chars as requested and applied the other suggestions you provided. Thanks for the review.

stuartmorgan (approved):
LGTM!

PR #4673, opened 2022-01-17
[local_auth] opening dialog fix 

cyrax111:
Fix an issue when on Android devices you try to authenticate with parameter `useErrorDialogs = true` and the device has hardware biometrics but it is `not enrolled` - the exception "not enrolled" appear but dialog for going to enroll does not.

This PR fixes #96646

## Pre-launch Checklist

- [V] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [V] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [V] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [V] I signed the [CLA].
- [V] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [V] I listed at least one issue that this PR fixes in the description above.
- [V] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [V] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [V] I updated/added relevant documentation (doc comments with `///`).

### DIFF
--- a/packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
+++ b/packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
@@ -294,7 +294,9 @@ public class LocalAuthPlugin implements MethodCallHandler, FlutterPlugin, Activi
 
   private boolean canAuthenticateWithBiometrics() {
     if (biometricManager == null) return false;
-    return biometricManager.canAuthenticate() == BiometricManager.BIOMETRIC_SUCCESS;
+    int biometricStatus = biometricManager.canAuthenticate();
+    return biometricStatus == BiometricManager.BIOMETRIC_SUCCESS 
+      || biometricStatus == BiometricManager.BIOMETRIC_STATUS_UNKNOWN;
   }
 
   private boolean hasBiometricHardware() {
--- a/packages/local_auth/CHANGELOG.md
+++ b/packages/local_auth/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.11
+
+* Fix the issue: "authenticate" with parameter "useErrorDialogs = true" does not open a dialog #96646
+
 ## 1.1.10
 
 * Removes dependency on `meta`.
--- a/packages/local_auth/pubspec.yaml
+++ b/packages/local_auth/pubspec.yaml
@@ -3,7 +3,7 @@ description: Flutter plugin for Android and iOS devices to allow local
   authentication via fingerprint, touch ID, face ID, passcode, pin, or pattern.
 repository: https://github.com/flutter/plugins/tree/master/packages/local_auth
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+local_auth%22
-version: 1.1.10
+version: 1.1.11
 
 environment:
   sdk: ">=2.14.0 <3.0.0"


packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
```diff
@@ -294,7 +294,9 @@ private boolean isDeviceSupported() {
 
   private boolean canAuthenticateWithBiometrics() {
     if (biometricManager == null) return false;
-    return biometricManager.canAuthenticate() == BiometricManager.BIOMETRIC_SUCCESS;
+    int biometricStatus = biometricManager.canAuthenticate();
+    return biometricStatus == BiometricManager.BIOMETRIC_SUCCESS
+        || biometricStatus == BiometricManager.BIOMETRIC_STATUS_UNKNOWN;
```
stuartmorgan, line 299: Can you explain why you are changing `canAuthenticateWithBiometrics`? If someone isn't enrolled, this method should return false, so this doesn't seem like a desirable change.

stuartmorgan (changes requested):
In the future, please don't remove items from the checklist; it's there for a reason.

This PR will need tests to move forward, as explained in the contribution docs.

packages/local_auth/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java
cyrax111: if `biometricStatus == BiometricManager.BIOMETRIC_STATUS_UNKNOWN` we don't know someone is enrolled or not

cyrax111: if `BIOMETRIC_STATUS_UNKNOWN` we can still authenticate with biometrics

cyrax111: as we can see in [documentation](https://developer.android.com/reference/androidx/biometric/BiometricManager#BIOMETRIC_STATUS_UNKNOWN)

> ... Applications that wish to enable biometric authentication on affected devices may still call BiometricPrompt#authenticate() after receiving this status code...

stuartmorgan: You've clipped out a critical part:
> Applications that wish to enable biometric authentication on affected devices may still call BiometricPrompt#authenticate() after receiving this status code **but should be prepared to handle possible errors.**

That means that "true" is no the correct return value here, because while it *might* be the case that biometric auth is available, it might also not. It's a different state. Reporting that would require API changes so that callers can decide whether to proceed or not.

cyrax111: @stuartmorgan I want to fix [#96646](https://github.com/flutter/flutter/issues/96646). Do you have any idea?

stuartmorgan: I don't understand the connection between this change and that issue. For instance, if your goal is to change handling of the "not enrolled case", why aren't you changing anything related to BIOMETRIC_ERROR_NONE_ENROLLED`?

You should clearly describe, either in the PR or in the issue, the specific flow whose behavior you are trying to change, how you determined that the change you are making is correct, and add thorough unit test coverage that shows (and asserts) the behaviors you are trying to create.

stuartmorgan:
@cyrax111 Are you still planning on updating this PR based on the feedback above?

stuartmorgan:

PR #4675, opened 2022-01-18
[image_picker_platform_interface] Add methods for picking both images and videos simultaneously.

BeMacized:
This PR adds two new functions to the image_picker plugin's platform interface, for picking images and videos from the same native UI. 

~~This PR is still in draft mode due to the lack of unit tests. I will be adding tests later.~~

Android and iOS implementations of these new methods can be found here: #4676

Relevant issue:
- flutter/flutter#89159

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
```diff
@@ -210,6 +210,81 @@ abstract class ImagePickerPlatform extends PlatformInterface {
...
+  ///   null, the image with the original quality will be returned. Compression
+  ///   is only supported for certain image types such as JPEG and on Android
+  ///   PNG and WebP. If compression is not supported for the image that is
+  ///   picked, a warning message will be logged.
+  ///
+  /// In Android, the MainActivity can be destroyed for various reasons. If that happens, the result will be lost
+  /// in this call. You can then call [getLostData] when your app relaunches to retrieve the lost data.
+  ///
+  /// If no image or video were picked, the return value is null.
+  Future<XFile?> getImageOrVideo({
```
stuartmorgan, line 242: Maybe we should call this `getGalleryImageOrVideo`, to make it more clear why this is a totally separate method instead of a replacement for `getImage` with an extra parameter for allowing video (which I was going to suggest until I realized it doesn't support `source`).

stuartmorgan: Although... is getting image or video but with the camera source something we could support in the future? If so, we should probably just make a replacement method.

packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
```diff
@@ -210,6 +210,81 @@ abstract class ImagePickerPlatform extends PlatformInterface {
...
+  ///   null, the image with the original quality will be returned. Compression
+  ///   is only supported for certain image types such as JPEG and on Android
+  ///   PNG and WebP. If compression is not supported for the image that is
+  ///   picked, a warning message will be logged.
+  ///
+  /// In Android, the MainActivity can be destroyed for various reasons. If that happens, the result will be lost
+  /// in this call. You can then call [getLostData] when your app relaunches to retrieve the lost data.
+  ///
+  /// If no images were picked, the return value is null.
+  Future<List<XFile>?> getMultiImageAndVideo({
```
stuartmorgan, line 279: Couldn't this just be a replacement for `getMultiImage`, rather than us having both methods forever?

stuartmorgan: And we could future-proof it by using a parameter object (e.g., `MultiImageSelectionOptions`). Over time we should migrate more of these to use objects (e.g., there should pretty clearly be an `ImageSettings` or similar as a sub-option that has width, height, and quality, given that we are repeating it in all these methods).

stuartmorgan (changes requested):

stuartmorgan:
@BeMacized Are you still planning on updating this?

BeMacized:
Right now, I am not, as I've temporarily been assigned to work on other projects.

yassinsameh:
@stuartmorgan if it's possible to just change your requested renaming of the method, while merging the method with getMultiImage at a later time? It would be great to have this feature live. Great work @BeMacized .

stuartmorgan:
@yassinsameh It is not possible to land PRs that don't go through a full review cycle, no.

This also wouldn't land without https://github.com/flutter/plugins/pull/4676 being reviewed (per [our contribution guide](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins)), and if a PR author doesn't currently have time to iterate on a small API PR, then they presumably don't have time to iterate on a PR that's over a thousand lines of diff across 30 files either.

Anyone interested in picking up PRs that the original authors don't currently have time to iterate on is always welcome to do so.

packages/image_picker/image_picker_platform_interface/lib/src/types/media_selection_options.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'package:image_picker_platform_interface/image_picker_platform_interface.dart';
+
+/// Options class for [ImagePickerPlatform.getMedia].
+class MediaSelectionOptions {
+  /// Construct a new MediaSelectionOptions instance.
+  MediaSelectionOptions({
+    List<RetrieveType>? types,
+    this.maxImageWidth,
+    this.maxImageHeight,
+    this.imageQuality,
```
stuartmorgan, line 14: We keep repeating this triplet of arguments; we should extract an `ImageAdjustmentOptions` (or similar; that's not a great name but I don't have a better suggestion offhand).

packages/image_picker/image_picker_platform_interface/lib/src/types/media_selection_options.dart
```diff
@@ -0,0 +1,48 @@
...
+    this.allowMultiple = false,
```
stuartmorgan, line 15: Do we need `maxDuration` here for videos? Or better, a `VideoSelectionOptions`?

More generally, it seems like we need a plan around how to handle type-specific options. Should this method always take the union of all options for the other methods?

packages/image_picker/image_picker_platform_interface/lib/src/types/media_selection_options.dart
stuartmorgan: Maybe `ImageSelectionOptions`.

packages/image_picker/image_picker_platform_interface/lib/src/types/media_selection_options.dart
```diff
@@ -0,0 +1,48 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:image_picker_platform_interface/image_picker_platform_interface.dart';
+
+/// Options class for [ImagePickerPlatform.getMedia].
+class MediaSelectionOptions {
+  /// Construct a new MediaSelectionOptions instance.
+  MediaSelectionOptions({
+    List<RetrieveType>? types,
```
stuartmorgan, line 11: We do we need both an array and a catch-all enum type? It seems like this allows any of `[media]`, `[images, video]`, `[images, media]`, `[video, media]`, or `[images, video, media]` to express the same thing, which is confusing.

stuartmorgan (changes requested):

packages/image_picker/image_picker_platform_interface/lib/src/types/media_selection_options.dart
stuartmorgan: (Or `[]` or `null`, in fact. We should also make this non-nullable to avoid having two ways for it to be empty.)

stuartmorgan:
Setting back to draft for now since there are open comments.

packages/image_picker/image_picker_platform_interface/lib/src/types/media_selection_options.dart
BeMacized: Agreed, moved it into `ImageSelectionOptions`.

laopaoer-wallet:
is it merged? waiting

stuartmorgan (changes requested):
Sorry for the delay, I needed to find time to at least skim the implementation PRs while reviewing this, and there was a lot there.

This mostly looks good, just some small comments.

packages/image_picker/image_picker_platform_interface/lib/src/types/media_selection_type.dart
```diff
@@ -0,0 +1,27 @@
...
+enum MediaSelectionType {
+  /// Static pictures.
+  image,
+
+  /// Videos.
+  video,
+}
+
+/// Serializes a [MediaSelectionType] value into a String value.
+String serializeMediaSelectionType(MediaSelectionType type) {
```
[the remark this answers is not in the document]
BeMacized: Moving to `method_channel_image_picker.dart`.

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
```diff
@@ -233,6 +233,57 @@ class MethodChannelImagePicker extends ImagePickerPlatform {
     return paths.map((dynamic path) => XFile(path as String)).toList();
   }
 
+  @override
+  Future<List<XFile>?> getMedia({
```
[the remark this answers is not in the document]
BeMacized: Initially, to stay consistent with the interface and behavior of `getMultiImage`. However, now I looked into it a bit that implementation makes no sense. 

Looking at the Android implementation for `getMultiImage`, which `getMedia` partly uses, I think the idea was for `null` to be returned rather than an empty list, when the user explicitly cancels the flow. However, at least on my device, it is not possible to pick 0 items. It's always either a non-empty list, or `null`. 
As for iOS, there is no such logic and it results in the same: either a non-empty list, or `null`.

It too would make more sense to me for it to return `[]` rather than `null`, so we could make this non-nullable, but that would mean the behavior would have to change for `getMultiImage` as well, which I'm guessing to be a breaking change.

packages/image_picker/image_picker_platform_interface/lib/src/types/media_selection_options.dart
```diff
@@ -0,0 +1,56 @@
...
+    this.allowMultiple = false,
+    this.imageAdjustmentOptions = const ImageAdjustmentOptions(),
+  }) : types = types ??
+            <MediaSelectionType>[
+              MediaSelectionType.image,
+              MediaSelectionType.video
+            ];
+
+  /// The types of allowed media to be picked.
+  /// Allows all types by default.
```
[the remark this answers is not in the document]
BeMacized: Fixed

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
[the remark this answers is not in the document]
stuartmorgan: It would be breaking for subclasses, yes. I think I'd rather be inconsistent than propagate the problematic use of a nullable collection to a new API though; less to clean up later.

stuartmorgan (commented):
We should put reviewing this on hold and do a fully combined PR, per the standard process documented at https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins. I should have caught that at the start; sorry for the oversight there.

It's fine to leave this and the implementation draft PRs open, but then to also combine those changes into the mega-PR along with the app-facing changes. Then we can evaluate the end-to-end API to make sure the entire flow makes sense.

Hixie:
Going to close this one as well for now just to get it off the review queue, but it's still mentioned in the issue in case anyone is ready to pick it up again in due course.

PR #4687, opened 2022-01-21
[camera]rename dispatchQueue to captureSessionQueue

hellohuanlin:
Rename the queue to be aligned with [the proposal](https://docs.google.com/document/d/1Znjp-BduXIb1Ix3JOE1YKyJi1Z0d6rB-2q6Lt7j8xSo/edit#). 

I am keeping this PR small and only scoped to renaming. 

Since it's just renaming, I think it's exempt from version bump and unit test

Issue: 

No version change: Because it's only renaming

## Pre-launch Checklist

- [x] I signed the [CLA].
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].*
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].*

### DIFF
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Minor internal code cleanup.
+
 ## 0.9.4+6
 
 * Fixes a crash in iOS when using image stream due to calling Flutter engine API on non-main thread. 
--- a/packages/camera/camera/ios/Classes/CameraPlugin.m
+++ b/packages/camera/camera/ios/Classes/CameraPlugin.m
@@ -22,23 +22,23 @@
 
 @interface FLTImageStreamHandler : NSObject <FlutterStreamHandler>
 // The queue on which `eventSink` property should be accessed
-@property(nonatomic, strong) dispatch_queue_t dispatchQueue;
-// `eventSink` property should be accessed on `dispatchQueue`.
+@property(nonatomic, strong) dispatch_queue_t captureSessionQueue;
+// `eventSink` property should be accessed on `captureSessionQueue`.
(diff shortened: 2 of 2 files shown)


jmagman:
> I think it's exempt from unit test

You're right, but you should still get a test exemption from @Hixie in Discord.

> If you need an exemption to this rule, contact Hixie on the #hackers channel in Chat.

@stuartmorgan we usually still do a CHANGELOG `NEXT` when the version isn't bumped?  Or do you not bother on little changes?

stuartmorgan:
> @stuartmorgan we usually still do a CHANGELOG `NEXT` when the version isn't bumped? Or do you not bother on little changes?

I moved us to a policy of generally requiring a CHANGELOG change for everything—see the last sentence of https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog for my rationale—which is why `publishable` is strict about it. But it's kind of a trial; if it turns out to be annoying in practice we can easily relax the policy and change the tooling to only require it when the version changes.

stuartmorgan:
In this case I would lean toward:

```
## NEXT

* Minor internal code cleanup.
```

or something like that. Mostly because if there were some subtle bug introduced by the rename that we didn't notice until the next `camera` release, it would make it easier for us to realize that there would be two PRs to go look at, rather than potentially being misled by the CHANGELOG and looking at only the last change.

Hixie:
test-exempt: code refactor with no semantic change

hellohuanlin:
@stuartmorgan updated the changelog. ptal thanks

stuartmorgan (approved):
LGTM

jmagman (approved):
LGTM

PR #4691, opened 2022-01-24
[webview_flutter] Remove dependency on flutter.dev and google.com

Hixie:
No CHANGELOG change: refactoring tests


stuartmorgan (approved):
LGTM! This is great :)

We'll need to duplicate this to the integration tests for the implementation packages (they are ~duplicates of this file; long story); you can either add that to this PR (https://github.com/flutter/plugins/search?q=primaryurl), or I can do it in a follow-up.

Hixie:
I'm loathe to be a bottleneck so I'm going to let you run with this. :-)

PR #4692, opened 2022-01-25
[camera] Fix disposed exceptions for the camera example on iOS

AlexV525:
No CHANGELOG change: only affects the example.
No version change: only affects the example.

Related to https://github.com/flutter/flutter/issues/97199.

Controllers were given to the field aggressively, which cause race conditions when initializing controllers. Now the code will remove the references ASAP when a controller has been disposed.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[...]

PR #4519, opened 2021-11-18
[camera] Fix ImageStream ImageFormatGroup is Ignored in iOS

zuvola:
Fixes [flutter/flutter#93842](https://github.com/flutter/flutter/issues/93842)

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[the auto-formatter]: https://github.com/flutter/plugins/blob/master/script/tool/README.md#format-code
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #4695, opened 2022-01-25
[camera]refactor flash mode and add unit test

hellohuanlin:
Breaking up `CameraPlugin.m` is part of optional cleanups discussed [in the proposal](https://docs.google.com/document/d/1Znjp-BduXIb1Ix3JOE1YKyJi1Z0d6rB-2q6Lt7j8xSo/edit#). The monolithic `CameraPlugin.m` makes it hard to unit test other sub-tasks, so we want to do this clean up sooner. 

This PR moves `FlashMode` type and its related util functions into a separate file. Also added unit tests for these util functions. 

No version change: just moving code

Issue here: https://github.com/flutter/flutter/issues/96429

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera/ios/Classes/FlashMode.h
```diff
@@ -0,0 +1,32 @@
...
+  FlashModeAuto,
+  FlashModeAlways,
+  FlashModeTorch,
+} FlashMode;
+
+/**
+ * Gets FlashMode from its string representation.
+ * @param mode a string representation of the FlashMode.
+ */
+FlashMode getFlashModeForString(NSString *mode);
```
jmagman, line 24: Needs `extern`.
```suggestion
extern FTLFlashMode FTLGetFlashModeForString(NSString *mode);
```

packages/camera/camera/ios/Classes/FlashMode.h
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Represents camera's flash mode. Mirrors `FlashMode` enum in flash_mode.dart.
+ */
+typedef enum {
+  FlashModeOff,
+  FlashModeAuto,
+  FlashModeAlways,
+  FlashModeTorch,
+} FlashMode;
```
jmagman, line 18: The Obj-C faux "name spacing" convention is a different three letter prefix per module.  So each plugin should have a different prefix, which should be different from Flutter's which is `FTL`.  But unfortunately they don't and it's not consistent. https://google.github.io/styleguide/objcguide.html#prefixes

So these symbols need some kind of prefix like `FTLFlashMode`, and the enum values should extend the typedef name for Swift compatibility.  Also use `NS_ENUM`. https://google.github.io/styleguide/objcguide.html#constants
```obj-c
typedef NS_ENUM(NSInteger, FTLFlashMode) {
  FTLFlashModeOff = 1,
  FTLFlashModeAuto = 2,
  FTLFlashModeAlways = 3,
  FTLFlashModeTorch = 4,
};
```

packages/camera/camera/ios/Classes/FlashMode.h
```diff
@@ -0,0 +1,32 @@
...
+ * Gets FlashMode from its string representation.
+ * @param mode a string representation of the FlashMode.
+ */
+FlashMode getFlashModeForString(NSString *mode);
+
+/**
+ * Gets AVCaptureFlashMode from FlashMode.
+ * @param mode flash mode.
+ */
+AVCaptureFlashMode getAVCaptureFlashModeForFlashMode(FlashMode mode);
```
jmagman, line 30:
```suggestion
extern AVCaptureFlashMode FTLGetAVCaptureFlashModeForFlashMode(FTLFlashMode mode);
```

packages/camera/camera/ios/Classes/camera-umbrella.h
```diff
@@ -8,6 +8,7 @@
 #import <camera/FLTThreadSafeFlutterResult.h>
 #import <camera/FLTThreadSafeMethodChannel.h>
 #import <camera/FLTThreadSafeTextureRegistry.h>
+#import <camera/FlashMode.h>
```
jmagman, line 11:
```suggestion
```

Looks like this is only needs to be exposed to the plugin implementation and a test, and we don't want these symbols exported to the wider Flutter app, or other plugins.
Fortunately most of the work for a `Test` module work was done in https://github.com/flutter/plugins/pull/4430 (which was cherry-picked with https://github.com/flutter/plugins/pull/4426#discussion_r727762574).

I think adding something like this to the modulemap should do it:
https://github.com/flutter/plugins/blob/d20d3bf48deb5432dfc44c8ae0debab0626e7830/packages/camera/camera/ios/Classes/CameraPlugin.modulemap#L7-L9
```diff
  explicit module Test {
    header "CameraPlugin_Test.h"
+   header "FlashMode.h"
  }
```

Then module imported in the test like this:
https://github.com/flutter/plugins/blob/d20d3bf48deb5432dfc44c8ae0debab0626e7830/packages/camera/camera/example/ios/RunnerTests/CameraOrientationTests.m#L6

And the header directly imported in the `CameraPlugin.m` implementation as you already have it.

See https://github.com/flutter/plugins/pull/4157 for more context about CocoaPods and modulemaps.

jmagman (changes requested):

packages/camera/camera/ios/Classes/FlashMode.h
stuartmorgan: Isn't it `FLT`?

But we could start using a different prefix for each plugin (or at least a different prefix for this repo vs others, like `FLTP`; as you said there's not much consistency right now (including Flutter itself not using a prefix in many cases, which drives me crazy) so we could just start incrementally improving.

packages/camera/camera/ios/Classes/FlashMode.h
hellohuanlin: ah yes

packages/camera/camera/ios/Classes/FlashMode.h
hellohuanlin: will go with `FLTFlashMode`

packages/camera/camera/ios/Classes/camera-umbrella.h
hellohuanlin: sounds reasonable. will take a look at how that works.

hellohuanlin:
> No version change: just moving code

@stuartmorgan Just double check - are you fine with not bumping the version here? Or do you think the added unit tests qualifies a version bump? There will be multiple similar PRs coming up.

packages/camera/camera/ios/Classes/FlashMode.h
hellohuanlin: Actually, should we go with `CAM` for camera?

stuartmorgan: @jmagman Do you just want to pick a prefixing policy for this repo that you're happy with that we can enforce going forward? (We can add a quick note about it in the ObjC section [here](https://github.com/flutter/plugins/blob/main/CONTRIBUTING.md#style)).

packages/camera/camera/ios/Classes/camera-umbrella.h
hellohuanlin: The modulemap is very convenient!

packages/camera/camera/ios/Classes/FlashMode.h
hellohuanlin: updated to `FLT` to be consistent with other types in camera for now, but happy to change to anything else

jmagman: I guess stick with `FLT` for now, I like `CAM` better but we can change it in another PR, and I can update the obj-c section as Stuart suggested.

> Isn't it `FLT`?

You're right it's not

packages/camera/camera/ios/Classes/FLTFlashMode.h
```diff
@@ -0,0 +1,32 @@
...
+  FLTFlashModeAuto,
+  FLTFlashModeAlways,
+  FLTFlashModeTorch,
+};
+
+/**
+ * Gets FLTFlashMode from its string representation.
+ * @param mode a string representation of the FLTFlashMode.
+ */
+extern FLTFlashMode getFLTFlashModeForString(NSString *mode);
```
jmagman, line 24: The function name also needs a prefix `FLTGetAVCaptureFlashModeForFlashMode`.
```suggestion
extern FLTFlashMode FLTGetAVCaptureFlashModeForFlashMode(NSString *mode);
```

packages/camera/camera/ios/Classes/FLTFlashMode.h
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Gets AVCaptureFlashMode from FLTFlashMode.
+ * @param mode flash mode.
+ */
+extern AVCaptureFlashMode getAVCaptureFlashModeForFLTFlashMode(FLTFlashMode mode);
```
jmagman, line 30: same.
```suggestion
extern AVCaptureFlashMode FLTGetAVCaptureFlashModeForFLTFlashMode(FLTFlashMode mode);
```

jmagman (approved):
LGTM with naming nit.
Thanks @hellohuanlin!

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -10,6 +10,7 @@
 #import <CoreMotion/CoreMotion.h>
 #import <libkern/OSAtomic.h>
 #import <uuid/uuid.h>
+#import "FLTFlashMode.h"
```
jmagman, line 13: I think `FlashMode.h` as you had it was better, there's no reason to prefix header names (that I'm aware of), even though a few other files have done so.

packages/camera/camera/ios/Classes/camera-umbrella.h
jmagman: Fortunately there was already one there.  Trust me it was really annoying to figure out how to make it work with CocoaPods the first time I did this. https://github.com/flutter/plugins/pull/4157 🙂

hellohuanlin:
This is strange. It is conflicting with https://github.com/flutter/plugins/pull/4692, which is not landed to main yet.

jmagman:
> This is strange. It is conflicting with #4692, which is not landed to main yet.

It's conflicting with #4519

hellohuanlin:
> > This is strange. It is conflicting with #4692, which is not landed to main yet.

Ah right. my mistake

packages/camera/camera/ios/Classes/CameraPlugin.m
jmagman: You're right, that is the pattern, and Apple also prefixes their header names.  Hard to keep it all straight. 🙂

PR #4682, opened 2022-01-19
[local_auth] updated canCheckBiometrics behavior

darwin-morocho:
On Android `canCheckBiometrics` returns true if the device doesn't have any biometric enrolled (For more info check https://github.com/flutter/flutter/issues/45497#issuecomment-1007310214). Now with this PR `canCheckBiometrics` returns false on Android if device has the biometric hardware but doesn't have any biometric enrolled.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[...]

PR #4697, opened 2022-01-26
[local_auth] Add platform interface to prepare for migration to federated architecture

BeMacized:
This PR is the first in a series to move the local_auth plugin to a federated architecture.

 1. **Move local_auth into a subfolder and add a platform interface**
 2. Add native implementations for the interface
    - iOS implementation (#4699)
3. Refactor app facing plugin to make use of the platform interface and native implementations.
(#4701)

Relevant issues:
 - flutter/flutter#68498
 - flutter/flutter#94224

## Pre-launch Checklist

- [x] I signed the [CLA].


BeMacized:
The failing `federated_safety` check should not be applicable here, as the app facing package does not yet make use of the platform interface.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,116 @@
...
+  static final Object _token = Object();
+
+  static LocalAuthPlatform _instance = DefaultLocalAuthPlatform();
+
+  /// The default instance of [LocalAuthPlatform] to use.
+  ///
+  /// Defaults to [DefaultLocalAuthPlatform].
+  static LocalAuthPlatform get instance => _instance;
+
+  /// Platform-specific plugins should set this with their own platform-specific
```
stuartmorgan, line 28: s/plugins/implementations/

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,116 @@
...
+  /// authentication (e.g. lack of relevant hardware). This might throw
+  /// [PlatformException] with error code [otherOperatingSystem] on the iOS
+  /// simulator.
+  Future<bool> authenticate({
+    required String localizedReason,
+    bool useErrorDialogs = true,
+    bool stickyAuth = false,
+    required Map<String, String> authStrings,
+    bool sensitiveTransaction = true,
+    bool biometricOnly = false,
```
stuartmorgan, line 81: Let's future-proof this based on lessons learned in other federated plugins by moving all of these booleans into a class called something like `AuthenticationOptions`. That way we can add new parameters without the breaking-change-or-duplicate method dance.

packages/local_auth/local_auth_platform_interface/lib/types/biometric_type.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Various types of biometric authentication.
```
stuartmorgan, line 5: This seems like a very good opportunity to add `strong` and `weak` per https://github.com/flutter/plugins/pull/3763#issuecomment-815033841

They won't be used yet, but it sets us up for a change we know we're going to make later anyway, without having to propagate it through the packages.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,116 @@
...
+  ///
+  /// Returns [Future] bool true or false:
+  Future<bool> stopAuthentication() async {
+    throw UnimplementedError('stopAuthentication() has not been implemented.');
+  }
+}
+
+/// The default interface implementation acting as a placeholder for
+/// the native implementation to be set.
+class DefaultLocalAuthPlatform extends LocalAuthPlatform {}
```
stuartmorgan, line 116: This seems like a good approach for a green-field plugin, but it would effectively be a regression in `local_auth` in some edge cases (such as existing third-party implementations of the method channel for other platforms). So as much as I'd love to not have a method channel sitting around that we don't actually use, I think we should make the default implementation be the current method channel (as with other plugins here).

stuartmorgan (changes requested):
> The failing federated_safety check should not be applicable here, as the app facing package does not yet make use of the platform interface.

This is true, but I'd rather we only override it when we really need to, and here there's a trivial solution of just doing the move as a separate initial PR before adding the platform interface package.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
BeMacized: Agreed, that would be much better. Changed.

packages/local_auth/local_auth_platform_interface/lib/types/biometric_type.dart
BeMacized: Added

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
BeMacized: Since no platform interface was published before, I'm assuming you mean forks of this plugin that make use of the previous method channel name in implementations for other platforms?

I've added a default implementation that should be 1:1 with the previous functionality. Please let me know if that is what you meant.

BeMacized:
> This is true, but I'd rather we only override it when we really need to, and here there's a trivial solution of just doing the move as a separate initial PR before adding the platform interface package.

Alright, this PR should now only contain the changes for the platform interface. The moved files have been taken care of separately.

packages/local_auth/local_auth_platform_interface/lib/default_method_channel_platform.dart
```diff
@@ -0,0 +1,65 @@
...
+        case 'iris':
+          biometrics.add(BiometricType.iris);
+          break;
+        case 'undefined':
+          break;
+      }
+    }
+    return biometrics;
+  }
+
```
stuartmorgan, line 57: Don't you need `canCheckBiometrics` in the platform interface, given this implementation? The way `canCheckBiometrics` works right now (on iOS; it's [currently broken on Android](https://github.com/flutter/plugins/pull/4682)) is that `'undefined'` signals that the device has biometrics capability, but nothing is enrolled. Since you're filtering at this layer, that can't be implemented in the app-facing plugin.

(I think that implementation is super-confusing, so I'm all for having `canCheckBiometrics` being in the platform interface so that implementations can move over time to something less weird without any changes to the interface.)

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
stuartmorgan: > I'm assuming you mean forks of this plugin that make use of the previous method channel name in implementations for other platforms?

Not forks, just implementations in general. See https://github.com/go-flutter-desktop/plugins/tree/master/url_launcher for an example. This is an implementation of `url_launcher` for a third-party embedding; it's not a fork, and it contains no Dart code, it's just a native implementation with a method channel listener that exactly matches what `url_launcher` sends. As long as that's registered by whatever plugin mechanism the embedding uses, and as long as the method channel doesn't change, this kind of thing Just Works.

It's fragile, but before federation introduced the platform interface concept it was the only way for other embeddings to support existing plugins.

(A more likely scenario in the case of local_auth is people mocking out the method channel in tests, which prior to platform interfaces was a common way of mocking out plugins in tests.)

What you've added here is indeed what I was looking for :)

packages/local_auth/local_auth_platform_interface/lib/default_method_channel_platform.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:flutter/services.dart';
+import 'package:local_auth_platform_interface/local_auth_platform_interface.dart';
+import 'package:local_auth_platform_interface/types/auth_options.dart';
+import 'package:local_auth_platform_interface/types/biometric_type.dart';
+
+const MethodChannel _channel = MethodChannel('plugins.flutter.io/local_auth');
+
+/// The default interface implementation acting as a placeholder for
+/// the native implementation to be set.
```
stuartmorgan, line 13: Let's add:

```
/// This implementation is not used by any of the implementations in this
/// repository, and exists only for backward compatibility with any
/// clients that were relying on internal details of the method channel
/// in the pre-federated plugin.
```

That way it's very clear why it's here even though we don't use it.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// register themselves.
+  static set instance(LocalAuthPlatform instance) {
+    PlatformInterface.verifyToken(instance, _token);
+    _instance = instance;
+  }
+
+  /// Authenticates the user with biometrics available on the device while also
+  /// allowing the user to use device authentication - pin, pattern, passcode.
+  ///
+  /// Returns a [Future] holding true, if the user successfully authenticated,
```
stuartmorgan, line 41: Let's just say "Returns true if [...]"; I think mentioning the Future just reads oddly, and it's a common pattern so I don't think it should cause confusion. (I know it's an existing comment, but we can take the opportunity to improve them.)

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// false otherwise.
+  ///
+  /// [localizedReason] is the message to show to user while prompting them
+  /// for authentication. This is typically along the lines of: 'Please scan
+  /// your finger to access MyApp.'. This must not be empty.
+  ///
+  /// Provide [authStrings] if you want to
+  /// customize messages in the dialogs.
```
stuartmorgan, line 49: We should say what the keys are supposed to be here.

stuartmorgan: Actually, looking at the current plugin... how is this intended to work? In the current implementation the keys here are platform-specific. How would a platform-agnostic front-end package call this with the correct things?

Is there a design proposal somewhere for how this will be structured?

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,91 @@
...
+    required String localizedReason,
+    required Map<String, String> authStrings,
+    AuthenticationOptions options = const AuthenticationOptions(),
+  }) async {
+    throw UnimplementedError('authenticate() has not been implemented.');
+  }
+
+  /// Returns a list of enrolled biometrics.
+  ///
+  /// Returns a [Future] List<BiometricType> with the following possibilities:
```
stuartmorgan, line 67: Let's remove the duplication and make this whole line: "Possible values include:"

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// - BiometricType.face
+  /// - BiometricType.fingerprint
+  /// - BiometricType.iris (not yet implemented)
```
stuartmorgan, line 70: Strong and weak should be listed.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,91 @@
...
+
+  /// Returns true if device is capable of checking biometrics or is able to
+  /// fail over to device credentials.
+  ///
+  /// Returns a [Future] bool true or false:
+  Future<bool> isDeviceSupported() async {
+    throw UnimplementedError('isDeviceSupported() has not been implemented.');
+  }
+
+  /// Returns true if auth was cancelled successfully.
```
stuartmorgan, line 84: This should start by saying what it *does*, since returning a value is not the primary purpose of the method.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// - BiometricType.iris (not yet implemented)
+  Future<List<BiometricType>> getAvailableBiometrics() async {
+    throw UnimplementedError(
+        'getAvailableBiometrics() has not been implemented.');
+  }
+
+  /// Returns true if device is capable of checking biometrics or is able to
+  /// fail over to device credentials.
+  ///
+  /// Returns a [Future] bool true or false:
```
stuartmorgan, line 79: Remove.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,91 @@
...
+  Future<bool> isDeviceSupported() async {
+    throw UnimplementedError('isDeviceSupported() has not been implemented.');
+  }
+
+  /// Returns true if auth was cancelled successfully.
+  /// Returns false if there was some error or no auth in progress.
+  ///
+  /// Returns [Future] bool true or false:
```
stuartmorgan, line 87: Remove.

packages/local_auth/local_auth_platform_interface/lib/types/auth_options.dart
```diff
@@ -0,0 +1,40 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Options wrapper for [LocalAuthPlatform.authenticate] parameters.
+class AuthenticationOptions {
+  /// Construct a new instance.
```
stuartmorgan, line 7: Constructs

packages/local_auth/local_auth_platform_interface/lib/types/auth_options.dart
```diff
@@ -0,0 +1,40 @@
...
+  const AuthenticationOptions({
+    this.useErrorDialogs = true,
+    this.stickyAuth = false,
+    this.sensitiveTransaction = true,
+    this.biometricOnly = false,
+  });
+
+  /// Whether the system will attempt to handle user fixable issues encountered
```
stuartmorgan, line 15: user-fixable

packages/local_auth/local_auth_platform_interface/lib/types/auth_options.dart
```diff
@@ -0,0 +1,40 @@
...
+  /// while authenticating. For instance, if fingerprint reader exists on the
+  /// phone but there's no fingerprint registered, the plugin will attempt to
```
stuartmorgan, line 17: s/phone/device/

packages/local_auth/local_auth_platform_interface/lib/types/auth_options.dart
```diff
@@ -0,0 +1,40 @@
...
+  /// Construct a new instance.
+  const AuthenticationOptions({
+    this.useErrorDialogs = true,
+    this.stickyAuth = false,
+    this.sensitiveTransaction = true,
+    this.biometricOnly = false,
+  });
+
+  /// Whether the system will attempt to handle user fixable issues encountered
+  /// while authenticating. For instance, if fingerprint reader exists on the
```
stuartmorgan, line 16: if a fingerprint reader

packages/local_auth/local_auth_platform_interface/lib/types/auth_options.dart
stuartmorgan, line 17: s/will/might/

packages/local_auth/local_auth_platform_interface/lib/types/auth_options.dart
```diff
@@ -0,0 +1,40 @@
...
+  /// phone but there's no fingerprint registered, the plugin will attempt to
+  /// take the user to settings to add one. Anything that is not user fixable,
+  /// such as no biometric sensor on device, will be returned
+  /// as a [PlatformException].
```
stuartmorgan, line 20: s/be returned as/still result in/

packages/local_auth/local_auth_platform_interface/lib/types/auth_options.dart
```diff
@@ -0,0 +1,40 @@
...
+  /// do something else.
+  final bool stickyAuth;
+
+  /// Whether platform specific precautions are enabled. For instance, on face
+  /// unlock, Android opens a confirmation dialog after the face is recognized
+  /// to make sure the user meant to unlock their phone.
+  final bool sensitiveTransaction;
+
+  /// Prevent authentications from using non-biometric local authentication
+  /// such as pin, passcode, and passcode.
```
stuartmorgan, line 38: s/and/or/
s/passcode/pattern/

packages/local_auth/local_auth_platform_interface/lib/types/auth_strings.dart
```diff
@@ -0,0 +1,13 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:intl/intl.dart';
+
+/// Message shown on a button that the user can click to go to settings pages
+/// from the current dialog. It is used on both Android and iOS sides.
```
stuartmorgan, line 8: s/. It is used on both Android and iOS sides/, on applicable platforms/

packages/local_auth/local_auth_platform_interface/lib/types/biometric_type.dart
```diff
@@ -0,0 +1,25 @@
...
```
stuartmorgan, line 5: We should expand this comment for the divergent types. Maybe add:

> Some platforms report specific biometric types, while others report only the classifications like `strong` and `weak`.

packages/local_auth/local_auth_platform_interface/lib/types/biometric_type.dart
```diff
@@ -0,0 +1,25 @@
...
+
+  /// Fingerprint authentication.
+  fingerprint,
+
+  /// Iris authentication.
+  iris,
+
+  /// Any biometric (e.g. fingerprint, iris, or face) on the device that meets
+  /// or exceeds the requirements for Class 3 (formerly Strong), as defined
+  /// by the Android CDD. Android only.
```
stuartmorgan, line 18: We shouldn't make claims about specific platforms in platform interfaces; who knows what future platforms will do. Although it's tricky since these classifications don't seem to be standards-based, just Android.

How about replacing "that meets [...]" with:
> that the platform API considers to be strong. For example, on android this corresponds to Class 3

Then the same for `weak` below.

packages/local_auth/local_auth_platform_interface/lib/types/auth_strings.dart
stuartmorgan: Why does this need to be in the platform interface though? I wouldn't expect this to be handled differently than the other strings.

packages/local_auth/local_auth_platform_interface/test/default_method_channel_platform_test.dart
```diff
@@ -0,0 +1,172 @@
...
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  const MethodChannel channel = MethodChannel(
+    'plugins.flutter.io/local_auth',
+  );
+
+  final List<MethodCall> log = <MethodCall>[];
+  late LocalAuthPlatform localAuthentication;
+
```
stuartmorgan, line 23: Please add a test that DefaultLocalAuthPlatform is automatically registered as the instance.

stuartmorgan (changes requested):

packages/local_auth/local_auth_platform_interface/lib/default_method_channel_platform.dart
BeMacized: Ah, I see. That behaviour is indeed a bit odd. I'll add it to the platform interface to allow for a custom implementation. 

For now I'll keep the same implementation in the iOS and Android packages, but I'll look into submitting a separate PR with a fix after they have been merged.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
BeMacized: This is indeed a bit of a weird one. The current plugin has two separate classes for iOS and Android which the `authenticate` method takes as optional parameters. These have been moved to their platform specific packages, but are still used in the app facing package as changing that would be a breaking change. As these classes are not platform agnostic, it doesn't seem right to use them in the platform interface, but that does lead to a situation like this where the keys are unknown. 

I'm not aware of any design proposal, and neither do I know what the right solution would be here.

packages/local_auth/local_auth_platform_interface/lib/types/auth_strings.dart
BeMacized: I moved it here because it is the only string that is shared between both iOS and Android. The others are all platform specific and therefore have been moved to their own respective platform packages. The native packages both require this string and the platform interface is their only shared dependency, hence why it went here.

Alternatively, we could just duplicate this string to both native packages so it can be removed from the platform interface?

stuartmorgan: Yes, we should not encode details of what happens to (currently) be true of the two platforms we happen to have implementations for into the platform interface.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
stuartmorgan: We'll want to deprecate that method and replace it with a new one that doesn't take the specific platforms.

I think the best option here is probably:
- Declare an empty base class for platform strings (`AuthMessages`) in the platform interface.
- Use that in the platform interface and app-facing APIs.
- Make a public subclass in each platform implementation package.
  - (To make moving the existing ones non-breaking, you'll need to re-`export` them for now from the app-facing package, with a TODO to remove it the next time there's a breaking change.)

Then clients would create and pass in the correct one via platform checks if they want to override strings.

Or actually, as a variant that's a bit easier to call: you could make the parameter be an `Iterable<AuthMessages>` instead of an `AuthMessages`, and then clients would just pass all of them in a list instead of needing to do platform checks. Each implementation would filter for only its own subclass within the list.

BeMacized: @stuartmorgan Agreed, that sounds like a more sane approach. I've updated all 4 PRs to implement this.

packages/local_auth/local_auth_platform_interface/lib/local_auth_platform_interface.dart
```diff
@@ -0,0 +1,97 @@
...
+  Future<bool> authenticate({
+    required String localizedReason,
+    required Iterable<AuthMessages> authMessages,
+    AuthenticationOptions options = const AuthenticationOptions(),
+  }) async {
+    throw UnimplementedError('authenticate() has not been implemented.');
+  }
+
+  /// Returns true if the device is capable of checking biometrics.
+  Future<bool> canCheckBiometrics() async {
```
stuartmorgan, line 66: Let's do some naming adjustment, so this API layer is as clear as possible even if it doesn't match the current app-facing API. I think `deviceSupportsBiometrics` is probably better here.

packages/local_auth/local_auth_platform_interface/lib/types/auth_messages.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Abstract class for storing platform specific strings.
+abstract class AuthMessages {
+  /// Constructs an instance of [AuthMessages].
+  const AuthMessages();
+
+  /// Returns all platform specific messages as a map.
```
stuartmorgan, line 10: Nit: platform-specific

stuartmorgan (changes requested):
Sorry for the churn, I have a couple of renames that I didn't think to suggest last time.

stuartmorgan (changes requested):
LGTM, thanks!

stuartmorgan (approved):
(Doh, wrong button)

PR #4698, opened 2022-01-26
[quick_actions] Extract the Android implementation into a separate package

mvanbeusekom:
Moves the native Android implementation from the quick_actions package into its own quick_actions_android package. I also duplicated the method channel implementation as suggested in issue flutter/flutter#94224.

Temporarily marks quick_actions as unpublishable to allow the implementations to be moved, rather than copied and deleted, in order to better preserve git history. A follow-up PR will restore it to publishable form.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/quick_actions/quick_actions/pubspec.yaml
```diff
@@ -15,12 +19,16 @@ flutter:
       android:
         package: io.flutter.plugins.quickactions
         pluginClass: QuickActionsPlugin
```
stuartmorgan, line 21: Remove these two lines.

packages/quick_actions/quick_actions_android/example/android/app/src/androidTest/java/io/flutter/plugins/quickactionsexample/QuickActionsTest.java
```diff
@@ -0,0 +1,23 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
```
stuartmorgan, line 4: I believe this file is for native unit tests, in which case it should be a move.

(Actually... is this even used? Java unit tests should be in the main source, not the example, and it looks like we already have that. So this file is likely a mistake. Try making it always fail and see if anything actually fails, and if not—which I expect is the case—you can just remove this, both here and the original.)

packages/quick_actions/quick_actions_android/example/integration_test/quick_actions_test.dart
```diff
@@ -0,0 +1,25 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// @dart = 2.9
```
stuartmorgan, line 5: As with iOS, can we remove this?

packages/quick_actions/quick_actions_android/example/lib/main.dart
```diff
@@ -0,0 +1,86 @@
...
+    quickActions.setShortcutItems(<ShortcutItem>[
+      // NOTE: This first action icon will only work on iOS.
+      // In a real world project keep the same file name for both platforms.
+      const ShortcutItem(
+        type: 'action_one',
+        localizedTitle: 'Action one',
+        icon: 'AppIcon',
+      ),
+      // NOTE: This second action icon will only work on Android.
+      // In a real world project keep the same file name for both platforms.
```
stuartmorgan, line 59: Same comments here as for iOS but reversed.

packages/quick_actions/quick_actions_android/example/pubspec.yaml
```diff
@@ -0,0 +1,29 @@
...
+environment:
+  sdk: ">=2.15.0 <3.0.0"
+  flutter: ">=2.8.0"
+
+dependencies:
+  flutter:
+    sdk: flutter
+  quick_actions_android:
+    # When depending on this package from a real application you should use:
+    #   quick_actions: ^x.y.z
```
stuartmorgan, line 14: quick_actions_android

packages/quick_actions/quick_actions_android/example/pubspec.yaml
```diff
@@ -0,0 +1,29 @@
...
+    # the parent directory to use the current plugin's version.
+    path: ../
+
+dev_dependencies:
+  espresso: ^0.1.0+2
+  flutter_driver:
+    sdk: flutter
+  integration_test:
+    sdk: flutter
+  pedantic: ^1.10.0
```
stuartmorgan, line 26: Remove

packages/quick_actions/quick_actions_android/example/test_driver/integration_test.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// @dart=2.9
```
stuartmorgan, line 5: Remove

packages/quick_actions/quick_actions_android/pubspec.yaml
```diff
@@ -0,0 +1,32 @@
...
+    sdk: flutter
+  quick_actions_platform_interface: ^1.0.0
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  integration_test:
+    sdk: flutter
+  mockito: ^5.0.0
+  pedantic: ^1.11.0
```
stuartmorgan, line 31: Remove

packages/quick_actions/quick_actions_android/pubspec.yaml
```diff
@@ -0,0 +1,32 @@
...
+  flutter:
+    sdk: flutter
+  quick_actions_platform_interface: ^1.0.0
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  integration_test:
+    sdk: flutter
+  mockito: ^5.0.0
```
stuartmorgan, line 30: I missed this on the iOS side... are we actually using mockito anywhere in this package?

packages/quick_actions/quick_actions_android/test/quick_actions_android_test.dart
```diff
@@ -0,0 +1,155 @@
...
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:quick_actions_android/quick_actions_android.dart';
+import 'package:quick_actions_platform_interface/quick_actions_platform_interface.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  group('$QuickActionsAndroid', () {
+    final QuickActionsAndroid quickActions = QuickActionsAndroid();
```
stuartmorgan, line 16: I missed this on the iOS side, but probably applies there too: please put this line in each test. Having the object under test be stateful across the tests is an antipattern and leads to a variety of problems (including tests passing when they shouldn't).

packages/quick_actions/quick_actions_android/test/quick_actions_android_test.dart
```diff
@@ -0,0 +1,155 @@
...
+
+    final List<MethodCall> log = <MethodCall>[];
```
stuartmorgan, line 18: Prefer making this `late` and created in `setUp` instead of being `final` and cleared in `setUp`, to reduce statefulness.

script/configs/exclude_all_plugins_app.yaml
```diff
@@ -8,3 +8,7 @@
 
 # This is a permament entry, as it should never be a direct app dependency.
 - plugin_platform_interface
+# TODO(mvanbeusekom): Remove the exclusion of the quick_actions_android and 
+#                     once the native implementation is removed from the 
+#                     quick_actions package.
+- quick_actions_android
```
stuartmorgan, line 14: This shouldn't be necessary now.

stuartmorgan (changes requested):

packages/quick_actions/quick_actions_android/example/android/app/src/androidTest/java/io/flutter/plugins/quickactionsexample/QuickActionsTest.java
mvanbeusekom: This is an interesting one. First of all the test is indeed run and will fail the tree if I force it to fail. Inside this file there is only one test that seems to verify if the QuickActions plugin is part of the `engine.getPlugins()` collection:

```java
  public void imagePickerPluginIsAdded() {
    final ActivityScenario<QuickActionsTestActivity> scenario =
        ActivityScenario.launch(QuickActionsTestActivity.class);
    scenario.onActivity(
        activity -> {
          assertTrue(activity.engine.getPlugins().has(QuickActionsPlugin.class));
        });
  }
```

Not sure if this test has a lot of value, I assume that loading the plugin it self is widely tested in the engine code it self, so the only thing it verifies now is that the plugin is added to the `pubspec.yaml` file (which I this doesn't really need a test).

If this test does have value, it should remain part of the example App since it tests if the plugin is loaded as part of the example App. It doesn't test any logic that is part of the native Java implementation. And doesn't it also make sense to keep the test in both the quick_actions_android and quick_actions example Apps? Both package contain their own example Apps which load the Android plugin in a slightly different way (meaning the registration in the `pubspec.yaml` is different).

Would like to hear your opinion on this issue.

stuartmorgan (approved):
LGTM once `QuickActionsTest.java` is converted to a move.

mvanbeusekom:
> LGTM once `QuickActionsTest.java` is converted to a move.

Done, added `waiting for tree to go green` label.

PR #4497, opened 2021-11-10
[path_provider] Publish fully federated version

stuartmorgan:
Now that the iOS and Android implementation packages are published,
switch to them and re-enable publishing.

Part of https://github.com/flutter/flutter/issues/68498

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[the auto-formatter]: https://github.com/flutter/plugins/blob/master/script/tool/README.md#format-code
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #4699, opened 2022-01-26
[local_auth] Add native iOS implementation of platform interface

BeMacized:
This PR is one in a series to move the local_auth plugin to a federated architecture.

 1. Move local_auth into a subfolder and add a platform interface 
 (#4697)
 2. **Add native implementations for the interface**
    - **iOS implementation**
3. Refactor app facing plugin to make use of the platform interface and native implementations. 
(#4701)

This PR remains in draft mode until the PR for the platform interface package has been merged and published (#4697).

Relevant issues:
 - flutter/flutter#68498
 - flutter/flutter#94224

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/local_auth/local_auth_ios/ios/Classes/FLTLocalAuthPlugin.m
```diff
@@ -0,0 +1,229 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
stuartmorgan, line 1: After the webview federation, where I was concerned about the impact to `git blame` of the copy-and-delete approach, I came up with a different way of doing this. See
https://github.com/flutter/plugins/pull/4465
and
https://github.com/flutter/plugins/pull/4497
for a pair of PRs that allows for preserving history by temporarily making the app-facing plugin unpublishable and using path-based deps to allow CI to still validate everything.

You *should* (I haven't tried, but don't see why it wouldn't work) be able to retrofit the PR to do this without re-doing anything, by doing a new commit that deletes the app-facing versions of these files and then squashing it with the commit that copied them.

/cc @mvanbeusekom

stuartmorgan (changes requested):
We should do the move of the app-facing package as a precursor PR to remove complexity from the substantive PR.

packages/local_auth/local_auth_ios/ios/Classes/FLTLocalAuthPlugin.m
mvanbeusekom: I can confirm that deleting the app-facing version and then squashing it with the previous commits works (see #4702).

packages/local_auth/local_auth_ios/lib/local_auth_ios.dart
```diff
@@ -0,0 +1,84 @@
...
+export 'package:local_auth_ios/types/auth_messages_ios.dart';
+export 'package:local_auth_platform_interface/types/auth_messages.dart';
+export 'package:local_auth_platform_interface/types/auth_options.dart';
+export 'package:local_auth_platform_interface/types/biometric_type.dart';
+
+const MethodChannel _channel =
+    MethodChannel('plugins.flutter.io/local_auth_ios');
+
+/// The implementation of [LocalAuthPlatform] for iOS.
+class LocalAuthIOS extends LocalAuthPlatform {
```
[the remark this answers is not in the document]
stuartmorgan: Actually, hm. I see that the strings, which are a public class, already use `IOS`. Since renaming that would be a breaking change, we'll need to stick with `IOS` there, in which case I think we should have package-level consistency override the style guide and use `IOS` in this instance too.

So this can stay as-is, but I'll leave all these comments so it's clear that it was a deliberate decision rather than an oversight.

packages/local_auth/local_auth_ios/test/local_auth_test.dart
```diff
@@ -0,0 +1,112 @@
...
+
+    setUp(() {
+      channel.setMockMethodCallHandler((MethodCall methodCall) {
+        log.add(methodCall);
+        return Future<dynamic>.value(true);
+      });
+      localAuthentication = LocalAuthIOS();
+      log.clear();
+    });
+
```
stuartmorgan, line 31: Looks like we're missing simple tests for the other methods; they should definitely be added.

stuartmorgan (changes requested):

packages/local_auth/local_auth_ios/lib/local_auth_ios.dart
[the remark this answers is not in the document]
stuartmorgan: Just to bring this conversation with myself full circle, the style guide was actually wrong; `IOS` is correct. We've now fixed it, and will use `IOS` going forward.

stuartmorgan:
You'll need to resolve with https://github.com/flutter/plugins/pull/3806, but it should be straightforward.

packages/local_auth/local_auth_ios/example/lib/main.dart
```diff
@@ -0,0 +1,236 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// ignore_for_file: public_member_api_docs
+
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:local_auth_ios/local_auth_ios.dart';
```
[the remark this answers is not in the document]
BeMacized: I've changed the code to make use of `LocalAuthPlatform.instance` instead, but I think I still need this import for accessing `IOSAuthMessages`.

packages/local_auth/local_auth/test/local_auth_test.dart
```diff
@@ -40,28 +40,14 @@ void main() {
...
-                'deviceCredentialsRequired':
-                    androidDeviceCredentialsRequiredTitle,
-                'deviceCredentialsSetupDescription':
-                    androidDeviceCredentialsSetupDescription,
-                'goToSetting': goToSettings,
-                'goToSettingDescription': androidGoToSettingsDescription,
-                'signInTitle': androidSignInTitle,
-              },
-            ),
+            isMethodCall('authenticate',
```
stuartmorgan, line 43: Tests in the app-facing package need to be rewritten to mock the platform interface, not the method channel. Whether any given implementation uses this platform channel format is an implementation detail of what is now a separate package.

packages/local_auth/local_auth_ios/example/ios/RunnerTests/FLTLocalAuthPluginTests.m
```diff
@@ -186,87 +186,4 @@ - (void)testFailedAuthWithoutBiometrics {
...
-                    result:^(id _Nullable result) {
-                      XCTAssertTrue([NSThread isMainThread]);
-                      XCTAssertTrue([result isKindOfClass:[NSNumber class]]);
-                      OCMVerify([mockAuthContext setLocalizedFallbackTitle:nil]);
-                      XCTAssertFalse([result boolValue]);
-                      [expectation fulfill];
-                    }];
-  [self waitForExpectationsWithTimeout:kTimeout handler:nil];
-}
-
```
stuartmorgan, line 271: Looks like you lost the new tests in the merge.

packages/local_auth/local_auth_ios/ios/Classes/FLTLocalAuthPlugin.m
stuartmorgan: This is not showing as a move; it looks like you forgot to delete local_auth/ios/

packages/local_auth/local_auth_ios/pubspec.yaml
```diff
@@ -0,0 +1,29 @@
+name: local_auth_ios
+description: iOS implementation of the local_auth plugin.
+repository: https://github.com/flutter/plugins/tree/master/packages/local_auth/local_auth_ios
+issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+local_auth%22
+# NOTE: We strongly prefer non-breaking changes, even at the expense of a
+# less-clean API. See https://flutter.dev/go/platform-interface-breaking-changes
```
stuartmorgan, line 6: This isn't a platform interface, so shouldn't have this comment.

stuartmorgan (changes requested):

packages/local_auth/local_auth/test/local_auth_test.dart
BeMacized: Yes, this has been done in #4701 which contains all changes for the app facing package.

stuartmorgan (changes requested):
This hasn't been updated to do a move, rather than copy, of the iOS implementation (which is why it's not suffering from https://github.com/flutter/plugins/pull/4700#pullrequestreview-934113270)

BeMacized:
I couldn't figure out how to update this PR to register the files as being moved, so I've created a new PR where this is fixed instead. Closing this PR as it is now superseded by:
- #5207.

PR #4700, opened 2022-01-26
[local_auth] Add native Android implementation of platform interface

BeMacized:
This PR is one in a series to move the local_auth plugin to a federated architecture.

 1. Move local_auth into a subfolder and add a platform interface 
 (#4697)
 2. **Add native implementations for the interface**
    - **Android implementation**
    - iOS implementation (#5207)
3. Refactor app facing plugin to make use of the platform interface and native implementations.
(#4701)

This PR remains in draft mode until #5195 has been merged.

Relevant issues:
 - flutter/flutter#68498
 - flutter/flutter#94224

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #4701, opened 2022-01-26
[local_auth] Refactor package to make use of new platform interface and native implementations

BeMacized:
This PR is the last in a series to move the local_auth plugin to a federated architecture.

 1. Move local_auth into a subfolder and add a platform interface 
 (#4697)
 2. Add native implementations for the interface
    - iOS implementation (#5207)
3. **Refactor app facing plugin to make use of the platform interface and native implementations.**

~~This PR remains in draft mode until the PR for the platform interface package (#4697) as well as the PRs for the native packages (#5207, #4700) have been merged and published.~~

This PR remains in draft mode until #5249 has been merged and published, so that the dependency version may be updated.

Relevant issues:
 - flutter/flutter#68498
 - flutter/flutter#94224

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/local_auth/local_auth/lib/local_auth.dart
```diff
@@ -1,145 +1,18 @@
...
+// to depend on the platform specific packages directly if they wish to use
+// these classes.
+export 'package:local_auth_android/types/auth_messages_android.dart'
+    show AndroidAuthMessages;
+// TODO(BeMacized): This should be removed the next time a breaking change
+// occurs and LocalAuth#authenticate is removed. Packages will be expected
+// to depend on the platform specific packages directly if they wish to use
+// these classes.
+export 'package:local_auth_ios/types/auth_messages_ios.dart'
+    show IOSAuthMessages;
```
BeMacized, line 14: @stuartmorgan As we're [bumping the major version](https://github.com/flutter/plugins/pull/5195#discussion_r845288546) here, should we already remove `authenticate` here without deprecating it first?

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,182 @@
...
+    );
+  }
+
+  /// Returns true if auth was cancelled successfully.
+  /// This api only works for Android.
+  /// Returns false if there was some error or no auth in progress.
+  ///
+  /// Returns [Future] bool true or false:
+  Future<bool> stopAuthentication() async {
+    if (_platform.isAndroid) {
```
stuartmorgan, line 155: Oops, this needs to be fixed; I missed it before. We should never do this kind of platform gating in the app-facing package.

We should instead just always pass through, and document that it is only supported on some platforms. The fact that it used to return `true` on iOS is weird since it's the opposite of what the docs say it would do. We should:
- Do a precursor PR to update `local_auth_ios` to return false, not throw (with a comment saying it's not supported), then:
- Remove this check.
- Make the minimum version of the iOS implementation required here the version with that fix.
- Explicitly list lack of platform support as one of the `false` cases in the comment above (arguably it's covered by "some error", but let's be explicit).

packages/local_auth/local_auth/pubspec.yaml
```diff
@@ -24,13 +20,9 @@ flutter:
...
   intl: ^0.17.0
-# Temporary path dependencies to allow moving Android and iOS implementations.
-  local_auth_android:
-    path: ../local_auth_android
-  local_auth_ios:
-    path: ../local_auth_ios
+  local_auth_android: ^1.0.0
+  local_auth_ios: ^ 1.0.0
   local_auth_platform_interface: ^1.0.1
   platform: ^3.0.0
```
stuartmorgan, line 27: We can then remove this as well.

stuartmorgan (changes requested):

packages/local_auth/local_auth/lib/src/local_auth.dart
BeMacized: Precursor PR: #5249

packages/local_auth/local_auth/CHANGELOG.md
```diff
@@ -1,7 +1,40 @@
-## NEXT
+## 2.0.0
 
+* Migrates plugin to federated architecture.
 * Adds OS version support information to README.
-* Switches over to default method implementation in new platform interface.
+* BREAKING CHANGE: Deprecated method `authenticateWithBiometrics` has been removed
```
stuartmorgan, line 5: Nit: missing period.

packages/local_auth/local_auth/CHANGELOG.md
```diff
@@ -1,7 +1,40 @@
...
+  Use `authenticate` instead.
+* BREAKING CHANGE: Parameters of `authenticate` have been changed.
+  
+  Example:
+```dart
```
stuartmorgan, line 10: Nit: I believe if you indent all of this code block, including the opening and closing backticks, two spaces (or maybe it's 4?) it will show as nested under the bullet point.

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,86 @@
...
+/// A Flutter plugin for authenticating the user identity locally.
+class LocalAuthentication {
+  /// Authenticates the user with biometrics available on the device while also
+  /// allowing the user to use device authentication - pin, pattern, passcode.
+  ///
+  /// Returns true if the user successfully authenticated, false otherwise.
+  ///
+  /// [localizedReason] is the message to show to user while prompting them
+  /// for authentication. This is typically along the lines of: 'Please scan
+  /// your finger to access MyApp.'. This must not be empty.
```
stuartmorgan, line 30: Nit: let's change "scan your finger" to "authenticate" so that this is not giving an example that's auth-type specific (and thus often the wrong thing to do).

packages/local_auth/local_auth/CHANGELOG.md
```diff
@@ -1,7 +1,40 @@
...
+  ],
+  options: const AuthenticationOptions(
+    useErrorDialogs: true,
+    stickyAuth: false,
+    sensitiveTransaction: true,
+    biometricOnly: false,
+  ),
+);
+```
+
```
stuartmorgan, line 36: We should add another breaking change highlighting that `strong` and `weak` have been added to the possible `getAvailableBiometrics` results, and that applications checking specific types should be updated to handle those as well.

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,86 @@
...
+      LocalAuthPlatform.instance.deviceSupportsBiometrics();
+
+  /// Returns true if device is capable of checking biometrics or is able to
+  /// fail over to device credentials.
+  ///
+  /// Returns a [Future] bool true or false:
+  Future<bool> isDeviceSupported() async =>
+      LocalAuthPlatform.instance.isDeviceSupported();
+
+  /// Returns a list of enrolled biometrics
```
stuartmorgan, line 78: Nit: missing period.

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,86 @@
...
+  ///
+  /// Returns a [Future] List<BiometricType> with the following possibilities:
+  /// - BiometricType.face
+  /// - BiometricType.fingerprint
+  /// - BiometricType.iris (not yet implemented)
```
stuartmorgan, line 83: Let's remove this entire section. This list is already wrong, and can become wrong again due to changes in a completely different package. People can just look at the enum to see the possibilities.

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,86 @@
...
+      ],
+      AuthenticationOptions options = const AuthenticationOptions()}) {
+    return LocalAuthPlatform.instance.authenticate(
+      localizedReason: localizedReason,
+      authMessages: authMessages,
+      options: options,
+    );
+  }
+
+  /// Returns true if auth was cancelled successfully.
```
stuartmorgan, line 55: Nit: blank line below, per doc conventions.

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,86 @@
...
+  /// This api only works for Android.
```
stuartmorgan, line 56: "This API is not supported by all platforms."

packages/local_auth/local_auth/lib/src/local_auth.dart
stuartmorgan, line 55:
```
/// Cancels any in-progress authentication, returning true if auth was
/// cancelled successfully.
```

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,86 @@
...
+  /// Returns false if there was some error, no authentication in progress,
+  /// or the current platform lacks support.
+  ///
+  /// Returns [Future] bool true or false:
+  Future<bool> stopAuthentication() async {
+    return LocalAuthPlatform.instance.stopAuthentication();
+  }
+
+  /// Returns true if device is capable of checking biometrics
```
stuartmorgan, line 65: Nit: missing period.

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,86 @@
...
+  ///
+  /// Returns a [Future] bool true or false:
```
stuartmorgan, line 67: Delete this; it's not useful to anyone who can read Dart code.

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,86 @@
...
+      options: options,
+    );
+  }
+
+  /// Returns true if auth was cancelled successfully.
+  /// This api only works for Android.
+  /// Returns false if there was some error, no authentication in progress,
+  /// or the current platform lacks support.
+  ///
+  /// Returns [Future] bool true or false:
```
stuartmorgan, line 60: Remove, as above.

packages/local_auth/local_auth/lib/src/local_auth.dart
```diff
@@ -0,0 +1,86 @@
...
+  /// Returns true if device is capable of checking biometrics
+  ///
+  /// Returns a [Future] bool true or false:
+  Future<bool> get canCheckBiometrics =>
+      LocalAuthPlatform.instance.deviceSupportsBiometrics();
+
+  /// Returns true if device is capable of checking biometrics or is able to
+  /// fail over to device credentials.
+  ///
+  /// Returns a [Future] bool true or false:
```
stuartmorgan, line 74: Remove.

stuartmorgan (changes requested):
Mostly looks good, but some small notes on the CHANGELOG. (And some minor cleanup to the API docs; I know those are mostly just moved as they were, but since there are some changes to docs related to platforms anyway, this is a good opportunity to do some quick cleanup.)

stuartmorgan (approved):
Looks great! Thanks for all the work on this; I know it ended up being a more complicated federation than some others have been, not least because I missed things during initial review.

PR #4707, opened 2022-01-27
[webview_flutter] Remove dependency on flutter.dev and google.com (Android/WKWebView)

BeMacized:
This is a follow up PR to this: 
- #4691

It applies the same changes that were made to the integration tests for the app facing package, to the identical integration tests in the webview_flutter_android and webview_flutter_wkwebview packages.


stuartmorgan (approved):
LGTM, thanks!

I added the same CHANGELOG override comment to the PR description; you'll just need to push some kind of update (e.g., amend the commit to make a trivial change to the commit message, then force push) to get the `publishable` check to pass.

PR #4712, opened 2022-01-28
[video_player] Federate mobile implementations

stuartmorgan:
Moves the iOS and Android implementations to their own federated packages. The iOS implementation package uses the name `avfoundation` rather than `ios` since it is likely that a macOS implementation could share a significant amount of code.

As part of the federation, migrates the mobile implementations to using in-package method channels rather than the shared method channel. Currently these are just duplicates of the existing method channels; updating them to new version of Pigeon will be done in a separate PR later.

Temporarily marks `video_player` as unpublishable to allow the implementations to be moved, rather than copied and deleted, in order to better preserve git history. A follow-up PR will restore it to publishable form.

Part of https://github.com/flutter/flutter/issues/68498
Part of https://github.com/flutter/flutter/issues/94224

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Some review notes:
- It's likely easier to review the individual commits here, rather than the entire diff all at once
- `MiniController` is just `VideoPlayerController`, stripped down to a smaller version that has what we need for the example. What's left is almost entirely unchanged, and it's only there to allow running the implementation packages examples for ease of local development, so I wouldn't worry too much about reviewing the two copies of that (they are identical; see the TODO comment there, but since sharing example code in implementation packages is a general problem we need to solve I didn't want to tackle it here)
- The tests are copies of exisiting tests, modified to not to use the app-facing package, and with tests that were only testing app-facing-package code removed.

packages/video_player/video_player_android/example/android/app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <domain-config cleartextTrafficPermitted="true">
+        <domain includeSubdomains="true">www.sample-videos.com</domain>
+        <domain includeSubdomains="true">184.72.239.149</domain>
```
gaaclarke, line 5: I don't think we're supposed to hit external servers in tests.

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -0,0 +1,172 @@
...
+import 'messages.dart';
+
+/// An iOS implementation of [VideoPlayerPlatform] that uses method channels.
+class AVFoundationVideoPlayer extends VideoPlayerPlatform {
+  final VideoPlayerApi _api = VideoPlayerApi();
+
+  /// Registers this class as the default instance of [VideoPlayerPlatform].
+  static void registerWith() {
+    VideoPlayerPlatform.instance = AVFoundationVideoPlayer();
+  }
```
gaaclarke, line 21: This is new code so we should have a test for it.  Ideally we'd have an integration test.  This is particularly important since if this code was deleted today everything would technically still work, no?

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -0,0 +1,173 @@
...
+
+/// An Android implementation of [VideoPlayerPlatform] that uses method
+/// channels.
+class AndroidVideoPlayer extends VideoPlayerPlatform {
+  final VideoPlayerApi _api = VideoPlayerApi();
+
+  /// Registers this class as the default instance of [PathProviderPlatform].
+  static void registerWith() {
+    VideoPlayerPlatform.instance = AndroidVideoPlayer();
+  }
```
gaaclarke, line 22: Same thing, needs a test.

packages/video_player/video_player_android/CONTRIBUTING.md
```diff
@@ -0,0 +1,33 @@
+## Updating pigeon-generated files
+
+If you update files in the pigeons/ directory, run the following
+command in this directory (ignore the errors you get about
+dependencies in the examples directory):
+
+```bash
+flutter pub upgrade
+flutter pub run pigeon --dart_null_safety --input pigeons/messages.dart
```
gaaclarke, line 9: This LGTM, but we should probably have a similar note in `video_player_platform_interface`.

packages/video_player/video_player_platform_interface/pubspec.yaml
```diff
@@ -19,3 +19,4 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   pedantic: ^1.10.0
+  pigeon: ^0.1.21
```
gaaclarke, line 22: This pigeon version should be pegged to a specific version since the wire format isn't guaranteed between versions.

gaaclarke (commented):
Okay, this looks good to me.  It was a pain to review but this makes sense.  Here is my overview of the changes:

1) the pigeon files are copied into the implementations
2) bindings are created for each platform between the platform interface and the generated api
3) federated so "video_player/iOS" and "video_player/android" deleted
4) use dart plugin registrant to set instance of video_player_platform_interface

I think we should have tests for the registrant, at a minimum a unit test, ideally an integration test where a video_player is created on iOS and the runtime type of the platform interface is checked.

packages/video_player/video_player_android/example/android/app/src/main/res/xml/network_security_config.xml
gaaclarke: (I suspect that this already existed but the git diff history is weird for it and it is showing up as new code?)

stuartmorgan: We have a bunch of tests that hit external servers. We've just started fixing them, but haven't gotten to video_player yet; I don't want to mix that into this PR.

It's showing as new code because video_player/video_player/example/android (which is what this is a copy of) still exists, so it's not a move. There's no concept of copies in git, unfortunately.

packages/video_player/video_player_android/CONTRIBUTING.md
stuartmorgan: This was just copied from `video_player`, but I did forget to copy it to `video_player_platform_interface`. Since it's no longer relevant in `video_player`, I've moved that copy to `video_player_platform_interface`, and added a warning about breaking changes to that copy.

packages/video_player/video_player_android/lib/src/android_video_player.dart
stuartmorgan: Added.

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -0,0 +1,172 @@
...
+import 'dart:async';
+import 'dart:ui';
+
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+import 'package:video_player_platform_interface/video_player_platform_interface.dart';
+
+import 'messages.dart';
+
+/// An iOS implementation of [VideoPlayerPlatform] that uses method channels.
```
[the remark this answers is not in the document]
stuartmorgan: Done.

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
stuartmorgan: Thanks, I forgot to add the usual unit test for this.

> Ideally we'd have an integration test.

The integration tests implicitly exercise this code, because the actual registration logic runs. It's what's setting the implementation that `MiniController` (and thus all of these tests) are relying on.

> This is particularly important since if this code was deleted today everything would technically still work, no?

Today it would work, because everything is currently the same as the default method channel. As soon as we update Pigeon, if `registerWith` weren't there everything would fail because we'd be using the Dart side of the old API, and the native side of the new one.

But I've added an explicit integration test for now, just to make 100% sure we're not using the default implementation.

stuartmorgan: > Today it would work, because everything is currently the same as the default method channel.

Which is what was happening, because I forgot to add the pubspec keys to register the Dart code 🤦🏻

stuartmorgan:
I completely missed that this was a general "looks good" not an actual "LGTM (approval) with the following things addressed" and landed it 😞  I'll follow up on any additional review comments today in a new PR.

gaaclarke:
Just reviewed the changes, LGTM, retroactive approval.

PR #4716, opened 2022-01-31
[google_maps_flutter] Switch app-facing package to new analysis options

TahaTesser:
Switches from legacy analysis options to current analysis options,
fixing all analysis issues it exposed.

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
We do not want custom analysis rules in flutter/plugins. See https://github.com/flutter/flutter/issues/76229 if you are interested in fixing analysis issues in the maps plugin, as that's the direction we want to go.

If there are things in flutter_lints that aren't in the repository settings, we should evaluate adding them there so the apply to all code.

TahaTesser:
@stuartmorgan 
Does it mean I update the all 3 packages in https://github.com/flutter/plugins/tree/main/packages/google_maps_flutter?

Since [custom_analysis.yaml](https://github.com/flutter/plugins/blob/main/script/configs/custom_analysis.yaml#L16-L18) has entire `google_maps_flutter`, not just its example 
I found that i need update interface package too since just migrating `google_maps_flutter` causes the tests to fail as it expects certain types during the test.

stuartmorgan:
> Does it mean I update the all the 3 maps packages in https://github.com/flutter/plugins/tree/main/packages/google_maps_flutter?
> 
> Since [custom_analysis.yaml](https://github.com/flutter/plugins/blob/main/script/configs/custom_analysis.yaml#L16-L18) has entire `google_maps_flutter`, not just its example

You need to do the entire package, not just the example. You don't need to do all three at once (and ideally shouldn't).

> I found that i need to update interface package too since just migrating `google_maps_flutter` causes the tests to fail as it expects certain types during the test.

I'm not sure what this means. What runtime type changes were you making?

TahaTesser:
@stuartmorgan 
Thanks for the instructions from the issue and notes, I am excited for this project.

I am entirely sure, I will do careful testing to be sure and update individual packages. 
Here one example of such error, analyzer is fine  in `google_maps_flutter` but test still fails, probably because i have assigned types to quite few maps in the codebase
```dart
00:05 +2 -64: /Users/tahatesser/AndroidStudioProjects/plugins/packages/google_maps_flutter/google_maps_flutter/test/google_map_test.dart: Can update zoomGesturesEnabled
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following PlatformException was thrown running a test:
PlatformException(error, type 'Null' is not a subtype of type 'double' in type cast, null, null)

#0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:607:7)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:177:18)
```

stuartmorgan:
> probably because i have assigned types to quite a few maps in the codebase

If this includes method channel code, you need to be very careful do do the right combinations of `as` with `dynamic` and then `cast`ing.

TahaTesser:
@stuartmorgan 
Following the instructions, I've made the changes. If it looks fine, I would move to other packages ASAP :)

stuartmorgan:
It looks like this doesn't pass tests currently.

TahaTesser:
@stuartmorgan 
fixed the changelog :)

packages/google_maps_flutter/google_maps_flutter/AUTHORS
```diff
@@ -64,3 +64,4 @@ Aleksandr Yurkovskiy <sanekyy@gmail.com>
 Anton Borries <mail@antonborri.es>
 Alex Li <google@alexv525.com>
 Rahul Raj <64.rahulraj@gmail.com>
+Taha Tesser <tessergmail.com>
```
stuartmorgan, line 67: This is missing an @

packages/google_maps_flutter/google_maps_flutter/example/lib/scrolling_map.dart
```diff
@@ -12,8 +12,10 @@ import 'package:google_maps_flutter/google_maps_flutter.dart';
 
 import 'page.dart';
 
+const LatLng center = LatLng(32.080664, 34.9563837);
```
stuartmorgan, line 15: Unless this is intended for use in other files, it should be `_center`.

packages/google_maps_flutter/google_maps_flutter/example/pubspec.yaml
```diff
@@ -7,18 +7,17 @@ environment:
...
     # When depending on this package from a real application you should use:
     #   google_maps_flutter: ^x.y.z
     # See https://dart.dev/tools/pub/dependencies#version-constraints
     # The example app is bundled with the plugin so we use a path dependency on
     # the parent directory to use the current plugin's version.
     path: ../
-  flutter_plugin_android_lifecycle: ^2.0.1
 
 dev_dependencies:
   espresso: ^0.1.0+2
```
stuartmorgan, line 23: `pedantic` should be remove below.

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -322,9 +325,13 @@ class _GoogleMapState extends State<GoogleMap> {
...
 
   @override
-  void dispose() async {
+  void dispose() {
+    disposeController();
     super.dispose();
-    GoogleMapController controller = await _controller.future;
+  }
+
+  Future<void> disposeController() async {
```
stuartmorgan, line 333: This should be private.

packages/google_maps_flutter/google_maps_flutter/test/polygon_updates_test.dart
```diff
@@ -122,14 +122,15 @@ void main() {
...
     final Polygon p1 = Polygon(
-      polygonId: PolygonId("polygon_1"),
-      points: <LatLng>[const LatLng(0.0, 0.0)],
+      polygonId: const PolygonId('polygon_1'),
+      points: _points,
     );
     await tester.pumpWidget(_mapWithPolygons(<Polygon>{p1}));
 
-    p1.points.add(const LatLng(1.0, 1.0));
+    _points.add(const LatLng(1.0, 1.0));
```
stuartmorgan, line 133: What was this set of changes fixing? The code seems less clear now, so I'm wondering if we can fix it differently.

packages/google_maps_flutter/google_maps_flutter/test/polyline_updates_test.dart
```diff
@@ -116,14 +116,15 @@ void main() {
     expect(update.geodesic, true);
   });
 
-  testWidgets("Mutate a polyline", (WidgetTester tester) async {
+  testWidgets('Mutate a polyline', (WidgetTester tester) async {
+    final List<LatLng> _points = <LatLng>[const LatLng(0.0, 0.0)];
```
stuartmorgan, line 120: Same.

stuartmorgan (changes requested):
Thanks for this cleanup! Just a few small things.

packages/google_maps_flutter/google_maps_flutter/test/polygon_updates_test.dart
TahaTesser: I meant to add `final List<LatLng> _points = <LatLng>[const LatLng(0.0, 0.0)];` to `p1`, to make _points list modifiable 
Otherwise, it throws `Unsupported operation: Cannot add to an unmodifiable list`


During local testing, I modified this line and forget to revert this line after making the p1 final above, thanks 

```diff
+   final List<LatLng> _points = <LatLng>[const LatLng(0.0, 0.0)];
     final Polygon p1 = Polygon(
-    polylineId: PolylineId("polyline_1"),
+   polygonId: const PolygonId('polygon_1'),
-    points: <LatLng>[const LatLng(0.0, 0.0)],
+   points: _points,
    );
    await tester.pumpWidget(_mapWithPolygons(<Polygon>{p1}));

    p1.points.add(const LatLng(1.0, 1.0));
```

stuartmorgan: I see, the reason this code looks strange is because it is; looking at the `blame` this is testing that we notice if someone modifies what is supposed to be an immutable object.

stuartmorgan (approved):
LGTM, thanks!

PR #4718, opened 2022-02-01
[image_picker] Default to gallery instead of camera when picking multiple images on pre-iOS 14 devices.

mvanbeusekom:
Solves two problems when using the `ImagePicker.pickMultiImage` method on pre-iOS 14 devices:

1. Picking multiple images now defaults to selecting images from the gallery instead of opening the camera (this is expected behaviour as it is not possible to select multiple images from the camera, see also issue flutter/flutter#96903);
2. Make sure the supplied `arguments` and `result` callback are made available to the `UIImagePicker` when picking images on devices running iOS 13 and lower.

Resolves flutter/flutter#96903
Replaces and closes #4706

**EDIT:** Converted this PR to draft as it depends on changes from PR #4722.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/image_picker/image_picker/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.8.4+6
+
+* Ensures the `result` callback and method call `arguments` send to iOS are 
+taken into consideration when picking multiple images on pre-iOS 14 devices;
+* Configures the `UIImagePicker` to default to gallery instead of camera on 
+pre-iOS 14 devices when picking multiple images on pre-iOS 14 devices.
```
stuartmorgan, line 6: Nit: you say "on pre-iOS 14 devices" twice in this sentence.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -101,13 +101,17 @@ - (void)pickImageWithPHPicker:(int)maxImagesAllowed API_AVAILABLE(ios(14)) {
 }
 
 - (void)pickImageWithUIImagePicker {
```
stuartmorgan, line 103: We should just not have this method. I really don't like the "put arguments in an ivar and read them later design; the bugs you are fixing here are exactly the sort of problem that kind of indirect system in notorious for. We should require the explicit argument, and over time completely eliminate `arguments` as an ivar.

AFAICT the calls are all from a context where we already have the arguments so this trampoline method shouldn't be needed.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -167,12 +170,9 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
     _imagePickerController.videoQuality = UIImagePickerControllerQualityTypeHigh;
 
-    self.result = result;
-    _arguments = call.arguments;
-
-    int imageSource = [[_arguments objectForKey:@"source"] intValue];
-    if ([[_arguments objectForKey:@"maxDuration"] isKindOfClass:[NSNumber class]]) {
-      NSTimeInterval max = [[_arguments objectForKey:@"maxDuration"] doubleValue];
+    int imageSource = [[self.arguments objectForKey:@"source"] intValue];
+    if ([[self.arguments objectForKey:@"maxDuration"] isKindOfClass:[NSNumber class]]) {
```
stuartmorgan, line 174: Please use `call.arguments`. It's far less fragile because then there's no order dependency with other code.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -23,6 +23,7 @@ - (UIViewController *)presentedViewController {
 
 @interface FLTImagePickerPlugin (Test)
 @property(copy, nonatomic) FlutterResult result;
+@property(copy, nonatomic) NSDictionary *arguments;
```
stuartmorgan, line 26: We should absolutely not make this design anti-pattern (see my other comments) something that tests rely on and prevent removing. We should test outcomes (expected calls to mocked system APIs, and/or expected return values based on those mocks), not implementation details.

stuartmorgan (changes requested):

mvanbeusekom:
Hi @stuartmorgan,

I fully agree with your comments and I have been struggling with this too. Ideally we'd separate the method channel implementation from the code that interacts directly with the `UIImagePicker` and `PHPicker` controls into separate classes. For example:

- A `MethodChannelHandler` class, responsible for communicating with the Flutter method channel and parsing arguments into strongly typed structures;
- A `FLTImagePicker` class, abstraction around the `UIImagePicker` and `PHPicker` controls exposing methods and callbacks accepting only strongly type and clearly named parameters.  

This would make testing the API surface between Dart and Native a lot easier. However this is also a large refactor. My reasoning was to solve this bug now by not introducing new parameters, ivars or logic but instead simply fixing the existing implementation (and add some tests to ensure it works) and discuss a bigger refactor with you offline.

stuartmorgan:
> Ideally we'd separate the method channel implementation from the code that interacts directly with the `UIImagePicker` and `PHPicker` controls into separate classes.

Sounds like a good way to go longer term, but it shouldn't actually be necessary for what I'm suggesting. Look at the local_auth iOS tests for instance, which have a single class but still do method-call-to-mock-system-API testing.

> My reasoning was to solve this bug now by not introducing new parameters, ivars or logic

Well, there's obviously new logic here. But I'm fine with the short term fix itself being pretty minimal (the non-tests changes I asked for should be quite small), the issue is that the change you made to the tests make the code *worse*. They make fixing the root problem harder, because they enshrine and encourage an anti-pattern. Small incremental changes are fine but they need to take us in the right direction.

> but instead simply fixing the existing implementation (and add some tests to ensure it works)

I think "ensure it works" is highly arguable. As an experiment I tried moving `self.arguments = call.arguments;` to the end of `handleMethodCall:...` (where it would do absolutely nothing), and 2/3 of your new tests still pass.

They ensure that part of the anti-pattern is being followed at some point, but not that it's the right point, or that it's used correctly.

packages/image_picker/image_picker/CHANGELOG.md
mvanbeusekom: Removed the redundant "on pre-iOS 14 devices".

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
mvanbeusekom: Removed this method as suggested.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -56,10 +56,32 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
+ * Initializes the _imagePickerController member with a new instance of the
+ * UIImagePickerController class.
+ *
+ * A new instance of the UIImagePickerController is created every time the
+ * initImagePickerController method is called. For testing purposes this can
+ * be overriden using the setImagePickerControllerOverride method, in which
+ * case the set instance of the UIImagePickerController is used to initialize
+ * the _imagePickerController member.
+ */
+- (void)initImagePickerController {
```
stuartmorgan, line 69: `init*` is a confusing name for a method in ObjC that's not a class initializer. I think the structure of having this have a side-effect also makes the code at the call side harder to follow (and it locks us into having the controller as an ivar, which I'm not convinced is actually a good idea; it seems like we could make this less plugin stateful and harder to order incorrectly by passing it around).

Let's use [this pattern](https://github.com/flutter/plugins/blob/main/packages/local_auth/local_auth/ios/Classes/FLTLocalAuthPlugin.m#L49-L60) instead, which keeps the assignment as it was before, and just replaces the creation with a method call.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -175,12 +194,9 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
 
-    self.result = result;
-    _arguments = call.arguments;
-
-    int imageSource = [[_arguments objectForKey:@"source"] intValue];
-    if ([[_arguments objectForKey:@"maxDuration"] isKindOfClass:[NSNumber class]]) {
-      NSTimeInterval max = [[_arguments objectForKey:@"maxDuration"] doubleValue];
+    int imageSource = [[call.arguments objectForKey:@"source"] intValue];
+    if ([[call.arguments objectForKey:@"maxDuration"] isKindOfClass:[NSNumber class]]) {
+      NSTimeInterval max = [[call.arguments objectForKey:@"maxDuration"] doubleValue];
```
stuartmorgan, line 199: Since you're touching these lines anyway, please replace `objectForKey:` with the modern `[]` syntax. We're opportunistically updating old style as we change code.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -45,6 +43,8 @@ @interface FLTImagePickerPlugin () <UINavigationControllerDelegate,
 typedef NS_ENUM(NSInteger, ImagePickerClassType) { UIImagePickerClassType, PHPickerClassType };
 
 @implementation FLTImagePickerPlugin {
+  NSDictionary *_arguments;
```
stuartmorgan, line 46: I believe @jmagman prefers we move *to* using properties for everything, rather than away from properties in favor of raw ivars.

stuartmorgan (changes requested):

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
mvanbeusekom: Moved back to using properties, my bad (Jenn indeed explained this to me once before).

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
mvanbeusekom: Done.

packages/image_picker/image_picker/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.8.4+8
+
+* Ensures the `result` callback and method call `arguments` send to iOS are 
+taken into consideration when picking multiple images on pre-iOS 14 devices;
```
stuartmorgan, line 4: Nit: This is a full sentence, so should end with a period.

packages/image_picker/image_picker/example/ios/RunnerTests/ImagePickerPluginTests.m
```diff
@@ -146,6 +146,32 @@ - (void)testPluginPickVideoDeviceFront {
                  UIImagePickerControllerCameraDeviceFront);
 }
 
+- (void)testPickMultiImageShouldUseUIImagePickerControllerOnPreiOS14 {
+  if (@available(iOS 14, *)) {
```
stuartmorgan, line 150: @jmagman Relevant to our discussions about test matrixes in this repository: we won't ever actually test this in CI currently.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -37,16 +37,18 @@ @interface FLTImagePickerPlugin () <UINavigationControllerDelegate,
 
 @property(strong, nonatomic) PHPickerViewController *pickerViewController API_AVAILABLE(ios(14));
 
+@property(strong, nonatomic) UIImagePickerController *imagePickerController;
```
stuartmorgan, line 40: Optional nit, since it's a pre-existing problem: Could you add a comment for this property? Properties and ivars should, by style, have declaration comments.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -37,16 +37,18 @@ @interface FLTImagePickerPlugin () <UINavigationControllerDelegate,
...
+
+@property(strong, nonatomic) NSMutableArray<UIImagePickerController *> *imagePickerControllerOverrides;
```
stuartmorgan, line 42: Needs a declaration comment, per style.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -141,47 +164,43 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
     } else {
-      [self pickImageWithUIImagePicker];
+      [self launchUIImagePickerWithSource:SOURCE_GALLERY];
     }
   } else if ([@"pickVideo" isEqualToString:call.method]) {
-    _imagePickerController = [[UIImagePickerController alloc] init];
-    _imagePickerController.modalPresentationStyle = UIModalPresentationCurrentContext;
-    _imagePickerController.delegate = self;
-    _imagePickerController.mediaTypes = @[
+    UIImagePickerController *imagePickerController = [self createImagePickerController];
```
stuartmorgan, line 191: Shouldn't this be `self.imagePickerController`?

If we need a stateful controller, I'm very concerned about our test coverage if this didn't break anything. Does it not actually need to be stateful?

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -484,9 +503,9 @@ - (void)imagePickerController:(UIImagePickerController *)picker
     self.result = nil;
     _arguments = nil;
   } else {
-    UIImage *image = [info objectForKey:UIImagePickerControllerEditedImage];
+    UIImage *image = info[UIImagePickerControllerEditedImage];
     if (image == nil) {
-      image = [info objectForKey:UIImagePickerControllerOriginalImage];
+      image = info[UIImagePickerControllerOriginalImage];
```
stuartmorgan, line 508: Thanks for fixing this everywhere :)

stuartmorgan (changes requested):

packages/image_picker/image_picker/CHANGELOG.md
mvanbeusekom: Removed the first sentence.

packages/image_picker/image_picker/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+* Configures the `UIImagePicker` to default to gallery instead of camera when
+picking multiple images on pre-iOS 14 devices.
```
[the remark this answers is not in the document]
mvanbeusekom: Removed the first sentence.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
mvanbeusekom: Added the declaration comments as required by style.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
mvanbeusekom: Added the declaration comments as required by style.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
mvanbeusekom: With some refactoring it is not necessary to keep the `imagePickerController` as state. I did the refactoring necessary which involves passing the current copy of the `imagePickerController` around to some other methods. Please have a look if you agree with this.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -31,22 +31,39 @@ @interface FLTImagePickerPlugin () <UINavigationControllerDelegate,
                                     PHPickerViewControllerDelegate,
                                     UIAdaptivePresentationControllerDelegate>
 
+/**
+ * Gets or sets the maximum amount of images that are allowed to be picked.
```
stuartmorgan, line 35: Minor nit: for all of these you can document them as if they were variables rather than methods; that's the normal style for properties. So this would be "The maximum number of images that are allowed to be picked." and similarly for the others below.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
stuartmorgan: Looks great, thanks! (Now we just need to do the same with arguments, and ideally result, but those are definitely out of scope.)

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
```diff
@@ -202,19 +224,19 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
   }
 }
 
-- (void)showCamera {
+- (void)showCameraWithImagePickerController:(UIImagePickerController *)imagePickerController {
```
stuartmorgan, line 227: Optional nit: you could make these slightly less verbose by making them `withImagePicker:`; there's not non-controller image picker (AFAIK) to confuse it with.

stuartmorgan (approved):
LGTM with nits, thanks for all the iteration and test work here!

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
mvanbeusekom: Fixed.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin.m
mvanbeusekom: Fixed.

PR #4721, opened 2022-02-01
[camera]writing file on background queue

hellohuanlin:
Writing photo files on background queue. A performance improvement similar to the previous PR that moves sample buffer handling to capture session queue. 

Very interesting learning related to mocking NSData: It turned out that we can't `OCMClassMock` NSData because XCTest infrastructure itself uses NSData (e.g. `XCTRunnerIDESession::logDebugMessage:`), which runs on a private queue (`com.apple.dt.xctest.default-debug-log-handler`). The race condition is pretty common and makes the test very fragile. Took me a while to debug. 

Issue: 

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.9.4+10
+
+* iOS performance improvement by writing captured photo file on a background IO queue. 
+
 ## 0.9.4+9
 
 * iOS performance improvement by moving sample buffer handling from the main queue to a background session queue. 
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.9.4+10
 
-* iOS performance improvement by writing captured photo file on a background IO queue. 
+* iOS performance improvement by moving file writing from the main queue to a background IO queue. 
 
 ## 0.9.4+9
 
--- a/packages/camera/camera/example/ios/RunnerTests/FLTSavePhotoDelegateTests.m
+++ b/packages/camera/camera/example/ios/RunnerTests/FLTSavePhotoDelegateTests.m
@@ -1,0 +1,135 @@
+//
+//  FLTSavePhotoDelegateTests.m
+//  RunnerTests
+//
+//  Created by Huan Lin on 1/31/22.
+//  Copyright © 2022 The Flutter Authors. All rights reserved.
+//
+
+@import camera;
+@import camera.Test;
+@import AVFoundation;
+@import XCTest;
+#import <OCMock/OCMock.h>
+
+@interface FLTSavePhotoDelegate : NSObject <AVCapturePhotoCaptureDelegate>
+@property(readonly, nonatomic) NSString *path;
+- initWithPath:(NSString *)path
+        result:(FLTThreadSafeFlutterResult *)result
+       ioQueue:(dispatch_queue_t)ioQueue;
+- (void)handlePhotoCaptureResultWithError:(nullable NSError *)error
+                        photoDataProvider:(NSData * (^)(void))photoDataProvider;
+@end
+
+@interface FLTSavePhotoDelegateTests : XCTestCase
+
+@end
+
+@implementation FLTSavePhotoDelegateTests
+
+- (void)testHandlePhotoCaptureResult_mustSendErrorIfFailedToCapture {
+  NSError *error = [NSError errorWithDomain:@"test" code:0 userInfo:nil];
+  dispatch_queue_t ioQueue = dispatch_queue_create("test", NULL);
+  id mockResult = OCMClassMock([FLTThreadSafeFlutterResult class]);
+  FLTSavePhotoDelegate *delegate = [[FLTSavePhotoDelegate alloc] initWithPath:@"test"
+                                                                       result:mockResult
+                                                                      ioQueue:ioQueue];
+
+  [delegate handlePhotoCaptureResultWithError:error
+                            photoDataProvider:^NSData * {
+                              return nil;
+                            }];
+  OCMVerify(times(1), [mockResult sendError:error]);
+}
+
+- (void)testHandlePhotoCaptureResult_mustSendErrorIfFailedToWrite {
+  XCTestExpectation *resultExpectation =
+      [self expectationWithDescription:@"Must send IOError to the result if failed to write file."];
+  dispatch_queue_t ioQueue = dispatch_queue_create("test", NULL);
+  id mockResult = OCMClassMock([FLTThreadSafeFlutterResult class]);
+  OCMStub([mockResult sendErrorWithCode:@"IOError" message:@"Unable to write file" details:nil])
+      .andDo(^(NSInvocation *invocation) {
+        [resultExpectation fulfill];
+      });
+  FLTSavePhotoDelegate *delegate = [[FLTSavePhotoDelegate alloc] initWithPath:@"test"
+                                                                       result:mockResult
+                                                                      ioQueue:ioQueue];
+
+  // We can't use OCMClassMock for NSData because some XCTest APIs uses NSData (e.g.
+  // `XCTRunnerIDESession::logDebugMessage:`) on a private queue.
+  id mockData = OCMPartialMock([NSData data]);
+  OCMStub([mockData writeToFile:[OCMArg any] atomically:[OCMArg any]]).andReturn(NO);
+  [delegate handlePhotoCaptureResultWithError:nil
+                            photoDataProvider:^NSData * {
+                              return mockData;
+                            }];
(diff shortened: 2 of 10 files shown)


packages/camera/camera/ios/Classes/FLTCam.m
```diff
@@ -162,6 +168,7 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
   }
   _enableAudio = enableAudio;
   _captureSessionQueue = captureSessionQueue;
+  _photoIOQueue = dispatch_queue_create("io.flutter.camera.photoIOQueue", NULL);
```
hellohuanlin, line 171: `captureSessionQueue` was injected because it's also used in `CameraPlugin.m`

packages/camera/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.9.4+10
```
hellohuanlin, line 1: @stuartmorgan not sure if we should keep going with `0.9.4+10` or use `0.9.5`. I only find `x.y.z` versioning in [The doc](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates).

hellohuanlin: Oh I found this: 

> The interpretation of each number is just shifted down one slot: going from 0.1.2 to 0.2.0 indicates a breaking change, going to 0.1.3 indicates a new feature, and going to 0.1.2+1 indicates a change that doesn’t affect the public API. For simplicity’s sake, avoid using + after the version reaches 1.0.0

Looks like we should keep bumping `+x` for bug fixes until we reach `1.0.0`

stuartmorgan: The controlling document here is this one (from PR checklist):
> * [...] with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning),

The second to last paragraph under https://dart.dev/tools/pub/versioning#semantic-versions covers this versioning.

hellohuanlin: wow this version conflict resolution is smart! I like that we document not just what the rules are, but also explain our thought process behind these rules. Very helpful and educative.

packages/camera/camera/example/ios/RunnerTests/FLTSavePhotoDelegateTests.m
```diff
@@ -0,0 +1,122 @@
...
+  id mockResult = OCMClassMock([FLTThreadSafeFlutterResult class]);
+  FLTSavePhotoDelegate *delegate = [[FLTSavePhotoDelegate alloc] initWithPath:@"test"
+                                                                       result:mockResult
+                                                                      ioQueue:ioQueue];
+
+  [delegate handlePhotoCaptureResultWithError:error
+                            photoDataProvider:^NSData * {
+                              return nil;
+                            }];
+  OCMVerify(times(1), [mockResult sendError:error]);
```
jmagman, line 29: I don't think you need the `times(1)`.
```suggestion
  OCMVerify([mockResult sendError:error]);
```

packages/camera/camera/example/ios/RunnerTests/FLTSavePhotoDelegateTests.m
```diff
@@ -0,0 +1,122 @@
...
+        [resultExpectation fulfill];
+      });
+  FLTSavePhotoDelegate *delegate = [[FLTSavePhotoDelegate alloc] initWithPath:@"test"
+                                                                       result:mockResult
+                                                                      ioQueue:ioQueue];
+
+  // We can't use OCMClassMock for NSData because some XCTest APIs uses NSData (e.g.
+  // `XCTRunnerIDESession::logDebugMessage:`) on a private queue.
+  id mockData = OCMPartialMock([NSData data]);
+  OCMStub([mockData writeToFile:[OCMArg any] atomically:[OCMArg any]]).andReturn(NO);
```
jmagman, line 48: fyi you can use `OCMOCK_ANY` instead of `[OCMArg any]`, it's the same.  Up to you though.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.h
```diff
@@ -0,0 +1,37 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
jmagman, line 1: This is now exposed in the plugin API, I don't think we want that.  I think keep most of this in the implementation and only expose `FLTSavePhotoDelegate_Test.h` in the test module.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.m
```diff
@@ -0,0 +1,64 @@
...
+
+@implementation FLTSavePhotoDelegate
+
+- (instancetype)initWithPath:(NSString *)path
+                      result:(FLTThreadSafeFlutterResult *)result
+                     ioQueue:(dispatch_queue_t)ioQueue {
+  self = [super init];
+  NSAssert(self, @"super init cannot be nil");
+  _path = path;
+  _selfReference = self;
```
jmagman, line 15: I know you didn't write this but this is a strange way to keep `self` alive, and it's shared between all the completion blocks, so it could be set to `nil` in one but not before another is complete.  You'd think it would be retained by `FLTCam` or something...
Anyway this is no worse than it was.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.m
```diff
@@ -0,0 +1,64 @@
...
+  }
+  dispatch_async(self.ioQueue, ^{
+    NSData *data = photoDataProvider();
+    bool success = [data writeToFile:self.path atomically:YES];
+
+    if (!success) {
+      [self.result sendErrorWithCode:@"IOError" message:@"Unable to write file" details:nil];
+      return;
+    }
+    [self.result sendSuccessWithData:self.path];
```
jmagman, line 36: Can you make all references to `bool` -> `BOOL`?

Probably makes to send back the error details, something like
```dart
    NSError *error;
    if ([data writeToFile:self.path options:NSDataWritingAtomic error:&error]) {
      [self.result sendSuccessWithData:self.path];
    } else {
      [self.result sendErrorWithCode:@"IOError" message:@"Unable to write file" details:error.localizedDescription];
    }
```

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.h
hellohuanlin: Oh I didn't add it to the umbrella header so it's not exposed outside of this package.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.m
hellohuanlin: This is discussed in the "Remove selfReference" part of [the proposal](https://docs.google.com/document/d/1Znjp-BduXIb1Ix3JOE1YKyJi1Z0d6rB-2q6Lt7j8xSo/edit#heading=h.p926krbv9zgw): 

> selfReference is used to keep FLTSavePhotoDelegate around. This is a nice trick, but should only be used as a workaround. In our case this delegate object has a clear owner - the delegating object (FLTCam), which controls the lifecycle of this delegate.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.m
hellohuanlin: Sharp eyes! I didn't notice the `bool`.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.h
hellohuanlin: Also removed thread safe wrappers from the umbrella header

jmagman: Oh right right I forgot this uses a real umbrella header and not one generated by CocoaPods.

I think you're still exposing it in the test module since you're `#import`ing it in the `_Test` header.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.m
jmagman: Ha well at least we were in agreement.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.h
hellohuanlin: yep

hellohuanlin: The original header and the test header are both exposed for unit test only

jmagman (approved):
LGTM

PR #4480, opened 2021-11-05
[webview_flutter] Implement loadRequest in iOS package.

BeMacized:
This PR is a partial resubmission of my colleague's PR: #4169. 

It implements the `loadRequest` method added in #4450 for the iOS package.

Related issue:
- flutter/flutter#27730

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Note that unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I [updated pubspec.yaml](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates) with an appropriate new version according to the [pub versioning philosophy].
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test exempt.
- [x] All existing and new tests are passing.

PR #4722, opened 2022-02-02
[image_picker] Refactored tests to expose private interface in separate test header.

mvanbeusekom:
Refactored the existing unit tests to expose private interfaces in a separate test header instead of an inline interface. This was first introduced in the google_sign_in plugin (#4157) and after that also applied in the camera plugin (#4426) and the webview_flutter plugin (#4480).

I have moved this refactoring out of the changes I am working on for issue flutter/flutter#96903 and PR #4718.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/image_picker/image_picker/CHANGELOG.md
+++ b/packages/image_picker/image_picker/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.8.4+7
 
-* Refactores unit test to expose private interface via a separate test header instead of the inline declaration.
+* Refactors unit test to expose private interface via a separate test header instead of the inline declaration.
 
 ## 0.8.4+6
 
--- a/packages/image_picker/image_picker/CHANGELOG.md
+++ b/packages/image_picker/image_picker/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.8.4+7
+
+* Refactores unit test to expose private interface via a separate test header instead of the inline declaration.
+
 ## 0.8.4+6
 
 * Fixes minor type issues in iOS implementation.
--- a/packages/image_picker/image_picker/example/ios/RunnerTests/ImageUtilTests.m
+++ b/packages/image_picker/image_picker/example/ios/RunnerTests/ImageUtilTests.m
@@ -5,6 +5,7 @@
 #import "ImagePickerTestImages.h"
 
 @import image_picker;
+@import image_picker.Test;
 @import XCTest;
 
 @interface ImageUtilTests : XCTestCase
--- a/packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin_Test.h
+++ b/packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin_Test.h
@@ -1,0 +1,16 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// This header is available in the Test module. Import via "@import image_picker.Test;"
+
+#import <image_picker/FLTImagePickerPlugin.h>
+
+/// Methods exposed for unit testing.
+@interface FLTImagePickerPlugin ()
+
+@property(copy, nonatomic) FlutterResult result;
+- (void)handleSavedPathList:(NSArray *)pathList;
+- (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
+
+@end
--- a/packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin_Test.h
+++ b/packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin_Test.h
@@ -6,11 +6,37 @@
 
 #import <image_picker/FLTImagePickerPlugin.h>
 
-/// Methods exposed for unit testing.
+/** Methods exposed for unit testing. */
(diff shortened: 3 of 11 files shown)


packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin_Test.h
```diff
@@ -0,0 +1,16 @@
...
+// This header is available in the Test module. Import via "@import image_picker.Test;"
+
+#import <image_picker/FLTImagePickerPlugin.h>
+
+/// Methods exposed for unit testing.
+@interface FLTImagePickerPlugin ()
+
+@property(copy, nonatomic) FlutterResult result;
+- (void)handleSavedPathList:(NSArray *)pathList;
+- (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker;
```
stuartmorgan, line 14: These should always have had declaration comments; could you add those while you are exposing them?

stuartmorgan: (Also, a small style nit: blank line between each one, in addition to the comment.

packages/image_picker/image_picker/ios/Classes/image_picker-umbrella.h
```diff
@@ -0,0 +1,10 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Foundation/Foundation.h>
+#import <image_picker/FLTImagePickerImageUtil.h>
+#import <image_picker/FLTImagePickerMetaDataUtil.h>
+#import <image_picker/FLTImagePickerPhotoAssetUtil.h>
+#import <image_picker/FLTImagePickerPlugin.h>
+#import <image_picker/FLTPHPickerSaveImageToPathOperation.h>
```
stuartmorgan, line 10: Are this header and the Util headers really intended to be public?

@jmagman Only public headers should be in the umbrella header, right (as with a framework master header)?

stuartmorgan (changes requested):

packages/image_picker/image_picker/ios/Classes/image_picker-umbrella.h
jmagman: > @jmagman Only public headers should be in the umbrella header, right (as with a framework master header)?

Correct.  Looks like it was public before this change, which isn't right.

packages/image_picker/image_picker/ios/Classes/FLTImagePickerPlugin_Test.h
mvanbeusekom: Added declaration comments to each of the methods. Note that I simply copied already existing documentation from Apple documentation on the `imagePickerControllerDidCancel` method.

packages/image_picker/image_picker/ios/Classes/image_picker-umbrella.h
mvanbeusekom: How would I solve this the best way? If I simply remove the util headers from the umbrella header I run into issues with the unit tests for those util classes. For example:

```
/packages/image_picker/image_picker/example/ios/RunnerTests/ImageUtilTests.m:17:24: Unknown receiver 'FLTImagePickerImageUtil'; did you mean 'FLTImagePickerPlugin'?
```

Should I create special test headers and add them to the `Test` module in the .modulemap?

packages/image_picker/image_picker/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.8.4+6
+
+* Refactores unit test to expose private interface via a separate test header instead of the inline declaration.
```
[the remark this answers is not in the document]
mvanbeusekom: Good catch, fixed it thank you.

packages/image_picker/image_picker/example/ios/RunnerTests/ImageUtilTests.m
```diff
@@ -6,6 +6,7 @@
 
 @import image_picker;
 @import XCTest;
+#import <image_picker/FLTImagePickerImageUtil.h>
```
stuartmorgan, line 9: Isn't this still relying on the bug in the framework (that it includes this header in the packaged framework at all)?

Pre module-map I would expect this to just be `#import "FLTImagePickerImageUtil.h"` with appropriate include paths for the target. I expect there's probably a cleaner way using the Test module though?

hellohuanlin: @mvanbeusekom i think you can put this header under the submodule `Test` that you've created, and just do `@import image_picker.Test;` here.

mvanbeusekom: The solution proposed by @hellohuanlin worked like a charm. Thanks

jmagman (approved):
LGTM, thanks for getting into the weeds with this.

jmagman:
@stuartmorgan this needs another review since you requested changes.

stuartmorgan (approved):
I'm not sure I'm sold on the specific things being exposed in the test header here, but we can tune that in the test PR if necessary; since the important thing is the structural change here, LGTM.

PR #4723, opened 2022-02-02
[shared_preferences] Move away from shared method channel implementation in native packages.

BeMacized:
This PR eliminates the use of the shared method channel implementation in each of the native packages for the shared_preferences plugin.

This PR's contents are based on the steps outlined in the following issue: 
- flutter/flutter#94224

No CHANGELOG change: The main package only has one of its tests changed.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/shared_preferences/shared_preferences_android/CHANGELOG.md
+++ b/packages/shared_preferences/shared_preferences_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.11
+
+* Removes use of shared method channel implementation. 
+
 ## 2.0.10
 
 * Removes dependency on `meta`.
--- a/packages/shared_preferences/shared_preferences_android/CHANGELOG.md
+++ b/packages/shared_preferences/shared_preferences_android/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 2.0.11
 
-* Removes use of shared method channel implementation. 
+* Switches to an in-package method channel implementation.
 
 ## 2.0.10
 
--- a/packages/shared_preferences/shared_preferences_android/test/shared_preferences_android_test.dart
+++ b/packages/shared_preferences/shared_preferences_android/test/shared_preferences_android_test.dart
@@ -1,0 +1,111 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:shared_preferences_android/shared_preferences_android.dart';
+import 'package:shared_preferences_platform_interface/method_channel_shared_preferences.dart';
+import 'package:shared_preferences_platform_interface/shared_preferences_platform_interface.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  group(MethodChannelSharedPreferencesStore, () {
+    const MethodChannel channel = MethodChannel(
+      'plugins.flutter.io/shared_preferences_android',
+    );
+
+    const Map<String, Object> kTestValues = <String, Object>{
+      'flutter.String': 'hello world',
+      'flutter.Bool': true,
+      'flutter.Int': 42,
+      'flutter.Double': 3.14159,
+      'flutter.StringList': <String>['foo', 'bar'],
+    };
+
+    late InMemorySharedPreferencesStore testData;
+
+    final List<MethodCall> log = <MethodCall>[];
+    late SharedPreferencesStorePlatform store;
+
+    setUp(() async {
+      testData = InMemorySharedPreferencesStore.empty();
+
+      channel.setMockMethodCallHandler((MethodCall methodCall) async {
+        log.add(methodCall);
+        if (methodCall.method == 'getAll') {
+          return await testData.getAll();
+        }
+        if (methodCall.method == 'remove') {
+          final String key = (methodCall.arguments['key'] as String?)!;
+          return await testData.remove(key);
+        }
+        if (methodCall.method == 'clear') {
+          return await testData.clear();
+        }
+        final RegExp setterRegExp = RegExp(r'set(.*)');
+        final Match? match = setterRegExp.matchAsPrefix(methodCall.method);
+        if (match?.groupCount == 1) {
+          final String valueType = match!.group(1)!;
+          final String key = (methodCall.arguments['key'] as String?)!;
+          final Object value = (methodCall.arguments['value'] as Object?)!;
+          return await testData.setValue(valueType, key, value);
+        }
+        fail('Unexpected method call: ${methodCall.method}');
+      });
+      store = SharedPreferencesAndroid();
+      log.clear();
+    });
+
+    tearDown(() async {
+      await testData.clear();
+    });
+
+    test('getAll', () async {
+      testData = InMemorySharedPreferencesStore.withData(kTestValues);
+      expect(await store.getAll(), kTestValues);
+      expect(log.single.method, 'getAll');
+    });
+
+    test('remove', () async {
+      testData = InMemorySharedPreferencesStore.withData(kTestValues);
+      expect(await store.remove('flutter.String'), true);
+      expect(await store.remove('flutter.Bool'), true);
+      expect(await store.remove('flutter.Int'), true);
+      expect(await store.remove('flutter.Double'), true);
+      expect(await testData.getAll(), <String, dynamic>{
+        'flutter.StringList': <String>['foo', 'bar'],
+      });
+
+      expect(log, hasLength(4));
+      for (final MethodCall call in log) {
+        expect(call.method, 'remove');
+      }
+    });
+
+    test('setValue', () async {
+      expect(await testData.getAll(), isEmpty);
+      for (final String key in kTestValues.keys) {
+        final Object value = kTestValues[key]!;
+        expect(await store.setValue(key.split('.').last, key, value), true);
+      }
+      expect(await testData.getAll(), kTestValues);
+
+      expect(log, hasLength(5));
+      expect(log[0].method, 'setString');
+      expect(log[1].method, 'setBool');
+      expect(log[2].method, 'setInt');
+      expect(log[3].method, 'setDouble');
+      expect(log[4].method, 'setStringList');
+    });
+
+    test('clear', () async {
+      testData = InMemorySharedPreferencesStore.withData(kTestValues);
+      expect(await testData.getAll(), isNotEmpty);
+      expect(await store.clear(), true);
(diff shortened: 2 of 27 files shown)


packages/shared_preferences/shared_preferences_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.11
+
+* Removes use of shared method channel implementation.
```
stuartmorgan, line 3: Same rewording in these CHANGELOGs as for the url_launcher PR please

packages/shared_preferences/shared_preferences_macos/lib/shared_preferences_macos.dart
```diff
@@ -0,0 +1,51 @@
...
+import 'package:flutter/services.dart';
+import 'package:shared_preferences_platform_interface/shared_preferences_platform_interface.dart';
+
+const MethodChannel _kChannel =
+    MethodChannel('plugins.flutter.io/shared_preferences_macos');
+
+/// An implementation of [SharedPreferencesStorePlatform] that uses method channels.
+class SharedPreferencesMacOS extends SharedPreferencesStorePlatform {
+  /// Registers this class as the default instance of [SharedPreferencesStorePlatform].
+  static void registerWith() {
```
stuartmorgan, line 16: Same note about unit testing these.

packages/shared_preferences/shared_preferences_linux/test/shared_preferences_linux_test.dart
```diff
@@ -81,3 +84,26 @@ void main() {
     expect(await _readTestFile(), '{}');
   });
 }
+
+/// Fake implementation of PathProviderWindows that returns hard-coded paths,
```
stuartmorgan, line 88: PathProviderLinux

packages/shared_preferences/shared_preferences_linux/test/shared_preferences_linux_test.dart
```diff
@@ -81,3 +84,26 @@ void main() {
...
+/// allowing tests to run on any platform.
+///
+/// Note that this should only be used with an in-memory filesystem, as the
+/// path it returns is a root path that does not actually exist on Windows.
```
stuartmorgan, line 92: Linux

packages/shared_preferences/shared_preferences_macos/pubspec.yaml
```diff
@@ -2,23 +2,26 @@ name: shared_preferences_macos
...
 
 dependencies:
   flutter:
     sdk: flutter
   shared_preferences_platform_interface: ^2.0.0
 
 dev_dependencies:
+  flutter_test:
+    sdk: flutter
   pedantic: ^1.10.0
```
stuartmorgan, line 27: Optional: Could you remove this while you are here? It's cruft that I missed when cleaning up analysis options.

stuartmorgan (changes requested):
Just minor things here as well.

packages/shared_preferences/shared_preferences_android/test/shared_preferences_android_test.dart
```diff
@@ -0,0 +1,117 @@
...
+    setUp(() async {
+      testData = InMemorySharedPreferencesStore.empty();
+
+      channel.setMockMethodCallHandler((MethodCall methodCall) async {
+        log.add(methodCall);
+        if (methodCall.method == 'getAll') {
+          return await testData.getAll();
+        }
+        if (methodCall.method == 'remove') {
+          final String key = (methodCall.arguments['key'] as String?)!;
```
stuartmorgan, line 41: Nit: you can write this slightly more simply as `methodCall.arguments['key']! as String`

packages/shared_preferences/shared_preferences_android/test/shared_preferences_android_test.dart
```diff
@@ -0,0 +1,117 @@
...
+
+    const Map<String, Object> kTestValues = <String, Object>{
+      'flutter.String': 'hello world',
+      'flutter.Bool': true,
+      'flutter.Int': 42,
+      'flutter.Double': 3.14159,
+      'flutter.StringList': <String>['foo', 'bar'],
+    };
+
+    late InMemorySharedPreferencesStore testData;
```
stuartmorgan, line 27: I was confused at first as to why a `InMemorySharedPreferencesStore` would be in this file at all; I think it's worth an explanatory comment here like:
```
// Create a dummy in-memory implementation to back the mocked method channel API
// to simplify validation of the expected calls.
```

packages/shared_preferences/shared_preferences_android/test/shared_preferences_android_test.dart
```diff
@@ -0,0 +1,117 @@
...
+        }
+        if (methodCall.method == 'clear') {
+          return await testData.clear();
+        }
+        final RegExp setterRegExp = RegExp(r'set(.*)');
+        final Match? match = setterRegExp.matchAsPrefix(methodCall.method);
+        if (match?.groupCount == 1) {
+          final String valueType = match!.group(1)!;
+          final String key = (methodCall.arguments['key'] as String?)!;
+          final Object value = (methodCall.arguments['value'] as Object?)!;
```
stuartmorgan, line 52: Same for these two.

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -16,14 +16,9 @@ import 'package:shared_preferences_platform_interface/shared_preferences_platfor
 ///
 /// This class implements the `package:shared_preferences` functionality for Windows.
 class SharedPreferencesWindows extends SharedPreferencesStorePlatform {
-  /// The default instance of [SharedPreferencesWindows] to use.
-  // TODO(egarciad): Remove when the Dart plugin registrant lands on Flutter stable.
-  // https://github.com/flutter/flutter/issues/81421
-  static SharedPreferencesWindows instance = SharedPreferencesWindows();
```
stuartmorgan, line 22: Here and for Linux: This is technically a breaking change, even though hopefully nobody was using it. You should just mark it deprecated instead of removing it. (You should still remove all the comments, and remove the use below, just leave this one line.)

You can make it officially deprecated, not just a comment, since we already aren't using it. You'll just need to make the version bumps on those two minor instead of bugfix per semver.

stuartmorgan (changes requested):
Sorry, I missed a few things.

packages/shared_preferences/shared_preferences_android/test/shared_preferences_android_test.dart
```diff
@@ -0,0 +1,117 @@
...
+          return await testData.setValue(valueType, key, value);
+        }
+        fail('Unexpected method call: ${methodCall.method}');
+      });
+      store = SharedPreferencesAndroid();
+      log.clear();
+    });
+
+    tearDown(() async {
+      await testData.clear();
```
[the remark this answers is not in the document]
BeMacized: Looks to me like this was already the case, but the teardown was just redundant

packages/shared_preferences/shared_preferences_android/test/shared_preferences_android_test.dart
```diff
@@ -48,33 +49,30 @@ void main() {
         final Match? match = setterRegExp.matchAsPrefix(methodCall.method);
         if (match?.groupCount == 1) {
           final String valueType = match!.group(1)!;
-          final String key = (methodCall.arguments['key'] as String?)!;
-          final Object value = (methodCall.arguments['value'] as Object?)!;
+          final String key = methodCall.arguments['key'] as String;
+          final Object value = methodCall.arguments['value'] as Object;
```
stuartmorgan, line 53: Don't these need `!`s? I would expect you to get an analyzer warning about using `as` to convert from nullable to non-nullable.

packages/shared_preferences/shared_preferences_macos/pubspec.yaml
```diff
@@ -2,23 +2,25 @@ name: shared_preferences_macos
 description: macOS implementation of the shared_preferences plugin.
 repository: https://github.com/flutter/plugins/tree/main/packages/shared_preferences/shared_preferences_macos
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+shared_preferences%22
-version: 2.0.2
+version: 2.0.3
```
stuartmorgan, line 5: The versions need to be bumped to 2.1.0 still (as noted in the comment about deprecating methods).

stuartmorgan (changes requested):

packages/shared_preferences/shared_preferences_android/test/shared_preferences_android_test.dart
BeMacized: I wondered the same thing, but as it gave me no warnings left it out.

packages/shared_preferences/shared_preferences_macos/pubspec.yaml
BeMacized: Ah yes, I forgot about that. You tagged the version for the macos package here though. Shouldn't the minor only be bumped for the linux and windows packages as those are the only ones who had that field deprecated?
Windows & Linux packages have had their pubspec versions and changelog updated.

stuartmorgan: Sorry, that was carelessness as I was scrolling through.

packages/shared_preferences/shared_preferences_linux/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.1.0
+
+* Updated to make use of `dartPluginClass` only.
```
stuartmorgan, line 3: I'm not sure what this means; it was already only using `dartPluginClass`. Isn't the change below the only actual change?

packages/shared_preferences/shared_preferences_windows/example/integration_test/shared_preferences_test.dart
```diff
@@ -30,8 +30,8 @@ void main() {
 
     late SharedPreferencesWindows preferences;
 
-    setUp(() async {
-      preferences = SharedPreferencesWindows.instance;
+    setUpAll(() async {
```
stuartmorgan, line 33: This makes the object under test global for all tests, which is exactly what we don't want. At the least this should be reverted; preferably `late SharedPreferencesWindows preferences;` should be removed entirely and the test object created in each test.

stuartmorgan (changes requested):

stuartmorgan (approved):
LGTM, sorry I kept missing things in earlier review stages.

PR #4725, opened 2022-02-02
[video_player] Adjust iOS test timeouts

stuartmorgan:
The native iOS tests load network resources, so need longer timeouts to
avoid flake.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player_avfoundation/CHANGELOG.md
+++ b/packages/video_player/video_player_avfoundation/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Adjusts test timeouts for network-dependent native tests to avoid flake.
+
 ## 2.2.17
 
 * Splits from `video_player` as a federated implementation.
--- a/packages/video_player/video_player_avfoundation/example/ios/RunnerTests/VideoPlayerTests.m
+++ b/packages/video_player/video_player_avfoundation/example/ios/RunnerTests/VideoPlayerTests.m
@@ -127,7 +127,7 @@
                           [initializedExpectation fulfill];
                         }
                       }];
-  [self waitForExpectationsWithTimeout:1.0 handler:nil];
+  [self waitForExpectationsWithTimeout:30.0 handler:nil];
 
   // Starts paused.
   AVPlayer *avPlayer = player.player;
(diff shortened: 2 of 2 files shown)


jmagman (approved):
LGTM

jmagman:
Actually now `testDeregistersFromPlayer` is failing https://cirrus-ci.com/task/5913296061595648?logs=native_test#L1724

jmagman:
This timeout should probably also be increased:
https://github.com/flutter/plugins/blob/4ebfcb53b93144cbe4a7e848c992fb394a697641/packages/video_player/video_player_avfoundation/example/ios/RunnerTests/VideoPlayerTests.m#L69-L70

Hixie:
why do we have timeouts at all?

jmagman:
> why do we have timeouts at all?

The iOS and macOS "expectation" APIs all require timeout numbers:

> [- waitForExpectations:timeout:](https://developer.apple.com/documentation/xctest/xctestcase/2806856-waitforexpectations?language=objc)
> Waits for the test to fulfill a set of expectations within a specified time.
> [- waitForExpectations:timeout:enforceOrder:](https://developer.apple.com/documentation/xctest/xctestcase/2806857-waitforexpectations?language=objc)
> Waits for the test to satisfy an array of expectations and specifies whether they must occur in the array’s order.
> [- waitForExpectationsWithTimeout:handler:](https://developer.apple.com/documentation/xctest/xctestcase/1500748-waitforexpectationswithtimeout?language=objc)
> Waits until the test fulfills all expectations or until it times out.
https://developer.apple.com/documentation/xctest/xctestcase?language=objc

We could make that timeout the max double to infinitely wait, but I've never seen that pattern in these kinds of tests.

stuartmorgan:
> We could make that timeout the max double to infinitely wait, but I've never seen that pattern in these kinds of tests.

It makes failures really obnoxious. Locally it seems like a hang, and on CI it makes it take whatever the bot timeout is to get results. And it guarantees that you only learn about the first failure per run which is often frustrating.

There's no legitimate case where an expectation like this takes longer than N for some reasonable N. If it's only supposed to take a second or two, we don't need to let it run for an hour to know it's broken.

(N might be longer than 30 seconds, but I certainly wouldn't want to set it to more than a minute or two.)

PR #4726, opened 2022-02-03
[video_player] Updated Pigeon version

gaaclarke:
Updated `video_player_android` and `video_player_avfoundation` to use the latest version of pigeon.

Test exempt: Refactor, no new features.

Changes:
1) Pigeon now 1.0.x.
1) Constructors are now used in the Dart code.
1) Some generated files have the ".g." segment in their name (the java generator has bug where I couldn't add it, it would generate the class "Messages.g")

## Pre-launch Checklist

- [x] I signed the [CLA].


gaaclarke:
I didn't look into the async or null safety usage, just getting everything updated.  There were a few things that were had tweaked that we should fix in pigeon:
1) `unnecessary_parenthesis` linter failure in dart test file
1) `import` in dart test file is relative which messes the compiler up.  it thinks there are 2 instances of the types.
1) generating a file named "Foo.g.java" makes class called "Foo.g"

packages/video_player/video_player_avfoundation/ios/Classes/messages.g.m
```diff
@@ -0,0 +1,555 @@
...
+  pigeonResult.uri = uri;
+  pigeonResult.packageName = packageName;
+  pigeonResult.formatHint = formatHint;
+  pigeonResult.httpHeaders = httpHeaders;
+  return pigeonResult;
+}
++ (FLTCreateMessage *)fromMap:(NSDictionary *)dict {
+  FLTCreateMessage *pigeonResult = [[FLTCreateMessage alloc] init];
+  pigeonResult.asset = dict[@"asset"];
+  if ((NSNull *)pigeonResult.asset == [NSNull null]) {
```
stuartmorgan, line 204: Pigeon fix (so, not for this PR): a simpler solution occurs to me for this type issue than local variables. You can generate a static helper C function at the top of this file and use it for the lookups:

```
NSObject *GetNullableObjectFromDictionary(NSDictionary *dict, NSString *key) {
  NSObject *value = dict[key];
  return value == [NSNull null] ? nil : value;
}
```

Then fromMap is just a series of:
```
pigeonResult.foo = GetNullableObjectFromDictionary(@"foo");
...
```

statements. Shorter, and no type abuse.

packages/video_player/video_player_avfoundation/ios/Classes/messages.g.m
```diff
@@ -0,0 +1,555 @@
...
+  return [[FLTVideoPlayerApiCodecWriter alloc] initWithData:data];
+}
+- (FlutterStandardReader *)readerWithData:(NSData *)data {
+  return [[FLTVideoPlayerApiCodecReader alloc] initWithData:data];
+}
+@end
+
+NSObject<FlutterMessageCodec> *FLTVideoPlayerApiGetCodec() {
+  static dispatch_once_t s_pred = 0;
+  static FlutterStandardMessageCodec *s_sharedObject = nil;
```
stuartmorgan, line 335: Pigeon fix: These names violate style; it should be `sPred` and `sSharedObject`. Hungarian-style prefixes (`s`, `k`) in ObjC still use normal naming rules.

stuartmorgan (changes requested):

packages/video_player/video_player_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.3.0
+
+* Update Pigeon to ^1.0.16.
```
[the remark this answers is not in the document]
gaaclarke: fyi: The style guide just says it should be the present tense, it doesn't specify which conjugation should be used (ie 3rd person singular, neutral).

packages/video_player/video_player_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.3.0
```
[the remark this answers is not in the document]
gaaclarke: My thinking was that by sheer percentage of code this with a bigger fix than bumping the last component.  It's not really a refactor, a bugfix, or a new feature.  Let me know if you want me to just increment the patch number.

packages/video_player/video_player_android/pigeons/messages.dart
```diff
@@ -2,44 +2,51 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// @dart = 2.9
-
-import 'package:pigeon/pigeon_lib.dart';
+import 'package:pigeon/pigeon.dart';
 
+@ConfigurePigeon(PigeonOptions(
+  dartOut: 'lib/src/messages.g.dart',
```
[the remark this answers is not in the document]
gaaclarke: I started moving to that because it makes review a bit easier if you can see that it's generated in the path name.  It's also easier when developing in an IDE to quickly distinguish between "messages.dart" and "messages.g.dart".

packages/video_player/video_player_android/CHANGELOG.md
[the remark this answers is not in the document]
gaaclarke: done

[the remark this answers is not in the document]
stuartmorgan: It doesn't just say it should be in present tense, it says it should be "present tense indicative". I assumed that the fact the package was the implied subject was obvious, but I'll add that.

It also specifically says:
> For example, "Adds cool new feature.", not "Add" or "Added"

So I'm not clear how it's ambiguous that this should be "Updates" rather than "Update". But I'll add even more specific grammar notes there if you like.

packages/video_player/video_player_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.3.0
+
+* Update Pigeon to 1.0.x.
```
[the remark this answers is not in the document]
gaaclarke: done (conjugation)

packages/video_player/video_player_android/pigeons/messages.dart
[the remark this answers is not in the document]
stuartmorgan: That seems pretty subtle. How about `.generated.dart`? Or maybe `.pigeon.dart` as a parallel to `.mocks.dart` for the Mockito-generated code.

packages/video_player/video_player_avfoundation/ios/Classes/messages.g.h
```diff
@@ -1,79 +1,95 @@
...
-@interface FLTCreateMessage : NSObject
-@property(nonatomic, copy, nullable) NSString *asset;
-@property(nonatomic, copy, nullable) NSString *uri;
-@property(nonatomic, copy, nullable) NSString *packageName;
-@property(nonatomic, copy, nullable) NSString *formatHint;
-@property(nonatomic, strong, nullable) NSDictionary *httpHeaders;
-@end
-
 @interface FLTLoopingMessage : NSObject
++ (instancetype)makeWithTextureId:(nullable NSNumber *)textureId
```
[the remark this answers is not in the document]
gaaclarke: Ah shoot, good call.  The reason I didn't use an `init` method is because you can't use properties in there and I didn't want to generate special logic for talking to raw ivars.  "make" was chosen since it follows the create rule.  I was thinking there was some prior art for it but there isn't.  The problem is that renaming it is technically a breaking change.

packages/video_player/video_player_avfoundation/ios/Classes/messages.g.m
gaaclarke: Good call.  I'll try to do some cleanup on the few pigeon things I saw today.  I'll add that to the list.

packages/video_player/video_player_android/pigeons/messages.dart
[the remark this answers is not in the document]
gaaclarke: ".g.dart" is what is used for the JSON generator.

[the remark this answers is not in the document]
stuartmorgan: Ah, okay. WFM then

gaaclarke:
@stuartmorgan friendly ping, I think we are good here. I talked about updating pigeon but we've landed the fix in pigeon and we'll grab it the next time.  The differences are trivial.

packages/video_player/video_player_avfoundation/ios/Classes/messages.g.h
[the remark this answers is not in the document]
stuartmorgan: Filed https://github.com/flutter/flutter/issues/97832 to capture this.

packages/video_player/video_player_android/pigeons/messages.dart
```diff
@@ -2,44 +2,51 @@
...
+  dartTestOut: 'test/test_api.dart',
+  javaOut: 'android/src/main/java/io/flutter/plugins/videoplayer/Messages.java',
+  javaOptions: JavaOptions(
+    package: 'io.flutter.plugins.videoplayer',
+  ),
+  copyrightHeader: 'pigeons/copyright.txt',
+))
 class TextureMessage {
-  int textureId;
+  int? textureId;
```
stuartmorgan, line 17: Shouldn't `textureId` be non-null in all of these, now that that's supported?

packages/video_player/video_player_android/pigeons/messages.dart
```diff
@@ -2,44 +2,51 @@
...
 }
 
 class LoopingMessage {
-  int textureId;
-  bool isLooping;
+  int? textureId;
+  bool? isLooping;
```
stuartmorgan, line 22: In fact, shouldn't *all* of the values in these classes be non-null?

packages/video_player/video_player_android/pigeons/messages.dart
```diff
@@ -2,44 +2,51 @@
...
-  String asset;
-  String uri;
-  String packageName;
-  String formatHint;
-  Map<String, String> httpHeaders;
+  String? asset;
+  String? uri;
+  String? packageName;
+  String? formatHint;
+  Map<String?, String?>? httpHeaders;
```
stuartmorgan, line 45: This should definitely be non-null so that we don't have the nil-vs-empty confusion.

packages/video_player/video_player_android/pigeons/messages.dart
```diff
@@ -2,44 +2,51 @@
...
 class CreateMessage {
-  String asset;
-  String uri;
-  String packageName;
-  String formatHint;
-  Map<String, String> httpHeaders;
+  String? asset;
+  String? uri;
+  String? packageName;
+  String? formatHint;
```
stuartmorgan, line 44: Except for some of these, which are nullable in `DataSource`.

packages/video_player/video_player_avfoundation/pigeons/messages.dart
```diff
@@ -2,68 +2,76 @@
...
+  dartOut: 'lib/src/messages.g.dart',
+  dartTestOut: 'test/test_api.dart',
+  objcHeaderOut: 'ios/Classes/messages.g.h',
+  objcSourceOut: 'ios/Classes/messages.g.m',
+  objcOptions: ObjcOptions(
+    prefix: 'FLT',
+  ),
+  copyrightHeader: 'pigeons/copyright.txt',
+))
 class TextureMessage {
```
stuartmorgan, line 17: Same nullability notes throughout this file as on Android.

stuartmorgan (changes requested):

packages/video_player/video_player_android/pigeons/messages.dart
gaaclarke: I mentioned that in my first comment I didn't investigate async or nullability changes.  I just want this PR to be an upgrade of pigeon, no semantics changes.

gaaclarke:
Hey @stuartmorgan , I wanted to address nullability and async changes in a different PR in order to separate the big pigeon upgrade from functionality changes.

stuartmorgan:
I understand why async changes would be separate since they involve rewriting logic, but I don't understand why nullability would be separate. A core part of this PR is that you are upgrading `messages.dart` from pre-NNBD to post-NNBD. When we do NNBD updates, actually using the correct nullability semantics is a core part of that.

In fact, I would expect that not adding a bunch of unwanted nullability annotations would make this diff *smaller*.

packages/video_player/video_player_android/pigeons/messages.dart
stuartmorgan: > - // @dart = 2.9

This is a semantics change though.

gaaclarke: It shouldn't be.  In Pigeon if you switch to NNBD then make all types nullable the semantics are the same, the syntax is changed only.  Pigeon is special since it doesn't allow logic, just declarations.

gaaclarke:
> In fact, I would expect that not adding a bunch of unwanted nullability annotations would make this diff smaller.

Yea, probably.  It's just that was explicitly listed as a non-goal of this PR in order to change less variables.  Let me queue up another PR for you.

gaaclarke:
@stuartmorgan I have a draft prepared for updating the nullability: https://github.com/flutter/plugins/pull/4742  I don't know if it's easy to see but there was code that had to be changed to accommodate the new semantics that didn't have to get changed in this PR since they were the same.

stuartmorgan:
> It's just that was explicitly listed as a non-goal of this PR in order to change less variables.

I understand that you wrote it down; that doesn't mean that it's actually desirable to have done so. Every single NNBD upgrade we did across flutter/plugin and flutter/packages involved auditing the *desired* nullability of each item and changing the code accordingly, not blindly adding `?`s everywhere to preserve semantics. Preserving semantics is a non-goal of an NNBD migration.

There is no reason to make a Pigeon update different in that regard. We should not migrate a file from non-NNBD to NNBD without giving it the desired nullability semantics.

> I don't know if it's easy to see but there was code that had to be changed to accommodate the new semantics

Yes, that's a normal part of an NNBD upgrade. (And it appears that they were incredibly minor, and primarily consistent of removing one `!` in each package which is entirely positive and trivial to reason about.)

stuartmorgan (changes requested):
(Updating review status to reflect comments above, since I forgot to do that as a review)

gaaclarke:
@stuartmorgan This isn't an NNBD upgrade, I know it looks like one but it isn't.  We are only using Dart for the front-end compiler, not the backend.  I've written the whole backend, this is not changing the semantics or the generated code.  It's just changing how the Dart front-end works.  We really should separate the change in semantics into a different PR.

stuartmorgan:
> This isn't an NNBD upgrade

I'm not talking about the package; I know that the package has already been converted. I'm talking about messages.dart, where you are making the following change:

```dart
-// @dart = 2.9
```

> this is not changing the semantics or the generated code

I am aware that it is not changing the generated code, but it *is* changing the semantics of everything in that file from "we aren't making any claims about the intended nullability of these values" to "we are explicitly claiming that all of these things are intended to be nullable". Those new semantics are wrong, because those things are not all intended to be nullable.

I'm not going to approve a PR that explicitly puts the wrong (as defined by the desired behavior of the API layer) annotations in messages.dart just to preserve behavior that we don't want, any more than I would have approved a package-level NNBD update that did the same thing to the rest of the code.

gaaclarke:
> "we aren't making any claims about the intended nullability of these values"

I think that's where our disconnect is.  That isn't what things meant before NNBD.  It meant "all of my values are nullable".  So the smallest change for upgrading would have been to keep those semantics "all of my values are nullable".

gaaclarke:
I talked offline with stuart.  Moving the fields from implicitly nullable to explicitly nullable means null values should be accommodated in the handlers.  Instead of going through the work to make that happen it's easier just to do the migration to nonnull values now when that was their intended usage.  So, while this is 2 big changes in one PR, it's less contradictory and more clear to land it in one PR.

gaaclarke:
@stuartmorgan I've adopted the null-safety annotations that match the usage.  Everything is non-null except for the fields in CreateMessage because that's how they are created in the dart side: https://github.com/flutter/plugins/blob/9d32c83af9af125981ef847abd9d1636b65aa8e1/packages/video_player/video_player_android/lib/src/android_video_player.dart#L36  and how they are used on the host side: https://github.com/flutter/plugins/blob/9d32c83af9af125981ef847abd9d1636b65aa8e1/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java#L84

packages/video_player/video_player_avfoundation/pigeons/messages.dart
```diff
@@ -2,68 +2,82 @@
...
-  String asset;
-  String uri;
-  String packageName;
-  String formatHint;
-  Map<String, String> httpHeaders;
+  String? asset;
+  String? uri;
+  String? packageName;
+  String? formatHint;
+  Map<String?, String?>? httpHeaders;
```
stuartmorgan, line 51: Same.

stuartmorgan (changes requested):

packages/video_player/video_player_avfoundation/pigeons/messages.dart
gaaclarke: Done.  FYI they were nonnull in the DataSource class, but in both the android implementation and the ios implementation they were nullable.

stuartmorgan: Yes, but the fact that it's traditionally been hard to reason about nullability across the platform channel boundary—leading to code that is either over-cautious like in this case, or not cautious enough and buggy as in other cases—is exactly why I think Pigeon NNBD support is a substantial opportunity.

packages/video_player/video_player_android/pigeons/messages.dart
```diff
@@ -43,11 +43,12 @@ class PositionMessage {
 }
 
 class CreateMessage {
+  CreateMessage({required this.httpHeaders});
```
stuartmorgan, line 46: Why is this required, rather than defaulting to an empty collection?

packages/video_player/video_player_avfoundation/ios/Classes/FLTVideoPlayerPlugin.m
```diff
@@ -177,7 +177,7 @@ - (void)createVideoOutputAndDisplayLink:(FLTFrameUpdater *)frameUpdater {
 
 - (instancetype)initWithURL:(NSURL *)url
                frameUpdater:(FLTFrameUpdater *)frameUpdater
-                httpHeaders:(NSDictionary<NSString *, NSString *> *)headers {
+                httpHeaders:(nonnull NSDictionary<NSString *, NSString *> *)headers {
   NSDictionary<NSString *, id> *options = nil;
   if (headers != nil && [headers count] != 0) {
```
stuartmorgan, line 182: The `headers != nil` can be removed.

(Technically it could have been removed either way since `[nil count]` is zero, but it should definitely be removed now.)

packages/video_player/video_player_android/pigeons/messages.dart
gaaclarke: Pigeon demands there be no logic here, it doesn't generate code for initialization.

packages/video_player/video_player_avfoundation/ios/Classes/FLTVideoPlayerPlugin.m
gaaclarke: done

stuartmorgan (approved):
LGTM given the current Pigeon limitations.

PR #4760, opened 2022-02-08
[url_launcher] Workaround for url_launcher_web 2.0.7 dependency on Flutter < 2.10.0

carlogrisetti:
This PR fixes the use of url_launcher in a web deployment using Flutter <2.10.0
With Flutter 2.5.3 (for example) the maximum compatible version reported is url_launcher_web 2.0.7, which still doesn't require Flutter 2.10.0+, but this is a broken requirement chain, since the correct dependency is inserted with url_launcher_web 2.0.8

IMHO, url_launcher_web 2.0.7 should be retired\yanked in the future, allowing to restore the ^2.0.0 version specifier

There is no current issue for this, since this is a breaking change introduced less than 5 days ago (and I am not able to create a new one)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/url_launcher/url_launcher/pubspec.yaml
+++ b/packages/url_launcher/url_launcher/pubspec.yaml
@@ -33,7 +33,7 @@ dependencies:
   url_launcher_linux: ^2.0.0
   url_launcher_macos: ^2.0.0
   url_launcher_platform_interface: ^2.0.3
-  url_launcher_web: ^2.0.0
+  url_launcher_web: '>=2.0.0 <2.0.7'  # 2.0.7 requires Flutter 2.10.0+ but it is not specified
   url_launcher_windows: ^2.0.0
 
 dev_dependencies:


stuartmorgan:
Thanks for the contribution, but this is not the correct way to address the problem with 2.0.7. See discussion in https://github.com/flutter/flutter/issues/97796

> IMHO, url_launcher_web 2.0.7 should be retired\yanked

It was, but you are choosing to use a version of Flutter that is too old to support package retraction.

carlogrisetti:
Damn... I came to the same conclusion (not the way to fix this) on the way home. Even if this PR were to be merged it would have done so in 2.0.9, this leaving 2.0.7 as-is

Thanks for the tip on package retraction. I didn't know this was a feature only available after a certain version of Flutter.

I'll push one more time to upgrade Flutter internally, but it has not been an accepted path recently.
Thanks!

PR #4780, opened 2022-02-09
[camera]remove "selfRef" for SavePhotoDelegate and ensure thread safety

hellohuanlin:
As discussed in [the proposal](https://docs.google.com/document/d/1Znjp-BduXIb1Ix3JOE1YKyJi1Z0d6rB-2q6Lt7j8xSo/edit#): 

> selfReference is used to keep FLTSavePhotoDelegate around. This is a nice trick, but should only be used as a workaround. In our case this delegate object has a clear owner - the delegating object (FLTCam), which controls the lifecycle of this delegate. 

The main idea is to have a map of refs in FLTCam (`inProgressSavePhotoDelegates`), and delete the ref in completion handler. We need to use the map (instead of a single pointer) because photo capture may overlap (e.g. live photos, though i don't think we support live photos yet)

Also fixes a thread safety issue when updating the delegate ref. Fix it together as it's an easy fix. 

Both issues are tracked here: https://github.com/flutter/flutter/issues/96429

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
```diff
@@ -0,0 +1,172 @@
...
+  // settings.uniqueID is used as the key for `inProgressSavePhotoDelegates` dictionary
+  AVCapturePhotoSettings *settings = [AVCapturePhotoSettings photoSettings];
+  id mockSettings = OCMClassMock([AVCapturePhotoSettings class]);
+  OCMStub([mockSettings photoSettings]).andReturn(settings);
+
+  // We need to make sure the delegate reference is actually saved in a real dictionary, so that we
+  // can call its completion handler later. Must use a partial mock, in order to forward invocation
+  // to the real object.
+  id mockDelegates = OCMPartialMock([NSMutableDictionary dictionary]);
+  OCMStub([mockDelegates setObject:OCMOCK_ANY forKeyedSubscript:OCMOCK_ANY])
```
jmagman, line 40: This seems like it's mocking too much of the underlying implementation.  Can you change it to validate a higher level: that when the photo is successfully taken, success or failure message is sent back to the engine?

hellohuanlin: This test is to verify that references are accessed on the right queue. The success/failure tests are the next 2 tests below.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.h
```diff
@@ -4,34 +4,35 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
+/// The completion handler block for save photo operations.
+/// Can be called from either main queue or IO queue.
+/// If success, `error` will be present and `path` will be nil. Otherewise, `error` will be nil and
+/// `path` will be present.
+/// @param error photo capture error or IO error.
+/// @param path the path for successfully saved photo file.
+typedef void (^FLTSavePhotoDelegateCompletionHandler)(NSError *_Nullable error,
+                                                      NSString *_Nullable path);
```
stuartmorgan, line 19: Nit: Usually errors are the last parameter in ObjC.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.m
```diff
@@ -3,37 +3,41 @@
 // found in the LICENSE file.
 
 #import "FLTSavePhotoDelegate.h"
+#import "FLTSavePhotoDelegate_Test.h"
+
+@interface FLTSavePhotoDelegate ()
+/// The file path for the captured photo.
+@property(readonly, nonatomic) NSString *path;
+/// The queue on which captured photos are wrote to disk.
```
stuartmorgan, line 11: Nit: written

packages/camera/camera/ios/Classes/FLTCam.m
```diff
@@ -220,11 +220,27 @@ - (void)captureToFile:(FLTThreadSafeFlutterResult *)result API_AVAILABLE(ios(10)
...
-                      delegate:[[FLTSavePhotoDelegate alloc] initWithPath:path
-                                                                   result:result
-                                                                  ioQueue:self.photoIOQueue]];
+  FLTSavePhotoDelegate *savePhotoDelegate = [[FLTSavePhotoDelegate alloc]
+           initWithPath:path
+                ioQueue:self.photoIOQueue
+      completionHandler:^(NSError *_Nullable error, NSString *_Nullable path) {
+        dispatch_async(self.captureSessionQueue, ^{
+          // Dispatch back to capture session queue to delete reference.
+          // Retain cycle is broken after the dictionary entry is cleared.
```
stuartmorgan, line 229: Why do we need a retain cycle in the first place? Do we actually want to deliver the results still if the camera has been destroyed?

packages/camera/camera/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
```diff
@@ -0,0 +1,172 @@
...
+      .andDo(^(NSInvocation *invocation) {
+        if (dispatch_get_specific(captureSessionQueueSpecific)) {
```
stuartmorgan, line 42: Why is all of this mocking necessary; couldn't you do essentially this same logic in the code itself with `NSAssert`s (instead of comments saying what the current queue is), and then have the unit test just exercise the actual codepath?

packages/camera/camera/ios/Classes/FLTCam.m
hellohuanlin: This was to keep the original behavior with `selfReference`, which is deallocated only after completion, even after camera has been destroyed. We can change the behavior and use `weak self` here.

stuartmorgan: I'm fine with preserving the behavior for now to minimize risk if we're not sure. It just seems odd to guarantee that callbacks would fine even on a deleted camera.

packages/camera/camera/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
hellohuanlin: Wow thank you for this explanation. It was very helpful! 

I will delete some of those "queue-verifying" unit tests for the ones that did not cause externally visible bugs. As long as they have other tests that goes through `NSAssert`, we should be confident.

packages/camera/camera/ios/Classes/QueueHelper.m
```diff
@@ -4,12 +4,26 @@
...
+
 + (void)ensureToRunOnMainQueue:(void (^)(void))block {
   if (!NSThread.isMainThread) {
     dispatch_async(dispatch_get_main_queue(), block);
   } else {
     block();
   }
 }
+
++ (dispatch_queue_t)createQueueWithLabel:(const char *)label specific:(const char *)specific {
```
jmagman, line 19: I generally don't like utility helpers that are only a few lines.  It essentially makes developers who are probably already familiar with gcd APIs now have to do the mental jump of understanding what the camera `createQueueWithLabel` method is doing.  This wraps a well-known standard library APIs around a bespoke helper method that doesn't do much and is more the letter than the spirit of DRY.

jmagman (commented):
The mocking looks much better.

packages/camera/camera/ios/Classes/QueueHelper.m
hellohuanlin: I don't have a strong opinion here, so will update. 

On the one hand, I think the benefit of `createQueueWithLabel:specific` util function is that it combines 2 calls, so that the caller site won't forget to set the queue specific data; And the benefit of `isCurrentlyOnQueueWithSpecific:` util function is that it makes the assertion readable like english sentence (e.g. `NSAssert([QueueHelper isCurrentlyOnQueueWithSpecific:fooSpec]) vs NSAssert(dispatch_get_specific(fooSpec))`

But on the other hand you also made a very good point that these are 2 extra functions that developers have to learn.

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.m
```diff
@@ -3,37 +3,41 @@
 // found in the LICENSE file.
 
 #import "FLTSavePhotoDelegate.h"
+#import "FLTSavePhotoDelegate_Test.h"
+
+@interface FLTSavePhotoDelegate ()
+/// The file path for the captured photo.
+@property(readonly, nonatomic) NSString *path;
```
jmagman, line 10:
```suggestion
@property(readonly, nonatomic, copy) NSString *path;
```

packages/camera/camera/ios/Classes/FLTCam.m
```diff
@@ -220,11 +221,30 @@ - (void)captureToFile:(FLTThreadSafeFlutterResult *)result API_AVAILABLE(ios(10)
...
+  FLTSavePhotoDelegate *savePhotoDelegate = [[FLTSavePhotoDelegate alloc]
+           initWithPath:path
+                ioQueue:self.photoIOQueue
+      completionHandler:^(NSString *_Nullable path, NSError *_Nullable error) {
+        dispatch_async(self.captureSessionQueue, ^{
+          // Dispatch back to capture session queue to delete reference.
+          // Retain cycle is broken after the dictionary entry is cleared.
+          // This is to keep the behavior with the previous `selfReference` approach in the
+          // FLTSavePhotoDelegate, where delegate is released only after capture completion.
+          self.inProgressSavePhotoDelegates[@(settings.uniqueID)] = nil;
```
jmagman, line 233: Or this, now that you're not mocking the dictionary subscripting. Not important though.
```suggestion
          [self.inProgressSavePhotoDelegates removeObjectForKey:@(settings.uniqueID)];
```

jmagman (approved):
LGTM

packages/camera/camera/ios/Classes/FLTSavePhotoDelegate.m
hellohuanlin: oh this is already `readonly` so i leave out `copy`

packages/camera/camera/ios/Classes/CameraPlugin.m
```diff
@@ -37,7 +38,10 @@ - (instancetype)initWithRegistry:(NSObject<FlutterTextureRegistry> *)registry
   NSAssert(self, @"super init cannot be nil");
   _registry = [[FLTThreadSafeTextureRegistry alloc] initWithTextureRegistry:registry];
   _messenger = messenger;
-  _captureSessionQueue = dispatch_queue_create("io.flutter.camera.captureSessionQueue", NULL);
+  _captureSessionQueue = dispatch_queue_create(io.flutter.camera.captureSessionQueue, NULL);
```
jmagman, line 41:
```
error: use of undeclared identifier 'io'
      _captureSessionQueue = dispatch_queue_create(io.flutter.camera.captureSessionQueue, NULL);
```

hellohuanlin: haha

PR #4782, opened 2022-02-09
Adds support for JavaScript channels to the new iOS implementation of `webview_flutter_wkwebview`.

bparrishMines:
Adds support for JavaScript channels to the new iOS implementation. 

No version change:
Part of https://github.com/flutter/flutter/issues/93732 and doesn't make any changes to the current implementation.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
+++ b/packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
@@ -2,6 +2,23 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+
+/// Times at which to inject script content into a webpage.
+///
+/// Wraps [WKUserScriptInjectionTime](https://developer.apple.com/documentation/webkit/wkuserscriptinjectiontime?language=objc).
+enum WKUserScriptInjectionTime {
+  /// Inject the script after the creation of the webpage’s document element, but before loading any other content.
+  ///
+  /// See https://developer.apple.com/documentation/webkit/wkuserscriptinjectiontime/wkuserscriptinjectiontimeatdocumentstart?language=objc.
+  atDocumentStart,
+
+  /// Inject the script after the document finishes loading, but before loading any other subresources.
+  ///
+  /// See https://developer.apple.com/documentation/webkit/wkuserscriptinjectiontime/wkuserscriptinjectiontimeatdocumentend?language=objc.
+  atDocumentEnd,
+}
+
 /// The media types that require a user gesture to begin playing.
 ///
 /// Wraps [WKAudiovisualMediaTypes](https://developer.apple.com/documentation/webkit/wkaudiovisualmediatypes?language=objc).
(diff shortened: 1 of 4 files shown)


packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit_webview_widget.dart
```diff
@@ -140,18 +146,89 @@ class WebKitWebViewPlatformController extends WebViewPlatformController {
       if (!requiresUserAction) WKAudiovisualMediaType.none,
     };
   }
+
+  @override
+  Future<void> addJavascriptChannels(Set<String> javascriptChannelNames) async {
```
bparrishMines, line 151: Relevant code from current implementation: https://github.com/flutter/plugins/blob/main/packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m#L424

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit_webview_widget.dart
```diff
@@ -140,18 +146,89 @@ class WebKitWebViewPlatformController extends WebViewPlatformController {
...
+            handler,
+            channelName,
+          );
+        },
+      ),
+    );
+  }
+
+  @override
+  Future<void> removeJavascriptChannels(
```
bparrishMines, line 194: relevant code from current implementation: https://github.com/flutter/plugins/blob/main/packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m#L433

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit_webview_widget.dart
```diff
@@ -78,15 +78,19 @@ class WebKitWebViewPlatformController extends WebViewPlatformController {
     required this.callbacksHandler,
     required this.javascriptChannelRegistry,
     WKWebViewConfiguration? configuration,
-    @visibleForTesting this.webViewProxy = const WebViewProxy(),
+    @visibleForTesting this.webViewProxy = const WebViewWidgetProxy(),
   }) : super(callbacksHandler) {
     _setCreationParams(
       creationParams,
-      configuration: configuration ?? WKWebViewConfiguration(),
-    ).then((_) => _initializationCompleter.complete());
```
bparrishMines, line 86: This method doesn't actually wait for any `Future` and  `_initializationCompleter` is not used. I missed this in the PR before this one.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit_webview_widget.dart
```diff
@@ -78,15 +78,19 @@ class WebKitWebViewPlatformController extends WebViewPlatformController {
...
+      configuration: configuration ??
+          WKWebViewConfiguration(
+            userContentController: WKUserContentController(),
+          ),
+    );
   }
 
-  final Completer<void> _initializationCompleter = Completer<void>();
```
bparrishMines, line 89: unused

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit_webview_widget.dart
```diff
@@ -140,18 +146,89 @@ class WebKitWebViewPlatformController extends WebViewPlatformController {
...
+    return addJavascriptChannels(remainingNames);
+  }
 }
 
-/// Handles constructing [WKWebView]s and calling static methods.
+/// Handles constructing objects and calling static methods.
 ///
 /// This should only be used for testing purposes.
 @visibleForTesting
-class WebViewProxy {
```
bparrishMines, line 149: Name change since this class constructs more than just `WKWebView`.

packages/webview_flutter/webview_flutter_wkwebview/test/src/web_kit_webview_widget_test.dart
```diff
@@ -49,7 +59,6 @@ void main() {
       CreationParams? creationParams,
       bool hasNavigationDelegate = false,
       bool hasProgressTracking = false,
-      bool useHybridComposition = false,
```
bparrishMines, line 52: unused

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -27,10 +42,149 @@ enum WKAudiovisualMediaType {
...
+  /// The body of the message.
+  ///
+  /// Allowed types are [num], [String], [List], [Map], and `null`.
+  final Object? body;
+}
+
+/// An interface for receiving messages from JavaScript code running in a webpage.
+///
+/// Wraps [WKScriptMessageHandler](https://developer.apple.com/documentation/webkit/wkscriptmessagehandler?language=objc)
+class WKScriptMessageHandler {
```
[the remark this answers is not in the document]
bparrishMines: I like this idea since it would make this simpler and easier to test. But my main concern would be that this protocol could eventually be updated with a new method, then we would have to convert it back to a class. Although this would be relatively trivial to do; ~50 line change.

So I wonder if you think it's better to play it safe and avoid larger changes in the future or use our intuition to convert one-method classes/protocols to typedefs. Java uses a lot of one method interfaces, so this policy could definitely apply to the Android code as well.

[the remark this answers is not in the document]
stuartmorgan: I was thinking that since it was entirely internal it wouldn't be a big deal if we had to convert it to a class (if it were public I'd definitely say we should keep this version). But I guess in cases where we did have to do that conversion later it would undermine your design goal of this approach of putting the language boundary as close to the OS APIs as possible in order to make updating it for small OS API changes a simple, mechanical change.

I guess it really becomes a gamble about whether we think new methods are likely or not. It's tricky since the pattern of using protocols for delegation, rather than blocks, is baked pretty deeply into ObjC norms, so the existence of a protocol isn't really an indication of whether the designers expect to add more methods.

I'll leave it to you to decide since I think there are good arguments both ways here, and the main deciding factor is unknowable.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit_webview_widget.dart
```diff
@@ -97,22 +101,24 @@ class WebKitWebViewPlatformController extends WebViewPlatformController {
...
+  }) {
     _setWebViewConfiguration(
       configuration,
       allowsInlineMediaPlayback: params.webSettings?.allowsInlineMediaPlayback,
       autoMediaPlaybackPolicy: params.autoMediaPlaybackPolicy,
     );
 
     webView = webViewProxy.createWebView(configuration);
+
+    addJavascriptChannels(params.javascriptChannelNames);
```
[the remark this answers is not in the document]
bparrishMines: Ah, good catch. I forgot I added this method. I changed the method to be async again. However, its only use is in the constructor so it won't be awaited.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit_webview_widget.dart
```diff
@@ -140,18 +146,89 @@ class WebKitWebViewPlatformController extends WebViewPlatformController {
...
+    // exposing support for custom scripts from applications.
+    webView.configuration.userContentController.removeAllUserScripts();
+    webView.configuration.userContentController
+        .removeAllScriptMessageHandlers();
+
+    javascriptChannelNames.forEach(_scriptMessageHandlers.remove);
+    final Set<String> remainingNames = _scriptMessageHandlers.keys.toSet();
+    _scriptMessageHandlers.clear();
+
+    return addJavascriptChannels(remainingNames);
```
stuartmorgan, line 213: Nit: Since this is a (Future) void function, wouldn't it be more idiomatic to `await` rather than `return` here?

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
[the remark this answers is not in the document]
bparrishMines: My preference is to play it safe so that it can remain easy to maintain for future maintainers and community developers. So I'll keep it as is for now.

PR #4803, opened 2022-02-11
[ios_platform_images] Support loading system images

cadenkriese:
I've implemented a method to retrieve system images known as 'sfsymbols' as one of the first steps in implementing https://flutter.dev/go/sf-symbols. At the same time, I've updated the example app to contain a demo, as well as added integration tests.

*Issues are fixed by this PR.*
flutter/flutter#82208
flutter/flutter#60034 (partially)

## Pre-launch Checklist

- [x] I signed the [CLA].
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy].
- [x] I added new tests to check the change I am making.


packages/ios_platform_images/lib/ios_platform_images.dart
```diff
@@ -138,6 +138,74 @@ class IosPlatformImages {
...
+  /// Multi-layer and multi-color symbols only work on iOS 15.0 and above.
+  /// **This method requires at least iOS 13.0**
+  ///
+  /// See [https://developer.apple.com/documentation/uikit/uiimage/configuring_and_displaying_symbol_images_in_your_ui?language=objc]
+  static ImageProvider loadSystemImage(
+    String name,
+    double pointSize, {
+    List<Color> colors = const <Color>[],
+    bool preferMulticolor = false,
+    int weightIndex = 1,
```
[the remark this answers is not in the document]
cadenkriese: My philosophy was to leave any widget-specific stuff to the eventual implementation in a separate `sfsymbols` package, which would have access to the BuildContext and the theme information. I imagined this as more of a bare-bones method, so I came up with some basic defaults.

That's just what made sense to me as I was writing it, but I could see an argument for getting the build context here.

packages/ios_platform_images/example/lib/main.dart
```diff
@@ -24,16 +24,47 @@ class _MyAppState extends State<MyApp> {
...
+              ),
+              Image(
+                image: IosPlatformImages.loadSystemImage(
+                  'ladybug.fill',
+                  100,
+                  preferMulticolor: true,
+                ),
+                semanticLabel: 'Ladybug',
+              ),
+            ],
```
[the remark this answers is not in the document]
cadenkriese: I tried to implement this in [c14cd6e](https://github.com/flutter/plugins/pull/4803/commits/c14cd6ec6c549c192195ae40204264c4011022a2), the test passes but prints an error message after the fact. Not sure if that's acceptable, so I'm not marking this as resolved quite yet.

packages/ios_platform_images/lib/ios_platform_images.dart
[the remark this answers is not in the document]
jmagman: > My philosophy was to leave any widget-specific stuff to the eventual implementation in a separate `sfsymbols` package, which would have access to the BuildContext and the theme information. I imagined this as more of a bare-bones method, so I came up with some basic defaults.
> 
> That's just what made sense to me as I was writing it, but I could see an argument for getting the build context here.

If someone didn't use your eventual-sfsymbols package, is there any advantage to making the user think in terms of `weightIndex` instead of `FontWeight`, or similar Flutter idioms?  It's importing Flutter (it's not a generic dart package) so may as well use the higher level abstractions Flutter provides.

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.m
```diff
@@ -40,6 +40,115 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
+          } else {
+            colors = [UIImageSymbolConfiguration configurationWithPaletteColors:colorArray];
+          }
+
+          UIImageSymbolConfiguration *final =
+              [pointSizeConfig configurationByApplyingConfiguration:colors];
+          finalImage = [UIImage systemImageNamed:name withConfiguration:final];
+        } else {
+          UIImage *image = [UIImage systemImageNamed:name withConfiguration:pointSizeConfig];
+          finalImage = [image imageWithTintColor:colorArray[0]];
```
jmagman, line 138: What happens when `colorArray` is empty (the default)?

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.m
```diff
@@ -40,6 +40,115 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
+        UIImage *finalImage;
+
+        if (@available(iOS 15, *)) {
+          NSNumber *preferMulticolor = call.arguments[5];
+          UIImageSymbolConfiguration *colors;
+
+          if ([preferMulticolor boolValue]) {
+            colors = [UIImageSymbolConfiguration configurationPreferringMulticolor];
+          } else {
+            colors = [UIImageSymbolConfiguration configurationWithPaletteColors:colorArray];
```
jmagman, line 130: What does this do if the array is empty?  And is the renderingMode set up correctly (maybe it is when you use  `-systemImageNamed:withConfiguration:` I've never used that API though)?

https://developer.apple.com/documentation/uikit/uiimagesymbolconfiguration/3810054-configurationwithpalettecolors?language=objc it says 

> For this color configuration to have an effect, your symbol image must have the following:
> - Its [renderingMode](https://developer.apple.com/documentation/uikit/uiimage/1624122-renderingmode?language=objc) set to [UIImageRenderingModeAlwaysTemplate](https://developer.apple.com/documentation/uikit/uiimagerenderingmode/uiimagerenderingmodealwaystemplate?language=objc) or [UIImageRenderingModeAutomatic](https://developer.apple.com/documentation/uikit/uiimagerenderingmode/uiimagerenderingmodeautomatic?language=objc).
> - Hierarchical layer annotations. If your symbol doesn’t have hierarchical layer annotations, the resulting image is a monochrome (template) symbol image.

packages/ios_platform_images/lib/ios_platform_images.dart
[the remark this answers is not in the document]
cadenkriese: Okay, this is a valid point I agree with; I've been trying to implement this across the board, but I've come to an issue regarding the size. From my understanding Flutter `iconSize` is in logical pixels but the SF Symbols take a font point size. 

I've spent the weekend trying to figure out how to convert from pixels to font size, but I couldn't find a reliable method. How would you recommend approaching this?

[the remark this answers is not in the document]
jmagman: Maybe pass in the build context and fetch `MediaQuery.of(context).devicePixelRatio`?

@goderbauer @Piinks this isn't in the `flutter` package but this feels framework-y, can you make a suggestion for how to make this API more idiomatic looking?

packages/ios_platform_images/lib/ios_platform_images.dart
```diff
@@ -138,6 +138,74 @@ class IosPlatformImages {
     return _FutureMemoryImage(bytesCompleter.future, scaleCompleter.future);
   }
 
+  /// Loads an image from the system.  The equivalent would be:
+  /// `[UIImage systemImageNamed:name]`.
+  ///
+  /// Throws an exception if the image can't be found.
+  ///
+  /// [colors] takes a list of colors to be applied to the corresponding layer
```
goderbauer, line 146: nit (here and elsewhere)L According to flutter style docs should be grammatically correct and therefore this sentence should start with a capital letter (may have to rephrase it a bit to not have it start with [color]).

packages/ios_platform_images/lib/ios_platform_images.dart
```diff
@@ -138,6 +138,74 @@ class IosPlatformImages {
...
+  /// of the icon, assuming that layer exists. An icon can have up to 3 layers,
+  /// if a layer color is not specified it will take the color of the most recent
+  /// valid index. Defaults to an empty list (black).
+  ///
+  /// For more information see [https://developer.apple.com/documentation/uikit/uiimagesymbolconfiguration/3810054-configurationwithpalettecolors?language=objc]
+  ///
+  /// [preferMulticolor] overrides [colors] and asks iOS to provide its preset
+  /// multicolor varient of the symbol. Depending on the symbol, these colors
+  /// *may not be mutable*. To find out, use the SF Symbols app found at
+  /// [https://developer.apple.com/sf-symbols/]
```
goderbauer, line 156: nit: missing `.` at end of sentence

packages/ios_platform_images/lib/ios_platform_images.dart
```diff
@@ -138,6 +138,74 @@ class IosPlatformImages {
...
+  /// Also note that regardless of mutability, the symbols default colors are
+  /// affected by the devices theme preference.
+  ///
+  /// Multi-layer and multi-color symbols only work on iOS 15.0 and above.
+  /// **This method requires at least iOS 13.0**
+  ///
+  /// See [https://developer.apple.com/documentation/uikit/uiimage/configuring_and_displaying_symbol_images_in_your_ui?language=objc]
```
goderbauer, line 163: nit: document all params (pointSize and weight appear to be missing).

packages/ios_platform_images/lib/ios_platform_images.dart
```diff
@@ -138,6 +138,74 @@ class IosPlatformImages {
...
+  /// Throws an exception if the image can't be found.
+  ///
+  /// [colors] takes a list of colors to be applied to the corresponding layer
+  /// of the icon, assuming that layer exists. An icon can have up to 3 layers,
+  /// if a layer color is not specified it will take the color of the most recent
+  /// valid index. Defaults to an empty list (black).
+  ///
+  /// For more information see [https://developer.apple.com/documentation/uikit/uiimagesymbolconfiguration/3810054-configurationwithpalettecolors?language=objc]
+  ///
+  /// [preferMulticolor] overrides [colors] and asks iOS to provide its preset
```
goderbauer, line 153: Does this mean if preferMulticolor is set, colors should never be provided? If so, add an assert that encodes that.

packages/ios_platform_images/lib/ios_platform_images.dart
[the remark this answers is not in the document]
goderbauer: For a low-level image loading API this API seems fine. I don't think I would pass in the build context here to retain the flexibility of just using this in the render layer directly (where you don't have a build context). However, it would be nice to have a higher-level wrapper API (maybe a widget) that does read some of these things from e.g. a theme.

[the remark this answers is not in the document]
cadenkriese: > For a low-level image loading API this API seems fine. I don't think I would pass in the build context here to retain the flexibility of just using this in the render layer directly (where you don't have a build context). However, it would be nice to have a higher-level wrapper API (maybe a widget) that does read some of these things from e.g. a theme.

Fair enough, that higher-level API will be my next PR after this one in a separate package; the discussion is on https://flutter.dev/go/sfsymbols. Do you have any thoughts on converting logical pixels to point size? 

I'm not sure if GitHub pings you anyway, but I'm going to add a @goderbauer for good measure.

packages/ios_platform_images/lib/ios_platform_images.dart
cadenkriese: Nothing breaks if the colors are specified, it just ignores them. Should I still add the assertion?

packages/ios_platform_images/lib/ios_platform_images.dart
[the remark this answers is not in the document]
goderbauer: What is "point size"? Is that physical pixels? If it is that you'd use the devicepixelratio.

packages/ios_platform_images/lib/ios_platform_images.dart
goderbauer: I would add it since it is confusing if you add colors later (e.g. while refactoring) and nothing changes because it got silently ignored.

packages/ios_platform_images/lib/ios_platform_images.dart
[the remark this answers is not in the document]
cadenkriese: Let me take a step back first; the ultimate issue is that a size 100 Material icon is not the same as a size 100 SFSymbol, which is unintuitive for anyone using the API.



The tl;dr is that it looks like iOS has a point size (pixels scaled up for 2x or 3x displays similar to logical pixels) and a font point size that is possibly defined as 1/72 of an inch.

I've done a bunch of digging and found that the [`UIImageSymbolConfiguration`](https://developer.apple.com/documentation/uikit/uiimagesymbolconfiguration/3294241-configurationwithpointsize?language=objc) takes a system font point size. Apple docs aren't entirely clear on what this means, but [some posts suggest](https://stackoverflow.com/a/1059323/5565806) it takes the definition of a font point as 1/72 of an inch. There was also a [discussion](https://stackoverflow.com/questions/11776767/what-exactly-is-uifonts-point-size) about the correlation between font and pixel sizes that may be useful.

I've also discussed this in the Flutter Discord in [this thread](https://canary.discord.com/channels/608014603317936148/946974368385351681).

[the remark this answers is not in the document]
cadenkriese: Friendly ping to @goderbauer; I'm pretty sure sizing should be one of if not the last tweak before this can land.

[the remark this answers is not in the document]
goderbauer: The "this thread" link is not going anywhere for me... So, I don't know what has been discussed there...

Ideally, the size you pass in here should be the same unit you use everywhere else in your Flutter app. So if I specify 100 here, that should be the same as a SizedBox with 100 as a dimension. Everything else will be very confusing. I don't know enough about iOS, though, to say how to convert from Flutter's unit to what iOS expects...

jmagman:
> Failed to stop: Error getting access token for service account: 400 Bad Request

Can you rebase onto main? We had to refresh our auth tokens.

packages/ios_platform_images/example/lib/main.dart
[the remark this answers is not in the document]
jmagman: Missed this, c14cd6e isn't linking to anything now.

packages/ios_platform_images/lib/ios_platform_images.dart
```diff
@@ -138,6 +138,89 @@ class IosPlatformImages {
...
+              color.green.toDouble() / 255,
+              color.blue.toDouble() / 255,
+              color.alpha.toDouble() / 255,
+            ])
+        .toList();
+
+    final Future<Map<String, dynamic>?> loadInfo =
+        _channel.invokeMapMethod<String, dynamic>(
+      'loadSystemImage',
+      <dynamic>[
```
jmagman, line 198: Use `Object?` instead of `dynamic` in null-safe files.

packages/ios_platform_images/example/lib/main.dart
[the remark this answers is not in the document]
cadenkriese: After rebasing, it's in 831cb53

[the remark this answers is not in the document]
jmagman: I have no idea how to test Flutter widgets, maybe @goderbauer has some more suggestions.  Ideally you could tell that it errored, but at least we know it didn't crash.  But that LGTM though.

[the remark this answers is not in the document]
goderbauer: > the test passes but prints an error message after the fact.

What is printing the error message? If the error is expected, you should catch it in the text and expect it (e.g. with https://master-api.flutter.dev/flutter/flutter_test/WidgetTester/takeException.html). Running the test should not print anything to the console.

This may also be a case where a screenshot test may be useful to verify that the expected icon does get rendered (not sure if screenshot/golden tests are available in this repository).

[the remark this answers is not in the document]
cadenkriese: @goderbauer 

That was my initial approach, but I couldn't get it to work due to what I assume are some asynchrony issues on the iOS end.

I've tried reimplementing this approach locally with the following code:
```dart
  testWidgets(
    'ios system image error case',
    (WidgetTester tester) async {
      app.runErrorCaseApp();
      await tester.pumpAndSettle();

      expect(tester.takeException(), isArgumentError);
      expect(find.bySemanticsLabel('Nonexistant Symbol'), findsNothing);
    },
  );
```
`app.runErrorCaseApp()` calls `runApp` on the IOSExampleImageError class. https://github.com/flutter/plugins/blob/b23953a63b4294460e9e8fb0f40426e58019ac14/packages/ios_platform_images/example/lib/main.dart#L78-L90
However, despite throwing the argument error, the `takeException` method returns null. 
<details><summary>Test logs</summary>

```
❯ flutter test integration_test/ios_platform_images.dart
00:01 +0: ... /Users/caden/Developer/flutter/plugins/packages/ios_platform_images/example/integration_test/ios_platform_images.dart            Warning: Missing build name (CFBundleShortVersionString).
Warning: Missing build number (CFBundleVersion).
Action Required: You must set a build name and number in the pubspec.yaml file version field before submitting to the App Store.
00:03 +0: ... /Users/caden/Developer/flutter/plugins/packages/ios_platform_images/example/integration_test/ios_platform_images.dart            Ru00:04 +0: ... /Users/caden/Developer/flutter/plugins/packages/ios_platform_images/example/integration_test/ios_platform_images.dart        494ms
00:16 +0: ... /Users/caden/Developer/flutter/plugins/packages/ios_platform_images/example/integration_test/ios_platform_images.dart
00:19 +0: ... /Users/caden/Developer/flutter/plugins/packages/ios_platform_images/example/integration_test/ios_platform_images.dart      2,845ms
Xcode build done.                                           15.1s
00:22 +0: load ios bundled image
file:///Users/caden/Library/Developer/CoreSimulator/Devices/7674071D-43B8-44F4-AF74-D37EFE3999FB/data/Containers/Bundle/Application/E54E0078-C213-4CCD-A250-25CB8D054FDB/Runner.app/textfile
00:22 +1: load ios bundled image
file:///Users/caden/Library/Developer/CoreSimulator/Devices/7674071D-43B8-44F4-AF74-D37EFE3999FB/data/Containers/Bundle/Application/E54E0078-C213-4CCD-A250-25CB8D054FDB/Runner.app/textfile
00:22 +2: load ios bundled image
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following ArgumentError was thrown running a test (but after the test had completed):
Invalid argument(s): System image couldn't be found: nonexistent.sfsymbol

#2      IosPlatformImages.loadSystemImage.<anonymous closure> (package:ios_platform_images/ios_platform_images.dart:210:24)
#10     IosPlatformImages.loadSystemImage (package:ios_platform_images/ios_platform_images.dart:208:14)
#11     IOSImageExampleError.build (package:ios_platform_images_example/main.dart:88:36)
#12     StatelessElement.build (package:flutter/src/widgets/framework.dart:4827:28)
#13     ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:4754:15)
#14     Element.rebuild (package:flutter/src/widgets/framework.dart:4477:5)
#15     ComponentElement._firstBuild (package:flutter/src/widgets/framework.dart:4735:5)
#16     ComponentElement.mount (package:flutter/src/widgets/framework.dart:4729:5)
#17     Element.inflateWidget (package:flutter/src/widgets/framework.dart:3790:14)
#18     Element.updateChild (package:flutter/src/widgets/framework.dart:3524:20)
#19     SingleChildRenderObjectElement.update (package:flutter/src/widgets/framework.dart:6291:14)
#20     Element.updateChild (package:flutter/src/widgets/framework.dart:3501:15)
#21     RenderObjectToWidgetElement._rebuild (package:flutter/src/widgets/binding.dart:1198:16)
#22     RenderObjectToWidgetElement.update (package:flutter/src/widgets/binding.dart:1175:5)
#23     RenderObjectToWidgetElement.performRebuild (package:flutter/src/widgets/binding.dart:1189:7)
#24     Element.rebuild (package:flutter/src/widgets/framework.dart:4477:5)
#25     BuildOwner.buildScope (package:flutter/src/widgets/framework.dart:2659:19)
#26     WidgetsBinding.drawFrame (package:flutter/src/widgets/binding.dart:882:21)
#27     RendererBinding._handlePersistentFrameCallback (package:flutter/src/rendering/binding.dart:363:5)
#28     SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1144:15)
#29     SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1081:9)
#30     LiveTestWidgetsFlutterBinding.handleDrawFrame (package:flutter_test/src/binding.dart:1458:13)
#31     SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:995:5)
#35     _invoke (dart:ui/hooks.dart:151:10)
#36     PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:308:5)
#37     _drawFrame (dart:ui/hooks.dart:115:31)
(elided 11 frames from dart:async and package:stack_trace)
════════════════════════════════════════════════════════════════════════════════════════════════════
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following ArgumentError was thrown running a test (but after the test had completed):
Invalid argument(s): System image couldn't be found: nonexistent.sfsymbol

#2      IosPlatformImages.loadSystemImage.<anonymous closure> (package:ios_platform_images/ios_platform_images.dart:213:24)
#10     IosPlatformImages.loadSystemImage (package:ios_platform_images/ios_platform_images.dart:208:14)
#11     IOSImageExampleError.build (package:ios_platform_images_example/main.dart:88:36)
#12     StatelessElement.build (package:flutter/src/widgets/framework.dart:4827:28)
#13     ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:4754:15)
#14     Element.rebuild (package:flutter/src/widgets/framework.dart:4477:5)
#15     ComponentElement._firstBuild (package:flutter/src/widgets/framework.dart:4735:5)
#16     ComponentElement.mount (package:flutter/src/widgets/framework.dart:4729:5)
#17     Element.inflateWidget (package:flutter/src/widgets/framework.dart:3790:14)
#18     Element.updateChild (package:flutter/src/widgets/framework.dart:3524:20)
#19     SingleChildRenderObjectElement.update (package:flutter/src/widgets/framework.dart:6291:14)
#20     Element.updateChild (package:flutter/src/widgets/framework.dart:3501:15)
#21     RenderObjectToWidgetElement._rebuild (package:flutter/src/widgets/binding.dart:1198:16)
#22     RenderObjectToWidgetElement.update (package:flutter/src/widgets/binding.dart:1175:5)
#23     RenderObjectToWidgetElement.performRebuild (package:flutter/src/widgets/binding.dart:1189:7)
#24     Element.rebuild (package:flutter/src/widgets/framework.dart:4477:5)
#25     BuildOwner.buildScope (package:flutter/src/widgets/framework.dart:2659:19)
#26     WidgetsBinding.drawFrame (package:flutter/src/widgets/binding.dart:882:21)
#27     RendererBinding._handlePersistentFrameCallback (package:flutter/src/rendering/binding.dart:363:5)
#28     SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1144:15)
#29     SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1081:9)
#30     LiveTestWidgetsFlutterBinding.handleDrawFrame (package:flutter_test/src/binding.dart:1458:13)
#31     SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:995:5)
#35     _invoke (dart:ui/hooks.dart:151:10)
#36     PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:308:5)
#37     _drawFrame (dart:ui/hooks.dart:115:31)
(elided 11 frames from dart:async and package:stack_trace)
════════════════════════════════════════════════════════════════════════════════════════════════════
00:22 +2: ios system image error case
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following TestFailure was thrown running a test:
Expected: <Instance of 'ArgumentError'>
  Actual: <null>
   Which: is not an instance of 'ArgumentError'

#4      main.<anonymous closure> (file:///Users/caden/Developer/flutter/plugins/packages/ios_platform_images/example/integration_test/ios_platform_images.dart:41:7)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)

This was caught by the test expectation on the following line:
  file:///Users/caden/Developer/flutter/plugins/packages/ios_platform_images/example/integration_test/ios_platform_images.dart line 41
The test description was:
  ios system image error case
════════════════════════════════════════════════════════════════════════════════════════════════════
00:22 +2 -1: ios system image error case [E]
  Test failed. See exception logs above.
  The test description was: ios system image error case
```
</details>

packages/ios_platform_images/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.0
```
stuartmorgan, line 1: This should be 0.2.1. 0.3.0 is a breaking change the way Dart treats pre-1.0.0 releases.

packages/ios_platform_images/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+* Added `loadSystemImage` to load system images, also known as 'sfsymbols'.
```
stuartmorgan, line 3: Adds

(See style guide linked from PR checklist.)

packages/ios_platform_images/example/integration_test/ios_platform_images.dart
```diff
@@ -0,0 +1,43 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+import 'package:ios_platform_images_example/main.dart' as app;
+
+void main() {
```
stuartmorgan, line 10: You need to update `script/configs/exclude_integration_ios.yaml`; currently this file isn't actually being run by our tooling. (Thanks for adding it!)

packages/ios_platform_images/example/integration_test/ios_platform_images.dart
```diff
@@ -0,0 +1,43 @@
...
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  testWidgets(
+    'load ios bundled image',
+    (WidgetTester tester) async {
+      app.main();
+      await tester.pumpAndSettle();
+
+      expect(find.bySemanticsLabel('Flutter logo'), findsOneWidget);
```
stuartmorgan, line 19: Are these tests actually doing anything? Isn't the *widget* there regardless of whether the image loads?

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.m
```diff
@@ -40,6 +40,99 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
       NSURL *url = [[NSBundle mainBundle] URLForResource:name withExtension:extension];
       result(url.absoluteString);
       return;
+    } else if ([@"loadSystemImage" isEqualToString:call.method]) {
+      if (@available(iOS 13, *)) {
+        NSString *name = call.arguments[0];
+        NSNumber *pixelSizeWithDouble = call.arguments[1];
```
stuartmorgan, line 46: Is there a reason this isn't just:
```
double pixelSize = [(NSNumber*)(call.arguments[1]) doubleValue];
```

? It seems odd to keep the intermediate value.

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.m
```diff
@@ -40,6 +40,99 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
+        // iOS adds 15% padding to the outside of the image so we scale down to match the requested
+        // size.
+        double pointSize = [pixelSizeWithDouble doubleValue] * 0.85;
+        NSNumber *weightIndex = call.arguments[2];
+
+        // Up to 3 rgb values for primary, seconday and tertiary colors.
+        // see
+        // https://developer.apple.com/documentation/uikit/uiimagesymbolconfiguration/3810054-configurationwithpalettecolors
+        NSArray *rgbaValuesList = call.arguments[3];
```
stuartmorgan, line 55: This array should use generics.

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.m
```diff
@@ -40,6 +40,99 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
       result(url.absoluteString);
       return;
+    } else if ([@"loadSystemImage" isEqualToString:call.method]) {
+      if (@available(iOS 13, *)) {
+        NSString *name = call.arguments[0];
+        NSNumber *pixelSizeWithDouble = call.arguments[1];
+        // iOS adds 15% padding to the outside of the image so we scale down to match the requested
+        // size.
+        double pointSize = [pixelSizeWithDouble doubleValue] * 0.85;
+        NSNumber *weightIndex = call.arguments[2];
```
stuartmorgan, line 50: Same here; this could just get the int directly.

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.m
```diff
@@ -40,6 +40,99 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
+          case 0:
+            weight = UIImageSymbolWeightUltraLight;
+            break;
+          case 1:
+            weight = UIImageSymbolWeightThin;
+            break;
+          case 2:
+            weight = UIImageSymbolWeightLight;
+            break;
+          // 3 is regular
```
stuartmorgan, line 79: This should be a `case 3:` right before `default:` rather than a comment.

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.m
```diff
@@ -40,6 +40,99 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
+
+          UIImageSymbolConfiguration *final =
+              [pointSizeConfig configurationByApplyingConfiguration:colors];
+          finalImage = [UIImage systemImageNamed:name withConfiguration:final];
+        } else {
+          UIImage *image = [UIImage systemImageNamed:name withConfiguration:pointSizeConfig];
+          finalImage = [image imageWithTintColor:colorArray.count > 0 ? colorArray[0] : nil];
+        }
+
+        NSData *data = UIImagePNGRepresentation(finalImage);
```
stuartmorgan, line 125: Is this documented as allowing `nil` as an argument?

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.m
```diff
@@ -40,6 +40,99 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
+        UIImageSymbolConfiguration *pointSizeConfig =
+            [UIImageSymbolConfiguration configurationWithPointSize:pointSize
+                                                            weight:weight
+                                                             scale:UIImageSymbolScaleDefault];
+
+        UIImage *finalImage;
+
+        if (@available(iOS 15, *)) {
+          NSNumber *preferMulticolor = call.arguments[4];
+          UIImageSymbolConfiguration *colors;
```
stuartmorgan, line 109: You can initialize this inline with a conditional so it's clear that it isn't ever nil:

```
UIImageSymbolConfiguration *colors = [preferMulticolor boolValue]
    ? ...
    : ...;
```

packages/ios_platform_images/lib/ios_platform_images.dart
```diff
@@ -138,6 +138,89 @@ class IosPlatformImages {
...
+    final Future<Map<String, Object?>?> loadInfo =
+        _channel.invokeMapMethod<String, Object?>(
+      'loadSystemImage',
+      <Object?>[
+        name,
+        size,
+        weight.index,
+        colorsRGBA,
+        preferMulticolor,
+      ],
```
stuartmorgan, line 204: Doing all of this by index is pretty ugly to track across the language boundary. Perhaps this would be a good time to introduce Pigeon into this plugin.

(Failing that, we usually do this with a map so there are named keys.)

stuartmorgan-g (changes requested):

packages/ios_platform_images/example/integration_test/ios_platform_images.dart
cadenkriese: You're right; should I check if the height and width are zero instead?

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.m
cadenkriese: It makes the image black for me, but it's not officially documented on https://developer.apple.com/documentation/uikit/uiimage/3327300-withtintcolor, so I changed it to use black directly.

stuartmorgan:
What's the status of this PR? Is this ready for another round of review?

cadenkriese:
Sorry, school has been getting intense for me. I implemented the trivial parts of your review and agree that Pigeon is the way to go, so I've looked at some examples but have yet to implement it.

stuartmorgan:
Okay thanks, I'll go ahead and mark this as a draft for now so it won't show up in triage sweeps, and you can mark it as ready for review once you've updated it.

packages/ios_platform_images/lib/ios_platform_images.dart
cadenkriese: @stuartmorgan I'm trying to figure out what a good Pigeon implementation would look like here, are you suggesting I move the Obj-C logic with the switch statement for indices into a Pigeon API? What would I pass through the platform channel in that case?

stuartmorgan: I'm not sure what you mean exactly. With Pigeon you would declare a method with arguments; there wouldn't be indexes at all.

stuartmorgan: I don't know what you mean by "another method". I'm saying that index-based arguments are very error-prone and hard to maintain, and that you should either:
- use a dictionary for the arguments, where at least there's no chance of accidentally extracting completely the wrong argument on the native side, or
- use Pigeon for this new API, since Pigeon manages the internals of passing parameters for you and totally eliminates this problem.

cadenkriese: Ah! I see now, I thought I would be using Pigeon in addition to platform channels, but I think I see now that you mean I should replace this platform channel with a Pigeon API that can handle my enum. Thank you for clarifying!

packages/ios_platform_images/example/integration_test/ios_platform_images_integration_test.dart
```diff
@@ -0,0 +1,58 @@
...
+      app.main();
+      await tester.pumpAndSettle();
+
+      expect(find.bySemanticsLabel('Smiling face'), findsOneWidget);
+      expect(find.bySemanticsLabel('Sprinting hare'), findsOneWidget);
+      expect(find.bySemanticsLabel('Ladybug'), findsOneWidget);
+    },
+  );
+
+  test(
```
stuartmorgan, line 39: The reason CI didn't fail here turns out to be that you use `test` instead of `testWidgets`; see the Note box at https://docs.flutter.dev/testing/integration-tests#project-setup

cadenkriese:
Apologies for my delays getting back to this. I believe the most significant parts of this PR are correct unless I went too far out of scope with the pigeon implementation.

The one thing I haven't been able to figure out on my own, however, is why the [error case integration test](https://github.com/flutter/plugins/blob/af3565bf8a92779a5de43409b148405e462c098e/packages/ios_platform_images/example/integration_test/ios_platform_images_integration_test.dart#L39) fails.

stuartmorgan:
Sorry for the delay in looking at the error you mentioned; unfortunately the logs have expired so I can't see what the failure was.

There are also a lot of other failures here (unit tests, format, etc.) Could you push an update to fix those issue, and then we can look at the new run of the integration test failure to see what the failure message is?

stuartmorgan:
It looks like the native code implementing the Pigeon interface isn't compiling in this version.

stuartmorgan:
@cadenkriese Is this still something you are planning on finishing? Should we mark it as a draft for now?

cadenkriese:
> @cadenkriese Is this still something you are planning on finishing? Should we mark it as a draft for now?

Yes, apologies for the delay (again). I'm trying to figure out unit testing with pigeon but haven't found any resources for that? On a macro level are unit tests making their own mock api implementation or does the framework handle that somehow?

stuartmorgan:
Dart unit tests, or native unit tests? For Dart, you can just use Mockito to generate a mock implementation of the `*Api` class. Pigeon has a way of generating a test implementation that lets you mock the method channel, but that's more of a transitional tool for migrating from a pre-Pigeon implementation without fully rewriting the tests; mocking the API class is less fragile than mocking the binary messenger.

cadenkriese:
@stuartmorgan, apologies for the silence on my end. I want to try to finish this since the 1-year mark is coming up. Here's a summary of the issues I'm facing and looking for guidance on:

Most importantly, I'm trying to figure out how to proceed with the failure integration test; I've tried 3 or 4 different implementations, but nothing can handle the asynchrony properly and report back to the test framework. The current state of the relevant code is here. https://github.com/flutter/plugins/blob/6b322afbaac00d8f3e3622562d13ac542074812b/packages/ios_platform_images/example/integration_test/ios_platform_images_integration_test.dart#L39-L58

I also can't get the integration test to run on my machine (Runner crashes), and it looks like LUCI can't either https://ci.chromium.org/ui/p/flutter/builders/try/Mac_x64%20ios_platform_tests_3_of_4%20master/817/overview. So ultimately, I'm very lost when it comes to all of this, I would appreciate some help getting back on track.

Also, the analyzer [isn't liking](https://ci.chromium.org/ui/p/flutter/builders/try/Mac_x64%20ios_platform_tests_2_of_4%20master/820/overview) that Pigeon generates an unused function, what should I do about this?
```
/Volumes/Work/s/w/ir/x/w/plugins/packages/ios_platform_images/ios/Classes/PlatformImagesApi.g.m:22:11: error: unused function 'GetNullableObject' [-Werror,-Wunused-function]
static id GetNullableObject(NSDictionary *dict, id key) {
          ^
1 error generated.
```

Thank you again for your persistence with this!

stuartmorgan:
> I also can't get the integration test to run on my machine (Runner crashes)

In general, or when running the specific integration test you referenced above?

> Also, the analyzer [isn't liking](https://ci.chromium.org/ui/p/flutter/builders/try/Mac_x64%20ios_platform_tests_2_of_4%20master/820/overview) that Pigeon generates an unused function, what should I do about this?

That's [a Pigeon bug](https://github.com/flutter/flutter/issues/119342); we can fix it and then you can update this PR with the new version of Pigeon.

stuartmorgan:
I don't know of a good way to do native debugging of a Dart integration test, but the workaround that I use is to locally copy the code from the problematic test into the example app itself, then run the example app in debug mode from Xcode, so you're attached with a native debugger. Then you can see what's crashing.

cadenkriese:
> I don't know of a good way to do native debugging of a Dart integration test, but the workaround that I use is to locally copy the code from the problematic test into the example app itself, then run the example app in debug mode from Xcode, so you're attached with a native debugger. Then you can see what's crashing.

That worked, thank you! Now I'm back where I was originally, the `ios system image error case` test is failing due to an ArgumentError, but I'm calling `await expectLater(completer.future, throwsArgumentError);`

packages/ios_platform_images/pigeons/platform_images.dart
```diff
@@ -0,0 +1,39 @@
...
+  objcHeaderOut: 'ios/Classes/PlatformImagesApi.g.h',
+  objcSourceOut: 'ios/Classes/PlatformImagesApi.g.m',
+  objcOptions: ObjcOptions(
+    prefix: 'FLT',
+  ),
+  copyrightHeader: 'pigeons/copyright.txt',
+))
+class PlatformImage {
+  double? scale;
+  Uint8List? bytes;
```
stuartmorgan, line 18: Why are these nullable? What does it mean for an image to have a null scale or null bytes?

cadenkriese: Yeah that doesn't make sense, I think I just did it without thinking to get rid of compiler errors.

Should I give them default values like the code below or add a `late` modifier?
```dart
double scale = 1;
Uint8List bytes = Uint8List(0);
```

stuartmorgan: You can define a Dart constructor for them with required parameters.

stuartmorgan:
> Now I'm back where I was originally, the `ios system image error case` test is failing due to an ArgumentError, but I'm calling `await expectLater(completer.future, throwsArgumentError);`

I didn't have time to fully debug this, but based on running it locally with breakpoints the errors don't seem to be propagating the way your test expects; `completer.completeError(exception);` is never even reached. I don't think your expectation is the problem; perhaps the error is not being passed through the future chain at some point.

cadenkriese:
@stuartmorgan I finally made progress! The `ImageStreamListener`'s `onError` method was getting called, but I had an (amateur) [syntax error](https://github.com/flutter/plugins/pull/4803/commits/38fe2c1c755a928d458ee429730209b452ece49b#diff-79c26e5693d1dadb3c1e39902fa89e27a3a0e2deb7a21b3133fe977b7c713ae7R47-R53) that slipped by. I'm still not sure how this is valid Dart code but oh well.

I also deleted the `completeError` call for the `scaleCompleter` since the `bytesCompleter` errored first and thus was caught, while the `scaleCompleter` errors later down the line. I'm unsure what the implications of this completer never completing or erroring are. I would think it best to error it as well and catch the error in the test but I would have no idea where to begin with that because I don't know what state the `scaleCompleter` is in after the `bytesCompleter` errors.

Finally, I [took a stab](https://github.com/flutter/plugins/pull/4803/commits/9293e2600b26893416de4bd10f31de1756caf0c2) at rearchitecting the other integration tests, but it's pretty bad. I'm all ears if you have any ideas on what would be best here.

packages/ios_platform_images/pubspec.yaml
```diff
@@ -12,12 +12,19 @@ flutter:
...
     sdk: flutter
 
 dev_dependencies:
+  build_runner: ^2.2.1
   flutter_test:
     sdk: flutter
+  integration_test:
+    sdk: flutter
+  mockito: ^5.3.2
+  pedantic: ^1.10.0
```
jmagman, line 29: `pedantic` has been deprecated, check out https://github.com/flutter/plugins/pull/6179
```suggestion
```

jmagman (commented):
> Finally, I [took a stab](https://github.com/flutter/plugins/pull/4803/commits/9293e2600b26893416de4bd10f31de1756caf0c2) at rearchitecting the other integration tests according to your previous feedback, but it's pretty bad. The previous state wasn't really testing anything at all, and now it's much too quirky and hardcoded.

I'm okay with hard-coded values in tests like this, as long as they pass in all expected environments.  The symbols and therefore dimensions may change with future iOS versions, we can investigate that when it happens though.

cadenkriese:
> I'm okay with hard-coded values in tests like this, as long as they pass in all expected environments. The symbols and therefore dimensions may change with future iOS versions, we can investigate that when it happens though.

They worked on two different iPhones in simulator but it's not passing in CI so I'm not sure if it's reliable.

jmagman:
> > I'm okay with hard-coded values in tests like this, as long as they pass in all expected environments. The symbols and therefore dimensions may change with future iOS versions, we can investigate that when it happens though.
> 
> They worked on two different iPhones in simulator but it's not passing in CI so I'm not sure if it's reliable.

```
flutter: ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
flutter: The following TestFailure was thrown running a test:
flutter: Expected: <96>
flutter:   Actual: <97>
```
Yeah... You could check it's greater than 0, or you could use a Matcher like https://api.flutter.dev/flutter/package-matcher_matcher/closeTo.html

stuartmorgan-g:

Please see [our instructions](https://github.com/flutter/flutter/wiki/Migrating-Plugins-repository-PRs-to-Packages) for an explanation of how to move your PR, and if you have any issues moving your PR please don't hesitate to reach out in the #hackers-ecosystem channel in [Discord](https://github.com/flutter/flutter/wiki/Chat).

Our apologies that your PR was caught in this one-time transition. We're aware that it's disruptive in the short term, and appreciate your help in getting us to a better long-term state!

PR #4806, opened 2022-02-11
[local_auth] Windows support.

azchohfi:
Adds Windows support for the local_auth plugin, using the Windows Hello APIs.

https://github.com/flutter/flutter/issues/70270

## Pre-launch Checklist

- [X] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [X] I signed the [CLA].
- [X] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [X] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].


azchohfi:
No idea why these are failing.

stuartmorgan:
> Any help here? I have not touched the ios platform.

It was likely flake (I'm not seeing the failure in the commit history, so I can't check.)

packages/local_auth/local_auth/windows/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+cmake_minimum_required(VERSION 3.15)
+set(PROJECT_NAME "local_auth")
+set(CPPWINRT_VERSION "2.0.210806.1")
+project(${PROJECT_NAME} LANGUAGES CXX)
+include(FetchContent)
+
+set(PLUGIN_NAME "${PROJECT_NAME}_plugin")
+
+FetchContent_Declare(nuget
+  URL "https://dist.nuget.org/win-x86-commandline/v6.0.0/nuget.exe"
```
stuartmorgan, line 10: Should we consider instead making NuGet a `doctor`-recommended package, and referencing the VS copy instead? /cc @cbracken for thoughts on this.

packages/local_auth/local_auth/windows/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+
+find_program(NUGET nuget)
+if (NOT NUGET)
+    message("Nuget.exe not found, trying to download or use cached version.")
+    FetchContent_MakeAvailable(nuget)
+    set(NUGET ${nuget_SOURCE_DIR}/nuget.exe)
+endif()
+
+execute_process(COMMAND
+    ${NUGET} install Microsoft.Windows.CppWinRT -Version ${CPPWINRT_VERSION} -OutputDirectory packages
```
stuartmorgan, line 23: Will this install be completely localized to the plugin build, or will it have machine-level effects?

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include <UserConsentVerifierInterop.h>
+#include <flutter/encodable_value.h>
+#include <flutter/method_channel.h>
+#include <flutter/plugin_registrar_windows.h>
+#include <flutter/standard_method_codec.h>
+#include <pplawait.h>
+#include <ppltasks.h>
+#include <windows.h>
+#include <winrt/Windows.Foundation.h>
+#include <winrt/Windows.Security.Credentials.UI.h>
```
stuartmorgan, line 16: We should clearly document the Windows version requirements in the README of the new package.

Since I'm not familiar with how WinRT linking works: what are the implications of the dependency here on support for older versions of Windows? Will a project that includes `local_auth` completely fail to launch on, e.g., Windows 7, or will it only be a problem if the plugin code is called?

(Pre-Windows-10 support definitely isn't required, I just want to make sure we know what will happen.)

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include <memory>
+#include <sstream>
+
+namespace {
+
+using namespace flutter;
+using namespace winrt;
+
+template <typename T>
+T GetArgument(const std::string arg, const EncodableValue* args, T fallback) {
```
stuartmorgan, line 29: Please make sure to document all functions and methods, per the C++ style guide.

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,187 @@
...
+    std::unique_ptr<MethodResult<EncodableValue>> result) {
+  try {
+    flutter::EncodableList biometrics;
+    auto ucvAvailability = co_await Windows::Security::Credentials::UI::
+        UserConsentVerifier::CheckAvailabilityAsync();
+    if (ucvAvailability == Windows::Security::Credentials::UI::
+                               UserConsentVerifierAvailability::Available) {
+      biometrics.push_back(EncodableValue("fingerprint"));
+      biometrics.push_back(EncodableValue("face"));
+      biometrics.push_back(EncodableValue("iris"));
```
stuartmorgan, line 163: This list is clearly not accurate. What actual information do we get about supported authentication options? We're in the process of changing the API (as part of the federation PRs) to also list `strong` and `weak` to match the Android options. Does one or both of those work instead? Is there a different set of things we should consider adding?

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,187 @@
...
+    result->NotImplemented();
+  }
+}
+
+winrt::fire_and_forget LocalAuthPlugin::Authenticate(
+    const MethodCall<EncodableValue>& method_call,
+    std::unique_ptr<MethodResult<EncodableValue>> result) {
+  auto reasonW = s2ws(GetArgument<std::string>(
+      "localizedReason", method_call.arguments(), std::string()));
+
```
stuartmorgan, line 118: You are ignoring `biometricOnly`, which would be a serious security bug for some clients. We either need to support it, or if it's not possible with the available Windows APIs, error out when it's set to true.

packages/local_auth/local_auth/windows/CMakeLists.txt
azchohfi: VS doesn't include its own version of NuGet.exe, so it needs to be downloaded somewhere. It would be beneficial to have NuGet.exe inside Flutter's installation, yes, but not sure where. Also, I do not know the work required in Flutter to add an exe as a dependency. IMO, this should not be an error, but only a warning. Also, if we go this route, it would be great if doctor checked if nuget.exe is in the path already, and use that as the correct path, allowing developers to override the nuget exe's version.

packages/local_auth/local_auth/windows/CMakeLists.txt
azchohfi: The only machine-level side effect of this is the download of CppWinRT's nuget file to Nuget's cache (_~\.nuget\packages_), and the unzipping of it to that same folder, but it should be that, and only that: a cache. Each Flutter project that is built will have its own plugin folder, which will have its own copy of that cached folder locally (that is what the -OutputDirectory parameter is doing there).

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
azchohfi: The existing Windows API being used here does not let us ask for a specific auth method, so it is up to the user to choose one that is available on their machine. That can be a combination of Fingerprint, Face, and or PIN.
I will ask internally here at Microsoft if there is guidance on querying those methods, but I'm assuming it would require some complex device GUID querying work, with likely false positives/negatives, but again, just an assumption.

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,187 @@
...
+  virtual ~LocalAuthPlugin();
+
+ private:
+  flutter::PluginRegistrarWindows* registrar_;
+
+  // Called when a method is called on this plugin's channel from Dart.
+  void HandleMethodCall(const MethodCall<EncodableValue>& method_call,
+                        std::unique_ptr<MethodResult<EncodableValue>> result);
+
+  winrt::fire_and_forget Authenticate(
```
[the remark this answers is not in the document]
azchohfi: Inside this method, I'm only calling a Windows Runtime IAsyncXxx method, so C++/WinRT guarantees that it will capture the calling context and ensure I'm still on it when the continuation resumes, so I'm only calling result-> anything on the same thread where each method was called from.

packages/local_auth/local_auth/windows/CMakeLists.txt
stuartmorgan: > VS doesn't include its own version of NuGet.exe, so it needs to be downloaded somewhere.

Ah, I just assumed it was since there's UI for installing packages with it in the IDE. If it's not bundled, this seems like the best approach; we generally don't ship toolchains.

We might want to add a doctor suggestion for works to install it themselves and put it in their path though (so that this wouldn't need to run most of the time).

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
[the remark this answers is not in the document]
stuartmorgan: Thanks for the explanation; I haven't really had any experience with WinRT so misunderstood the behavior.

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
stuartmorgan: If there's any reasonable correspondence with either the "weak" or "strong" designations that Android uses now we could just return that.

If we can't get *any* information would be to add something like "unknown" to the set of possible options in the API. Then clients can decide how they want to handle it.

packages/local_auth/local_auth/windows/CMakeLists.txt
azchohfi: Visual Studio uses "MSBuild.exe /restore", not nuget.exe. AFAIK, VS's UI edits the csproj (in case of C# projects) and calls the restore, basically.
CMake 3.15 does support adding nuget package references, which are restored by msbuild, but there are issues with vcxprojs (C++) and PackageReferences (that is why vcxprojs use a packages.config file), so this doesn't actually work (unfortunately). https://stackoverflow.com/a/61910170/1272111

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
azchohfi: I've started an internal thread on how this should be handled, and I'll get back to you with the recommendations, but I believe this is either gonna be a new value that needs to be created only for windows or, for better compatibility, 'weak'.

packages/local_auth/local_auth/windows/local_auth_plugin.cpp
azchohfi: I'm failing when this is true on Windows now.

stuartmorgan:
> * Our policy for adding new platforms to a plugin in this repository is that we federate it first. `local_auth` is being federated right now, so this will need to be reworked as a new package once that lands.

FYI, this has been taking longer than expected to get finishing and landed, but is still happening.

stuartmorgan:
The person working on the local_auth federation has picked up the PRs again recently, and they are looking very close, so hopefully this will be unblocked next week.

stuartmorgan:
We've now completed and published the fully federated version of `local_auth`, so converting this to a `local_auth_windows` is unblocked. Sorry that took longer than expected!

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  auto userConsentVerifierInterop = winrt::get_activation_factory<
+      Windows::Security::Credentials::UI::UserConsentVerifier,
+      IUserConsentVerifierInterop>();
+
+  auto hWnd = GetRootWindow(registrar_->GetView());
+
+  HSTRING hReason;
+  WindowsCreateString(reasonW.c_str(), (uint32_t)reasonW.size(), &hReason);
```
cbracken, line 159: Check the return value for failure.

packages/local_auth/local_auth_windows/example/lib/main.dart
```diff
@@ -0,0 +1,238 @@
...
+  bool _isAuthenticating = false;
+
+  @override
+  void initState() {
+    super.initState();
+    LocalAuthPlatform.instance.isDeviceSupported().then(
+          (bool isSupported) => setState(() => _supportState = isSupported
+              ? _SupportState.supported
+              : _SupportState.unsupported),
+        );
```
cbracken, line 37: nit: consider using a block here rather than arrow notation to reduce the wrapping and make it very marginally more readable:

```dart
LocalAuthPlatform.instance.isDeviceSupported().then((bool isSupported) {
  setState(() {
    _supportState = isSupported
      ? _SupportState.supported
      : _SupportState.unsupported),
  });
});
```

packages/local_auth/local_auth_windows/example/lib/main.dart
```diff
@@ -0,0 +1,238 @@
...
+        _authorized = 'Error - ${e.message}';
+      });
+      return;
+    }
+    if (!mounted) {
+      return;
+    }
+
+    setState(
+        () => _authorized = authenticated ? 'Authorized' : 'Not Authorized');
```
cbracken, line 107: nit: Consider using a block here rather than arrow notation.

packages/local_auth/local_auth_windows/example/lib/main.dart
```diff
@@ -0,0 +1,238 @@
...
+          children: <Widget>[
+            Column(
+              mainAxisAlignment: MainAxisAlignment.center,
+              children: <Widget>[
+                if (_supportState == _SupportState.unknown)
+                  const CircularProgressIndicator()
+                else if (_supportState == _SupportState.supported)
+                  const Text('This device is supported')
+                else
+                  const Text('This device is not supported'),
```
cbracken, line 172: Use braces for conditionals with an `else` clause:
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#separate-the-if-expression-from-its-statement

They're optional for one-line `if` statements with no else clause.

cbracken (approved):
Overall lgtm modulo comments. Thanks for your patience on the review.

cbracken (changes requested):
Apologies; selected the wrong radio button -- overall looks good so far modulo my comments. Let us know if you need suggestions on the Windows native test side.

packages/local_auth/local_auth_windows/lib/types/auth_messages_windows.dart
```diff
@@ -0,0 +1,191 @@
...
+      biometricHint.hashCode ^
+      biometricNotRecognized.hashCode ^
+      biometricRequiredTitle.hashCode ^
+      biometricSuccess.hashCode ^
+      cancelButton.hashCode ^
+      deviceCredentialsRequiredTitle.hashCode ^
+      deviceCredentialsSetupDescription.hashCode ^
+      goToSettingsButton.hashCode ^
+      goToSettingsDescription.hashCode ^
+      signInTitle.hashCode;
```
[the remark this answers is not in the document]
azchohfi: This file is mostly a copy of the android version of this file, so if this is an issue, then it already exists in the main branch. I'm happy to change this, but just wanted to let you know.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+// Converts the given UTF-8 string to UTF-16.
+std::wstring s2ws(const std::string& s) {
+  int len;
+  int slength = (int)s.length() + 1;
+  len = MultiByteToWideChar(CP_ACP, 0, s.c_str(), slength, 0, 0);
+  std::wstring r(len, L'\0');
+  MultiByteToWideChar(CP_ACP, 0, s.c_str(), slength, &r[0], len);
+  return r;
+}
```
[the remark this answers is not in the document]
azchohfi: Yeah, I've replaced that implementation with the one from the URL launcher, which is exactly the same as this.

packages/local_auth/local_auth_windows/lib/types/auth_messages_windows.dart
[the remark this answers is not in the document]
cbracken: `Object.hash` is a fairly recent addition to the API of Object, so makes sense that we didn't do it before, but we should now.

packages/local_auth/local_auth_windows/example/lib/main.dart
azchohfi: I thought If statements inside widgets didn't have curly brackets, by design. Is there a way to do this? Also, this file is a copy of the other platform's example, so this issue is already in main. Should this be a "cleaning" PR on top of master, and not part of this PR?

cbracken: Apologies -- you're entirely correct; I'd somehow glossed over the fact this was conditional inclusion in a collection. This is correct as-is.

Up to you whether you'd like to send cleanups in a separate patch or include them in this one.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+  return result;
+}
+
+// Returns the window's HWND for a given FlutterView
+HWND GetRootWindow(flutter::FlutterView* view) {
+  return ::GetAncestor(view->GetNativeWindow(), GA_ROOT);
+}
+
+// Converts the given UTF-8 string to UTF-16.
+std::wstring s2ws(const std::string& s) {
```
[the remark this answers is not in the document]
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+  int len;
+  int slength = (int)s.length() + 1;
```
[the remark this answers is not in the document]
azchohfi: Fixed.

packages/local_auth/local_auth_windows/lib/types/auth_messages_windows.dart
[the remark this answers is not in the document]
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+  len = MultiByteToWideChar(CP_ACP, 0, s.c_str(), slength, 0, 0);
+  std::wstring r(len, L'\0');
```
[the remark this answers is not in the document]
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+// Returns the window's HWND for a given FlutterView
+HWND GetRootWindow(flutter::FlutterView* view) {
+  return ::GetAncestor(view->GetNativeWindow(), GA_ROOT);
+}
+
+// Converts the given UTF-8 string to UTF-16.
+std::wstring s2ws(const std::string& s) {
+  int len;
+  int slength = (int)s.length() + 1;
+  len = MultiByteToWideChar(CP_ACP, 0, s.c_str(), slength, 0, 0);
```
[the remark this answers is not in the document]
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+  std::wstring r(len, L'\0');
+  MultiByteToWideChar(CP_ACP, 0, s.c_str(), slength, &r[0], len);
```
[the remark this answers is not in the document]
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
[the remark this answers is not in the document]
azchohfi: Fixed.

azchohfi: Fixed.

packages/local_auth/local_auth_windows/example/lib/main.dart
azchohfi: Fixed.

azchohfi:
@cbracken Thanks for the review. I've mimicked what the camera plugin does, so I got native mocked tests for the plugin running now. Let me know what you think. I think that is a good start.
Also, let me know what we should do to fix the publishable task for this PR.

stuartmorgan:
`publishable` is failing because you are changing `local_auth`, but not changing the `local_auth` version, which is against our policy.

But this also needs to be two different PRs: one to add `local_auth_windows`, and a follow-up to endorse it in `local_auth`.

azchohfi:
Ok, I just reverted those changes. We can add them back in a future PR.

azchohfi:
@stuartmorgan / @cbracken Anything else needed here?

packages/local_auth/local_auth_windows/AUTHORS
```diff
@@ -0,0 +1,67 @@
...
+Théo Champion <contact.theochampion@gmail.com>
+Kazuki Yamaguchi <y.kazuki0614n@gmail.com>
+Eitan Schwartz <eshvartz@gmail.com>
+Chris Rutkowski <chrisrutkowski89@gmail.com>
+Juan Alvarez <juan.alvarez@resideo.com>
+Aleksandr Yurkovskiy <sanekyy@gmail.com>
+Anton Borries <mail@antonborri.es>
+Alex Li <google@alexv525.com>
+Rahul Raj <64.rahulraj@gmail.com>
+Bodhi Mulders <info@bemacized.net>
```
stuartmorgan, line 67: A new AUTHORS file should just be the authors of the specific new package, which would be the standard Google.com entry plus you.

(The giant list we have in most places is because for a long time we only had one AUTHORS file in the whole repo.)

packages/local_auth/local_auth_windows/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 1.0.0
+
+* Initial release from migration to federated architecture.
```
stuartmorgan, line 3: Initial release of Windows support.

packages/local_auth/local_auth_windows/lib/local_auth_windows.dart
```diff
@@ -0,0 +1,89 @@
...
+      if (messages is WindowsAuthMessages) {
+        args.addAll(messages.args);
+      }
+    }
+    return (await _channel.invokeMethod<bool>('authenticate', args)) ?? false;
+  }
+
+  @override
+  Future<bool> deviceSupportsBiometrics() async {
+    return (await getEnrolledBiometrics()).isNotEmpty;
```
stuartmorgan, line 52: This is broken; we fixed it recently in iOS and Android (where it was a regression from federation); it's supposed to call the method channel invocation directly so that it can pick up 'undefined'.

But the whole 'undefined' thing was awful in the first place, and we're moving away from it. Instead, you should define a new method channel call for `deviceSupportsBiometrics`, and call that here, and get rid of 'undefined' anywhere it's referenced.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,223 @@
...
+    std::unique_ptr<MethodResult<EncodableValue>> result) {
+  try {
+    flutter::EncodableList biometrics;
+    auto ucvAvailability =
+        co_await user_consent_verifier_->CheckAvailabilityAsync();
+    if (ucvAvailability == Windows::Security::Credentials::UI::
+                               UserConsentVerifierAvailability::Available) {
+      biometrics.push_back(EncodableValue("fingerprint"));
+      biometrics.push_back(EncodableValue("face"));
+      biometrics.push_back(EncodableValue("iris"));
```
stuartmorgan, line 205: Looks like https://github.com/flutter/plugins/pull/4806#discussion_r814057427 is still not resolved.

packages/local_auth/local_auth_windows/lib/types/auth_messages_windows.dart
```diff
@@ -0,0 +1,191 @@
...
+    this.biometricHint,
+    this.biometricNotRecognized,
+    this.biometricRequiredTitle,
+    this.biometricSuccess,
+    this.cancelButton,
+    this.deviceCredentialsRequiredTitle,
+    this.deviceCredentialsSetupDescription,
+    this.goToSettingsButton,
+    this.goToSettingsDescription,
+    this.signInTitle,
```
stuartmorgan, line 25: Are these all actually used? Are any of them?

packages/local_auth/local_auth_windows/windows/local_auth.h
```diff
@@ -0,0 +1,72 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <UserConsentVerifierInterop.h>
+
+#include "include/local_auth_windows/local_auth_plugin.h"
+
+// This must be included before many other Windows headers.
```
stuartmorgan, line 9: This comment is cruft; it doesn't apply to the line it's on.

packages/local_auth/local_auth_windows/windows/local_auth.h
```diff
@@ -0,0 +1,72 @@
...
+#include <map>
+#include <memory>
+#include <sstream>
+
+namespace local_auth_windows {
+
+using namespace flutter;
+using namespace winrt;
+
+class UserConsentVerifier {
```
stuartmorgan, line 30: This class and all of its methods need declaration comments, per style guide.

packages/local_auth/local_auth_windows/windows/local_auth.h
```diff
@@ -0,0 +1,72 @@
...
+#include <winstring.h>
+
+#include <map>
+#include <memory>
+#include <sstream>
+
+namespace local_auth_windows {
+
+using namespace flutter;
+using namespace winrt;
```
stuartmorgan, line 28: These violate the style guide: https://google.github.io/styleguide/cppguide.html#Namespaces

packages/local_auth/local_auth_windows/windows/local_auth.h
```diff
@@ -0,0 +1,72 @@
...
+  UserConsentVerifier(const UserConsentVerifier&) = delete;
+  UserConsentVerifier& operator=(const UserConsentVerifier&) = delete;
+};
+
+class LocalAuthPlugin : public Plugin {
+ public:
+  static void RegisterWithRegistrar(PluginRegistrarWindows* registrar);
+
+  LocalAuthPlugin(std::function<HWND()> window_provider);
+  LocalAuthPlugin(std::unique_ptr<UserConsentVerifier> user_consent_verifier);
```
stuartmorgan, line 52: These could use comments explaining the different use cases.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include "local_auth.h"
+
+namespace local_auth_windows {
+
+using namespace flutter;
+using namespace winrt;
+
+template <typename T>
+// Helper method for getting an argument from an EncodableValue
+T GetArgument(const std::string arg, const EncodableValue* args, T fallback) {
```
stuartmorgan, line 13: All of these private helpers and classes should be in an anonymous namespace to give them internal linkage.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,223 @@
...
+class UserConsentVerifierImpl : public UserConsentVerifier {
+ public:
+  explicit UserConsentVerifierImpl(std::function<HWND()> window_provider)
+      : get_root_window_(std::move(window_provider)){};
+  virtual ~UserConsentVerifierImpl() = default;
+
+  Windows::Foundation::IAsyncOperation<
+      Windows::Security::Credentials::UI::UserConsentVerificationResult>
+  RequestVerificationForWindowAsync(std::wstring localizedReason) override {
+    auto userConsentVerifierInterop = winrt::get_activation_factory<
```
stuartmorgan, line 62: There's a lot of use of auto here where it's not actually obvious what the type is; see the "decision" section of https://google.github.io/styleguide/cppguide.html#Type_deduction

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,223 @@
...
+        Windows::Security::Credentials::UI::UserConsentVerifier,
+        IUserConsentVerifierInterop>();
+
+    auto hWnd = get_root_window_();
```
stuartmorgan, line 66: These variable names (hWnd, hReason) violate style guide naming rules.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    auto consentResult =
+        co_await winrt::capture<Windows::Foundation::IAsyncOperation<
+            Windows::Security::Credentials::UI::UserConsentVerificationResult>>(
+            userConsentVerifierInterop,
+            &::IUserConsentVerifierInterop::RequestVerificationForWindowAsync,
+            hWnd, hReason);
+
+    WindowsDeleteString(hReason);
```
stuartmorgan, line 83: Is there a unique_ptr adapter for this out of the box that we could use?

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,223 @@
...
+  } else {
+    result->NotImplemented();
+  }
+}
+
+// Starts authentication process
+winrt::fire_and_forget LocalAuthPlugin::Authenticate(
+    const MethodCall<EncodableValue>& method_call,
+    std::unique_ptr<MethodResult<EncodableValue>> result) {
+  auto reasonW = Utf16FromUtf8(GetArgument<std::string>(
```
stuartmorgan, line 150: Names here also need to use C++ style guide naming (`reason`, `biometric_only`, etc.)

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,223 @@
...
+        co_await user_consent_verifier_->CheckAvailabilityAsync();
+    if (ucvAvailability == Windows::Security::Credentials::UI::
+                               UserConsentVerifierAvailability::Available) {
+      biometrics.push_back(EncodableValue("fingerprint"));
+      biometrics.push_back(EncodableValue("face"));
+      biometrics.push_back(EncodableValue("iris"));
+    }
+    result->Success(biometrics);
+  } catch (const std::exception& e) {
+    result->Error("no_biometrics_available", e.what());
```
stuartmorgan, line 209: Does this case not map to one of the established error codes?

stuartmorgan (changes requested):

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
azchohfi: I'm using the same as the Android one:
https://github.com/flutter/plugins/blob/bde713a230d3931f17105f2b4f8439cf448d031a/packages/local_auth/local_auth_android/android/src/main/java/io/flutter/plugins/localauth/LocalAuthPlugin.java#L269

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
azchohfi: Yes, there is, I'll switch to it.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth.h
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
azchohfi: Done.

packages/local_auth/local_auth_windows/lib/local_auth_windows.dart
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
azchohfi: I'm now returning both weak and strong, but they act the same, just to increase compatibility, as Windows Hello doesn't support picking the auth security "level". Do you believe we should actually add the "unknown" option?

stuartmorgan: What is the spectrum of possible types of authentication? We certainly shouldn't return a level that suggests stronger auth is available than actually is.

Consider, say, a banking application that wants to allow using local_auth-based authentication only if a sufficiently strong biometric option is available.

azchohfi: From the documentation: "Microsoft Passport PIN, Windows Hello biometric, or fingerprint reader". This APIs is the recommended one for secure authentication on Windows 10+.
https://docs.microsoft.com/en-us/windows/uwp/security/microsoft-passport#2-what-is-windows-hello

It is important to notice that not all Windows 10 devices support Windows Hello, as it will only work if a TPM module is present, and the biometric credentials are properly configured. You can't only use a PIN. Here is a list of devices that support it:
https://www.windowscentral.com/complete-list-laptops-support-windows-hello

The list is long, and growing, but developers should be aware that this is not guaranteed, and should provide a different auth method, just as they should with the existing API exposed by this plugin on other platforms.

This **is** the API we recommend banking apps to use if they want a strong auth validation, as it is the same one that will be used if available during the Windows Sign-in process.

stuartmorgan: Thanks, that's useful context.

Since we documented `strong` as:

```
  /// Any biometric (e.g. fingerprint, iris, or face) on the device that the
  /// platform API considers to be strong.
```

I'm fine with `weak` and `strong` if you think that description applies for Windows (which it sounds like is the case).

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
stuartmorgan: Ah. That's probably a bug, but since we clearly need to do a sweep to enforce more consistency already, matching Android is fine.

packages/local_auth/local_auth_windows/example/README.md
```diff
@@ -0,0 +1,8 @@
+# local_auth_example
+
+Demonstrates how to use the local_auth plugin.
+
+## Getting Started
```
stuartmorgan, line 5: Sorry, missed this before. This whole section is boilerplate and should be removed. (I need to scrub the whole repo for it and clean it all up.)

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
azchohfi: I would say so, but it is important to notice that a PIN can still be used and will always be available (as long as Hello is enabled). This is mostly in case the user had an injury and can't use the biometric device, for example, but it is the user's choice, not the developers.

stuartmorgan: That should be fine, since Android and iOS allow that as well with `biometricOnly` set to false.

packages/local_auth/local_auth_windows/lib/types/auth_messages_windows.dart
stuartmorgan: Answering my own question: none of them are, unless I missed something. So this class can be empty for now.

Or not exist, but that's probably more confusing since people would probably look for it and not be able to find it. I would leave it here, but empty (with `args` returning an empty dict) and a comment saying that it's currently unused.

packages/local_auth/local_auth_windows/pubspec.yaml
```diff
@@ -0,0 +1,27 @@
...
+    implements: local_auth
+    platforms:
+      windows:
+        pluginClass: LocalAuthPlugin
+        dartPluginClass: LocalAuthWindows
+
+dependencies:
+  flutter:
+    sdk: flutter
+  intl: ^0.17.0
```
stuartmorgan, line 22: Which means you can drop this dependency.

packages/local_auth/local_auth_windows/test/local_auth_test.dart
```diff
@@ -0,0 +1,171 @@
...
+      );
+    });
+
+    test('stopAuthentication returns false', () async {
+      final bool result = await localAuthentication.stopAuthentication();
+      expect(result, false);
+    });
+
+    group('With device auth fail over', () {
+      test('authenticate with no args.', () async {
```
stuartmorgan, line 78: Apparently the existing tests in the implementation packages have completely incorrect names 😢  let's not duplicate the mistake here. (I suspect they made more sense at some point on an earlier version of the API, then drifted as things changed.)

Let's eliminate the two groups, and just have one 'authenticate' group, with two tests of for now: 'authenticate with no arguments passes expected defaults' (which actually passes no arguments, unlike now) and 'authenticate passes all options' (which has all non-default values for options).

stuartmorgan (changes requested):
I did another pass through the non-C++ parts; it looks like the C++ code still has some open comments from the last round so I'll wait on those.

packages/local_auth/local_auth_windows/test/local_auth_test.dart
azchohfi: Fixed.

packages/local_auth/local_auth_windows/pubspec.yaml
azchohfi: Fixed.

packages/local_auth/local_auth_windows/lib/types/auth_messages_windows.dart
azchohfi: Fixed.

packages/local_auth/local_auth_windows/example/README.md
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth.h
azchohfi: Done.

packages/local_auth/local_auth_windows/windows/local_auth.h
azchohfi: Done.

azchohfi:
@stuartmorgan I think I fixed it all. Let me know if there is anything I missed.

packages/local_auth/local_auth_windows/windows/local_auth.h
```diff
@@ -47,7 +54,12 @@ class LocalAuthPlugin : public flutter::Plugin {
  public:
   static void RegisterWithRegistrar(flutter::PluginRegistrarWindows* registrar);
 
+  // Real implementation that creates a plugin instance that will create the
+  // dialog and associate it with the HWND returned from the provided function
```
cbracken, line 58: Pedantic nit: end with a period.

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#:~:text=Each%20line%20should%20end%20with%20a%20period.

packages/local_auth/local_auth_windows/windows/local_auth.h
```diff
@@ -60,11 +72,16 @@ class LocalAuthPlugin : public flutter::Plugin {
  private:
   std::unique_ptr<UserConsentVerifier> user_consent_verifier_;
 
+  // Starts authentication process
```
cbracken, line 75: Also here.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -85,6 +89,7 @@ class UserConsentVerifierImpl : public UserConsentVerifier {
     return consent_result;
   }
 
+  // Calls the native Windows API to check for the Windows Hello availability
```
cbracken, line 92: And here.

packages/local_auth/local_auth_windows/windows/local_auth.h
```diff
@@ -0,0 +1,89 @@
...
+  // Disallow copy and move.
+  UserConsentVerifier(const UserConsentVerifier&) = delete;
+  UserConsentVerifier& operator=(const UserConsentVerifier&) = delete;
+};
+
+class LocalAuthPlugin : public flutter::Plugin {
+ public:
+  static void RegisterWithRegistrar(flutter::PluginRegistrarWindows* registrar);
+
+  // Real implementation that creates a plugin instance that will create the
```
cbracken, line 57: nit: I'd just start with "Creates a plugin instance ..." since if it creates an instance it's non-abstract and thus real.

packages/local_auth/local_auth_windows/windows/local_auth.h
```diff
@@ -0,0 +1,89 @@
...
+  // dialog and associate it with the HWND returned from the provided function
+  LocalAuthPlugin(std::function<HWND()> window_provider);
+
+  // Creates a plugin instance with the given UserConsentVerifier instance.
+  // Exists for unit testing with mock implementations.
+  LocalAuthPlugin(std::unique_ptr<UserConsentVerifier> user_consent_verifier);
+
+  // Called when a method is called on this plugin's channel from Dart.
```
cbracken, line 65: Maybe "Handles method calls from Dart on this plugin's channel."

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
```diff
@@ -0,0 +1,236 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+#include <winstring.h>
+
+#include "local_auth.h"
+
+namespace {
+
+template <typename T>
+// Helper method for getting an argument from an EncodableValue
```
cbracken, line 11: Pedantic nit: here and a few more places below. End with a period.

cbracken (approved):
Overall LGTM on the latest pass modulo a few pedantic nits around the doc comments.

packages/local_auth/local_auth_windows/windows/local_auth.h
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth.h
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth.h
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth.h
azchohfi: Fixed.

packages/local_auth/local_auth_windows/windows/local_auth_plugin.cpp
azchohfi: Fixed.

azchohfi:
@cbracken fixed them all. I wish the flutter_plugin_tools' format command fixed this.

cbracken:
It would definitely be nice -- or at least try to call them out. The tricky part is block comments in comments and ASCII art. This is also why for example the `dart format` command won't rewrap comments.

stuartmorgan:
> I wish the flutter_plugin_tools' format command fixed this.

FYI for C++ it's just running clang-format. The tool isn't a formatter, it's just a wrapper to call the right formatter on the right files.

azchohfi:
Anything else needed? I've merged the main branch again just to keep it up-to-date.

stuartmorgan:
It's been published now, so is ready for an endorsement PR :)

azchohfi:
@stuartmorgan #5776

PR #4809, opened 2022-02-11
[webview_flutter_wkwebview] Add support for `WKUIDelegate`

bparrishMines:
Add support for `WKUIDelegate`.

No version change:
Part of https://github.com/flutter/flutter/issues/93732 and doesn't make any changes to the current implementation.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit_webview_widget.dart
```diff
@@ -87,6 +87,16 @@ class WebKitWebViewPlatformController extends WebViewPlatformController {
...
+    webView.uiDelegate = uiDelegate;
+    uiDelegate.setOnCreateWebView((
+      WKWebViewConfiguration configuration,
+      WKNavigationAction navigationAction,
+    ) {
+      if (!navigationAction.targetFrame.isMainFrame) {
+        webView.loadRequest(navigationAction.request);
+      }
+    });
   }
```
bparrishMines, line 100: Relevant code from current implementation:
https://github.com/flutter/plugins/blob/main/packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m#L699
https://github.com/flutter/plugins/blob/main/packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FlutterWebView.m#L121

packages/webview_flutter/webview_flutter_wkwebview/lib/src/foundation/foundation.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:typed_data';
+
+/// A URL load request that is independent of protocol or URL scheme.
+///
+/// Wraps [NSUrlRequest](https://developer.apple.com/documentation/foundation/nsurlrequest?language=objc).
+class NSUrlRequest {
```
stuartmorgan, line 10: @immutable

packages/webview_flutter/webview_flutter_wkwebview/lib/src/foundation/foundation.dart
```diff
@@ -0,0 +1,32 @@
...
+  /// Constructs an [NSUrlRequest].
+  NSUrlRequest({
```
stuartmorgan, line 12: const

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -42,6 +44,34 @@ enum WKAudiovisualMediaType {
   all,
 }
 
+/// An object that contains information about an action that causes navigation to occur.
+///
+/// Wraps [WKNavigationAction](https://developer.apple.com/documentation/webkit/wknavigationaction?language=objc).
+class WKNavigationAction {
```
stuartmorgan, line 50: @immutable

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -42,6 +44,34 @@ enum WKAudiovisualMediaType {
...
+  /// Constructs a [WKNavigationAction].
+  WKNavigationAction({required this.request, required this.targetFrame});
```
stuartmorgan, line 52: const

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -42,6 +44,34 @@ enum WKAudiovisualMediaType {
...
+  /// Construct a [WKFrameInfo].
+  WKFrameInfo({required this.isMainFrame});
+
+  /// Indicates whether the frame is the web site's main frame or a subframe.
+  final bool isMainFrame;
+}
+
 /// A script that the web view injects into a webpage.
 ///
 /// Wraps [WKUserScript](https://developer.apple.com/documentation/webkit/wkuserscript?language=objc).
```
stuartmorgan, line 77: Could you also fix this in the two classes just below?

I've been hitting this a lot in doing analysis option updates, and it's hard to add later if it ends up public (which shouldn't happen for these, but better safe than sorry. E.g., in case we do decide to someday vend this layer as a plugin.)

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -202,6 +232,22 @@ class WKWebViewConfiguration {
   }
 }
 
+/// The methods for presenting native user interface elements on behalf of a webpage.
+///
+/// Wraps [WKUIDelegate](https://developer.apple.com/documentation/webkit/wkuidelegate?language=objc).
+class WKUIDelegate {
+  /// Indicates a new [WebView] was requested to be created with [configuration].
+  void setOnCreateWebView(
```
stuartmorgan, line 240: Why is this a function named `setFoo` rather than a `set foo`? (This may apply to code from previous PRs and I just didn't notice it then.)

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit_webview_widget.dart
stuartmorgan: Weird. I'm surprised nobody has ever filed a bug about this behavior, because it's pretty surprising.

packages/webview_flutter/webview_flutter_wkwebview/test/src/web_kit_webview_widget_test.dart
```diff
@@ -84,6 +89,26 @@ void main() {
       await buildWidget(tester);
     });
 
+    testWidgets('WKUIDelegate.onCreateWebView loadRequest',
```
stuartmorgan, line 92: Can we name this something more descriptive of the behavior it's testing, like 'requests to open a new window instead open inline'?

stuartmorgan (approved):
LGTM with nits

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
bparrishMines: That's my mistake. It was supposed to have a `Future<void>`. But, I suppose using `set foo` IS more Darty. But it also would be an asynchronous function so I'm not sure which is better. I can't think of an instance where it would need to be awaited though.

PR #4838, opened 2022-02-14
[image_picker] Update app-facing and web analysis options

stuartmorgan:
Switches from legacy analysis options to current analysis options,
fixing all analysis issues it exposed.

Also opportunistically fixes one potential null safety issue highlighted by the warnings.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
+++ b/packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
@@ -18,14 +18,6 @@ const String _kAcceptVideoMimeType = 'video/3gpp,video/x-m4v,video/mp4,video/*';
 ///
 /// This class implements the `package:image_picker` functionality for the web.
 class ImagePickerPlugin extends ImagePickerPlatform {
-  final ImagePickerPluginTestOverrides? _overrides;
-
-  bool get _hasOverrides => _overrides != null;
-
-  late html.Element _target;
-
-  late ImageResizer _imageResizer;
-
   /// A constructor that allows tests to override the function that creates file inputs.
   ImagePickerPlugin({
     @visibleForTesting ImagePickerPluginTestOverrides? overrides,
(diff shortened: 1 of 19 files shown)


stuartmorgan:
The last commit here is an actual behavioral change. The warnings highlighted that a helper method that was typed to return a nullable value was internally force-unwrapping a nullable value instead of passing that null through, so I fixed it instead of adding an explicit force-unwrap under the assumption that it was a mistake.

ditman:
Currently looking at this

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
```diff
@@ -232,24 +236,24 @@ class ImagePickerPlugin extends ImagePickerPlatform {
   /// Handles the OnChange event from a FileUploadInputElement object
   /// Returns a list of selected files.
   List<html.File>? _handleOnChangeEvent(html.Event event) {
-    final html.FileUploadInputElement input =
-        event.target as html.FileUploadInputElement;
-    return _getFilesFromInput(input);
+    final html.FileUploadInputElement? input =
+        event.target as html.FileUploadInputElement?;
+    return input == null ? null : _getFilesFromInput(input);
```
ditman, line 241: I'm not aware that input here can be null (html.Event has a nullable target because you can create your own custom events without setting it up, but AFAIK the DOM Events should always contain a target); another acceptable solution would have been:

```dart
return _getFilesFromInput(input!);
```

ditman (approved):
Hit it! Thanks for the cleanup!

packages/image_picker/image_picker_for_web/lib/image_picker_for_web.dart
stuartmorgan: I did this based on the return type being nullable. Maybe that was wrong too? But all the callers already have handling for it, so if this should be force-unwrapped there's cleanup that should happen throughout this file.

ditman: Nope, this is good. The return type is nullable because the behavior in general is nullable (except in this particular case (and maybe others) where the event comes from the browser, and it will always fill the `target`).

PR #4859, opened 2022-02-15
[webview_flutter] Fix debuggingEnabled on Android

stuartmorgan:
The recent Dart rewrite accidentally dropped a parameter value in the
call chain, causing disabling debugging to enabled it instead.

This fixes the issue, and adds testing of that portion of the call
chain (which was entirely missing), as well as adding testing of `false`
at various other points just for added coverage.

As opportunistic cleanup, changes the test group naming, as the pattern
of using `'$FooClass'` breaks test result display in VSCode.

Fixes https://github.com/flutter/flutter/issues/98521

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/webview_flutter_android/CHANGELOG.md
+++ b/packages/webview_flutter/webview_flutter_android/CHANGELOG.md
@@ -1,5 +1,6 @@
-## NEXT
+## 2.8.3
 
+* Fixes a bug causing `debuggingEnabled` to always be set to true.
 * Fixes an integration test race condition.
 
 ## 2.8.2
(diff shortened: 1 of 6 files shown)


stuartmorgan:
We should look into re-running code coverage analysis of this, and (before switching to it) running it on the new iOS implementation, and look for gaps like this class being completely untested. The downside of this new structure is that there are more layers where a bug like this can sneak in, and coverage analysis will help catch things like this in advance.

(Although in this case all the tests of the rest of the chain of calls also only tested `true`, so coverage wouldn't have found that anyway. But at least it would have flagged the need to have tests of this object.)

bparrishMines (approved):
LGTM

PR #4863, opened 2022-02-15
[image_picker] Added image_picker_for_windows.

azchohfi:
This PR adds a very simplistic Windows implementation of the image_picker plugin, which uses the file_selector plugin to get images or videos from the "gallery". This initial implementation doesn't yet support ImageSource.camera (falls back to gallery).

Fixes https://github.com/flutter/flutter/issues/92108

## Pre-launch Checklist

- [X] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [X] I signed the [CLA].
- [X] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [X] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].


packages/image_picker/image_picker_for_windows/AUTHORS
```diff
@@ -0,0 +1,67 @@
...
+Théo Champion <contact.theochampion@gmail.com>
+Kazuki Yamaguchi <y.kazuki0614n@gmail.com>
+Eitan Schwartz <eshvartz@gmail.com>
+Chris Rutkowski <chrisrutkowski89@gmail.com>
+Juan Alvarez <juan.alvarez@resideo.com>
+Aleksandr Yurkovskiy <sanekyy@gmail.com>
+Anton Borries <mail@antonborri.es>
+Alex Li <google@alexv525.com>
+Rahul Raj <64.rahulraj@gmail.com>
+Balvinder Singh Gambhir <balvindersi2@gmail.com>
```
stuartmorgan, line 67: The authors list for a new plugin should only contain Google (the first entry here) and the plugin author (if different).

packages/image_picker/image_picker_for_windows/README.md
```diff
@@ -0,0 +1,36 @@
+# image\_picker\_for\_windows
```
stuartmorgan, line 1: We should call it `image_picker_windows`, as that's our standard naming. The "_for_web" in some packages was to work around the fact that in some cases a third party published something under the "_web" name before we added web support.

packages/image_picker/image_picker_for_windows/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Usage
+
+### Import the package
+
+This package is not yet [endorsed](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#endorsed-federated-plugin), which means you need to add 
+not only the `image_picker`, as well as the `image_picker_for_windows`.
+
+### Use the plugin
+
+You should be able to use `package:image_picker` _almost_ as normal.
```
stuartmorgan, line 20: What's the difference here relative to normal `image_picker` code?

packages/image_picker/image_picker_for_windows/example/lib/main.dart
```diff
@@ -0,0 +1,465 @@
...
+// found in the LICENSE file.
+
+// ignore_for_file: public_member_api_docs
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:image_picker/image_picker.dart';
```
stuartmorgan, line 12: We don't use the app-facing platform in implementation examples (there are some legacy violations, but we don't do it any more), as it creates a circular dependency that causes problems for testing changes. This should instead be written directly against `image_picker_platform_interface`.

packages/image_picker/image_picker_for_windows/example/lib/main.dart
```diff
@@ -0,0 +1,465 @@
...
+      } else {
+        controller = VideoPlayerController.file(File(file.path));
+      }
+      _controller = controller;
+      // In web, most browsers won't honor a programmatic call to .play
+      // if the video has a sound track (and is not muted).
+      // Mute the video so it auto-plays in web!
+      // This is not needed if the call to .play is the result of user
+      // interaction (clicking on a "play" button, for example).
+      final double volume = kIsWeb ? 0.0 : 1.0;
```
stuartmorgan, line 72: No need for web-specific code in the Windows example.

packages/image_picker/image_picker_for_windows/example/lib/main.dart
```diff
@@ -0,0 +1,465 @@
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(widget.title!),
+      ),
+      body: Center(
+        child: !kIsWeb && defaultTargetPlatform == TargetPlatform.android
```
stuartmorgan, line 244: Or Android.

packages/image_picker/image_picker_for_windows/lib/image_picker_for_windows.dart
```diff
@@ -0,0 +1,143 @@
...
+import 'dart:async';
+
+import 'package:image_picker_platform_interface/image_picker_platform_interface.dart';
+import 'package:file_selector_platform_interface/file_selector_platform_interface.dart';
+
+/// The Windows implementation of [ImagePickerPlatform].
+///
+/// This class implements the `package:image_picker` functionality for Windows.
+class ImagePickerWindowsPlugin extends ImagePickerPlatform {
+  /// A constructor that allows tests to override the function that creates file inputs.
```
stuartmorgan, line 14: This comment seems wrong?

packages/image_picker/image_picker_for_windows/lib/image_picker_for_windows.dart
```diff
@@ -0,0 +1,143 @@
...
+  ImagePickerWindowsPlugin();
+
+  List<String> _imageFormats = ['jpg', 'jpeg', 'png', 'bmp'];
```
stuartmorgan, line 17: We should add at least webp, gif, tiff (and tif), and maybe apng.

packages/image_picker/image_picker_for_windows/lib/image_picker_for_windows.dart
```diff
@@ -0,0 +1,143 @@
...
+  List<String> _videoFormats = ['mov', 'wmv', 'mkv', 'mp4'];
```
stuartmorgan, line 18: webm, avi, mpeg, mpg

packages/image_picker/image_picker_for_windows/lib/image_picker_for_windows.dart
```diff
@@ -0,0 +1,143 @@
...
+
+import 'dart:async';
+
+import 'package:image_picker_platform_interface/image_picker_platform_interface.dart';
+import 'package:file_selector_platform_interface/file_selector_platform_interface.dart';
+
+/// The Windows implementation of [ImagePickerPlatform].
+///
+/// This class implements the `package:image_picker` functionality for Windows.
+class ImagePickerWindowsPlugin extends ImagePickerPlatform {
```
stuartmorgan, line 13: We generally don't include a `Plugin` suffix here.

packages/image_picker/image_picker_for_windows/lib/image_picker_for_windows.dart
```diff
@@ -0,0 +1,143 @@
...
+  /// The `source` argument controls where the image comes from. This can
+  /// be either [ImageSource.camera] or [ImageSource.gallery].
+  ///
+  /// Note that the `maxWidth`, `maxHeight` and `imageQuality` arguments are not supported on the Windows. If any of these arguments is supplied, it'll be silently ignored by the Windows version of the plugin.
+  ///
+  /// Use `preferredCameraDevice` to specify the camera to use when the `source` is [ImageSource.camera].
+  /// The `preferredCameraDevice` is ignored when `source` is [ImageSource.gallery]. It is also ignored if the chosen camera is not supported on the device.
+  /// Defaults to [CameraDevice.rear].
+  ///
+  /// If no images were picked, the return value is null.
```
stuartmorgan, line 36: We generally don't duplicate declaration comments on overriden methods.

I would just put a non-documentation comment (`//`) with the note about the arguments that are currently ignored.

packages/image_picker/image_picker_for_windows/lib/image_picker_for_windows.dart
```diff
@@ -0,0 +1,143 @@
...
+  Future<PickedFile?> pickImage({
+    required ImageSource source,
+    double? maxWidth,
+    double? maxHeight,
+    int? imageQuality,
+    CameraDevice preferredCameraDevice = CameraDevice.rear,
+  }) async {
+    final typeGroup = XTypeGroup(label: 'images', extensions: _imageFormats);
+    final file = await FileSelectorPlatform.instance
+        .openFile(acceptedTypeGroups: [typeGroup]);
```
stuartmorgan, line 47: Prefer implementing this by calling `getImage` so we don't have two copies of the implementation.

packages/image_picker/image_picker_for_windows/lib/image_picker_for_windows.dart
```diff
@@ -0,0 +1,143 @@
...
+  /// If no images were picked, the return value is null.
+  @override
+  Future<PickedFile?> pickVideo({
+    required ImageSource source,
+    CameraDevice preferredCameraDevice = CameraDevice.rear,
+    Duration? maxDuration,
+  }) async {
+    final typeGroup = XTypeGroup(label: 'videos', extensions: _videoFormats);
+    final file = await FileSelectorPlatform.instance
+        .openFile(acceptedTypeGroups: [typeGroup]);
```
stuartmorgan, line 74: Same.

packages/image_picker/image_picker_for_windows/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
+name: image_picker_for_windows
+description: Windows platform implementation of image_picker
+repository: https://github.com/flutter/plugins/tree/main/packages/image_picker/image_picker_for_windows
+issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+image_picker%22
+version: 0.0.1
```
stuartmorgan, line 5: Nit: Let's make it 0.1.0; we usually start with a minor version.

packages/image_picker/image_picker_for_windows/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+environment:
+  sdk: ">=2.12.0 <3.0.0"
+  flutter: ">=2.0.0"
```
stuartmorgan, line 9: 2.5 (required for Dart registration)

packages/image_picker/image_picker_for_windows/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
...
+    implements: image_picker
+    platforms:
+      windows:
+        dartPluginClass: ImagePickerWindowsPlugin
+
+dependencies:
+  flutter:
+    sdk: flutter
+  image_picker_platform_interface: ^2.4.3
+  file_selector: ^0.8.3
```
stuartmorgan, line 22: This should be `file_selector_platform_interface`; we don't want to pull in the native `file_selector` implementations for other platforms and bloat people's apps unnecessarily.

packages/image_picker/image_picker_for_windows/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
...
+  file_selector_windows: ^0.8.2
+  pedantic: ^1.10.0
```
stuartmorgan, line 24: This isn't necessary.

stuartmorgan (changes requested):
Looks good! This will need Dart unit tests that sets a mock `FileSelectorPlatform` implementation as `FileSelectorPlatform.instance` and validates that the expected calls are made. And also that `registerWith` works (as in other Dart plugin implementation tests).

While the logic is currently pretty trivial, having tests in place is important as we change it over time.

packages/image_picker/image_picker_for_windows/README.md
azchohfi: I just wanted to say that it is not fully featured yet. I'll change comment to make it clearer.

packages/image_picker/image_picker_windows/README.md
```diff
@@ -0,0 +1,36 @@
...
+### Import the package
+
+This package is not yet [endorsed](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#endorsed-federated-plugin), which means you need to add 
+not only the `image_picker`, as well as the `image_picker_windows`.
+
+### Use the plugin
+
+You should be able to use `package:image_picker` _almost_ as normal, since this package is not fully featured yet.
+
+If you want to use the path directly, your code would need look like this:
```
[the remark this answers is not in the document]
azchohfi: Agreed. I can just remove the "Use the plugin" section.

packages/image_picker/image_picker_windows/example/lib/main.dart
```diff
@@ -0,0 +1,405 @@
...
+  }
+
+  dynamic _pickImageError;
+  bool isVideo = false;
+
+  VideoPlayerController? _controller;
+  VideoPlayerController? _toBeDisposed;
+  String? _retrieveDataError;
+
+  final ImagePickerPlatform _picker = ImagePickerWindows();
```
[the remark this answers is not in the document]
azchohfi: I've tried doing this, but then I get an exception about the file_selector plugin being missing, during runtime. It seems like the FIleSelectorWindows plugin is not registering itself. If I add a:
```dart
FileSelectorWindows.registerWith();
```
at ImagePickerWindows.registerWith() method, then everything works, but it feels like wrong. Am I missing something here?

azchohfi:
Oh, I believe I need to add the image_picker_windows to the exclude_integration_win32.yaml file.

azchohfi:
@stuartmorgan anything else needed here?

stuartmorgan:
Sorry for the slow response, I haven't had much time for reviews because of the build issue with VS 17.1.0.

packages/image_picker/image_picker_windows/example/lib/main.dart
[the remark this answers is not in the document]
stuartmorgan: There may well be a bug with Dart registration for the edge case of depending directly on a federated implementation (without the app-facing package). I'll investigate tomorrow or Friday.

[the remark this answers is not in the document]
stuartmorgan: Okay, I know why this happens now, but I need to think about it more to figure out if it's actually a bug. There are some weird subtleties.

For the purposes of this PR, let's assume it's not a bug. In which case, the way to handle it is that you shouldn't be using `FileSelectorPlatform` anywhere in your implementation, but instead use a `FileSelectorWindows` instance directly.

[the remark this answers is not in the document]
stuartmorgan: (Filed https://github.com/flutter/flutter/issues/99083 for the reason this happens)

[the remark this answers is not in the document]
azchohfi: Should I just create my own FileSelectorWindows instance, or should I try to get the FileSelectorPlatform.instance first? A very low overhead since the MethodChannel itself is already a const outside the instance anyway, but I wonder how much this can impact other plugins that follow the same pattern.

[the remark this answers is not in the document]
stuartmorgan: You can just make one; it should be minimal cost.

> I wonder how much this can impact other plugins that follow the same pattern

Yes, longer term I do think we'll need to fix the underlying issue. (I don't expect this to be a super common pattern though.)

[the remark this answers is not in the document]
azchohfi: Ok, tests are passing, so this should be good now.

packages/image_picker/image_picker_windows/README.md
```diff
@@ -0,0 +1,16 @@
+# image\_picker\_windows
+
+A Windows implementation of [`image_picker`][1].
+
```
stuartmorgan, line 4: This should mention that image source is not currently supported.

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
```diff
@@ -38,10 +38,11 @@ class ImagePickerWindows extends ImagePickerPlatform {
   /// Constructs a ImagePickerWindows.
   ImagePickerWindows();
 
+  static FileSelectorWindows get _fileSelectorInstance => FileSelectorWindows();
```
stuartmorgan, line 41: This doesn't do what I expect you intended; this is basically just an alias for the constructor. Assuming you wanted a shared instance you want:
```
static late FileSelectorWindows = FileSelectorWindows();
```

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
```diff
@@ -0,0 +1,142 @@
...
+  ImagePickerWindows();
+
+  static FileSelectorWindows get _fileSelectorInstance => FileSelectorWindows();
+
+  /// Registers this class as the default instance of [ImagePickerPlatform].
+  static void registerWith() {
+    ImagePickerPlatform.instance = ImagePickerWindows();
+  }
+
+  // Note that the `maxWidth`, `maxHeight` and `imageQuality` arguments are
```
stuartmorgan, line 48: This list is missing `source`

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
```diff
@@ -0,0 +1,142 @@
...
+        maxDuration: maxDuration);
+    if (file != null) {
+      return PickedFile(file.path);
+    }
+    return null;
+  }
+
+  // Note that the `source`, `maxWidth`, `maxHeight`, `imageQuality`, and
+  // `preferredCameraDevice` arguments are not supported on Windows. If
+  // any of these arguments is supplied, it'll be silently ignored by the
```
stuartmorgan, line 93: For `source`, since ignoring it will do completely the wrong thing (as opposed to the other arguments which are less critical), I think you should throw an `UnimplementedError` when it's not the gallery.

(Ideally we would delegate to `camera_windows`, but without the camera controller that would be quite tricky. Longer term we might have to consider some refactoring there to share code.)

stuartmorgan (changes requested):
A few other small things, but otherwise it should be good from my side.

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
azchohfi: Oh, my bad. I want a simple singleton, and I don't see many reasons to late initialize this. I believe this is fine:
```dart
  static final FileSelectorWindows _fileSelectorInstance = FileSelectorWindows();
```

azchohfi:
@stuartmorgan I think I got it all, let me know if there is anything else needed.

packages/image_picker/image_picker_windows/README.md
azchohfi: Done.

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
azchohfi: Done.

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
azchohfi: I've added the UnimplementedError.

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
```diff
@@ -0,0 +1,156 @@
...
+  Future<XFile?> getImage({
+    required ImageSource source,
+    double? maxWidth,
+    double? maxHeight,
+    int? imageQuality,
+    CameraDevice preferredCameraDevice = CameraDevice.rear,
+  }) async {
+    if (source != ImageSource.gallery) {
+      throw UnimplementedError(
+          'source is only implemented for `ImageSource.gallery` on Windows');
```
stuartmorgan, line 110: Nit (here and below): How about "ImageSource.gallery is currently the only supported source on Windows"

packages/image_picker/image_picker_windows/test/image_picker_windows_test.dart
```diff
@@ -0,0 +1,187 @@
...
+              'tiff',
+              'apng'
+            ]);
+
+        await plugin.pickImage(source: ImageSource.gallery);
+
+        expect(
+          log,
+          <Matcher>[
+            isMethodCall('openFile', arguments: <String, dynamic>{
```
stuartmorgan, line 57: You are testing implementation details of `file_selector` here; these tests would all break if FileSelectorWindows changed its implementation (e.g., switched to a Dart implementation, or to using Pigeon).

Even though you can't use `FileSelectorPlatform.instance` in the implementation, you can still write the implementation against a `FileSelectorPlatform` variable, and just manually instantiate it with a `FileSelectorWindows`. Then you can make that instance settable by tests, and insert a mock implementation in these tests and validate against that mock. Then it doesn't matter how `FileSelectorWindows` works, because you'll just be verifying that you're using its API correctly.

stuartmorgan (changes requested):

packages/image_picker/image_picker_windows/test/image_picker_windows_test.dart
azchohfi: @stuartmorgan I just pushed the changes. I believe that is what you means, which I agree that is a way better approach.

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
azchohfi: Done.

azchohfi:
@stuartmorgan anything needed here?

packages/image_picker/image_picker_windows/test/image_picker_windows_test.dart
```diff
@@ -0,0 +1,104 @@
...
+
+        expect(
+            verify(mockFileSelectorPlatform.openFile(
+                    acceptedTypeGroups: captureAnyNamed('acceptedTypeGroups')))
+                .captured
+                .single[0]
+                .extensions,
+            ImagePickerWindows.videoFormats);
+      });
+    });
```
stuartmorgan, line 102: It would be good to add tests that passing `camera` as a source throws `UnimplementedError` for all the relevant methods.

stuartmorgan (approved):
One comment about adding a few more tests, but LGTM otherwise. The main thing this is waiting on is just the second review from @cbracken.

packages/image_picker/image_picker_windows/test/image_picker_windows_test.dart
azchohfi: Done.

azchohfi:
@cbracken anything else needed here?

azchohfi:
Merged main again, just to keep this updated. Unlike the local_auth plugin, the image_picker plugin is already federated. @stuartmorgan / @cbracken what is keeping this PR from being reviewed/merged?

stuartmorgan:
> @stuartmorgan / @cbracken what is keeping this PR from being reviewed/merged?

It's still @cbracken having time to review; Flutter policy requires two reviews for PRs from people outside the org before landing.

packages/image_picker/image_picker_windows/example/lib/main.dart
```diff
@@ -0,0 +1,404 @@
...
+
+class AspectRatioVideoState extends State<AspectRatioVideo> {
+  VideoPlayerController? get controller => widget.controller;
+  bool initialized = false;
+
+  void _onVideoControllerUpdate() {
+    if (!mounted) {
+      return;
+    }
+    if (initialized != controller!.value.isInitialized) {
```
cbracken, line 373: Here and below: `controller` is nullable; is it safe to assume it's non-null in all these cases? If so, consider making it non-nullable which pushes the null-check up into `_previewVideo`, so no one needs to reason about nullability here.

packages/image_picker/image_picker_windows/example/lib/main.dart
```diff
@@ -0,0 +1,404 @@
...
+      );
+    }
+  }
+
+  Widget _handlePreview() {
+    if (isVideo) {
+      return _previewVideo();
+    } else {
+      return _previewImages();
+    }
```
cbracken, line 199: Optionally simplify down to:
```dart
Widget _handlePreview() => isVideo ? _previewVideo() : _previewImages();
```

packages/image_picker/image_picker_windows/example/lib/main.dart
```diff
@@ -0,0 +1,404 @@
...
+  bool isVideo = false;
+
+  VideoPlayerController? _controller;
+  VideoPlayerController? _toBeDisposed;
+  String? _retrieveDataError;
+
+  final ImagePickerPlatform _picker = ImagePickerPlatform.instance;
+  final TextEditingController maxWidthController = TextEditingController();
+  final TextEditingController maxHeightController = TextEditingController();
+  final TextEditingController qualityController = TextEditingController();
```
cbracken, line 55: Consider making these private.

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
```diff
@@ -0,0 +1,163 @@
...
+
+/// The Windows implementation of [ImagePickerPlatform].
+///
+/// This class implements the `package:image_picker` functionality for
+/// Windows.
+class ImagePickerWindows extends ImagePickerPlatform {
+  /// Constructs a ImagePickerWindows.
+  ImagePickerWindows();
+
+  /// List of image extensions used when picking images
```
cbracken, line 20: nit: here and below, add trailing periods on doc comments.

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
```diff
@@ -0,0 +1,163 @@
...
+    'mp4',
+    'webm',
+    'avi',
+    'mpeg',
+    'mpg'
+  ];
+
+  /// The FileSelectorPlatform used by the ImagePickerWindows.
+  @visibleForTesting
+  static late FileSelectorPlatform fileSelectorInstance = FileSelectorWindows();
```
cbracken, line 49: Since this is never reassigned, you can make this `static late final`.

cbracken, line 49: Consider dropping `Instance` from the name; it doesn't add any information, since all variables are instances of some type. Plus it might save a line-wrap some day :)

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
```diff
@@ -0,0 +1,163 @@
...
+    'wmv',
+    'mkv',
+    'mp4',
+    'webm',
+    'avi',
+    'mpeg',
+    'mpg'
+  ];
+
+  /// The FileSelectorPlatform used by the ImagePickerWindows.
```
cbracken, line 47: This comment doesn't provide much info that isn't in the declaration itself. Consider documenting how this is used. Even something super short like:
```dart
/// The file selector used to prompt the user to select images or videos.
```

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#documentation-dartdocs-javadocs-etc

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
```diff
@@ -0,0 +1,163 @@
...
+  @visibleForTesting
+  static late FileSelectorPlatform fileSelectorInstance = FileSelectorWindows();
+
+  /// Registers this class as the default instance of [ImagePickerPlatform].
+  static void registerWith() {
+    ImagePickerPlatform.instance = ImagePickerWindows();
+  }
+
+  // Note that the `maxWidth`, `maxHeight`, `imageQuality`
```
cbracken, line 56: Here and below: Drop the 'Note that the'.

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-empty-prose

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
```diff
@@ -0,0 +1,163 @@
...
+  Future<XFile?> getImage({
+    required ImageSource source,
+    double? maxWidth,
+    double? maxHeight,
+    int? imageQuality,
+    CameraDevice preferredCameraDevice = CameraDevice.rear,
+  }) async {
+    if (source != ImageSource.gallery) {
+      throw UnimplementedError(
+          'ImageSource.gallery is currently the only supported source on Windows');
```
cbracken, line 117: Is the intent to add other image sources in the future? If so, please file an issue to track this, add a `TODO` here and link the issue.

```
// TODO(username): Support ImageSource.foo, ImageSource.bar.
//                 See http://link/to/a/bug/123
```

If we will never support other image sources, throw `UnsupportedError` instead.

cbracken (changes requested):
Overall looks good! Just a few comments.

packages/image_picker/image_picker_windows/example/lib/main.dart
```diff
@@ -0,0 +1,404 @@
...
+    maxHeightController.dispose();
+    qualityController.dispose();
+    super.dispose();
+  }
+
+  Future<void> _disposeVideoController() async {
+    if (_toBeDisposed != null) {
+      await _toBeDisposed!.dispose();
+    }
+    _toBeDisposed = _controller;
```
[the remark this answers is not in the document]
azchohfi: I honestly don't know. I copied this example app from the other image_picker samples that are already in this repository. I'm not the original developer of it.

packages/image_picker/image_picker_windows/lib/image_picker_windows.dart
azchohfi: Can't be final since the tests need to replace this instance.

azchohfi:
@cbracken let me know what you think.

packages/image_picker/image_picker_windows/example/lib/main.dart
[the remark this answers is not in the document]
cbracken: Ah makes sense. Ignore then, and we can do a separate pass over the lot of them once this has landed. Thanks for clarifying.

